PR #6, opened 2016-08-20
MINFI-85: Add ListenSyslog processor

benqiu2016:


apiri:
will review

apiri:
Was able to test this a bit and it looks pretty good.  Will make some additions of including the AS license on the header of the provided config file and renaming ListenSysLog.cpp to ListenSyslog for consistency with the other files and class names.

PR #9, opened 2016-08-25
Minifi 87: add start/stop script

benqiu2016:
add MiNiFi start/stop script


apiri:
This also had the the syslog items in it.  Will incorporate this with the efforts of #8 / MINIFI-63 and merge though.  Thanks for contributing!

PR #10, opened 2016-08-29
MINIFI-97 minifi.sh stop fix

pvillard31:


apiri:
reviewing

apiri:
Looks good, thanks for finding and patching up.

Will merge in shortly.

PR #14, opened 2016-09-26
MINIFI-34 Establishing CMake build system to provide build functional…

apiri:
MINIFI-34 Establishing CMake build system to provide build functionality equivalent to pre-existing Makefile.


trixpan:
@apiri another note: Seems like CMake is a must before we get Windows build in shape:

According to [this](http://help.appveyor.com/discussions/problems/1467-building-with-makefile) and [this](http://stackoverflow.com/questions/12881854/how-to-use-gnu-make-on-windows), using gnu make on appveyor is sort of a funny endeavour.

apiri:
Yep, sounds fair.  From the reading and research I've done, CMake certainly feels like it makes things possible.  Travis is being a little odd with CMake.  Will work to flush out those issues so that those changes are also included with this PR.

PR #18, opened 2016-10-07
MINIFI-34 - attempt to progress the CMake environment

trixpan:


trixpan:
@apiri we are very close. 

The current issue now is the use of `INTERFACE `(libminifi). This feature is only available when running CMake 3 (not available on RHEL6 and also not on Ubuntu LTS).

Also, you will see a few adjustments around `VERSION`.

 The CMake policy required for that to work is not available on CMake 2.x (again, RHEL6 & Ubuntu LTS), reason why we saw [this](https://travis-ci.org/apache/nifi-minifi-cpp/builds/165648184#L629):

I left the code as a condition but I reckon we will have to get rid of it in the long run in order to achieve better compatibility.

apiri:
Hmm, okay.  Good information and agree on supporting the listed environments.  I think I can work around the INTERFACE usage.

trixpan:
@apiri indeed. solving the INTERFACE was a matter of using what seems to be called the "infamous `dummy.cpp `".

Good news is that this should be working now (it works locally).

once you confirm travis-ci precise is working, I will be happy to proceed with enabling trusty tahr again and setup a docker based test to ease cross platform development.

Major changes of the last commit

- the CMake has been adjusted to compile against existing libuuid
- travis adjusted to use whatever version of libboost-all-dev is available

trixpan:
ok. So precise(12.04) does not work out of the box, only with add on packages. 

I think we should push this with 14.04 (travis' `dist: tahr`) and add 12.04 as a docker container for testing. This should also help documenting the required compilation steps on 12.04

apiri:
Sounds fair to me.  Indeed we have some obstacles to conquer handling all the varied environments but as you mentioned, the lessons learned should be a great help to folks who wish to build on their machine.

trixpan:
@apiri should we merge this?

apiri:
Yeah, seems like we have a good start.  Will do some testing around on a few systems and get it merged in today.

PR #21, opened 2016-10-20
MINIFI-124 The active_pid() function kill -s 0 <PID> returns 0 on success (PID i…

codingogre:
MINIFI-124.  In the active_pid() function kill -s 0 <PID> returns 0 on success (the PID is currently running), yet the if statement in the stop case statement is checking for 0 as a failure condition.


apiri:
reviewing

apiri:
Looks good here and thanks for catching!  Will merge.

PR #22, opened 2016-10-20
MINIFI-123 Pthreads linker fix

codingogre:
MINIFI-123.  On the Raspbian platform (Debian based) the linker is unsuccessful in linking pthreads with the MiNiFi executable because CMake does not specify -lpthreads.  The following change should create a platform agnostic way of making sure posix threads are linked correctly.  This change should be compatible with CMake 2.8.x and above.


apiri:
reviewing

apiri:
Was able to setup a similar environment to test and verified the fix!  Thanks for fixing this up and I have merged it into master.

PR #25, opened 2016-11-24
MINIFI-141 ignore build, bin, target, *.o, and *.a files in project

jdye64:


apiri:
Looks good to me.  Will merge and thanks for taking the time to add these!

PR #26, opened 2016-11-30
MINIFI-139 - Resolved OS X 10.12 build failures by introducing Cmake

jdye64:
find modules for LevelDB and UUID for both includes and libraries.


apiri:
reviewing

apiri:
hey @jdye64 things look overall and would like to avoid the absolute path.

could you let me know the version of XCode you have in your OS X environment?  
`xcodebuild -version`

apiri:
@jdye64 changes look good and the find package seems like the best path forward.  will merge this in and thanks!

PR #27, opened 2016-12-07
MINIFI-159: Create AppendHostInfo processor

randerzander:


randerzander:
I believe the CI failure is due to broken appveyor build configuration

apiri:
reviewing

libminifi/src/AppendHostInfo.cpp
```diff
@@ -0,0 +1,89 @@
...
+}
+
+void AppendHostInfo::onTrigger(ProcessContext *context, ProcessSession *session)
+{
+	FlowFileRecord *flow = session->get();
+	if (!flow)
+	  return;
+
+	//Get Hostname
+	char hostname[1024];
```
apiri: Is there a reason 1024 is used?  By spec, it is no more than 255 but may be lower depending on OS implementation.  limits.h has a HOST_NAME_MAX that contains this to get more precision on this instantiation

libminifi/src/AppendHostInfo.cpp
```diff
@@ -0,0 +1,89 @@
...
+	hostname[1023] = '\0'; 
+	gethostname(hostname, 1023);
+	struct hostent* h;
+	h = gethostbyname(hostname);
+  std::string hostAttribute;
+  context->getProperty(HostAttribute.getName(), hostAttribute);
+	flow->addAttribute(hostAttribute.c_str(), h->h_name);
+
+	//Get IP address for the specified interface
+  std::string iface;
```
apiri: What is the appropriate path if the interface specified does not exist?  Before adjusting, this currently returns 1.0.0.0.  Would it make sense, should the interface not be found, to omit the inclusion of the attribute and value?

apiri:
Hey @randerzander. Changes look good.  Will get this merged in.  Thanks for the contrib, definitely seems like it would be helpful.

PR #30, opened 2016-12-19
MINIFI-165 Updating licensing terms to include files bundled with spdlog

apiri:
MINIFI-165 Updating licensing terms to include files bundled with spdlog.

mpmc_bounded_q comes from http://www.1024cores.net/home/lock-free-algorithms/queues/bounded-mpmc-queue
and has a separate LICENSE file available from http://sites.google.com/site/1024cores/mpmc_bounded.zip


jdye64:
reviewing

PR #34, opened 2016-12-28
MINIFI-150 Adjusting ignored source ignore files

apiri:
MINIFI-150 Adjusting ignored source ignore files to prevent inclusion of git and supporting CMake directories.


jdye64:
reviewing

PR #35, opened 2016-12-30
MINIFI-149 Updating project README to aid in build.

apiri:
MINIFI-149 Updating project README with updates to build and needed dependencies.


jdye64:
reviewing

PR #36, opened 2017-01-02
MINIFI-180 Duplicate Variable causes MINIFI_HOME to be NULL if ENV variable is not present

jdye64:
Duplicate Variable causes MINIFI_HOME to be NULL if ENV variable is not
present


apiri:
reviewing

PR #37, opened 2017-01-03
MINIFI-171  Dynamic Properties support for processors

jdye64:
Added support in Processor.cpp to both set and get dynamic properties.
The logic will first attempt to set the parsed Property from the .yml
file as a supported property. IF that parsed Property is not in the
whitelisted supported list AND if the “nifi.accept.dynamic.properties”
configuration is set to true then the Property will be added to a
Map<String, Property> of dynamic properties that can be accessed in a
manner similar to the supported properties from any processor logic
that extends Processor.cpp

### DIFF
--- a/conf/minifi.properties
+++ b/conf/minifi.properties
@@ -16,6 +16,7 @@
 # Core Properties #
 nifi.version=0.2.0
 nifi.flow.configuration.file=./conf/config.yml
+nifi.accept.dynamic.properties=true
 nifi.administrative.yield.duration=30 sec
 # If a component has no work to do (is "bored"), how long should we wait before checking again for work?
 nifi.bored.yield.duration=10 millis
--- a/libminifi/src/Processor.cpp
+++ b/libminifi/src/Processor.cpp
@@ -29,6 +29,7 @@
 #include "Processor.h"
 #include "ProcessContext.h"
 #include "ProcessSession.h"
+#include "Configure.h"
 
 Processor::Processor(std::string name, uuid_t uuid)
 : _name(name)
@@ -206,6 +207,29 @@ bool Processor::getProperty(std::string name, std::string &value)
 	}
 }
 
+bool Processor::getDynamicProperty(std::string name, std::string &value)
+{
+	if (isRunning())
+		// Because set property only allowed in non running state, we need to obtain lock avoid rack condition
+		_mtx.lock();
+
+	std::map<std::string, Property>::iterator it = _dynamicProperties.find(name);
+	if (it != _dynamicProperties.end())
+	{
+		Property item = it->second;
+		value = item.getValue();
+		if (!isRunning())
+			_mtx.unlock();
+		return true;
+	}
+	else
+	{
+		if (!isRunning())
+			_mtx.unlock();
+		return false;
+	}
+}
+
 bool Processor::setProperty(std::string name, std::string value)
 {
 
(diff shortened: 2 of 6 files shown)


conf/minifi.properties
```diff
@@ -16,6 +16,7 @@
 # Core Properties #
 nifi.version=0.2.0
 nifi.flow.configuration.file=./conf/config.yml
+nifi.accept.dynamic.properties=true
```
apiri: Any reason why we shouldn't support dynamic properties by default? I understand we are trying to delineate between just blindly accepting new properties, but feel like the model present in NiFi is the best where processors can choose to support it.

jdye64:
@apiri I agree. That was my intent was to allow each Processor itself to determine how/if it used the dynamic properties that was the reason for creating the new set in Processor.cpp for the dynamic properties instead of just lumping them all into the same "supportedProperties" set. The only reason for the framework "switch" was just since the CPP version is meant to run on embedded and lightweight hardware maybe the guard against preventing someone from flooding the yml flow with so many dynamic properties that the embedded device ran out of memory and crashed. 

I see your point and in my own shallow mind I thought that was what I was trying to do but maybe your right and this implementation doesn't make sense? What were you thinking?

apiri:
not discounting the possibility, but with access to the YAML, there are other potential angles that are also at hand of a similar nature.  I do think the impact can certainly be mitigated during processor instantiation as to whether or not it is populated and incorporated; whether or not processors support these items is not something that changes on a given startup.

overall, I think my preference is for consistency in handling

libminifi/src/Processor.cpp
```diff
@@ -206,6 +207,29 @@ bool Processor::getProperty(std::string name, std::string &value)
 	}
 }
 
+bool Processor::getDynamicProperty(std::string name, std::string &value)
+{
+	if (isRunning())
+		// Because set property only allowed in non running state, we need to obtain lock avoid rack condition
+		_mtx.lock();
```
phrocker: Hi, any particular reason you did not use a lock guard ? http://en.cppreference.com/w/cpp/thread/lock_guard 
It's scoped on the stack pop so you don't need to unlock.

libminifi/src/Processor.cpp
```diff
@@ -206,6 +207,29 @@ bool Processor::getProperty(std::string name, std::string &value)
...
+	{
+		Property item = it->second;
+		value = item.getValue();
+		if (!isRunning())
+			_mtx.unlock();
+		return true;
+	}
+	else
+	{
+		if (!isRunning())
```
phrocker: I'm new to this, so forgive the question, but why only unlock when it is not running if the above conditional works  only when it is running? This would imply the mutex is locked when you are running. I likely missed something, though.

libminifi/src/Processor.cpp
```diff
@@ -226,6 +250,39 @@ bool Processor::setProperty(std::string name, std::string value)
 	}
 }
 
+bool Processor::setDynamicProperty(std::string name, std::string value)
+{
+	std::lock_guard<std::mutex> lock(_mtx);
+	std::map<std::string, Property>::iterator it = _dynamicProperties.find(name);
```
phrocker: Just a general comment about iterators. We could make this simpler with C++11 type deduction:
auto it = _dynamicProperties.find(name).

jdye64:
@phrocker thanks a ton for the suggestions. Updated with your recommendations.

jdye64:
@apiri Where do you think we stand on this? I know you had some concerns about the implementation approach in the beginning. Is that still the case? If so what can we do to resolve those?

apiri:
@jdye64 Didn't mean to leave things hanging.  My thought process is still on removing the configuration option and letting each processor choose how/if to interpret dynamic properties. With access to and/or controlling the YAML specified, I'm not sure what we are avoiding and worry that the configuration is going to be a "gotcha" that will throw off people's expectations as to functionality.

jdye64:
@apiri so do you think we should just close this PR then? Trying to tidy up a few PRs this morning

apiri:
@jdye64 Certainly in favor for the inclusion and support for dynamic properties.  Some of that work in Processor could be used toward that functionality.  Would obviously need to adapt things a bit but could be a nice start for framework level support to then delegate to the components to utilize.  Let me know your thoughts.

achristianson:
Taking this on.

phrocker:
@jdye64 feel free to close this. Thanks!

PR #38, opened 2017-01-05
MINIFI-175 Create official Apache NiFi MiNiFi C++ Dockerfile

jdye64:
Created version 1 of official NiFi-MiNiFi-CPP docker image. The image
can be built by running a custom target “make docker” please note that
“make package” must first be ran to ensure the binary assembly is
present.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -129,6 +129,6 @@ enable_testing(test)
 # Create a custom build target called "docker" that will invoke DockerBuild.sh and create the NiFi-MiNiFi-CPP Docker image
 add_custom_target(
     docker
-    COMMAND ${CMAKE_SOURCE_DIR}/docker/DockerBuild.sh 1000 1000 0.2.0 nifi-minifi-cpp-0.2.0-bin.tar.gz ${CMAKE_SOURCE_DIR}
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -88,3 +88,10 @@ set(CPACK_ARCHIVE_COMPONENT_INSTALL ON)
 set(CPACK_COMPONENTS_ALL bin)
 
 include(CPack)
+
+# Create a custom build target called "docker" that will invoke DockerBuild.sh and create the NiFi-MiNiFi-CPP Docker image
+add_custom_target(
+    docker
+    COMMAND ${CMAKE_SOURCE_DIR}/docker/DockerBuild.sh 1000 1000 0.2.0 nifi-minifi-cpp-0.2.0-bin.tar.gz ${CMAKE_SOURCE_DIR}
```
apiri: Would be good to to make use of the defined project version variables to avoid having to update this.  A bit more verbose, but would automatically be taken care of on release/version incrementing.

phrocker:
Going through some PRs to learn the lay of the land. I can't approve but I really like this one. It excites me +1

jdye64:
@apiri I think we are good for a first pass on the Docker image in my opinion. I think after we use it in the wild a couple of times we will have a better understanding of how people want to use it an can adapt as necessary.

apiri:
reviewing

apiri:
hey @jdye64... I did not have much luck getting the image to run.  I realize we have some bits at play with libs and such.  I had originally built on OS X but then realized this would be an OS X packaged binary.  Ran again on a CentOS system with similar issues, presumably to similar build related items.

jdye64:
@apiri I have made some changes. This should solve our problems with building and running the docker image.

apiri:
Looks good here.  Thanks for making the adjustment.  Will get this merged.

PR #39, opened 2017-01-05
MINIFI-181 Created initial implementation of PutFile

achristianson:
Wrote an initial implementation of PutFile. Not all features of NiFi's PutFile are supported at this point, but this is tested/functional and lays out the groundwork.


apiri:
Build looks good and processor works as anticipated.  Makes for a nice start on this and thanks for contributing.  Will get this merged in shortly.

PR #40, opened 2017-01-11
MINIFI-182 Implemented event-driven scheduler

achristianson:
Implemented the event-driven scheduler. Refactored the thread-related code into a base class as most of it is reused between the event-driven scheduler and timer-driven. The event-driven scheduler has a similar loop to the timer-driven, but instead of sleeping, it blocks until notified by another thread that work is available.

Results in logs like this:

    [2017-01-11 12:37:59.917] [minifi log] [info] GetFile process /tmp/getfile/test
    [2017-01-11 12:37:59.918] [minifi log] [info] PutFile using temporary file /tmp/putfile/.test.5a709b5b-b517-4710-bd56-2f0be2b754c0
    [2017-01-11 12:37:59.918] [minifi log] [info] PutFile writing file test into directory /tmp/putfile
    [2017-01-11 12:37:59.918] [minifi log] [info] PutFile committing put file operation to /tmp/putfile/test
    [2017-01-11 12:37:59.918] [minifi log] [info] PutFile commit put file operation to /tmp/putfile/test succeeded

Note that the PutFile was triggered immediately after GetFile picked up the file, which is what we want.

### DIFF
--- a/libminifi/include/EventDrivenSchedulingAgent.h
+++ b/libminifi/include/EventDrivenSchedulingAgent.h
@@ -1,6 +1,6 @@
 /**
- * @file TimerDrivenSchedulingAgent.h
- * TimerDrivenSchedulingAgent class declaration
+ * @file EventDrivenSchedulingAgent.h
+ * EventDrivenSchedulingAgent class declaration
  *
  * Licensed to the Apache Software Foundation (ASF) under one or more
  * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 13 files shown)


achristianson:
@benqiu2016 Good point. Fixed.

benqiu2016:
@achristianson looks good to me.

apiri:
reviewing

apiri:
changes and build looks good.  works as anticipated and looks like a solid foundation.  will get this merged in shortly.  Thanks!

PR #43, opened 2017-01-31
MINIFI-183 Implemented ListenHTTP

achristianson:
Implemented ListenHTTP. Supports the following features:

- Connection keepalive (to be friendly with clients using connection pooling)
- Two-way SSL
- Configurable minimum TLS version
- Pattern-based DN authorization
- Pattern-based extraction of HTTP headers to FlowFile attrs


achristianson:
@apiri Just resolved some conflicts with the master branch that emerged with other changes. Tested and should be good to go now.

apiri:
great, thanks.  will scope out shortly

libminifi/include/ListenHTTP.h
```diff
@@ -0,0 +1,116 @@
...
+		const struct mg_request_info *_reqInfo;
+	};
+
+protected:
+
+private:
+	//! Logger
+	Logger *_logger;
+
+	std::unique_ptr<CivetServer> _server;
```
phrocker: You have no idea how happy I am to see you use unique_ptr :)

libminifi/include/ProcessSessionFactory.h
```diff
@@ -0,0 +1,52 @@
...
+{
+public:
+	//! Constructor
+	/*!
+	 * Create a new process session factory
+	 */
+	ProcessSessionFactory(ProcessContext *processContext) : _processContext(processContext) {}
+
+	//! Create the session
+	std::unique_ptr<ProcessSession> createSession();
```
phrocker: It is my goal to make pointer usage clear. Thanks for making this exceptionally clear with unique_ptr here

libminifi/include/ProcessSessionFactory.h
```diff
@@ -0,0 +1,52 @@
...
+
+//! ProcessSessionFactory Class
+class ProcessSessionFactory
+{
+public:
+	//! Constructor
+	/*!
+	 * Create a new process session factory
+	 */
+	ProcessSessionFactory(ProcessContext *processContext) : _processContext(processContext) {}
```
phrocker: Do you want to mark this explicit?

libminifi/src/Processor.cpp
```diff
@@ -439,31 +447,25 @@ bool Processor::flowFilesOutGoingFull()
 	return false;
 }
 
-void Processor::onTrigger()
+void Processor::onTrigger(ProcessContext *context, ProcessSessionFactory *sessionFactory)
 {
-	ProcessContext *context = new ProcessContext(this);
-	ProcessSession *session = new ProcessSession(context);
+	auto session = sessionFactory->createSession();
```
phrocker: Thanks. I can remove this off of my my todo. fantastic!

libminifi/CMakeLists.txt
```diff
@@ -48,6 +49,19 @@ file(GLOB SPD_SOURCES "../include/spdlog/*")
 add_library(spdlog STATIC ${SPD_SOURCES})
 add_library(minifi STATIC ${SOURCES})
 
+# Include Boost System
+find_package(Boost COMPONENTS system REQUIRED)
```
phrocker: There seemed to be a little hesitation about using boost. My guess is that -Os added to flags won't actually cause the expected increase; however, it does potentially cause some bloat.

libminifi/src/ThreadedSchedulingAgent.cpp
```diff
@@ -64,11 +68,18 @@ void ThreadedSchedulingAgent::schedule(Processor *processor)
 		return;
 	}
 
+	auto processContext = std::make_shared<ProcessContext>(processor);
```
phrocker: Why are you getting the raw (stored -- not managed ) pointer if you're making a shared ptr?

CMakeLists.txt
```diff
@@ -63,6 +66,8 @@ find_package(UUID REQUIRED)
 file(GLOB SPD_SOURCES "include/spdlog/*")
 
 add_subdirectory(thirdparty/yaml-cpp-yaml-cpp-0.5.3)
+set(CIVETWEB_ENABLE_CXX ON CACHE BOOL "Enable civet C++ library")
```
phrocker: Could we download and then build  civet using ExternalProject instead of bundling it?

libminifi/CMakeLists.txt
phrocker: I assume this was included for civet? all uses of pointer management in libminifi appeared to be std::

libminifi/src/ListenHTTP.cpp
```diff
@@ -0,0 +1,395 @@
...
+			options.push_back("ssl_verify_peer");
+			options.push_back("yes");
+		}
+
+		if (sslMinVer.compare("SSL2") == 0)
+		{
+			options.push_back("ssl_protocol_version");
+			options.push_back(std::to_string(0));
+		}
+		else if (sslMinVer.compare("SSL3") == 0)
```
phrocker: @apiri  Are we concerned about POODLE? Do we want to disallow any version < TLS1.2?

phrocker:
If you happen to update try  -Os ( don't feel like you need to in this PR ) https://issues.apache.org/jira/browse/MINIFI-197 is a corresponding ticket for it. I was going to do this and prune out warnings that weren't important.

achristianson:
@phrocker @apiri Thanks for the feedback. All makes sense, but had one question re: civet as an ExternalProject. I did it as a thirdparty/ dir mainly to be consistent with how the current thirdparty dependencies are handled, e.g. yaml and leveldb. If we're OK with breaking that consistency, and perhaps adding future tickets to extract out the current thirdparty projects, then I'll move forward with that.

The only other concern regarding the ExternalProject is that it does make it harder to build offline or if the remote host goes down. We would probably want to add some build instructions for how and where to place the external code manually to cover these build-time use cases.

phrocker:
@achristianson I thought I created a ticket to break these out...or maybe it was an E-mail, but I think that's something we could discuss. When I sent a message to @apiri  earlier I said that the ExternalProject comment was mostly a desire and that this PR followed precedent and as such it's not a deal breaker, it's just a direction I would like to see things go. With that said, the operative word is suggestion and we can thus have a ticket to discuss this further. I'll create that ticket and we can discuss it there. 

TL;DR on my thoughts: having people build offline isn't a big deal because that likely wouldn't happen very frequently. In cases where it is anticipated, you would likely need to move code to said device, and in this case we could codify the download so that it's easy to stage. In my opinion, worrying about those types of systems also brings the worry about those which don't have OpenSSL or boost readily available. Since we don't package those and those certainly aren't standard fare on many lightweight distros, are we really breaking the mold to have ExternalProjects?

phrocker:
Created MINIFI-213 to have the CMake 3rd party discussion.

achristianson:
@phrocker Sounds good. I'll take a look at the rest and also fix this new merge conflict.

libminifi/CMakeLists.txt
achristianson: @phrocker Yes, that's for civet.

libminifi/include/ProcessSessionFactory.h
achristianson: @phrocker No problem. It makes it a lot easier to avoid memory management bugs.

libminifi/include/ProcessSessionFactory.h
achristianson: @phrocker Done.

libminifi/src/ThreadedSchedulingAgent.cpp
achristianson: @phrocker Mainly, creating a shared ProcessContext and SessionFactory which will be used in this thread as well as the threads created in the loop below. I wanted these objects to be cleaned up once they go out of scope in all threads. The raw pointers are only used to integrate with the existing Processor pointer that is passed-in, as well as passing raw pointers to the constructors of the new shared objects. I wanted to minimize function/constructor changes that might have impacts elsewhere in the code base. If we want to move more to having managed pointers across the code base, then some of this can be cleaned up.

achristianson:
@apiri @phrocker Addressed the comments, updated the code, and rebased. The biggest thing is cleanly setting up an ExternalProject system, but I think that's best addressed under the ticket @phrocker created.

apiri:
@achristianson sounds fair concerning the ExternalProjects, will scope this out

apiri:
We will need to update the LICENSE for the civetweb source inclusion.  That does not appear to be present currently.  My understanding is that we are not using any of the optional items in the build, so the following portion should be sufficient: https://github.com/civetweb/civetweb/blob/master/LICENSE.md#included-with-all-features  Please clarify if my understanding is incorrect.

achristianson:
@apiri Correct. Not using any of the optional features listed there. Want me to update the LICENSE file, or were you planning to do it? I don't mind, just want to avoid a merge conflict.

libminifi/CMakeLists.txt
```diff
@@ -48,6 +49,19 @@ file(GLOB SPD_SOURCES "../include/spdlog/*")
...
+target_link_libraries(minifi ${Boost_SYSTEM_LIBRARY})
+
+# Include libxml2
```
apiri: Does civet need libxml2?  Otherwise this should no longer be needed.

achristianson: @apiri I *think* this was something that was there already, not something required for civet. If you're sure it's not needed elsewhere, I can take it out and test to make sure civet doesn't need it.

apiri: Okay, may have just been an artifact.  This was removed as per MINIFI-187 (https://github.com/apache/nifi-minifi-cpp/commit/0b022245d46371567077efd128cbf21a1657aa4d)

apiri:
@achristianson feel free to update the LICENSE when you have the opportunity if you are going to adjust the libxml items

achristianson:
@apiri Will do. Expecting to get an updated PR out here in a couple minutes.

achristianson:
@apiri Done.

apiri:
@achristianson great, thanks.  still working through testing functionality

achristianson:
@apiri Sure thing. These cmds might help for part of your testing:

```sh
dd if=/dev/urandom of=./testdat bs=1M count=1
```
```sh
curl -vvv -X POST --data-binary @./testdat http://127.0.0.1:8888/contentListener
```

Here's a link with info to create test TLS certs: http://stackoverflow.com/questions/14267010/how-to-create-self-signed-ssl-certificate-for-test-purposes

Here's the flow I was testing with:

```markdown
Flow Controller:
  name: MiNiFi Flow
Processors:
    # - name: GetFile
    #   class: org.apache.nifi.processors.standard.GetFile
    #   max concurrent tasks: 1
    #   scheduling strategy: TIMER_DRIVEN
    #   scheduling period: 1 sec
    #   penalization period: 30 sec
    #   yield period: 1 sec
    #   run duration nanos: 0
    #   auto-terminated relationships list:
    #   Properties:
    #       Input Directory: /tmp/getfile
    #       Keep Source File: false
    #       File Filter: ".*"
    - name: ListenHTTP
      class: org.apache.nifi.processors.standard.ListenHTTP
      max concurrent tasks: 16
      scheduling strategy: TIMER_DRIVEN
      scheduling period: 1 sec
      penalization period: 30 sec
      yield period: 1 sec
      run duration nanos: 0
      auto-terminated relationships list:
      Properties:
          Listening Port: 8888
        #   Authorized DN Pattern: /C=AU/ST=Some-State/O=Internet Widgits Pty Ltd/CN=localhost|/C=US/ST=Maryland/O=Internet Widgits Pty Ltd/CN=localhost
        #   SSL Certificate: /home/ubuntu/workspace/build/server.pem
        #   SSL Certificate Authority: /home/ubuntu/workspace/build/server.pem
        #   SSL Verify Peer: yes
        #   SSL Minimum Version: TLS1.2
          HTTP Headers to receive as Attributes (Regex): X-Test2.*
    - name: LogAttribute
      class: org.apache.nifi.processors.standard.LogAttribute
      max concurrent tasks: 1
      scheduling strategy: EVENT_DRIVEN
      scheduling period: 1 sec
      penalization period: 30 sec
      yield period: 1 sec
      run duration nanos: 0
      auto-terminated relationships list:
    - name: PutFile
      class: org.apache.nifi.processors.standard.PutFile
      max concurrent tasks: 1
      scheduling strategy: EVENT_DRIVEN
      scheduling period: 1 sec
      penalization period: 30 sec
      yield period: 1 sec
      run duration nanos: 0
      auto-terminated relationships list: ["success", "failure"]
      Properties:
          Output Directory: /tmp/putfile
Connections:
    # - name: TransferFiles
    #   source name: GetFile
    #   source relationship name: success
    #   destination name: PutFile
    #   max work queue size: 0
    #   max work queue data size: 1 MB
    #   flowfile expiration: 60 sec
    - name: TransferHttpToLogAttribute
      source name: ListenHTTP
      source relationship name: success
      destination name: LogAttribute
      max work queue size: 0
      max work queue data size: 1 MB
      flowfile expiration: 60 sec
    - name: LogAttributeToPutFile
      source name: LogAttribute
      source relationship name: success
      destination name: PutFile
      max work queue size: 0
      max work queue data size: 1 MB
      flowfile expiration: 60 sec
Remote Processing Groups: []
```

apiri:
Great, thanks for all the resources.  If you are not aware, the community also has the TLS Toolkit that is super helpful as well:  http://nifi.apache.org/docs/nifi-docs/html/administration-guide.html#tls-generation-toolkit

apiri:
@achristianson had some other things that interrupted, but looks good.  will get these squashed and merged.  appreciate you working through all the iterations with us!

PR #45, opened 2017-02-03
MINIFI-187: Remove XML References and deps.

phrocker:
Remove XML references and dependencies. This
includes changing the reload functionality to use
YAML instead of XML.


apiri:
reviewing

apiri:
changes look good and verified config after the fact.  additionally going to remove the references in the readme to the removed XML deps on merge.  

Thanks for cleaning up!

PR #46, opened 2017-02-03
MINIFI-188: Incorporate CATCH and example tests

phrocker:
This PR introduces catch and a few example tests.
Additional tests will follow, but they currently cause
several segfaults without a large number of corresponding
changes in the code. Therefore to minimize the bootstrapping
I will include the cmake changes in addition to the dep.

to run type make; make test 

You can also run ./tests in the build directory for more information

CATCH is Boost license, let me know if this is a problem. Thanks!


apiri:
reviewing

libminifi/cmake/FindLeveldb.cmake
```diff
@@ -0,0 +1,50 @@
...
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless qrequired by applicable law or agreed to in writing,
```
apiri: typo

phrocker: Thanks! I'll fix that.

apiri:
reviewing updates

libminifi/cmake/FindLeveldb.cmake
phrocker: I see this same typo in other files. I'll make that change in either this PR or the next. Apparently I propagated the typo.

apiri: Ah, gotcha.  Good with this one or a follow-on.  Your call!

apiri: Actually, I can just take care of those items as part of the merge.

phrocker: Oh I already force pushed the change.

phrocker: I updated the commit to include a blurb about it so that it was squashed into a single commit.

apiri:
@phrocker Thanks for cleaning up the typos.  The changes all look good and builds are working on Linux for me.

Boost license is perfectly fine as per http://www.apache.org/legal/resolved.html#category-a

To that end, and apologize for overlooking this initially, but we will need to update the project LICENSE as we are including the source as part of the codebase.  After that, I believe we should be good for merging.

Thanks!

apiri:
Thanks for taking care of licensing.  Will get this merged in!

PR #49, opened 2017-02-13
MINIFI-198: Update root CMAKE so testing can occur

phrocker:
Update root CMake to bring OpenSSL location to top level. Keeping find_package(...) in libminifi CMakeLists.txt to avoid issues building only libminifi.


apiri:
reviewing

apiri:
@phrocker changes look great.  thanks for fixing.  will get this merged

PR #50, opened 2017-02-13
MINIFI-195: Convert uuid_unparse usage to uuid_unparse_lower

phrocker:
This will ensure that we use lower case UUIDs. Undefining
the macro won't work and we likely don't want to change
the default behavior.


apiri:
reviewing

apiri:
rebased on master and built on a couple environments.  changes and test cases look good.  will get this merged.  thanks!

PR #53, opened 2017-02-21
Minifi 215

benqiu2016:


benqiu2016:
i rebase the benqiu2016/nifi-minifi-cpp to nifi-minifi-cpp and create branch MINIFI-215 under benqiu2016 to track the change.

benqiu2016:
need rebase

PR #52, opened 2017-02-21
MINIFI-218: Support for GPSD integration

jdye64:

PR #54, opened 2017-02-22
MINIFI-206: Implement Extensible Comm Mechanism

phrocker:
Create initial communication mechanism that's
initially based on BSD sockets.

MINIFI-201: Change how hostname is resolved
and ensure that it is not null

MINIFI-199: Utilize zlib's crc32 implementation,
which is more portable and faster. we can explore
alternatives when that time arrives.

MINIFI-200: Put host "get" into initialization.
Any change wouldn't be picked up by a running process
most likely.


phrocker:
Due to the size of this PR I'm submitting it in hopes of continuing tests on TLS and submitting that as a separate commit before merge.

libminifi/CMakeLists.txt
```diff
@@ -36,18 +36,34 @@ ENDIF(POLICY CMP0048)
...
+else()
+ message(STATUS "The compiler ${CMAKE_CXX_COMPILER} has no C++11 support. Please use a different C++ compiler.")
+endif()
+
 include_directories(../include)
 include_directories(../thirdparty/yaml-cpp-yaml-cpp-0.5.3/include)
 include_directories(include)
 
-file(GLOB SOURCES "src/*.cpp")
+file(GLOB_RECURSE SOURCES "src/*.cpp")
```
phrocker: I generally don't like to see this. Since we only have two sub directories, we can call them out.

apiri:
reviewing

apiri:
Hey @phrocker, for the most part things look good.  Thanks for going through the tedium to get everything a bit more consistent.  Overall, nothing worth holding things up and protracting the PR.

As per one of your previous comments, what is the anticipated sequence of events concerning TLS before merging?

libminifi/CMakeLists.txt
```diff
@@ -36,12 +36,23 @@ ENDIF(POLICY CMP0048)
...
+ message(STATUS "The compiler ${CMAKE_CXX_COMPILER} has no C++11 support. Please use a different C++ compiler.")
+endif()
+
 include_directories(../include)
 include_directories(../thirdparty/yaml-cpp-yaml-cpp-0.5.3/include)
 include_directories(../thirdparty/civetweb-1.9.1/include)
 include_directories(include)
 
-file(GLOB SOURCES "src/*.cpp")
+file(GLOB_RECURSE SOURCES "src/*.cpp")
```
jdye64: Will think conflict with conditional builds like in https://github.com/apache/nifi-minifi-cpp/pull/52/files#diff-2672ba1ce10995865ff39a95c67bc5dcR72 since those features are in fact subdirectories of src?

phrocker:
@apiri I tested it locally. I would like to add tests for it in unit tests ala the java client. That may require a little bit more work though to stand up a server, but that will be a subsequent PR

@jdye64  I will make this change

phrocker:
@apiri FYI the test protocol I used was to test s2s, bidirectionally, in secure and non-secure mode. I tested in linux and under mac. It inspired me to try and build an integration test

apiri:
All right, did some more poking around and looks good.  Think we are good to get this rolled in.  Thanks again for all the efforts.

PR #55, opened 2017-02-22
MINIFI-222: Upgrade spdlog to latest release.

phrocker:
Modified Logger so that it removed the previously required flush argument.

Copyright headers changed but copyright license is the same.


apiri:
hey @phrocker,

Just so I can better scope out licensing and such, is this version 0.12?

apiri:
I see that it is in the code and is 0.12 (https://github.com/apache/nifi-minifi-cpp/pull/55/files#diff-ec88af83a0f1976c3c605e8b1027c454R10)

include/spdlog/sinks/ansicolor_sink.h
```diff
@@ -0,0 +1,116 @@
+//
+// Copyright(c) 2016 Kevin M. Godby (a modified version by spdlog).
```
apiri: Seems like we would need to call this out similar to https://github.com/apache/nifi-minifi-cpp/blob/ff7aa7fc3bf55681398427f1de6af7f352a100fa/LICENSE#L394

include/spdlog/sinks/dist_sink.h
```diff
@@ -0,0 +1,71 @@
+//
+// Copyright (c) 2015 David Schury, Gabi Melman
```
apiri: Believe we need to include David Schury in the LICENSE in addition to Gabi Melman: https://github.com/apache/nifi-minifi-cpp/blob/ff7aa7fc3bf55681398427f1de6af7f352a100fa/LICENSE#L243

include/spdlog/sinks/msvc_sink.h
```diff
@@ -0,0 +1,50 @@
+//
+// Copyright(c) 2016 Alexander Dalshov.
```
apiri: Also should be covered in LICENSE

include/spdlog/sinks/wincolor_sink.h
```diff
@@ -0,0 +1,116 @@
+//
+// Copyright(c) 2016 spdlog
```
apiri: LICENSE adjustment for this entity as well

phrocker: thanks!

phrocker:
I put the authors with the corresponding license multiple times. I feel like it would be better to include the Copyrights and the license once, but I've seen it done like this in other apache projects. advise if the alternative is preferred.

apiri:
I think the license once is preferred to kind of limit the noise and still provides the full coverage.  Reviewing the ListenHTTP I saw CivetWeb (https://github.com/civetweb/civetweb/blob/master/LICENSE.md#included-with-all-features) seemed to do this and I think is clear from a legality standpoint about spdlog as a whole.

apiri:
@phrocker changes look good, thanks, will get this merged

PR #57, opened 2017-02-24
MINIFI-83: Allow appenders to be specified and auto configured.

phrocker:
This PR will modify logging so that we can have configurable
appenders and have them be modified via custom configuration
for that appender. Eventually these can be modified real time.


apiri:
reviewing

libminifi/include/LogAppenders.h
```diff
@@ -0,0 +1,292 @@
...
+
+			if (use_std_err == "true") {
+				std::string err_unique_name =
+						getUniqueName<OutputStreamAppender>();
+				auto error_ostream_sink = std::make_shared<
+						spdlog::sinks::ostream_sink_mt>(std::cerr);
+				stderr_ = std::make_shared<spdlog::logger>(err_unique_name,
+						error_ostream_sink);
+			}
+		} else
```
apiri: minor:  believe we should make use of braces around else clause as well

libminifi/include/LogAppenders.h
```diff
@@ -0,0 +1,292 @@
...
+		logger_ = std::make_shared<spdlog::logger>(unique_name, ostream_sink);
+
+		stderr_ = nullptr;
+
+		std::string log_level;
+		if (NULL != config
+				&& config->get(BaseLogger::nifi_log_level, log_level)) {
+			setLogLevel(log_level);
+		} else
+			setLogLevel("info");
```
apiri: minor:  believe we should make use of braces around else clause as well

libminifi/include/LogAppenders.h
phrocker: Good catch, thanks. I'll update this when you're finished commenting so I don't make your life harder.

phrocker: I did that in a few places apparently...

apiri:
Looks great, thanks.  Will get this merged.

PR #58, opened 2017-02-24
MINIFI-145 - Updating Travis to perform builds on both OS X and Linux

apiri:
MINIFI-145 - Updating Travis to perform builds on both OS X and Linux


jdye64:
Looks good Aldrin. Thanks for adding this was needed for sure.

PR #62, opened 2017-03-02
MINIFI-231: Add Flow Persistent, Using id instead of name to load the…

benqiu2016:
… flow from YAML

### DIFF
--- a/libminifi/include/Connection.h
+++ b/libminifi/include/Connection.h
@@ -84,6 +84,10 @@ public:
 		else
 			return false;
 	}
+	//! Get UUID Str
+	std::string getUUIDStr() {
+		return _uuidStr;
+	}
 	//! Set Connection Source Processor
 	void setSourceProcessor(Processor *source) {
 		_srcProcessor = source;
@@ -180,7 +184,8 @@ protected:
 	std::atomic<uint64_t> _maxQueueDataSize;
 	//! Flow File Expiration Duration in= MilliSeconds
 	std::atomic<uint64_t> _expiredDuration;
-
+	//! UUID string
+	std::string _uuidStr;
 
 private:
 	//! Mutex for protection
--- a/libminifi/include/FlowController.h
+++ b/libminifi/include/FlowController.h
@@ -47,6 +47,7 @@
 #include "FlowControlProtocol.h"
 #include "RemoteProcessorGroupPort.h"
 #include "Provenance.h"
+#include "FlowFileRepository.h"
 #include "GetFile.h"
 #include "PutFile.h"
 #include "TailFile.h"
@@ -64,6 +65,7 @@
 #define CONFIG_YAML_PROCESSORS_KEY "Processors"
 
 struct ProcessorConfig {
+	std::string id;
 	std::string name;
 	std::string javaClass;
 	std::string maxConcurrentTasks;
@@ -129,6 +131,10 @@ public:
 	virtual ProvenanceRepository *getProvenanceRepository() {
 		return this->_provenanceRepo;
 	}
+	//! Get the flowfile repository
+	virtual FlowFileRepository *getFlowFileRepository() {
+		return this->_flowfileRepo;
+	}
 	//! Load flow xml from disk, after that, create the root process group and its children, initialize the flows
 	virtual void load() = 0;
 
@@ -172,7 +178,6 @@ public:
 		_protocol->setSerialNumber(number);
 	}
 
-
 protected:
   
 	//! A global unique identifier
@@ -197,6 +202,8 @@ protected:
 	std::atomic<bool> _initialized;
 	//! Provenance Repo
 	ProvenanceRepository *_provenanceRepo;
+	//! FlowFile Repo
+	FlowFileRepository *_flowfileRepo;
 	//! Flow Engines
 	//! Flow Timer Scheduler
 	TimerDrivenSchedulingAgent _timerScheduler;
@@ -212,7 +219,7 @@ protected:
 
 	FlowController() :
 			_root(0), _maxTimerDrivenThreads(0), _maxEventDrivenThreads(0), _running(
-					false), _initialized(false), _provenanceRepo(0), _protocol(
+					false), _initialized(false), _provenanceRepo(0), _flowfileRepo(0), _protocol(
 					0), logger_(Logger::getLogger()){
 	}
 
@@ -246,6 +253,8 @@ public:
 	void unload();
 	//! Load new xml
 	void reload(std::string yamlFile);
+	//! Load Flow File from persistent Flow Repo
+	void loadFlowRepo();
 	//! update property value
 	void updatePropertyValue(std::string processorName,
 			std::string propertyName, std::string propertyValue) {
@@ -283,6 +292,7 @@ private:
 	//! Logger
 	std::shared_ptr<Logger> logger_;
 	Configure *configure_;
+
 	//! Process Processor Node YAML
 	void parseProcessorNodeYaml(YAML::Node processorNode, ProcessGroup *parent);
 	//! Process Port YAML
--- a/libminifi/include/FlowFileRecord.h
+++ b/libminifi/include/FlowFileRecord.h
@@ -108,11 +108,11 @@ public:
 	/*!
 	 * Create a new flow record
 	 */
-	FlowFileRecord(std::map<std::string, std::string> attributes, ResourceClaim *claim = NULL);
+	explicit FlowFileRecord(std::map<std::string, std::string> attributes, ResourceClaim *claim = NULL);
 	/*!
 	 * Create a new flow record from repo flow event
 	 */
-	FlowFileRecord(FlowFileEventRecord *event);
+	explicit FlowFileRecord(FlowFileEventRecord *event);
 	//! Destructor
 	virtual ~FlowFileRecord();
 	//! addAttribute key is enum
--- a/libminifi/include/FlowFileRecord.h
+++ b/libminifi/include/FlowFileRecord.h
@@ -37,6 +37,7 @@
 
 class ProcessSession;
 class Connection;
+class FlowFileEventRecord;
 
 #define DEFAULT_FLOWFILE_PATH "."
 
@@ -108,6 +109,10 @@ public:
 	 * Create a new flow record
 	 */
 	FlowFileRecord(std::map<std::string, std::string> attributes, ResourceClaim *claim = NULL);
+	/*!
+	 * Create a new flow record from repo flow event
+	 */
+	FlowFileRecord(FlowFileEventRecord *event);
 	//! Destructor
 	virtual ~FlowFileRecord();
 	//! addAttribute key is enum
@@ -175,6 +180,15 @@ public:
 	{
 		return _lineageIdentifiers;
 	}
+	//! Check whether it is stored to DB already
+	bool isStoredToRepository()
+	{
+		return _isStoredToRepo;
+	}
+	void setStoredToRepository(bool value)
+	{
+		_isStoredToRepo = value;
+	}
 
 protected:
 
@@ -202,6 +216,8 @@ protected:
 	std::string _uuidStr;
 	//! UUID string for all parents
 	std::set<std::string> _lineageIdentifiers;
+	//! whether it is stored to DB
+	bool _isStoredToRepo;
 	//! duplicate the original flow file
 	void duplicate(FlowFileRecord *original);
 
--- a/libminifi/include/FlowFileRepository.h
+++ b/libminifi/include/FlowFileRepository.h
@@ -1,0 +1,208 @@
+/**
+ * @file FlowFileRepository 
+ * Flow file repository class declaration
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef __FLOWFILE_REPOSITORY_H__
+#define __FLOWFILE_REPOSITORY_H__
+
+#include <ftw.h>
+#include <uuid/uuid.h>
+#include <atomic>
+#include <cstdint>
+#include <cstring>
+#include <iostream>
+#include <map>
+#include <set>
+#include <string>
+#include <thread>
+#include <vector>
+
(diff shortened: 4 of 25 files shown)


libminifi/include/Repository.h
```diff
@@ -0,0 +1,294 @@
...
+		leveldb::Slice value((const char *) buf, bufLen);
+		leveldb::Status status;
+		status = _db->Put(leveldb::WriteOptions(), key, value);
+		if (status.ok())
+			return true;
+		else
+			return false;
+	}
+	//! Delete
+	virtual bool Delete(std::string key)
```
phrocker: const std::string &key ?

libminifi/include/Repository.h
```diff
@@ -0,0 +1,294 @@
...
+		status = _db->Get(leveldb::ReadOptions(), key, &value);
+		if (status.ok())
+			return true;
+		else
+			return false;
+	}
+	//! destroy
+	void destroy()
+	{
+		if (_db)
```
phrocker: This doesn't appear to be thread safe. other state variables are not changed here. How is this limited with a public function?

libminifi/include/Repository.h
```diff
@@ -0,0 +1,294 @@
...
+	Configure *configure_;
+	int64_t _maxPartitionMillis;
+	//! max db size
+	int64_t _maxPartitionBytes;
+	//! purge period
+	uint64_t _purgePeriod;
+	//! level DB database
+	leveldb::DB* _db;
+	//! thread
+	std::thread *_thread;
```
phrocker: Why does this need to be a pointer? The move operator allows you to do thread_ = std::thread{ func, var }

libminifi/include/Repository.h
```diff
@@ -0,0 +1,294 @@
...
+	//! whether it is running
```
phrocker: Can you better explain the difference of running and enable? Preferably in comments?

libminifi/include/Repository.h
```diff
@@ -0,0 +1,294 @@
...
+	bool _running;
+	//! whether it is enable 
+	bool _enable;
+	//! whether stop accepting provenace event
+	std::atomic<bool> _repoFull;
+	//! repoSize
+	uint64_t repoSize();
+	//! size of the directory
+	static uint64_t _repoSize[MAX_REPO_TYPE];
```
phrocker: Can this be configuration based?

libminifi/include/Repository.h
```diff
@@ -0,0 +1,294 @@
...
+	//! call back for directory size
+	static int repoSumProvenance(const char *fpath, const struct stat *sb, int typeflag)
+	{
+		_repoSize[PROVENANCE] += sb->st_size;
```
phrocker: Is this state that can be maintained in metadata of the repo versus directory?

libminifi/src/FlowFileRecord.cpp
```diff
@@ -74,6 +76,43 @@ FlowFileRecord::FlowFileRecord(std::map<std::string, std::string> attributes, Re
...
+	_offset = event->getFileOffset();
+	_lineageIdentifiers = event->getLineageIdentifiers();
+	_attributes = event->getAttributes();
+    _snapshot = false;
+    _uuidStr = event->getFlowFileUuid();
+    uuid_parse(_uuidStr.c_str(), _uuid);
+
+    if (_size > 0)
+    {
+    	_claim = new ResourceClaim();
```
phrocker: Why does claim_ need to be a pointer? You should use a unique_ptr if that is required.

libminifi/src/Repository.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+void Repository::stop() {
+	if (!_running)
+		return;
+	_running = false;
+	logger_->log_info("%s Repository Monitor Thread Stop", RepositoryTypeStr[_type]);
+}
+
+void Repository::run(Repository *repo) {
+	// threshold for purge
```
phrocker: What is this calculation based on?

libminifi/src/Repository.cpp
```diff
@@ -0,0 +1,140 @@
...
+							"NiFi %s retrieve event %s fail",
+							RepositoryTypeStr[repo->_type],
+							key.c_str());
+						purgeList.push_back(key);
+					}
+				}
+			}
+			delete it;
+			std::vector<std::string>::iterator itPurge;
+			for (itPurge = purgeList.begin(); itPurge != purgeList.end();
```
phrocker: why not use auto?

libminifi/include/FlowController.h
```diff
@@ -197,6 +202,8 @@ class FlowController {
 	std::atomic<bool> _initialized;
 	//! Provenance Repo
 	ProvenanceRepository *_provenanceRepo;
+	//! FlowFile Repo
+	FlowFileRepository *_flowfileRepo;
```
phrocker: Why does this need to be a pointer? Why not use unique/shared pointer depending on the need?

phrocker:
I'll continue to go through it. I made a very cursory look. Good changes overall!

libminifi/include/Connection.h
```diff
@@ -180,7 +184,8 @@ class Connection
 	std::atomic<uint64_t> _maxQueueDataSize;
 	//! Flow File Expiration Duration in= MilliSeconds
 	std::atomic<uint64_t> _expiredDuration;
-
+	//! UUID string
+	std::string _uuidStr;
```
phrocker: as i was going through it again I wanted to make clear that this is informational and not a huge deal.

benqiu2016: all the other private members in the Connection.h are like _XXX based. Just want to keep it consistent.

libminifi/include/FlowFileRecord.h
```diff
@@ -108,6 +109,10 @@ class FlowFileRecord
 	 * Create a new flow record
 	 */
 	FlowFileRecord(std::map<std::string, std::string> attributes, ResourceClaim *claim = NULL);
```
benqiu2016: will do.

libminifi/include/FlowFileRecord.h
```diff
@@ -108,6 +109,10 @@ class FlowFileRecord
...
+	/*!
+	 * Create a new flow record from repo flow event
+	 */
+	FlowFileRecord(FlowFileEventRecord *event);
```
benqiu2016: will use explicit, i am not using smart pointer here is that if you look at where we construct FlowFileRecord from FlowFileEventRecord, i am using a FlowFileEventRecord in the stack to. I want to avoid to create a new FlowFileEventRecord every time.

libminifi/include/FlowFileRecord.h
```diff
@@ -202,6 +216,8 @@ class FlowFileRecord
 	std::string _uuidStr;
 	//! UUID string for all parents
 	std::set<std::string> _lineageIdentifiers;
+	//! whether it is stored to DB
+	bool _isStoredToRepo;
```
benqiu2016: Same comment as above, keep it consistent.

libminifi/include/FlowFileRepository.h
```diff
@@ -0,0 +1,208 @@
...
+#include "Connection.h"
+#include "FlowFileRecord.h"
+#include "Logger.h"
+#include "Property.h"
+#include "ResourceClaim.h"
+#include "io/Serializable.h"
+#include "utils/TimeUtil.h"
+#include "Repository.h"
+
+class FlowFileRepository;
```
benqiu2016: will do.

libminifi/include/FlowFileRepository.h
```diff
@@ -0,0 +1,208 @@
...
+
+//! FlowFile Event Record
+class FlowFileEventRecord : protected Serializable
+{
+public:
+	friend class ProcessSession;
```
benqiu2016: will remove that.

libminifi/include/FlowFileRepository.h
```diff
@@ -0,0 +1,208 @@
...
+	{
+		_eventTime = getTimeMillis();
+		logger_ = Logger::getLogger();
+	}
+
+	//! Destructor
+	virtual ~FlowFileEventRecord() {
+	}
+	//! Get Attributes
+	std::map<std::string, std::string> getAttributes() {
```
benqiu2016: if we call getAttributes and somehow the event which holding that attributes was deleted, we will have trouble.

libminifi/include/FlowFileRepository.h
```diff
@@ -0,0 +1,208 @@
...
+	bool DeSerialize(DataStream &stream)
+	{
+		return DeSerialize(stream.getBuffer(),stream.getSize());
+	}
+	//! DeSerialize
+	bool DeSerialize(FlowFileRepository *repo, std::string key);
+
+protected:
+
+	//! Date at which the event was created
```
benqiu2016: to keep consistent with current code

libminifi/include/FlowFileRepository.h
```diff
@@ -0,0 +1,208 @@
...
+
+//! FlowFile Repository
+class FlowFileRepository : public Repository
+{
+public:
+	//! Constructor
+	/*!
+	 * Create a new provenance repository
+	 */
+	FlowFileRepository()
```
benqiu2016: It is just a default constructor.

libminifi/include/Repository.h
```diff
@@ -0,0 +1,294 @@
...
+#include "FlowFileRecord.h"
+#include "Logger.h"
+#include "Property.h"
+#include "ResourceClaim.h"
+#include "io/Serializable.h"
+#include "utils/TimeUtil.h"
+#include "utils/StringUtils.h"
+
+//! Repository
+class Repository
```
benqiu2016: some flow repository behave is different from provenance repository. so we need to have that extend repository.

libminifi/include/Repository.h
```diff
@@ -0,0 +1,294 @@
...
+				return false;
+			}
+		}
+
+		return true;
+	}
+	//! Put
+	virtual bool Put(std::string key, uint8_t *buf, int bufLen)
+	{
+		if (!_enable)
```
benqiu2016: enable is set while the repo is initialized from minifi property, so we are safe here

libminifi/include/Repository.h
benqiu2016: sure.

libminifi/include/Repository.h
benqiu2016: we only call destroy when we delete the repo, will change to a private function.

libminifi/include/Repository.h
benqiu2016: just to keep code persistent with other way to do the std::thread in ThreadSchedulingAgent and make sure that we start the thread not on constructor.

libminifi/include/Repository.h
benqiu2016: Sure.

libminifi/include/Repository.h
benqiu2016: we only supported two repo as NiFI

libminifi/include/Repository.h
benqiu2016: LevelDB does not expose a API to check DB size, also the levelDB has log and extra file when we commit the record, it was better to check the size of the directory to get the whole size.

libminifi/src/FlowFileRecord.cpp
benqiu2016: the claim was pass around while flow was dup or clone, etc

libminifi/src/Repository.cpp
benqiu2016: uint64_t purgeThreshold = repo->_maxPartitionBytes * 3 / 4;

# Provenance Repository #
nifi.provenance.repository.directory.default=./provenance_repository
nifi.provenance.repository.max.storage.time=1 MIN
nifi.provenance.repository.max.storage.size=1 MB
So once the repo size exceed 75%, we start to purge the record if the record was stay in the repo exceed the max storage time.
If repo size is full, we mark the repo full and we will stop to put the record in repo.

libminifi/src/Repository.cpp
benqiu2016: will do.

libminifi/include/FlowController.h
benqiu2016: keep it consistent with other context in FlowController

libminifi/include/Connection.h
phrocker: We're making a transition to the google code style. I don't think consistency in a class is a concern now since we're in such flux. Doing that will help me avoid having to do new members.

libminifi/include/FlowFileRepository.h
phrocker: what do you mean somehow? can you explain a path? Why would you be worrying about an event record being deleted? If that is the case we have bigger problems. perhaps we should discuss this.

libminifi/src/Repository.cpp
phrocker: I think a comment should be placed to reflect this.

libminifi/include/Connection.h
phrocker: Again this isn't a big deal but will be changed.

benqiu2016:
@phrocker Add unitest, please review and approve. I would like to get it merge before your big name space change. Thanks a lot.

libminifi/test/unit/ProcessorTests.cpp
```diff
@@ -33,10 +35,18 @@ TEST_CASE("Test Find file", "[getfileCreate2]"){
 
 	TestController testController;
 
-	testController.enableDebug();
+	Configure *config = Configure::getConfigure();
 
-	ProvenanceTestRepository repo;
-	TestFlowController controller(repo);
+	config->set(BaseLogger::nifi_log_appender,"rollingappender");
```
phrocker: @apiri  @benqiu2016  Can this be removed? either by Bin or Aldrin on the merge. Logs shouldn't be needed for tests.

libminifi/test/unit/ProvenanceTestHelper.h
```diff
@@ -20,6 +20,62 @@
 
 #include "Provenance.h"
 #include "FlowController.h"
+#include "FlowFileRepository.h"
+
+/**
+ * Test repository
+ */
+class FlowTestRepository : public FlowFileRepository
```
benqiu2016: the FlowFileRepo construct to populate the default repo type and parameters for the repo 
/*!
	 * Create a new provenance repository
	 */
	FlowFileRepository()
	 : Repository(Repository::FLOWFILE, FLOWFILE_REPOSITORY_DIRECTORY,
			MAX_FLOWFILE_REPOSITORY_ENTRY_LIFE_TIME, MAX_FLOWFILE_REPOSITORY_STORAGE_SIZE, FLOWFILE_REPOSITORY_PURGE_PERIOD)
	{
	}

libminifi/test/unit/ProcessorTests.cpp
benqiu2016: sure. we can remove that. For my last test, i found test is failed and have trouble to find the log for the test that just ran.

benqiu2016:
@apiri could you please merge the PR once you have free time. No rush. But it would be good if it can get in before Marc name space change.

apiri:
reviewing

apiri:
No substantial comments beyond those already covered.  Looks good and verified functionality.  Will merge, thanks for the work, @benqiu2016!

PR #66, opened 2017-03-09
MINIFI-215: IOS port, add OPENSSL_SUPPORT/YAML_SUPPORT/LEVELDB_SUPPORT into cmake

benqiu2016:

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/src/io/TLSSocket.cpp
```diff
@@ -20,23 +20,29 @@
 #include "Configure.h"
```
phrocker: The entire class should be excluded.

phrocker:
Instead of commenting on every macro, perhaps an alternative is to, at build time, include files that are appropriate to your unit. If you want YAML, have a configuration class that's only built as a result of Yaml. This may require having a different directory or including specific files but that's better than the definitions here.

phrocker:
Here's an idea. We could include the files as a result. You could make it so that you have only one or two macros that define, perhaps, template metaprogramming. Look into SFINAE. At the very least you could have only a handful of macro definitions instead of the numerous that exist here

libminifi/include/io/TLSSocket.h
```diff
@@ -24,8 +24,10 @@
 #include <mutex>
```
benqiu2016: it is difficult because we need to add more # define in other files. I tried that before

libminifi/src/io/TLSSocket.cpp
benqiu2016: same as above

benqiu2016:
Instead of commenting on every macro, perhaps an alternative is to, at build time, include files that are appropriate to your unit. If you want YAML, have a configuration class that's only built as a result of Yaml. This may require having a different directory or including specific files but that's better than the definitions here. 
Above is ideal, still we need to have some #define in the caller to indicate which one to call or we need to have two set of files, one is just stub if the feature was disable, one has the feature dependent stuff.

bqiust:
@apiri @phrocker any feedback? i updated the above comments. Thanks. Bin

benqiu2016:
i think we may need to revisit the way to do MARCO in the later stage.

PR #85, opened 2017-04-28
MINIFI-275 Bugfix for YAML Configs without Component IDs

kevdoran:
This commit fixes a bug in which loading config YAML that did not
specify IDs (UUIDs) for components caused an exception. The logic
previously treated component IDs as required fields by loading them
without checking for existence. This commit updates the logic to
generate IDs when they are not specified in the YAML config. As part
of this fix, the logic for loading connections from YAML config was
modified to search by name or remote port id in the absence of source
id or remote id in the YAML config.

For fields that are required, useful error messages were added when
those fields are missing to assist users in self-diagnosing issues
related to invalid config files.

Some minor tweaks to the top level .gitignore are included with this
commit.

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/include/core/yaml/YamlConfiguration.h
```diff
@@ -88,10 +102,17 @@ class YamlConfiguration : public FlowConfiguration {
...
   // Process Provenance Report YAML
-  void parseProvenanceReportingYaml(YAML::Node *reportNode, core::ProcessGroup * parentGroup);
+  void parseProvenanceReportingYaml(YAML::Node *reportNode,
+                                    core::ProcessGroup * parentGroup);
   // Parse Properties Node YAML for a processor
   void parsePropertiesNodeYaml(YAML::Node *propertiesNode,
                                std::shared_ptr<core::Processor> processor);
+  // Helper function to get or generate a component's id field
+  std::string getOrGenerateId(YAML::Node *yamlNode, std::string idField = "id");
+  // Get or generate a component's id field
```
kevdoran: Good catch on the comment, and yeah, will switch to const std::string &str

libminifi/test/unit/YamlCongifurationTests.cpp
```diff
@@ -0,0 +1,182 @@
...
+#include <core/RepositoryFactory.h>
+#include "core/yaml/YamlConfiguration.h"
+#include "../TestBase.h"
+
+static const std::shared_ptr<core::Repository> TEST_PROV_REPO = core::createRepository("provenancerepository", true);
+static const std::shared_ptr<core::Repository> TEST_FF_REPO = core::createRepository("flowfilerepository", true);
+
+TEST_CASE("Test YAML Config 1", "[testyamlconfig1]") {
+
+  static const std::string TEST_YAML_WITHOUT_IDS = "MiNiFi Config Version: 1\n"
```
kevdoran: will do

libminifi/src/core/yaml/YamlConfiguration.cpp
```diff
@@ -491,14 +510,18 @@ void YamlConfiguration::parsePortYaml(YAML::Node *portNode,
...
+      "is specificy an input port for a NiFi Remote Process Group, the port "
+      "id should match the id of of the input port in the NiFi configuration. "
+      "This is a UUID of the format XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX.");
+  auto portId = inputPortsObj["id"].as<std::string>();
   uuid_parse(portId.c_str(), uuid);
 
-  port = new minifi::RemoteProcessorGroupPort(nameStr.c_str(), uuid);
+  port = new minifi::RemoteProcessorGroupPort(nameStr, uuid);
 
   processor = (std::shared_ptr<core::Processor>) port;
```
kevdoran: ok. the same style cast occurs in a few other places in the file, eg, getRoot(...). I assume you prefer just avoiding raw pointers altogether when possible?

libminifi/include/core/yaml/YamlConfiguration.h
```diff
@@ -88,10 +102,17 @@ class YamlConfiguration : public FlowConfiguration {
   void parseRemoteProcessGroupYaml(YAML::Node *node,
                                    core::ProcessGroup * parent);
   // Process Provenance Report YAML
-  void parseProvenanceReportingYaml(YAML::Node *reportNode, core::ProcessGroup * parentGroup);
+  void parseProvenanceReportingYaml(YAML::Node *reportNode,
+                                    core::ProcessGroup * parentGroup);
   // Parse Properties Node YAML for a processor
   void parsePropertiesNodeYaml(YAML::Node *propertiesNode,
                                std::shared_ptr<core::Processor> processor);
+  // Helper function to get or generate a component's id field
```
kevdoran: I like that. I'll update all existing function comments in YamlConfiguration.h while I'm here

libminifi/src/core/yaml/YamlConfiguration.cpp
```diff
@@ -538,14 +561,62 @@ void YamlConfiguration::parsePropertiesNodeYaml(
...
+    uuid_generate(uuid);
+    char uuid_str[37];
+    uuid_unparse(uuid, uuid_str);
+    id = uuid_str;
+    logger_->log_debug("Generating random ID: id => [%s]", id);
+  }
+  return id;
+}
+
+void YamlConfiguration::checkRequiredField(
```
kevdoran: Hey @phrocker,

If I follow you correctly, the concern is about the conditional on line 610?

operator[] returns a bool for the version of yaml-cpp we are using:
https://github.com/jbeder/yaml-cpp/wiki/Tutorial

In older versions of the yaml-cpp api, it did in fact throw an exception as you describe:
https://github.com/jbeder/yaml-cpp/wiki/How-To-Parse-A-Document-(Old-API)#optional-keys

As to the function in general, it puts the logic in one place for generating helpful error messages and eliminated the need for if/else or try/catch blocks throughout the code. The idea is you would call this prior to using a required field; if the field is not present, a runtime error with a useful user-facing error message is generated, otherwise you are safe to use the field without a try/catch or if/else after calling checkRequiredField(...)

Does that answer your question?

I would like input on the function interface (aside from the types, which I am changing to const std::string &str).. I ended up tacking on a couple optional fields solely for the use in the error message generation, but it does seem like a lot to ask the caller just for an assertion check. Let me know if you see a more elegant way to achieve the desired functionality here.

libminifi/src/core/yaml/YamlConfiguration.cpp
phrocker: I don't see the cast you are referring to in getRoot. I only see the use of the unique_ptr, which  okay as that is currently the only way to create a unique pointer. What I'm referencing is the explicit cast of a pointer to a shared_ptr object, which could have negative consequences at some point. If this is an endemic issue we can create a ticket, but I didn't see another case of this specific cast. The appropriate thing to do is std::make_shared<minifi::RemoteProcessorGroupPort>(nameStr,uuid);

kevdoran: I updated my earlier comment. You're right, this is a one-off not an endemic issue.

is this cast ok:

`std::shared_ptr<core::Processor> processor = NULL;
std::shared_ptr<minifi::RemoteProcessorGroupPort> port = NULL;
port = new minifi::RemoteProcessorGroupPort(nameStr, uuid);
processor = (std::shared_ptr<core::Processor>) port;
`

libminifi/src/core/yaml/YamlConfiguration.cpp
phrocker: Sorry, the double reply was likely confusing. I'm good with the function itself. The code that I'm looking at in our third party directory returns a Node
template <typename Key>
  const Node operator[](const Key& key) const;

The exception I reference is below and throws an InvalidNode if the node is invalid some how. Is this also handled?
`template <typename Key>
inline const Node Node::operator[](const Key& key) const {
  if (!m_isValid)
    throw InvalidNode();
  EnsureNodeExists();
  detail::node* value = static_cast<const detail::node&>(*m_pNode)
                            .get(detail::to_value(key), m_pMemory);
  if (!value) {
    return Node(ZombieNode);
  }
  return Node(*value, m_pMemory);
}`

phrocker: While it's touching the same code, I believe this may be out of the scope of this PR and ticket. We can simply create another ticket for this case.

libminifi/src/core/yaml/YamlConfiguration.cpp
phrocker: No that cast is not okay. That should be
```
port = std::make_shared<minifi::RemoteProcessorGroupPort>(nameStr,uuid);
processor = std::static_pointer_cast<core::Processor>(port);
```

phrocker: This will create a true shared pointer object for port and perform a compile time cast. I think the only reason port exists is to call RPG specific functions later, which predates me.

libminifi/src/core/yaml/YamlConfiguration.cpp
kevdoran: Oh I see. Nope, not checking for InvalidNode. Will take a look at that and might add some basic handling, but may need to spend some more time thinking about how to best implement that as the [] operator is used throughout this file, not just in this function. So yeah, might open a ticket instead

kevdoran:
Thanks for the review and feedback @phrocker!

apiri:
reviewing

apiri:
@kevdoran looks good here!  thanks for fixing this up and associated test cases.

will get this merged.

libminifi/test/unit/YamlCongifurationTests.cpp
apiri: @kevdoran @phrocker if there isn't one, can we be sure to capture this in a JIRA issue?

kevdoran: @apiri, I'm opening a JIRA now

PR #88, opened 2017-05-02
MINIFI-286: Fix compilation error when GCC version is greater than 4.9

jdye64:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/src/processors/GetFile.cpp
+++ b/libminifi/src/processors/GetFile.cpp
@@ -24,13 +24,6 @@
 #include <dirent.h>
 #include <limits.h>
 #include <unistd.h>
-#if  (__GNUC__ >= 4)
-#if (__GNUC_MINOR__ < 9)
-#include <regex.h>
-#else
-#include <regex>
-#endif
-#endif
 #include <vector>
 #include <queue>
 #include <map>
@@ -267,9 +260,6 @@ bool GetFile::acceptFile(std::string fullName, std::string name,
     if (request.keepSourceFile == false && access(fullName.c_str(), W_OK) != 0)
       return false;
 
-#ifdef __GNUC__
-#if (__GNUC__ >= 4)
-#if (__GNUC_MINOR__ < 9)
     regex_t regex;
     int ret = regcomp(&regex, request.fileFilter.c_str(), 0);
     if (ret)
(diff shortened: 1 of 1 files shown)


apiri:
reviewing

apiri:
Hey @jdye64,

I was not able to make the current codebase fail under Ubuntu 17.04 which has gcc 6.3.0 (12ubuntu2).  Any other considerations of your environment?

phrocker:
@apiri The logic can be improved  @jdye64 I'm in favor of removing the conditional expression entirely and using the c style for all versions since that will work. There's no point in using the C++ regex matching since it won't be available for all builds. If you really want to maintain it, you can do this:

```
diff --git a/libminifi/src/processors/GetFile.cpp b/libminifi/src/processors/GetFile.cpp
index 2740793..9988029 100644
--- a/libminifi/src/processors/GetFile.cpp
+++ b/libminifi/src/processors/GetFile.cpp
@@ -24,8 +24,9 @@
 #include <dirent.h>
 #include <limits.h>
 #include <unistd.h>
-#if  (__GNUC__ >= 4)
-#if (__GNUC_MINOR__ < 9)
+// if an older version of gcc that doesn't support <regex>
+#ifdef __GNUC__
+#if (__GNUC__ < 4 || (__GNUC__ == 4 && __GNUC_MINOR__ < 9))
 #include <regex.h>
 #else
 #include <regex>
@@ -268,8 +269,7 @@ bool GetFile::acceptFile(std::string fullName, std::string name,
       return false;
 
 #ifdef __GNUC__
-#if (__GNUC__ >= 4)
-#if (__GNUC_MINOR__ < 9)
+#if (__GNUC__ < 4 || (__GNUC__ == 4 && __GNUC_MINOR__ < 9))
     regex_t regex;
     int ret = regcomp(&regex, request.fileFilter.c_str(), 0);
     if (ret)
@@ -280,7 +280,7 @@ bool GetFile::acceptFile(std::string fullName, std::string name,
       return false;
 #else
     try {
-      std::regex re(fileFilter);
+      std::regex re(request.fileFilter);
 
       if (!std::regex_match(name, re)) {
         return false;
@@ -290,7 +290,6 @@ bool GetFile::acceptFile(std::string fullName, std::string name,
       return false;
     }
 #endif
-#endif
 #else
     logger_->log_info("Cannot support regex filtering");
 #endif
```

jdye64:
I like just removing the conditional expression. I'm going to update to that now.

apiri:
cool, looks good and was able to verify the issue beforehand with a specific docker environment that hit the bad logic.  I did notice there was a linter issue but I can take care of that on merge

Thanks for correcting this!

PR #89, opened 2017-05-03
MINIFI-293 Remove thirdparty dependencies no longer utilized

apiri:
MINIFI-293 Remove thirdparty dependencies no longer utilized inclusive of libxml2 and all of civetweb thirdparty.


benqiu2016:
+1

PR #91, opened 2017-05-04
MINIFI-258 - Removing Configure, StreamFactory, TLSContext singletons

brosander:

### For all changes:
     in the commit message?

- [X] Does your PR title start with MINIFI-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.

- [X] Has your PR been rebased against the latest commit within the target branch (typically master)?

### For code changes:
- [X] - N/A - If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?
- [X] - N/A - If applicable, have you updated the LICENSE file?
- [X] - N/A - If applicable, have you updated the NOTICE file?

- [X] - N/A - Have you ensured that format looks appropriate for the output in which it is rendered?

### Note:

### DIFF
--- a/libminifi/include/EventDrivenSchedulingAgent.h
+++ b/libminifi/include/EventDrivenSchedulingAgent.h
@@ -38,8 +38,8 @@ class EventDrivenSchedulingAgent : public ThreadedSchedulingAgent {
   /*!
    * Create a new processor
    */
-  EventDrivenSchedulingAgent(std::shared_ptr<core::Repository> repo)
-      : ThreadedSchedulingAgent(repo) {
+  EventDrivenSchedulingAgent(std::shared_ptr<core::Repository> repo, std::shared_ptr<Configure> configure)
+      : ThreadedSchedulingAgent(repo, configure) {
   }
   // Destructor
   virtual ~EventDrivenSchedulingAgent() {
--- a/libminifi/include/EventDrivenSchedulingAgent.h
+++ b/libminifi/include/EventDrivenSchedulingAgent.h
@@ -38,7 +38,7 @@ class EventDrivenSchedulingAgent : public ThreadedSchedulingAgent {
   /*!
    * Create a new processor
    */
-  EventDrivenSchedulingAgent(std::shared_ptr<core::Repository> repo, std::shared_ptr<Configure> configure)
+  EventDrivenSchedulingAgent(const std::shared_ptr<core::Repository> &repo, const std::shared_ptr<Configure> &configure)
       : ThreadedSchedulingAgent(repo, configure) {
   }
   // Destructor
--- a/libminifi/include/EventDrivenSchedulingAgent.h
+++ b/libminifi/include/EventDrivenSchedulingAgent.h
@@ -38,7 +38,7 @@ class EventDrivenSchedulingAgent : public ThreadedSchedulingAgent {
   /*!
    * Create a new processor
    */
-  EventDrivenSchedulingAgent(const std::shared_ptr<core::Repository> &repo, const std::shared_ptr<Configure> &configure)
+  EventDrivenSchedulingAgent(std::shared_ptr<core::Repository> repo, std::shared_ptr<Configure> configure)
       : ThreadedSchedulingAgent(repo, configure) {
   }
   // Destructor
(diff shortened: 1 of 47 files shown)


libminifi/include/core/Processor.h
```diff
@@ -243,7 +244,7 @@ class Processor : public Connectable, public ConfigurableComponent,
   std::atomic<bool> _triggerWhenEmpty;
 
 //! obtainSite2SiteProtocol for use
```
phrocker: In this case we may want to use a const reference to avoid the need to increment and decrement the reference counter. This will avoid performance issues, eventually. This isn't always required, but in places where the callgraph may enter this function frequently, it's usually good practice.

libminifi/src/io/ClientSocket.cpp
```diff
@@ -39,7 +40,7 @@ namespace io {
 
 char *Socket::HOSTNAME = const_cast<char*>(Socket::getMyHostName(0).c_str());
 
-Socket::Socket(const std::string &hostname, const uint16_t port,
+Socket::Socket(std::shared_ptr<SocketContext> context, const std::string &hostname, const uint16_t port,
```
phrocker: same, may want to pass by const reference here.

libminifi/src/io/tls/TLSSocket.cpp
```diff
@@ -31,14 +32,11 @@ namespace nifi {
...
-std::mutex TLSContext::context_mutex;
-
-TLSContext::TLSContext()
-    : error_value(0),
+TLSContext::TLSContext(std::shared_ptr<Configure> configure)
+    : SocketContext(configure), error_value(0),
       ctx(0),
       logger_(logging::Logger::getLogger()),
-      configuration(Configure::getConfigure()) {
+      configure_(configure) {
```
phrocker: for some reason configure_ seems wrong here. sounds petty but Configure as the name of the Configuration class seems odd to me.

phrocker:
Looks good overall. I'm gonna take another pass through it in a bit with some more caffeinated eyes.

libminifi/include/core/yaml/YamlConfiguration.h
```diff
@@ -41,10 +42,12 @@ namespace core {
 class YamlConfiguration : public FlowConfiguration {
 
  public:
-  YamlConfiguration(std::shared_ptr<core::Repository> repo,
-                    std::shared_ptr<core::Repository> flow_file_repo,
+  YamlConfiguration(const std::shared_ptr<core::Repository> &repo,
```
phrocker: might have to do the same here to avoid removing const ( which you can do obviously )

libminifi/include/FlowController.h
```diff
@@ -64,8 +64,9 @@ class FlowController : public core::CoreComponent {
   /**
    * Flow controller constructor
    */
-  FlowController(std::shared_ptr<core::Repository> provenance_repo,
-                 std::shared_ptr<core::Repository> flow_file_repo,
+  FlowController(const std::shared_ptr<core::Repository> &provenance_repo,
```
phrocker: we could move the original std shared ptr to subsume ownership, but I'm not sure we want that. We want to say we're sharing ownership with this object and thus this object ( FlowController ) helps in determining the lifetime of shared_ptr.

libminifi/include/EventDrivenSchedulingAgent.h
```diff
@@ -38,8 +38,8 @@ class EventDrivenSchedulingAgent : public ThreadedSchedulingAgent {
   /*!
    * Create a new processor
    */
-  EventDrivenSchedulingAgent(std::shared_ptr<core::Repository> repo)
-      : ThreadedSchedulingAgent(repo) {
+  EventDrivenSchedulingAgent(const std::shared_ptr<core::Repository> &repo, const std::shared_ptr<Configure> &configure)
```
phrocker: These schedulers stop in a way in which we don't guarantee threads to stop. We wait in flow controller 1s to stop threads, thus we can't be 100% certain that the scheduled threads have stopped. They "probably" did, but we likely want to keep this pass by value to ensure that we are guaranteed shared ownership of the shared repo. Since the threads are daemon threads, they will stop eventually, but we probably want to avoid the error of a dereferencing deleted memory if we can to aid in eventual recoverability ( if that's a thing ).

brosander:
Thanks for feedback, changed those back and am rerunning build/tests

apiri:
reviewing

apiri:
build works, ran some simple integration flows.  code changes look good.  thanks for the refactoring, this should be quite helpful moving forward!

will get this merged shortly

PR #93, opened 2017-05-06
MINIFI-301 Removing extraneous third party resources

apiri:
MINIFI-301 Removing extraneous third party resources from the project source tree.


benqiu2016:
+1

PR #94, opened 2017-05-07
MINIFI-302 Exclude generated docs from the source tarball

apiri:
MINIFI-302 Exclude generated docs from the source tarball via CPack ignore.


trkurc:
code looks good - testing now

PR #95, opened 2017-05-09
MINIFI-290 Refactor Tests

kevdoran:
- Correct doxygen function comment in YamlConfiguration
    - Correct a log message in YamlConfiguration
    - Tweak ubuntu Dockerfile to fix docker build target
    - Add --output-on-failure flag to Travis config

### For all changes:
     in the commit message?

### For code changes:

### Note:


apiri:
reviewing

apiri:
Hey @kevdoran. This is certainly a lot cleaner and a much welcomed improvement.  Might you be able to rebase on the changes that went in recently for the updates to how configure is handled and take care of renaming YamlConfiguration while we are messing with it in this commit?

Thanks!

kevdoran:
Thanks for the feedback @apiri. Will do

kevdoran:
@apiri and @phrocker - requested changes have been made. let me know if there is anything else here

apiri:
@kevdoran Thanks for rebasing/adjusting.  Will get this merged.

apiri:
Saw that the Travis build for Linux failed.  Doing some additional builds/runs to make sure we're clear first.

kevdoran:
I'll run some more linux tests as well. Should we look into configuring the CTests to dump verbose output on failures so we could get more information out of Travis?

phrocker:
@kevdoran that ticket already exists. It's part of another PR.

kevdoran:
@apiri and @phrocker:

Here is the verbose test 

```
-------------------------------------------------------------------------------
Test YAML Config Processing
  loading YAML without optional component IDs works
-------------------------------------------------------------------------------
/home/travis/build/kevdoran/nifi-minifi-cpp/libminifi/test/unit/YamlConfigurationTests.cpp:25
...............................................................................

/home/travis/build/kevdoran/nifi-minifi-cpp/libminifi/test/unit/YamlConfigurationTests.cpp:34: FAILED:
due to a fatal error condition:
  SIGSEGV - Segmentation violation signal

===============================================================================
test cases: 11 | 10 passed | 1 failed
assertions: 48 | 47 passed | 1 failed

1/6 Test #1: LibMinifiTests ...................***Failed    0.05 sec
```

It looks like this is failing for the same reason identified in MINIFI-304, which is resolved in @phrocker's pull request #96 [1]. Is that ready to merge?

[1] https://github.com/apache/nifi-minifi-cpp/pull/96

apiri:
Hey @kevdoran,

#96 is now merged.  Would you be able to rebase and see if that solves the problem you were encountering before?

kevdoran:
Will do.

kevdoran:
@apiri @phrocker - update on this: 

The approach of setting a hardcoded working directory when adding tests to ctest in the CMakeLists.txt file proved to be too brittle. If you invoke the test executable directly, it fails. After some exploration, I could not find a way to pass the resource location to the Catch unit tests in a manner that I was satisfied with. Two alternatives I explored included passing the location as an environment variable and passing it as a command line argument. The environment variable approach was a no-go as it it would require us to setup the "test" target in CMake using something other than enable_testing() and instead creating our own target, which seemed undesirable at this time. The argument approach would require us to define our own main for Catch unit tests, rather than using the `#define CATCH_CONFIG_MAIN`, and because `#define CATCH_CONFIG_MAIN` does more than just inline a main function, I didn't want to do that either in case assumptions made today break with other versions of Catch.hpp in the future.

The benefits here seemed minor, and could not justify the extra complexity introduced by the workarounds I explored (IMO), so I reverted the yaml config to being hardcoded inline strings. For now, I recommend closing MINIFI-290 as "won't do" and merge this PR which has a few other changes that I cam across while investigating this. Alternatively, we could move MINIFI-290 back to "open" if we want to revisit this in the future... but I'm sure if the need arises we would reopen it or make a new ticket.

apiri:
@kevdoran thanks for all the thought process and rationale.  totally seems fair. will review the updates.  thanks!

libminifi/test/TestBase.h
```diff
@@ -29,6 +29,13 @@
 #include "core/Core.h"
 #include "properties/Configure.h"
 
+/* Defining the root location of test resource files here in case it moves.   *
```
apiri: Is this a relic of the original approach?  Don't believe it's currently being used any more.

kevdoran: Good catch! Will push a commit removing that to my branch

kevdoran:
Updated

apiri:
@kevdoran cool, thanks will get merged!

PR #103, opened 2017-05-23
MINIFI-177 Incorporating Apache RAT to build with a custom target of

apiri:
apache-rat.  Updating non-compliant files and/or listing exclusions
in .rat-excludes

This commit introduces a separate CMake file to manage the acquisition
of the RAT binary.

### DIFF
--- a/cmake/RunApacheRAT.cmake
+++ b/cmake/RunApacheRAT.cmake
@@ -1,0 +1,44 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# Find the preferred Apache mirror to use for the download by querying the list of mirrors and filtering out 'preferred'
+execute_process(COMMAND curl -s https://www.apache.org/dyn/closer.lua/?asjson=1
+        COMMAND grep preferred
+        COMMAND awk "{print $2}"
+        COMMAND tr -d "\""
+        TIMEOUT 10
+        OUTPUT_STRIP_TRAILING_WHITESPACE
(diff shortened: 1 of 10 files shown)


apiri:
After performing cmake configuration, the RAT check can be performed via

```
make apache-rat

<lots of output>
*****************************************************
Summary
-------
Generated at: 2017-05-23T13:21:38-04:00

Notes: 2
Binaries: 9
Archives: 0
Standards: 194

Apache Licensed: 194
Generated Documents: 0

JavaDocs are generated, thus a license header is optional.
Generated files do not require license headers.

0 Unknown Licenses
```

apiri:
Also of note is that this isn't a pass/fail operation but provides output of the analysis.

apiri:
@phrocker Good call.

apiri:
may have to rework structure a bit to cache the binary and/or make use of the mirror provided by the ASF closer.lua for getting this.  I think the selected endpoint in this case blacklisted the node or had issues.

apiri:
Updated to pull the artifact from a mirror provided by the ASF closer.lua script.

cmake/RunApacheRAT.cmake
```diff
@@ -0,0 +1,44 @@
...
+# Find the preferred Apache mirror to use for the download by querying the list of mirrors and filtering out 'preferred'
+execute_process(COMMAND curl -s https://www.apache.org/dyn/closer.lua/?asjson=1
+        COMMAND grep preferred
+        COMMAND awk "{print $2}"
+        COMMAND tr -d "\""
+        TIMEOUT 10
+        OUTPUT_STRIP_TRAILING_WHITESPACE
+        OUTPUT_VARIABLE MIRROR_URL )
+
+ExternalProject_Add(
```
brosander: @apiri can we make it so this is only downloaded when we run the rat target?

cmake/RunApacheRAT.cmake
```diff
@@ -0,0 +1,44 @@
...
+        rat-binary
+        PREFIX "apache-rat"
+        URL "${MIRROR_URL}creadur/apache-rat-0.12/apache-rat-0.12-bin.tar.gz"
+        URL_HASH SHA1=e84dffe8b354871c29f5078b823a726508474a6c
```
brosander: @apiri can we use sha256 or sha512? if it's not available from the apache mirror, we could precalculate it since it shouldn't change until we upgrade rat

apiri: yep, can do

cmake/RunApacheRAT.cmake
apiri: adjusted this to use the script mode of cmake and perform that check/download only when the specific target is invoked.  what is a little tricky is that CMake handles the file downloading across at configuration time and also has the built in facilities to handle this wherever CMake runs.  May be a better way, but this one came to mind without having to sling around additional arguments/variables for the invoking command

brosander:
+1 LGTM

PR #114, opened 2017-06-21
site2site port negotiation

benqiu2016:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/RemoteProcessorGroupPort.h
+++ b/libminifi/include/RemoteProcessorGroupPort.h
@@ -23,6 +23,7 @@
 #include <mutex>
 #include <memory>
 #include <stack>
+#include "utils/HTTPUtils.h"
 #include "concurrentqueue.h"
 #include "FlowFileRecord.h"
 #include "core/Processor.h"
@@ -42,20 +43,30 @@ class RemoteProcessorGroupPort : public core::Processor {
   /*!
    * Create a new processor
    */
-  RemoteProcessorGroupPort(const std::shared_ptr<io::StreamFactory> &stream_factory, std::string name, uuid_t uuid = nullptr)
+  RemoteProcessorGroupPort(const std::shared_ptr<io::StreamFactory> &stream_factory, std::string name, std::string url, std::shared_ptr<Configure> configure, uuid_t uuid = nullptr)
       : core::Processor(name, uuid),
+        configure_(configure),
         direction_(SEND),
         transmitting_(false),
-        logger_(logging::LoggerFactory<RemoteProcessorGroupPort>::getLogger()) {
+        logger_(logging::LoggerFactory<RemoteProcessorGroupPort>::getLogger()),
+        url_(url),
+        securityConfig_(configure) {
     stream_factory_ = stream_factory;
     if (uuid != nullptr) {
       uuid_copy(protocol_uuid_, uuid);
     }
+    site2site_port_ = -1;
+    site2site_secure_ = false;
+    site2site_peer_index_ = -1;
+    // REST API port and host
+    port_ = -1;
+    utils::parse_url(url_, host_, port_, protocol_);
   }
   // Destructor
   virtual ~RemoteProcessorGroupPort() {
 
   }
+
   // Processor Name
   static const char *ProcessorName;
   // Supported Properties
@@ -84,6 +95,17 @@ class RemoteProcessorGroupPort : public core::Processor {
   void setTransmitting(bool val) {
     transmitting_ = val;
   }
+  // setURL
+  void setURL(std::string val) {
+    url_ = val;
+    utils::parse_url(url_, host_, port_, protocol_);
+  }
+
+  // refresh remoteSite2SiteInfo via nifi rest api
+  void refreshRemoteSite2SiteInfo();
+
+  // refresh site2site peer list
+  void refreshPeerList();
 
  protected:
 
@@ -93,6 +115,7 @@ class RemoteProcessorGroupPort : public core::Processor {
 
   moodycamel::ConcurrentQueue<std::unique_ptr<Site2SiteClientProtocol>> available_protocols_;
 
+  std::shared_ptr<Configure> configure_;
   // Logger
   std::shared_ptr<logging::Logger> logger_;
   // Transaction Direction
@@ -104,9 +127,21 @@ class RemoteProcessorGroupPort : public core::Processor {
 
   uuid_t protocol_uuid_;
 
+  // rest API end point info
   std::string host_;
+  int port_;
+  std::string protocol_;
+  std::string url_;
 
-  uint16_t port_;
+  // Remote Site2Site Info
+  int site2site_port_;
+  bool site2site_secure_;
+  std::vector<minifi::Site2SitePeerStatus> site2site_peer_status_list_;
+  std::atomic<int> site2site_peer_index_;
+  std::mutex site2site_peer_mutex_;
+  std::string rest_user_name_;
+  std::string rest_password_;
+  minifi::utils::HTTPSecurityConfiguration securityConfig_;
 
 };
 
--- a/libminifi/include/RemoteProcessorGroupPort.h
+++ b/libminifi/include/RemoteProcessorGroupPort.h
@@ -99,6 +99,14 @@ class RemoteProcessorGroupPort : public core::Processor {
   void setURL(std::string val) {
     url_ = val;
     utils::parse_url(url_, host_, port_, protocol_);
+    if (port_ == -1) {
+      if (protocol_.find("https") != std::string::npos) {
+        port_ = 443;
+      }
+      else if (protocol_.find("http") != std::string::npos) {
+        port_ = 80;
+      }
+    }
   }
 
   // refresh remoteSite2SiteInfo via nifi rest api
--- a/libminifi/include/Site2SiteClientProtocol.h
+++ b/libminifi/include/Site2SiteClientProtocol.h
@@ -43,6 +43,7 @@
 #include "core/ProcessContext.h"
 #include "core/ProcessSession.h"
 #include "io/CRCStream.h"
+#include "utils/Id.h"
 
 namespace org {
 namespace apache {
@@ -301,7 +302,7 @@ class Transaction {
     char uuidStr[37];
 
     // Generate the global UUID for the transaction
-    uuid_generate(_uuid);
+    id_generator_->generate(_uuid);
     uuid_unparse_lower(_uuid, uuidStr);
     _uuidStr = uuidStr;
   }
@@ -363,7 +364,7 @@ class Transaction {
   int _transfers;
   // Number of content bytes
   uint64_t _bytes;
-
+  static std::shared_ptr<utils::IdGenerator> id_generator_;
 };
 
 /**
@@ -630,6 +631,7 @@ class Site2SiteClientProtocol {
   // Only support pass by reference or pointer
   Site2SiteClientProtocol(const Site2SiteClientProtocol &parent);
   Site2SiteClientProtocol &operator=(const Site2SiteClientProtocol &parent);
+  static std::shared_ptr<utils::IdGenerator> id_generator_;
 };
 
 } /* namespace minifi */
--- a/libminifi/include/Site2SiteClientProtocol.h
+++ b/libminifi/include/Site2SiteClientProtocol.h
@@ -387,6 +387,15 @@ class DataPacket {
 
 };
 
+/**
+  * Site2Site Peer
+  */
+ typedef struct Site2SitePeerStatus {
+   std::string host_;
+   int port_;
+   bool isSecure_;
+ } Site2SitePeerStatus;
+
 // Site2SiteClientProtocol Class
 class Site2SiteClientProtocol {
  public:
@@ -475,6 +484,8 @@ class Site2SiteClientProtocol {
   }
   // bootstrap the protocol to the ready for transaction state by going through the state machine
   bool bootstrap();
+  // get peerList
+  bool getPeerList(std::vector<minifi::Site2SitePeerStatus> &peer);
   // establish
   bool establish();
   // handShake
--- a/libminifi/include/utils/HTTPUtils.h
+++ b/libminifi/include/utils/HTTPUtils.h
@@ -21,7 +21,15 @@
 
 #include <curl/curl.h>
 #include <vector>
+#include <iostream>
+#include <string>
+#include <curl/curlbuild.h>
+#include <curl/easy.h>
+#include <openssl/ssl.h>
 #include "ByteInputCallBack.h"
+#include "core/logging/Logger.h"
+#include "core/logging/LoggerConfiguration.h"
+#include "properties/Configure.h"
 
 namespace org {
 namespace apache {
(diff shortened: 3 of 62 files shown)


libminifi/src/core/yaml/YamlConfiguration.cpp
```diff
@@ -314,10 +314,8 @@ void YamlConfiguration::parseProvenanceReportingYaml(YAML::Node *reportNode, cor
   auto schedulingStrategyStr = node["scheduling strategy"].as<std::string>();
   checkRequiredField(&node, "scheduling period", CONFIG_YAML_PROVENANCE_REPORT_KEY);
   auto schedulingPeriodStr = node["scheduling period"].as<std::string>();
-  checkRequiredField(&node, "host", CONFIG_YAML_PROVENANCE_REPORT_KEY);
-  auto hostStr = node["host"].as<std::string>();
-  checkRequiredField(&node, "port", CONFIG_YAML_PROVENANCE_REPORT_KEY);
-  auto portStr = node["port"].as<std::string>();
+  checkRequiredField(&node, "url", CONFIG_YAML_PROVENANCE_REPORT_KEY);
```
phrocker: We should maintain prior functionality. Though we're not GA, we should avoid jarring consumers who may be using a prior config. We should able to support both conditionally.

libminifi/src/RemoteProcessorGroupPort.cpp
```diff
@@ -54,19 +57,29 @@ bool create = true) {
...
+      }
     }
+    available_protocols_.try_dequeue(nextProtocol);
   }
   return std::move(nextProtocol);
 }
 
 void RemoteProcessorGroupPort::returnProtocol(std::unique_ptr<Site2SiteClientProtocol> return_protocol) {
-  if (available_protocols_.size_approx() >= max_concurrent_tasks_) {
+  if (available_protocols_.size_approx() >= (max_concurrent_tasks_ * site2site_peer_status_list_.size())) {
```
phrocker: Why should this change? 
With 10 concurrent tasks ( threads ) and 10 peers, you need 100 queued elements? Shouldn't you only need ten 
 ( +/- a few ) since you only have ten threads to pull objects?

libminifi/src/RemoteProcessorGroupPort.cpp
```diff
@@ -54,19 +57,29 @@ bool create = true) {
...
   if (!available_protocols_.try_dequeue(nextProtocol)) {
     if (create) {
+      refreshPeerList();
       // create
-      nextProtocol = std::unique_ptr<Site2SiteClientProtocol>(new Site2SiteClientProtocol(nullptr));
-      nextProtocol->setPortId(protocol_uuid_);
-      std::unique_ptr<org::apache::nifi::minifi::io::DataStream> str = std::unique_ptr<org::apache::nifi::minifi::io::DataStream>(stream_factory_->createSocket(host_, port_));
-      std::unique_ptr<Site2SitePeer> peer_ = std::unique_ptr<Site2SitePeer>(new Site2SitePeer(std::move(str), host_, port_));
-      nextProtocol->setPeer(std::move(peer_));
+      for (auto peer : site2site_peer_status_list_) {
```
phrocker: Could you not create one on demand and simply return it versus enqueueing it and dequeuing it?

libminifi/src/RemoteProcessorGroupPort.cpp
```diff
@@ -76,27 +89,20 @@ void RemoteProcessorGroupPort::returnProtocol(std::unique_ptr<Site2SiteClientPro
...
   std::set<core::Property> properties;
-  properties.insert(hostName);
-  properties.insert(port);
   properties.insert(portUUID);
   setSupportedProperties(properties);
 // Set the supported relationships
   std::set<core::Relationship> relationships;
   relationships.insert(relation);
   setSupportedRelationships(relationships);
+  curl_global_init(CURL_GLOBAL_DEFAULT);
```
phrocker: the function is not thread safe. initialization isn't guaranteed to be thread safe either since other components could initialize concurrently. It's probably fine for now but seems like this should be done in a thread safe way. 
https://curl.haxx.se/libcurl/c/curl_global_init.html

libminifi/src/RemoteProcessorGroupPort.cpp
benqiu2016: i want to prepopulate to reduce the time to establish the socket/negotiation. Also we need to round robin them.

libminifi/src/RemoteProcessorGroupPort.cpp
benqiu2016: i need to round robin the peer even for a single thread to distribute the load cross all peers.

libminifi/src/RemoteProcessorGroupPort.cpp
benqiu2016: OK.

libminifi/src/Site2SiteClientProtocol.cpp
```diff
@@ -38,7 +39,8 @@ namespace minifi {
 
 bool Site2SiteClientProtocol::establish() {
   if (_peerState != IDLE) {
-    logger_->log_error("Site2Site peer state is not idle while try to establish");
+    logger_->log_error(
```
benqiu2016: OK, i will change the format, i want to reduce the line length to make it easy to read

libminifi/src/Site2SiteClientProtocol.cpp
```diff
@@ -278,26 +301,28 @@ bool Site2SiteClientProtocol::handShake() {
...
+    logger_->log_info("Site2Site HandShake Completed");
+    _peerState = HANDSHAKED;
+    return true;
+  case PORT_NOT_IN_VALID_STATE:
+  case UNKNOWN_PORT:
+  case PORTS_DESTINATION_FULL:
+    logger_->log_error(
+        "Site2Site HandShake Failed because destination port is either invalid or full");
+    ret = -1;
+    /*
```
benqiu2016: will remove that.

libminifi/src/RemoteProcessorGroupPort.cpp
```diff
@@ -150,6 +147,87 @@ void RemoteProcessorGroupPort::onTrigger(core::ProcessContext *context, core::Pr
...
+void RemoteProcessorGroupPort::refreshPeerList() {
+  refreshRemoteSite2SiteInfo();
+  if (site2site_port_ == -1)
+    return;
+
+  this->site2site_peer_status_list_.clear();
+
+  std::unique_ptr < Site2SiteClientProtocol> protocol;
+  protocol = std::unique_ptr < Site2SiteClientProtocol
+      > (new Site2SiteClientProtocol(nullptr));
```
benqiu2016: refresh is done once, i would like to avoid hanging a protocol around.

libminifi/src/Site2SiteClientProtocol.cpp
```diff
@@ -319,6 +344,64 @@ void Site2SiteClientProtocol::tearDown() {
...
+
+    uint32_t number;
+    status = peer_->read(number);
+
+    if (status <= 0) {
+      tearDown();
+      return false;
+    }
+
+    for (int i = 0; i < number; i++) {
```
benqiu2016: it is config by the NiFI cluster server to let client evenly distribute the node cross peers.
MiNiFI java is doing the similar thing.

libminifi/src/Site2SiteClientProtocol.cpp
```diff
@@ -344,7 +427,8 @@ int Site2SiteClientProtocol::readRequestType(RequestType &type) {
   return -1;
 }
 
-int Site2SiteClientProtocol::readRespond(RespondCode &code, std::string &message) {
+int Site2SiteClientProtocol::readRespond(RespondCode &code,
```
benqiu2016: i will revert that.

libminifi/src/core/yaml/YamlConfiguration.cpp
benqiu2016: i want to avoid confusion so that we have a consistent way to config the same.

benqiu2016:
@phrocker please review.

libminifi/src/core/yaml/YamlConfiguration.cpp
phrocker: If we tagged today would people who have tried MiNiFi in cpp 0.2.0 be able to use the same config files in cpp 0.3.0?

kevdoran: I agree it would be good to support both. We might be able to support both in the same config loader by preferring the new way and falling back to the old way, or if not we could version the flow file YAML schema similar to the Java implementation of this YAML config loader: https://github.com/apache/nifi-minifi/tree/master/minifi-commons/minifi-commons-schema/src/main/java/org/apache/nifi/minifi/commons/schema

libminifi/src/RemoteProcessorGroupPort.cpp
phrocker: I think the logic works for a single thread, but what happens in the case where you have ten threads, returning at different times? Typical algorithms use multiple queues for fairness.

libminifi/src/RemoteProcessorGroupPort.cpp
phrocker: What is the time to establish sockets?

benqiu2016:
@phrocker please review

libminifi/include/utils/HTTPUtils.h
```diff
@@ -88,6 +90,40 @@ struct HTTPRequestResponse {
 
 };
 
+static void parse_url(std::string &url, std::string &host, int &port, std::string &protocol) {
```
kevdoran: This URL parsing method is a subset of valid URLs. Are we ok with that? For example, it will fail if passed a Literal IPv6 URL as defined by [RFC 3986](https://tools.ietf.org/html/rfc3986) or [RFC 2732](https://tools.ietf.org/html/rfc2732), e.g.:

 `http://[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]:80/nifi`

If we are ok with this parser being designed to work with URLs of a certain format, we should document those assumptions of inputs.

If we would like this part of the code to be a bit more robust for cases we haven't thought of (both present and future), we could consider adding a dependency to a RFC-compliant URI parser, or, if we want to avoid adding a dependency, just expanding the logic to cover more valid URLs.

Lastly, should the url input arg be const and checked for null?

libminifi/include/Site2SiteClientProtocol.h
```diff
@@ -387,6 +387,15 @@ class DataPacket {
...
 };
 
+/**
+  * Site2Site Peer
+  */
+ typedef struct Site2SitePeerStatus {
+   std::string host_;
+   int port_;
+   bool isSecure_;
+ } Site2SitePeerStatus;
```
kevdoran: What about flowFileCount, the number of flow files the peer holds? That is discoverable using the `/site-to-site/peers` endpoint and could be useful if the client wants to use it for load balancing decisions.

libminifi/src/RemoteProcessorGroupPort.cpp
```diff
@@ -150,6 +194,87 @@ void RemoteProcessorGroupPort::onTrigger(core::ProcessContext *context, core::Pr
...
+  }
+  curl_easy_cleanup(http_session);
+}
+
+void RemoteProcessorGroupPort::refreshPeerList() {
+  refreshRemoteSite2SiteInfo();
+  if (site2site_port_ == -1)
+    return;
+
+  this->site2site_peer_status_list_.clear();
```
kevdoran: Can the logic below this fail under certain conditions? If so, is it work waiting to clear the current list until we are sure we have an updated list?

libminifi/src/RemoteProcessorGroupPort.cpp
```diff
@@ -150,6 +194,87 @@ void RemoteProcessorGroupPort::onTrigger(core::ProcessContext *context, core::Pr
   return;
 }
 
+void RemoteProcessorGroupPort::refreshRemoteSite2SiteInfo() {
+  if (this->host_.empty() || this->port_ == -1 || this->protocol_.empty())
+      return;
+
+  std::string fullUrl = this->protocol_ + this->host_ + ":" + std::to_string(this->port_) + "/nifi-api/controller/";
```
kevdoran: If port is not available this should default to 80 for http and 443 for https.

Also, I'm new to the NiFi REST API... what does /nifi-api/controller return? the site2site impl's I see pull from /site-to-site/peers for this info. curious about the difference (will try it out later today)

libminifi/include/RemoteProcessorGroupPort.h
```diff
@@ -84,6 +93,17 @@ class RemoteProcessorGroupPort : public core::Processor {
   void setTransmitting(bool val) {
     transmitting_ = val;
   }
+  // setURL
+  void setURL(std::string val) {
+    url_ = val;
+    utils::parse_url(url_, host_, port_, protocol_);
```
benqiu2016: if the val is empty, the url will be empty and other will empty also.

libminifi/include/utils/HTTPUtils.h
benqiu2016: now it was designed for support ipv4 now. If we need a RFC compliant URL parser, we can certainly use a open source compliant URL parser.

libminifi/include/Site2SiteClientProtocol.h
benqiu2016: yes, there is a flowFileCount, i want to reduce the complexity of looking at flowFileCount and doing round robin. If we need to use flowFileCount, it means that we need refresh the peer status periodically and even that, we can only catch a snapshot of the flowFileCount based on the refreshing interval, it is not accurate.

libminifi/src/RemoteProcessorGroupPort.cpp
benqiu2016: the /nifi-api/controller return 
it return the site2site port and we can use that to do the site2site negotiation with master to find the peer info
{
  "revision": {
    "clientId": "d40fb824-070b-4547-9b1c-f50f5ba0a677"
  },
  "controller": {
    "id": "fe4a3a42-53b6-4af1-a80d-6fdfe60de97f",
    "name": "NiFi Flow",
    "comments": "",
    "runningCount": 3,
    "stoppedCount": 2,
    "invalidCount": 0,
    "disabledCount": 0,
    "inputPortCount": 1,
    "outputPortCount": 1,
    "remoteSiteListeningPort": 10001,
    "siteToSiteSecure": false,
    "instanceId": "9d841c51-ab00-422e-811c-53c6dc2f8e59",
    "inputPorts": [
      {
        "id": "471deef6-2a6e-4a7d-912a-81cc17e3a204",
        "name": " From Node A",
        "comments": "",
        "state": "RUNNING"
      }
    ],
    "outputPorts": [
      {
        "id": "75f88005-0a87-4fef-8320-6219cdbcf18b",
        "name": "To A",
        "comments": "",
        "state": "STOPPED"
      }
    ]
  }
}

libminifi/src/RemoteProcessorGroupPort.cpp
benqiu2016: refreshRemoteSite2SiteInfo can fail.
if it is failed. it will set the this->site2site_port_ = -1;
if that's the case, we do not need to clear the list.

libminifi/include/utils/HTTPUtils.h
kevdoran: IPv6 was just one example. Another case would be username:password@example.com (Not that that is a typical NiFi deployment case, but for the sake of argument as we can't anticipate how these tools will be deployed in every scenario.) I am not saying we necessarily need full RFC URI/L compliance, but I do think whatever is / is not supported in terms of remote s2s urls should be documented somewhere that the users / admins can find it, or logged if parsing fails so that users can more easily discover at runtime why they cannot connect to their IPv6 URL :)

libminifi/src/RemoteProcessorGroupPort.cpp
```diff
@@ -20,6 +20,9 @@
 
 #include "../include/RemoteProcessorGroupPort.h"
 
+#include <curl/curl.h>
+#include <curl/curlbuild.h>
+#include <curl/easy.h>
```
kevdoran: Encapsulating the curl / http communication in this implementation has some pros and cons. On the upside, it abstracts the details of the network communication from the users of the interface, making calling it simple. A downside is that it's harder to unit test the logic in this class as the networking component is not mockable / stub-able without setting up an HTTP server. Also if we ever want to change HTTP client libraries, it is more work as it is tightly coupled here. I don't know if it's worth changing now as it would be a significant refactoring, and it's certainly not just in this case but something that goes for a lot of our code base.  It would be nice if there was more decoupled modularity in these components.

libminifi/include/utils/HTTPUtils.h
benqiu2016: for now, we can document the same in the README.md to specify the URL format if it is OK with you. 
As for the access control, i looked at the doc, it looks like we need to use token, etc for a secure cluster. I will need to ask around.

libminifi/src/RemoteProcessorGroupPort.cpp
benqiu2016: for now, i think we are tied couple curl with our code like others InvokeHttp, etc.

phrocker: InvokeHTTP's sole responsibility is HTTP communications. I don't think the same translates to RemoteProcessGroupPort.

libminifi/src/RemoteProcessorGroupPort.cpp
```diff
@@ -84,19 +107,41 @@ void RemoteProcessorGroupPort::initialize() {
   std::set<core::Relationship> relationships;
   relationships.insert(relation);
   setSupportedRelationships(relationships);
+  curl_global_init(CURL_GLOBAL_DEFAULT);
+  site2site_peer_mutex_.lock();
```
phrocker: why did you choose to use lock over a  lock guard?

phrocker:
Looks good, but am curious about the use of lock vs a stack based lock guard. What was the reason for that?

libminifi/include/utils/HTTPUtils.h
kevdoran: Sure, I think documentation of the YAML format is a good thing to have, and probably also a method comment on the interface for HTTPUtils for future developers to know what format the method expects to be passed.

libminifi/src/RemoteProcessorGroupPort.cpp
kevdoran: Thanks for posting that @benqiu2016. I learned something new about the NiFI REST API :).

Are you planning to add support for default protocol ports?

benqiu2016:
@phrocker if we use a stack based lock guard, in order to control the period of lock, i need to add scope for the lock guard inside {}. it make code less readable.
Anyway. i can change my code to use that if you prefer the same.

libminifi/src/RemoteProcessorGroupPort.cpp
kevdoran: I looked into it further and found out that `/controller` is a NiFi 0.x endpoint. The `/site-to-site` endpoint was introduced in NiFi 1.x, and NiFi 1.x maps `/controller` to `/site-to-site` so that NiFi 0.x instances can communicate with NiFI 1.x instances.

I wasn't sure what minifi-cpp should do. `/controller` currently works for both major versions of NiFi, but there is no guarantee it will forever, and `/controller` is not documented in the 1.x REST API, making the intended behavior of this client code a little less obvious/discoverable for future developers (it was not obvious to me just from the API documentation). The `/site-to-site` endpoint is documented, but would not be available on 0.x NiFi instances.

I decided to look at the minifi-java implementation. It uses the site-to-site client module from NiFi, which has this behavior: It tries `/site-to-site` first, and if it gets a 404 (resource not found) error, then it falls back to `/controller`. From [SiteToSiteRestAPiClient](https://github.com/apache/nifi/blob/master/nifi-commons/nifi-site-to-site-client/src/main/java/org/apache/nifi/remote/util/SiteToSiteRestApiClient.java):

```java
private ControllerDTO getController() throws IOException {
        try {
            HttpGet get = this.createGetControllerRequest();
            return ((ControllerEntity)this.execute(get, ControllerEntity.class)).getController();
        } catch (SiteToSiteRestApiClient.HttpGetFailedException var3) {
            if(404 == var3.getResponseCode()) {
                logger.debug("getController received NOT_FOUND, trying to access the old NiFi version resource url...");
                HttpGet get = this.createGet("/controller");
                return ((ControllerEntity)this.execute(get, ControllerEntity.class)).getController();
            } else {
                throw var3;
            }
        }
    }

    private HttpGet createGetControllerRequest() {
        HttpGet get = this.createGet("/site-to-site");
        get.setHeader("x-nifi-site-to-site-protocol-version", String.valueOf(this.transportProtocolVersionNegotiator.getVersion()));
        return get;
    }
```

That seems like a reasonable approach for minifi-cpp as well. I'm interested in hearing thoughts from others. @apiri - should we try to keep compatibility with 0.x NiFi's in this implementation? If not, we could just go with `/site-to-site` in this implementation.

libminifi/src/RemoteProcessorGroupPort.cpp
kevdoran: I agree this seems to be a good use case for lock_guard as the lock is obtained at the start of initialize() and released just before returning.

benqiu2016:
@phrocker @kevdoran use lock, extract token code, set default port. we need to support /nifi-api/controller which is common to 1.x and 0.x.

kevdoran:
@benqiu2016 thanks for making that round of changes! They look good. I'll be happy to test the branch when a get a chance to setup a NiFi environment that I can use to test some of the more advanced configurations that are supported (clustered peers, secure client). I should have a chance to do that this week, hopefully in the next couple days. As a general comment, it looks like there are quite a few cases that are not covered by corresponding automated tests so a lot of this will rely manual verification. You may want to open a JIRA ticket for adding some additional unit tests and integration tests as a way to verify this feature over time as changes are made to the codebase. Thanks!

benqiu2016:
Thanks for the review. Can we merge the PR?

phrocker:
@benqiu2016 Can you rebase? https://github.com/apache/nifi-minifi-cpp/commit/4042aa0ed2cd16aee740978e3575cc58a7e9cff7 appears to be a merge commit since there are two parents for this commit. Can you rebase and we'll get it merged, thanks.

benqiu2016:
@phrocker used https://github.com/apache/nifi-minifi-cpp/pull/119 which has a single merge.

PR #117, opened 2017-07-05
MINIFI-338: Convert processor threads to use thread pools

phrocker:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/utils/ThreadPool.h
+++ b/libminifi/include/utils/ThreadPool.h
@@ -129,6 +129,10 @@ class Worker {
   virtual uint64_t getTimeSlice() {
     return time_slice_;
   }
+  
+  virtual uint64_t getWaitTime(){
+    return run_determinant_->wait_time();
+  }
 
   Worker<T>(const Worker<T>&) = delete;
   Worker<T>& operator =(const Worker<T>&) = delete;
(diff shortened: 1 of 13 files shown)


libminifi/include/utils/ThreadPool.h
```diff
@@ -246,15 +349,67 @@ void ThreadPool<T>::startWorkers() {
...
+    }
+    // if we have to wait we re-queue the worker.
+    if (wait_to_run) {
+      {
+        std::unique_lock<std::mutex> lock(worker_queue_mutex_);
+        if (!task_status_[task.getIdentifier()]) {
+          continue;
+        }
+      }
+      worker_queue_.enqueue(std::move(task));
```
benqiu2016: do we need to enqueue to head?

libminifi/include/utils/ThreadPool.h
```diff
@@ -246,15 +349,67 @@ void ThreadPool<T>::startWorkers() {
...
   auto waitperiod = std::chrono::milliseconds(1) * 100;
+  uint64_t wait_decay_ = 0;
   while (running_.load()) {
 
+    // if we are spinning, perform a wait. If something changes in the worker such that the timeslice has changed, we will pick that information up. Note that it's possible
+    // we could starve for processing time if all workers are waiting. In the event that the number of workers far exceeds the number of threads, threads will spin and potentially
+    // wait until they arrive at a task that can be run. In this case we reset the wait_decay and attempt to pick up a new task. This means that threads that recently ran should
+    // be more likely to run. This is intentional.
+    if (wait_decay_ > 1000) {
+      std::this_thread::sleep_for(std::chrono::nanoseconds(wait_decay_));
```
benqiu2016: we increase wait_decay if there is not task to run. so the wait_decay may become a very large number if we do not have task to run for a long time.

phrocker: @benqiu2016 thanks. I thought I took care of that.Thanks!

libminifi/include/utils/ThreadPool.h
phrocker: Need? Not really we could run with the same task, but the premise is to enqueue in the event that something else could be pulled off if another task exists, if this one is dequeued, then we run it unless the timeslice has again said "come back later." Admittedly it's waste of a queue, but we won't know if a task is available after the wait period.

benqiu2016: OK. it is possible to sort the queue or somehow to make it such that the head of the queue is the first to expire.
In this case, we can avoid enqueue/dequeue for all the items in the queues.

benqiu2016:
overall looks good. may be some optimization for the queue.
Please some tests in long duration to make sure it is not breaking the master because it is big change.

libminifi/include/utils/ThreadPool.h
phrocker: Unfortunately that would require a locking queue or dequeuing everything in order to sort. Since it is a lock free queue tasks can be enqueued and dequeued with only a std::move with relatively low cost. An alternative to this would be to make multiple queues, but then that would require a prioritized strategy to access each queue. I think that would be a follow on activity if the need arose. 

The biggest negative thus far is that shutdown takes longer because we are now deterministically awaiting threads to end.

phrocker:
@benqiu2016 I've only run this for two hours. It dramatically improved throughput on that test when configuring several concurrent tasks for my threads. Do you have an example flow that you like to run with? I would be happy to run that overnight.

benqiu2016:
@phrocker the normal flow that i run is attached. one get file connected to a RPG

benqiu2016:
[config.txt](https://github.com/apache/nifi-minifi-cpp/files/1166193/config.txt)

phrocker:
@benqiu2016 I've run this for > 24 hrs without issues. Is there additional testing you would like to see?

PR #124, opened 2017-08-08
MINIFI-367 port tests to use boost::filesystem vs. stat.h for better portability

achristianson:
This closes MINIFI-367.


phrocker:
+1

PR #125, opened 2017-08-09
MINIFI-368 exclude hidden files when scanning for src files

achristianson:
Updated regex matching when scanning for source files so that we don't pick up hidden files such as vim swap files.


apiri:
reviewing

apiri:
looks good.  replicated issue with vim editing.  merged.  thanks for fixing

PR #152, opened 2017-10-25
MINIFICPP-52 basic ExtractText processor

calebj:
Regex support blocked by dynamic properties, #37

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/processors/ExtractText.h
+++ b/libminifi/include/processors/ExtractText.h
@@ -1,0 +1,91 @@
+/**
+ * @file ExtractText.h
+ * ExtractText class declaration
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef __EXTRACT_TEXT_H__
+#define __EXTRACT_TEXT_H__
+
+#include "FlowFileRecord.h"
+#include "core/Processor.h"
+#include "core/ProcessSession.h"
+#include "core/Resource.h"
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace processors {
+
+//! ExtractText Class
+class ExtractText : public core::Processor {
+public:
+    //! Constructor
+    /*!
+     * Create a new processor
+     */
+    ExtractText(std::string name, uuid_t uuid = NULL)
+    : Processor(name, uuid)
+    {
+        logger_ = logging::LoggerFactory<ExtractText>::getLogger();
+    }
+    //! Destructor
+    virtual ~ExtractText()
+    {
+    }
+    //! Processor Name
+    static constexpr char const* ProcessorName = "MergeContent";
+    //! Supported Properties
+    static core::Property Attribute;
+    //! Supported Relationships
+    static core::Relationship Success;
+
+    //! OnTrigger method, implemented by NiFi ExtractText
+    virtual void onTrigger(core::ProcessContext *context, core::ProcessSession *session);
+    //! Initialize, over write by NiFi ExtractText
+    virtual void initialize(void);
+
+    class ReadCallback : public InputStreamCallback {
+    public:
+        ReadCallback(std::shared_ptr<core::FlowFile> flowFile, core::ProcessContext *ct);
+        ~ReadCallback() { delete[] _buffer; }
+        int64_t process(std::shared_ptr<io::BaseStream> stream);
+
+    private:
+        std::shared_ptr<logging::Logger> logger_;
+        std::shared_ptr<core::FlowFile> _flowFile;
+        core::ProcessContext *_ctx;
+        uint8_t *_buffer;
+        int64_t _max_read;
+    };
+
+protected:
+
+private:
+    //! Logger
+    std::shared_ptr<logging::Logger> logger_;
+};
+
+REGISTER_RESOURCE(ExtractText);
+
+} /* namespace processors */
+} /* namespace minifi */
+} /* namespace nifi */
+} /* namespace apache */
+} /* namespace org */
+
+#endif
--- a/libminifi/include/processors/ExtractText.h
+++ b/libminifi/include/processors/ExtractText.h
@@ -38,39 +38,34 @@ public:
     /*!
      * Create a new processor
      */
-    ExtractText(std::string name, uuid_t uuid = NULL)
+    explicit ExtractText(std::string name, uuid_t uuid = nullptr)
     : Processor(name, uuid)
     {
         logger_ = logging::LoggerFactory<ExtractText>::getLogger();
     }
-    //! Destructor
-    virtual ~ExtractText()
-    {
-    }
     //! Processor Name
-    static constexpr char const* ProcessorName = "MergeContent";
+    static constexpr char const* ProcessorName = "ExtractText";
     //! Supported Properties
     static core::Property Attribute;
     //! Supported Relationships
     static core::Relationship Success;
 
     //! OnTrigger method, implemented by NiFi ExtractText
-    virtual void onTrigger(core::ProcessContext *context, core::ProcessSession *session);
+    void onTrigger(core::ProcessContext *context, core::ProcessSession *session);
     //! Initialize, over write by NiFi ExtractText
-    virtual void initialize(void);
+    void initialize(void);
 
     class ReadCallback : public InputStreamCallback {
     public:
         ReadCallback(std::shared_ptr<core::FlowFile> flowFile, core::ProcessContext *ct);
-        ~ReadCallback() { delete[] _buffer; }
+        ~ReadCallback() { delete[] buffer_; }
         int64_t process(std::shared_ptr<io::BaseStream> stream);
 
     private:
-        std::shared_ptr<logging::Logger> logger_;
-        std::shared_ptr<core::FlowFile> _flowFile;
-        core::ProcessContext *_ctx;
-        uint8_t *_buffer;
-        int64_t _max_read;
+        std::shared_ptr<core::FlowFile> flowFile_;
+        core::ProcessContext *ctx_;
+        uint8_t *buffer_;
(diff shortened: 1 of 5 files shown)


libminifi/src/processors/ExtractText.cpp
```diff
@@ -0,0 +1,105 @@
...
+    std::string contentStr;
+
+    while (read_size < _flowFile->getSize()) {
+        ret = stream->read(_buffer, _max_read);
+        if (ret < 0) {
+            return -1;
+        }
+
+        if (ret > 0) {
+            contentStream.write(reinterpret_cast<const char*>(_buffer), ret);
```
phrocker: I see that this appears to be a partial implementation of the JAVA class; however, one component that I think is necessary will be the limits on write. As it stands, the stringstream is unbounded based on the size of the input file. Don't you think we should bound or cap this with a configurable limit to avoid memory issues when users aren't always aware of the size of the content coming through?

libminifi/src/processors/ExtractText.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include "core/ProcessSession.h"
+#include "core/FlowFile.h"
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace processors {
+
+core::Property ExtractText::Attribute("Attribute", "Attribute to set from content (TEMPORARY)", "");
```
phrocker: what does TEMPORARY mean in this context?

libminifi/src/processors/ExtractText.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+    contentStr = contentStream.str();
+    _flowFile->setAttribute(attrKey, contentStr);
+    return read_size;
+}
+
+ExtractText::ReadCallback::ReadCallback(std::shared_ptr<core::FlowFile> flowFile, core::ProcessContext *ctx) {
+    _max_read = getpagesize() * sizeof(uint8_t);
+    _buffer = new uint8_t[_max_read];
+    logger_ = logging::LoggerFactory<ReadCallback>::getLogger();
```
phrocker: Are you using the logger in this class? If not, is there a reason to instantiate the facilities in the logger?

libminifi/include/processors/ExtractText.h
```diff
@@ -0,0 +1,91 @@
...
+
+    class ReadCallback : public InputStreamCallback {
+    public:
+        ReadCallback(std::shared_ptr<core::FlowFile> flowFile, core::ProcessContext *ct);
+        ~ReadCallback() { delete[] _buffer; }
+        int64_t process(std::shared_ptr<io::BaseStream> stream);
+
+    private:
+        std::shared_ptr<logging::Logger> logger_;
+        std::shared_ptr<core::FlowFile> _flowFile;
```
phrocker: ^our convention is the google code style, so this would be ctx_ or buffer_

libminifi/src/processors/ExtractText.cpp
calebj: Not sure, ask @achristianson. I got it from [here](https://github.com/NiFiLocal/nifi-minifi-cpp/commit/ec03fd15f80e33f1dba3be9c1e2520745e732930#diff-3d4ecb8ad9604bc01f1c18e283d8bc4fR31).

libminifi/src/processors/ExtractText.cpp
calebj: Sure. Should it be a constant somewhere, or a property?

libminifi/src/processors/ExtractText.cpp
achristianson: @calebj @phrocker In the [reference NiFi ExtractText](https://nifi.apache.org/docs/nifi-docs/components/org.apache.nifi/nifi-standard-nar/1.4.0/org.apache.nifi.processors.standard.ExtractText/index.html), the attribute to set from the extracted text, as well as the regular expression to match, is set with dynamic properties. Until MiNiFi - C++ supports dynamic properties, the intent was for ExtractText to support one regular expression and one attribute. The "TEMPORARY," if I recall correctly, was to document that this property is liable to be deprecated in the future. Looking back at it, we should probably clarify this, or remove the text altogether. Also suggest we prioritize the dynamic attributes feature.

libminifi/include/processors/ExtractText.h
```diff
@@ -0,0 +1,91 @@
...
+    : Processor(name, uuid)
+    {
+        logger_ = logging::LoggerFactory<ExtractText>::getLogger();
+    }
+    //! Destructor
+    virtual ~ExtractText()
+    {
+    }
+    //! Processor Name
+    static constexpr char const* ProcessorName = "MergeContent";
```
achristianson: This should say ExtractText, not MergeContent.

libminifi/include/processors/ExtractText.h
achristianson: In similar vein, the indentation needs to be changed to Google style.

libminifi/include/processors/ExtractText.h
```diff
@@ -0,0 +1,91 @@
...
+    //! Supported Properties
+    static core::Property Attribute;
+    //! Supported Relationships
+    static core::Relationship Success;
+
+    //! OnTrigger method, implemented by NiFi ExtractText
+    virtual void onTrigger(core::ProcessContext *context, core::ProcessSession *session);
```
achristianson: Suggest we use override instead of virtual, to be more modern/idiomatic c++.

libminifi/include/processors/ExtractText.h
```diff
@@ -0,0 +1,91 @@
...
+     */
+    ExtractText(std::string name, uuid_t uuid = NULL)
+    : Processor(name, uuid)
+    {
+        logger_ = logging::LoggerFactory<ExtractText>::getLogger();
+    }
+    //! Destructor
+    virtual ~ExtractText()
+    {
+    }
```
achristianson: Clang-tidy suggests using override here. Do we even need this destructor?

libminifi/include/processors/ExtractText.h
```diff
@@ -0,0 +1,91 @@
...
+namespace processors {
+
+//! ExtractText Class
+class ExtractText : public core::Processor {
+public:
+    //! Constructor
+    /*!
+     * Create a new processor
+     */
+    ExtractText(std::string name, uuid_t uuid = NULL)
```
achristianson: Clang-tidy suggests marking as explicit since this can be called with a single arg.

achristianson: Also should use nullptr rather than NULL.

libminifi/src/processors/ExtractText.cpp
```diff
@@ -0,0 +1,105 @@
...
+    _flowFile = flowFile;
```
achristianson: Suggest we use a [constructor initializer list](http://en.cppreference.com/w/cpp/language/initializer_list) to assign _flowFile as well as other members where possible. Also needs google style here.

libminifi/src/processors/ExtractText.cpp
achristianson: I think we should add that as a processor property. Something like "Size Limit" with a good description so that the user becomes conscious of the cost.

libminifi/test/unit/ExtractTextTest.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+    plan->runNextProcessor();  // GetFile
+    plan->runNextProcessor();  // ExtractText
+    plan->runNextProcessor();  // LogAttribute
+
+    std::stringstream ss2;
+    ss2 << "key:" << TEST_ATTR << " value:" << TEST_TEXT;
+    std::string log_check = ss2.str();
+
+    REQUIRE(true == LogTestController::getInstance().contains(log_check));
```
achristianson: Suggest we simplify to
```c++
REQUIRE(LogTestController::getInstance().contains(log_check));
```

libminifi/test/unit/ExtractTextTest.cpp
```diff
@@ -0,0 +1,101 @@
...
+const char* TEST_TEXT = "Test text\n";
+const char* TEST_FILE = "test_file.txt";
+const char* TEST_ATTR = "ExtractedText";
+
+TEST_CASE("Test Creation of ExtractText", "[extracttextCreate]") {
+    TestController testController;
+    std::shared_ptr<core::Processor> processor = std::make_shared<org::apache::nifi::minifi::processors::ExtractText>("processorname");
+    REQUIRE(processor->getName() == "processorname");
+    uuid_t processoruuid;
+    REQUIRE(true == processor->getUUID(processoruuid));
```
achristianson: Suggest we simplify to

```c++
REQUIRE(processor->getUUID(processoruuid));
```

libminifi/test/unit/ExtractTextTest.cpp
```diff
@@ -0,0 +1,101 @@
...
+        test_file << TEST_TEXT << std::endl;
+        test_file.close();
+    }
+
+    plan->runNextProcessor();  // GetFile
+    plan->runNextProcessor();  // ExtractText
+    plan->runNextProcessor();  // LogAttribute
+
+    std::stringstream ss2;
+    ss2 << "key:" << TEST_ATTR << " value:" << TEST_TEXT;
```
achristianson: Good.

achristianson:
@calebj Thanks for updating all this, adding the test, and opening the PR. Overall looks good, mostly stylistic/idiom issues due to the style I was using at the time.

libminifi/include/processors/ExtractText.h
```diff
@@ -0,0 +1,91 @@
...
+    public:
+        ReadCallback(std::shared_ptr<core::FlowFile> flowFile, core::ProcessContext *ct);
+        ~ReadCallback() { delete[] _buffer; }
+        int64_t process(std::shared_ptr<io::BaseStream> stream);
+
+    private:
+        std::shared_ptr<logging::Logger> logger_;
+        std::shared_ptr<core::FlowFile> _flowFile;
+        core::ProcessContext *_ctx;
+        uint8_t *_buffer;
```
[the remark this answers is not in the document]
achristianson: @phrocker Good idea. Doesn't seem to be much of a performance impact and improves the memory safety characteristics.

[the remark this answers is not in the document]
calebj: I wasn't able to get `stream->readData(buf_vec)` working. It appears to fill the vector with null bytes. I'll take care of the other things and try it again from scratch.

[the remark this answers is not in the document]
achristianson: Interesting. Maybe that's a bug. Although I think what you did should work, you could try readData(&buf_vec[0], size) (might need some casting).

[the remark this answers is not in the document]
calebj: I was finally able to get it working. readData doesn't complain, but I think I'll stick to calling `write()` on the block directly. The alternative is to use
```
std::copy(buffer_.begin(), buffer_.end(), std::ostream_iterator<uint8_t>(contentStream));
```
which checks each character individually before copying, instead of doing it as one block.

PR #158, opened 2017-10-28
MINIFICPP-60: Add initial implementation of Site to Site changes.

phrocker:
MINIFICPP-60: Update interfaces to allow for seamless extension

MINIFICPP-60: Update functions to accomodate changed signatures

### For all changes:
     in the commit message?

### For code changes:

### Note:


phrocker:
I'll be going through this myself over the next few days as well, but I like to leave the comments for others to see on the PR

extensions/http-curl/client/HTTPClient.cpp
```diff
@@ -175,19 +201,53 @@ bool HTTPClient::submit() {
...
-
+  if (callback == nullptr) {
+    content_.ptr = &read_callback_;
+    curl_easy_setopt(http_session_, CURLOPT_WRITEFUNCTION, &utils::HTTPRequestResponse::recieve_write);
+    curl_easy_setopt(http_session_, CURLOPT_WRITEDATA, static_cast<void*>(&content_));
+  }
   curl_easy_setopt(http_session_, CURLOPT_HEADERFUNCTION, &utils::HTTPHeaderResponse::receive_headers);
   curl_easy_setopt(http_session_, CURLOPT_HEADERDATA, static_cast<void*>(&header_response_));
-
+  curl_easy_setopt(http_session_, CURLOPT_TCP_KEEPALIVE, 0L);
```
phrocker: This shouldn't be necessary.

libminifi/test/curl-tests/sitetositehttp/HTTPHandlers.h
```diff
@@ -0,0 +1,322 @@
...
+
+  explicit DeleteTransactionResponder(const std::string base_url, std::string response_code, moodycamel::ConcurrentQueue<std::shared_ptr<FlowObj>> *feed)
+      : base_url(base_url),
+        response_code(response_code),
+        flow_files_feed_(feed) {
+  }
+
+  bool handleDelete(CivetServer *server, struct mg_connection *conn) {
+
+    std::string site2site_rest_resp = "";
```
phrocker: Should do some additional verification here if possible.

achristianson:
Nothing glaring on initial glance. There are a lot of changes. Can you summarize generally what the changes are for? Just want to get the right understanding to put it into context.

extensions/http-curl/client/HTTPCallback.h
```diff
@@ -0,0 +1,187 @@
...
+    std::vector<char> vec;
+
+    if (size > 0) {
+      vec.resize(size);
+
+      memcpy(vec.data(), vector, size);
+
+      size_t added_size = vec.size();
+
+      byte_arrays_.enqueue(std::move(vec));
```
minifirocks: do we need to lock the mutex_ before enqueue?

extensions/http-curl/client/HTTPCallback.h
```diff
@@ -0,0 +1,187 @@
...
+    return rolling_count_;
+  }
+
+ private:
+
+  inline void load_buffer() {
+    std::unique_lock<std::recursive_mutex> lock(mutex_);
+    cv.wait(lock, [&] {return byte_arrays_.size_approx() > 0 || is_alive_==false;});
+    if (!is_alive_ && byte_arrays_.size_approx() == 0) {
+      return;
```
minifirocks: unlock the mutex before return?

phrocker: wait releases the mutex

extensions/http-curl/client/HTTPCallback.h
phrocker: byte_arrays_ is a lock free queue. There can be concerns with using lock free queues like moodycamel, particularly with ordering, but I've mitigated this by how I'm using it.

extensions/http-curl/client/HTTPCallback.h
minifirocks: i thought after wait fired, we own the lock, that's reason you did unlock afterwards.

phrocker: Sorry I thought you meant in the wait, assuming it can't get the lock. My apologies I misunderstood.

I didn't put a release there since every other usage of the lock would free and we wouldn't continue operations if we have no data or are not alive; however, I should be consistent.

extensions/http-curl/sitetosite/HTTPProtocol.cpp
```diff
@@ -0,0 +1,312 @@
...
+}
+
+int HttpSiteToSiteClient::readRespond(const std::shared_ptr<Transaction> &transaction, RespondCode &code, std::string &message) {
+  if (current_code == FINISH_TRANSACTION) {
+
+    if (transaction->getDirection() == SEND) {
+      auto stream = dynamic_cast<io::HttpStream*>(peer_->getStream());
+      stream->closeStream();
+      auto client = stream->getClient();
+      logger_->log_info("Faking it till I makin it 2, response code %d", client->getResponseCode());
```
minifirocks: the log info print is not clear

extensions/http-curl/sitetosite/HTTPProtocol.cpp
```diff
@@ -0,0 +1,312 @@
...
+      if (client->getResponseCode() == 202) {
+        code = CONFIRM_TRANSACTION;
+        message = std::string(client->getResponseBody().data(), client->getResponseBody().size());
+      } else {
+        code = UNRECOGNIZED_RESPONSE_CODE;
+      }
+      return 1;
```
minifirocks: return -1 for failure

extensions/http-curl/sitetosite/HTTPProtocol.cpp
```diff
@@ -0,0 +1,312 @@
...
+    return 1;
+  }
+  return SiteToSiteClient::readRespond(transaction, code, message);
+
+}
+// write respond
+int HttpSiteToSiteClient::writeRespond(const std::shared_ptr<Transaction> &transaction, RespondCode code, std::string message) {
+  current_code = code;
+  if (code == CONFIRM_TRANSACTION || code == FINISH_TRANSACTION) {
+    logger_->log_info("Faking it till I makin it");
```
minifirocks: log print is not clear

extensions/http-curl/sitetosite/HTTPProtocol.cpp
phrocker: If we return -1 we won't have an opportunity to provide an error message. -1 would be for the case where there is a comms error. here we have no need to read and write for respond. But you did remind me to change that name to readResponse.

extensions/http-curl/client/HTTPClient.cpp
```diff
@@ -204,7 +264,9 @@ const char *HTTPClient::getContentType() {
 }
 
 const std::vector<char> &HTTPClient::getResponseBody() {
```
achristianson: c++11+ guarantees std::string bytes are sequential. May not need a vector<char> in lieu of string.

achristianson: Not in these changes, but wanted to note...

extensions/http-curl/processors/InvokeHTTP.cpp
```diff
@@ -33,6 +33,7 @@
 #include <utility>
 #include <vector>
 
+#include "../../../libminifi/include/utils/ByteArrayCallback.h"
```
achristianson: Awkward include path. Perhaps include path could be set better in CMakeLists.txt?

extensions/http-curl/processors/InvokeHTTP.h
```diff
@@ -23,14 +23,14 @@
 #include <regex>
 
 #include <curl/curl.h>
+#include "../../../libminifi/include/utils/ByteArrayCallback.h"
```
achristianson: Ditto.

extensions/http-curl/processors/InvokeHTTP.cpp
phrocker: cmake lists? huh?

no, that's eclipse. it does that to me. thanks!

extensions/http-curl/protocols/RESTSender.h
```diff
@@ -22,12 +22,13 @@
 #include "json/writer.h"
 #include <string>
 #include <mutex>
+
+#include "../../../libminifi/include/utils/ByteArrayCallback.h"
```
achristianson: Ditto.

extensions/http-curl/client/HTTPClient.cpp
phrocker: yes and it changes again in 17...some not for the best unfortunately.

herb sutter has some really good talks about small string optimization and how you shouldn't fight it, but I'm not too certain we can really avoid the pessimization.  what are your thoughts on the SOO causing pessimization against strings that are larger than the buffer on things like sizeof, etc? I've seen profiles where it can really cause unnecessary havoc.

extensions/http-curl/protocols/RESTSender.h
phrocker: oh man sometimes eclipse really does this to me. thanks for pointing these out. I should really disable that feature

extensions/http-curl/sitetosite/HTTPProtocol.cpp
```diff
@@ -0,0 +1,310 @@
+/**
```
achristianson: General organizational note --it seems to be a bit of a mixing of concerns to have s2s-specific code under the http-curl extension.

Granted, we probably have/will have dependencies between extensions which may or may not be easy to express using cmake constructs.

phrocker: this HTTPProtocol implementation uses CURL to do its work, hence why it exists here. While we could certainly have cross compilation between extensions, it doesn't seem like it's worth the increase in build time to change. One thing I want to avoid is having single function extensions and instead of have classes of extensions.

extensions/http-curl/sitetosite/HTTPProtocol.h
```diff
@@ -0,0 +1,197 @@
...
+        logger_(logging::LoggerFactory<HttpSiteToSiteClient>::getLogger()) {
+    peer_ = std::move(peer);
+    peer_state_ = READY;
+  }
+  // Destructor
+  virtual ~HttpSiteToSiteClient() {
+
+  }
+
+  void setPeer(std::unique_ptr<SiteToSitePeer> peer) {
```
achristianson: setPeer is not virtual.

extensions/http-curl/sitetosite/HTTPProtocol.h
```diff
@@ -0,0 +1,197 @@
...
+    peer_ = std::move(peer);
+  }
+
+  virtual bool getPeerList(std::vector<PeerStatus> &peers);
```
achristianson: New c++ idiom is to mark this override rather than virtual. Helps better express intent and catch bugs at compile time.

phrocker: Good point. I don't want to yet do that in core but the intent was to do that in extensions. Thanks for bringing that up.

extensions/http-curl/sitetosite/HTTPProtocol.h
```diff
@@ -0,0 +1,197 @@
...
+    uri.append(std::to_string(peer_->getPort()));
+    uri.append("/nifi-api/");
+    return uri;
+  }
+
+  virtual void tearDown();
+
+  const std::string parseTransactionId(const std::string &uri);
+
+  std::unique_ptr<utils::HTTPClient> create_http_client(const std::string &uri, const std::string method = "POST", bool setPropertyHeaders = false) {
```
achristianson: Make method a reference?

achristianson:
+1

minifirocks:
@phrocker please resolve the conflict and merge into simple commit, update readme for the same.

PR #181, opened 2017-11-09
MINIFICPP-297: Remove Boost dependencies since we have very little re…

phrocker:
…liance on it

MINIFICPP-297: Ensure posix compliance

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/utils/file/FileManager.h
+++ b/libminifi/include/utils/file/FileManager.h
@@ -1,0 +1,113 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef LIBMINIFI_INCLUDE_UTILS_FILEMANAGER_H_
+#define LIBMINIFI_INCLUDE_UTILS_FILEMANAGER_H_
+
+#ifdef BOOST_VERSION
+#include <boost/filesystem.hpp>
+#else
+#include <cstdlib>
+#endif
+#include <cstdio>
+#include <unistd.h>
+#include <fcntl.h>
+#include "io/validation.h"
+#include "utils/Id.h"
+#include "utils/StringUtils.h"
+#ifdef WIN32
+#define stat _stat
+#endif
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace utils {
+namespace file {
+
+/**
+ * Simple implementation of simple file manager utilities.
+ *
+ * unique_file is not a static implementation so that we can support scope driven temporary files.
+ */
+class FileManager {
+ public:
+
+  FileManager() {
+  }
+
+  ~FileManager() {
+    for (auto file : unique_files_) {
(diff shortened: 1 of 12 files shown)


phrocker:
Still doing some validation on older *nix releases to validate that the stat structure is the same everywhere. In some cases it will not be, so need to finish validation against all releases.

achristianson:
This looks great. My only critique is that FilesystemUtils is used in a stateful way (presumably just to keep a cache of previous unique files?) I think keeping it entirely stateless would be preferable. Looking at the same issue, keeping a stateful instance of "utilities" isn't very meaningful. If part of it really needs to be stateful, I would consider making a nested class in FilesystemUtils i.e. FilesystemUtils::UniqueFileGenerator.

Overall the change is great and removes a huge dependency without adding a lot of complexity.

phrocker:
@achristianson statefulness helps use RAII to eliminate temp files, whereas other techniques don't do this. It's a balance between cleaning them up automatically or requiring the user to. As I saw previously with the code I removed, this was not done, and I fear most users won't properly do this. I'm with you. My initial implementation was just a simple create file without tracking. I'm still on the fence about that. We should require users to remove files, but that's not an expectation from creating temp files.

phrocker:
@achristianson what are your thoughts on that? I'm so on the fence I need a second opinion about that.

libminifi/include/utils/FileSystemUtils.h
```diff
@@ -0,0 +1,122 @@
...
+  std::string unique_file(const std::string &location, bool keep = false) {
+
+    if (!IsNullOrEmpty(location)) {
+      std::string file_name = location + "/" + non_repeating_string_generator_.generate();
+      while (!verify_not_exist(file_name)) {
+        file_name = location + "/" + non_repeating_string_generator_.generate();
+      }
+      if (!keep)
+        unique_files_.push_back(file_name);
+      return file_name;
```
achristianson: Return std::move()?

libminifi/include/utils/FileSystemUtils.h
```diff
@@ -0,0 +1,122 @@
...
+    } else {
+      std::string file_name = "/tmp/" + non_repeating_string_generator_.generate();
+      while (!verify_not_exist(file_name)) {
+        file_name = "/tmp/" + non_repeating_string_generator_.generate();
+      }
+      if (!keep)
+        unique_files_.push_back(file_name);
+      return file_name;
```
achristianson: Return std::move(file_name)?

achristianson:
@phrocker my thought is to definitely keep the RAII to have simpler/safer resource management. I see the unlink in the destructor. I would just tweak the naming & organization a bit. Now I see this is actually a unique *temporary* file session, so I would put that in the name. I wouldn't get rid of FileSystemUtils, but would create a nested class i.e. FileSystemUtils::TempFileManager. Usage like this:

```c++
{
  FileSystemUtils::TempFileManager mgr;
  auto tmp_file = mgr.unique_file("/tmp", false);
  // Do something with tmp_file
} // All tmp files created with mgr are cleaned up (RAII)
```

libminifi/include/utils/FileSystemUtils.h
phrocker: We cannot since it may go into a map. Then we move conditionally and that's generally not an accepted move.

phrocker: Sorry a better way to put my previous statement is that I'm relying on CE and RVO, especially with the push_back. It is an optimization, but one the compiler can choose.

libminifi/include/utils/file/FileManager.h
```diff
@@ -0,0 +1,113 @@
...
+  }
+  std::string unique_file(const std::string &location, bool keep = false) {
+
+    if (!IsNullOrEmpty(location)) {
+      std::string file_name = location + "/" + non_repeating_string_generator_.generate();
+      while (!verify_not_exist(file_name)) {
+        file_name = location + "/" + non_repeating_string_generator_.generate();
+      }
+      if (!keep)
+        unique_files_.push_back(file_name);
```
minifirocks: why we need to do unlink later in the destructor

phrocker: These are temporary files. If we don't specify keep it will be a temp file.

PR #183, opened 2017-11-12
MINIFICPP-301 Added initial implementation of TFApplyGraph which appl…

achristianson:
…ies arbitrary TensorFlow graphs to tensor protocol bufferss in the data flow

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -166,6 +166,13 @@ if (ENABLE_USB_CAMERA)
     createExtension(USB-CAMERA-EXTENSIONS "USB CAMERA EXTENSIONS" "This enables USB camera support" "extensions/usb-camera" "${TEST_DIR}/usb-camera-tests" "TRUE" "thirdparty/libuvc-0.0.6")
 endif()
 
+## TensorFlow extensions
+## Disabled by default because TF can be complex/environment-specific to build
+option(ENABLE_TENSORFLOW "Disables the TensorFlow extensions." OFF)
+if (ENABLE_TENSORFLOW)
+    createExtension(TENSORFLOW-EXTENSIONS "TENSORFLOW EXTENSIONS" "This enables TensorFlow support" "extensions/tensorflow" "${TEST_DIR}/tensorflow-tests")
+endif()
+
 ## NOW WE CAN ADD LIBRARIES AND EXTENSIONS TO MAIN
 
 add_subdirectory(main)
--- a/extensions/tensorflow/TFApplyGraph.cpp
+++ b/extensions/tensorflow/TFApplyGraph.cpp
@@ -1,0 +1,227 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "TFApplyGraph.h"
+
+#include "tensorflow/cc/ops/standard_ops.h"
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace processors {
+
+core::Property TFApplyGraph::InputNode(  // NOLINT
+    "Input Node",
+    "The node of the TensorFlow graph to feed tensor inputs to", "");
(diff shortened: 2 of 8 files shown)


achristianson:
Looks like I need to touch up the createExtension to account for the latest changes.

achristianson:
Fixed & ready for review.

CMakeLists.txt
```diff
@@ -155,6 +155,13 @@ if (ENABLE_USB_CAMERA)
     createExtension(USB-CAMERA-EXTENSIONS "USB CAMERA EXTENSIONS" "This enables USB camera support" "extensions/usb-camera" "${TEST_DIR}/usb-camera-tests" "TRUE" "thirdparty/libuvc-0.0.6")
 endif()
 
+## TensorFlow extensions
+## Disabled by default because TF can be complex/environment-specific to build
+option(ENABLE_TENSORFLOW "Disables the TensorFlow extensions." OFF)
+if (ENABLE_TENSORFLOW)
+    createExtension(TENSORFLOW-EXTENSIONS "TENSORFLOW EXTENSIONS" "This enables TensorFlow support" "extensions/tensorflow" "${TEST_DIR}/tensorflow-tests")
```
minifirocks: does TF provide source built as third party library. in case we want to build for other target like ARM, etc

achristianson: This extension should be compatible with that platform. The key thing would be building TF on (or for, with cross-compiling) the target.

We are dynamically linking TF so it just has to be available in the target linking environment.

As for build time, our custom FindTensorFlow.cmake needs to find the libtensorflow_cc.so and headers.

Ultimately, it would be nice to have the ability to build TF in thirdparty/ and link statically, but there's currently no way to static link TF without giving up important features such as GPU acceleration.

My thought was to keep it as an externally-provided and dynamically-linked library, and keep an eye on libtensorflow_cc as it matures and hopefully supports static linking options and cleaner embedded cmake files we could put in thirdparty/.

achristianson: @phrocker also look at BUILDING.md if you haven't looked at it yet. While there are only instructions for CentOS 7 currently, the procedure should be similar for other environments (installing TF-deps like CUDA/cuDNN, building & installing TF to /usr/local).

extensions/tensorflow/TFApplyGraph.cpp
```diff
@@ -0,0 +1,227 @@
...
+    std::string tf_type;
+    flow_file->getAttribute("tf.type", tf_type);
+
+    std::shared_ptr<tensorflow::GraphDef> graph_def;
+    uint32_t graph_version;
+
+    {
+      std::lock_guard<std::mutex> guard(graph_def_mtx_);
+
+      if ("graph" == tf_type) {
```
minifirocks: after we reload the graph, do we need to clean the context which associated with the old graph

extensions/tensorflow/TFApplyGraph.h
```diff
@@ -0,0 +1,117 @@
...
+namespace processors {
+
+class TFApplyGraph : public core::Processor {
+ public:
+  explicit TFApplyGraph(const std::string &name, uuid_t uuid = nullptr)
+      : Processor(name, uuid),
+        logger_(logging::LoggerFactory<TFApplyGraph>::getLogger()) {
+  }
+
+  static void hello();
```
minifirocks: unused hello

extensions/tensorflow/TFApplyGraph.cpp
achristianson: Later down, when it pulls the context off the queue, it lets it expire if the graph is outdated:

```c++
      if (ctx->graph_version != graph_version) {
         logger_->log_info("Allowing session with stale graph to expire");
         ctx = nullptr;
       }
```

All resources should be freed. Does that answer your question?

extensions/tensorflow/TFApplyGraph.cpp
```diff
@@ -0,0 +1,227 @@
...
+  context->getProperty(OutputNode.getName(), output_node_);
+
+  if (output_node_.empty()) {
+    logger_->log_error("Invalid output node");
+  }
+}
+
+void TFApplyGraph::onTrigger(const std::shared_ptr<core::ProcessContext> &context,
+                             const std::shared_ptr<core::ProcessSession> &session) {
+  auto flow_file = session->get();
```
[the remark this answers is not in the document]
achristianson: Good question. Whatever is feeding this processor could simply fork its output relationship to another processor to preserve the input. Another alternative is to add an "original" relationship to this processor, which follows a common design pattern from NiFi processors.

In my opinion, for now, we should just allow this processor to be destructive with its input, and expect the user to fork off the original upstream of the processor if they have a need to do so.

When we flesh out our user docs for the minifi-cpp processors, we can detail this behavior as well as common use cases, such as loading a graph from the filesystem with a GetFile. The "success" relationship of said GetFile can be sent anywhere to preserve the original graph. Likewise for tensor inputs.

extensions/tensorflow/TFApplyGraph.h
achristianson: Good catch. Will fix.

minifirocks:
@achristianson please merge to one commit

achristianson:
@minifirocks done.

PR #188, opened 2017-11-14
MINIFICPP-49 Added initial implementation of NiFi expression language

achristianson:
This implementation covers core language syntax such as embedding ${expressions}, referencing attributes, calling functions ${expr:toUpper()}, function arguments, quoted attribute IDs.

It has a very minimal implementation of functions (hostname() and toUpper()), and minimal integration with processor properties (PutFile directory and InvokeHTTP URL). The intent is to expand coverage and integration over future commits.

It is implemented as an extension and can be entirely disabled at compile time with no memory or compute impact in the resulting binary when compared with before this PR. The cmake code is somewhat, but not much, different than other extensions due to the API touching ProcessContext.

--

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,10 @@ Perspectives of the role of MiNiFi should be from the perspective of the agent a
   * 4.8.4 or greater
 * g++
   * 4.8.4 or greater
+* bison
+  * 3.0 or greater
+* flex
+  * 2.5 or greater
   
 **NOTE** if Lua support is enabled, then a C++ compiler with support for c++-14 must be used. If using GCC, version 6.x
 or greater is recommended.
@@ -121,6 +125,8 @@ using a devtools-* package from the Software Collections (SCL).
 # ~/Development/code/apache/nifi-minifi-cpp on git:master
 $ yum install cmake \
   gcc gcc-c++ \
+  bison \
+  flex \
   libcurl-devel \
   rocksdb-devel rocksdb \
   libuuid libuuid-devel \
@@ -150,6 +156,8 @@ Aptitude based Linux Distributions
 # ~/Development/code/apache/nifi-minifi-cpp on git:master
 $ apt-get install cmake \
   gcc g++ \
+  bison \
+  flex \
   libcurl-dev \
   librocksdb-dev librocksdb4.1 \
   uuid-dev uuid \
@@ -176,6 +184,8 @@ OS X Using Homebrew (with XCode Command Line Tools installed)
 ```
 # ~/Development/code/apache/nifi-minifi-cpp on git:master
 $ brew install cmake \
+  bison \
+  flex \
   rocksdb \
   ossp-uuid \
   boost \
--- a/extensions/expression-language/Expression.cpp
+++ b/extensions/expression-language/Expression.cpp
@@ -1,0 +1,164 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <utility>
+#include <iostream>
+
+#include <expression/Expression.h>
+#include "Driver.h"
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace expression {
+
+Expression compile(const std::string &expr_str) {
+  std::stringstream expr_str_stream(expr_str);
+  Driver driver(&expr_str_stream);
+  Parser parser(&driver);
+  parser.parse();
+  return driver.result;
+}
+
+Expression make_static(std::string val) {
+  return Expression(std::move(val));
+}
+
+Expression make_dynamic(std::function<std::string(const Parameters &params)> val_fn) {
+  return Expression("", std::move(val_fn));
+}
+
+Expression make_dynamic_attr(const std::string &attribute_id) {
+  return make_dynamic([attribute_id](const Parameters &params) -> std::string {
+    std::string result;
+    params.flow_file.lock()->getAttribute(attribute_id, result);
+    return result;
+  });
+}
+
+std::string expr_hostname(const std::vector<std::string> &args) {
+  char hostname[1024];
+  hostname[1023] = '\0';
+  gethostname(hostname, 1023);
+  return std::string(hostname);
+}
+
+std::string expr_toUpper(const std::vector<std::string> &args) {
+  std::string result = args[0];
+  std::transform(result.begin(), result.end(), result.begin(), ::toupper);
+  return result;
+}
+
+template<std::string T(const std::vector<std::string> &)>
+Expression make_dynamic_function_incomplete(const std::string &function_name,
+                                            const std::vector<Expression> &args,
+                                            std::size_t num_args) {
+  if (args.size() == num_args) {
+    return make_dynamic([=](const Parameters &params) -> std::string {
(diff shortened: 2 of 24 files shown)


extensions/expression-language/ProcessContextExpr.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace core {
+
+bool ProcessContext::getProperty(const std::string &name, std::string &value,
+                                 const std::shared_ptr<FlowFile> &flow_file) {
+  if (expressions_.find(name) == expressions_.end()) {
```
minifirocks: where is the expressions_ map populated?

minifirocks: I did not see the code path for EL for property handling. it will make sense to change README to illustrate the example use case for EL

README.md
```diff
@@ -149,6 +155,10 @@ Aptitude based Linux Distributions
 # ~/Development/code/apache/nifi-minifi-cpp on git:master
 $ apt-get install cmake \
   gcc g++ \
+  bison \
+  flex \
+  libcurl-dev \
```
phrocker: Was this a copy and paste error?

extensions/expression-language/Expression.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+Expression make_dynamic(std::function<std::string(const Parameters &params)> val_fn) {
+  return Expression("", std::move(val_fn));
+}
+
+Expression make_dynamic_attr(const std::string &attribute_id) {
+  return make_dynamic([attribute_id](const Parameters &params) -> std::string {
+    std::string result;
+    params.flow_file.lock()->getAttribute(attribute_id, result);
+    return std::move(result);
```
phrocker: This will not allow for RVO

extensions/expression-language/Expression.cpp
```diff
@@ -0,0 +1,164 @@
...
+  char hostname[1024];
+  hostname[1023] = '\0';
+  gethostname(hostname, 1023);
+  return std::move(std::string(hostname));
+}
+
+std::string expr_toUpper(const std::vector<std::string> &args) {
+  std::string result = args[0];
+  std::transform(result.begin(), result.end(), result.begin(), ::toupper);
+  return std::move(result);
```
phrocker: This will not allow the object to be elided during RVO.

libminifi/CMakeLists.txt
```diff
@@ -74,6 +74,7 @@ target_link_libraries(core-minifi ${UUID_LIBRARIES} ${JSONCPP_LIB} yaml-cpp )
 find_package(ZLIB REQUIRED)
 include_directories(${ZLIB_INCLUDE_DIRS})
 
+target_link_libraries(core-minifi minifi-expression-language-extensions)
```
phrocker: I'm surprised to see this here. This means the extension cannot be fully disabled, so then we rely on the noop?

phrocker: Ah seems that way from the cmake.

libminifi/test/expression-language-tests/ExpressionLanguageTests.cpp
```diff
@@ -0,0 +1,234 @@
...
+ * limitations under the License.
+ */
+
+#include <memory>
+#include <string>
+
+#include "../TestBase.h"
+#include <ExtractText.h>
+
+namespace expression = org::apache::nifi::minifi::expression;
```
phrocker: ^ you can add this to the Core.h if it helps.

README.md
achristianson: Hmm... possibly. Will check it out.

extensions/expression-language/Expression.cpp
achristianson: Good catch. My brain logic was, done with the object, so move it. But now I'm thinking C++ logic. Will fix.

libminifi/CMakeLists.txt
achristianson: Yeah we rely on the noop. Design intent is for noop to be optimized out at compile time, but would have to double-check that's the case. Maybe it should be explicitly inlined so we inline nothing. I'm open to suggestions.

extensions/expression-language/ProcessContextExpr.cpp
achristianson: Good catch. I think it may not be populated. Will fix.

achristianson: ...actually you'll notice a few lines down:

```c++
expressions_.emplace(name, expression::compile(expression_str));
```

So it's def. populating the map.

achristianson:
Addressed review notes.

extensions/expression-language/ProcessContextExpr.cpp
minifirocks: OK.

minifirocks:
looks good to me +1

minifirocks:
@achristianson here is conflict.

achristianson:
@minifirocks fixed.

libminifi/CMakeLists.txt
phrocker: Should be able to use enable_if to to completely eliminate the code and the function calls, but I'm not terribly worried. Cost will be low and can be changed in the future. 

It won't be completely optimized. With -O3 you'll get a stack setup, checks for the std::__throw_bad_function_call() and delete in the assembler. It's not the same as calling an empty function that does nothing, which can be optimized out, but in my opinion it's premature to concern ourselves over the cost.

extensions/expression-language/Expression.cpp
```diff
@@ -0,0 +1,164 @@
...
+}
+
+Expression Expression::operator+(const Expression &other_expr) const {
+  if (isDynamic() && other_expr.isDynamic()) {
+    auto val_fn = val_fn_;
+    auto other_val_fn = other_expr.val_fn_;
+    return make_dynamic([val_fn, other_val_fn](const Parameters &params) -> std::string {
+      std::string result = val_fn(params);
+      result.append(other_val_fn(params));
+      return std::move(result);
```
phrocker: The move prevents copy elision through RVO.

extensions/expression-language/Expression.cpp
```diff
@@ -0,0 +1,164 @@
...
+    });
+  } else if (isDynamic() && !other_expr.isDynamic()) {
+    auto val_fn = val_fn_;
+    auto other_val = other_expr.val_;
+    return make_dynamic([val_fn, other_val](const Parameters &params) -> std::string {
+      std::string result = val_fn(params);
+      result.append(other_val);
+      return std::move(result);
```
phrocker: The move prevents copy elision through RVO.

extensions/expression-language/Expression.cpp
```diff
@@ -0,0 +1,164 @@
...
+    });
+  } else if (!isDynamic() && other_expr.isDynamic()) {
+    auto val = val_;
+    auto other_val_fn = other_expr.val_fn_;
+    return make_dynamic([val, other_val_fn](const Parameters &params) -> std::string {
+      std::string result(val);
+      result.append(other_val_fn(params));
+      return std::move(result);
```
phrocker: The move prevents copy elision through RVO.

phrocker: "
In a return statement or a throw-expression, if the compiler cannot perform copy elision but the conditions for copy elision are met or would be met, except that the source is a function parameter, the compiler will attempt to use the move constructor even if the object is designated by an lvalue; see return statement for details."

extensions/expression-language/Expression.cpp
```diff
@@ -0,0 +1,164 @@
...
+    });
+  } else if (!isDynamic() && !other_expr.isDynamic()) {
+    std::string result(val_);
+    result.append(other_expr.val_);
+    return make_static(std::move(result));
```
phrocker: The move may prevent copy elision here too.

extensions/expression-language/Expression.cpp
```diff
@@ -0,0 +1,164 @@
...
+    params.flow_file.lock()->getAttribute(attribute_id, result);
+    return result;
+  });
+}
+
+std::string expr_hostname(const std::vector<std::string> &args) {
+  char hostname[1024];
+  hostname[1023] = '\0';
+  gethostname(hostname, 1023);
+  return std::move(std::string(hostname));
```
phrocker: ^ same for this move sorry I didn't notice before.

achristianson:
@phrocker fixed.

PR #231, opened 2018-01-08
MINIFICPP-359 Generate connection name field if none provided, thereb…

achristianson:
…y enabling anonymous connections.

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/src/core/yaml/YamlConfiguration.cpp
+++ b/libminifi/src/core/yaml/YamlConfiguration.cpp
@@ -469,10 +469,16 @@ void YamlConfiguration::parseConnectionYaml(YAML::Node *connectionsNode, core::P
 
         // Configure basic connection
         uuid_t uuid;
-        checkRequiredField(&connectionNode, "name",
-                           CONFIG_YAML_CONNECTIONS_KEY);
-        std::string name = connectionNode["name"].as<std::string>();
         std::string id = getOrGenerateId(&connectionNode);
+
+        // Default name to be same as ID
+        std::string name = id;
+
+        // If name is specified in configuration, use the value
+        if (connectionNode["name"]) {
+          name = connectionNode["name"].as<std::string>();
+        }
+
         uuid_parse(id.c_str(), uuid);
         connection = this->createConnection(name, uuid);
         logger_->log_debug("Created connection with UUID %s and name %s", id, name);


libminifi/include/core/yaml/YamlConfiguration.h
```diff
@@ -245,11 +245,11 @@ class YamlConfiguration : public FlowConfiguration {
    *
    * @param yamlNode a pointer to the YAML::Node that will be checked for the
    *                   presence of an idField
-   * @param idField  the string of the name of the idField to check for. This
+   * @param field  the string of the name of the idField to check for. This
    *                   is optional and defaults to 'id'
    * @return         the parsed or generated UUID string
    */
-  std::string getOrGenerateId(YAML::Node *yamlNode, const std::string &idField = "id");
+  std::string getOrGenerateField(YAML::Node *yamlNode, const std::string &field = "id");
```
phrocker: What is the end goal of this PR? It seems that you want to generalize the function name to eventually be used elsewhere.

achristianson: End goal is outlined in https://issues.apache.org/jira/browse/MINIFICPP-359. getOrGenerateField is used in this PR for name now, not just id: `std::string name = getOrGenerateField(&connectionNode, "name");`

phrocker: I see that you added the assignment, but I don't see how we are generating fields other than ID fields with that function. In this case it is okay because name will exist at this point in the code by virtue of the checkRequiredFields call. 

In the case where someone else uses it, there is little to indicate other than const std::string &field = "id" that they will get a uuid field value per the conditional in getOrGenerateField. In my opinion they are getting or generating an ID in the current function, hence my confusion.

achristianson: The checkRequiredField call was removed:

```diff
-        checkRequiredField(&connectionNode, "name",
 -        CONFIG_YAML_CONNECTIONS_KEY);
```

So the current behavior is that name can be left unspecified. In that case it will be generated.

achristianson:
@phrocker removed getOrGenerateField change.

achristianson:
@phrocker fixed conflct should be ready for re-review.

PR #242, opened 2018-01-16
MINIFICPP-374: Commit Linux power management service that enables the…

phrocker:
… threadpools to reduce the number of threads

and throttle active tasks using linux powr constructs.

MINIFICPP-374: Add readme entry

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/README.md
+++ b/README.md
@@ -609,6 +609,21 @@ Additionally, a unique hexadecimal uid.minifi.device.segment should be assigned 
 	  	class: ControllerServiceClass
 	  	Properties:
 
+### Linux Power Manager Controller Service
+  The linux power manager controller service can be configured to monitor the battery level and status ( discharging or charging ) via the following configuration.
+  Simply provide the capacity path and status path along with your threshold for the trigger and low battery alarm and you can monitor your battery and throttle
+  the threadpools within MiNiFi C++. Note that the name is identified must be ThreadPoolManager.
+
+   Controller Services:
+    - name: ThreadPoolManager
+      id: 2438e3c8-015a-1000-79ca-83af40ec1888
+      class: LinuxPowerManagerService
+      Properties:
+          Battery Capacity Path: /path/to/battery/capacity
+          Battery Status Path: /path/to/battery/status
+          Trigger Threshold: 90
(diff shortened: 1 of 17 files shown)


README.md
```diff
@@ -609,6 +609,21 @@ Additionally, a unique hexadecimal uid.minifi.device.segment should be assigned
...
+  Simply provide the capacity path and status path along with your threshold for the trigger and low battery alarm and you can monitor your battery and throttle
+  the threadpools within MiNiFi C++. Note that the name is identified must be ThreadPoolManager.
+
+   Controller Services:
+    - name: ThreadPoolManager
+      id: 2438e3c8-015a-1000-79ca-83af40ec1888
+      class: LinuxPowerManagerService
+      Properties:
+          Battery Capacity Path: /path/to/battery/capacity
+          Battery Status Path: /path/to/battery/status
```
achristianson: Is it feasible to detect these paths automatically if they aren't specified by the user? I.e. use heuristics to check some standard locations, but still allow the user to override if they're set manually.

phrocker: Yeah, for linux I think there is a small(ish) enough set that we can do that. Good idea.

minifirocks: overall looks good. One question I have is why we implement using a controller service instead of a processor. It looks like more like processor to monitor the standard pro filesystem and perform some user config actions

phrocker: Controller services provide a shared service. In this case the evaluation of information to make a determination of whether or not we should reduce threads. It has been discussed that we can move gathering of this information into a Processor, but then we have to transfer it somewhere. In that flow we need to build logic and feed that to elements such as the thread executor, power management, etc. We would need to create a pipeline of logic that flows beyond the processors into structural elements of MiNiFi C++. This is better suited for a controller service. Controller services, by way of the ControllerServiceProvider, are already designed to be part of the thread pools ( scheduling agents ) amongst other structural components. 

This allows a single shared state to make an informed decision. While I can see the case where we feed the controller service by way of the processor output, we'd have to do so at the cost of incurring FlowFile management ( we could have a single processor, but what's the point of that? ).

PR #259, opened 2018-02-05
MINIFICPP-393: Add security support for MQTT

minifirocks:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/mqtt/AbstractMQTTProcessor.cpp
+++ b/extensions/mqtt/AbstractMQTTProcessor.cpp
@@ -41,6 +41,11 @@ core::Property AbstractMQTTProcessor::KeepLiveInterval("Keep Alive Interval", "D
 core::Property AbstractMQTTProcessor::ConnectionTimeOut("Connection Timeout", "Maximum time interval the client will wait for the network connection to the MQTT server", "30 sec");
 core::Property AbstractMQTTProcessor::QOS("Quality of Service", "The Quality of Service(QoS) to send the message with. Accepts three values '0', '1' and '2'", "MQTT_QOS_0");
 core::Property AbstractMQTTProcessor::Topic("Topic", "The topic to publish the message to", "");
+core::Property AbstractMQTTProcessor::SecurityProtocol("Security Protocol", "Protocol used to communicate with brokers", "");
+core::Property AbstractMQTTProcessor::SecurityCA("Security CA", "File or directory path to CA certificate(s) for verifying the broker's key", "");
+core::Property AbstractMQTTProcessor::SecurityCert("Security Cert", "Path to client's public key (PEM) used for authentication", "");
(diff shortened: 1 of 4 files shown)


extensions/mqtt/ConsumeMQTT.cpp
```diff
@@ -35,7 +35,7 @@ namespace nifi {
 namespace minifi {
 namespace processors {
 
-core::Property ConsumeMQTT::MaxQueueSize("Max Flow Segment Size", "Maximum flow content payload segment size for the MQTT record", "");
+core::Property ConsumeMQTT::MaxQueueSize("Max Queue Size", "Maximum receive queue size for the MQTT record", "");
```
phrocker: May not want to change a variable name. If someone is using this option it may break their configuration.

extensions/mqtt/AbstractMQTTProcessor.cpp
```diff
@@ -119,6 +126,38 @@ void AbstractMQTTProcessor::onSchedule(core::ProcessContext *context, core::Proc
     qos_ = valInt;
     logger_->log_debug("AbstractMQTTProcessor: QOS [%ll]", qos_);
   }
+  value = "";
+
+  if (context->getProperty(SecurityProtocol.getName(), value) && !value.empty()) {
+    if (value == MQTT_SECURITY_PROTOCOL_SSL) {
+      sslEnabled_ = true;
```
phrocker: what happens if a cert isn't specified or the CA isn't specified but needs to be?

extensions/mqtt/ConsumeMQTT.cpp
minifirocks: we need to correct that because the variable is not right in the first place. also if user was using that option, it will default to the default value.

extensions/mqtt/AbstractMQTTProcessor.cpp
minifirocks: the SSL handshake will fail and based on the error, user need to config the right certs.

extensions/mqtt/ConsumeMQTT.cpp
phrocker: Do we intend to follow semver?

minifirocks: The previous version of the parameter configuration is wrong so that is no backward compatible.

phrocker: I see your point that the variable name doesn't equal the option name "Max Flow Segment Size" vs "Max Queue Size" , but if a user happens to be using "Max flow segment size" per the option we've changed the semantics of what they want. Does that make sense? We should have that option available to them until we can deprecate it. Further, "receive queue size" seems to not connote if the size is in the number of flow files or the number of bytes. That should be clarified.

minifirocks: Fixed in latest commit

PR #279, opened 2018-03-13
Minificpp 424

phrocker:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/http-curl/client/HTTPClient.cpp
+++ b/extensions/http-curl/client/HTTPClient.cpp
@@ -334,6 +334,8 @@ void HTTPClient::configure_secure_connection(CURL *http_session) {
   logger_->log_debug("Using certificate file %s", ssl_context_service_->getCertificateFile());
   curl_easy_setopt(http_session, CURLOPT_SSL_CTX_FUNCTION, &configure_ssl_context);
   curl_easy_setopt(http_session, CURLOPT_SSL_CTX_DATA, static_cast<void*>(ssl_context_service_.get()));
+  curl_easy_setopt(http_session, CURLOPT_CAINFO, 0);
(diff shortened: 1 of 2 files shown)


extensions/http-curl/client/HTTPClient.cpp
```diff
@@ -334,6 +334,8 @@ void HTTPClient::configure_secure_connection(CURL *http_session) {
   logger_->log_debug("Using certificate file %s", ssl_context_service_->getCertificateFile());
   curl_easy_setopt(http_session, CURLOPT_SSL_CTX_FUNCTION, &configure_ssl_context);
   curl_easy_setopt(http_session, CURLOPT_SSL_CTX_DATA, static_cast<void*>(ssl_context_service_.get()));
+  curl_easy_setopt(http_session, CURLOPT_CAINFO, 0);
```
apiri: I was able to find in the curl docs (https://curl.haxx.se/libcurl/c/CURLOPT_CAINFO.html) for what this behavior does for NSS  and curl 7.55+ and get how this applies but didn't see any such docs for openssl variants or other versions.  Is the behavior pretty consistent from that standpoint?

phrocker: There is little documentation about this but this is to ensure we don't have any ambiguous error messages. Users should be using libcurl-openssl; however, RHEL variants don't have this built into the repos, so we will have to download the source and build it. As a result, the default implementation uses NSS which is not how we've implemented the code.

apiri: Okay, thanks for clarifying.   Looks good to me.  Will merge.

PR #286, opened 2018-03-20
MINIFICPP-438: Preform string to correct test failure for verifying t…

phrocker:
…hat json embedded strings will be unescaped

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/test/expression-language-tests/ExpressionLanguageTests.cpp
+++ b/libminifi/test/expression-language-tests/ExpressionLanguageTests.cpp
@@ -1025,18 +1025,19 @@ TEST_CASE("Encode JSON", "[expressionEncodeJSON]") {  // NOLINT
   REQUIRE("This is a \\\"test!\\\"" == expr({flow_file_a}).asString());
 }
 
+
 TEST_CASE("Decode JSON", "[expressionDecodeJSON]") {  // NOLINT
   auto expr = expression::compile("${message:unescapeJson()}");
 
   auto flow_file_a = std::make_shared<MockFlowFile>();
-  flow_file_a->addAttribute("message", "This is a \\\"test!\\\"");
-  REQUIRE("This is a \"test!\"" == expr({flow_file_a}).asString());
+  flow_file_a->addAttribute("message", "{This is a \\\"test!\\\"}");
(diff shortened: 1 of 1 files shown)


phrocker:
My intent was to not form a valid JSON document but to instead continue the spirit of the test.

achristianson:
I looked into this and it's really a problem with the escape/unescape JSON. We're not aiming to parse/unparse a JSON object, but rather to prepare a string for use in JSON. The escape/unescape EL functions use RapidJSON functions to handle the escaping/unescaping by putting the target value into a string JSON object and stripping the quotes.

Your change is correct in spirit that wrapping the string in an object container makes it compatible with RFC4627 and RFC7159.

The change needs to happen in one of these, from Expression.cpp:

```c++
Value expr_escapeJson(const std::vector<Value> &args) {
  const std::string &arg_0 = args[0].asString();
  rapidjson::StringBuffer buf;
  rapidjson::Writer<rapidjson::StringBuffer> writer(buf);
  writer.String(arg_0.c_str());
  std::string result(buf.GetString());
  return Value(result.substr(1, result.length()-2));
}

Value expr_unescapeJson(const std::vector<Value> &args) {
  std::stringstream arg_0_ss;
  arg_0_ss << "\"" << args[0].asString() << "\"";
  rapidjson::Reader reader;
  rapidjson::StringStream ss(arg_0_ss.str().c_str());
  rapidjson::Document doc;
  doc.ParseStream(ss);
  if (doc.IsString()) {
    return Value(std::string(doc.GetString()));
  } else {
    return Value();
  }
}
```

Either the object wrap/unwrap code should be moved here (instead of the test case), or a different technique altogether should be used for JSON escape/unescape.

Feel free to update this PR/add a new PR or I can handle it if you want.

achristianson:
...also I think it would need to be

```json
["string val"]
```
or
```json
{"dummy_key": "string_val"}
```
to really be universally valid, not just

```json
{"string val"}
```
...because there is no key/val pair in that object.

phrocker:
@achristianson i agreed that something else should likely be done. Feel free to take that on. My goal was to solve the test failure.

achristianson:
Fixed the decode method in #290

PR #296, opened 2018-04-09
MINIFICPP-449 Add cURL external project build with static linking

achristianson:

### For all changes:
     in the commit message?

### For code changes:

### Note:


CMakeLists.txt
```diff
@@ -117,6 +118,49 @@ else ()
   message( FATAL_ERROR "OpenSSL was not found. Please install OpenSSL" )
 endif (OPENSSL_FOUND)
 
+if(NOT USE_SYSTEM_CURL)
```
phrocker: This should be in http-curl. If curl is disabled there is no point in adding it as an external project

main/CMakeLists.txt
```diff
@@ -66,7 +66,7 @@ else ()
 	target_link_libraries (minifiexe -Wl,--whole-archive minifi -Wl,--no-whole-archive)
 endif ()
 
-target_link_libraries(minifiexe yaml-cpp ${UUID_LIBRARIES} ${OPENSSL_LIBRARIES})
+target_link_libraries(minifiexe yaml-cpp ${UUID_LIBRARIES} ${CURL_LIBRARIES} ${OPENSSL_LIBRARIES})
```
phrocker: This should not be included here. It should be passed transitively through the minifi-http-curl iff it is enabled.

extensions/http-curl/CMakeLists.txt
```diff
@@ -42,9 +42,9 @@ if(CMAKE_THREAD_LIBS_INIT)
 endif()
 
 if (CURL_FOUND)
-        include_directories(${CURL_INCLUDE_DIRS})
-        target_link_libraries (minifi-http-curl ${CURL_LIBRARIES})
-endif(CURL_FOUND)
+  include_directories(${CURL_INCLUDE_DIRS})
+  target_link_libraries(minifi-http-curl ${CURL_LIBRARIES})
```
phrocker: ^ Even in the case of a system library we could rely on the static library if it exists. We don't do it now to ensure we have a small binary size, but we should make it clear that while downloading the source and building manually is an option in most cases a static curl lib likely already exists on the build system. In my case I've just checked my centos 7,6, u16-18, fedora, and suse VMs and they all have static libs. There should be a little work here to ensure we get that IF the user wants that. Perhaps that necessitates an added option?

achristianson: Agreed that we should have the option to static link when not using bundled. I think that should be done in a new ticket that covers all dynamic libs that have static linking options when linking system libs, possibly with a general option e.g. STATIC_LINK_ALL.

achristianson:
Moved to curl extension and fixed external static lib cmake dependencies by using PROPERTIES INTERFACE_LINK_LIBRARIES.

PR #307, opened 2018-04-24
MINIFICPP-469 Added encode/decode base64 EL functions

achristianson:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/expression-language/base64.h
+++ b/extensions/expression-language/base64.h
@@ -1,0 +1,188 @@
+/***************************************************************************
+ *                                  _   _ ____  _
+ *  Project                     ___| | | |  _ \| |
+ *                             / __| | | | |_) | |
+ *                            | (__| |_| |  _ <| |___
+ *                             \___|\___/|_| \_\_____|
+ *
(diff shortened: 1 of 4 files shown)


extensions/expression-language/base64.h
```diff
@@ -0,0 +1,188 @@
+/***************************************************************************
```
phrocker: Why is this class here?

achristianson: This is the cURL base64 encode/decode. Here's why:

1) Avoid reinventing the wheel
2) It's a private function in cURL, so that's why it's copied rather than exposing/calling into cURL lib
3) We already have cURL included & in LICENSE, so it's simple to use this one
4) Simpler implementation to call than OpenSSL implementation

achristianson: Rebased to fix conflicts. The reason it's here rather than 3rd-party is that it's specific functions copied/pasted, not a third party lib. We own/adopted the code now and it won't be maintained by a third party.

phrocker: Thanks for rebasing I will merge.

apiri:
reverified builds on OS X and Debian.  will merge.

PR #333, opened 2018-05-21
MINIFICPP-501 Incorporate dependent property metadata into agent info…

achristianson:
…rmation

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/core/state/nodes/AgentInformation.h
+++ b/libminifi/include/core/state/nodes/AgentInformation.h
@@ -297,9 +297,19 @@ class ComponentManifest : public DeviceInformation {
             descriptorRequired.name = "required";
             descriptorRequired.value = prop.second.getRequired();
 
+            SerializedResponseNode descriptorDependentProperties;
+            descriptorDependentProperties.name = "dependentProperties";
+
+            for (const auto &propName : prop.second.getDependentProperties()) {
+              SerializedResponseNode descriptorDependentProperty;
+              descriptorDependentProperty.name = propName;
+              descriptorDependentProperties.children.push_back(descriptorDependentProperty);
+            }
+
             child.children.push_back(descriptorName);
             child.children.push_back(descriptorDescription);
             child.children.push_back(descriptorRequired);
+            child.children.push_back(descriptorDependentProperties);
 
             props.children.push_back(child);
           }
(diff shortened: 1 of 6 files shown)


libminifi/src/processors/PutFile.cpp
```diff
@@ -51,7 +51,9 @@ core::Property PutFile::CreateDirs(
     "Create Missing Directories",
     "If true, then missing destination directories will be created. "
         "If false, flowfiles are penalized and sent to failure.",
-    "true");
+    "true",
+    true,
```
phrocker: Before this property was not required but had a default value of true. Now it's required?

libminifi/include/core/state/nodes/AgentInformation.h
```diff
@@ -297,9 +297,20 @@ class ComponentManifest : public DeviceInformation {
             descriptorRequired.name = "required";
             descriptorRequired.value = prop.second.getRequired();
 
+            SerializedResponseNode descriptorDependentProperties;
+            descriptorDependentProperties.name = "dependentProperties";
+
+            for (const auto &propName : prop.second.getDependentProperties()) {
+              SerializedResponseNode descriptorDependentProperty;
+              descriptorDependentProperty.name = propName;
+              descriptorDependentProperty.value = propName;
```
phrocker: why is the value the property name? You don't need a value if you just need a list.

libminifi/src/processors/PutFile.cpp
achristianson: It's required per documentation https://github.com/apache/nifi-minifi-cpp/blob/master/PROCESSORS.md#putfile. There are other properties out of sync with documentation/spec, which will be reconciled in future commits.

libminifi/include/core/state/nodes/AgentInformation.h
achristianson: Fixed.

achristianson:
Fixed the name/value redundancy.

libminifi/src/processors/PutFile.cpp
achristianson: Also note the case where a prop is required & has a default, and no value is given in the config is covered in unit tests. It will use the default value and validation will pass. So this change has no operational effect unless the user explicitly gives an empty value ("") in the config file.

PR #347, opened 2018-05-29
MINIFIPP-514 Incorporated regex property validation information into agent manifest

achristianson:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/http-curl/processors/InvokeHTTP.cpp
+++ b/extensions/http-curl/processors/InvokeHTTP.cpp
@@ -70,6 +70,7 @@ core::Property InvokeHTTP::SSLContext("SSL Context Service",
                                       "information for TLS/SSL (https) connections.",
                                       "",
                                       false,
+                                      "",
                                       {},
                                       {{"Remote URL", "^http:.*$"}});
 core::Property InvokeHTTP::ProxyHost("Proxy Host", "The fully qualified hostname or IP address of the proxy server", "");
--- a/libminifi/include/core/Property.h
+++ b/libminifi/include/core/Property.h
@@ -53,15 +53,17 @@ class Property {
   /*!
    * Create a new property
    */
-  Property(const std::string name,
-           const std::string description,
+  Property(std::string name,
+           std::string description,
            std::string value,
            bool is_required,
-           std::vector<std::string> &&dependent_properties,
-           std::vector<std::pair<std::string, std::string>> &&exclusive_of_properties)
-      : name_(name),
-        description_(description),
+           std::string valid_regex,
+           std::vector<std::string> dependent_properties,
+           std::vector<std::pair<std::string, std::string>> exclusive_of_properties)
+      : name_(std::move(name)),
+        description_(std::move(description)),
         is_required_(is_required),
+        valid_regex_(std::move(valid_regex)),
         dependent_properties_(std::move(dependent_properties)),
         exclusive_of_properties_(std::move(exclusive_of_properties)),
         is_collection_(false) {
@@ -97,6 +99,7 @@ class Property {
   std::string getDescription() const;
   std::string getValue() const;
   bool getRequired() const;
+  std::string getValidRegex() const;
   std::vector<std::string> getDependentProperties() const;
   std::vector<std::pair<std::string, std::string>> getExclusiveOfProperties() const;
   std::vector<std::string> &getValues();
@@ -375,6 +378,7 @@ class Property {
   std::string name_;
   std::string description_;
   bool is_required_;
+  std::string valid_regex_;
   std::vector<std::string> dependent_properties_;
   std::vector<std::pair<std::string, std::string>> exclusive_of_properties_;
   bool is_collection_;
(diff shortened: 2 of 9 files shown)


extensions/http-curl/processors/InvokeHTTP.cpp
```diff
@@ -65,7 +65,13 @@ core::Property InvokeHTTP::FollowRedirects("Follow Redirects", "Follow HTTP redi
...
                                             " headers in the request. If not defined, no attributes are sent as headers.",
                                             "");
-core::Property InvokeHTTP::SSLContext("SSL Context Service", "The SSL Context Service used to provide client certificate information for TLS/SSL (https) connections.", "");
+core::Property InvokeHTTP::SSLContext("SSL Context Service",
+                                      "The SSL Context Service used to provide client certificate "
+                                      "information for TLS/SSL (https) connections.",
+                                      "",
+                                      false,
+                                      {},
+                                      {{"Remote URL", "^http:.*$"}});
```
phrocker: shouldn't this be https? I also think that URL regexes tend to be much more complicated than .*

phrocker: Actually, is this option part of SSL Context Service? Seems like the wrong variable name for the property.

phrocker: @apiri mentioned his interpretation of this. let's change this and use a builder instead. This is difficult to read.

apiri: would be in favor of that

achristianson: This is confusing because the PR is based on MINIFICPP-500 mutually-exclusive. This PR is for MINIFICPP-514 so that Remote URL mutually-exclusive property is correct.

A URL regex could get complex, but in this case we're just checking for HTTP protocol as per RFC3986 (which is exclusive of this SSL property, for mutual-exclusivity as defined by MINIFICPP-467 E.g. "If I set certificates in InvokeHTTP, I shouldn’t allow http as the target protocol.").

Should it use a controller service? Probably. That's out of scope for this ticket but I agree with a new ticket for that.

All these notes apply for MINIFICPP-500, not MINIFICPP-514.

achristianson: To clarify, the regex is just checking for positive incidence of HTTP protocol, nothing more. It isn't there to validate a URL (again this is very confusing due to the incidental git basing).

phrocker: I'm afraid I'm going to disagree. I'm -1 on this PR until that issue is addressed. Completely happy with it being a separate ticket.

achristianson:
Build fail looks unrelated.

phrocker:
@achristianson I've seen that when there are port collisions in tests. I've made https://issues.apache.org/jira/browse/MINIFICPP-518 and will take a look.

achristianson:
Looks like some conflict(s) cropped up. Taking a look.

achristianson:
Fixed.

apiri:
@achristianson would you mind rebasing this one last time so we can merge it in?

phrocker:
@achristianson and @apiri I don't believe the requests made in these pull requests were ever responded to. It includes this PR and two others that are all the same commits, just layered. I'm okay pushing those requests to another ticket as long as that ticket is a blocker for this release. I don't want to push to a ticket just for the reason that "it's not in the scope of this ticket." and that get forgotten. IMO my request for these tickets was a pretty important request that wasn't responded to.

apiri:
@phrocker I missed those when they got folded up by the GitHub view.  Apologies for that.

achristianson:
@phrocker Please elaborate what exactly your issues are. As you can see from 20 days ago, all concerns were addressed to the best of my knowledge.

extensions/http-curl/processors/InvokeHTTP.cpp
phrocker: I see this comment chain can be a bit confusing -- there were offline discussions that make it a bit worse. My request for a separate ticket was that this and other PRs have made the usage of these classes very difficult to read and use. I believe that it's bad enough that it warrants resolution in this ticket or subsequent work. To make code worse and then fix it doesn't seem correct; however, I'm okay with a subsequent ticket. I'm -1 with my opinion merit especially since there was no discussion of this. I think Aldrin and others can disagree and move forward with their +1s, through which I'll provide an implicit agreement to move forward -- I'm also open to the possibility that I'm completely wrong so let's actually have that discussion. My goal is not to stop this but there was no discussion regarding how this and others have made the code difficult to read and use.

apiri:
It looks like this went in before we had outstanding items tended to, the least of which would be an issue which I do not currently see in JIRA.  We should either get those tasks captured/logged or revert this commit until we can reach a general consensus.

achristianson:
@apiri This was merged on the basis of "I'll provide an implicit agreement to move forward." MINIFICPP-577 has been created to refactor the Property constructor to reduce the # of arguments. I don't see that it was actually clarified in the conversation as to what the issue was, but that is my best guess. Pleas update ticket if incorrect.

PR #348, opened 2018-05-30
MINIFICPP-517: Add RTIMULib and create basic functionality.

phrocker:
Ported from other work, this package contains two processors that create
simple flow files. This approach will likely evolve to improve accessing
the sensor data from the Processors and decorating flow files.

I'm not against using ExternalProject here, but there are some modifications to RTIMULib. We could, alternatively, fork RTIMULib while trying to get changes merged there with ExternalProject or keep a diff in our repo. Since I haven't set myself on either of these approaches I've included the library itself, which I would like to avoid for most dependencies due to the size. Input here is appreciated as I evolve this package. 

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/sensors/GetEnvironmentalSensors.cpp
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  humidity_sensor_ = RTHumidity::createHumidity(&settings);
+  if (humidity_sensor_) {
+    humidity_sensor_->humidityInit();
+  } else {
+    throw std::runtime_error("RTHumidity could not be initialized");
+  }
+
+  pressure_sensor_ = RTPressure::createPressure(&settings);
```
achristianson: Are these sensor objects thread safe? I don't see issue, but wanted to double-check.

phrocker: That's a good point. I will investigate and respond if they are or fix if not.

extensions/sensors/GetEnvironmentalSensors.cpp
```diff
@@ -0,0 +1,156 @@
...
+  if (pressure_sensor_) {
+    pressure_sensor_->pressureInit();
+  } else {
+    throw std::runtime_error("RTPressure could not be initialized");
+  }
+
+}
+
+GetEnvironmentalSensors::~GetEnvironmentalSensors() {
```
achristianson: Possible (but unlikely??) memleak as nothing precludes onSchedule from being called twice. Maybe embed in unique_ptr?

extensions/sensors/GetEnvironmentalSensors.cpp
```diff
@@ -0,0 +1,156 @@
...
+  delete humidity_sensor_;
+  delete pressure_sensor_;
+}
+
+void GetEnvironmentalSensors::onTrigger(const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSession> &session) {
+
+  auto flow_file_ = session->create();
+
+  flow_file_->setSize(0);
```
achristianson: Isn't 0 size the default?

phrocker: It was not when this PR originated. I fixed that I believe in master, so I'll check and remove if that is the case. thanks

extensions/sensors/GetEnvironmentalSensors.cpp
```diff
@@ -0,0 +1,156 @@
...
+    }
+  }
+
+  if (have_sensor) {
+
+    WriteCallback callback("EnvironmentalSensors");
+    session->write(flow_file_, &callback);
+    session->transfer(flow_file_, Success);
+  }
+
```
achristianson: Wondering if there are any exceptions thrown by RT lib or error cases we'd want to log in onTrigger and transfer to fail or at least log->error.

extensions/sensors/GetMovementSensors.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+  setSupportedProperties(properties);
+  // Set the supported relationships
+  std::set<core::Relationship> relationships;
+  relationships.insert(Success);
+  setSupportedRelationships(relationships);
+}
+
+void GetMovementSensors::onSchedule(const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSessionFactory> &sessionFactory) {
+  imu = RTIMU::createIMU(&settings);
```
achristianson: Does this createIMU alloc and give us owner? Didn't catch a delete in the destructor like the other one. If so, same comment applies w/r/t unique_ptr to give more of a RAII memory assurance given that nothing technically precludes a N call to onSchedule where N > 1.

extensions/sensors/GetMovementSensors.cpp
```diff
@@ -0,0 +1,109 @@
...
+      auto vector = imuData.accel;
+      std::string degrees = RTMath::displayDegrees("acceleration", vector);
+      flow_file_->addAttribute("ACCELERATION", degrees);
+    }
+    if (imuData.gyroValid) {
+      auto vector = imuData.gyro;
+      std::string degrees = RTMath::displayDegrees("gyro", vector);
+      flow_file_->addAttribute("GYRO", degrees);
+    }
+
```
achristianson: Same type of comment w/r/t error or exception cases. Didn't notice anything wrong but I imagine there could be failures on sensor polls that we'd want to at least log a warn or error for.

extensions/sensors/SensorLoader.cpp
```diff
@@ -0,0 +1,30 @@
...
+#include "SensorLoader.h"
+
+#include "core/FlowConfiguration.h"
+
+bool SensorFactory::added = core::FlowConfiguration::add_static_func("createSensorFactory");
+
+extern "C" {
+
+void *createSensorFactory(void) {
+  return new SensorFactory();
```
achristianson: Maybe return a moved unique_ptr to make ownership semantics clearer?

phrocker: Can't really do that here due to where this could be called in C code.

extensions/sensors/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+# under the License.
+#
+
+cmake_minimum_required(VERSION 2.6)
+
+
+set(CMAKE_EXE_LINKER_FLAGS "-Wl,--export-all-symbols")
+set(CMAKE_SHARED_LINKER_FLAGS "-Wl,--export-symbols")
+
+include_directories(../../libminifi/include  ../../libminifi/include/core/yaml  ../../libminifi/include/core  ../../thirdparty/spdlog-20170710/include ../../thirdparty/concurrentqueue ../../thirdparty/yaml-cpp-yaml-cpp-0.5.3/include ${CIVET_THIRDPARTY_ROOT}/include ../../thirdparty/)
```
achristianson: Not unique to this PR or ticket, but a lot of this extension boilerplate is really getting duplicated a lot... might be worth considering or adding a ticket for. (Not just this line but quite a bit of various duplicated cruft in an extension CMakeLists.txt).

extensions/sensors/GetMovementSensors.cpp
phrocker: Afraid there is very little from this library that we can translate, aside from a valid value being collected.

extensions/sensors/GetEnvironmentalSensors.cpp
phrocker: Yeah this was one of the places I was hoping to get feedback on. having an invalid read of an IMU is not an error condition, so should I make this trace and log that pressure wasn't valid ? A lot of this comes from the specific IMU and how quickly they can be polled. Don't have many error details since the library hides some of that..but a trace to say "couldn't pull data source?"

extensions/sensors/GetEnvironmentalSensors.cpp
phrocker: My memory served me ( I hadn't touched this in six months ) but these calls are thread safe.

extensions/sensors/CMakeLists.txt
phrocker: yeah I agree. I didn't take a healthy look at the PR before taking an experimental branch because I'm reasonably certain some of that cmake file isn't necessary. It would definitely help others to simplify some of that stuff. I'd also like to automate the acceptance of the extension dir and avoid people having to do more than define their dependencies for their extension -- i.e. not touching the root cmake.

extensions/sensors/GetMovementSensors.cpp
phrocker: If memory serves me this was a singelton, but I updated the dependency version from a very old copy in the original PR, so I'll check and report back or fix. Thanks

extensions/sensors/GetEnvironmentalSensors.cpp
apiri: any value in having a property that still generates a flowfile that might indicate the issue as well.  having the ability to "know" that things aren't able to be collected seems like it be useful but not sure if that is overly contrived.

PR #359, opened 2018-06-15
MINIFICPP-515 Use emplace_back instead of push_back

achristianson:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/core/Property.h
+++ b/libminifi/include/core/Property.h
@@ -55,7 +55,7 @@ class Property {
    */
   Property(const std::string name,
            const std::string description,
-           const std::string value,
+           std::string value,
            bool is_required,
            std::vector<std::string> &&dependent_properties,
            std::vector<std::pair<std::string, std::string>> &&exclusive_of_properties)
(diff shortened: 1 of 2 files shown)


libminifi/include/core/Property.h
```diff
@@ -65,15 +65,15 @@ class Property {
         dependent_properties_(std::move(dependent_properties)),
         exclusive_of_properties_(std::move(exclusive_of_properties)),
         is_collection_(false) {
-    values_.push_back(std::string(value.c_str()));
```
phrocker: I'm curious why this existed the way it did, but my guess is it's overriding the C++11 pointer copy to take ownership of that pointer. We have a named variable already and std::string has a move constructor so most of this is implicitly unnecessary [1], but I'm more concerned with the impetus for the ptr and std::string. That may have been someone ( could be me ) taking advantage of the ptr copy. I'll do al little research into that to validate, but in general push_back of already constructed objects with a move constructor should be fine; however, even in the case above I would have preferred emplacing the pointer than pushing since we're constructing an object to be appended, so I'll take a look at  the string specs before approving to see if I can gather what the original intent was, here. My be a micro-optimization for pointer copies...but I think the move constructor does that for for SSOs, which many of these may be. 

[1] http://ptspts.blogspot.com/2017/02/fast-vector-append.html

achristianson:
@phrocker I read through the blog post.

The high level objective of this PR is to achieve the "copy free" state. The push_back( T&& value ) is what we want.

"in general push_back of already constructed objects with a move constructor should be fine"

It would copy, because the string is still valid after we add it (this would be "Case B" from the article):

 "Case B is not affected (it still copies), and that's good, because we want to print s1 to cout below, so we want that data there."

"Case G" is the fit here, since the strings would still be valid. We need std::move to say that we're done with the string in this scope and it's OK for it to go into arbitrary post-move state.

I'll update to call push_back( T&& value ) which should be copy-free and is appropriate because we're not constructing an object.

achristianson:
Fixed.

phrocker:
@achristianson I feel like there is something I'm missing because there had to have been a reason c_str() existed. On Friday I took another look and couldn't statically determine why that was one. So I'm inclined to merge this since there is no test or comment that alludes to why the newly constructed string was formulated.

PR #360, opened 2018-06-16
MINIFICPP-427 - add PublishKafka headers support

dtrodrigues:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/librdkafka/PublishKafka.cpp
+++ b/extensions/librdkafka/PublishKafka.cpp
@@ -42,6 +42,7 @@ core::Property PublishKafka::MaxMessageSize("Max Request Size", "Maximum Kafka p
 core::Property PublishKafka::RequestTimeOut("Request Timeout", "The ack timeout of the producer request in milliseconds", "");
 core::Property PublishKafka::ClientName("Client Name", "Client Name to use when communicating with Kafka", "");
 core::Property PublishKafka::BatchSize("Batch Size", "Maximum number of messages batched in one MessageSet", "");
+core::Property PublishKafka::AttributeNameRegex("Attributes to Send as Headers", "Any attribute whose name matches the regex will be added to the Kafka messages as a Header", "");
 core::Property PublishKafka::QueueBufferMaxTime("Queue Buffering Max Time", "Delay to wait for messages in the producer queue to accumulate before constructing message batches", "");
 core::Property PublishKafka::QueueBufferMaxSize("Queue Max Buffer Size", "Maximum total message size sum allowed on the producer queue", "");
 core::Property PublishKafka::QueueBufferMaxMessage("Queue Max Message", "Maximum number of messages allowed on the producer queue", "");
@@ -64,6 +65,7 @@ void PublishKafka::initialize() {
   properties.insert(MaxMessageSize);
   properties.insert(RequestTimeOut);
   properties.insert(ClientName);
+  properties.insert(AttributeNameRegex);
   properties.insert(BatchSize);
   properties.insert(QueueBufferMaxTime);
   properties.insert(QueueBufferMaxSize);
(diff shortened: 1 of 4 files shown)


extensions/librdkafka/PublishKafka.h
```diff
@@ -28,6 +28,7 @@
 #include "core/Property.h"
 #include "core/logging/LoggerConfiguration.h"
 #include "rdkafka.h"
+#include <regex>
```
phrocker: This does not work with all variants of GCC that we support. Therefore you will either need to switch to regex.h ( C variant ) or put a note in the ReadMe that not all versions of gcc ( I don't recall, but remember the issues with GetFile ) may support this processor. Will then require some checks in bootstrap to avoid users building this suite of processors for those versions. When I have some time and am back home I can attempt a build across the variants. I am leaning toward 4.8.4 having issues, so I will attempt that first.

extensions/librdkafka/PublishKafka.h
```diff
@@ -115,14 +121,35 @@ class PublishKafka: public core::Processor {
       buffer.reserve(max_seg_size_);
       read_size_ = 0;
       status_ = 0;
+      rd_kafka_resp_err_t err;
+
+      for (auto kv : flowFile_->getAttributes()) {
+        if (regex_match(kv.first, attributeNameRegex_)) {
+          if (!hdrs) {
+            hdrs = rd_kafka_headers_new(8);
```
phrocker: Was eight randomly chosen? If not please add a comment for any maintainers to understand why you chose this number -- even if any number can be added or removed per  their API.

extensions/librdkafka/PublishKafka.cpp
```diff
@@ -262,7 +269,7 @@ void PublishKafka::onTrigger(const std::shared_ptr<core::ProcessContext> &contex
   if (flowFile->getAttribute(KAFKA_KEY_ATTRIBUTE, value))
     kafkaKey = value;
 
-  PublishKafka::ReadCallback callback(flowFile->getSize(), max_seg_size_, kafkaKey, rkt_);
+  PublishKafka::ReadCallback callback(flowFile->getSize(), max_seg_size_, kafkaKey, rkt_, rk_, flowFile, attributeNameRegex);
```
phrocker: Do we need to send the size and the flow file?

dtrodrigues: Now that we're sending the flowFile so we have access to the attributes, we don't need to send the size by itself anymore. I'll update this.

extensions/librdkafka/PublishKafka.h
dtrodrigues: 8 was mostly arbitrary, I used the value from https://github.com/edenhill/librdkafka/blob/master/examples/rdkafka_example.c#L416 since some value needed to be used. 0 is also an option but it looks like if 0 is chosen, any additions will grow the list to 16 elements. If a number besides 0 is chosen, when list runs out of room, the list is grown by 2*the list size.

extensions/librdkafka/PublishKafka.h
dtrodrigues: I believe GCC prior to 4.9 didn't support C++11 regex. I'm leaning towards not supporting older compilers for this processor but can look into using regex.h instead if desired.

phrocker: I'm fine with limiting it to 4.9 forward.

phrocker: @dtrodrigues Did you need help with limiting to 4.9 insofar that we can limit it in the cmake and documentation?

dtrodrigues:
I've pushed a commit which updates the README regarding the GCC version required and refactored the extension code slightly. I'm not sure of the best way to incorporate the GCC version requirement into the bootstrap script -- I'm not seeing how it enforces the appropriate compiler versions for Lua/bustache currently.

phrocker:
@dtrodrigues Cool, I'll take a look at the changes tomorrow. I'll probably just make a follow on ticket for the bootstrap and merge if everything else looks good. Thanks

phrocker:
@dtrodrigues thanks for pointing out the readme update for bustache. I'll add that via MINIFICPP-582 and merge this.

PR #365, opened 2018-06-26
MINIFICPP-543: Implement C2 based update policy enforcement

phrocker:

### For all changes:
     in the commit message?

### For code changes:

### Note:


apiri:
reviewing

libminifi/src/controllers/UpdatePolicyControllerService.cpp
```diff
@@ -0,0 +1,106 @@
...
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace controllers {
+
+core::Property UpdatePolicyControllerService::EnableAllProperties("Enable All Properties", "Enables all properties", "false");
+core::Property UpdatePolicyControllerService::AllowedProperties("Allowed Properties", "Properties for which we will allow updates");
+core::Property UpdatePolicyControllerService::DisallowedProperties("Disallowed Properties", "Properties for which we will not allow updates");
+core::Property UpdatePolicyControllerService::PersistUpdates("Persist Updates", "Property that dictates whether updates should persist a restart");
```
apiri: a -> after

libminifi/src/controllers/UpdatePolicyControllerService.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+void UpdatePolicyControllerService::onEnable() {
+  std::string enableStr, persistStr;
+
+  bool enable_all = false;
+  if (getProperty(EnableAllProperties.getName(), enableStr)) {
+    enable_all = utils::StringUtils::StringToBool(enableStr, enable_all);
+  }
+
+  if (getProperty(PersistUpdates.getName(), persistStr)) {
```
apiri: I saw this is being handled via MINIFICPP-544.  It seems like this should be used in conjunction with the temp/permanent considerations.  Scanned through the C2 wiki page but am a little hazy about how this interacts with temp/permanent.  I assume the policy service would get precedence but am uncertain of conditions such as I request allowing permanent but the instance is not configured to persist config changes.

libminifi/test/unit/UpdatePolicyTests.cpp
```diff
@@ -0,0 +1,78 @@
...
+TEST_CASE("TestEnableAll", "[test1]") {
+  auto controller = std::make_shared<minifi::controllers::UpdatePolicyControllerService>("TestService");
+  std::shared_ptr<minifi::Configure> configuration = std::make_shared<minifi::Configure>();
+  controller->initialize();
+  controller->setProperty(minifi::controllers::UpdatePolicyControllerService::EnableAllProperties, "true");
+  controller->onEnable();
+  REQUIRE(true == controller->canUpdate("anyproperty"));
+}
+
+TEST_CASE("TestEnableAllFalls", "[test1]") {
```
apiri: Falls -> Fails

libminifi/src/controllers/UpdatePolicyControllerService.cpp
```diff
@@ -0,0 +1,106 @@
...
+    persist_updates_ = utils::StringUtils::StringToBool(persistStr, persist_updates_);
+  }
+
+  auto builder = state::UpdatePolicyBuilder::newBuilder(enable_all);
+
+  core::Property all_prop("Allowed Properties", "Properties for which we will allow updates");
+  core::Property dall_prop("Disallowed Properties", "Properties for which we will allow updates");
```
apiri: allow -> disallow

libminifi/src/controllers/UpdatePolicyControllerService.cpp
```diff
@@ -0,0 +1,106 @@
...
+#endif
+#include "core/state/UpdatePolicy.h"
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace controllers {
+
+core::Property UpdatePolicyControllerService::EnableAllProperties("Enable All Properties", "Enables all properties", "false");
```
apiri: Not sure I am tracking the delineation between "enable" and "allow".  It seems that enable all might be better suited as AllowAll?

libminifi/src/controllers/UpdatePolicyControllerService.cpp
phrocker: This is the intransitive verb.

libminifi/src/controllers/UpdatePolicyControllerService.cpp
phrocker: "I assume the policy service would get precedence but am uncertain of conditions such as I request allowing permanent but the instance is not configured to persist config changes."

I'm not sure I follow...but perhaps I should explain this a little better before adding a comment for the code to ensure it makes sense. Updates can be controlled via the flow via the controller service. There may be cases where you do not want configuration updates to be persisted. An easy case would be one in which there is no volatile storage. That's a bit trite as it over exemplifies the argument; however, a less whimsical example may be one in which you only want updates to persist through transferring the minifi.properties file. This ONLY controls updates performed through mechanisms that use the controller service. Does that make sense? If so I'd like to document that on the wiki and in the code as your question points out that it's a point of confusion.

libminifi/src/controllers/UpdatePolicyControllerService.cpp
phrocker: COPY PASTA! Thanks!

libminifi/test/unit/UpdatePolicyTests.cpp
phrocker: Oh wow. Thanks. This is what I get for using a 4k monitor.

libminifi/src/controllers/UpdatePolicyControllerService.cpp
phrocker: I typed a long response that Allow All doesn't fit what I intended with disallow, below; however, it was lost. I think the gist of that response was I will think about a better name to convey its meaning.

phrocker:
@apiri I made some updates to the wiki entry.

apiri:
@phrocker thank you for the updates, definitely clears things up.  will get this merged in

PR #417, opened 2018-10-14
MINIFICPP-638 - C API: add unit tests

arpadboda:

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/src/capi/api.cpp
```diff
@@ -237,25 +239,58 @@ void update_attribute(flow_file_record *ff, char *key, void *value, size_t size)
...
+  if (!attribute_map || attribute_map->empty()) {
+    return nullptr;
+  }
+  attribute_set * attr_set = new attribute_set;
+  attr_set->size = attribute_map->size();
+  attr_set->attributes = new attribute[attr_set->size];
+  std::transform(attribute_map->begin(), attribute_map->end(), attr_set->attributes,
+      [](const string_map::value_type& key_value) -> attribute {
+    attribute ret_val;
+    ret_val.key = key_value.first.data();
```
phrocker: I don't think I would trust the lifetime of this data, especially given how this function could be used.

libminifi/src/capi/api.cpp
```diff
@@ -237,25 +239,58 @@ void update_attribute(flow_file_record *ff, char *key, void *value, size_t size)
...
 
+attribute_set *get_all_attributes(const flow_file_record* ff) {
+  if (ff == nullptr) {
+    return nullptr;
+  }
+  auto attribute_map = static_cast<string_map*>(ff->attributes);
+  if (!attribute_map || attribute_map->empty()) {
+    return nullptr;
+  }
+  attribute_set * attr_set = new attribute_set;
```
phrocker: I'd prefer to avoid allocs for attributes in the API....

libminifi/src/capi/api.cpp
```diff
@@ -32,6 +32,7 @@
 using string_map = std::map<std::string, std::string>;
 
 class API_INITIALIZER {
+ public:
```
phrocker: Why was this needed?

libminifi/include/capi/cstructs.h
```diff
@@ -86,11 +86,16 @@ typedef struct {
  */
 
 typedef struct {
-  char *key;
+  const char *key;
```
phrocker: Should avoid changing structs until 1.0

libminifi/test/capi/CAPITests.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+TEST_CASE("Set valid and invalid properties", "[setProcesssorProperties]") {
+  nifi_instance *instance = create_instance("random_instance", create_port("12345"));
+  REQUIRE(instance != nullptr);
+  flow *test_flow = create_flow(instance, nullptr);
+  REQUIRE(test_flow != nullptr);
+  processor *test_proc = add_processor(test_flow, "GenerateFlowFile");
+  REQUIRE(test_proc != nullptr);
+  REQUIRE(set_property(test_proc, "Data Format", "Text") == 0);  // Valid value
+  // TODO(aboda): add this two below when property handling is made strictly typed
```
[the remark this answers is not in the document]
arpadboda: That's the format linter requires, although this file is going to be excluded anyway, so will remove.

libminifi/src/capi/api.cpp
```diff
@@ -361,6 +396,18 @@ processor *add_processor(flow *flow, const char *processor_name) {
   }
   return nullptr;
 }
+
+processor *add_processor_with_linkage(flow *flow, const char *processor_name) {
+  ExecutionPlan *plan = static_cast<ExecutionPlan*>(flow->plan);
+  auto proc = plan->addProcessor(processor_name, processor_name, core::Relationship("success", "description"), true);
```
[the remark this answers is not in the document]
arpadboda: Makes sense, but why don't we simply require to create a new flow in that case? In my opinion a flow with separate (not connected) processors hardly make any sense.

libminifi/src/capi/api.cpp
arpadboda: Don't see any other way of flattening a map to a C compatible struct.

libminifi/include/capi/cstructs.h
arpadboda: The problem here is without the const specifier you can't write the following:
```
key = "some text"
```
As you end up with a non-const pointer to a string literal. This is pretty error-prone as well as any modification on that will compile, but most probably lead to a bus error or sg. alike.

libminifi/src/capi/api.cpp
arpadboda: Not needed, just left there, will remove, good point.

libminifi/src/capi/api.cpp
```diff
@@ -212,7 +214,7 @@ void free_flowfile(flow_file_record *ff) {
  * @param size size of value
  * @return 0 or -1 based on whether the attributed existed previously (-1) or not (0)
  */
-uint8_t add_attribute(flow_file_record *ff, char *key, void *value, size_t size) {
+uint8_t add_attribute(flow_file_record *ff, const char *key, void *value, size_t size) {
```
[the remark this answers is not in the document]
arpadboda: This is quite the same issue as I've described below (the change in cstructs.h): add_attribute(ff, "my_key", ...) would compile with warning. 

Moreover this function doesn't take ownership (the internal c++ string representation copies the key), so this signature gives better suggestions on ownership to the user.

libminifi/src/capi/api.cpp
```diff
@@ -237,25 +239,58 @@ void update_attribute(flow_file_record *ff, char *key, void *value, size_t size)
...
+    ret_val.key = key_value.first.data();
+    ret_val.value = static_cast<void*>(const_cast<char*>(key_value.second.data()));
+    ret_val.value_size = key_value.second.size();
+    return ret_val;
+  });
+  return attr_set;
+}
+
+void free_attribute_set(attribute_set * attr_set) {
+  delete[] attr_set->attributes;
```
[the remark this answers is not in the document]
arpadboda: Agreed, will add.

libminifi/src/capi/api.cpp
```diff
@@ -237,25 +239,58 @@ void update_attribute(flow_file_record *ff, char *key, void *value, size_t size)
...
+    return nullptr;
+  }
+  auto attribute_map = static_cast<string_map*>(ff->attributes);
+  if (!attribute_map || attribute_map->empty()) {
+    return nullptr;
+  }
+  attribute_set * attr_set = new attribute_set;
+  attr_set->size = attribute_map->size();
+  attr_set->attributes = new attribute[attr_set->size];
+  std::transform(attribute_map->begin(), attribute_map->end(), attr_set->attributes,
```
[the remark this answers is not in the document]
arpadboda: That depends on preference, in my opinion the name of std functions (transform, accumulate, any, all) give a hint on what's happening in the scope, so for me it's more talkative than a simple range based for. 
I'm fine with either way, so can change if you prefer the other.

libminifi/src/capi/api.cpp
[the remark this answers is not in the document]
arpadboda: The reason I implemented this way is saving the cost of allocations, let the user copy anything he would like to keep for further usage.
This approach seemed to match what we do in "create_flowfile_na". 
If you think this is more dangerous than useful, I can change to do deep copy.

libminifi/test/CPPLINT.cfg
```diff
@@ -1,4 +1,5 @@
 set noparent
 filter=-build/include_order,-build/include_alpha
 exclude_files=Server.cpp
-exclude_files=TestBase.cpp
\ No newline at end of file
+exclude_files=TestBase.cpp
+exclude_files=CAPITests.cpp
```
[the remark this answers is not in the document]
arpadboda: Because of void* -> char* conversion we do there to verify the value of the attribute.

libminifi/src/capi/api.cpp
```diff
@@ -237,25 +239,58 @@ void update_attribute(flow_file_record *ff, char *key, void *value, size_t size)
...
+    attribute ret_val;
+    ret_val.key = key_value.first.data();
+    ret_val.value = static_cast<void*>(const_cast<char*>(key_value.second.data()));
+    ret_val.value_size = key_value.second.size();
+    return ret_val;
+  });
+  return attr_set;
+}
+
+void free_attribute_set(attribute_set * attr_set) {
```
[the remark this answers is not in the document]
arpadboda: The first thing to decide here is whether we provide shallow or deep copy of the internal data (such as attributes, values, content, etc). In case anything we do goes with shallow approach (which we do in case of attributes), it becomes our responsibility to properly free things we allocate, we can't let the user calling free. (Needs good documentation we can point to?)

libminifi/src/capi/api.cpp
phrocker: Sorry, meant avoid it by having user perform this. May require different accesses, but then we avoid allocating all memory for them.

arpadboda: Ok, that sounds sensible.

libminifi/include/capi/cstructs.h
phrocker: I see that the function prototype has changed. We should make clear known issue in comments. I'm concerned about the lifetime of these pointers and ownership. The original idea was that key can and would change in attribute due to dynamic properties, hence why key was sent in addition to attribute. i'll have to give this more thought. I think there may be some diverging thoughts on this. It's also making me wonder about the utility of get_all_attributes. 

After more thought we should be making the internal API more C like, so in general use of new on a returned object is probably not ideal anyway ( kind of goes to an earlier comment ). Since we're crossing the boundary there aren't guarantees that we would be safely freeing that -- same goes for the other create_functions where new was already used -- so that can be done later, though ( https://issues.apache.org/jira/browse/MINIFICPP-645).

phrocker:
@arpadboda so as not to slow you down here I've made a number of blockers for this release to address before we can close out -- so if you fix the linter issues I can merge. We can discuss the other issues in time.

PR #420, opened 2018-10-17
MINIFICPP-647: Begin process of deprecating functions and remove thos…

phrocker:
…e which were added recently that aren't needed

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/src/capi/api.cpp
+++ b/libminifi/src/capi/api.cpp
@@ -61,7 +61,7 @@ class DirectoryConfiguration {
  * Creates a NiFi Instance from the url and output port.
  * @param url http URL for NiFi instance
  * @param port Remote output port.
- * @Deprecated for 0.6.0 in favor of
+ * @Deprecated for API version 0.2 in favor of the following prototype
  * nifi_instance *create_instance(nifi_port const *port) {
  */
 nifi_instance *create_instance(const char *url, nifi_port *port) {
--- a/libminifi/src/capi/api.cpp
+++ b/libminifi/src/capi/api.cpp
@@ -57,38 +57,28 @@ class DirectoryConfiguration {
   }
 };
 
-nifi_port *create_port(const char *port) {
-  if (nullptr == port)
-    return nullptr;
-  nifi_port *p = new nifi_port();
-  p->port_id = new char[strlen(port) + 1];
-  memset(p->port_id, 0x00, strlen(port) + 1);
-  strncpy(p->port_id, port, strlen(port));
-  return p;
-}
-
-int free_port(nifi_port *port) {
-  if (port == nullptr)
-    return -1;
-  delete[] port->port_id;
-  delete port;
-  return 0;
-}
-
 /**
  * Creates a NiFi Instance from the url and output port.
  * @param url http URL for NiFi instance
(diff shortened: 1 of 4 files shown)


libminifi/test/capi/CAPITests.cpp
```diff
@@ -33,8 +32,14 @@
 #include <chrono>
 #include <thread>
 
+static nifi_instance *create_instance_obj(const char *name = "random_instance") {
+  nifi_port port;
+  port.port_id = "12345";
+  return create_instance("random_instance", &port);
```
arpadboda: How can you set port after this?

python/minifi/__init__.py
```diff
@@ -66,9 +66,9 @@ def setBase(self, proc):
     def get(self, session):
         ff = self._minifi.get(self._instance.get_instance(),self._flow, session)
         if ff:
```
arpadboda: Python also supports ternary:
```
return FlowFile(self._minifi, ff) if ff else None
```

libminifi/test/capi/CAPITests.cpp
phrocker: Are you asking how can they change the port in the instance?

python/minifi/__init__.py
phrocker: Yeah i tend to avoid ternary operators in general since my experience with production code has always led maintainers at some point to result in error. Looks great -- but more errors than positives in my experience I'm afraid.

libminifi/test/capi/CAPITests.cpp
arpadboda: My bad, just didn't notice this being in the test code.

python/minifi/__init__.py
arpadboda: Fair point.

libminifi/src/capi/api.cpp
```diff
@@ -57,38 +57,28 @@ class DirectoryConfiguration {
...
-  delete[] port->port_id;
-  delete port;
-  return 0;
-}
-
 /**
  * Creates a NiFi Instance from the url and output port.
  * @param url http URL for NiFi instance
  * @param port Remote output port.
+ * @Deprecated for 0.6.0 in favor of
```
apiri: Sorry to nitpick, but shall we refer to minifi releases or API version, could see this getting a little confusing at some point?  How could we best document these version relationships?

phrocker: Oh that's a very good point.

apiri: Well rereading the deprecation note, it makes more sense than my first take on things.  I guess I anticipate this maybe being problematic in the near future when both would reach the 1.0 mark.  I know there is connectedness between the both the capi and the agent itself but would almost prefer just saying that it was done for 0.02.  I could see the API reaching a certain level of maturity where the agent would continue to evolve but the API itself is fairly static.  

As a general thought, we could probably use some exploration of minifi cpp's versioning/support guarantees especially as this API gets built out.

apiri:
Thanks for updating.  Will get this merged in.

PR #421, opened 2018-10-18
Minificpp 641 - C API: add support to register failure callback

arpadboda:

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/include/capi/Plan.h
```diff
@@ -93,11 +95,50 @@ class ExecutionPlan {
...
+  class FailureHandler {
+   public:
+    FailureHandler() {
+      callback_ = nullptr;
+    }
+    void setCallback(void (*onerror_callback)(const flow_file_record*)) {
+      callback_=onerror_callback;
+    }
+    void operator()(const processor_session* ps)
+    {
```
phrocker: hmm, this doesn't give the option of what's called a rollback ( go back and potentially re-try). I'm curious if this could be done more simply with a function ptr and that gives them the option and provide them a set of ready-made ones that rollback and retry, drop, and continue, etc?

libminifi/include/capi/api.h
```diff
@@ -79,6 +79,13 @@ processor *add_processor_with_linkage(flow *flow, const char *processor_name);
 
 processor *add_python_processor(flow *, void (*ontrigger_callback)(processor_session *session));
 
+/**
+* Register your callback to received flow files that the flow failed to process
+* The flow file is deleted after the callback is executed, make sure to copy all the data you need!
+* The first callback should be registered before the flow is used. Can be changed later during runtime.
+*/
```
phrocker: what if failure is an optional condition? Why remove and not reuse? curious about comment above.

libminifi/include/capi/Plan.h
```diff
@@ -93,11 +95,50 @@ class ExecutionPlan {
...
 
+  static std::shared_ptr<core::Processor> createProcessor(const std::string &processor_name, const std::string &name);
+
  protected:
+  class FailureHandler {
+   public:
+    FailureHandler() {
+      callback_ = nullptr;
+    }
+    void setCallback(void (*onerror_callback)(const flow_file_record*)) {
```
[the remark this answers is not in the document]
arpadboda: Done

libminifi/include/capi/Plan.h
```diff
@@ -93,11 +95,50 @@ class ExecutionPlan {
...
+        ffr->ffp = ff.get();
+        callback_(ffr);
+      }
+      // This deletes the content of the flowfile as ff gets out of scope
+      // It's the users responsibility to copy all the data
+      ses->remove(ff);
+
+    }
+   private:
+    void (*callback_)(const flow_file_record*);
```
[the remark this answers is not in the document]
arpadboda: Done

libminifi/src/capi/Plan.cpp
```diff
@@ -202,29 +180,31 @@ std::shared_ptr<core::ProcessSession> ExecutionPlan::getCurrentSession() {
...
-  if (setDest)
-    connection->setDestinationUUID(uuid_copy);
+    std::shared_ptr<processors::CallbackProcessor> callback_proc = std::static_pointer_cast<processors::CallbackProcessor>(failure_proc);
+    callback_proc->setCallback(nullptr, std::bind(&FailureHandler::operator(), failure_handler_, std::placeholders::_1));
 
-  processor->addConnection(connection);
-  return connection;
-}
+    for (const auto& proc : processor_queue_) {
+      relationships_.push_back(connectProcessors(proc, failure_proc, core::Relationship("failure", "failure collector"), true));
```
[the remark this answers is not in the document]
arpadboda: Done

libminifi/include/capi/api.h
arpadboda: Not sure I understand the Q properly. 
The current implementation only requires to add the callback before finalising the flow (as there is no way to add the callback processor later). Later the callback can be replaced.

libminifi/include/capi/Plan.h
arpadboda: Makes sense, but I'm not aware of all possible use-cases here, so let's discuss before doing further implementation.

libminifi/include/capi/api.h
phrocker: Sorry that was a bit terse -- I'm asking about the choice to remove the flow file by default. It seems that there is no choice to handle rolling back events ( to retry ) or keeping the flow file content for future operations without completely copying the data. That seems like it should be something that's optional

libminifi/include/capi/Plan.h
phrocker: These are features that exist within NiFi and are things users will undoubtedly expect. Rollback is a feature that enables us to roll back that flow file. As you state in the comment in capi.h it is the dev's responsibility to copy the content -- but the content could be modified at this point and not the original content -- that's something that should happen at rollback. Therefore even left it at at this state, they are potentially losing any ability to correct or even copy the data before a processor may have failed or modified the content. Happy to move this to a ticket referencing this PR if you want to create one, that way we can implement that discussion before release and avoid keeping this PR alive forever. Also happy to keep discussion here, too.

libminifi/include/capi/api.h
phrocker: ^ again, happy to move this discussion to that same ticket referenced above and merge this. Seems like any of these ideas can be done after more discussion on this topic. Thanks!

PR #429, opened 2018-10-29
MINIFICPP-654 - C API: failure callback improvements

arpadboda:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/capi/Plan.h
+++ b/libminifi/include/capi/Plan.h
@@ -46,6 +46,41 @@
 #include "capi/cstructs.h"
 #include "capi/api.h"
 
+using failure_callback_type = std::function<void(flow_file_record*)>;
+using content_repo_sptr = std::shared_ptr<core::ContentRepository>;
+
+namespace {
+
+  void failureStrategyAsIs(core::ProcessSession *session, failure_callback_type user_callback, content_repo_sptr cr_ptr) {
+    auto ff = session->get();
+    if (ff == nullptr) {
+      return;
+    }
+
+    auto claim = ff->getResourceClaim();
+
+    if (claim != nullptr && user_callback != nullptr) {
+      claim->increaseFlowFileRecordOwnedCount();
+      // create a flow file.
+      auto path = claim->getContentFullPath();
+      auto ffr = create_ff_object_na(path.c_str(), path.length(), ff->getSize());
+      ffr->attributes = ff->getAttributesPtr();
+      ffr->ffp = ff.get();
+      auto content_repo_ptr = static_cast<std::shared_ptr<minifi::core::ContentRepository>*>(ffr->crp);
+      *content_repo_ptr = cr_ptr;
+      user_callback(ffr);
+    }
+    session->remove(ff);
+  }
+
+  void failureStrategyRollback(core::ProcessSession *session, failure_callback_type user_callback, content_repo_sptr cr_ptr) {
+    session->rollback();
+    failureStrategyAsIs(session, user_callback, cr_ptr);
+  }
+}
+
+static const std::map<FailureStrategy, const std::function<void(core::ProcessSession*, failure_callback_type, content_repo_sptr)>> FailureStrategies =
+    { { FailureStrategy::AS_IS, failureStrategyAsIs }, {FailureStrategy::ROLLBACK, failureStrategyRollback } };
 
 class ExecutionPlan {
  public:
@@ -67,7 +102,9 @@ class ExecutionPlan {
 
   bool runNextProcessor(std::function<void(const std::shared_ptr<core::ProcessContext>, const std::shared_ptr<core::ProcessSession>)> verify = nullptr);
 
-  bool setFailureCallback(void (*onerror_callback)(const flow_file_record*));
+  bool setFailureCallback(failure_callback_type onerror_callback);
+
+  bool setFailureStrategy(FailureStrategy start);
 
   std::set<provenance::ProvenanceEventRecord*> getProvenanceRecords();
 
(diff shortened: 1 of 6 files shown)


arpadboda:
It's not final, further changes expected as I'm not yet satisfied with the code quality, but the concept can be reviewed at least.

phrocker:
@arpadboda is this ready for review? You stated, "it's not final"

arpadboda:
Yes, I plan some further changes, but most probably in scope of a different PR.

libminifi/include/capi/Plan.h
```diff
@@ -16,6 +16,8 @@
  * limitations under the License.
  */
 
+#pragma once
```
phrocker: The more I think about this for a public, open source project, we should probably avoid pragma once given that it's implementation defined and not language spec. We don't control all compiler variants, especially for what will be this portion of the API. I realize it's nearly "universally supported" but it seems like an unnecessary risk, albeit small.

libminifi/include/capi/Plan.h
```diff
@@ -46,11 +48,50 @@
 #include "capi/cstructs.h"
 #include "capi/api.h"
 
+enum FailureStrategy { AS_IS, ROLLBACK };
+
+bool intToFailureStragey(int in, FailureStrategy *out);
```
phrocker: *spelling.
Is this something to be used in C? Perhaps we should avoid booleans? stdbool wasn't introduced until C99..

libminifi/include/capi/Plan.h
```diff
@@ -46,11 +48,50 @@
 #include "capi/cstructs.h"
 #include "capi/api.h"
 
+enum FailureStrategy { AS_IS, ROLLBACK };
```
phrocker: Do you think AS_IS could be clarified in the scope of the enumeration to better indicate the intent?

libminifi/include/capi/Plan.h
```diff
@@ -46,11 +48,50 @@
...
+
+bool intToFailureStragey(int in, FailureStrategy *out);
+
+using callback_type = std::function<void(flow_file_record*)>;
```
phrocker: ^I'm not inherently against aliasing but I think 'callback_type' does little to improve readability, and since readability is in the eye of the beholder ( and I don't have a preference if others use it ) is there a way to at least better indicate that this is a function used for a certain type of callback?

libminifi/include/capi/Plan.h
```diff
@@ -46,11 +48,50 @@
...
+    session->remove(ff);
+  }
+
+  void failureStrategyRollback(core::ProcessSession *session, callback_type user_callback, content_repo_sptr cr_ptr) {
+    session->rollback();
+    failureStrategyAsIs(session, user_callback, cr_ptr);
+  }
+}
+
+static const std::map<FailureStrategy, const std::function<void(core::ProcessSession*, callback_type, content_repo_sptr)>> FailureStrategies =
```
phrocker: If there are only two is an ordered map or any map at all necessary? Would a simple switch suffice?

libminifi/src/capi/api.cpp
```diff
@@ -170,31 +165,31 @@ flow_file_record* create_ff_object_na(const char *file, const size_t len, const
   new_ff->attributes = nullptr;
   new_ff->contentLocation = new char[len + 1];
   snprintf(new_ff->contentLocation, len + 1, "%s", file);
-  std::ifstream in(file, std::ifstream::ate | std::ifstream::binary);
   // set the size of the flow file.
   new_ff->size = size;
+  new_ff->crp = static_cast<void*>(new std::shared_ptr<minifi::core::ContentRepository>);
```
phrocker: make_shared ?

libminifi/src/capi/api.cpp
```diff
@@ -400,11 +395,19 @@ processor *add_processor_with_linkage(flow *flow, const char *processor_name) {
   return nullptr;
 }
 
-int add_failure_callback(flow *flow, void (*onerror_callback)(const flow_file_record*)) {
+int add_failure_callback(flow *flow, void (*onerror_callback)(flow_file_record*)) {
   ExecutionPlan *plan = static_cast<ExecutionPlan*>(flow->plan);
   return plan->setFailureCallback(onerror_callback) ? 0 : 1;
 }
 
+int set_failure_strategy(flow *flow, int strategy) {
```
phrocker: Why is strategy not typed to an enumeration? Might change the dynamics of the return value but it might make conversions more safe.

libminifi/include/capi/Plan.h
```diff
@@ -46,11 +48,50 @@
...
+    { { FailureStrategy::AS_IS, failureStrategyAsIs }, {FailureStrategy::ROLLBACK, failureStrategyRollback } };
 
 class ExecutionPlan {
  public:
 
-  explicit ExecutionPlan(std::shared_ptr<core::ContentRepository> content_repo, std::shared_ptr<core::Repository> flow_repo, std::shared_ptr<core::Repository> prov_repo);
+  explicit ExecutionPlan(content_repo_sptr content_repo, std::shared_ptr<core::Repository> flow_repo, std::shared_ptr<core::Repository> prov_repo);
```
phrocker: ^ Seems harder to read. Not sure if it helps but we have our linter set to 200 line characters ( to match Apache NiFi ), so we're not that tight on line space.

phrocker:
Looks good. minor comments, nothing that's a big deal

libminifi/include/capi/Plan.h
arpadboda: Nope, it's hidden behind the C API, the API wraps it to int return value.

libminifi/include/capi/Plan.h
arpadboda: I agree with you that it's not the best naming as doesn't properly describe what happens, but no better idea. @apiri - you are our last hope. :)

libminifi/include/capi/Plan.h
arpadboda: Sure, I will rename this, should be failure callback type or something alike.

libminifi/src/capi/api.cpp
arpadboda: As there are no arguments provided, I couldn't save specifying the template parameter. Moreover it's an empty shared ptr, so it doesn't even allocate twice.

libminifi/src/capi/api.cpp
arpadboda: The reason is pretty simple: I didn't have enough coffee when I wrote this, so forgot that C also supports enums. :) Will replace.

libminifi/include/capi/Plan.h
arpadboda: For extendability and readability reasons I would prefer to keep the map. As it's failure handling, I don't think the minor overhead introduced by the map matters.

libminifi/include/capi/Plan.h
arpadboda: Agreed, it doesn't really help here.

libminifi/include/capi/Plan.h
arpadboda: As it's Plan.h, this will surely be compiled with C++ compilers, which properly support according to my experience. I don't plan to add it to C API.

phrocker: We should keep to the standards.

arpadboda: Removed.

PR #430, opened 2018-10-29
MINIFICPP-659: Break out CAPI into nanofi

phrocker:
Break out structure -- facilitating Python and new nanofi structure

### For all changes:
     in the commit message?

### For code changes:

### Note:


nanofi/examples/CMakeLists.txt
```diff
@@ -61,18 +61,18 @@ add_executable(generate_flow generate_flow.c)
...
 
 add_executable(transmit_flow transmit_flow.c)
 
-target_link_libraries(transmit_flow capi ${CMAKE_THREAD_LIBS_INIT} ${LINK_FLAGS} minifi ${LINK_END_FLAGS})
+target_link_libraries(transmit_flow nanofi ${CMAKE_THREAD_LIBS_INIT})
 
 add_executable(monitor_directory monitor_directory.c)
 
-target_link_libraries(monitor_directory capi ${CMAKE_THREAD_LIBS_INIT} ${LINK_FLAGS} minifi ${LINK_END_FLAGS})
+target_link_libraries(monitor_directory nanofi ${CMAKE_THREAD_LIBS_INIT})
```
arpadboda: Are you sure this is fine?

I see that nanofi is linked with whole-archive flag to make sure it contains all symbols from minifi, however when linking happens here and monitor_directory binary is created, the linker is still free to drop unused symbols. 

Fix me, I would also be happy if we could skip doing that ugly flag-magic here. :)

thirdparty/rocksdb/CMakeLists.txt
```diff
@@ -592,16 +592,16 @@ if(WIN32)
...
+  #add_library(${ROCKSDB_SHARED_LIB} SHARED ${SOURCES})
  # target_link_libraries(${ROCKSDB_SHARED_LIB}
 #    ${THIRDPARTY_LIBS} ${SYSTEM_LIBS})
-  set_target_properties(${ROCKSDB_SHARED_LIB} PROPERTIES
-                        LINKER_LANGUAGE CXX
-                        VERSION ${ROCKSDB_VERSION}
-                        SOVERSION ${ROCKSDB_VERSION_MAJOR}
-                        CXX_STANDARD 11
-                        OUTPUT_NAME "rocksdb")
+#  set_target_properties(${ROCKSDB_SHARED_LIB} PROPERTIES
```
arpadboda: Why commented?

If we comment it, even temporary, let's add some comment that justifies.

nanofi/examples/CMakeLists.txt
phrocker: Cmake implants those target libs in place. Therefore, I get

/usr/bin/c++   -std=c++11 -DOPENSSL_SUPPORT -std=c++11 -std=c++11  -rdynamic CMakeFiles/monitor_directory.dir/monitor_directory.c.o  -o monitor_directory -Wl,-rpath,/home/marc/dev/nifi-minifi-cpp/build/thirdparty/uuid:/home/marc/dev/nifi-minifi-cpp/build/thirdparty/yaml-cpp-yaml-cpp-20171024 ../libnanofi.a -lpthread -Wl,--whole-archive ../../libminifi/libminifi.a ../../libminifi/libcore-minifi.a -ldl ../../thirdparty/uuid/libuuid.so ../../thirdparty/yaml-cpp-yaml-cpp-20171024/libyaml-cpp.so.0.5.3 ../../extensions/expression-language/noop/libminifi-expression-language-extensions.a -lz -lssl -lcrypto -Wl,--no-whole-archive 

Even though I had the above build statement. This is because of the way CMAKE injects the targets, so there isn't the need for the extraneous target protections.  On my system monitror_directory works appropriately after testing, but this does mean that MINIFICPP-660 should probably be changed to enable better testing these examples in docker images. I will take a look at that since I've already started 660

thirdparty/rocksdb/CMakeLists.txt
phrocker: My OSX machine is having some issues so I left this here and remove once OSX passes. Was hoping I could do a quick change but it seems that OSX is failing. It will be removed with a comment telling maintainers why the removal happened.

nanofi/examples/CMakeLists.txt
arpadboda: Thanks, this looks good.

apiri:
reviewing

apiri:
@phrocker would you mind rebasing this? didn't want to make any bad assumptions between this and current master, otherwise looks good and build was fine

phrocker:
@apiri rebasing  now, thanks.

phrocker:
@apiri rebased but don't see the updates in the PR. Do you see them?

apiri:
@phrocker it doesn't seem like I was unable to pull them... unfortunately, may just have to open up a new PR with that commit

PR #432, opened 2018-10-30
MINIFICPP-648 - add processor and add processor with linkage nomencla…

arpadboda:
…ture is confusing

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/CMakeLists.txt
+++ b/libminifi/CMakeLists.txt
@@ -141,11 +141,14 @@ endif()
 SET (LIBMINIFI core-minifi PARENT_SCOPE)
 
 if (ENABLE_PYTHON)
-if (NOT APPLE)
 #### shared
 
 add_library(core-minifi-shared SHARED ${SOURCES})
-target_link_libraries(core-minifi-shared ${CMAKE_DL_LIBS} uuid-shared yaml-cpp)
+if (APPLE)
+	target_link_libraries(core-minifi-shared ${CMAKE_DL_LIBS} yaml-cpp)
+else()
+	target_link_libraries(core-minifi-shared ${CMAKE_DL_LIBS} uuid-shared yaml-cpp)
+endif()
 
 find_package(ZLIB REQUIRED)
 include_directories(${ZLIB_INCLUDE_DIRS})
@@ -175,5 +178,5 @@ endif()
 
 set_property(TARGET core-minifi-shared PROPERTY POSITION_INDEPENDENT_CODE ON)
 set_property(TARGET minifi-shared PROPERTY POSITION_INDEPENDENT_CODE ON)
-endif()
+#endif()
 endif(ENABLE_PYTHON)
--- a/nanofi/include/core/cxxstructs.h
+++ b/nanofi/include/core/cxxstructs.h
@@ -15,22 +15,27 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#include "cxx/CallbackProcessor.h"
-#include "core/cxxstructs.h"
-namespace org {
-namespace apache {
-namespace nifi {
-namespace minifi {
-namespace processors {
 
-void CallbackProcessor::onTrigger(core::ProcessContext *context, core::ProcessSession *session) {
-  if (callback_ != nullptr) {
-    callback_(session);
-  }
-}
+#ifndef NIFI_MINIFI_CPP_CXXSTRUCTS_H
+#define NIFI_MINIFI_CPP_CXXSTRUCTS_H
 
-} /* namespace processors */
-} /* namespace minifi */
-} /* namespace nifi */
-} /* namespace apache */
-} /* namespace org */
+#include "cstructs.h"
+#include "cxx/Plan.h"
+
+struct flow : public ExecutionPlan {
+  using ExecutionPlan::ExecutionPlan;
+};
+
+struct standalone_processor : public core::Processor {
+  using core::Processor::Processor;
+};
+
+struct processor : public core::Processor {
+  using core::Processor::Processor;
+};
+
+struct processor_session : public core::ProcessSession {
+  using core::ProcessSession::ProcessSession;
+};
+
+#endif //NIFI_MINIFI_CPP_CXXSTRUCTS_H
--- a/nanofi/include/cxx/CallbackProcessor.h
+++ b/nanofi/include/cxx/CallbackProcessor.h
@@ -65,7 +65,7 @@ class CallbackProcessor : public core::Processor {
 
  public:
 
-  void setCallback(void *obj,std::function<void(processor_session*)> ontrigger_callback) {
+  void setCallback(void *obj,std::function<void(core::ProcessSession*)> ontrigger_callback) {
     objref_ = obj;
     callback_ = ontrigger_callback;
   }
@@ -82,7 +82,7 @@ class CallbackProcessor : public core::Processor {
 
  protected:
   void *objref_;
-  std::function<void(processor_session*)> callback_;
+  std::function<void(core::ProcessSession*)> callback_;
  private:
   // Logger
   std::shared_ptr<logging::Logger> logger_;
--- a/nanofi/include/cxx/Plan.h
+++ b/nanofi/include/cxx/Plan.h
@@ -44,7 +44,6 @@
 #include "core/ProcessSession.h"
 #include "core/ProcessorNode.h"
 #include "core/reporting/SiteToSiteProvenanceReportingTask.h"
-#include "core/cstructs.h"
 #include "api/nanofi.h"
 
 using failure_callback_type = std::function<void(flow_file_record*)>;
@@ -88,7 +87,7 @@ class ExecutionPlan {
 
   explicit ExecutionPlan(std::shared_ptr<core::ContentRepository> content_repo, std::shared_ptr<core::Repository> flow_repo, std::shared_ptr<core::Repository> prov_repo);
 
-  std::shared_ptr<core::Processor> addCallback(void *, std::function<void(processor_session*)>);
+  std::shared_ptr<core::Processor> addCallback(void *, std::function<void(core::ProcessSession*)>);
 
   std::shared_ptr<core::Processor> addProcessor(const std::shared_ptr<core::Processor> &processor, const std::string &name,
                                                 core::Relationship relationship = core::Relationship("success", "description"),
@@ -144,7 +143,7 @@ class ExecutionPlan {
     return it != proc_plan_map_.end() ? it->second : nullptr;
   }
 
-  static void addProcWithPlan(const std::string& uuid, std::shared_ptr<ExecutionPlan> plan) {
+  static void addProcessorWithPlan(const std::string &uuid, std::shared_ptr<ExecutionPlan> plan) {
     proc_plan_map_[uuid] = plan;
   }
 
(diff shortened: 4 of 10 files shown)


libminifi/src/capi/Plan.cpp
```diff
@@ -23,19 +23,8 @@
 #include <set>
 #include <string>
 
-bool intToFailureStragey(int in, FailureStrategy *out) {
```
arpadboda: This was just forgotten in the previously merged PR, the declaration was already removed from the header.

libminifi/CMakeLists.txt
```diff
@@ -141,11 +141,14 @@ endif()
 SET (LIBMINIFI core-minifi PARENT_SCOPE)
 
 if (ENABLE_PYTHON)
-if (NOT APPLE)
 #### shared
 
 add_library(core-minifi-shared SHARED ${SOURCES})
-target_link_libraries(core-minifi-shared ${CMAKE_DL_LIBS} uuid-shared yaml-cpp)
+if (APPLE)
```
phrocker: I think uuid-shared is already linked elsewhere, so I guess it was just unnecessary?

arpadboda: Indeed, it ended up with double definition errors. However works fine in Linux.

phrocker: Hmm. Maybe we should add -DENABLE_PYTHON to our travis builds

arpadboda: I strongly agree, but in that case Python related tests (the ones that don't transport, just create instance, processor, flow files) could also be added to cover that.

phrocker: Indeed -- https://issues.apache.org/jira/browse/MINIFICPP-660 already exists. I've pre-empted that in favor of CoAP, but should get back to that soon. Python is kind of WIP. I do have some tests from that branch that cover what you are talking about. I'll try and re-surface them as soon as I get away from CoAP.

phrocker:
@arpadboda is this good? I'm good with this otherwise.

arpadboda:
> @arpadboda is this good? I'm good with this otherwise.

Good, but not complete, I will amend some stuff soon. (today)

nanofi/src/cxx/Plan.cpp
```diff
@@ -162,6 +153,21 @@ bool ExecutionPlan::runNextProcessor(std::function<void(const std::shared_ptr<co
   }
   std::shared_ptr<core::ProcessSession> current_session = std::make_shared<core::ProcessSession>(context);
   process_sessions_.push_back(current_session);
+  if (input_ff != nullptr) {
+    auto content_repo = static_cast<std::shared_ptr<minifi::core::ContentRepository>*>(input_ff->crp);
+    std::shared_ptr<minifi::ResourceClaim> claim = std::make_shared<minifi::ResourceClaim>(input_ff->contentLocation, *content_repo);
+    auto stream = (*content_repo)->read(claim);
```
arpadboda: This copies the content of the incoming flow file. 
Naturally this leaves space for improvement, in case the content repos of the current flow and the incoming flow file are the same, the copy is needless. 
Given the size of the change I would prefer to do it later.

arpadboda:
> 
> 
> @arpadboda is this good? I'm good with this otherwise.

Now it is.

nanofi/include/core/cxxstructs.h
```diff
@@ -0,0 +1,41 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef NIFI_MINIFI_CPP_CXXSTRUCTS_H
+#define NIFI_MINIFI_CPP_CXXSTRUCTS_H
+
+#include "cstructs.h"
+#include "cxx/Plan.h"
+
```
phrocker: A lot of this works seems counter to tickets to move away from C++. These structs were created extending C++ classes. and the use of cxx structs seems unnecessary as a result.

nanofi/include/cxx/CallbackProcessor.h
```diff
@@ -65,7 +65,7 @@ class CallbackProcessor : public core::Processor {
 
  public:
 
-  void setCallback(void *obj,std::function<void(processor_session*)> ontrigger_callback) {
+  void setCallback(void *obj,std::function<void(core::ProcessSession*)> ontrigger_callback) {
```
phrocker: It's been a while since I've looked at that work, but did the python bindings call this correctly when running python processors?

nanofi/include/cxx/Plan.h
```diff
@@ -44,7 +44,7 @@
 #include "core/ProcessSession.h"
 #include "core/ProcessorNode.h"
 #include "core/reporting/SiteToSiteProvenanceReportingTask.h"
-#include "core/cstructs.h"
```
phrocker: Should remove the rem'd line.

nanofi/include/cxx/Plan.h
```diff
@@ -133,8 +133,29 @@ class ExecutionPlan {
...
+  }
+
   static std::shared_ptr<core::Processor> createProcessor(const std::string &processor_name, const std::string &name);
 
+  static std::shared_ptr<ExecutionPlan> getPlan(const std::string& uuid) {
+    auto it = proc_plan_map_.find(uuid);
+    return it != proc_plan_map_.end() ? it->second : nullptr;
+  }
+
+  static void addProcWithPlan(const std::string& uuid, std::shared_ptr<ExecutionPlan> plan) {
```
phrocker: addProcessorWithPlan

nanofi/src/api/nanofi.cpp
```diff
@@ -89,6 +93,39 @@ nifi_instance *create_instance(const char *url, nifi_port *port) {
...
 }
 
+standalone_processor *create_processor(const char *name) {
+  static int proc_counter = 0;
+  auto ptr = ExecutionPlan::createProcessor(name, name);
+  if (!ptr) {
+    return nullptr;
+  }
+  if (standalone_instance == nullptr) {
+    nifi_port port;
```
phrocker: this is an input port identifier.

nanofi/src/api/nanofi.cpp
```diff
@@ -323,55 +360,45 @@ int transmit_flowfile(flow_file_record *ff, nifi_instance *instance) {
...
 flow * create_new_flow(nifi_instance * instance) {
   auto minifi_instance_ref = static_cast<minifi::Instance*>(instance->instance_ptr);
-  flow *new_flow = (flow*) malloc(sizeof(flow));
-
-  auto execution_plan = new ExecutionPlan(minifi_instance_ref->getContentRepository(), minifi_instance_ref->getNoOpRepository(), minifi_instance_ref->getNoOpRepository());
-
-  new_flow->plan = execution_plan;
-
-  return new_flow;
+  return new flow(minifi_instance_ref->getContentRepository(), minifi_instance_ref->getNoOpRepository(), minifi_instance_ref->getNoOpRepository());
```
phrocker: https://issues.apache.org/jira/browse/MINIFICPP-645 exists because public facing objects that are turned from a C API can be malloc'd by users. If a consumer of our API doesn't use free as designed, we should guarantee that they only cause memory leaks, not introduce undefined behavior.

nanofi/src/api/nanofi.cpp
```diff
@@ -323,55 +360,45 @@ int transmit_flowfile(flow_file_record *ff, nifi_instance *instance) {
...
   if (nullptr == instance || nullptr == instance->instance_ptr) {
     return nullptr;
   }
   auto minifi_instance_ref = static_cast<minifi::Instance*>(instance->instance_ptr);
-  flow *new_flow = (flow*) malloc(sizeof(flow));
-
-  auto execution_plan = new ExecutionPlan(minifi_instance_ref->getContentRepository(), minifi_instance_ref->getNoOpRepository(), minifi_instance_ref->getNoOpRepository());
 
-  new_flow->plan = execution_plan;
+  flow *new_flow = new flow(minifi_instance_ref->getContentRepository(), minifi_instance_ref->getNoOpRepository(), minifi_instance_ref->getNoOpRepository());
```
phrocker: same malloc / new

nanofi/src/api/nanofi.cpp
```diff
@@ -323,55 +360,45 @@ int transmit_flowfile(flow_file_record *ff, nifi_instance *instance) {
...
 
   if (first_processor != nullptr && strlen(first_processor) > 0) {
     // automatically adds it with success
-    execution_plan->addProcessor(first_processor, first_processor);
```
phrocker: Combined C and C++ implies we're moving more toward C++. If that were the case this API would be drastically different. We should move in the other direction since C++ functions as a temporary interconnect.

nanofi/src/cxx/Plan.cpp
```diff
@@ -162,6 +153,21 @@ bool ExecutionPlan::runNextProcessor(std::function<void(const std::shared_ptr<co
...
+    std::shared_ptr<minifi::FlowFileRecord> flowFile = std::static_pointer_cast<minifi::FlowFileRecord>(current_session->create());
+    auto map = static_cast<std::map<std::string, std::string>*>(input_ff->attributes);
+    if (map) {
+      for (const auto& attr: *map) {
+        flowFile->setAttribute(attr.first, attr.second);
+      }
+    }
+    current_session->importFrom(*stream, flowFile);
+    current_session->transfer(flowFile, core::Relationship("success", "success"));
+    relationships_[relationships_.size()-1]->put(std::static_pointer_cast<core::FlowFile>(flowFile));
```
phrocker: is there a safeguard to ensure a relationship exists?

nanofi/include/cxx/CallbackProcessor.h
arpadboda: Yes, tested and works.

nanofi/src/api/nanofi.cpp
arpadboda: Created with dummy value. Do you think having a valid input port is valid usecase for an instance we don't even expose? (As we use it to "store" standalone processors)

nanofi/include/cxx/Plan.h
arpadboda: Thanks, done.

nanofi/src/api/nanofi.cpp
arpadboda: I guess that only declaring structures via the API leaves us the freedom to change the internal representation anyhow without breaking it. On the other side it's now faster than it was.

nanofi/src/cxx/Plan.cpp
arpadboda: The finalise() call introduces a new relationship anyway, that's the only one we expect to have in case of standalone processors.

nanofi/include/cxx/Plan.h
arpadboda: Renamed

nanofi/include/core/cxxstructs.h
phrocker: What do you mean by "I don't think is moving that way, just simplifies the current structure." ? You don't see it moving away from C++?

nanofi/src/api/nanofi.cpp
phrocker: What are you basing the statement, "it's now faster than it was?" What profiler did you use to gather this, and what benefit does that have over readability concerns?

PR #436, opened 2018-11-12
MINIFICPP-667: Add structural definitions to work out plan for C migr…

phrocker:
…ation

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/nanofi/include/core/cstructs.h
+++ b/nanofi/include/core/cstructs.h
@@ -19,6 +19,8 @@
 #ifndef LIBMINIFI_SRC_CAPI_CSTRUCTS_H_
 #define LIBMINIFI_SRC_CAPI_CSTRUCTS_H_
 
+#include "uthash.h"
+#include "utlist.h"
 /**
  * NiFi Port struct
  */
@@ -30,7 +32,7 @@ typedef struct {
  * Nifi instance struct
  */
 typedef struct {
-
+  // should go away as we transition from C++ Instances to C implementations
   void *instance_ptr;
 
   nifi_port port;
@@ -63,10 +65,12 @@ typedef struct {
  */
 
 typedef struct {
+  // should go away as we transition from C++ Instances to C implementations
   void *processor_ptr;
 } processor;
 
 typedef struct {
+  // should go away as we transition from C++ Instances to C implementations
   void *session;
 } processor_session;
 
@@ -75,6 +79,12 @@ typedef struct {
  *  FLOWFILE OPERATIONS
  * ##################################################################
  */
+
+typedef struct{
+  char *key;
+  char *value;
+  UT_hash_handle handle;
+} key_value;
 
 typedef struct {
   const char *key;
@@ -92,6 +102,17 @@ typedef struct {
  *
  */
 typedef struct {
+  // structural definitions for moving way from C++
+  unsigned modified:1;
+  unsigned flushed:1;
+  unsigned inmem:1;
+  uint64_t id;
+  uint64_t entry_date;
+  uint64_t lineage_start;
+  uint64_t penalty_ms;
+  key_value *attribute_map;
+  // structural definitions for moving way from C++
+
   uint64_t size; /**< Size in bytes of the data corresponding to this flow file */
 
   void * in;
@@ -102,7 +123,7 @@ typedef struct {
 
   void *attributes; /**< Hash map of attributes */
 
-  void *ffp;
+  void *ffp;// should go away as we transition from C++ Instances to C implementations
 
 } flow_file_record;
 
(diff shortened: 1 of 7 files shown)


nanofi/include/core/utlist.h
```diff
@@ -0,0 +1,1073 @@
+/*
```
arpadboda: I would prefer to place this file somewhere else, as it's not really part of core, but a 3rd party library we use. Could you add to a folder that implies this?

nanofi/include/core/cstructs.h
```diff
@@ -76,6 +80,12 @@ typedef struct {
  * ##################################################################
  */
 
+typedef struct{
+  char *key;
```
arpadboda: Can we have this as const char *?
A bit better user experience if you can write something like 
```
key_value.key = "some string" 
```

phrocker: The example code I took this from had this as char*, not sure we can do this. I'd have to look at the library.

nanofi/src/api/nanofi.cpp
```diff
@@ -72,19 +41,17 @@ class DirectoryConfiguration {
...
   /**
    * This API will gradually move away from C++, hence malloc is used for nifi_instance
    * Since minifi::Instance is currently being used, then we need to use new in that case.
    */
-  instance->instance_ptr = new minifi::Instance(url, port->port_id);
+  instance->instance_ptr = create_cxx_instance(url,port);
   // may have to translate port ID here in the future
   // need reinterpret cast until we move to C for this module.
-  instance->port.port_id = reinterpret_cast<char*>(malloc(strlen(port->port_id) + 1));
+  instance->port.port_id = (char*)(malloc(strlen(port->port_id) + 1));
```
arpadboda: I expect this to fail linter, do we already have nanofi out of range of that?

phrocker: Yes. We won't be able to use the C++ linter on C files. I haven't pushed the recent commit, but I'm moving toward renaming this to nanofi.c

nanofi/include/core/cstructs.h
phrocker: I'll add that I'm not against this, but the code pointer's management was a little different. Again, I'll take a look at if this makes sense. core will be things that don't reflect a user experience though. That is code that's not part of our public API.

nanofi/include/core/utlist.h
phrocker: Ah thanks for calling this out. I have a more recent commit with a lot of changes, but didn't move this yet and probably would have forgotten.

phrocker: We have a third party dir in the main directory, which is where this should go.

arpadboda: Not yet in nanofi :) 
Guess this library is not going to be used in the C++ part.

phrocker: sorry CMAKE_SOURCE_DIR. It can go there and should. I think I updated the NOTICE and we direct people to the root source thirdparty, so we can probably just use that. No reason to create levels of thirdparty source dirs yet.

arpadboda: Okay, fine for me. 👍 Happy to approve as it gets moved.

phrocker: As per the title this is to work out a plan so that it's out in the open on Apache. I won't be merging this.

nanofi/src/cxx/nanofi_cpp_layer.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+#include "cxx/nanofi_cpp_layer.h"
+
+class API_INITIALIZER {
+ public:
+  static int initialized;
+};
+
+int API_INITIALIZER::initialized = initialize_api();
+
```
phrocker: @arpadboda I'm going to merge this PR since some of this has made it's way elsewhere and this was just a proof of concept; however, it may make sense to begin isolating the cpp layer a little more like this for the eventuality that we remote it entirely.

phrocker:
Was simply proof of concept to work out plan and visually see ideas in a diff. never intended to merge, so closing as it's OBE.

nanofi/include/core/cstructs.h
```diff
@@ -92,6 +102,17 @@ typedef struct {
  *
  */
 typedef struct {
+  // structural definitions for moving way from C++
+  unsigned modified:1;
```
phrocker: @arpadboda changes in cstructs were the only thing of merit here, where we have attributes...are you incorporating something like this into your site to site work?

nanofi/src/cxx/nanofi_cpp_layer.cpp
arpadboda: You mean NOT going to merge?

Totally agree, my current site2site work introduces UT collections anyway to help working in C.

nanofi/include/core/cstructs.h
arpadboda: No, didn't touch C structures yet, I think I won't need to do either as there is C API to get attributes.

PR #440, opened 2018-11-15
MINIFICPP-676 - Cleanup and fix serializable interface implementation

arpadboda:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/http-curl/client/HTTPCallback.h
+++ b/extensions/http-curl/client/HTTPCallback.h
@@ -78,7 +78,7 @@ class HttpStreamingCallback : public ByteInputCallBack {
 
   }
 
-  virtual int64_t process(uint8_t *vector, size_t size) {
+  virtual int64_t process(const uint8_t * const vector, size_t size) {
 
     std::vector<char> vec;
 
--- a/extensions/http-curl/client/HTTPStream.cpp
+++ b/extensions/http-curl/client/HTTPStream.cpp
@@ -57,12 +57,12 @@ int HttpStream::writeData(std::vector<uint8_t> &buf, int buflen) {
   if ((int)buf.capacity() < buflen) {
     return -1;
   }
-  return writeData(reinterpret_cast<uint8_t *>(&buf[0]), buflen);
+  return writeData((&buf[0]), buflen);
 }
 
 // data stream overrides
 
-int HttpStream::writeData(uint8_t *value, int size) {
+int HttpStream::writeData(const uint8_t * const value, int size) {
   if (!IsNullOrEmpty(value)) {
     if (!started_) {
       std::lock_guard<std::mutex> lock(mutex_);
--- a/extensions/http-curl/client/HTTPStream.h
+++ b/extensions/http-curl/client/HTTPStream.h
@@ -107,7 +107,7 @@ class HttpStream : public io::BaseStream {
    * @param value value to write
    * @param size size of value
    */
-  virtual int writeData(uint8_t *value, int size) override;
+  virtual int writeData(const uint8_t * const value, int size) override;
 
   /**
    * Returns the underlying buffer
--- a/libminifi/include/io/CRCStream.h
+++ b/libminifi/include/io/CRCStream.h
@@ -84,7 +84,7 @@ class CRCStream : public BaseStream {
    * @param value value to write
    * @param size size of value
    */
-  virtual int writeData(uint8_t *value, int size);
+  virtual int writeData(const uint8_t * const value, int size);
 
   /**
    * write 4 bytes to stream
@@ -195,14 +195,13 @@ int CRCStream<T>::readData(uint8_t *buf, int buflen) {
 
 template<typename T>
 int CRCStream<T>::writeData(std::vector<uint8_t> &buf, int buflen) {
-
   if ((int)buf.capacity() < buflen)
     buf.resize(buflen);
-  return writeData((uint8_t*) &buf[0], buflen);
+  return writeData(&buf[0], buflen);
 }
 
 template<typename T>
-int CRCStream<T>::writeData(uint8_t *value, int size) {
+int CRCStream<T>::writeData(const uint8_t * const value, int size) {
 
   int ret = child_stream_->write(value, size);
   crc_ = crc32(crc_, value, size);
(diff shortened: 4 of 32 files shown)


arpadboda:
A step forward, there is improvement left (a lot of code duplication exists in stream hierarchy), although this is enough for one PR.

extensions/http-curl/client/HTTPStream.cpp
```diff
@@ -53,16 +53,16 @@ void HttpStream::seek(uint64_t offset) {
   throw std::exception();
 }
 
-int HttpStream::writeData(std::vector<uint8_t> &buf, int buflen) {
+int HttpStream::writeData(const std::vector<uint8_t> &buf, int buflen) {
   if ((int)buf.capacity() < buflen) {
     return -1;
   }
-  return writeData(reinterpret_cast<uint8_t *>(&buf[0]), buflen);
```
phrocker: this fails on some compilers. I spent a long time and saw failures on older versions ( not sure if specifically this one ). Did this cause a specific bug? These are non const for a reason and that's to tell the caller that their data may change. You're making an assumption that it shouldn't be const.

extensions/http-curl/client/HTTPStream.h
```diff
@@ -100,14 +100,14 @@ class HttpStream : public io::BaseStream {
    * @param buflen buffer to write
    *
    */
-  virtual int writeData(std::vector<uint8_t> &buf, int buflen);
+  virtual int writeData(const std::vector<uint8_t> &buf, int buflen);
```
phrocker: Same. C and C++ developers see const and they know what that means. They see the lack of const and know their data may change. If it's non const that doesn't mean it should be changed without discussion. I think these should be non const as we reserve the right to take ownership in the internal API. I would agree for external API.

libminifi/include/io/CRCStream.h
```diff
@@ -194,15 +194,12 @@ int CRCStream<T>::readData(uint8_t *buf, int buflen) {
 }
 
 template<typename T>
-int CRCStream<T>::writeData(std::vector<uint8_t> &buf, int buflen) {
-
-  if ((int)buf.capacity() < buflen)
```
phrocker: This existed for a reason. Did you test this in flight ? This causes a failure. We have non const because we reserve the right to adjust buffers. That's the intent behind making the signatures the way they are.

libminifi/include/io/Serializable.h
```diff
@@ -22,11 +22,36 @@
 #include <string>
 #include "EndianCheck.h"
 #include "DataStream.h"
+
+namespace {
+  template<typename Integral, typename std::enable_if<
```
phrocker: Far less readable and godbolt and my i7 showed no benefit to this. It seems like syntactical candy. Profiling shows no benefit whatsoever. Having functions with specific functors makes it easy on the development side to specifically type what we set. Your ticket "Type-unsafe template functions" I disagree. I think this is more type unsafe and less easy for developers to read. 

"Code duplication" I think this is true, but it results in perhaps some easier development for some. I think we should keep this as is and what was added is syntactical sugar with the potential negative that it can be more confusing during development.

libminifi/include/io/Serializable.h
```diff
@@ -173,8 +147,22 @@ class Serializable {
...
+  * @param value reference in which will set the result
+  * @param stream stream from which we will read
+  * @return resulting read size
+  **/
+  template<typename Integral, typename std::enable_if<
+      (sizeof(Integral) > 1) &&
+      std::is_integral<Integral>::value &&
+      !std::is_signed<Integral>::value
+      ,Integral>::type* = nullptr>
+  int read(Integral &value, DataStream *stream, bool is_little_endian = EndiannessCheck::IS_LITTLE) {
```
phrocker: My IDEs does not make this clear the function to use. In some ways java made this easier with their streams "writeInt, writeShort, writeUnsignedShort" and we can do the same with write(short) write(int), write(long), but my one IDE correctly provides this while others don't.  I agree that generally reducing "Needless functions" is a good thing, but here it served a purpose. Happy to further discuss how we can benefit that purpose.

libminifi/include/io/tls/TLSSocket.h
```diff
@@ -173,19 +173,19 @@ class TLSSocket : public Socket {
    * @param buflen buffer to write
    *
    */
-  int writeData(std::vector<uint8_t> &buf, int buflen);
+  int writeData(const std::vector<uint8_t> &buf, int buflen);
```
phrocker: if we chose ( and we eventually will in further tickets ) to take this memory and do a memmove ( in a different stream -- can discuss later ) we can't do that with the changes. This was intentional and not an oversight.

libminifi/src/io/FileStream.cpp
```diff
@@ -82,16 +82,16 @@ void FileStream::seek(uint64_t offset) {
   file_stream_->seekp(offset_);
 }
 
-int FileStream::writeData(std::vector<uint8_t> &buf, int buflen) {
+int FileStream::writeData(const std::vector<uint8_t> &buf, int buflen) {
   if (static_cast<int>(buf.capacity()) < buflen) {
     return -1;
   }
-  return writeData(reinterpret_cast<uint8_t *>(&buf[0]), buflen);
```
phrocker: fails to compile on some of my archs. perhaps run make docker ?

extensions/http-curl/client/HTTPCallback.h
```diff
@@ -78,7 +78,7 @@ class HttpStreamingCallback : public ByteInputCallBack {
 
   }
 
-  virtual int64_t process(uint8_t *vector, size_t size) {
+  virtual int64_t process(const uint8_t *vector, size_t size) {
```
phrocker: in general when i see something like const uint8_t *ptr, then I wonder what the true intent may be there. You have a pointer that can be modified. Should also make that pointer const too, but as with all comments I'm not sure that's the best approach here.

phrocker:
hmmm, I should have checked appveyor because I would have saved some time building on windows...but as I suspected we have issues there too.

libminifi/include/io/Serializable.h
arpadboda: You are right in the part that this doesn't improve either performance or readibility.
This change became useful in the template write(Integral) function, where this can be called easily, so I don't need to copy-paste writeData calls and could remove the local template writeData implementation. 
Overall I see more gain here.

PR #448, opened 2018-11-26
MINIFICPP-682 - C API: provide functions to create custom processors

arpadboda:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/nanofi/include/api/nanofi.h
+++ b/nanofi/include/api/nanofi.h
@@ -37,6 +37,9 @@ extern "C" {
  */
 #define API_VERSION "0.02"
 
+#define SUCCESS_RELATIONSHIP "success"
+#define FAILURE_RELATIONSHIP "failure"
+
 void enable_logging();
 
 void set_terminate_callback(void (*terminate_callback)());
@@ -97,7 +100,7 @@ int add_failure_callback(flow *flow, void (*onerror_callback)(flow_file_record*)
 * Set failure strategy. Please use the enum defined in cstructs.h
 * Return values: 0 (success), -1 (strategy cannot be set - no failure callback added?)
 * Can be changed runtime.
-* The defailt strategy is AS IS.
+* The default strategy is AS IS.
 */
 int set_failure_strategy(flow *flow, FailureStrategy strategy);
 
@@ -106,6 +109,8 @@ int set_property(processor *, const char *, const char *);
 int set_standalone_property(standalone_processor*, const char*, const char *);
 
 int set_instance_property(nifi_instance *instance, const char*, const char *);
+
+char * get_property(const processor_context *  context, const char * name);
 
 int free_flow(flow *);
 
@@ -135,13 +140,18 @@ flow_file_record* create_ff_object(const char *file, const size_t len, const uin
 
 flow_file_record* create_ff_object_na(const char *file, const size_t len, const uint64_t size);
 
+/**
+ * Get incoming flow file. To be used in processor logic callbacks.
+ */
+flow_file_record* get_flowfile(processor_session* session, processor_context* context);
+
 void free_flowfile(flow_file_record*);
 
 uint8_t add_attribute(flow_file_record*, const char *key, void *value, size_t size);
 
 void update_attribute(flow_file_record*, const char *key, void *value, size_t size);
 
-uint8_t get_attribute(flow_file_record *ff, attribute *caller_attribute);
+uint8_t get_attribute(const flow_file_record *ff, attribute *caller_attribute);
 
 int get_attribute_qty(const flow_file_record* ff);
 
@@ -164,6 +174,12 @@ uint8_t remove_attribute(flow_file_record*, char *key);
  */
 
 int transmit_flowfile(flow_file_record *, nifi_instance *);
+
+int add_custom_processor(const char * name, processor_logic* logic);
+
+int delete_custom_processor(const char * name);
+
+int transfer_to_relationship(flow_file_record * ffr, processor_session * ps, const char * relationship);
 
 /****
  * ##################################################################
--- a/nanofi/include/api/nanofi.h
+++ b/nanofi/include/api/nanofi.h
@@ -40,8 +40,15 @@ extern "C" {
 #define SUCCESS_RELATIONSHIP "success"
 #define FAILURE_RELATIONSHIP "failure"
 
+/**
+ * Enables logging (disabled by default)
+ */
 void enable_logging();
 
+/**
+ * Sets terminate callback. The callback is executed upon termination (undhandled exception in C++ backend)
+ * @param terminate_callback the callback to execute
+ */
 void set_terminate_callback(void (*terminate_callback)());
 
 /****
@@ -50,11 +57,27 @@ void set_terminate_callback(void (*terminate_callback)());
  * ##################################################################
  */
 
+/**
+ * Creates a new MiNiFi instance
+ * @param url remote URL the instance connects to
+ * @param port remote port the instance connects to
+ * @return pointer to the new instance
+ */
 nifi_instance *create_instance(const char *url, nifi_port *port);
 
-void initialize_instance(nifi_instance *);
+/**
+ * Initialize remote connection of instance for transfers
+ * @param instance
+ */
+void initialize_instance(nifi_instance * instance);
 
-void free_instance(nifi_instance*);
+/**
+ * Frees instance
+ * @attention Any action on flows that belong to the freed instance are undefined after this is done!
+ * It's recommended to free all flows before freeing the instance.
+ * @param instance instance to be freed
+ */
+void free_instance(nifi_instance * instance);
 
 /****
  * ##################################################################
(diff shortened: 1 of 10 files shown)


nanofi/src/api/nanofi.cpp
```diff
@@ -223,13 +226,15 @@ void free_flowfile(flow_file_record *ff) {
     return;
   }
   auto content_repo_ptr = static_cast<std::shared_ptr<minifi::core::ContentRepository>*>(ff->crp);
-  if (content_repo_ptr->get()) {
+  if (content_repo_ptr->get() && (ff->keepContent == 0)) {
```
arpadboda: @phrocker : I don't link this, but I have no better idea yet. We provide a flow_file_record to access content and attributes in the processor logic callback. This means that the content cannot be deleted in case the flowfile is transfered via a relationship.

nanofi/include/cxx/CallbackProcessor.h
```diff
@@ -65,24 +67,19 @@ class CallbackProcessor : public core::Processor {
...
-    core::Relationship Success("success", "description");
-    relationships.insert(Success);
-    setSupportedRelationships(relationships);
-  }
+  virtual void initialize() override;
 
  protected:
   void *objref_;
-  std::function<void(core::ProcessSession*)> callback_;
+  std::function<void(core::ProcessSession*, core::ProcessContext *context)> callback_;
```
arpadboda: Context is required in the callback to:
-Access content
-Access properties (not yet supported, but will be)

nanofi/src/api/nanofi.cpp
phrocker: Yeah that seems like it could cause issues when freeing the flow file, but if they were to do all of this, this means that they've either threaded the reads of content or are calling free in the middle of accessing that content? I think it's reasonable for each read to "keep the content" because it's not the owner. A better way might be to use the reference counting in ResourceClaim but we don't presently have that since we make new claim objects. I think what you've done is not unreasonable.

arpadboda:
> I just skimmed the latest commit. Really good stuff. This will go a long way toward helping consumers of this API. Very excited!
> 
> Is this ready for review? If so I'll try to take a look this weekend sometime. Thanks!

Not yet, plan to finish today.

nanofi/include/api/nanofi.h
```diff
@@ -71,62 +94,173 @@ typedef int c2_start_callback(char *);
...
 void enable_async_c2(nifi_instance *, C2_Server *, c2_stop_callback *, c2_start_callback *, c2_update_callback *);
 
+/**
+ * Creates a new, empty flow
+ * @param instance the instance new flow will belong to
+ * @return a pointer to the created flow
+ **/
+flow *create_new_flow(nifi_instance * instance);
 
-uint8_t run_processor(const processor *processor);
```
arpadboda: I removed this as it has no implementation, so pretty sure noone ever used.

nanofi/include/api/nanofi.h
```diff
@@ -71,62 +94,173 @@ typedef int c2_start_callback(char *);
...
-processor *add_processor(flow *, const char *);
+/**
+ * Extend a flow with a new processor
+ * @param flow the flow to be extended with the new processor
+ * @param name name of the new processor
+ * @return pointer to the new processor or nullptr in case it cannot be instantiated (wrong name?)
+ */
+processor *add_processor(flow * flow, const char * name);
 
 processor *add_python_processor(flow *, void (*ontrigger_callback)(processor_session *session));
```
arpadboda: Didn't document as I'm not sure this is the right place for this.
This is only used by Python libraries, so I don't think this should be public and documented.

nanofi/include/api/nanofi.h
```diff
@@ -71,62 +94,173 @@ typedef int c2_start_callback(char *);
...
+ * Get all flow files of the given flow
+ * @param instance the instance the flow belongs to
+ * @param flow the flow to get flowfiles from
+ * @param flowfiles target area to copy the flowfiles to
+ * @param size the maximum number of flowfiles to copy to target (size of target)
+ * @return the number of flow files copies to target. Less or equal to size.
+ **/
+size_t get_flow_files(nifi_instance * instance, flow * flow, flow_file_record ** flowfiles, size_t size);
 
 flow_file_record *get(nifi_instance *,flow *, processor_session *);
```
arpadboda: This is Python only as well.

phrocker:
taking a look

nanofi/include/api/nanofi.h
```diff
@@ -68,60 +94,173 @@ typedef int c2_start_callback(char *);
...
+flow *create_new_flow(nifi_instance * instance);
 
-uint8_t run_processor(const processor *processor);
-
-flow *create_new_flow(nifi_instance *);
 
-flow *create_flow(nifi_instance *, const char *);
+/**
+ * Creates new flow and adds the first processor in case a valid name is provided
+ * @deprecated  as there is no proper indication of processor adding errors,
```
phrocker: I'm unclear about deprecation. Does this mean that it's deprecated because you have no way of indicating the processor could not be instantiated? Wouldn't that arrive at the case where flow is null? Flow in create_new_flow could be null by virtue of a malloc error, so why deprecate a function that results in the same behavior and fewer function calls?

nanofi/include/api/nanofi.h
```diff
@@ -68,60 +94,173 @@ typedef int c2_start_callback(char *);
...
+ * @attention in case first processor is empty or doesn't name any existing processor, an empty flow is returned.
+ * @return a pointer to the created flow
+ **/
+DEPRECATED flow *create_flow(nifi_instance * instance, const char * first_processor);
 
-flow *create_getfile(nifi_instance *instance, flow *parent, GetFileConfig *c);
+/**
+ * Add a getfile processor to "parent" flow.
+ * Creates new flow in instance in case "parent" is nullptr
+ * @deprecated as getfile processor can be added using "add_processor" function,
```
phrocker: This is precisely something we want to keep by design. Let's talk about this before merging this.

nanofi/include/api/nanofi.h
phrocker: This was starved off due to other work, but we will eventually need directory specific accessor functions for languages , I would imagine. This could be a candidate for that.

nanofi/include/api/nanofi.h
arpadboda: The problematic case is flow being instantiated properly, but the processor doesn't (the name is wrong for eg.).
In this case a valid flow pointer is returned, but the flow doesn't contain any processor.
To make it worse:
-There is no function in the API to get (at least the number) of processors in the flow
-As a valid ptr was returned, the caller could expect everything to be fine, but that's not the case.

nanofi/include/api/nanofi.h
```diff
@@ -68,60 +94,173 @@ typedef int c2_start_callback(char *);
...
+ */
+void free_standalone_processor(standalone_processor* processor);
 
 /**
-* Register your callback to received flow files that the flow failed to process
-* The flow file ownership is transferred to the caller!
-* The first callback should be registered before the flow is used. Can be changed later during runtime.
-*/
+ * Register your callback to received flow files that the flow failed to process
+ * The flow file ownership is transferred to the caller!
```
phrocker: "transferred to the caller" of the callback or of this function?

arpadboda: Good shout, thanks, I will reword this.

nanofi/include/api/nanofi.h
```diff
@@ -68,60 +94,173 @@ typedef int c2_start_callback(char *);
...
+ * @param name name of the property
+ * @param value value of the property
+ * @return 0 in case of success, -1 otherwise. Always succeeds unless instance or name is nullptr/emtpy.
+ **/
+int set_instance_property(nifi_instance *instance, const char * name, const char * value);
 
-int free_flow(flow *);
+/**
+ * Get a property. Should be used in custom processor logic callbacks.
+ * @attention The returned value transfers ownership, it's the callers responsibility to free it!
```
phrocker: Am I interpreting this correctly in that the library calls malloc for a new character array and copies the memory as opposed to requiring the API user do that allocation?  This doesn't seem standard in the way that C does this.

nanofi/include/api/nanofi.h
phrocker: I think this will be used across languages and very much by custom C processors.

nanofi/include/api/nanofi.h
```diff
@@ -68,60 +94,173 @@ typedef int c2_start_callback(char *);
...
 flow_file_record *invoke_file(standalone_processor* proc, const char* path);
 
-flow_file_record *invoke_chunck(standalone_processor* proc, uint8_t* buf, uint64_t);
+/**
+ * Invoke a standalone processor with some in-memory data
+ * @param buf specifies the beginning of the input buffer
+ * @param size specifies the size of the buffer
+ * @return a flow file record or nullptr in case no flowfile was generated
+ **/
+flow_file_record *invoke_chunck(standalone_processor* proc, uint8_t* buf, uint64_t size);
```
phrocker: chunk?

nanofi/include/api/nanofi.h
```diff
@@ -135,16 +274,62 @@ flow_file_record* create_ff_object(const char *file, const size_t len, const uin
...
 flow_file_record* create_ff_object_na(const char *file, const size_t len, const uint64_t size);
 
-void free_flowfile(flow_file_record*);
+/**
+ * Get incoming flow file. To be used in processor logic callbacks.
+ * @param session current processor session
+ * @param context current processor context
+ * @return a flow file record or nullptr in case there is none in the session
+ **/
+flow_file_record* get_flowfile(processor_session* session, processor_context* context);
```
phrocker: So get has been replaced by get_flowfile ? Any reason we can't use this in python?

nanofi/include/api/nanofi.h
```diff
@@ -135,16 +274,62 @@ flow_file_record* create_ff_object(const char *file, const size_t len, const uin
...
+ * @return 0 in case of success, -1 otherwise (no such attribute)
+ **/
+uint8_t get_attribute(const flow_file_record *ff, attribute *caller_attribute);
 
+/**
+ * Get the quantity of attributes
+ * @param ff flow file
+ * @return the number of attributes
+ **/
 int get_attribute_qty(const flow_file_record* ff);
```
phrocker: I thought there was a ticket for this ( can't find it so maybe it was never made ?? ) but we shouldn't be using qty or similar within an API. This should be count or quantity.

nanofi/include/api/nanofi.h
```diff
@@ -165,6 +355,31 @@ uint8_t remove_attribute(flow_file_record*, char *key);
 
 int transmit_flowfile(flow_file_record *, nifi_instance *);
 
+/**
+ * Adds a custom processor for later instantiation
+ * @param name name of the processor
+ * @param logic the callback to be invoked when the processor is triggered
+ * @return 0 on success, -1 otherwise (name already in use for eg.)
+ **/
+int add_custom_processor(const char * name, processor_logic* logic);
```
phrocker: processor_logic doesn't encapsulate the potential of "scheduling" May want to keep that fundamentally so that we allow configuration items to be loaded only once. The Python code left that as a todo

nanofi/include/core/cstructs.h
```diff
@@ -114,4 +126,6 @@ typedef enum FS {
   ROLLBACK
 } FailureStrategy;
 
+typedef void (processor_logic)(processor_session*, processor_context *);
```
phrocker: You have a comment, below, about accessing properties not being supported. Does that mean that processor_logic will go in tandem with processor configuration somehow?

nanofi/include/api/nanofi.h
arpadboda: I think we can, will give it a try

nanofi/include/api/nanofi.h
arpadboda: I don't think custom C processors can benefit from this at the moment as ptrs to "instance" and "flow" are not available in processor logic callbacks. That's the reason I created "get_flowfile".
Would you mind deprecating this in case Python works with the new one, too?

nanofi/include/api/nanofi.h
arpadboda: Good spot, fixed!

nanofi/include/api/nanofi.h
arpadboda: Ok, removed deprecation.

nanofi/include/api/nanofi.h
arpadboda: Unfortunately yes, you are correct.
The problem is "getProperty" returning string (not reference), which is a temporary object, so c_str() cannot be used as a return value. 
That's why there is an "attention" mark here, otherwise it would require an additional callback or change the internal API (change getProperty)

nanofi/include/api/nanofi.h
phrocker: The nomenclature that's common in NiFi is get, so we should work toward keeping some commonality, but get(...) can be changed to use the structs you've defined as it's new within this release and there are no guarantees. I'll create a follow on ticket (MINIFICPP-692). At the very least the consistency should exist be the language python, ruby, or C in how funcitons get accessed for obtaining the flowfile. Naming is secondary to that need.

nanofi/include/api/nanofi.h
phrocker: Could we not change it to return a null pointer?

nanofi/include/api/nanofi.h
phrocker: MINIFICPP-693 is a follow on for this.

nanofi/include/api/nanofi.h
phrocker: Typically the session can define a mapping of properties and can call a "scheduling" piece before every iteration. In some cases, this could be at the outset of execution of a custom processor. Effectively before their actions are invoked, the processor should be configured. We can make that stipulation a priori unless properties are dynamic, in which case we've sort of tabled that for the interim. MINIFICPP-694 is a good follow on to that.

phrocker:
Also added https://issues.apache.org/jira/browse/MINIFICPP-695 since examples don't appear to work in testing. Merging anyway so we can fix that as a follow on

PR #453, opened 2018-11-29
POC: static property improvements

arpadboda:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/core/Property.h
+++ b/libminifi/include/core/Property.h
@@ -33,6 +33,8 @@
 #include <stdlib.h>
 #include <math.h>
 #include "utils/StringUtils.h"
+#include "PropertyUtil.h"
+#include "PropertyValidator.h"
 
 namespace org {
 namespace apache {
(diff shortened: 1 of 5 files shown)


arpadboda:
Overview:
-Please read tests first. They should give an idea of the motivation of the change
-This is POF, some parts of the code might leave space for improvement, some are not yet implemented.

Pros of suggested solution:
-Fully backward-compatible, no change required in processors. 
-All property types can be assigned to Property (string property) without loosing the validator.
-Values are still stored in string, so no change in overhead of getter
-Easily extendable with further validators,  properties
-Validators can provide data in any format for server-side validation, no duplication of data

Cons:
-Templates, templates, templates everywhere... Added some static asserts to help development with meaningful error messages.

phrocker:
Thanks for submitting this. I think there is definite overlap between work I previously had with MINIFICPP-479 (https://github.com/apache/nifi-minifi-cpp/compare/master...phrocker:MINIFICPP-479?expand=1) I collected commits from across computers and squashed them, but has the side effect that it does validation at flow file load time ( something we've desperately needed) -- giving us a much earlier failure. a bit different implementation -- and it informs c2 of java validators to be used.

what is the motivation of this PR beyond that work that we discussed offline? I think the same "pros" exist for both works. That work has been around since June -- so it's probably stale in its ideas, so I'll take a look at this at some point. Thanks again.

PR #455, opened 2018-12-05
MINIFICPP-689 - Make minifi::Exception constructible with string param

arpadboda:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/Exception.h
+++ b/libminifi/include/Exception.h
@@ -60,16 +60,17 @@ class Exception : public std::exception {
  public:
   // Constructor
   /*!
-   * Create a new flow record
+   * Create a new exception
    */
-  Exception(ExceptionType type, const char *errorMsg)
+  Exception(ExceptionType type, std::string errorMsg)
       : _type(type),
-        _errorMsg(errorMsg) {
+        _errorMsg(std::move(errorMsg)) {
   }
+
   // Destructor
-  virtual ~Exception() throw () {
+  virtual ~Exception() noexcept {
   }
-  virtual const char * what() const throw () {
+  virtual const char * what() const noexcept {
 
     _whatStr = ExceptionTypeToString(_type);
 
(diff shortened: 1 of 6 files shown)


libminifi/include/Exception.h
```diff
@@ -60,16 +60,17 @@ class Exception : public std::exception {
...
   // Constructor
   /*!
-   * Create a new flow record
+   * Create a new exception
    */
-  Exception(ExceptionType type, const char *errorMsg)
+  Exception(ExceptionType type, std::string errorMsg)
       : _type(type),
-        _errorMsg(errorMsg) {
+        _errorMsg(std::move(errorMsg)) {
```
phrocker: Just saw this when I was typing the one below, I'll run the build before merging obviously, but does this not generate an elision warning?

libminifi/include/Exception.h
```diff
@@ -60,16 +60,17 @@ class Exception : public std::exception {
...
   }
+
   // Destructor
-  virtual ~Exception() throw () {
+  virtual ~Exception() noexcept {
```
phrocker: I didn't notice that throw was here. Odd choice. The compiler usually generates an implicit noexcept unless told otherwise, so I'm tempted to do a little more testing of this across compilers. I'll do this before merge.

libminifi/include/Exception.h
arpadboda: As far as I remember that warning is only applied for return values, where moving could prevent copy-elision, but it's not the case.

libminifi/include/Exception.h
arpadboda: throw() == noexcept, however throw() is deprecated since C++11.

phrocker: I'm not sure how that applies to my comment...but I think the end result is that I will still run basic tests across compilers. Are you suggesting we needn't run tests for verification and validation?

libminifi/include/Exception.h
phrocker: Ah yah. I don't remember either, but that sounds right  If I don't see the warning when I run the build then all is well that ends well. Thanks.

PR #460, opened 2018-12-12
MINIFICPP-479: Add processor property descriptor updates with c2 validation

phrocker:

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/include/core/state/Value.h
```diff
@@ -74,94 +157,215 @@ class BoolValue : public Value {
...
   template<typename T>
-  auto operator=(
-      const T ref) -> typename std::enable_if<std::is_same<T, int >::value ||
-      std::is_same<T, uint32_t >::value ||
-      std::is_same<T, uint64_t >::value ||
-      std::is_same<T, bool >::value ||
-      std::is_same<T, char* >::value ||
-      std::is_same<T, const char* >::value ||
-      std::is_same<T, std::string>::value,ValueNode&>::type {
+  auto operator=(const T ref) -> typename std::enable_if<std::is_same<T, int >::value ||
```
arpadboda: Guess std::is_integral<T> would help here reducing this.

libminifi/include/core/state/Value.h
```diff
@@ -74,94 +157,215 @@ class BoolValue : public Value {
...
+    return true;
+  }
+
+  virtual bool getValue(bool &ref) {
+    return false;
+  }
 
-static inline std::shared_ptr<Value> createValue(
-    const bool &object) {
+  int64_t value;
```
arpadboda: Honestly I don't really like to hanem a member with different type but same name in each derived type. 
Why don't simply create value as a template class and use "using" directive to create Int64Value as value<int64_t> ? 
In my opinion it would require less code duplication.

libminifi/src/processors/LogAttribute.cpp
```diff
@@ -38,13 +38,23 @@ namespace apache {
...
-core::Property LogAttribute::AttributesToLog("Attributes to Log", "A comma-separated list of Attributes to Log. If not specified, all attributes will be logged.", "");
-core::Property LogAttribute::AttributesToIgnore("Attributes to Ignore", "A comma-separated list of Attributes to ignore. If not specified, no attributes will be ignored.", "");
-core::Property LogAttribute::LogPayload("Log Payload", "If true, the FlowFile's payload will be logged, in addition to its attributes;"
-                                        "otherwise, just the Attributes will be logged.",
-                                        "false");
-core::Property LogAttribute::LogPrefix("Log prefix", "Log prefix appended to the log lines. It helps to distinguish the output of multiple LogAttribute processors.", "");
+
+core::Property LogAttribute::LogLevel(
+    core::PropertyBuilder::createProperty("Log Level")->withDescription("The Log Level to use when logging the Attributes")->withAllowedValue<std::string>("info")->withAllowedValue("trace")
+        ->withAllowedValue("error")->withAllowedValue("warn")->withAllowedValue("debug")->withDefaultValue("info")->build());
```
arpadboda: I don't like this, withAllowedValue should be implemented as 
```
template<typename T>
withAllowedValue(std::set<T>) {
  ...
}
```

This way it can be done as:
```
withAllowedValue({"Red", "Green", "Blue"))
```

libminifi/include/core/PropertyValidation.h
```diff
@@ -0,0 +1,347 @@
...
+ public:
+  PortValidator(const std::string &name)
+      : LongValidator(name, 1, 65535) {
+  }
+  virtual ~PortValidator() {
+
+  }
+};
+
+class DataTimeValidator : public PropertyValidator {
```
arpadboda: Dat**e**TimeValidator?

libminifi/src/processors/LogAttribute.cpp
phrocker: I think that would be "withAllowedValues". That's a good point. What you don't like is an artifact of a fluent API and is something that withAllowedValues would solve. Good idea.

libminifi/include/core/state/Value.h
phrocker: I think that's probably a differing opinion of style. I don't like using "using" directives and don't think the duplication is enough to warrant its usage. I have and do use "using" but very sparingly.

libminifi/include/core/state/Value.h
phrocker: There were a few excluded types. Once I get around to implementing those I'd prefer to use integral too

libminifi/src/c2/C2Agent.cpp
```diff
@@ -747,6 +752,20 @@ void C2Agent::handle_update(const C2ContentResponse &resp) {
...
+ */
+bool C2Agent::update_property(const std::string &property_name, const std::string &property_value, bool persist) {
+  if (update_service_->canUpdate(property_name)) {
+    configuration_->set(property_name, property_value);
+    if (persist) {
+      configuration_->persistProperties();
+      return true;
+    }
+  }
+  return false;
```
arpadboda: This looks a bit strange.
Why does this function always return false in case persist=false?

libminifi/include/core/FlowConfiguration.h
```diff
@@ -70,7 +70,11 @@ class FlowConfiguration : public CoreComponent {
         logger_(logging::LoggerFactory<FlowConfiguration>::getLogger()) {
     controller_services_ = std::make_shared<core::controller::ControllerServiceMap>();
     service_provider_ = std::make_shared<core::controller::StandardControllerServiceProvider>(controller_services_, nullptr, configuration);
-    flow_version_ = std::make_shared<state::response::FlowVersion>("", "default", "");
+    std::string flowUrl = "", bucket_id = "default", flowId = "";
```
arpadboda: String are default constructed as empty, why do we need this?

extensions/civetweb/processors/ListenHTTP.cpp
```diff
@@ -40,12 +40,12 @@ core::Property ListenHTTP::SSLCertificate("SSL Certificate", "File containing PE
...
-    core::PropertyBuilder::createProperty("SSL Verify Peer")->withDescription("Whether or not to verify the client's certificate (yes/no)")->isRequired(false)->withAllowedValue<std::string>("yes")
-        ->withAllowedValue("no")->withDefaultValue("no")->build());
+    core::PropertyBuilder::createProperty("SSL Verify Peer")->withDescription("Whether or not to verify the client's certificate (yes/no)")->isRequired(false)->withAllowableValue<std::string>("yes")
+        ->withAllowableValue("no")->withDefaultValue("no")->build());
 
 core::Property ListenHTTP::SSLMinimumVersion(
     core::PropertyBuilder::createProperty("SSL Minimum Version")->withDescription("Minimum TLS/SSL version allowed (SSL2, SSL3, TLS1.0, TLS1.1, TLS1.2)")->isRequired(false)
-        ->withAllowedValue<std::string>("SSL2")->withAllowedValue("SSL3")->withAllowedValue("TLS1.0")->withAllowedValue("TLS1.1")->withAllowedValue("TLS1.2")->withDefaultValue("SSL2")->build());
+        ->withAllowableValue<std::string>("SSL2")->withAllowableValue("SSL3")->withAllowableValue("TLS1.0")->withAllowableValue("TLS1.1")->withAllowableValue("TLS1.2")->withDefaultValue("SSL2")->build());
```
arpadboda: This could be changed to use the new version with set param.

libminifi/src/c2/C2Agent.cpp
phrocker: because it wasn't persisted and that's the semantics

libminifi/include/core/FlowConfiguration.h
phrocker: I used auto and then let it autobox ( from the assignment ) but windows failed on the compilation of that -- I changed it without regard to the sets. I'll try to remember changing this. thanks!

extensions/civetweb/processors/ListenHTTP.cpp
phrocker: I left it there as an example that both ways are allowed.

apiri:
reviewing

libminifi/include/core/state/Value.h
```diff
@@ -46,26 +56,99 @@ class Value {
...
+    return true;
+  }
+
+  virtual bool getValue(int64_t &ref) {
+    ref = std::stoll(string_value);
+    return true;
+  }
+
+  virtual bool getValue(uint64_t &ref) {
+    ref = std::stoull(string_value);
```
arpadboda: These throw. What's the intent here?

Are we happy with an unhandled exception in case the getvalue call fails or shall we catch and set return value based on that?

libminifi/include/core/state/Value.h
```diff
@@ -46,26 +56,99 @@ class Value {
...
+    return true;
+  }
+
+  virtual bool getValue(bool &ref) {
+    std::istringstream(string_value) >> std::boolalpha >> ref;
```
arpadboda: stream >> operator has a return value to detect error, I guess that should be the return value. 

It's also not consistent with the std:stol approach above, which throws.

libminifi/include/core/state/Value.h
```diff
@@ -74,94 +157,215 @@ class BoolValue : public Value {
...
     return value;
   }
  protected:
+
+  virtual bool getValue(int &ref) {
+    return false;
+  }
+
+  virtual bool getValue(int64_t &ref) {
+    if (value < (std::numeric_limits<int64_t>::max)()) {
```
arpadboda: <= ? (just to make sure :) )

libminifi/include/core/state/Value.h
```diff
@@ -74,94 +157,215 @@ class BoolValue : public Value {
...
+    ref = value;
+    return true;
+  }
+
+  virtual bool getValue(uint64_t &ref) {
+    if (value >= 0) {
+      ref = value;
+      return true;
+    }
+    return true;
```
arpadboda: I would expect this to be false

libminifi/include/io/validation.h
```diff
@@ -27,6 +27,21 @@
  * A checker that will, at compile time, tell us
  * if the declared type has a size method.
```
arpadboda: This comment belongs to the function below (size_function_functor_checker )

libminifi/src/c2/protocols/RESTProtocol.cpp
```diff
@@ -261,29 +280,49 @@ rapidjson::Value RESTProtocol::serializeJsonPayload(const C2Payload &payload, ra
...
 
-  std::map<std::string, std::list<rapidjson::Value*>> children;
+  std::vector<ValueObject> children;
 
   for (const auto &nested_payload : payload.getNestedPayloads()) {
     rapidjson::Value* child_payload = new rapidjson::Value(serializeJsonPayload(nested_payload, alloc));
 
-    children[nested_payload.getLabel()].push_back(child_payload);
+    if (nested_payload.isCollapsible()) {
+      bool combine = false;
```
arpadboda:
```
    bool combine = false;
    if (nested_payload.isCollapsible()) {
      for (auto &subordinate : children) {
        if (subordinate.name == nested_payload.getLabel()) {
          subordinate.values.push_back(child_payload);
          combine = true;
          break;
        }
      }
    } 
   if(!combine)
   {
      ValueObject obj;
      obj.name = nested_payload.getLabel();
      obj.values.push_back(child_payload);
      children.push_back(obj);
    }
```

Code duplication can be avoided this way.

libminifi/include/io/validation.h
```diff
@@ -27,6 +27,21 @@
...
  */
+template<typename T>
+class empty_function_functor_checker {
```
arpadboda: What's the reason for this? To use empty(), which is guaranteed to be O(1), while size() doesn't?

libminifi/src/core/PropertyValidation.cpp
```diff
@@ -0,0 +1,40 @@
...
+namespace minifi {
+namespace core {
+
+std::shared_ptr<PropertyValidator> StandardValidators::VALID = std::make_shared<AlwaysValid>(true, "VALID");
+StandardValidators::StandardValidators() {
+  INVALID = std::make_shared<AlwaysValid>(false, "INVALID");
+  INTEGER_VALIDATOR = std::make_shared<IntegerValidator>("INTEGER_VALIDATOR");
+  LONG_VALIDATOR = std::make_shared<LongValidator>("LONG_VALIDATOR");
+  UNSIGNED_LONG_VALIDATOR = std::make_shared<UnsignedLongValidator>("LONG_VALIDATOR");
+  SIZE_VALIDATOR = std::make_shared<DataSizeValidator>("DATA_SIZE_VALIDATOR");
```
apiri: would be good to make this DATA_SIZE_VALIDATOR for consistency and to pair with the data size value

libminifi/include/io/validation.h
phrocker: yep just a switch to empty.

libminifi/include/core/state/Value.h
phrocker: That would be inconsistent with prior behavior I'm afraid.

libminifi/include/core/state/Value.h
phrocker: Yeah, and that's handled by the call path. Logs should make that evident when run in these cases (hopefully).

libminifi/include/core/TypedValues.h
```diff
@@ -0,0 +1,232 @@
...
+  // Convert String to Integer
+  template<typename T>
+  static bool StringToInt(const std::string &input, T &output) {
+    if (input.size() == 0) {
+      return false;
+    }
+
+    const char *cvalue = input.c_str();
+    char *pEnd;
+    auto ival = std::strtoll(cvalue, &pEnd, 0);
```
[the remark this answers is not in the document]
phrocker: strtoll is a signed value.

libminifi/include/core/state/Value.h
phrocker: thanks!

libminifi/include/FlowController.h
```diff
@@ -175,21 +175,27 @@ class FlowController : public core::controller::ControllerServiceProvider, publi
...
   }
 
+  virtual std::string getComponentUUID() const {
+    utils::Identifier ident;
+    root_->getUUID(ident);
+    return ident.to_string();
+  }
+
   // get version
   virtual std::string getVersion() {
```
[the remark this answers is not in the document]
phrocker: I didn't due to this existing a while, but will make the change in 707. thanks!

libminifi/include/core/state/Value.h
phrocker: thanks!

PR #470, opened 2019-01-09
MINIFICPP-706 - RawSiteToSite: remove code duplication

arpadboda:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/sitetosite/RawSocketProtocol.h
+++ b/libminifi/include/sitetosite/RawSocketProtocol.h
@@ -112,9 +112,6 @@ class RawSiteToSiteClient : public sitetosite::SiteToSiteClient {
 
   }
 
-  void setPeer(std::unique_ptr<SiteToSitePeer> peer) {
-    peer_ = std::move(peer);
-  }
   /**
    * Provides a reference to the time out
    * @returns timeout
@@ -146,18 +143,14 @@ class RawSiteToSiteClient : public sitetosite::SiteToSiteClient {
   virtual int writeRequestType(RequestType type);
   // read Request Type
   virtual int readRequestType(RequestType &type);
+
   // read Respond
   virtual int readRespond(const std::shared_ptr<Transaction> &transaction, RespondCode &code, std::string &message);
   // write respond
   virtual int writeRespond(const std::shared_ptr<Transaction> &transaction, RespondCode code, std::string message);
   // getRespondCodeContext
   virtual RespondCodeContext *getRespondCodeContext(RespondCode code) {
-    for (unsigned int i = 0; i < sizeof(SiteToSiteRequest::respondCodeContext) / sizeof(RespondCodeContext); i++) {
-      if (SiteToSiteRequest::respondCodeContext[i].code == code) {
-        return &SiteToSiteRequest::respondCodeContext[i];
-      }
-    }
-    return NULL;
+    return SiteToSiteClient::getRespondCodeContext(code);
   }
 
   // Creation of a new transaction, return the transaction ID if success,
--- a/libminifi/include/sitetosite/SiteToSiteClient.h
+++ b/libminifi/include/sitetosite/SiteToSiteClient.h
@@ -221,12 +221,8 @@ class SiteToSiteClient : public core::Connectable {
   // deleteTransaction
   virtual void deleteTransaction(std::string transactionID);
 
-  virtual void tearDown();
-
-  // write Request Type
-  virtual int writeRequestType(RequestType type);
-  // read Request Type
-  virtual int readRequestType(RequestType &type);
+  virtual void tearDown() = 0;
+  
   // read Respond
   virtual int readResponse(const std::shared_ptr<Transaction> &transaction, RespondCode &code, std::string &message);
   // write respond
--- a/libminifi/src/sitetosite/RawSocketProtocol.cpp
+++ b/libminifi/src/sitetosite/RawSocketProtocol.cpp
@@ -395,97 +395,37 @@ bool RawSiteToSiteClient::getPeerList(std::vector<PeerStatus> &peers) {
   }
 }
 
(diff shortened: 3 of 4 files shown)


phrocker:
Changes inherently look fine. Will run some tests before approval. Hope to send an email soon re a release so I'll wait for 0.7 to merge it. Thanks

libminifi/include/sitetosite/RawSocketProtocol.h
```diff
@@ -143,21 +140,21 @@ class RawSiteToSiteClient : public sitetosite::SiteToSiteClient {
   // tearDown
   virtual void tearDown();
   // write Request Type
-  virtual int writeRequestType(RequestType type);
+  virtual int writeRequestType(RequestType type) {
+    return SiteToSiteClient::writeRequestType(type);
```
phrocker: I think the code being in SiteToSiteClient is vestigial. It probably shouldn't be there. That also would reduce the risk of any testing. Just remove it from the base class. The only other implementation ( and others I have in mind ) wouldn't need this.

libminifi/src/sitetosite/RawSocketProtocol.cpp
```diff
@@ -394,98 +394,12 @@ bool RawSiteToSiteClient::getPeerList(std::vector<PeerStatus> &peers) {
     return false;
   }
 }
-
-int RawSiteToSiteClient::writeRequestType(RequestType type) {
```
phrocker: ^ Same as comments above, This can stay and we can remove the duplicate in the base.

libminifi/include/sitetosite/RawSocketProtocol.h
arpadboda: Okay, good shout, this helped to remove even more duplications. :]

libminifi/src/sitetosite/RawSocketProtocol.cpp
arpadboda: Done, it seems github failed to handle the diff visualisaiton properly.

libminifi/include/sitetosite/SiteToSiteClient.h
```diff
@@ -221,12 +221,8 @@ class SiteToSiteClient : public core::Connectable {
   // deleteTransaction
   virtual void deleteTransaction(std::string transactionID);
 
-  virtual void tearDown();
-
-  // write Request Type
-  virtual int writeRequestType(RequestType type);
-  // read Request Type
-  virtual int readRequestType(RequestType &type);
+  virtual void tearDown() = 0;
```
phrocker: i don't think this is needed at all. is it used elsewhere? Teardown is protocol specific of when and how it occurs.

libminifi/src/sitetosite/RawSocketProtocol.cpp
phrocker: There probably don't need to be any changes in this file, right? Seems that there have been.

libminifi/include/sitetosite/SiteToSiteClient.h
arpadboda: It exist in both http and raw, with different implementation, so I guess it's fine here as pure virtual.

libminifi/src/sitetosite/RawSocketProtocol.cpp
```diff
@@ -395,97 +395,37 @@ bool RawSiteToSiteClient::getPeerList(std::vector<PeerStatus> &peers) {
...
-      return ret;
-    }
+    return peer_->writeUTF(SiteToSiteRequest::RequestTypeStr[type]);
   }
 
-  return -1;
-}
-
-int RawSiteToSiteClient::readRespond(const std::shared_ptr<Transaction> &transaction, RespondCode &code, std::string &message) {
-  uint8_t firstByte;
```
phrocker: This is again the case where the base class doesn't need to deal with the the innards of reading bytes. Probably means they can be removed from the base and made pure virtual since I believe they are overridden in http site to site if I'm not mistaken.

phrocker: This isn't valid any longer. I think some parent class details changed, but I will have to further investigate the HTTP portion. I want to fully understand this. If it is a bug that may push when this should be merged, but once I take a closer look at HTTP it may become obvious. What the parent currently does, is an invalid reliance from the end of HTTP, but since things changed in some of the accessory code it may never get there, meaning we can make it pure virtual; however, I'll take a closer look at it while running to validate that it's not a bug before going that route.

phrocker: I'll also take another look at this PR closely and approve for merge in 0.7.0 if it doesn't change status quo. Thanks.

phrocker: I didn't see a response for this. I have some time in the next few days to run tests. What have you been able to run? Thanks!

arpadboda: Two things I did:
-Execute site2site related unit tests
-Verified transfers from MiNiFi to NiFi using the C examples (which still rely on C++ S2S implementation atm)

phrocker: Yep, I usually do full scope: docker tests, site to site secure, unsecure, transfers -- in this case I'll do raw and http just to be safe ( in minifi note C agent ). -- once all that is done I'll take another look at the PR and approve for merge in 0.7.0. Thanks.

phrocker:
@arpadboda can you force travis and appveyor to re-run on this so that we can do a little more eval and before merging this?

arpadboda:
> @arpadboda can you force travis and appveyor to re-run on this so that we can do a little more eval and before merging this?

Sure, rebased to master, pushed.

PR #472, opened 2019-01-10
MINIFICPP-710 - Fix errors in Property StringToInt conversion

arpadboda:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/core/TypedValues.h
+++ b/libminifi/include/core/TypedValues.h
@@ -141,6 +141,47 @@ class TimePeriodValue : public TransformableValue, public state::response::UInt6
  * format <numeric> <byte size>.
  */
 class DataSizeValue : public TransformableValue, public state::response::UInt64Value {
+private:
+
+  //Signed
+  template<typename T, typename std::enable_if<std::is_signed<T>::value, T>::type* = nullptr>
+  static bool AssignIfFits(int64_t base, uint64_t mult, T& output) {
+    int64_t val = base * mult;
+    if(val >= (std::numeric_limits<T>::min)() && val <= (std::numeric_limits<T>::max)()) {
+      output = val;
+      return true;
+    }
+    return false;
+  }
+
+  //Unsigned
+  template<typename T, typename std::enable_if<!std::is_signed<T>::value, T>::type* = nullptr>
+  static bool AssignIfFits(uint64_t base, uint64_t mult, T& output) {
+    uint64_t val = base * mult;
+    if(val <= (std::numeric_limits<T>::max)()) {
+      output = val;
+      return true;
+    }
+    return false;
+  }
+
+  //Signed
+  template<typename T, typename std::enable_if<std::is_signed<T>::value, T>::type* = nullptr>
+  static bool StringToNum(const char* str_begin, char **str_end, T& output) {
+    int64_t val = std::strtoll(str_begin, str_end, 0);
+    return AssignIfFits<T>(val, 1, output);
+  }
+
+  //Unsigned
+  template<typename T, typename std::enable_if<!std::is_signed<T>::value, T>::type* = nullptr>
+  static bool StringToNum(const char* str_begin, char **str_end, T& output) {
+    if((str_begin[0] == '\0') || (str_begin[0] == '-')) {
+      return false;
+    }
+    uint64_t val = std::strtoull(str_begin, str_end, 0);
+    return AssignIfFits<T>(val, 1, output);
+  }
+
  public:
   static const std::type_index type_id;
 
@@ -163,7 +204,11 @@ class DataSizeValue : public TransformableValue, public state::response::UInt64V
 
     const char *cvalue = input.c_str();
     char *pEnd;
-    auto ival = std::strtoll(cvalue, &pEnd, 0);
+
+    T ival = 0;
+    if(!StringToNum(cvalue, &pEnd, ival)) {
+      return false;
+    }
 
     if (pEnd[0] == '\0') {
       output = ival;
@@ -195,8 +240,7 @@ class DataSizeValue : public TransformableValue, public state::response::UInt64V
             }
           }
         }
-        output = ival * multiplier;
-        return true;
+        return AssignIfFits<T>(ival, multiplier, output);
 
       } else if ((pEnd[1] == 'b' || pEnd[1] == 'B') && (pEnd[2] == '\0')) {
 
@@ -214,8 +258,7 @@ class DataSizeValue : public TransformableValue, public state::response::UInt64V
             }
           }
         }
-        output = ival * multiplier;
-        return true;
+        return AssignIfFits<T>(ival, multiplier, output);
       }
     }
 
--- a/libminifi/src/c2/protocols/RESTProtocol.cpp
+++ b/libminifi/src/c2/protocols/RESTProtocol.cpp
@@ -51,7 +51,7 @@ const C2Payload RESTProtocol::parseJsonResponse(const C2Payload &payload, const 
         identifier = root["identifier"].GetString();
       }
       if (root["requested_operations"].Size() == 0 && root["requestedOperations"].Size() == 0)
-        return std::move(C2Payload(payload.getOperation(), state::UpdateState::READ_COMPLETE, true));
+        return C2Payload(payload.getOperation(), state::UpdateState::READ_COMPLETE, true);
 
       C2Payload new_payload(payload.getOperation(), state::UpdateState::NESTED, true);
 
@@ -129,7 +129,7 @@ const C2Payload RESTProtocol::parseJsonResponse(const C2Payload &payload, const 
   } catch (...) {
   }
 #endif
-  return std::move(C2Payload(payload.getOperation(), state::UpdateState::READ_COMPLETE, true));
+  return C2Payload(payload.getOperation(), state::UpdateState::READ_COMPLETE, true);
 }
 
 void setJsonStr(const std::string& key, const state::response::ValueNode& value, rapidjson::Value& parent, rapidjson::Document::AllocatorType& alloc) {  // NOLINT
--- a/libminifi/test/unit/PropertyTests.cpp
+++ b/libminifi/test/unit/PropertyTests.cpp
@@ -107,3 +107,28 @@ TEST_CASE("Test Trimmer Left", "[testTrims]") {
   REQUIRE(test.c_str()[1] == ' ');
 }
 
+TEST_CASE("Test int conversion", "[testStringToInt]") {
+  using org::apache::nifi::minifi::core::Property;
+
+  uint64_t uint64_var = 0;
+  REQUIRE(Property::StringToInt("-2", uint64_var) == false);  // Negative shouldn't be converted to uint
+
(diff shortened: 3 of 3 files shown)


libminifi/test/unit/PropertyTests.cpp
```diff
@@ -107,3 +107,28 @@ TEST_CASE("Test Trimmer Left", "[testTrims]") {
   REQUIRE(test.c_str()[1] == ' ');
 }
 
+TEST_CASE("Test int conversion", "[testStringToInt]") {
```
arpadboda: Most of these would fail without this change.

libminifi/src/c2/protocols/RESTProtocol.cpp
```diff
@@ -129,7 +129,7 @@ const C2Payload RESTProtocol::parseJsonResponse(const C2Payload &payload, const
   } catch (...) {
   }
 #endif
-  return std::move(C2Payload(payload.getOperation(), state::UpdateState::READ_COMPLETE, true));
```
arpadboda: These generated copy-elision warnings in clang, so removed.

libminifi/test/unit/PropertyTests.cpp
phrocker: Almost all of these are invalid and shouldn't be accepted, though. This was the point of https://issues.apache.org/jira/browse/MINIFICPP-707. I don't want to trounce on this PR, but the reason I created the ticket was that some of these SHOULD be prevented from being used at all in configs, fail early. Therefore it's not as simple as doing what is done here. 710 should be OBE in favor of 707.

libminifi/include/core/TypedValues.h
```diff
@@ -141,6 +141,47 @@ class TimePeriodValue : public TransformableValue, public state::response::UInt6
  * format <numeric> <byte size>.
  */
 class DataSizeValue : public TransformableValue, public state::response::UInt64Value {
+private:
+
+  //Signed
+  template<typename T, typename std::enable_if<std::is_signed<T>::value, T>::type* = nullptr>
+  static bool AssignIfFits(int64_t base, uint64_t mult, T& output) {
```
phrocker: These I like and think we should use this.

libminifi/include/core/TypedValues.h
```diff
@@ -163,7 +204,11 @@ class DataSizeValue : public TransformableValue, public state::response::UInt64V
 
     const char *cvalue = input.c_str();
     char *pEnd;
-    auto ival = std::strtoll(cvalue, &pEnd, 0);
+
+    T ival = 0;
+    if(!StringToNum(cvalue, &pEnd, ival)) {
```
phrocker: This is only applicable if we reach the null terminator and it is not a byte string

libminifi/src/c2/protocols/RESTProtocol.cpp
phrocker: Good catch. I rarely build with clang these days...

libminifi/include/core/TypedValues.h
```diff
@@ -141,6 +141,47 @@ class TimePeriodValue : public TransformableValue, public state::response::UInt6
...
+  template<typename T, typename std::enable_if<std::is_signed<T>::value, T>::type* = nullptr>
+  static bool StringToNum(const char* str_begin, char **str_end, T& output) {
+    int64_t val = std::strtoll(str_begin, str_end, 0);
+    return AssignIfFits<T>(val, 1, output);
+  }
+
+  //Unsigned
+  template<typename T, typename std::enable_if<!std::is_signed<T>::value, T>::type* = nullptr>
+  static bool StringToNum(const char* str_begin, char **str_end, T& output) {
+    if(str_begin[0] == '-') {
```
phrocker: can't assume that it's zero, and that it's non-empty

libminifi/include/core/TypedValues.h
```diff
@@ -141,6 +141,47 @@ class TimePeriodValue : public TransformableValue, public state::response::UInt6
...
+  template<typename T, typename std::enable_if<std::is_signed<T>::value, T>::type* = nullptr>
+  static bool AssignIfFits(int64_t base, uint64_t mult, T& output) {
+    int64_t val = base * mult;
+    if(val >= std::numeric_limits<T>::min() && val <= std::numeric_limits<T>::max()) {
+      output = val;
+      return true;
+    }
+    return false;
+  }
+
```
phrocker: Does this code compile in windows? I'm not sure that it will.

libminifi/test/unit/PropertyTests.cpp
phrocker: I didn't think my last comment was entirely helpful so I went through and tried to parse out the parts that are appropriate and SHOULD be accepted. Hopefully I was more clear on what we can immediately and what we can't accept without more discussion. If we can parse this out then we can approve and merge the good parts.

libminifi/include/core/TypedValues.h
phrocker: Appveyor showed that it didn't. Do you have access to windows? I can probably help with that since I do.

libminifi/test/unit/PropertyTests.cpp
```diff
@@ -107,3 +107,28 @@ TEST_CASE("Test Trimmer Left", "[testTrims]") {
...
+  using org::apache::nifi::minifi::core::Property;
+
+  uint64_t uint64_var = 0;
+  REQUIRE(Property::StringToInt("-2", uint64_var) == false);  // Negative shouldn't be converted to uint
+
+  uint32_t uint32_var = 0;
+  REQUIRE(Property::StringToInt("3  GB", uint32_var) == true);  // Test skipping of spaces, too
```
phrocker: Some of these tests exist in yaml if I recall correctly ( but I didn't write them ), if you're interested.

libminifi/test/unit/PropertyTests.cpp
```diff
@@ -107,3 +107,28 @@ TEST_CASE("Test Trimmer Left", "[testTrims]") {
   REQUIRE(test.c_str()[1] == ' ');
 }
 
+TEST_CASE("Test int conversion", "[testStringToInt]") {
+  using org::apache::nifi::minifi::core::Property;
+
+  uint64_t uint64_var = 0;
+  REQUIRE(Property::StringToInt("-2", uint64_var) == false);  // Negative shouldn't be converted to uint
```
phrocker: I realize 710 kind of had a different focus, but your comment on the prior PR in which you mentioned this was to open 707 in hopes of opening a discussion for better dealing with this. For example, it may be the case where sometimes a negative value makes sense AND the conversion make sense. I will do some research, as I know these cases exist. In others, that is NOT true. 

While the above test is totally valid from the perspective of 710, the hope was that we could discuss and more intelligently handle these cases, perhaps with the validators themselves. What is your perspective? Would you rather fix the low hanging fruit and move on or discuss alternatives?

libminifi/include/core/TypedValues.h
```diff
@@ -141,6 +141,47 @@ class TimePeriodValue : public TransformableValue, public state::response::UInt6
...
+  //Signed
+  template<typename T, typename std::enable_if<std::is_signed<T>::value, T>::type* = nullptr>
+  static bool StringToNum(const char* str_begin, char **str_end, T& output) {
+    int64_t val = std::strtoll(str_begin, str_end, 0);
+    return AssignIfFits<T>(val, 1, output);
```
phrocker: why have any multiplication at all if it is 1?

libminifi/include/core/TypedValues.h
arpadboda: Yes, please! I can check Appveyor err messages, but that's not so efficient in longer term.

libminifi/include/core/TypedValues.h
arpadboda: Good shout, thanks!
This also means adding a tectcase to check with empty and invalid strings to make sure it handles errors gracefully.

libminifi/include/core/TypedValues.h
arpadboda: Not sure I get the point, this does the same strto(u)ll call inside that was here before. 
The only change is the return value, which comes from AssignIfFits.

libminifi/test/unit/PropertyTests.cpp
arpadboda: Sure, let's focus on the long term solution.

Just scratched this as I spotted errors in the code, however as you correctly pointed out, my change doesn't include the integration approach.

libminifi/test/unit/PropertyTests.cpp
```diff
@@ -107,3 +107,28 @@ TEST_CASE("Test Trimmer Left", "[testTrims]") {
...
+
+  uint32_t uint32_var = 0;
+  REQUIRE(Property::StringToInt("3  GB", uint32_var) == true);  // Test skipping of spaces, too
+  uint32_t expected_value = 3u * 1024 * 1024 * 1024;
+  REQUIRE(uint32_var == expected_value);
+
+  int32_t int32_var = 0;
+  REQUIRE(Property::StringToInt("3GB", int32_var) == false);  // Doesn't fit
+
+  REQUIRE(Property::StringToInt("-1 G", int32_var) == true);
```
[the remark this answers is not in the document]
arpadboda: Okay, thanks, will change.

libminifi/test/unit/PropertyTests.cpp
```diff
@@ -107,3 +107,28 @@ TEST_CASE("Test Trimmer Left", "[testTrims]") {
...
+  REQUIRE(int32_var == -1 * 1000 * 1000 * 1000);
+
+  REQUIRE(Property::StringToInt("-1G", uint32_var) == false);  // Negative to uint
```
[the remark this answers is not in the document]
arpadboda: Agreed, however as you see, it's not allowed, this conversion fails.

Although I think the current implementation is still more forgiving than it should be according to the whole picture, so let's try to explore that before going ahead with this.

libminifi/test/unit/PropertyTests.cpp
```diff
@@ -107,3 +107,28 @@ TEST_CASE("Test Trimmer Left", "[testTrims]") {
...
+  uint64_t uint64_var = 0;
+  REQUIRE(Property::StringToInt("-2", uint64_var) == false);  // Negative shouldn't be converted to uint
+
+  uint32_t uint32_var = 0;
+  REQUIRE(Property::StringToInt("3  GB", uint32_var) == true);  // Test skipping of spaces, too
+  uint32_t expected_value = 3u * 1024 * 1024 * 1024;
+  REQUIRE(uint32_var == expected_value);
+
+  int32_t int32_var = 0;
+  REQUIRE(Property::StringToInt("3GB", int32_var) == false);  // Doesn't fit
```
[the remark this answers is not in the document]
arpadboda: If we only aim for uint64, I think we should consider having this function only for uint_64t instead of the current template solution and provide possibility to safely cast for smaller types (like I did in AssignIfFits). 

As this would introduce a breaking change, not sure when it can be done properly.

szaszm:
Closing this because of inactivity

PR #482, opened 2019-02-11
Minificpp 658 - Port Raw Site to Site to C

arpadboda:

### For all changes:
     in the commit message?

### For code changes:

### Note:


nanofi/src/sitetosite/CRawSocketProtocol.c
```diff
@@ -0,0 +1,1016 @@
...
+      return -1;
+  }
+
+  // All known error cases handled here
+  //client->logger_->log_error("Site2Site HandShake Failed because destination port, %s, is %s", client->port_id_str_, error);
+  return -2;
+}
+
+
+/*bool CRawSiteToSiteClient::getPeerList(std::vector<CPeerStatus> &peers) {
```
arpadboda: Left this commented here as it's planned to be added as a follow-up.

nanofi/include/core/cstream.h
```diff
@@ -0,0 +1,73 @@
...
+
+#ifndef NIFI_MINIFI_CPP_CSTREAM_H
+#define NIFI_MINIFI_CPP_CSTREAM_H
+
+#include "cstructs.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
```
arpadboda: This helps unit-testing the C implementation in C++ env

nanofi/include/sitetosite/CSiteToSite.h
```diff
@@ -0,0 +1,423 @@
...
+
+static int write_uint16t(CTransaction * transaction, uint16_t base_value) {
+  const uint16_t value = is_little_endian() == 1 ? htons(base_value) : base_value;
+  const uint8_t *buf = (uint8_t *) (&value);
+
+  return writeData(transaction, buf, sizeof(uint16_t));
+}
+
+static int write_UTF_len(CTransaction * transaction, const char * str, size_t len, enum Widen widen) {
+  if (len > 65535) {
```
arpadboda: This reflects some of the changes in MINIFICPP-676 - payload doesn't get copied during transfer.

phrocker:
Will take a look thanks!

CMakeLists.txt
```diff
@@ -246,6 +246,7 @@ else()
 	include_directories("thirdparty/uuid/include/posix")
 endif()
 add_subdirectory("thirdparty/uuid")
+include_directories(thirdparty/ut)
```
phrocker: Is ut used within the main project(s)?

libminifi/include/sitetosite/RawSocketProtocol.h
```diff
@@ -178,9 +178,9 @@ class RawSiteToSiteClient : public sitetosite::SiteToSiteClient {
   virtual bool handShake();
 
  private:
-
+  /*
   // Mutex for protection
```
phrocker: Might as well remove it if it was not referenced anywhere.

nanofi/examples/transmit_payload.c
```diff
@@ -0,0 +1,77 @@
...
+
+  char *host = argv[1];
+  char *port_str = argv[2];
+  char *nifi_port_str = argv[3];
+
+  uint16_t port_num = atoi(port_str);
+
+  struct SiteToSiteCPeer peer;
+
+  initPeer(&peer, NULL, host, port_num, "");
```
phrocker: We should probably refine this over time as people likely won't want to define their "peers". That's mostly abstracted now, but I appreciate this being part of the API for sure.

nanofi/examples/transmit_payload.c
```diff
@@ -0,0 +1,77 @@
...
+
+  if(transmitPayload(&protocol, payload, &as) == 0){
+    printf("Packet successfully sent\n");
+  } else {
+    printf("Failed to send packet\n");
+  }
+
+  tearDown(&protocol);
+
+  freePeer(&peer);
```
phrocker: Is this freeing internal components of peer? I question because we're calling free on an object that wasn't "created" but instead was "inited"

nanofi/include/core/log.h
```diff
@@ -0,0 +1,76 @@
...
+  off = 0,
+  critical = 1,
+  err = 2,
+  warn = 3,
+  info = 4,
+  debug = 5,
+  trace = 6,
+} log_level;
+
+extern int global_log_level;
```
phrocker: may want to hint to the optimizer with volatile, generally this is a single CPU instruction but I wouldn't guarantee that to always be the case.

nanofi/include/sitetosite/CPeer.h
```diff
@@ -0,0 +1,134 @@
...
+  // URL
+  char * url_;
+
+  char * host_;
+
+  uint16_t port_;
+
+  enum Bool owns_socket_;
+
+  //io::NetworkInterface local_network_interface_;
```
phrocker: not needed? same for logger?

nanofi/include/sitetosite/CPeer.h
```diff
@@ -0,0 +1,134 @@
...
+  size_t host_len = strlen(hostname);
+  peer->host_ = (char*)malloc(host_len + 1); // +1 for trailing zero
+  peer->url_ = (char*)malloc(host_len + 14); // +1 for trailing zero, 1 for ':', at most 5 for port, 7 for "nifi://" suffix
+  memset(peer->url_, 0, host_len + 14); // make sure to have zero padding no matter the length of the port
+  strncpy(peer->host_, hostname, host_len);
+  strncpy(peer->url_, "nifi://", 7);
+  strncpy(peer->url_ + 7, hostname, host_len);
+  peer->host_[host_len] = '\0';
+  peer->url_[host_len + 7] = ':';
+  if(peer->port_ != 0) {
```
phrocker: Not sure it's ever appropriate to ever have an empty port, unless using HTTP,  in which case the scheme above would be incorrect. Further, why not just use a single snprintf versus strncpy(s). Further, what limits port to 5 digits? Is that done elsewhere?

nanofi/include/sitetosite/CPeer.h
```diff
@@ -0,0 +1,134 @@
...
+    sprintf(peer->url_ + host_len + 8, "%d", peer->port_);
+  }
+  return;
+}
+
+static void setPort(struct SiteToSiteCPeer * peer, uint16_t port) {
+  peer->port_ = port;
+  if(peer->url_ != NULL) {
+    int i;
+    for(i = 8; i < strlen(peer->url_); ++i) { //8 to begin searching for ':' after the "nifi://" suffix
```
phrocker: If code changes elsewhere then this specifically has to change too. Why not start from zero? Is this function used frequently enough that the starting point is material?

nanofi/include/sitetosite/CPeer.h
```diff
@@ -0,0 +1,134 @@
...
+      if(peer->url_[i] == ':'){
+        memset(peer->url_ + i + 1, 0, 6); // zero the port area  - the new port can be shorter
+        sprintf(peer->url_ + i + 1, "%d", peer->port_);
```
phrocker: any reason snprintf isn't used here?

nanofi/include/sitetosite/CPeer.h
arpadboda: It would just made the code ugly. 
The URL looks like:
```
nifi://<host>:<port>
```

Where the "nifi://" prefix is put by the code, host is user input. 

This code just looks for the 2nd occurrence of ":" in the string. As the position of the first is known, it was easier this way. 
If you have a better idea to do this, feel free to share.

nanofi/include/sitetosite/CPeer.h
arpadboda: Changed to snprintf, thanks. 
Anyway the parameter is uint16_t, so it's expected to be at most 5 chars (<65Kb)

nanofi/include/sitetosite/CPeer.h
arpadboda: Port is by default initialized to 0 in initPeer function in case port string is null. 
This just leaves the possibility to specify/change the port later. 

snprintf: what do we gain with that? it requires a format string as well, so the code is just getting longer. Furthermore I doubt it's any better performance-wise, given it has to deal with a format str and variadic args. 
Using the format string as a const string input generates warning on some compilers, so I would prefer not doing that.

Port: the input is uint16_t, which is always expected to fit 5 chars.

nanofi/include/sitetosite/CPeer.h
arpadboda: Removed, thanks.

nanofi/include/core/log.h
arpadboda: Good shout, added, thanks!

nanofi/examples/transmit_payload.c
arpadboda: It can happen, yes. Peer supports injecting socket. In case there is no socket injected (for eg.: the case above), peer created the internal socket based on the host and the port. In this case peer owns the socket, so closes and calls free. 
On naming, I agree it's not the best. 
Do you think tearDownPeer or shutDownPeer or something alike that would be more appropriate for this functionality?

libminifi/include/sitetosite/RawSocketProtocol.h
arpadboda: Removed, thanks!

CMakeLists.txt
arpadboda: No, but due to some cmake magic including in cmakelist.txt of nanofi itself wasn't enough. Not sure about the reasons, this way it worked.

nanofi/examples/transmit_payload.c
arpadboda: We can modify "initRawClient" to support creating the peer internally in case some additional arguments (host, port) are provided. 
It would be the same concept as peer creates the underlying socket now. 
Let me add it as a follow-up.

nanofi/include/sitetosite/CPeer.h
phrocker: If the host is the user input start from the end?

arpadboda: Thanks, changed.

nanofi/examples/transmit_payload.c
phrocker: I guess the answer to that question is would you, as a consumer of an API, care about peers and protocol objects?

nanofi/include/sitetosite/CPeer.h
```diff
@@ -0,0 +1,129 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+static const char MAGIC_BYTES[] = { 'N', 'i', 'F', 'i' };
+
+struct SiteToSiteCPeer;
+
+// open connection to the peer
+int openPeer(struct SiteToSiteCPeer * peer);
```
phrocker: Const pointer for these and below?

nanofi/include/sitetosite/CPeer.h
```diff
@@ -0,0 +1,129 @@
...
+  cstream * stream_;
+
+  // URL
+  char * url_;
+
+  char * host_;
+
+  uint16_t port_;
+
+  enum Bool owns_socket_;
```
phrocker: This may not always be socket based. Looking forward maybe a different name to convey ownership?

nanofi/include/sitetosite/CPeer.h
```diff
@@ -0,0 +1,129 @@
...
+        memset(peer->url_ + i + 1, 0, 6); // zero the port area  - the new port can be shorter
+        snprintf(peer->url_ + i + 1, 5, "%d", peer->port_);
+      }
+    }
+  }
+}
+
+static void initPeer(struct SiteToSiteCPeer * peer, cstream * injected_socket, const char * host, uint16_t port, const char * ifc) {
+  peer->stream_ = injected_socket;
+  //peer->local_network_interface_= std::move(io::NetworkInterface(ifc, nullptr));
```
phrocker: Is this comment needed or will a todo suffice?

nanofi/include/sitetosite/CPeer.h
```diff
@@ -0,0 +1,129 @@
...
+  setPort(peer, port);
+
+  if(peer->stream_ == NULL) {
+    peer->owns_socket_ = True;
+  }
+}
+
+static void freePeer(struct SiteToSiteCPeer * peer) {
+  closePeer(peer);
+  setHostName(peer, NULL);
```
phrocker: Freeing host names is followed from a setter? That leads me to confusion when reading the code. Do you feel comfortable with the possibility that someone free may be called within setHostName even if we can't own that memory? I ask because if a user calls free on peer peer, it's safe to call free and if it fails that is a developer error, but if they set hostnames in peer and then called in it they could unwillingly arrive down the same path. I'm on the fence with that API perspective.

nanofi/include/sitetosite/CRawSocketProtocol.h
```diff
@@ -0,0 +1,198 @@
...
+      return &respondCodeContext[i];
+    }
+  }
+  return NULL;
+}
+
+// RawSiteToSiteClient Class
+struct CRawSiteToSiteClient {
+  // Batch Count
+  uint64_t _batchCount;
```
phrocker: I know this was copied and changed to a struct, but can we change this to use the rest of the project's guidelines for variable naming?

nanofi/include/sitetosite/CPeer.h
arpadboda: A todo should be fine, thanks!

nanofi/include/sitetosite/CPeer.h
arpadboda: Good spot, thanks, will add where it makes sense!

nanofi/include/sitetosite/CPeer.h
arpadboda: Not sure I get your point. 
Peer copies the user-provided hostname into a "member", so we can always assume that the struct owns hostname. 
Here I could also write a null-check and free hostname in case it's not null, but this is exactly what setHostName does.

nanofi/include/sitetosite/CPeer.h
arpadboda: Makes sense, thanks!

nanofi/examples/transmit_payload.c
arpadboda: I have to admit that your question is more than reasonable. :)

I will update the API to handle peer inside the client, so client will be the only struct that the user needs to deal with (call init and teardown).

nanofi/CMakeLists.txt
```diff
@@ -25,14 +25,15 @@ ENDIF(POLICY CMP0048)
...
+include_directories(../thirdparty/ut)
 
 if(WIN32)
 include_directories(../libminifi/opsys/win)
 else()
 include_directories(../libminifi/opsys/posix)
 endif()
 
-file(GLOB NANOFI_SOURCES  "src/api/*.cpp" "src/core/*.cpp" "src/cxx/*.cpp")
+file(GLOB NANOFI_SOURCES  "src/api/*.cpp" "src/core/*.c*" "src/cxx/*.cpp" "src/sitetosite/*.c*")
```
phrocker: Glad to see this!

nanofi/include/sitetosite/CRawSocketProtocol.h
arpadboda: Done, made it consistent.

nanofi/examples/transmit_payload.c
```diff
@@ -0,0 +1,67 @@
...
+
+  set_log_level(info);
+
+  char *host = argv[1];
+  char *port_str = argv[2];
+  char *nifi_port_str = argv[3];
+
+  uint16_t port_num = atoi(port_str);
+
+  struct CRawSiteToSiteClient * client = createClient(host, port_num, nifi_port_str);
```
arpadboda: Don't think I can make it any more comfortable :)

nanofi/include/sitetosite/CPeer.h
phrocker: I think the update you made to the API calls per the example alleviate any issue so no need to clarify.

nanofi/include/sitetosite/CRawSocketProtocol.h
```diff
@@ -0,0 +1,219 @@
+/**
+ * @file RawSiteToSiteClient.h
```
phrocker: Not a big deal but if you have another commit might want to resolve some of these.

nanofi/src/sitetosite/CRawSocketProtocol.c
```diff
@@ -0,0 +1,1026 @@
...
+
+  RespondCode code;
+
+  ret = readResponse(client, &code);
+
+  if (ret <= 0) {
+    return -1;
+  }
+
+  RespondCodeContext *resCode = getRespondCodeContext(code);
```
phrocker: This function returns null, as do the other instances but no checks, only dereferences.

nanofi/src/sitetosite/CRawSocketProtocol.c
```diff
@@ -0,0 +1,1026 @@
...
+
+  HASH_ITER(hh, properties, current, tmp) {
+    if(writeUTF(current->name, strlen(current->name), False, client->_peer->_stream) <= 0) {
+      return -1;
+    }
+    if(writeUTF(current->value, strlen(current->value), False, client->_peer->_stream) <= 0) {
+      return -1;
+    }
+    //client->logger_->log_debug("Site2Site Protocol Send handshake properties %s %s", current->name, current->value);
+    HASH_DEL(properties, current);
```
phrocker: In the cases where return -1 is used, no free is called. Is that going to cause a memory leak?

nanofi/src/sitetosite/CRawSocketProtocol.c
```diff
@@ -0,0 +1,1026 @@
...
+
+    if (transaction->current_transfers_ > 0) {
+      ret = writeResponse(client, CONTINUE_TRANSACTION, "CONTINUE_TRANSACTION");
+      if (ret <= 0) {
+        return -1;
+      }
+    }
+    // start to read the packet
+    uint32_t numAttributes = packet->_attributes->size;
+    ret = write_uint32t(transaction, numAttributes);
```
phrocker: Is write_uint32t a variation of write_uint32_t?

phrocker:
I'm gonna run some tests with this locally now. I see a test failure but assume that's could be old.

nanofi/src/sitetosite/CRawSocketProtocol.c
arpadboda: Short answer: yes.
Long answer: I needed dedicated write function for transaction to maintain crc checksum.
However there is no overload in C, so the function to write to transaction has to have a different name than the one that's used to write to streams.

nanofi/src/sitetosite/CRawSocketProtocol.c
arpadboda: Yes, it would. Fixed, thanks for spotting!

nanofi/src/sitetosite/CRawSocketProtocol.c
arpadboda: Thanks, added checks.

nanofi/include/sitetosite/CRawSocketProtocol.h
arpadboda: Done, thanks

PR #499, opened 2019-03-05
MINIFICPP-748 - Nanofi: add custom C processor example

arpadboda:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/nanofi/examples/hash_file.c
+++ b/nanofi/examples/hash_file.c
@@ -1,0 +1,121 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+    *
+    *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+#ifdef OPENSSL_SUPPORT
+
+#include "api/nanofi.h"
+#include <openssl/md5.h>
+#include <unistd.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <sys/stat.h>
+
+void custom_processor_logic(processor_session * ps, processor_context * ctx) {
+  flow_file_record * ffr = get(ps, ctx);
+  if(ffr == NULL) {
+    return;
+  }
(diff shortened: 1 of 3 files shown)


nanofi/examples/hash_file.c
```diff
@@ -0,0 +1,121 @@
...
+  free_flowfile(ffr);
+
+  attribute attr;
+  attr.key = checksum_attr_name;
+  attr.value_size = 0;
+  get_attribute(new_ff, &attr);
+
+  char md5value[50];
+
+  snprintf(md5value, attr.value_size, "%s", (const char*)attr.value+1);
```
arpadboda: As a user I find it pretty painful to use.
@phrocker I think we should consider changing values to "const char *"
Not sure what's the right timing for that release-wise. 
If I remember well we already have a Jira for that.

nanofi/src/api/nanofi.cpp
```diff
@@ -371,9 +371,18 @@ int8_t remove_attribute(flow_file_record *ff, const char *key) {
...
   auto content_repo = static_cast<std::shared_ptr<minifi::core::ContentRepository>*>(ff->crp);
-  std::shared_ptr<minifi::ResourceClaim> claim = std::make_shared<minifi::ResourceClaim>(ff->contentLocation, *content_repo);
-  auto stream = (*content_repo)->read(claim);
-  return stream->read(target, size);
+  if(ff->crp && (*content_repo)) {
+    std::shared_ptr<minifi::ResourceClaim> claim = std::make_shared<minifi::ResourceClaim>(ff->contentLocation,
+                                                                                           *content_repo);
+    auto stream = (*content_repo)->read(claim);
+    return stream->read(target, size);
+  } else {
```
arpadboda: The problem here is that ff->contentlocation is used for two different purpose:
-A simple filename to show where the content of the flowfile in the filesystem is
-An internal string that represent the location of the file within the content repo

I think this should be separated.
@phrocker do you have any longer term plan to resolve this in nanofi?

nanofi/examples/hash_file.c
phrocker: Totally agree -- and no objection from my side since it's an internal struct and assignment likely wouldn't cause any impact to currently existing code. 

You're right though, I do think there may be a JIRA for it.

nanofi/src/api/nanofi.cpp
phrocker: Imagine if there were no c++ involved, and as a result, no content repo, would that change your concern?

arpadboda: I wasn't aware of the nanofi related visions here.
In case we don't plan to support content repo in Nanofi, just have flow files store content in a simple file, that's fine. Although I would expect people to in IoT world to use nanofi to trasmit data without any disk i/o operations. 
As this struct is most probably going to change anyway in the near future, it's fine as is.

nanofi/examples/hash_file.c
arpadboda: Just found this: https://issues.apache.org/jira/browse/MINIFICPP-646
This might fit. 

However my question was mostly about timing: this is most probably going to be a breaking change in the API and 0.6.0 is to be released soon, I wonder if we try to add it or do it later in 0.7.0 or 1.0? The later we do it the more we break.

PR #511, opened 2019-03-15
Minificpp 764: Add docker builds for different systems.

phrocker:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ MiNiFi is a child project effort of Apache NiFi.  This repository is for a nativ
 - [Getting Started](#getting-started)
   - [System Requirements](#system-requirements)
   - [Bootstrapping](#bootstrapping)
+  - [Building For Other Distros](#building-for-other-distros)
   - [Cleaning](#cleaning)
   - [Configuring](#configuring)
   - [Running](#running)
@@ -457,6 +458,20 @@ Built target docker
 ~/Development/code/apache/nifi-minifi-cpp/build
 $ make docker-verify
 ```
+
+### Building For Other Distros
+If you have docker installed on your machine you can build for CentOS 7, Fedoera 29, Ubuntu 16, Ubuntu 18, and Debian 9 via our make docker commands. The following table
+provides the command to build your distro and the output file in your build directoiry. Since the versions are limited ( except for Ubuntu ) we output the archive based on the distros name.  
+
+
+| Distro         | command           | Output File  |
+| ------------- |:-------------| :-----|
+| CentOS 7  | make centos | nifi-minifi-cpp-centos-$VERSION-bin.tar.gz
+| Debian 9  | make debian | nifi-minifi-cpp-debian-$VERSION-bin.tar.gz
+| Fedora 29  | make fedora | nifi-minifi-cpp-fedora-$VERSION-bin.tar.gz
+| Ubuntu 16  | make u16 | nifi-minifi-cpp-xenial-$VERSION-bin.tar.gz
+| Ubuntu 16  | make u18 | nifi-minifi-cpp-bionic-$VERSION-bin.tar.gz
+
 
 ### Cleaning
 Remove the build directory created above.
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -458,6 +459,20 @@ Built target docker
 $ make docker-verify
 ```
 
+### Building For Other Distros
+If you have docker installed on your machine you can build for CentOS 7, Fedoera 29, Ubuntu 16, Ubuntu 18, and Debian 9 via our make docker commands. The following table
+provides the command to build your distro and the output file in your build directoiry. Since the versions are limited ( except for Ubuntu ) we output the archive based on the distros name.
```
apiri: minor: directory typo

cmake/DockerConfig.cmake
```diff
@@ -20,6 +20,35 @@ add_custom_target(
     docker
     COMMAND ${CMAKE_SOURCE_DIR}/docker/DockerBuild.sh 1000 1000 ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH} minificppsource ${CMAKE_SOURCE_DIR}
     WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/docker/)
+    
+
+add_custom_target(
```
apiri: Do you think it might be better to just have these as their own Makefile?  If we did that, folks could get all these builds with just Docker whereas they would also need to go through the cmake'ing and need its dependencies when it's just formulating the targets.

docker/ContainerBuild.sh
```diff
@@ -0,0 +1,63 @@
...
+                             --build-arg GID=$GID_ARG \
+                             --build-arg MINIFI_VERSION=$MINIFI_VERSION \
+                             --build-arg MINIFI_SOURCE_CODE=$MINIFI_SOURCE_CODE \
+                             --build-arg DUMP_LOCATION=$DUMP_LOCATION \
+                             -t \
+                             apacheminificpp:${DISTRO_NAME}-$MINIFI_VERSION ."
+echo "Docker Command: '$DOCKER_COMMAND'"
+${DOCKER_COMMAND}
+
+docker run --rm --entrypoint cat apacheminificpp:${DISTRO_NAME}-$MINIFI_VERSION /opt/minifi/build/nifi-minifi-cpp-$MINIFI_VERSION-bin.tar.gz > ${DUMP_LOCATION}/nifi-minifi-cpp-${DISTRO_NAME}-$MINIFI_VERSION-bin.tar.gz
```
apiri: Might be handy to allow an optional maven repository to be specified and then we provide that here as a volume when we run.  Would save the need to download all the artifacts repeatedly when doing the JNI bits.

cmake/DockerConfig.cmake
phrocker: That was a consideration, but then make all in the build process would not be able to make everything. We'd have to pull up a parent Make to launch cmake produced targets, which seems less than ideal. Further, even not withstanding these considerations, the users would have two places to call make as a result.

docker/centos/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+# First stage: the build environment
+# Edge required for rocksdb
+FROM centos:latest AS builder
```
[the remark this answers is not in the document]
phrocker: yes agreed.

cmake/DockerConfig.cmake
apiri: sounds reasonable

phrocker:
@apiri My preference is to do that in a follow on. It's a really good idea, but I think building without JNI and with an attached repo can be a follow on.

apiri:
@phrocker Sounds good. Will take care of the typos and specifying 7 for CentOS on merge.  Good bit of functionality!

docker/ContainerBuild.sh
apiri: https://issues.apache.org/jira/browse/MINIFICPP-765

PR #515, opened 2019-03-17
MINIFICPP-646 - Re-evaluate passing attributes. Potentially deprecate…

arpadboda:
… attribute usage functions

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/nanofi/examples/transmit_payload.c
+++ b/nanofi/examples/transmit_payload.c
@@ -47,9 +47,7 @@ int main(int argc, char **argv) {
   attribute attribute1;
 
   attribute1.key = "some_key";
-  const char * attr_value = "some value";
-  attribute1.value = (void *)attr_value;
-  attribute1.value_size = strlen(attr_value);
+  attribute1.value = "some value";
 
   attribute_set as;
   as.size = 1;
--- a/nanofi/include/api/nanofi.h
+++ b/nanofi/include/api/nanofi.h
@@ -321,7 +321,16 @@ void free_flowfile(flow_file_record* ff);
  * @size size size of the data pointed by "value"
  * @return 0 in case of success, -1 otherwise (already existed)
  **/
-int8_t add_attribute(flow_file_record*, const char *key, void *value, size_t size);
+int8_t add_attribute(flow_file_record*, const char *key, const char *value, size_t size);
+
+/**
+ * Adds an attribute, fails in case there is already an attribute with the given key.
+ * @param ff flow file
+ * @param key name of attribute
+ * @param value location of value - expected to be null-terminated string
+ * @return 0 in case of success, -1 otherwise (already existed)
+ **/
+int8_t add_attribute_str(flow_file_record*, const char *key, const char *value);
 
 /**
  * Updates an attribute (adds if it hasn't existed before)
@@ -330,7 +339,15 @@ int8_t add_attribute(flow_file_record*, const char *key, void *value, size_t siz
  * @param value location of value
  * @size size size of the data pointed by "value"
  **/
-void update_attribute(flow_file_record* ff, const char *key, void *value, size_t size);
+void update_attribute(flow_file_record* ff, const char *key, const char *value, size_t size);
+
+/**
+ * Updates an attribute (adds if it hasn't existed before)
+ * @param ff flow file
+ * @param key name of attribute
+ * @param value location of value - expected to be null-terminated string
+ **/
+void update_attribute_str(flow_file_record* ff, const char *key, const char *value);
 
 /**
  * Get the value of an attribute. Value and value size are written to parameter "caller_attribute"
--- a/nanofi/include/core/cstructs.h
+++ b/nanofi/include/core/cstructs.h
@@ -89,8 +89,7 @@ typedef struct processor_context processor_context;
 
 typedef struct {
   const char *key;
-  void *value;
-  size_t value_size;
+  const char *value;
 } attribute;
 
 typedef struct {
--- a/nanofi/src/api/nanofi.cpp
+++ b/nanofi/src/api/nanofi.cpp
@@ -289,12 +289,17 @@ void free_flowfile(flow_file_record *ff) {
  * @param size size of value
  * @return 0 or -1 based on whether the attributed existed previously (-1) or not (0)
  */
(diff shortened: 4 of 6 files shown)


arpadboda:
This is for 1.0, just coded quickly while I remember what I wanted to do. Can stay open for a while.

nanofi/include/core/cstructs.h
```diff
@@ -89,8 +89,7 @@ typedef struct processor_context processor_context;
 
 typedef struct {
   const char *key;
-  void *value;
-  size_t value_size;
+  const char *value;
```
phrocker: I would probably leave the size because a) you don't know if the value is null terminated ( it may not be ) -- and b) as we move away from C++ we will likely manage the memory and update_attribute at that point could/may be a realloc.

arpadboda: My idea here was to make sure that whatever we store is null-terminated.
We provide API to add non-terminated str with length, (something like strncpy), but inside store null-terminated things.

Two reasons behind:
-no need to store len in internal structures
-whatever we return is null-terminated, which makes it easy to copy/log

In my opinion this leads to a more compact  and cleaner API, but I can be convinced about the opposite way, too.

phrocker: What would convince you?

bakaid:
@arpadboda 

I agree that attributes should be owned by the flowfile structure, otherwise it would be very error-prone to handle lifetimes. It is more than worth the small overhead.
Because of this, we must provide setters.
With getters we have a choice: either we can make the internal storage of the attributes part of the API, and we don't have to provide getters, or we make the attribute structure opaque and provide getters for the attributes (which can either copy the values, or return a const pointer to them, with the proper lifetime warning part of the specification of the getter).
This choice does not influence how we store the attributes, but it influences whether we can change how we store them without breaking the API.
From a setter standpoint I see the merits of having a buffer + size setter along with the C string setter.
From a getter standpoint I think it is easier to use the API if it returns a C string.
Therefore my preference would be to:
 - make the structure opaque
 - create getters
 - make getters return a C string
 - because we can now modify the internal struct it does not really matter how we store the data internally, but because we are returning a C string in the getter, if there are no other considerations it would be worth storing it that way for the time being.

nanofi/include/api/nanofi.h
```diff
@@ -321,7 +321,16 @@ void free_flowfile(flow_file_record* ff);
...
+int8_t add_attribute(flow_file_record*, const char *key, const char *value, size_t size);
+
+/**
+ * Adds an attribute, fails in case there is already an attribute with the given key.
+ * @param ff flow file
+ * @param key name of attribute
+ * @param value location of value - expected to be null-terminated string
+ * @return 0 in case of success, -1 otherwise (already existed)
+ **/
+int8_t add_attribute_str(flow_file_record*, const char *key, const char *value);
```
phrocker: So this functions as purely a convenience function?

phrocker:
@arpadboda Value needn't be null terminated, especially given the sink and source where nanofi may exist; however, I don't actually have a strong opinion on that for this PR -- especially if @bakaid and @msharee9 agree with that approach. NiFi is not the penultimate input/output of NanoFi code -- but as we evolve we can live in the confines of that decision for attributes.

Think of the input/output of nanofi edge controller units as being site to site, kafka, mqtt, pipe, and file ( and maybe redis eventually ). In any of these cases attributes can be of an intermediate form. That means that attributes could carry binary data ( hence why they were initially not typed and instead a bag of keys and values  -- I think this is how redis stores them internally ). I think you can draw a line in the sand that attributes are null terminated and require that another mechanism be built. This may alleviate your concerns about having non textual data in the attributes -- so then we can simply create a follow on to store intermediate data that are attributes of the data that are binary. 

The initial API facilitated C strings that were null terminated and arbitrary binary data. If you feel separating these concerns is important we can make a follow on ticket for the release that references 646. I can certainly see the value in isolating those APIs.

I don't have a super strong opinion on how you structure this "struct" -- as long as we have a way to ultimately carry attributes that are not null terminated. if you think it helps to add API calls for something like "add_binary_attribute", then I'm okay with that.

I don't really foresee a great deal of cost with ownership as you present it. Since we own that data structure and if it were opaque as @bakaid suggests, then we can realloc more safely. 

All code that uses the variants of the API defined in the last release should compile in this minor release. How everything is represented internally is up to you.

arpadboda:
@phrocker : what you say in aspect of Redis for eg. makes sense, however I think that's already messed up. 
Attributes are handled as UTF strings on S2S, so when transferring them, we handle them as strings, which means cutting at the first 0 byte. 
Even if we fix that I'm unsure whether we would achieve the behaviour we want or not given NiFi handles attributes as strings, too. 

To deal with binary formats I think I would prefer to offer base64 encoding for attribute settings and keep the internal representation (and network representation) as string.

phrocker:
@arpadboda Base64 encoding the underlying data is fine, I think, as long as the API allows them to get/set these as binary data. I think that's a good idea.

nanofi/include/core/cstructs.h
szaszm: > We provide API to add non-terminated str with length, (something like strncpy), but inside store null-terminated things.

We can not provide non-null-terminated API for null-terminated storage, because the former allows for null bytes in the payload while the latter does not.
+1 for keeping `void*` + `size_t` if the attribute value can ever contain a null byte as part of the payload, -1 if it is always a string that can be null-terminated.

nanofi/examples/transmit_payload.c
```diff
@@ -47,9 +47,7 @@ int main(int argc, char **argv) {
   attribute attribute1;
 
   attribute1.key = "some_key";
-  const char * attr_value = "some value";
-  attribute1.value = (void *)attr_value;
-  attribute1.value_size = strlen(attr_value);
+  attribute1.value = "some value";
```
szaszm: I've expected `attribute` to be an owning container of `key`/`value`. We really need to clarify ownership semantics and enforce them through the API.
I support the opaque struct proposal and would provide copying and optionally adopting setters for the struct.

szaszm:
Closing this because of inactivity

PR #518, opened 2019-03-20
MINIFICPP-781 - Fix documentation typos found in 0.6.0

arpadboda:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,5 +1,5 @@
 Apache NiFi MiNiFi
-Copyright 2017 The Apache Software Foundation
+Copyright 2017-2019 The Apache Software Foundation
 
 This product includes software developed at
 The Apache Software Foundation (http://www.apache.org/).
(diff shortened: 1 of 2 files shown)


NOTICE
```diff
@@ -1,5 +1,5 @@
 Apache NiFi MiNiFi
-Copyright 2017 The Apache Software Foundation
+Copyright 2019 The Apache Software Foundation
```
phrocker: This is a copyright notice, so that's the year of the first publication. Section 4.3 of ALV2 doesn't stipulate that a date is needed; however, the original notice was from the original notice. Typically copyrights are good for the lifetime of author, so a range doesn't particularly make sense in any scenario for copyrights. Does ASF provide any additional information on this @apiri ?

apiri: The convention I've seen is that it spans the range of the project. I'll have to dig around a bit more to see if I can find the actual basis.

phrocker: Thanks. If the date of first publication is fine as per copy right law, I imagine a range is fine too, in which case I would prefer a singular date to avoid giving the impression that we "haven't updated"

apiri: Looks like the policy is the range of current and prior versions as per https://www.apache.org/legal/src-headers.html#notice.

arpadboda: Thanks, updated according to that!

PR #527, opened 2019-03-27
MINIFICPP-672 - Bootstrap agent through Python

arpadboda:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -1,0 +1,237 @@
+#!/usr/bin/python
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+from __future__ import print_function
+
+MINIFI_SUBFOLDER = '/nifi/nifi-minifi-cpp/'
+APACHE_CLOSER_REPO_JSON_URL = 'https://www.apache.org/dyn/closer.cgi?as_json=1&path=/nifi/nifi-minifi-cpp'
+APACHE_MIRROR_LIST = "http://www.apache.org/mirrors/"
+
+import argparse
+import sys
+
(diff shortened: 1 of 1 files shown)


arpadboda:
No integration pushed yet, it's only a "tool" to find nearest mirrror and download a binary matching your system.

bootstrap.py
```diff
@@ -0,0 +1,298 @@
...
+
+try:
+  import distro
+
+  distro_available = True
+except:
+  distro_available = install_package("distro")
+
+try:
+  from bs4 import BeautifulSoup
```
arpadboda: I'm not really happy with this, as bs4 installation requires pip and I'm not sure we can assume that our users have it.
I was wondering on adding bs4 as a third party source code, which is not nice, but significantly decreases the dependencies the script have. The rest of the libs used are either available in Python by default or not mandatory.

@apiri , @phrocker what do you think?

phrocker: If you visit a link like this: https://www.apache.org/dyn/closer.cgi?as_json=1&path=/nifi/nifi-minifi-cpp/

You will get the preferred mirror. You can then go to that mirror and list the directories: http://mirrors.sonic.net/apache/nifi/nifi-minifi-cpp/0.6.0 for example. You can probably get the latest release from git and looking at the latest release number if you don't want to introduce a new dependency.

arpadboda: The directory listing itself also requires html parsing in case the mirror is http.

phrocker: You needn't perform a directory listing since you are able to get verification and a preferred mirror for the individual files. If it helps we can standardize the release documents to allow us to have specific release files to go to.

arpadboda: That's a fair point, however still implies that the script contains hardcoded information about:
-operating systems and versions we provide release for (centos, xenial, bionic, etc). Or do we have information available somewhere about this?
-different releases for a given platform. (w/o JNI for eg.)

And, as you highlighted above, we need to standardize the name of binaries. 

In my opinion that's a bit too much hardcoded information and restriction, but I'm open if we can somehow resolve the above two.

phrocker: I'm afraid I don't understand your last sentiment. Let's take a step back and look at the spirit of the ticket. Interrogating some service ( should be non specific ) to get a specific arch and operating system. We only support a small set of operating systems at the moment. We'd send this file over to a machine, let it take over by downloading tagged version, or installing the dependency. With that said if it can be done that way, and without installing dependencies on the user system that aren't bootstrap specific, then I think we're golden.

arpadboda: Found a trick using the link you provided in your comment before, we don't need to know the version to find the nearest ftp, which is great news. 
I modified the script, no it only selects FTP mirror. Which means no html parsing is needed any more, so I could remove that part completely. 
Other than that:
-Added an argument to specify a given version (default is latest)
-Made the script compatible with Python 2.7 as well. We should be good running it on a smart toaster as well. :)

phrocker: cool I'm gonna try this out, sorry for the delay

PR #533, opened 2019-04-08
MINIFICPP-726 - Enhance ExtractText to have more feature parity with …

arpadboda:
…the Java impl

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/processors/ExtractText.h
+++ b/libminifi/include/processors/ExtractText.h
@@ -73,7 +73,7 @@ public:
 
     class ReadCallback : public InputStreamCallback {
     public:
-        ReadCallback(std::shared_ptr<core::FlowFile> flowFile, core::ProcessContext *ct);
+        ReadCallback(std::shared_ptr<core::FlowFile> flowFile, core::ProcessContext *ct, std::shared_ptr<logging::Logger> lgr);
         ~ReadCallback() {}
         int64_t process(std::shared_ptr<io::BaseStream> stream);
 
@@ -81,6 +81,7 @@ public:
         std::shared_ptr<core::FlowFile> flowFile_;
         core::ProcessContext *ctx_;
         std::vector<uint8_t> buffer_;
+        std::shared_ptr<logging::Logger> logger_;
     };
 
 protected:
--- a/libminifi/include/processors/ExtractText.h
+++ b/libminifi/include/processors/ExtractText.h
@@ -50,15 +50,26 @@ public:
     //! Supported Properties
     static core::Property Attribute;
     static core::Property SizeLimit;
+
+    static core::Property RegexMode;
+    static core::Property IgnoreCaptureGroupZero;
+    static core::Property InsensitiveMatch;
+    static core::Property MaxCaptureGroupLen;
+    static core::Property EnableRepeatingCaptureGroup;
+
     //! Supported Relationships
     static core::Relationship Success;
     //! Default maximum bytes to read into an attribute
     static constexpr int DEFAULT_SIZE_LIMIT = 2 * 1024 * 1024;
 
     //! OnTrigger method, implemented by NiFi ExtractText
-    void onTrigger(core::ProcessContext *context, core::ProcessSession *session) override;
+    void onTrigger(core::ProcessContext *context, core::ProcessSession *session);
     //! Initialize, over write by NiFi ExtractText
-    void initialize(void) override;
+    void initialize(void);
+
+    virtual bool supportsDynamicProperties() {
+      return true;
+    };
 
     class ReadCallback : public InputStreamCallback {
     public:
--- a/libminifi/src/processors/ExtractText.cpp
+++ b/libminifi/src/processors/ExtractText.cpp
@@ -31,6 +31,12 @@
 #include "core/ProcessSession.h"
 #include "core/FlowFile.h"
 
+#if !defined(_WIN32)
+#if __cplusplus <= 201103L
+#include <regex.h>
+#endif
+#endif
+
 namespace org {
 namespace apache {
 namespace nifi {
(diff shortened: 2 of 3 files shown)


libminifi/src/processors/ExtractText.cpp
```diff
@@ -104,7 +143,64 @@ int64_t ExtractText::ReadCallback::process(std::shared_ptr<io::BaseStream> strea
...
+        }
+        if(!repeatingcapture) {
+          break;
+        }
+        workStr = matches.suffix();
+      }
+    }
+    for(const auto& kv : regexAttributes) {
+      flowFile_->setAttribute(kv.first, kv.second);
+    }
```
arpadboda: The only thing missing here is routing to "unmatched" in case there is no match.
However I think that would be a breaking change, so I would prefer to do that in scope of a follow-up scheduled for 1.0.
This change can be part of the next minor release. 
@apiri do you agree or do you have different opinion?

libminifi/include/processors/ExtractText.h
```diff
@@ -50,15 +50,26 @@ class ExtractText : public core::Processor {
...
     //! Supported Relationships
     static core::Relationship Success;
     //! Default maximum bytes to read into an attribute
     static constexpr int DEFAULT_SIZE_LIMIT = 2 * 1024 * 1024;
 
     //! OnTrigger method, implemented by NiFi ExtractText
-    void onTrigger(core::ProcessContext *context, core::ProcessSession *session) override;
+    void onTrigger(core::ProcessContext *context, core::ProcessSession *session);
     //! Initialize, over write by NiFi ExtractText
-    void initialize(void) override;
```
arpadboda: @phrocker : just found that I accidentally left these here when I fixed https://issues.apache.org/jira/browse/MINIFICPP-684 . It's part of core, It was released in scope of 0.6.0 and it seems we received no complains. Do you think this fact can be used to justify https://issues.apache.org/jira/browse/MINIFICPP-29 ? I'm happy to do further verification in this area, this just made me guess that override should be fine.

libminifi/src/processors/ExtractText.cpp
```diff
@@ -21,6 +21,7 @@
 #include <string>
 #include <memory>
 #include <set>
+#include <regex>
```
phrocker: I don't think this is supported for all versions we allow. Should be easy to check on u14 and u16

libminifi/src/processors/ExtractText.cpp
```diff
@@ -74,11 +111,13 @@ void ExtractText::onTrigger(core::ProcessContext *context, core::ProcessSession
 int64_t ExtractText::ReadCallback::process(std::shared_ptr<io::BaseStream> stream) {
   int64_t ret = 0;
   uint64_t read_size = 0;
+  bool regex_mode = RegexMode.getDefaultValue();
```
phrocker: Why assign to the default value, just because you want to assign it to something?

libminifi/include/processors/ExtractText.h
```diff
@@ -50,15 +50,26 @@ class ExtractText : public core::Processor {
...
+
     //! Supported Relationships
     static core::Relationship Success;
     //! Default maximum bytes to read into an attribute
     static constexpr int DEFAULT_SIZE_LIMIT = 2 * 1024 * 1024;
 
     //! OnTrigger method, implemented by NiFi ExtractText
-    void onTrigger(core::ProcessContext *context, core::ProcessSession *session) override;
+    void onTrigger(core::ProcessContext *context, core::ProcessSession *session);
     //! Initialize, over write by NiFi ExtractText
```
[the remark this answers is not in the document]
arpadboda: Fair point, although I would prefer removing such comments and use the standard way to indicate these.
Further details in: https://issues.apache.org/jira/browse/MINIFICPP-29
Most probably going to be scoped for 1.0

libminifi/src/processors/ExtractText.cpp
arpadboda: Because getProperty call doesn't assign value to the reference in case the property is not found. 
Yes, it should always be found as these properties have default values, but I prefer not to leave such traps in the  code. 

I was wondering on introducing a function call with a fallback value in signature, like:
```
T TryGetProperty(const std::string name, const T& fallBackValue);
```

Which would make such codes a bit nicer.

libminifi/src/processors/ExtractText.cpp
arpadboda: Regex support was added in gcc4.9, u16 uses 5.3, so that's more than fine, u14 is although uses 4.8, so yes, it's not gonna work. 
The question here is urgency: if we need it out pretty soon, I naturally do it. 
In case we don't, u14 is EOL at the end of this month, so I don't think we are going to support it for long, most probably not with 1.0. In this case it can wait. 
I'm happy with both options, what's your opinion?

phrocker: U14 and U16 are just examples. In the case of U14, even if it is EOL that is not the only system that uses GCC 4.8. In any case we should honor what is listed in our public Readme

libminifi/src/processors/ExtractText.cpp
phrocker: What trap are you referring to with using default values when properties aren't specified via configuration?

libminifi/src/processors/ExtractText.cpp
arpadboda: Added 4.8 compatible implementation as well - verified both with the unit tests.

libminifi/src/processors/ExtractText.cpp
arpadboda: Maybe the fact I'm a bit afraid of is the scenario when the property is not required and neither have a default value, so getProperty might return no value. 
In this case it can't occur unless someone removes the defaults, so I removed them here.

libminifi/src/processors/ExtractText.cpp
```diff
@@ -30,20 +31,55 @@
 #include "core/ProcessSession.h"
 #include "core/FlowFile.h"
 
+#if (__cplusplus <= 201103L) || !defined(WIN32)
```
phrocker: I saw an appveyor error related to this, but didn't look into it.

arpadboda: Fixed it.
MSVC++ 14 (Visual Studio 2015) claims C++ version to be '98... Some famous quotes of Linus came to my mind when I realised this.

PR #537, opened 2019-04-17
MINIFICPP-804 - C API Python examples are broken

arpadboda:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/python/minifi/__init__.py
+++ b/python/minifi/__init__.py
@@ -60,7 +60,7 @@ class Processor(object):
         self._minifi = minifi
 
     def set_property(self, name, value):
-        self._minifi.set_property( self._proc, name.encode("UTF-8"), value.encode("UTF-8"))
+        return self._minifi.set_property( self._proc, name.encode("UTF-8"), value.encode("UTF-8")) == 0
 
 class PyProcessor(object):
     def __init__(self, minifi, flow):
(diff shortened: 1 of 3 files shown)


python/minifi/__init__.py
```diff
@@ -81,6 +81,9 @@ def get(self, session, context):
     def transfer(self, session, ff, rel):
         self._minifi.transfer(session, self._flow, rel.encode("UTF-8"))
```
phrocker: This looks like a bug. what is _flow? Can we not simply call transfer on the flowfile? transfer_to_relationship is a bit redundant since transfer is intended to account for this.

arpadboda: Flow is the C flow structure (ExecutionPlan), which the C API was trying to get the flowfile from, but it doesn't really make sense in my opinion in case we already have the flowfile referenced here. 
Naming: happy to rename, but I would prefer to have some suffix to avoid mixing transfer (relationship) and transmit (remote)

phrocker: Ah yeah that might not be necessary since we have the session. I think this one should remain transfer, but hopefully the functions for RPG are transmit*. I'll take a closer look to make sure that is the case. thanks

arpadboda: Okay, renamed to transfer.

PR #540, opened 2019-04-19
MINIFICPP-809: Better support python capabilities and recordreader/wr…

phrocker:
…iters

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/pythonprocessors/google/SentimentAnalyzer.py
```diff
@@ -0,0 +1,64 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+  Install the following with pip ( or pip3 )
+  
+  pip install google-cloud-language
```
arpadboda: Import can be added to try-cache blocks, so error messages can be printed to warn the user to install.

libminifi/src/io/FileStream.cpp
```diff
@@ -28,12 +28,15 @@ namespace nifi {
...
 
-FileStream::FileStream(const std::string &path)
+FileStream::FileStream(const std::string &path, bool append)
     : logger_(logging::LoggerFactory<FileStream>::getLogger()),
       path_(path),
       offset_(0) {
   file_stream_ = std::unique_ptr<std::fstream>(new std::fstream());
-  file_stream_->open(path.c_str(), std::fstream::out | std::fstream::binary);
+  if (append)
+    file_stream_->open(path.c_str(), std::fstream::in | std::fstream::out | std::fstream::app | std::fstream::binary);
```
arpadboda: Why do we need "in" for appending? 

To avoid duplication can it be done like:
```
auto flags  = std::fstream::out | std::fstream::binary;
if (append) {
  flags |= std::fstream::in | std::fstream::app
}
file_stream_->open(path.c_str(), flags);
```
?

phrocker: That bit sets the flag to allow output operations to append at the end but there is nothing to suggest we won't be reading. Further, we need to be able to read and seek to get length. We don't want to obviate user's intent.

extensions/pythonprocessors/google/SentimentAnalyzer.py
phrocker: I actually had it catch and attempt install...that didn't work in the confines of where it was executing. Is that something that's generally ill advised?

A warning should be printed that there is a missing dependency. This won't prevent startup, but will be in the log file.

PR #542, opened 2019-04-24
MINIFICPP-799 - Fixed bison install script in MacOS bootstrap

bakaid:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/darwin.sh
+++ b/darwin.sh
@@ -50,12 +50,9 @@ install_bison() {
     fi
   fi
   if [ "$BISON_INSTALLED" = "false" ]; then
-    ## ensure that the toolchain is installed
-    INSTALL_BASE="sudo zypper in -y gcc gcc-c++"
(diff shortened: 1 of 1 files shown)


bakaid:
The install_bison function was copied from suse.sh and tried to use zypper. This has been removed.
Bison version 3.0.4 has failed to compile with the latest XCode toolchain, so it has been updated to 3.0.5 
There are also other issues with this script, but they are not blocking ones. We install curl via brew and then try to force link it to /usr/local, but that will actually fail, as brew refuses to link it, because there is a system-provided curl.
If we would try to install bison via brew, it would also refuse to link it, because it is “keg-only” and may cause other formulae to fail if linked into /usr/local. The way we install bison causes it to be installed to /usr/local, so it could theoretically cause other brew formulae to fail as well.

PR #554, opened 2019-05-15
MINIFICPP-842 - CivetWeb tests should use random port

arpadboda:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/civetweb/processors/ListenHTTP.cpp
+++ b/extensions/civetweb/processors/ListenHTTP.cpp
@@ -27,11 +27,16 @@ namespace minifi {
 namespace processors {
 
 core::Property ListenHTTP::BasePath(
-    core::PropertyBuilder::createProperty("Base Path")->withDescription("Base path for incoming connections")->isRequired(false)->withDefaultValue<std::string>("contentListener")->build());
+    core::PropertyBuilder::createProperty("Base Path")
+        ->withDescription("Base path for incoming connections")
+        ->isRequired(false)
+        ->withDefaultValue<std::string>("contentListener")->build());
 
 core::Property ListenHTTP::Port(
-    core::PropertyBuilder::createProperty("Listening Port")->withDescription("The Port to listen on for incoming connections")->isRequired(true)->withDefaultValue<int>(
-        80, core::StandardValidators::PORT_VALIDATOR())->build());
+    core::PropertyBuilder::createProperty("Listening Port")
+        ->withDescription("The Port to listen on for incoming connections. 0 means port is going to be selected randomly.")
+        ->isRequired(true)
+        ->withDefaultValue<int>(80, core::StandardValidators::LISTEN_PORT_VALIDATOR())->build());
 
 core::Property ListenHTTP::AuthorizedDNPattern("Authorized DN Pattern", "A Regular Expression to apply against the Distinguished Name of incoming"
                                                " connections. If the Pattern does not match the DN, the connection will be refused.",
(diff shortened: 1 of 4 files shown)


extensions/civetweb/processors/ListenHTTP.cpp
```diff
@@ -27,11 +27,22 @@ namespace minifi {
 namespace processors {
 
 core::Property ListenHTTP::BasePath(
-    core::PropertyBuilder::createProperty("Base Path")->withDescription("Base path for incoming connections")->isRequired(false)->withDefaultValue<std::string>("contentListener")->build());
+    core::PropertyBuilder::createProperty("Base Path")->
```
phrocker: I know we don't quite yet lint extensions ( this will happen soon ) but I'm pretty sure this will need //NOLINT if you intend to keep it this way.

extensions/civetweb/processors/ListenHTTP.cpp
```diff
@@ -27,11 +27,22 @@ namespace minifi {
...
-    core::PropertyBuilder::createProperty("Listening Port")->withDescription("The Port to listen on for incoming connections")->isRequired(true)->withDefaultValue<int>(
-        80, core::StandardValidators::PORT_VALIDATOR())->build());
+    core::PropertyBuilder::createProperty("Listening Port")->
+    withDescription("The Port to listen on for incoming connections")->
+    isRequired(true)->
+    withDefaultValue<int>(80, core::StandardValidators::PORT_VALIDATOR())->build());
+
+core::Property ListenHTTP::RandomPort(
+    core::PropertyBuilder::createProperty("Random Port")->
+    withDescription("Choose random port to listen - for testing purposes only!")->
```
phrocker: I don't think we should have testing code available to production environments. This will be relayed through the C2 framework. Perhaps eliminating it when not building tests will work.

extensions/civetweb/processors/ListenHTTP.cpp
```diff
@@ -191,6 +211,16 @@ void ListenHTTP::onSchedule(core::ProcessContext *context, core::ProcessSessionF
   server_.reset(new CivetServer(options));
   handler_.reset(new Handler(basePath, context, sessionFactory, std::move(authDNPattern), std::move(headersAsAttributesPattern)));
   server_->addHandler(basePath, handler_.get());
+
+  if (randomPort) {
+    const auto& vec = server_->getListeningPorts();
+    if (vec.size() != 1) {
+      logger_->log_error("Random port is set, but there is no listening port! Server most probably failed to start!");
```
phrocker: If this is listening on http and https and a test will use either/both how will it know those random ports?

extensions/civetweb/processors/ListenHTTP.cpp
```diff
@@ -441,6 +471,10 @@ int64_t ListenHTTP::WriteCallback::process(std::shared_ptr<io::BaseStream> strea
   return nlen;
 }
 
+std::string ListenHTTP::getPort() const {
+  return listeningPort;
```
phrocker: Why is a port a string? Is it because ports are appended with s?

extensions/civetweb/processors/ListenHTTP.cpp
```diff
@@ -40,12 +51,18 @@ core::Property ListenHTTP::SSLCertificate("SSL Certificate", "File containing PE
 core::Property ListenHTTP::SSLCertificateAuthority("SSL Certificate Authority", "File containing trusted PEM-formatted certificates", "");
 
 core::Property ListenHTTP::SSLVerifyPeer(
-    core::PropertyBuilder::createProperty("SSL Verify Peer")->withDescription("Whether or not to verify the client's certificate (yes/no)")->isRequired(false)->withAllowableValue<std::string>("yes")
-        ->withAllowableValue("no")->withDefaultValue("no")->build());
+    core::PropertyBuilder::createProperty("SSL Verify Peer")->
+    withDescription("Whether or not to verify the client's certificate (yes/no)")->
```
phrocker: I'm not against this style change at all and the removal of the linted code, but let's move that to a separate PR so it can be tracked better through git.

phrocker: * sorry in the future. leave these as-is

extensions/civetweb/processors/ListenHTTP.cpp
arpadboda: There is no such option. 
The processor currently only supports one port property, which is validated using port validator.
http XOR https depends on whether SSL cert property is provided or not. In case it is, the processor automatically switches to https. 
To make sure I will add a getter for this, too.

extensions/civetweb/processors/ListenHTTP.cpp
arpadboda: Yes.

extensions/civetweb/processors/ListenHTTP.cpp
arpadboda: Not "just" style, changed these to use allowableValue**s** to avoid calling _withAllowableValue_ many times for the same prop.

extensions/civetweb/processors/ListenHTTP.cpp
arpadboda: That's a good point, thanks!
Checked other processors, it should be the opposite way ( -> operator is at the beginning of new lines) to be linter-compatible, so fixed.

extensions/civetweb/processors/ListenHTTP.cpp
arpadboda: Good point. Added an ifdef to only add it to the set of supported properties in case of building with tests.

extensions/civetweb/processors/ListenHTTP.cpp
phrocker: That is a fluent API.

arpadboda:
> Was about to merge and realized that SKIP_TESTS is used as the ifdef. The logic there is that if tests are built include this parameter. This means that RandomPort will be included via the manifest if tests were built. Could you just set listening port to zero via the config of the test, use the function to get the port you added to later config the client? This way no randomport option is needed. The same port config is used.

That was my initial idea as well, although the port validator we have at the moment doesn't accept 0, so that wouldn't work. 
Which makes sense in a way as 0 is not a valid port. 

We could make that a global rule in MiNiFi - port validator allows 0 and that means the processor has to choose a random free port. We might consider this in the future, but it's a breaking change and clearly doesn't fit the scope of this ticket.

PR #553, opened 2019-05-14
MINIFICPP-287: Add property for flow files to log

phrocker:
Thank you for submitting a contribution to Apache NiFi - MiNiFi C++.

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [ ] Is there a JIRA ticket associated with this PR? Is it referenced
     in the commit message?

- [ ] Does your PR title start with MINIFICPP-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.

- [ ] Has your PR been rebased against the latest commit within the target branch (typically master)?

- [ ] Is your initial contribution a single, squashed commit?

### For code changes:
- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?
- [ ] If applicable, have you updated the LICENSE file?
- [ ] If applicable, have you updated the NOTICE file?

### For documentation related changes:
- [ ] Have you ensured that format looks appropriate for the output in which it is rendered?

### Note:
Please ensure that once the PR is submitted, you check travis-ci for build issues and submit an update to your PR as soon as possible.

PR #558, opened 2019-05-16
MINIFICPP-542 - Add PutSFTP processor

bakaid:

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/sftp/processors/PutSFTP.cpp
```diff
@@ -0,0 +1,846 @@
...
+    }
+  }
+
+  session->transfer(flow_file, Success);
+  put_connection_back_to_cache();
+  return true;
+}
+
+void PutSFTP::onTrigger(const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSession> &session) {
+  for (size_t i = 0; i < batch_size_; i++) {
```
arpadboda: Please add an unsigned validator for BatchSize property. 
Moreover 0 should either be forbidden or we should process as many flowfiles as available. 
Doing the latter does have consequences, please check this PR, quite the same topic was discussed:
https://github.com/apache/nifi-minifi-cpp/pull/553

CMakeLists.txt
```diff
@@ -376,6 +379,94 @@ if (CURL_FOUND)
   include_directories("${CURL_INCLUDE_DIRS}")
 endif()
 
+if (ENABLE_ALL OR ENABLE_SFTP)
```
phrocker: Can this be put into a cmake module. We'll slowly move in that direction like libressl did.

extensions/sftp/client/SFTPClient.h
```diff
@@ -0,0 +1,183 @@
...
+    static SFTPClientInitializer initializer;
+    return &initializer;
+  }
+  void initialize() {
+
+  }
+ private:
+  SFTPClientInitializer() {
+    libssh2_init(0);
+    curl_global_init(CURL_GLOBAL_DEFAULT);
```
phrocker: curl_global_init is not thread safe. The singleton pattern appears to model that which is in http-curl, which is lazily initialized.

libminifi/include/utils/file/FileUtils.h
```diff
@@ -386,6 +415,141 @@ class FileUtils {
...
+    if (last_separator == std::string::npos || last_separator < root_pos) {
+      return "";
+    }
+    return path.substr(0, last_separator + 1);
+  }
+
+  /*
+   * Returns the absolute path of the current executable
+   */
+  static std::string get_executable_path() {
```
phrocker: We should be using MINIFI_HOME instead of any system resources like this.

extensions/sftp/processors/PutSFTP.h
```diff
@@ -0,0 +1,183 @@
...
+  static core::Property ProxyPort;
+  static core::Property HttpProxyUsername;
+  static core::Property HttpProxyPassword;
+
+  // Supported Relationships
+  static core::Relationship Success;
+  static core::Relationship Reject;
+  static core::Relationship Failure;
+
+  virtual bool supportsDynamicProperties() override {
```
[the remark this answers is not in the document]
phrocker: In extensions it's okay. It's only in core (libminifi) code where we have any concern.

extensions/sftp/tests/tools/sftp-test-server/src/main/java/InsecureRandom.java
```diff
@@ -0,0 +1,56 @@
...
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.util.Random;
+
+import org.apache.sshd.common.random.AbstractRandom;
+
+public class InsecureRandom extends AbstractRandom {
```
phrocker: May be wrong, but doesn't org.apache.sshd.common.random.SingletonRandomFactory do this already and implement both interface?

libminifi/include/utils/ScopeGuard.h
```diff
@@ -0,0 +1,64 @@
...
+#ifndef __SCOPE_GUARD_H__
+#define __SCOPE_GUARD_H__
+
+#include <functional>
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+
```
phrocker: namespace utils for this file ?

libminifi/include/utils/ScopeGuard.h
```diff
@@ -0,0 +1,64 @@
...
+ public:
+  ScopeGuard(std::function<void()>&& func)
+    : enabled_(true)
+    , func_(std::move(func)) {
+  }
+
+  ~ScopeGuard() {
+    if (enabled_) {
+      try {
+        func_();
```
phrocker: I loved scoped guards when used in C++ but you're calling the closure of c code, have you validated that the close functions on libssh don't trap and shut the agent down if a closure were to happen after some errant transfer ? It seems like those closures would be safe, but want to double check, thanks!

CMakeLists.txt
bakaid: I think so, I'll try to do that.

libminifi/include/utils/file/FileUtils.h
bakaid: This is only used from tests where I need test resources relative to the test executable (the test SFTP server, keys, hosts files, etc.).

extensions/sftp/processors/PutSFTP.cpp
bakaid: Nice catch, thank you!

extensions/sftp/client/SFTPClient.cpp
```diff
@@ -0,0 +1,713 @@
...
+
+bool SFTPClient::putFile(const std::string& path, io::BaseStream& input, bool overwrite, int64_t expected_size /*= -1*/) {
+  int flags = 0;
+  if (overwrite) {
+    flags = LIBSSH2_FXF_WRITE | LIBSSH2_FXF_CREAT | LIBSSH2_FXF_TRUNC;
+  } else {
+    flags = LIBSSH2_FXF_WRITE | LIBSSH2_FXF_CREAT | LIBSSH2_FXF_EXCL;
+  }
+  logger_->log_trace("Opening remote file \"%s\"", path.c_str());
+  LIBSSH2_SFTP_HANDLE *file_handle = libssh2_sftp_open(sftp_session_, path.c_str(), flags, 0644);
```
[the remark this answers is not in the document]
bakaid: There is no remote default, it works pretty much the same way open (http://man7.org/linux/man-pages/man2/open.2.html) does on the local system: if you want to create files you have specify a mode which will be masked with the umask in effect.
0644 is pretty standard when you want to create non-executable files, curl uses that too: https://curl.haxx.se/libcurl/c/CURLOPT_NEW_FILE_PERMS.html

extensions/sftp/client/SFTPClient.cpp
```diff
@@ -0,0 +1,713 @@
...
+                              source_path.length(),
+                              target_path.c_str(),
+                              target_path.length(),
+                              flags) != 0) {
+    auto err = libssh2_sftp_last_error(sftp_session_);
+    if (overwrite && err == LIBSSH2_FX_FILE_ALREADY_EXISTS && !tried_deleting) {
+      /* It couldn't overwrite the file, let's delete it and try again */
+      tried_deleting = true;
+      if (!this->removeFile(target_path)) {
+        return false;
```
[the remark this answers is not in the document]
bakaid: I agree, but I would only do it on a debug level, not error, because in my experience if a system does not support overwriting files with rename, it will consistently refuse it, and the logs would be full of this not-really-error.

CMakeLists.txt
```diff
@@ -426,12 +517,14 @@ if(BOOTSTRAP)
     return()
 endif()
 
+message("-------- BOOTSTRAP BEGIN --------")
```
phrocker: hmm thanks for the reminder. I don't think this process is needed any longer.

cmake/libssh2/sys/FindLibSSH2.cmake
```diff
@@ -0,0 +1,35 @@
...
+find_path(LIBSSH2_INCLUDE_DIR libssh2.h
+        )
+
+find_library(LIBSSH2_LIBRARY NAMES ssh2
+        )
+
+if(LIBSSH2_INCLUDE_DIR)
+    file(STRINGS "${LIBSSH2_INCLUDE_DIR}/libssh2.h" libssh2_version_str REGEX "^#define[\t ]+LIBSSH2_VERSION_NUM[\t ]+0x[0-9][0-9][0-9][0-9][0-9][0-9].*")
+
+    string(REGEX REPLACE "^.*LIBSSH2_VERSION_NUM[\t ]+0x([0-9][0-9]).*$" "\\1" LIBSSH2_VERSION_MAJOR "${libssh2_version_str}")
```
phrocker: this makes me think this is a derived file ( and not just replicating the variables ) -- then will need something in NOTICE

extensions/sftp/CMakeLists.txt
```diff
@@ -0,0 +1,98 @@
...
+		message("Including ${CURL_LIBRARY}")
+		target_link_libraries (minifi-sftp ${CURL_LIBRARY})
+		set_target_properties(minifi-sftp PROPERTIES LINK_FLAGS "${LINK_FLAGS} /WHOLEARCHIVE:${CURL_LIBRARY}")
+	else()
+		target_link_libraries(minifi-sftp ${CURL_LIBRARIES})
+	endif()
+endif ()
+
+# Include UUID
+find_package(UUID REQUIRED)
```
phrocker: A lot of extensions were copy/paste, but i don't think things like uuid are required everywhere since these aren't (currently) releasable targets. In any case you can leave it here unless you have a burning desire to remove them.

thirdparty/libssh2/include/libssh2_publickey.h
```diff
@@ -0,0 +1,119 @@
+/* Copyright (c) 2004-2006, Sara Golemon <sarag@libssh2.org>
```
phrocker: Is this patch from their code base or from elsewhere? If the latter, then it needs to be documented accordingly, otherwise it's covered by the LICENSE block....though I suspect they're actually from a different version?

libminifi/include/utils/file/FileUtils.h
```diff
@@ -196,6 +200,29 @@ class FileUtils {
...
+      return true;
+    }
+    return false;
+  }
+#endif
+
+#ifndef WIN32
+  static bool get_uid_gid(const std::string &path, uint64_t &uid, uint64_t &gid) {
+    struct stat result;
+    if (stat(path.c_str(), &result) == 0) {
```
phrocker: I'm guessing the Windows GUID doesn't work here?

cmake/libssh2/sys/FindLibSSH2.cmake
bakaid: You're right, this was a thing I was going to ask about. This is from curl, and we already mention we bundle curl in the LICENSE file. Should this module be mentioned specifically somewhere?

libminifi/include/utils/file/FileUtils.h
bakaid: I'm not sure what you mean. The st_uid and st_uid fields will always be 0 in the Windows version of stat (https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/stat-functions?view=vs-2019). I may be able to figure the owner out with GetSecurityInfo, but a simple uid-gid is not really applicable on Windows AFAIK.

extensions/sftp/client/SFTPClient.cpp
```diff
@@ -0,0 +1,713 @@
...
+
+bool SFTPClient::rename(const std::string& source_path, const std::string& target_path, bool overwrite) {
+  int flags = 0;
+  if (overwrite) {
+    flags = LIBSSH2_SFTP_RENAME_ATOMIC | LIBSSH2_SFTP_RENAME_NATIVE | LIBSSH2_SFTP_RENAME_OVERWRITE;
+  } else {
+    flags = LIBSSH2_SFTP_RENAME_ATOMIC | LIBSSH2_SFTP_RENAME_NATIVE;
+  }
+  bool tried_deleting = false;
+  while (libssh2_sftp_rename_ex(sftp_session_,
```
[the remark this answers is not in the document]
bakaid: Yeah, a loop seemed more elegant than parameterizing and calling the rename again.

LIBSSH2_ERROR_EAGAIN: we use libssh2 in blocking mode, which means that it shouldn't return with EAGAIN, as evidenced by this: https://github.com/libssh2/libssh2/blob/libssh2-1.8.2/src/session.h#L42

libminifi/include/utils/ScopeGuard.h
bakaid: I love them too.
libssh2_sftp_close should be called on every handle even if there were transfer errors, otherwise resources wouldn't be freed.
I assume you meant the closure question somewhat independently from using ScopeGuards though - if they would trap they would do so even if the function wouldn't be called from ~ScopeGuard.

CMakeLists.txt
bakaid: It took me a while to figure out why things were called twice with different parameterization, that's why I put a separator there.

extensions/sftp/processors/PutSFTP.h
```diff
@@ -0,0 +1,183 @@
...
+        const std::string& conflict_resolution);
+    ~ReadCallback();
+    virtual int64_t process(std::shared_ptr<io::BaseStream> stream) override;
+    bool commit();
+
+   private:
+    std::shared_ptr<logging::Logger> logger_;
+    bool write_succeeded_;
+    const std::string target_path_;
+    std::shared_ptr<utils::SFTPClient> client_;
```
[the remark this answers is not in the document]
bakaid: Yeah, you're completely right. There was a version of the connection cache where shared_ptr was really needed, but I have rewritten it once, and it no longer is.

.travis.yml
```diff
@@ -49,6 +49,7 @@ matrix:
           - libusb-1.0-0-dev
           - libpng12-dev
           - libpcap-dev
+          - libssh2-1-dev
```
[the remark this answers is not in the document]
bakaid: It is done by bootstrap if SFTP is selected and everything works fine, but so is libusb and libpcap, and they're both here. I am not sure why.

libminifi/include/utils/ScopeGuard.h
bakaid: Yes, you are quire right.

extensions/sftp/processors/PutSFTP.cpp
bakaid: 0 is treated as unlimited (uint64 max), in accordance with the NiFi implementation.

extensions/sftp/processors/PutSFTP.h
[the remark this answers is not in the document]
bakaid: Thanks for the heads-up, I will keep this as-is then, and remember not do it in the core.

thirdparty/libssh2/include/libssh2_publickey.h
bakaid: The include directory is copied verbatim from the same version we use in the ExternalProject. I've seen this done with all of our thirdparty dependencies, and I suspected that this is because we don't want to depend on includes only available in the build directory after building the external project. If this is not the case, I will revise this.
The patch file and the copied CMakeLists.txt, which are applied in the PATCH phase are my own modifications (on Windows we don't have a patch command, so the modified CMakeLists.txt is copied over the one in the downloaded sources).

extensions/sftp/client/SFTPClient.cpp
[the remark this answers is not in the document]
bakaid: Added with log_debug.

CMakeLists.txt
bakaid: Done that, added use_bundled_libssh2(), but unlike LibreSSL it must only be called after ZLIB and the SSL implementation (OpenSSL/LibreSSL) has been processed. We're currently calling the function the proper way, but I am not sure if I can programatically protect it from being called at the wrong place.

extensions/sftp/tests/tools/sftp-test-server/src/main/java/InsecureRandom.java
bakaid: It does implement both interfaces, but it is just a wrapper for another random factory, it does not in itself implement random generation (https://github.com/apache/mina-sshd/blob/master/sshd-common/src/main/java/org/apache/sshd/common/random/SingletonRandomFactory.java).
The point of InsecureRandom and InsecureRandomFactory is to make SshServer use java.util.Random, an insecure but fast random implementation.
This was needed because tests magically started to fail on Linux because the SSH server took more and more time to start and then refused to start entirely. The problem then fixed itself after a few minutes. I realized that the problem was that it used /dev/random to seed its internal random engine, which ran out of entropy after a while...
It turns out that is change is not enough, because even though SshServer itself will use this random generator, BouncyCastle will still use its secure one, that's why the server is started with "-Djava.security.egd=file:/dev/./urandom", to make it use /dev/urandom.
I still kept this because I'm not sure what will happen on Windows, and we might need this there.

extensions/sftp/client/SFTPClient.h
bakaid: Neither is libssh2_init thread-safe.
You're right, I have modeled the SFTPClientInitializer on HTTPClientInitializer from http-curl, and I don't see how it would be more or less thread-safe in itself than that (a block-scope static object provides synchronization). It is even initialized in the same way (lazily, both by the respective client and in the constructor of the Processor that uses that client).
However, now I do realize, that globally it became less thread-safe, because if both initializers would be constructed at the same time on different threads, there would be no synchronization between them, and curl_global_init could indeed be called in an unsafe manner.
If this could happen, we need to create global initializers for these third-party libraries. Creating a separate CurlInitializer and LibSSH2Initializer won't solve the problem, however, because even calling curl_global_init and libssh2_init at the same time is unsafe, because they both potentially call OpenSSL's initialization, which is not thread-safe in itself.
To solve this we need to find a way to synchronize all thirdparty global initializations and deinitializations.
What do you think, @phrocker ?

extensions/sftp/CMakeLists.txt
bakaid: Could you elaborate on this please? PutSFTP's contructor has a (default) utils::Identifier uuid parameter which with the delegated Processor constructor is called. Is this unnecessary? What do you mean by uuid only needed in releasable targets?

.travis.yml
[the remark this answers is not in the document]
phrocker: It's vestigial. we need to do some cleanup in travis so we can sort of deal with that another time.

cmake/libssh2/sys/FindLibSSH2.cmake
phrocker: Yes, we follow this http://www.apache.org/licenses/LICENSE-2.0.html#redistribution -- which is not technically required but a derived work should be called out in the notice. I can take care of that on merge, though

extensions/sftp/CMakeLists.txt
phrocker: Sorry that wasn't clear, UUID is transitively included via libminifi -- and at this point the extensions are not releasable -- but I plan on them being so at some point. We can discuss this further in tickets. I don't think a change is needed it was more FYI

bakaid:
@phrocker I do, there are still unresolved (and valid) issues I did not yet fix, but I've added a comment to them asking about how they should be done.

bakaid:
Specifically the thread safety of the global initializers needs to fixed - I am not sure what would the best method be, but I've added comments -, and I would like to run a manual test of stopping and restarting the processor.

extensions/sftp/client/SFTPClient.h
phrocker: @bakaid Thanks for pointing that out. One option is create a varying dependency tree of lazy initializers -- but I think as you point out that will be error prone.

Extensions were intended to be groups of functionality and one other issue is that the bootstrap script will eventually run out of characters. I'd like to get to a point where more of these extensions that rely on the same dependencies are combined. 

Another option here could be combining sftp and http-curl into a curl extension, and using the HTTPCurlLoader functionality to eagerly initialize these components. We can still have SFTP be an included component within that extension, but then all curl based functionality will be included within a single extension with the ability to pick and choose sub-features as needed. Does that make sense? 

There will be more features that use curl, so having http-curl and other curl reliant extensions doesn't make a ton of sense to me. We can still have the same ability to limit features included and also have the byproduct that initialization can be done via the ObjectFactory ( a component of the class loader that is loaded before all processors start ). 

While we certainly could have the case where something "like this" happens again, I'm not sure we can avoid that as that's a consistent issue amongst Java Classloaders, too. What are your thoughts on this? It can be something we do after this is merged, labeling it a blocker to get it done, but I'll leave that up to you since there is a ton of good work here and I don't want to slow this up for such a change unless you are open to do that. 

I'll give it more thought, but those are two options I can think of that would solve this and other issues that will arise by having the split we have.

bakaid: @phrocker On grouping the extensions by dependencies: from our standpoint that could make sense (it could solve our current problem, for example), as long as extensions do not depend on different sets of multiple external dependencies. The sftp extension depends on curl and libssh2, so we could group it under extensions depending on curl, with an extra attribute of depending on libssh2.
But what would happen if in the future we want to create an extension which depends on libssh2 but does not depend on curl? (I don’t think this particular case will happen, but I’m trying to figure out a general architecture.) In that case we will still have to group it under curl-dependent extensions and make it dependent on curl, as we could not create a different initializer for it that calls libssh2_init, because that could potentially be called in parallel with HTTPCurlLoader.
Even worse, If our extension would depend on libssh2 and libfoobar, we would have to group all extensions depending on libfoobar with the extensions depending both on curl and libssh2, as we couldn’t allow having two initializers calling libfoobar_init.
From the standpoint of the user I would imagine (but I could be terribly wrong in this case), that they do not care about the thirdparty dependencies we use to implement processors - they care about the particular processor and the functionality it brings. They don’t want to have all functionality that could be implemented with curl - they want SFTP, FTP or HTTP.
So to recap: grouping similar processors into the same extensions, especially if they have the same dependencies could make sense on a case-to-case basis but I don’t think that grouping all processors depending on a particular external dependency into the same extension solves our problems or brings much to the table (we could solve running out of letters in bootstrap.sh by other means), but brings its own set of problems.

A way to solve this based on your recommendation perhaps would be to eagerly initialize the components in the loader AND call these initialization functions under a global lock (move the lock in ClassLoader::registerResource up a few lines). This way we wouldn’t have to change the architecture drastically, but still wouldn’t have to group all processors depending on curl (and all other transitive dependencies) in the same extension.

I too will think about this more.

cmake/libssh2/sys/FindLibSSH2.cmake
bakaid: I have updated the NOTICE file.

bakaid:
The XCode 10 Travis job timed out, but the tests passed:
[0;32m100% tests passed[0;0m, 0 tests failed[0;0m out of 69

phrocker:
@bakaid that happens quite a bit. good to merge?

bakaid:
@phrocker Yes, it should be good to go, thank you!

bstrp_functions.sh
```diff
@@ -279,7 +280,7 @@ show_supported_features() {
 
 read_feature_options(){
   local choice
-  read -p "Enter choice [ A - T or 1-3 ] " choice
+  read -p "Enter choice [ A - T or 1-4 ] " choice
```
arpadboda: Should be A - V :)

phrocker:
Merged, created https://issues.apache.org/jira/browse/MINIFICPP-879 due to a test failure on U18

bstrp_functions.sh
phrocker: Easy change in https://issues.apache.org/jira/browse/MINIFICPP-880.

PR #560, opened 2019-05-17
MINIFICPP-852: Add details to restful response. CoAP will not include…

phrocker:
… details for the time being

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/http-curl/tests/C2FailedUpdateTest.cpp
+++ b/extensions/http-curl/tests/C2FailedUpdateTest.cpp
@@ -57,8 +57,31 @@ class ConfigHandler : public CivetHandler {
   ConfigHandler() {
     calls_ = 0;
   }
-  bool handlePost(CivetServer *server, struct mg_connection *conn) {
+  virtual bool handlePost(CivetServer *server, struct mg_connection *conn) override {
     calls_++;
+    const struct mg_request_info *req_info = mg_get_request_info(conn);
+    long long remainlen;
+    long long readlen = 0;
+    long long contentlen = req_info->content_length;
+    char buf[1024];
+
+    std::string data;
+    while (readlen < contentlen) {
+      remainlen = contentlen - readlen;
+      if (remainlen > sizeof(buf)) {
+        remainlen = sizeof(buf);
+      }
+      remainlen = mg_read(conn, buf, (size_t) remainlen);
+      if (remainlen <= 0) {
+        break;
+      }
+      readlen += remainlen;
+      data += std::string(buf, remainlen);
+    }
+    if (data.find("operationState") != std::string::npos) {
+      assert(data.find("state\": \"NOT_APPLIED") != std::string::npos);
+    }
+
     if (responses.size() > 0) {
       std::string top_str = responses.back();
       responses.pop_back();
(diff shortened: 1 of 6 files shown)


extensions/http-curl/tests/C2FailedUpdateTest.cpp
```diff
@@ -57,8 +58,30 @@ class ConfigHandler : public CivetHandler {
...
+    long long contentlen = req_info->content_length;
+    char buf[1024];
+
+    std::string data;
+    while (readlen < contentlen) {
+      remainlen = contentlen - readlen;
+      if (remainlen > sizeof(buf)) {
+        remainlen = sizeof(buf);
+      }
+      remainlen = mg_read(conn, buf, (size_t) contentlen);
```
arpadboda: I think this was meant to be:
```
      remainlen = mg_read(conn, buf, (size_t) remainlen);
      if (remainlen <= 0) {
        break;
      }
      readlen += remainlen;
```

libminifi/include/core/logging/LoggingInterceptor.h
```diff
@@ -0,0 +1,117 @@
...
+
+class Interceptor {
+ public:
+  void intercept_error(const std::string &msg) {
+    std::lock_guard<std::mutex> lock(guard_);
+    errors_.emplace_back(msg);
+  }
+
+  const std::vector<std::string> &getErrors() const {
+    return errors_;
```
arpadboda: I'm a bit afraid of this. 
When errors are logged the intercept_error func above will lock the mutex, although whoever calling this function will get a reference to the vector without knowing about the mutex. 
I would prefer lock and copy here.  A match function (to be able to search for specific errors using regex) would be nice, too - that can do the job using the lock and avoid copies.

libminifi/include/core/logging/LoggingInterceptor.h
```diff
@@ -0,0 +1,117 @@
...
+class LoggingInterceptor : public Interceptor {
+
+ public:
+
+  LoggingInterceptor() {
+    LoggingInterceptors::getInstance()->addInterceptor(this);
+  }
+
+  ~LoggingInterceptor() {
+    std::lock_guard<std::mutex> lock(guard_);
```
arpadboda: This is a possible deadlock. 
LoggingInterceptor is being destructed while LoggingInterceptors::intercept_error is being executed. 
Both of these grab's it's own lock before trying to call a member of the other to lock that, too.

libminifi/include/core/logging/LoggingInterceptor.h
phrocker: That's a good catch. This was unintentional

libminifi/include/core/logging/LoggingInterceptor.h
phrocker: Yah that's a bad design, will fix. Though in thinking about it the logging interceptor is not what I want to do. So I will likely remove it entirely.  It should be more purposeful.n

phrocker: Errors should be captured mode purposefully, rather n

PR #563, opened 2019-05-21
WiP: MINIFICPP-839 - Fix memleaks

bakaid:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/ResourceClaim.h
+++ b/libminifi/include/ResourceClaim.h
@@ -54,26 +54,35 @@ class ResourceClaim : public std::enable_shared_from_this<ResourceClaim> {
   /*!
    * Create a new resource claim
    */
-  //explicit ResourceClaim(std::shared_ptr<core::StreamManager<ResourceClaim>> claim_manager, const std::string contentDirectory);
+  //explicit ResourceClaim(std::weak_ptr<core::StreamManager<ResourceClaim>> claim_manager, const std::string contentDirectory);
 
-  explicit ResourceClaim(std::shared_ptr<core::StreamManager<ResourceClaim>> claim_manager);
+  explicit ResourceClaim(std::weak_ptr<core::StreamManager<ResourceClaim>> claim_manager);
 
-  explicit ResourceClaim(const std::string path, std::shared_ptr<core::StreamManager<ResourceClaim>> claim_manager, bool deleted = false);
+  explicit ResourceClaim(const std::string path, std::weak_ptr<core::StreamManager<ResourceClaim>> claim_manager, bool deleted = false);
   // Destructor
   ~ResourceClaim() {
   }
   // increaseFlowFileRecordOwnedCount
   void increaseFlowFileRecordOwnedCount() {
-    claim_manager_->incrementStreamCount(shared_from_this());
+    auto claim_manager_shared = claim_manager_.lock();
+    if (claim_manager_shared) {
+      claim_manager_shared->incrementStreamCount(shared_from_this());
+    }
   }
   // decreaseFlowFileRecordOwenedCount
   void decreaseFlowFileRecordOwnedCount() {
-    claim_manager_->decrementStreamCount(shared_from_this());
-
+    auto claim_manager_shared = claim_manager_.lock();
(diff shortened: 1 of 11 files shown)


libminifi/src/core/repository/VolatileContentRepository.cpp
```diff
@@ -81,10 +81,7 @@ void VolatileContentRepository::start() {
     return;
   if (running_)
     return;
-  thread_ = std::thread(&VolatileContentRepository::run, shared_from_parent<VolatileContentRepository>());
-  thread_.detach();
```
phrocker: What was the point of removing this?

libminifi/include/core/FlowFile.h
```diff
@@ -254,29 +254,29 @@ class FlowFile : public core::Connectable, public ReferenceContainer {
    * Sets the original connection with a shared pointer.
    * @param connection shared connection.
```
phrocker: Hmm, the flowfile should complete its activity. Why does this need to be weak?

libminifi/src/ResourceClaim.cpp
```diff
@@ -41,11 +42,15 @@ void setDefaultDirectory(std::string path) {
   default_directory_path = path;
 }
 
-ResourceClaim::ResourceClaim(std::shared_ptr<core::StreamManager<ResourceClaim>> claim_manager)
```
phrocker: These claims should not be persisted on stops. They should complete, so making them weak seems counterintuitive, though I admit it may be necessary -- why is the claim required to be weak ?

libminifi/src/core/ProcessSession.cpp
```diff
@@ -805,7 +805,7 @@ void ProcessSession::commit() {
         continue;
       }
 
-      connection = std::static_pointer_cast<Connection>(record->getConnection());
+      connection = std::static_pointer_cast<Connection>(record->getConnection().lock());
```
phrocker: Session should be a hard pointer. How did you arrive at this?

libminifi/src/core/Processor.cpp
```diff
@@ -176,7 +176,7 @@ void Processor::removeConnection(std::shared_ptr<Connectable> conn) {
     // Connection is destination to the current processor
     if (_incomingConnections.find(connection) != _incomingConnections.end()) {
       _incomingConnections.erase(connection);
-      connection->setDestination(NULL);
```
phrocker: hmmm, why this change? I would prefer nullptr here. It's easier to read IMO, so I think we should either leave this as-is or use nullptr.

libminifi/include/core/FlowFile.h
phrocker: I presume this is because the processor is weak?

libminifi/include/ResourceClaim.h
```diff
@@ -112,7 +122,7 @@ class ResourceClaim : public std::enable_shared_from_this<ResourceClaim> {
   // Full path to the content
   std::string _contentFullPath;
 
-  std::shared_ptr<core::StreamManager<ResourceClaim>> claim_manager_;
```
phrocker: hmmm why are we creating a non-owning reference here? We could potentially lose data in some routes. Seems worrisome. I'll do some testing to validate.

libminifi/src/core/repository/VolatileContentRepository.cpp
bakaid: As far as I could gather it served no discernible purpose, but because a shared_ptr instance is capture in a detached thread it causes valgrind to issue a warning (as valgrind will evaluate memory leaks at exit()).

libminifi/src/core/Processor.cpp
bakaid: I prefer nullptr too, but setDestination now gets a weak_ptr and weak_ptr has no constructor from nullptr_t (or any pointer for that matter).

szaszm:
Closing this because of inactivity. Some of the leaks have been fixed on main, and I'll make sure to check this as reference when similar efforts are underway.

PR #573, opened 2019-05-29
minificpp-873 Implmented.

am-c-p-p:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/windows-event-log/CollectorInitiatedSubscription.cpp
+++ b/extensions/windows-event-log/CollectorInitiatedSubscription.cpp
@@ -616,7 +616,7 @@ bool CollectorInitiatedSubscription::subscribe(const std::shared_ptr<core::Proce
     return false;
   }
 
-  lastActivityTimestamp_ = GetTickCount();
+  lastActivityTimestamp_ = GetTickCount64();
 
   return true;
 }
@@ -636,10 +636,7 @@ int CollectorInitiatedSubscription::processQueue(const std::shared_ptr<core::Pro
     }
 
     int64_t process(std::shared_ptr<io::BaseStream> stream) {
-      auto len = stream->writeData((uint8_t*)&str_[0], str_.size());
-      if (len < 0)
-        status_ = -1;
-      return len;
+      return stream->writeData((uint8_t*)&str_[0], str_.size());
     }
 
     std::string str_;
--- a/extensions/windows-event-log/CollectorInitiatedSubscription.cpp
+++ b/extensions/windows-event-log/CollectorInitiatedSubscription.cpp
@@ -22,7 +22,6 @@
 #include <vector>
 #include <queue>
 #include <map>
-#include <vector>
 #include <set>
 #include <sstream>
 #include <stdio.h>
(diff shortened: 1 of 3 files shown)


extensions/windows-event-log/CollectorInitiatedSubscription.cpp
```diff
@@ -0,0 +1,777 @@
...
+  core::PropertyBuilder::createProperty("Delivery MaxLatency Time")->
+  isRequired(true)->
+  withDefaultValue<core::DataSizeValue>("10 min")->
+  withDescription("How long, in milliseconds, the event source should wait before sending events.")->
+  build());
+
+static SupportedProperty s_heartbeatInterval(
+  core::PropertyBuilder::createProperty("Heartbeat Interval")->
+  isRequired(true)->
+  withDefaultValue<core::DataSizeValue>("10 min")->
```
phrocker: I believe this is core::TimePeriodValue for the time based ones.

extensions/windows-event-log/CollectorInitiatedSubscription.cpp
```diff
@@ -0,0 +1,777 @@
...
+  core::PropertyBuilder::createProperty("Max Delivery Items")->
+  isRequired(true)->
+  withDefaultValue<core::DataSizeValue>("1000")->
+  withDescription("Determines the maximum number of items that will forwarded from an event source for each request.")->
+  build());
+
+static SupportedProperty s_deliveryMaxLatencyTime(
+  core::PropertyBuilder::createProperty("Delivery MaxLatency Time")->
+  isRequired(true)->
+  withDefaultValue<core::DataSizeValue>("10 min")->
```
arpadboda: I think TimePeriodValidator would be more appropriate for this usecase.

extensions/windows-event-log/CollectorInitiatedSubscription.cpp
```diff
@@ -0,0 +1,777 @@
...
+
+
+CollectorInitiatedSubscription::CollectorInitiatedSubscription(const std::string& name, utils::Identifier uuid)
+  : core::Processor(name, uuid), logger_(logging::LoggerFactory<CollectorInitiatedSubscription>::getLogger()) {
+  char buff[MAX_COMPUTERNAME_LENGTH + 1];
+  DWORD size = sizeof(buff);
+  if (GetComputerName(buff, &size)) {
+    computerName_ = buff;
+  } else {
+    logWindowsError(__LINE__, "GetComputerName");
```
arpadboda: Could you add a macro to avoid repeating __LINE__ for every usage?

extensions/windows-event-log/CollectorInitiatedSubscription.cpp
```diff
@@ -0,0 +1,777 @@
...
+      return;
+    }
+  }
+
+  // Check subscription runtime status.
+  checkSubscriptionRuntimeStatus();
+
+  const auto flowFileCount = processQueue(session);
+
+  const auto now = GetTickCount();
```
arpadboda: As far as I remember gettickkount _overflows_ after a while (at about 50 days). Running on Windows servers that might be exceeded causing issues here. 
I don't care about being platform-independent here given this code is only useful on Win, but std::chrono might be a safer choice for timing. 
If you are aware of Windows-specific cons of chrono, please don't hesitate to share!

extensions/windows-event-log/CollectorInitiatedSubscription.cpp
```diff
@@ -0,0 +1,777 @@
...
+{
+  std::function<void(void)> functor_;
+public:
+  OnScopeExit(const decltype(functor_)& functor) : functor_(functor) {}
+  ~OnScopeExit() { functor_(); }
+};
+
+#define SCOPE_EXIT_CONCAT(x, y) x##y
+#define SCOPE_EXIT_UNIQUE_NAME(name, counter) SCOPE_EXIT_CONCAT(name, counter)
+#define ON_SCOPE_EXIT(f) OnScopeExit SCOPE_EXIT_UNIQUE_NAME(onScopeExit, __COUNTER__)([&]{f;})
```
phrocker: utils/ScopedGuard.h might provide this same functionality

extensions/windows-event-log/CollectorInitiatedSubscription.cpp
phrocker: Good point from Arpad. My addition: utils/TimeUtil.h provides getTimeMillis which uses chrono, so that is an option, Alex, though if that if GetTickCount64 is an option I guess.

extensions/windows-event-log/CollectorInitiatedSubscription.cpp
```diff
@@ -0,0 +1,789 @@
...
+    }
+  }
+
+  operator std::set<core::Property>() const
+  {
+    return sSupportedProperties_;
+  }
+};
+
+static SuportedProperties s_supportedProperties;
```
arpadboda: Why do we need this as a global static? 

Another concern I have here is someone configuring two of these processors. 
As all the properties are shared that's not gonna work as expected. 

In case the goal is to to have at most one of these, we should:
-Clearly state that in docs
-Avoid instantiation of a 2nd one
-Have to think over how this gonna work via C2 manifest as it should also block designing a flow that contains two of this.

extensions/windows-event-log/CollectorInitiatedSubscription.cpp
```diff
@@ -0,0 +1,789 @@
...
+  T value() {
+    return t_;
+  }
+};
+
+void SupportedProperty<std::wstring>::Init(const std::shared_ptr<core::ProcessContext>& context) {
+  std::string val;
+  context->getProperty(getName(), val);
+
+  t_ = std::wstring(val.begin(), val.end());
```
bakaid: Can we be sure that the property value is ASCII-only? If not, then this has to be replaced with a proper conversion.

extensions/windows-event-log/CollectorInitiatedSubscription.cpp
```diff
@@ -0,0 +1,789 @@
...
+};
+
+void SupportedProperty<TimeProperty>::Init(const std::shared_ptr<core::ProcessContext>& context) {
+  std::string strTime;
+  context->getProperty(getName(), strTime);
+
+  int64_t time{};
+  core::TimeUnit unit;
+  core::Property::StringToTime(strTime, time, unit);
+  core::Property::ConvertTimeUnitToMS(time, unit, time);
```
phrocker: For those that are TimePeriodValue you can do context->getProperty(MinAge.getName(), uint64_t); and the conversion will be performed for you

extensions/windows-event-log/CollectorInitiatedSubscription.cpp
```diff
@@ -0,0 +1,789 @@
...
+
+          if (!EvtRender(NULL, hEvent, EvtRenderEventXml, size, 0, &used, &propertyCount)) {
+            if (ERROR_INSUFFICIENT_BUFFER == GetLastError()) {
+              if (used > s_maxBufferSize.value()) {
+                logger->log_error("Dropping event %x because it couldn't be rendered within %ll bytes.", hEvent, s_maxBufferSize.value());
+                return 0UL;
+              }
+
+              size = used;
+              std::vector<char> buf(size);
```
bakaid: Is there a reason to use a char buffer here? I think this breaks strict aliasing rules, (accessing wchar_t* through char* would be OK, because char can alias any type, but it is not OK do it the other way).
The official example I found also uses LPWSTR: https://docs.microsoft.com/en-us/windows/desktop/wes/rendering-events, so I think this should be `std::vector<wchar_t> buf(size/2);` (or `std::vector<wchar_t> buf(size/2 + 1)`, if we think the Windows API is really stupid.)

extensions/windows-event-log/CollectorInitiatedSubscription.cpp
```diff
@@ -0,0 +1,789 @@
...
+
+#define LOG_SUBSCRIPTION_ERROR(error) logError(__LINE__, error)
+#define LOG_SUBSCRIPTION_WINDOWS_ERROR(info) logWindowsError(__LINE__, info)
+
+struct ISupportedProperty {
+  void virtual Init(const std::shared_ptr<core::ProcessContext>& context) = 0;
+};
+
+class SuportedProperties {
+  std::vector<ISupportedProperty*> vISupportedProperty_;
```
msharee9: Is this a convention followed to subscript with v for vector member variable? Just curious to know.
If it is not a convention followed it would be better to name the variable as supportedPropertyList_ etc.

extensions/windows-event-log/CollectorInitiatedSubscription.cpp
```diff
@@ -0,0 +1,789 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "CollectorInitiatedSubscription.h"
+#include <vector>
+#include <queue>
+#include <map>
+#include <vector>
```
msharee9: <vector> included twice.

extensions/windows-event-log/CollectorInitiatedSubscription.cpp
```diff
@@ -0,0 +1,789 @@
...
+namespace processors {
+
+#define LOG_SUBSCRIPTION_ERROR(error) logError(__LINE__, error)
+#define LOG_SUBSCRIPTION_WINDOWS_ERROR(info) logWindowsError(__LINE__, info)
+
+struct ISupportedProperty {
+  void virtual Init(const std::shared_ptr<core::ProcessContext>& context) = 0;
+};
+
+class SuportedProperties {
```
msharee9: nitpick: SuportedProperties spelling

extensions/windows-event-log/CollectorInitiatedSubscription.cpp
```diff
@@ -0,0 +1,789 @@
...
+  std::vector<ISupportedProperty*> vISupportedProperty_;
+  std::set<core::Property> sSupportedProperties_;
```
msharee9: Since you are adding objects of derived type in this set which is declared with Base type template, there will be object slicing problem here. The derived class part of the object won't be copied.
Is it not better to use ISupportedProperty* here as well as you use for the vector.

extensions/windows-event-log/CollectorInitiatedSubscription.cpp
```diff
@@ -0,0 +1,789 @@
...
+static SuportedProperties s_supportedProperties;
+
+template <typename T>
+class SupportedProperty: public ISupportedProperty, public core::Property {
+  T t_;
+
+public:
+  template <typename ...Args>
+  SupportedProperty(const Args& ...args) : core::Property(args...) {
+    s_supportedProperties.add(*this);
```
msharee9: Please look at comment on line 57

extensions/windows-event-log/CollectorInitiatedSubscription.cpp
```diff
@@ -0,0 +1,789 @@
...
+    sSupportedProperties_.insert(t);
+  }
+
+  void init(const std::shared_ptr<core::ProcessContext>& context) {
+    for (auto pProp : vISupportedProperty_) {
+      pProp->Init(context);
+    }
+  }
+
+  operator std::set<core::Property>() const
```
msharee9: What does this function do? Can you please make a comment on top of this function. Is it doing type casting? but it is not very clear how it would be used.

msharee9: Or may be just define GetSupportedProperties function and return the set ?

extensions/windows-event-log/CollectorInitiatedSubscription.cpp
```diff
@@ -0,0 +1,789 @@
...
+  utils::ScopeGuard guard_hArray([hArray]() { EcClose(hArray); });
+
+  // Get the EventSources array size (number of elements).
+  DWORD dwEventSourceCount{};
+  if (!EcGetObjectArraySize(hArray, &dwEventSourceCount)) {
+    LOG_SUBSCRIPTION_WINDOWS_ERROR("EcGetObjectArraySize");
+    return false;
+  }
+
+  auto getArrayProperty = [this](EC_OBJECT_ARRAY_PROPERTY_HANDLE hArray, EC_SUBSCRIPTION_PROPERTY_ID propID, DWORD arrayIndex, DWORD flags, std::vector<BYTE>& buffer, PEC_VARIANT& vProperty)
```
msharee9: Is it better to have the lambda's return type specified explicitly? -> bool

extensions/windows-event-log/CollectorInitiatedSubscription.cpp
```diff
@@ -0,0 +1,789 @@
...
+  auto getStatus = [this](const std::wstring& eventSource, EC_SUBSCRIPTION_RUNTIME_STATUS_INFO_ID statusInfoID, DWORD flags, std::vector<BYTE>& buffer, PEC_VARIANT& vStatus) {
+    buffer.clear();
+    buffer.resize(sizeof(EC_VARIANT));
+    DWORD dwBufferSize{};
+    if (!EcGetSubscriptionRunTimeStatus(
+          s_subscriptionName.value().c_str(), 
+          statusInfoID, 
+          eventSource.c_str(),
+          flags,
+          (DWORD)buffer.size(),
```
msharee9: nitpick: If this is doing casting, prefer to use c++ style cast?

extensions/windows-event-log/CollectorInitiatedSubscription.cpp
```diff
@@ -0,0 +1,789 @@
...
+  };
+
+  for (DWORD i = 0; i < dwEventSourceCount; i++) {
+    std::vector<BYTE> eventSourceBuffer;
+    PEC_VARIANT vProperty = NULL;
+    if (!getArrayProperty(hArray, EcSubscriptionEventSourceAddress, i, 0, eventSourceBuffer, vProperty)) {
+      return false;
+    }
+
+    if (vProperty->Type != EcVarTypeNull && vProperty->Type != EcVarTypeString) {
```
msharee9: Are we sure vProperty will not be NULL here?

extensions/windows-event-log/CollectorInitiatedSubscription.cpp
```diff
@@ -0,0 +1,789 @@
...
+    if (!getStatus(eventSource.c_str(), EcSubscriptionRunTimeStatusActive, 0, buffer, vProperty)) {
+      return false;
+    }
+
+    if (vProperty->Type != EcVarTypeUInt32) {
+      logInvalidSubscriptionPropertyType(__LINE__, vProperty->Type);
+      return false;
+    }
+
+    auto runtimeStatus = vProperty->UInt32Val;
```
msharee9: local variable prefer const

extensions/windows-event-log/CollectorInitiatedSubscription.cpp
```diff
@@ -0,0 +1,789 @@
...
+
+    std::wstring strRuntimeStatus;
+
+    switch (vProperty->UInt32Val)
```
msharee9: Use the variable on line 427 here or we don't need that line at all.

extensions/windows-event-log/CollectorInitiatedSubscription.cpp
```diff
@@ -0,0 +1,789 @@
...
+  }
+
+  return true;
+}
+
+bool CollectorInitiatedSubscription::subscribe(const std::shared_ptr<core::ProcessContext> &context)
+{
+  logger_->log_debug("CollectorInitiatedSubscription: maxBufferSize_ %lld", s_maxBufferSize.value());
+
+  provenanceUri_ = "winlog://" + computerName_ + "/" + to_string(s_channel.value()) + "?" + to_string(s_query.value());
```
msharee9: Although this is not a big deal, I would use std::string::append here.

extensions/windows-event-log/CollectorInitiatedSubscription.cpp
```diff
@@ -0,0 +1,789 @@
...
+    MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
+    (LPTSTR)&lpMsg,
+    0, 
+    NULL);
+
+  logger_->log_error("Line %d: '%s': error %d: %s\n", line, info.c_str(), (int)error, (char *)lpMsg);
+
+  LocalFree(lpMsg);
+
+  int z = 11;
```
msharee9: Not sure why this is here.

extensions/windows-event-log/CollectorInitiatedSubscription.cpp
am-c-p-p: If `getArrayProperty` returns true, `vProperty` points to first element of `eventSourceBuffer`.

extensions/windows-event-log/CollectorInitiatedSubscription.cpp
am-c-p-p: Since it is not a big deal, will leave it as it is :-)

extensions/windows-event-log/CollectorInitiatedSubscription.cpp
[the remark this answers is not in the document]
am-c-p-p: This was fixed in last commit.

phrocker:
@msharee9 @bakaid @arpadboda I stepped away and am seeing a lot of minor comments. Are all major issues addressed so that this can move forward?

arpadboda:
> @msharee9 @bakaid @arpadboda I see some minor comments at this point. Are all major issues addressed? If so can we +1 and merge? Thanks!

@phrocker : my main concern was storing the properties as global static, so only one instance of the given processor can be instantiated without messing things up. @am-c-p-p  marked it as resolved, but nor I see related changes, neither a reply to prove me wrong (maybe I missed something), so I would prefer to have that concern of mine resolved before merging.

bakaid:
> @msharee9 @bakaid @arpadboda I see some minor comments at this point. Are all major issues addressed? If so can we +1 and merge? Thanks!
@phrocker All of my issues have been addressed, but I would also like to see the resolution of @arpadboda 's global static issue.

phrocker:
@arpadboda thanks. are you referring to this? https://github.com/apache/nifi-minifi-cpp/pull/573/files#diff-27c9e86bea7d1c8fc73b13cc6e695fe4R90 where the properties are a private member?

arpadboda:
> @arpadboda thanks. are you referring to this? https://github.com/apache/nifi-minifi-cpp/pull/573/files#diff-27c9e86bea7d1c8fc73b13cc6e695fe4R90 where the properties are a private member?

Yes, thanks, I missed that! Looks good now.

PR #585, opened 2019-06-06
MINIFICPP-910 - Extend StringUtils with string join capability

arpadboda:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/utils/StringUtils.h
+++ b/libminifi/include/utils/StringUtils.h
@@ -43,6 +43,26 @@ namespace nifi {
 namespace minifi {
 namespace utils {
 
+namespace {
+  template<class Char>
+  struct string_traits;
+  template<>
+  struct string_traits<char>{
+    template<class T>
+    static std::string convert_to_string(T&& t){
+      return std::to_string(std::forward<T>(t));
+    }
+  };
+
+  template<>
+  struct string_traits<wchar_t>{
+    template<class T>
+    static std::wstring convert_to_string(T&& t){
+      return std::to_wstring(std::forward<T>(t));
+    }
+  };
+}
+
 /**
  * Stateless String utility class.
  *
@@ -257,6 +277,71 @@ class StringUtils {
 
     return result_string;
   }
+
+  /**
+ * Concatenates strings stored in an arbitrary container using the provided separator.
+ * @tparam TChar char type of the string (char or wchar_t)
+ * @tparam U arbitrary container which has string or wstring value type
+ * @param separator that is inserted between each elements. Type should match the type of strings in container.
+ * @param container that contains the strings to be concatenated
+ * @return the result string
+ */
+  template<class TChar, class U, typename std::enable_if<std::is_same<typename U::value_type, std::basic_string<TChar>>::value>::type* = nullptr>
+  static std::basic_string<TChar> join(const std::basic_string<TChar>& separator, const U& container) {
+    typedef typename U::const_iterator ITtype;
+    ITtype it = container.cbegin();
+    std::basic_stringstream<TChar> sstream;
+    while(it != container.cend()) {
+      sstream << (*it);
+      ++it;
+      if(it != container.cend()) {
+        sstream << separator;
+      }
+    }
+    return sstream.str();
+  };
+
+  /**
+   * Just a wrapper for the above function to be able to create separator from const char* or const wchar_t*
+   */
+  template<class TChar, class U, typename std::enable_if<std::is_same<typename U::value_type, std::basic_string<TChar>>::value>::type* = nullptr>
+  static std::basic_string<TChar> join(const TChar* separator, const U& container) {
+    return join(std::basic_string<TChar>(separator), container);
+  };
+
+
+  /**
+   * Concatenates string representation of integrals stored in an arbitrary container using the provided separator.
+   * @tparam TChar char type of the string (char or wchar_t)
+   * @tparam U arbitrary container which has any integral value type
+   * @param separator that is inserted between each elements. Type of this determines the result type. (wstring separator -> wstring)
+   * @param container that contains the integrals to be concatenated
+   * @return the result string
+   */
+  template<class TChar, class U, typename std::enable_if<std::is_integral<typename U::value_type>::value>::type* = nullptr,
+      typename std::enable_if<!std::is_same<U, std::basic_string<TChar>>::value>::type* = nullptr>
+  static std::basic_string<TChar> join(const std::basic_string<TChar>& separator, const U& container) {
+    typedef typename U::const_iterator ITtype;
+    ITtype it = container.cbegin();
+    std::basic_stringstream<TChar> sstream;
+    while(it != container.cend()) {
+      sstream << string_traits<TChar>::convert_to_string(*it);
+      ++it;
+      if(it != container.cend()) {
+        sstream << separator;
+      }
+    }
+    return sstream.str();
+  }
+
+  /**
+   * Just a wrapper for the above function to be able to create separator from const char* or const wchar_t*
+   */
+  template<class TChar, class U, typename std::enable_if<std::is_integral<typename U::value_type>::value>::type* = nullptr,
+      typename std::enable_if<!std::is_same<U, std::basic_string<TChar>>::value>::type* = nullptr>
+  static std::basic_string<TChar> join(const TChar* separator, const U& container) {
+    return join(std::basic_string<TChar>(separator), container);
+  };
 
 };
 
--- a/libminifi/test/unit/StringUtilsTests.cpp
+++ b/libminifi/test/unit/StringUtilsTests.cpp
@@ -17,6 +17,7 @@
  */
 
 #include <string>
+#include <list>
 #include <vector>
 #include <cstdlib>
 #include "../TestBase.h"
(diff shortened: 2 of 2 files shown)


libminifi/src/utils/StringUtils.cpp
```diff
@@ -0,0 +1,162 @@
+/**
```
arpadboda: The content of this file is mostly moved from the header, the changes I made were to make linter happy. Some of these functions still leave space for improvement.

phrocker: hmm, can we minimize those changes? Like the idea but it seems that there is scope creep. Is there an actual motivation behind another ticket for which this is necessary? I think I'm -1 on general cleanup until the follow on release since we have so much to test in this release.

libminifi/test/unit/StringUtilsTests.cpp
```diff
@@ -17,32 +17,36 @@
...
 TEST_CASE("TestStringUtils::split3", "[test split multiple delimiter]") {
   std::vector<std::string> expected = { "hello", "world", "I'm", "a", "unit", "test" };
-  REQUIRE(expected == StringUtils::split("hello world I'm a unit test", " "));
+  REQUIRE(expected == split("hello world I'm a unit test", " "));
 }
 
 TEST_CASE("TestStringUtils::split4", "[test split classname]") {
   std::vector<std::string> expected = { "org", "apache", "nifi", "minifi", "utils", "StringUtils" };
-  REQUIRE(expected == StringUtils::split(org::apache::nifi::minifi::core::getClassName<org::apache::nifi::minifi::utils::StringUtils>(), "::"));
+  REQUIRE(expected == split("org::apache::nifi::minifi::utils::StringUtils", "::"));
```
phrocker: This allows us to inline test for a separate function, so the change seemed unnecessary.

libminifi/test/unit/StringUtilsTests.cpp
```diff
@@ -91,5 +95,19 @@ TEST_CASE("TestStringUtils::testEnv5", "[test split classname]") {
...
+TEST_CASE("TestStringUtils::testJoin", "[test string join]") {
+  std::set<std::string> strings = {"3", "2", "1"};
+  REQUIRE(join(",", strings) == "1,2,3");
+
+  std::wstring sep = L"é";
+  std::vector<std::wstring> wstrings = {L"1", L"2"};
+  REQUIRE(join(sep, wstrings) == L"1é2");
+
+  std::list<uint64_t> ulist = {1, 2};
+  REQUIRE(join(sep, ulist) == L"1é2");
```
phrocker: what purpose does join serve here? The semantics of a join seem clear to allow strings.

phrocker:
@arpadboda Another concern of mine is that this failed compilation on my RPI 32 bit in LE mode. I will try and get that info off the device.

libminifi/include/utils/StringUtils.h
```diff
@@ -46,43 +47,53 @@ namespace utils {
...
+}
+
+namespace StringUtils {
   /**
    * Converts a string to a boolean
    * Better handles mixed case.
    * @param input input string
    * @param output output string.
    */
-  static bool StringToBool(std::string input, bool &output) {
```
phrocker: As I suspected this caused issues. We're giving weak guidance to the linker. So this caused issues when running some extensions that are loaded dynamically ( where the local linking guidance causes no issues ) -- so while this isn't a bad, that must be resolved first.

libminifi/include/utils/StringUtils.h
```diff
@@ -92,172 +103,104 @@ class StringUtils {
...
+    return sstream.str();
+  };
+
+  template<class TChar, class U, typename std::enable_if<std::is_same<typename U::value_type, std::basic_string<TChar>>::value>::type* = nullptr>
+  std::basic_string<TChar> join(const TChar* separator, const U& container) {
+    return join(std::basic_string<TChar>(separator), container);
+  };
+
+
+  template<class TChar, class U, typename std::enable_if<std::is_integral<typename U::value_type>::value>::type* = nullptr,
```
phrocker: This needs to be well documented as what we are doing. The contract here allows for joining, but that doesn't come through with the name or the way this is written, so I think very explanatory comments need to indicate what is done in every case.

libminifi/src/utils/StringUtils.cpp
```diff
@@ -0,0 +1,162 @@
...
+    });
+
+    for (const auto &replacement : replacements) {
+      result_string = source_string.replace(replacement.first, replacement.second.first, replacement.second.second);
+    }
+
+    return result_string;
+  }
+
+  bool StringToFloat(std::string input, float &output, FailurePolicy cp) {  // NOLINT
```
phrocker: As a project we historically do not use NOLINT unless in test code.

libminifi/test/unit/StringUtilsTests.cpp
```diff
@@ -91,5 +95,19 @@ TEST_CASE("TestStringUtils::testEnv5", "[test split classname]") {
 
   std::string expected = "hello world ";
 
-  REQUIRE(expected == StringUtils::replaceEnvironmentVariables(test_string));
+  REQUIRE(expected == replaceEnvironmentVariables(test_string));
+}
+
+TEST_CASE("TestStringUtils::testJoin", "[test string join]") {
+  std::set<std::string> strings = {"3", "2", "1"};
+  REQUIRE(join(",", strings) == "1,2,3");
```
phrocker: tests for empty empty sets, and empty delimiters would be nice here, too.

libminifi/src/utils/StringUtils.cpp
phrocker: quite nice.

libminifi/src/utils/StringUtils.cpp
arpadboda: Yes it was, just in the header, where linter didn't catch this.

libminifi/test/unit/StringUtilsTests.cpp
arpadboda: That's true, but it basically it wasn't a nice idea to add a bunch of util function to a class instead of a namespace. 

I agree that this is a useful test, I will find another class and use split on the name of that.

libminifi/test/unit/StringUtilsTests.cpp
arpadboda: Good idea, thanks, will add!

libminifi/include/utils/StringUtils.h
```diff
@@ -92,172 +103,104 @@ class StringUtils {
...
+      typename std::enable_if<!std::is_same<U, std::basic_string<TChar>>::value>::type* = nullptr>
+  std::basic_string<TChar> join(const std::basic_string<TChar>& separator, const U& container) {
+    typedef typename U::const_iterator ITtype;
+    ITtype it = container.cbegin();
+    std::basic_stringstream<TChar> sstream;
+    while(it != container.cend()) {
+      sstream << string_traits<TChar>::convert_to_string(*it);
```
phrocker: Some good comments, thanks. So I have a better understanding ( and now that I'm not watching a movie trying to review this from my phone ), why not allow the stringstream to do the conversion? It's, in my compiler impls, faster. I agree that to_string can be cleaner for simple or single conversions, but you are already using a stream. Other implementations of join I've seen for strings in c++ typically use iterators and then a stream and concat. What was the impetus for the inline conversion?

arpadboda: For eg. uint8_t. 
Which is handled as char, so given the following:
```
std::list<uint_8t> l = { 10, 11 }
```
Using operator<< a new line and a tabulator would be printed instead of the numbers, while to_string properly converts them to string representation of the given number.

phrocker: This seems to enforce a semantic. I think basic stringstream assumes character if it is one. In that case, I don't see the need for the complexity as you could have one function that assumes what streamstream does and an overload that allows uint8_t to be interpreted with to_string. Two functions, much simpler and cleaner. Removing the rest of the changes will alleviate any linker issues.

arpadboda: That could help removing the two template structs that provide convert_to_string methods, the price would be two different template specialisations (one for char and uint8_t, the other is wchar_t and uint8_t), so it can be done, but not sure if that makes it any simpler. 

The rest of the changes are coming from the fact that the this was a class, which cannot be extended with template functions. 
Changing to namespace means that the linkage changes, which naturally need to be verified on all platforms, but it cannot be avoided.

phrocker: Why would you need specializations for "*char" ? So that they can be dealt with as numerics? 

Can't have templates in classes???

arpadboda: char: the specialisation is needed for string (aka std::basic_string<char>) and wstring (aka std::basic_string<wchar_t>). The latter one is especially needed as are to support Windows. 

Templates in classes: they are basically allowed, but full specialisation is not (https://en.cppreference.com/w/cpp/language/member_template), so it would fail any way: nor I can specialise the function for uint8_t, neither can add the current "convert_to_string" functions.

phrocker: I'm sorry I don't follow why that's absolutely necessary . We should have a discussion of what the differences are in a different format as I'm still -1 on this change even without the arm failures ( down to only two devices after an OS upgrade ).

arpadboda: Reduced the impact by keeping the class (will add a follow-up), should be good now.

arpadboda:
@phrocker : thanks, usage will come soon! :)

phrocker:
@arpadboda I hope to get this merged tomorrow. One of the builds passed, but the others are working on higher priority PRs.

PR #589, opened 2019-06-10
MINIFICPP-914: Incur wait while awaiting minifi shutdown during restart

phrocker:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/bin/minifi.sh
+++ b/bin/minifi.sh
@@ -89,6 +89,12 @@ active_pid() {
   fi
 }
 
+endnow() {
+   echo "Killing MiNiFi..."
+   kill -9 ${saved_pid} > /dev/null 2>&1
+   
+}
+
 install() {
     detectOS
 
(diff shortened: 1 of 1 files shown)


bin/minifi.sh
```diff
@@ -252,7 +252,12 @@ case "\$1" in
     restart)
         echo Restarting MiNiFi service
         \${bin_dir}/minifi.sh stop
-        \${bin_dir}/minifi.sh start
+	if [ "${saved_pid}" -gt 0 ]; then
+	        while [ $(active_pid ${saved_pid}) -eq 0 ]; do
```
arpadboda: Wonder if a timeout makes sense here. (If there is any chance for MiNiFi to get stuck during shutdown)

phrocker: My initial thought was that the user can ctrl+c either the service case or the agent restart case to determine why it didn't restart; however, I think we can give them the option to ctrl+c to forcibly terminate, even in the service case. kill -9 is safe to the processes. I'll make that quick update and you can take a look.

 There will always be some who have the opinion that we should wait and terminate and others will have the opinion that we should ctrl+c and allow investigation ( I had this same issue on another open source project I support ) -- but I think we should make the service and agent restart paths the same no matter the community preferred route.

PR #590, opened 2019-06-11
MINIFICPP-621 Nanofi Tailfile example

msharee9:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/ResourceClaim.h
+++ b/libminifi/include/ResourceClaim.h
@@ -59,6 +59,7 @@ class ResourceClaim : public std::enable_shared_from_this<ResourceClaim> {
   explicit ResourceClaim(std::shared_ptr<core::StreamManager<ResourceClaim>> claim_manager);
 
   explicit ResourceClaim(const std::string path, std::shared_ptr<core::StreamManager<ResourceClaim>> claim_manager, bool deleted = false);
+
   // Destructor
   ~ResourceClaim() {
   }
--- a/nanofi/CMakeLists.txt
+++ b/nanofi/CMakeLists.txt
@@ -37,6 +37,8 @@ file(GLOB NANOFI_SOURCES  "src/api/*.cpp" "src/core/*.c*" "src/cxx/*.cpp" "src/s
 
 file(GLOB NANOFI_EXAMPLES_SOURCES  "examples/*.c" )
 
+file(GLOB NANOFI_ECU_SOURCES "ecu/*.c")
+
 include(CheckCXXCompilerFlag)
 if (WIN32)
   if ((MSVC_VERSION GREATER "1900") OR (MSVC_VERSION EQUAL "1900"))
@@ -103,4 +105,5 @@ endif(ENABLE_PYTHON AND NOT STATIC_BUILD)
 
 if (NOT DISABLE_CURL)
 add_subdirectory(examples)
+add_subdirectory(ecu)
 endif()
--- a/nanofi/ecu/tail_file.c
+++ b/nanofi/ecu/tail_file.c
@@ -1,0 +1,206 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+    *
+    *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+#include "api/nanofi.h"
+#include "core/string_utils.h"
+#include "core/cstructs.h"
+#include "core/file_utils.h"
+#include <unistd.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <errno.h>
+#include <limits.h>
+#include <signal.h>
+#include <sys/stat.h>
+
+struct flow_file_records * flowfiles = NULL;
+nifi_instance * instance = NULL;
+standalone_processor * proc = NULL;
+int file_offset = 0;
+int stopped = 0;
+flow_file_list ff_list;
+token_list tks;
+
+void signal_handler(int signum) {
+    if (signum == SIGINT || signum == SIGTERM) {
+        stopped = 1;
+    }
+}
+
+void transmit_flow_files(nifi_instance * instance) {
+    flow_file_list_node * head = ff_list.head;
+    while (head) {
+        transmit_flowfile(head->ff_record, instance);
+        head = head->next;
+    }
+}
+
+void set_offset(int offset) {
+    file_offset = offset;
+}
+
+int get_offset() {
+    return file_offset;
+}
+
+void on_trigger_callback(processor_session * ps, processor_context * ctx) {
+
+    char file_path[4096];
+    char delimiter[3];
+
+    if (get_property(ctx, "file_path", file_path, sizeof(file_path)) != 0) {
+        return;
+    }
+
+    if (get_property(ctx, "delimiter", delimiter, sizeof(delimiter)) != 0) {
+        return;
+    }
+
+    if (strlen(delimiter) == 0) {
+        printf("Delimiter not specified or it is empty\n");
+        return;
+    }
+    char delim = delimiter[0];
+
+    if (delim == '\\') {
+          if (strlen(delimiter) > 1) {
+            switch (delimiter[1]) {
+              case 'r':
+                delim = '\r';
+                break;
+              case 't':
+                delim = '\t';
+                break;
+              case 'n':
+                delim = '\n';
+                break;
+              case '\\':
+                delim = '\\';
+                break;
+              default:
+                break;
+            }
+        }
+    }
+
+    tks = tail_file(file_path, delim, get_offset());
+
+    if (!validate_list(&tks)) return;
+
+    set_offset(get_offset() + tks.total_bytes);
+
+    token_node * head;
+    for (head = tks.head; head && head->data; head = head->next) {
+        flow_file_record * ffr = generate_flow_file(instance, proc);
+        const char * flow_file_path = ffr->contentLocation;
+        FILE * ffp = fopen(flow_file_path, "wb");
+        if (!ffp) {
+            printf("Cannot open flow file at path %s to write content to.\n", flow_file_path);
+            break;
+        }
+
+        int count = strlen(head->data);
+        int ret = fwrite(head->data, 1, count, ffp);
+        if (ret < count) {
+            fclose(ffp);
+            break;
+        }
+        fseek(ffp, 0, SEEK_END);
+        ffr->size = ftell(ffp);
+        fclose(ffp);
+        add_flow_file_record(&ff_list, ffr);
+    }
+    free_all_tokens(&tks);
+}
+
+int main(int argc, char** argv) {
+
+    if (argc < 6) {
+        printf("Error: must run ./tail_file <file> <interval> <delimiter> <nifi instance url> <remote port>\n");
+        exit(1);
+    }
+
+    char * file = argv[1];
+    char * interval = argv[2];
+    char * delimiter = argv[3];
+    char * instance_str = argv[4];
+    char * port_str = argv[5];
+
+    if (access(file, F_OK) == -1) {
+        printf("Error: %s doesn't exist!\n", file);
+        exit(1);
+    }
+
+    struct stat stats;
+    errno = 0;
+    int ret = stat(file, &stats);
+
+    if (ret == -1) {
+        printf("Error occurred while getting file status {file: %s, error: %s}\n", file, strerror(errno));
+        exit(1);
+    }
+    // Check for file existence
+    if (S_ISDIR(stats.st_mode)){
+        printf("Error: %s is a directory!\n", file);
+        exit(1);
+    }
+
+    errno = 0;
+    unsigned long intrvl = strtol(interval, NULL, 10);
+
+    if (errno == ERANGE || intrvl == LONG_MAX || intrvl == LONG_MIN) {
+        printf("Invalid interval value specified\n");
+        return 0;
+    }
+
+    struct sigaction action;
+    memset(&action, 0, sizeof(sigaction));
+    action.sa_handler = signal_handler;
+    sigaction(SIGTERM, &action, NULL);
+    sigaction(SIGINT, &action, NULL);
+
+    nifi_port port;
+
+    port.port_id = port_str;
+
+    instance = create_instance(instance_str, &port);
+
+    const char * processor_name = "TailFile";
+
+    add_custom_processor(processor_name, on_trigger_callback);
+
+    proc = create_processor(processor_name);
+
+    set_standalone_property(proc, "file_path", file);
+    set_standalone_property(proc, "delimiter", delimiter);
+
+    set_offset(0);
+    while (!stopped) {
+        flow_file_record * new_ff = invoke(proc);
+        transmit_flow_files(instance);
+        free_flow_file_list(&ff_list);
+        free_flowfile(new_ff);
+        sleep(intrvl);
+    }
+
+    printf("tail file processor stopped\n");
+    free_standalone_processor(proc);
+    free(instance);
+
+    return 0;
+}
--- a/nanofi/include/core/cstructs.h
+++ b/nanofi/include/core/cstructs.h
@@ -149,4 +149,41 @@ typedef struct cstream {
   SOCKET socket_;
 } cstream;
 
+/****
+ * ##################################################################
+ *  STRING OPERATIONS
+ * ##################################################################
+ */
+
+typedef struct token_node {
+    char * data;
+    struct token_node * next;
+} token_node;
+
+typedef struct token_list {
+    struct token_node * head;
+    struct token_node * tail;
+    uint64_t size;
+    uint64_t total_bytes;
+    int has_non_delimited_token;
+} token_list;
+
+/****
+ * ##################################################################
+ *  FLOWFILE OPERATIONS
+ * ##################################################################
+ */
+
+typedef struct flow_file_list_node {
+    flow_file_record * ff_record;
+    struct flow_file_list_node * next;
+} flow_file_list_node;
+
+typedef struct flow_file_list {
+    flow_file_list_node * head;
+    flow_file_list_node * tail;
+    int len;
+    int offset;
+} flow_file_list;
+
 #endif /* LIBMINIFI_SRC_CAPI_CSTRUCTS_H_ */
--- a/nanofi/include/core/string_utils.h
+++ b/nanofi/include/core/string_utils.h
@@ -1,0 +1,94 @@
+/**
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+#ifndef NIFI_MINIFI_CPP_STRING_UTILS_H
+#define NIFI_MINIFI_CPP_STRING_UTILS_H
+
+#include "cstructs.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define MAX_BYTES_READ 4096
+
+/**
+ * Tokenizes a delimited string and returns a list of tokens
+ * @param str the string to be tokenized
+ * @param delim the delimiting character
+ * @return a list of tokens
+ */
+token_list tokenize_string(const char * str, char delim);
+
+/**
+ * Tokenizes a delimited string and returns a list of tokens but excludes
+ * the last token if it is not delimited by the delimiter. This function
+ * is used by tailfile processor
+ * @param str the cstring to tokenize
+ * @param delim the delimiter to tokenize by
+ * @return a list of tokens
+ */
+token_list tokenize_string_tailfile(const char * str, char delim);
+
+/**
+ * Adds a token to the token list
+ * @param tks the token list to add the token to
+ * @param begin the beginning of the token
+ * @param len the length of the token
+ */
+void add_token_to_list(token_list * tks, const char * begin, uint64_t len);
+
+/**
+ * Deallocate one token node
+ * @param node the node in the list to be deallocated
+ */
+void free_token_node(token_node * node);
+
+/**
+ * Deallocate the dynamically allocated token list
+ * @param tks the token list to be freed
+ */
+void free_all_tokens(token_list * tks);
+
+/**
+ * Remove the tail node from the list
+ * @param tks the linked list of token nodes
+ */
+void remove_last_node(token_list * tks);
+
+/**
+ * Validate a linked list
+ * @param tks_list the list to be validated
+ * @return 1 if the list if valid else 0
+ */
+int validate_list(token_list * tk_list);
+
+/**
+ * Append one list to other (Chaining)
+ * @param to, the destination list to append to
+ * @param from, the source list
+ * @attention, if the to list is empty, to and from will be same after appending
+ */
+void attach_lists(token_list * to, token_list * from);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif //NIFI_MINIFI_CPP_STRING_UTILS_H
--- a/nanofi/src/core/file_utils.c
+++ b/nanofi/src/core/file_utils.c
@@ -1,0 +1,63 @@
+/**
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <errno.h>
+
+#include "api/nanofi.h"
+#include "core/string_utils.h"
+#include "core/file_utils.h"
+
+token_list tail_file(const char * file_path, char delim, int curr_offset) {
+    token_list tkn_list;
+    memset(&tkn_list, 0, sizeof(struct token_list));
+
+    if (!file_path) {
+        return tkn_list;
+    }
+
+    char buff[MAX_BYTES_READ + 1];
+    memset(buff, 0, MAX_BYTES_READ+1);
+    errno = 0;
+    FILE * fp = fopen(file_path, "rb");
+    if (!fp) {
+        printf("Cannot open file: {file: %s, reason: %s}\n", file_path, strerror(errno));
+        return tkn_list;
+    }
+    fseek(fp, curr_offset, SEEK_SET);
+
+    int bytes_read = 0;
+    int i = 0;
+    while ((bytes_read = fread(buff, 1, MAX_BYTES_READ, fp)) > 0) {
+        buff[bytes_read] = '\0';
+        struct token_list tokens = tokenize_string_tailfile(buff, delim);
+        if (tokens.size > 0) {
+            attach_lists(&tkn_list, &tokens);
+        }
+        tkn_list.total_bytes += tokens.total_bytes;
+        if (tokens.total_bytes > 0) {
+            curr_offset += tokens.total_bytes;
+            fseek(fp, curr_offset, SEEK_SET);
+        }
+        memset(buff, 0, MAX_BYTES_READ);
+    }
+    fclose(fp);
+    return tkn_list;
+}
--- a/nanofi/src/core/flowfiles.c
+++ b/nanofi/src/core/flowfiles.c
@@ -1,0 +1,53 @@
+/**
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "core/flowfiles.h"
+#include <string.h>
+
+void add_flow_file_record(flow_file_list * ff_list, flow_file_record * record) {
+    if (!ff_list || !record) return;
+
+    struct flow_file_list_node * new_node = (struct flow_file_list_node *)malloc(sizeof(struct flow_file_list_node));
+    new_node->ff_record = record;
+    new_node->next = NULL;
+
+    if (!ff_list->head || !ff_list->tail) {
+        ff_list->head = ff_list->tail = new_node;
+        ff_list->len = 1;
+        return;
+    }
+
+    ff_list->tail->next = new_node;
+    ff_list->tail = new_node;
+    ff_list->len++;
+}
+
+void free_flow_file_list(flow_file_list * ff_list) {
+    if (!ff_list || !ff_list->head) {
+        return;
+    }
+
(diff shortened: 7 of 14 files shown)


nanofi/examples/tail_file.c
```diff
@@ -0,0 +1,270 @@
...
+    if (flowfiles) {
+        for (int i = 0; i < flowfiles->len; ++i) {
+            if (flowfiles->records[i]) {
+                transmit_flowfile(flowfiles->records[i], instance);
+            }
+        }
+    }
+}
+
+char * get_string(char * begin, char * end) {
```
arpadboda: After I read the implementation I understood what this function does. 
Maybe _get_substring_ or _copy_substring_ as a name would be closer to what it does, but I hope you guys can come up with something even better. 
In the meanwhile a doxygen-style comment like in nanofi.h would be nice to describe the different usecases of the function (the string is expected to be null-terminated in case end is not specified).

I would also prefer to have const char * arguments given we don't modify the source. 
With the current signature the following wouldn't work:

```
const char * text = "nice code";
const char * text_part = get_string(text+5, NULL);
```

As this function is surely useful I think this should be moved to some of the nanofi core files (maybe a new one) to support future ourselves. 

For the cherry on top it could be trickily made variadic arg to allow the caller to only specify begin. :) (not mandatory at all)

nanofi/examples/tail_file.c
```diff
@@ -0,0 +1,270 @@
...
+    char * buff = NULL;
+
+    if (begin == end) {
+        buff = (char *)malloc(2 * sizeof(char));
+        strncpy(buff, begin, 1);
+        buff[1] = '\0';
+        return buff;
+    }
+
+    if (end) {
```
arpadboda: In these two blocks only the first line is different.

```
        len = end ? (end-begin) : strlen(begin);
        buff = (char *)malloc((len+1) * sizeof(char));
        strncpy(buff, begin, len);
        buff[len] = '\0';
```

nanofi/examples/tail_file.c
```diff
@@ -0,0 +1,270 @@
...
+    else {
+        len = strlen(begin);
+        buff = (char *)malloc((len+1) * sizeof(char));
+        strncpy(buff, begin, len);
+        buff[len] = '\0';
+    }
+    return buff;
+}
+
+void split_string(char * str, char delim, char ** collector, int * num_strings)
```
arpadboda: This is also a cool a powerful utility for nanofi, so I would prefer to have it in core.
Please make the source const.

To return multiple values the current way we do it in Nanofi is to return a struct that contains these (for eg. a ptr and the corresponding size). Please note that this is only my preference, feel free to convince me about the opposite way. 
In case you prefer to stay with the current signature please make sure to add comments to differentiate in and out args.

nanofi/examples/tail_file.c
```diff
@@ -0,0 +1,270 @@
...
+    standalone_processor *standalone_proc = create_processor(processor_name);
+
+    set_standalone_property(standalone_proc, "file_path", file);
+    set_standalone_property(standalone_proc, "delimiter", delimiter);
+
+    fileoffset = (file_offset *)malloc(sizeof(file_offset));
+    set_offset(0);
+    for (;;) {
+        flow_file_record * new_ff = invoke(standalone_proc);
+        transmit_flow_files(instance);
```
arpadboda: 👍 
One step closer to do something useful with Nanofi!

nanofi/examples/tail_file.c
```diff
@@ -0,0 +1,270 @@
...
+
+void free_all_strings(char ** strings, int num_strings) {
+    for (int i = 0; i < num_strings; ++i) {
+        if (strings[i]) free(strings[i]);
+    }
+}
+
+void on_trigger_callback(processor_session * ps, processor_context * ctx) {
+
+    char file_path[50];
```
bakaid: 49 as the maximum path length is a little short I think. Linux can have unlimited path length, but having at least 4096 (the usual PATH_MAX on unixes) as the buffer size would be preferable.

nanofi/examples/tail_file.c
```diff
@@ -0,0 +1,270 @@
...
+      }
+    }
+
+    int curr_offset = get_offset();
+    int max_bytes_read = 4096;
+    char buff[max_bytes_read];
+    FILE * fp = fopen(file_path, "rb");
+    if (!fp) return;
+    fseek(fp, curr_offset, SEEK_SET);
+    while (fgets(buff, max_bytes_read, fp) != NULL) {
```
bakaid: fgets will stop reading if it encounters a '\n', even if \n is not our delimiter.

nanofi/examples/tail_file.c
```diff
@@ -0,0 +1,270 @@
...
+        set_offset(ftell(fp));
+        char * strings[max_bytes_read];
```
bakaid: The number of delimiters is usually much smaller than the number of characters in the stream.
Preallocating as many possible substrings as there are characters is wasteful.
I would prefer for split_string to have a different contract, see my comments there.

nanofi/examples/tail_file.c
```diff
@@ -0,0 +1,270 @@
...
+    struct stat stats;
+    stat(file, &stats);
+
+    // Check for file existence
+    if (S_ISDIR(stats.st_mode)){
+        printf("Error: %s is a directory!\n", file);
+        exit(1);
+    }
+
+    unsigned long intrvl = strtol(interval, NULL, 10);
```
bakaid: According to the documentation of strtol, because all of its return values are potentially valid, you have to set errno to 0 before using it, so that you can differentiate errors from valid cases.
Not zeroing errno and using it afterwards can have unexpected consequences (using a potentially uninitialized value or using an errno set by a previous function).

nanofi/examples/tail_file.c
```diff
@@ -0,0 +1,270 @@
...
+    char * instance_str = argv[4];
+    char * port_str = argv[5];
+
+    if (access(file, F_OK) == -1) {
+        printf("Error: %s doesn't exist!\n", file);
+        exit(1);
+    }
+
+    struct stat stats;
+    stat(file, &stats);
```
bakaid: Nitpicking here, but the file could potentially change (be deleted, permissions changed) between calling access and stat, so it would be better to check stat's return value as well.

nanofi/examples/tail_file.c
bakaid: I would also like to this either documented or changed.
I have a few issues with this function:
 - the contract is not clear (you should preallocate collector, and preallocate it so that it could accommodate all potential substrings, and num_strings have to be zeroed before passing it to the function)
 - preallocating the collector so that it could store all potential substrings is wasteful. As we have the whole string in memory I would prefer for this function to iterate through the string, count the substrings, and allocate the return array itself.
 - based on its usage, duplicating the substrings is not needed. Returning with an array of offsets and lengths would be more efficient.
 - if we want to keep the source string unmodified, the current implementation is fine, but if we are ok with modifing it, the whole implementation could be replaced with strtok

nanofi/examples/tail_file.c
```diff
@@ -0,0 +1,270 @@
...
+    fileoffset->offset = offset;
+}
+
+int get_offset() {
+    return fileoffset->offset;
+}
+
+void free_all_strings(char ** strings, int num_strings) {
+    for (int i = 0; i < num_strings; ++i) {
+        if (strings[i]) free(strings[i]);
```
bakaid: Nitpicking here, but calling free with NULL is NOP (and therefore OK) according to the standard and works on all reasonable platforms.

nanofi/examples/tail_file.c
bakaid: My issue with the main business logic of this implementation is that it potentially cuts the file into arbitrary pieces. It is not guaranteed that there will be a delimiter in the 4096 bytes we read, neither is it guaranteed that the last character of the buffer will be a delimiter. This means that we will cut long lines into 4096-byte pieces, and cut last lines in a buffer in two (or more, if the other process writing the file is currently writing the line).
Data should be buffered until there is a delimiter, and a flowfile should only be created when we find one.

nanofi/examples/tail_file.c
```diff
@@ -0,0 +1,270 @@
...
+    fileoffset = (file_offset *)malloc(sizeof(file_offset));
+    set_offset(0);
+    for (;;) {
+        flow_file_record * new_ff = invoke(standalone_proc);
+        transmit_flow_files(instance);
+        free_flowfile(new_ff);
+        sleep(intrvl);
+    }
+
+    free(fileoffset);
```
bakaid: I am not sure about how signal handling is implemented in nanofi (if at all), but I do not see how we could reach this code.

nanofi/examples/tail_file.c
bakaid: Also, I am not sure if this works on Windows, I think it only has _stat and struct _stat.

nanofi/examples/tail_file.c
```diff
@@ -0,0 +1,270 @@
...
+        exit(1);
+    }
+
+    char * file = argv[1];
+    char * interval = argv[2];
+    char * delimiter = argv[3];
+    char * instance_str = argv[4];
+    char * port_str = argv[5];
+
+    if (access(file, F_OK) == -1) {
```
bakaid: I think there is only _access on Windows.

nanofi/examples/tail_file.c
arpadboda: I think nanofi examples are not *yet* meant to be compiled on Win.

bakaid: I did not know about that. The issue specified to "Create examples that can tail log files ( perhaps like TailFile, but strictly in C ) that works in **windows** and *nix"

nanofi/examples/tail_file.c
arpadboda: I don't know about any handler for sigterm (which would apply here), so this is code is unreachable.

nanofi/examples/tail_file.c
arpadboda: Seems we leave some space for improvement in Jira handling:
https://issues.apache.org/jira/browse/MINIFICPP-777

libminifi/include/ResourceClaim.h
```diff
@@ -59,6 +59,7 @@ class ResourceClaim : public std::enable_shared_from_this<ResourceClaim> {
   explicit ResourceClaim(std::shared_ptr<core::StreamManager<ResourceClaim>> claim_manager);
 
   explicit ResourceClaim(const std::string path, std::shared_ptr<core::StreamManager<ResourceClaim>> claim_manager, bool deleted = false);
+
```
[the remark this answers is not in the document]
msharee9: Yes, it is uninteded. Will fix that.

nanofi/examples/tail_file.c
```diff
@@ -0,0 +1,270 @@
...
+#include <errno.h>
+#include <limits.h>
+#include <sys/stat.h>
+
+typedef struct flow_file_records {
+    flow_file_record ** records;
+    uint64_t len;
+} flow_file_records;
+
+typedef struct file_offset {
```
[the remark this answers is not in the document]
msharee9: We can use int for sure here. Will change that.

nanofi/examples/tail_file.c
msharee9: This function is gone from latest update.

nanofi/examples/tail_file.c
msharee9: This function is no more present in the latest update.

nanofi/examples/tail_file.c
msharee9: @bakaid please review the updated changes. If we return an array of subsets wouldn't it make this function complex to use? The user of this function then has to deal with offsets if he wants to extract the strings. We can customize this function if we want to make it efficient by not duplicating the strings, but at present the usage from nanofi tailfile is to split blocks of chars of size 4KB. I don't think that is big enough to think of efficiency. As we move along and  encounter use cases we will refine this function and make it efficient.

nanofi/examples/tail_file.c
msharee9: Will take a look at that.

nanofi/examples/tail_file.c
msharee9: Changed the implementation. Now using fread. Please see the updated code.

nanofi/examples/tail_file.c
msharee9: Updated code.

nanofi/examples/tail_file.c
msharee9: Updated code.

nanofi/examples/tail_file.c
msharee9: Agreed. Updated code to reflect this.

nanofi/src/core/string_utils.c
```diff
@@ -0,0 +1,87 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "core/cstructs.h"
+#include "core/string_utils.h"
+#include <string.h>
+#include <stdlib.h>
+
+tokens * tokenize_string(char * str, const char * delimiter) {
```
arpadboda: The input str should be const as well.

nanofi/src/core/string_utils.c
```diff
@@ -0,0 +1,87 @@
...
+    if (!str || !delimiter) return NULL;
+
+    char delim = delimiter[0];
```
arpadboda: This doesn't belong here in my opinion, I would prefer the delimiter being a single char and this logic kept outside. (in tailfile)

nanofi/include/core/string_utils.h
```diff
@@ -0,0 +1,35 @@
...
+#ifndef NIFI_MINIFI_CPP_STRING_UTILS_H
+#define NIFI_MINIFI_CPP_STRING_UTILS_H
+
+#include "cstructs.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+tokens * tokenize_string(char * str, const char * delimiter);
```
arpadboda: Docstring as well to describe the contract. Quite important given this function allocates memory, but the caller has to free it.

nanofi/include/core/cstructs.h
```diff
@@ -149,4 +149,15 @@ typedef struct cstream {
...
+/****
+ * ##################################################################
+ *  STRING OPERATIONS
+ * ##################################################################
+ */
+
+typedef struct tokens {
+    char ** str_list;
+    uint64_t num_strings;
+    uint64_t total_bytes;
```
arpadboda: What's the purpose of this?

nanofi/src/core/string_utils.c
```diff
@@ -0,0 +1,87 @@
...
+    while ((end = strchr(begin, delim))) {
+        if (begin == end) {
+            begin++;
+            continue;
+        }
+        begin = (end+1);
+        num_strings++;
+    }
+
+    struct tokens * tks = (struct tokens *)malloc(sizeof(tokens));
```
arpadboda: Why allocate this?
This struct just encapsulates the string pointer with the correspondig size, the struct itself can be returned simply.

nanofi/src/core/string_utils.c
```diff
@@ -0,0 +1,87 @@
...
+    tks->str_list = malloc(sizeof(char *) * (num_strings));
+    tks->num_strings = 0;
+    tks->total_bytes = 0;
+
+    for (int i = 0; i < num_strings; ++i) {
```
arpadboda: If you allocate the block with calloc, you get it filled with zeros, so you don't need this at all.

arpadboda:
Please also fix the build error in Travis:
```
/home/travis/build/apache/nifi-minifi-cpp/nanofi/src/core/string_utils.c: In function â€˜tokenize_stringâ€™:
/home/travis/build/apache/nifi-minifi-cpp/nanofi/src/core/string_utils.c:66:5: error: â€˜forâ€™ loop initial declarations are only allowed in C99 mode
     for (int i = 0; i < num_strings; ++i) {
     ^
/home/travis/build/apache/nifi-minifi-cpp/nanofi/src/core/string_utils.c:66:5: note: use option -std=c99 or -std=gnu99 to compile your code
make[2]: *** [nanofi/CMakeFiles/nanofi.dir/src/core/string_utils.c.o] Error 1
```

nanofi/src/core/string_utils.c
```diff
@@ -0,0 +1,87 @@
...
+
+    begin = str;
+    end = NULL;
+    while ((end = strchr(begin, delim))) {
+        if (begin == end) {
+            begin++;
+            tks->total_bytes++;
+            continue;
+        }
+        int len = end - begin;
```
bakaid: `tks->str_list[tks->num_strings++] = strndup(begin, end - begin);`

nanofi/src/core/string_utils.c
bakaid: In my opinion this function works unexpectedly if:
 - there is no delimiter in the string (I think it should return with a single token containing the entire string)
 - for the last token after the last delimiter, before the terminating '\0': it does not create a token for it

I assume that it works this way because of the business logic we want to achieve in TailFile (namely not creating a token until we encounter a delimiter), but I don't think this function works the way most people would expect.

nanofi/include/core/cstructs.h
```diff
@@ -149,4 +149,15 @@ typedef struct cstream {
   SOCKET socket_;
 } cstream;
 
+/****
+ * ##################################################################
+ *  STRING OPERATIONS
+ * ##################################################################
+ */
+
+typedef struct tokens {
```
bakaid: As this struct contains dynamically allocated memory, please provide a free_tokens function for it: it is less error-prone than manually freeing its members every time.

nanofi/examples/tail_file.c
```diff
@@ -0,0 +1,223 @@
...
+    if (!fp) return;
+    fseek(fp, curr_offset, SEEK_SET);
+
+    int bytes_read = fread(buff, 1, max_bytes_read, fp);
+    if (bytes_read > 0) {
+        struct tokens * tks = tokenize_string(buff, delimiter);
+
+        if (!tks || tks->num_strings == 0) return;
+
+        set_offset(get_offset() + tks->total_bytes);
```
bakaid: I think the entire processor will stop working if there is a line longer than max_bytes_read-1 bytes in the file: you only create tokens for entire lines, and seek the file after the end of the last processed line. But if there is no delimiter in max_bytes_read-1 bytes, you will never be able to exit this condition and will try to read this line and fail every time you are triggered.

bakaid: I think there is no harm caching the current line in memory if it has not ended in the current read.
Also, I don't think you should only read once every trigger: you should read and create FlowFile until you hit EOF.

nanofi/examples/tail_file.c
```diff
@@ -0,0 +1,223 @@
...
+
+    int curr_offset = get_offset();
+    int max_bytes_read = 4096;
+    char buff[max_bytes_read];
+    memset(buff,'\0', max_bytes_read);
+    FILE * fp = fopen(file_path, "rb");
+    if (!fp) return;
+    fseek(fp, curr_offset, SEEK_SET);
+
+    int bytes_read = fread(buff, 1, max_bytes_read, fp);
```
bakaid: fread, unlike fgets does NOT append a terminating null character. If you were to read a file at least max_bytes_read long, buff will not be a valid c string and a buffer overflow will occur.

nanofi/src/core/string_utils.c
phrocker: sorry @msharee9 I haven't been able to take a look as others have -- but it seems like they're offering good comments. I skimmed the review quickly and this caught my eye due to strndup.

I don't believe strndup is commonly implemented. A project that I've worked on previously ( https://github.com/rougier/freetype-gl/blob/master/platform.c ) has this file for this specific reason. I've seen similar implementations across other projects so may be a nice string util if and only if you feel this a worthwhile change.

nanofi/examples/tail_file.c
```diff
@@ -0,0 +1,257 @@
...
+
+struct flow_file_records * flowfiles = NULL;
+nifi_instance * instance = NULL;
+standalone_processor * proc = NULL;
+int file_offset = 0;
+int stopped = 0;
+
+void signal_handler(int signum) {
+    if (signum == SIGINT || signum == SIGTERM) {
+        stopped = 1;
```
phrocker: Can make a follow on ticket for this ( especially since I'm just skimming and don't have full context ): sig_atomic_t is typically used here as it's a cross platform way to achieve atomicity; however, I see you are using int, which is probably going to be written in a single instruction, anyway.

 Since this aims to achieve cross platform compilation anyway, I think a ticket may be warranted to explore if this is actually an issue or not. I'll leave that up to you to decide if a ticket for exploration is warranted. It may be the case that int will never provide a problem.

nanofi/examples/tail_file.c
```diff
@@ -0,0 +1,257 @@
...
+    }
+}
+
+void transmit_flow_files(nifi_instance * instance) {
+    NULL_CHECK( ,flowfiles);
+    int i;
+    for (i = 0; i < flowfiles->len; ++i) {
+        NULL_CHECK( ,flowfiles->records[i]);
```
phrocker: This seems like a case where a NULL_CHECK would indicate a bug somewhere in our or the implementation of this library. That seems to indicate we really should be performing an assertion. This function seems to also exist within comms.h "transmit_to_nifi".  Can create a follow on -- but there is a "blocks" directory for some of our code deemed as building blocks but not part of the primary API, some of the code here that is useful for other programs like this can be placed there. Further, can we move this from example to a directory named, "ECU". Eventually we can break that out into a separate entity.

nanofi/src/core/string_utils.c
```diff
@@ -0,0 +1,149 @@
...
+    }
+}
+
+token_list tokenize_string(const char * str, char delim) {
+    token_list tks;
+    memset(&tks, 0, sizeof(struct token_list));
+
+    if (!str) return tks;
+
+    char * begin = (char *)str;
```
arpadboda: Please leave this const!
Pointer to const array (aka const char *) can be incremented as you need. 
Const pointers to char * (akar char * const) cannot, but you don't have any of those.

nanofi/src/core/string_utils.c
```diff
@@ -0,0 +1,149 @@
...
+    while (tmp_head->next && (tmp_head->next != tmp_tail)) {
+        tmp_head = tmp_head->next;
+    }
+
+    struct token_node * tail_node = tmp_tail;
+    tks->tail = tmp_head;
+    tks->tail->next = NULL;
+
+    tks->size--;
+    tks->total_bytes -= (strlen(tail_node->data));
```
bakaid: This does not account for the skipped multiple delimiters, causing an erroneus offset.

nanofi/src/core/file_utils.c
```diff
@@ -0,0 +1,54 @@
...
+
+    int max_bytes_read = 4096;
+    char buff[max_bytes_read + 1];
+    memset(buff,'\0', max_bytes_read);
+    FILE * fp = fopen(file_path, "rb");
+    if (!fp) return tkn_list;
+    fseek(fp, curr_offset, SEEK_SET);
+
+    int bytes_read = 0;
+    while ((bytes_read = fread(buff, 1, max_bytes_read, fp)) > 0) {
```
bakaid: I think this modification preserved and amplified the original problem with the logic of this processor.
TailFile, in my interpretation, should do the following on each on_trigger:
 * Restore its previous state (its offset in the file)
 * Repeat until it reaches EOF (or an error occurs):
   * Read the file until a delimiter is reached.
   * Create a FlowFile from the token.
 * Store the position in the file after the end of the last token from which a FlowFile was created

Unless I am mistaken, what this implementation does is:
 * Repeat until it reaches EOF (or an error occurs):
   * Read at most 4096 bytes from the file
   * Find all completely delimited tokens in the resulting buffer
   * Discard the last part of the buffer (or all of it, if there was no delimiter in the read bytes)

This means that every buffer will be skipped that does not contain a delimiter, and the last non-delimited parts of every buffer will be also skipped.
The last version of this PR introduces an additional bug: because the total_bytes contains only the non-skipped parts of the buffers, the resulting total_bytes, which will be later used for determining the offset in the file will be practically a random offset in the file.

Please review and fix the business logic and write tests, including tests for long (>> buffer) lines, delimiters occuring at buffer-boundary and delimiters occuring at non buffer-boundary positions.

phrocker: I haven't fully reviewed this but I think some of this can be a follow on discussion. Calling this "tail file" is a bit of a misnomer in regards to what we're attempting to achieve. There will be slight differences -- and those should be captured a little further in may be some follow on discussion. I do think that the current logic if laid out as such is potentially a bug -- but I envision slight changes in how this functionality is implemented/used

nanofi/tests/CTailFileTests.cpp
```diff
@@ -0,0 +1,365 @@
...
+    const std::string filePath = fm.getFilePath();
+    fm.CloseStream();
+
+    struct token_list tokens = tail_file(filePath.c_str(), '-', 0);
+    test_lists_equal(&tokens, std::vector<std::string>{std::move(s)});
+}
+
+// Although there is no delimiter within the string that is equal to 4096 bytes or longer
+// tail_file creates a flow file for each subsequent 4096 byte chunk. It leaves the last chunk
+// if it is smaller than 4096 bytes and not delimited
```
bakaid: Both NiFi (https://github.com/apache/nifi/blob/ea9b0db2f620526c8dd0db595cf8b44c3ef835be/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/TailFile.java#L832) and MiNiFi (https://github.com/apache/nifi-minifi-cpp/blob/master/libminifi/src/core/ProcessSession.cpp#L560) implement the TailFile processor to only ingest data (create flow files) when it encounters a delimiter. It is also in the specification of the processor (https://nifi.apache.org/docs/nifi-docs/components/org.apache.nifi/nifi-standard-nar/1.9.2/org.apache.nifi.processors.standard.TailFile/).
I think that in nanofi we can deviate from the behaviour of NiFi or MiNiFi when it is necessary or beneficial.
In this case, however, I do not see how this behaviour (creating flow files from the data if it is larger than or equal to 4096 bytes even if there was no delimiter) would be beneficial. On the contrary, I think it is an unexpected behaviour, and makes the processor really hard to use, as it can potentially chunk your data into arbitrary pieces which are hard to parse and manage later on.
I am willing to be convinced that this behaviour is better (I do not think it is necessary), but so far I have not seen a reason to believe that.

bakaid: I am sorry, I have misread how MiNiFi's TailFile works. It has two modes: with delimiter and without delimiter. If a delimiter is not set, it will read until EOF and create a flow file from all of that data.
If a delimiter is set, it will indeed only try to read at max 4095+1 bytes to find a delimiter.
Unfortunately that implementation is wrong: std::ifstream's getline will set the failbit, if it doesn't find the delimiter in the given size. We will check for this failbit, and stop reading the file without creating any FlowFiles.
And because we only set TailFile's current offset in the file based on the generated FlowFiles, this means that if TailFile encounters a line longer than 4095 bytes without a delimiter, it will stop working indefinitely (this was a bug in a previous version of this PR too).
I have just tried this out, and it indeed fails. :(
This means that MiNiFi's TailFile will have to be fixed: at a minimum, to fix the implementation, but I propose to change it to match NiFi's behaviour, and only ingest data when a delimiter is found.

nanofi/src/core/string_utils.c
```diff
@@ -0,0 +1,165 @@
...
+
+token_list tokenize_string(const char * begin, char delim) {
+    token_list tks;
+    memset(&tks, 0, sizeof(struct token_list));
+
+    if (!begin) return tks;
+
+    char * end = NULL;
+
+    while ((end = strchr(begin, delim))) {
```
arpadboda: Doesn't it produce warning?
const char * is assigned to char * (end should be const, too)

nanofi/tests/CTailFileTests.cpp
```diff
@@ -0,0 +1,365 @@
...
+    REQUIRE(tknlist->size == sv.size());
+    for (const auto& s : sv) {
+        if (tknlist->head) {
+            REQUIRE(strcmp(s.c_str(), tknlist->head->data) == 0);
+            tknlist->head = tknlist->head->next;
+        }
+    }
+}
+
+std::string join_strings(const std::vector<std::string>& strings, const std::string& token) {
```
arpadboda: https://github.com/apache/nifi-minifi-cpp/pull/585

One more reason to move that PR forward (which provides a much more generic and efficient way for doing the same).

nanofi/ecu/tail_file.c
```diff
@@ -0,0 +1,214 @@
...
+int get_offset() {
+    return file_offset;
+}
+
+void on_trigger_callback(processor_session * ps, processor_context * ctx) {
+
+    char file_path[4096];
+    char delimiter[2];
+
+    if (get_property(ctx, "file_path", file_path, 50) != 0) {
```
bakaid: The buffer has been extended to 4096 bytes but here the size has been left 50.
Perhaps using sizeof(file_path) (and sizeof(delimiter) below) as the sizes would be less error-prone.

nanofi/ecu/tail_file.c
```diff
@@ -0,0 +1,214 @@
...
+
+    if (get_property(ctx, "delimiter", delimiter, 2) != 0) {
+        return;
+    }
+
+    if (strlen(delimiter) == 0) {
+        printf("Delimiter not specified or it is empty\n");
+        return;
+    }
+    char delim = '\0';
```
bakaid: This entire section until checking whether the delimiter is an escape sequence is unnecessary.
You have already returned if `strlen(delimiter) == 0`, so `strlen(delimiter) > 0` is true.
As `strlen(delimiter) > 0` is true, the first character of the string cannot be a terminating null character, so the `if (delim == '\0')` check is unnecessary as well.
The entire section can be substituted with `char delim = delimiter[0];`

nanofi/ecu/tail_file.c
```diff
@@ -0,0 +1,214 @@
...
+              case 't':
+                delim = '\t';
+                break;
+              case 'n':
+                delim = '\n';
+                break;
+              case '\\':
+                delim = '\\';
+                break;
+              default:
```
bakaid: Returning an error here instead of silently using \ as the delimiter would be better, but I can live with it.

nanofi/ecu/tail_file.c
```diff
@@ -0,0 +1,214 @@
...
+
+    if (access(file, F_OK) == -1) {
+        printf("Error: %s doesn't exist!\n", file);
+        exit(1);
+    }
+
+    struct stat stats;
+    int ret = stat(file, &stats);
+
+    errno = 0;
```
bakaid: As you set errno here to 0, strerror(errno) will not return with stat's error.

nanofi/src/core/flowfiles.c
```diff
@@ -0,0 +1,52 @@
...
+void free_flow_file_list(flow_file_list * ff_list) {
+    if (!ff_list || !ff_list->head) {
+        return;
+    }
+
+    flow_file_list_node * head = ff_list->head;
+    while (head) {
+        free_flowfile(head->ff_record);
+        free(head);
+        head = head->next;
```
bakaid: This is a use-after-free.
You have to save the next pointer before freeing the head structure.

nanofi/src/core/string_utils.c
```diff
@@ -0,0 +1,165 @@
...
+    tk_list->size++;
+    tk_list->total_bytes += len;
+}
+
+void free_token_node(struct token_node * node) {
+    if (node) {
+        free(node->data);
+    }
+    free(node);
+    memset(node, 0, sizeof(struct token_node));
```
bakaid: This is a use-after-free, you can't null a struct after you have freed it.

nanofi/src/core/string_utils.c
```diff
@@ -0,0 +1,165 @@
...
+        free_token_node(node);
+    }
+}
+
+void print_token_list(token_list * tokens) {
+    if (tokens) {
+        token_node * head = tokens->head;
+        int i = 0;
+        while (head) {
+            printf("Token %d : %s Length = %d\n", i, head->data, strlen(head->data));
```
bakaid: strlen returns a size_t, the printf format specifier for it should be %zu, not %d.

nanofi/src/core/string_utils.c
bakaid: strchr returns const char*, so I agree, end should be const too.

bakaid:
@phrocker please see my last review. There are - among other things - multiple use-after-free issues in the code. It should not be merged until those are fixed.

phrocker:
@msharee9 per @bakaid there are some use after frees that probably should be addressed. Thanks!

nanofi/src/core/string_utils.c
```diff
@@ -0,0 +1,165 @@
...
+        free_token_node(node);
+    }
+}
+
+void print_token_list(token_list * tokens) {
+    if (tokens) {
+        token_node * head = tokens->head;
+        int i = 0;
+        while (head) {
+            printf("Token %d : %s Length = %lu\n", i, head->data, strlen(head->data));
```
bakaid: l is the length sub-specifier for long, strlen return a size_t, and the sub-specifier for size_t is z, so this should be %zu not %lu

nanofi/CMakeLists.txt
```diff
@@ -103,4 +105,5 @@ endif(ENABLE_PYTHON AND NOT STATIC_BUILD)
 
 if (NOT DISABLE_CURL)
 add_subdirectory(examples)
+add_subdirectory(ecu)
```
phrocker: Seeing this gets me super excited!

nanofi/tests/CTailFileTests.cpp
```diff
@@ -0,0 +1,365 @@
...
+    const uint64_t totalStringsSize = s1.size() + s2.size() + d1.size();
+    const std::string filePath = fm.getFilePath();
+    const uint64_t bytesWrittenToStream = fm.GetFileSize();
+    REQUIRE(bytesWrittenToStream == totalStringsSize);
+
+    struct token_list tokens = tail_file(filePath.c_str(), '-', 0);
+    test_lists_equal(&tokens, std::vector<std::string>{std::move(s1), std::move(s2)});
+}
+
+TEST_CASE("Test tail file having more than 4096 bytes with delimiter and second chunk less than 4096", "[testTailFileWithDelimitedString]") {
```
phrocker: Taking another casual look to make it look like I'm useful. I like the breadth of tests! Nice!

nanofi/ecu/tail_file.c
```diff
@@ -0,0 +1,206 @@
...
+}
+
+int get_offset() {
+    return file_offset;
+}
+
+void on_trigger_callback(processor_session * ps, processor_context * ctx) {
+
+    char file_path[4096];
+    char delimiter[2];
```
bakaid: This includes the terminating null character, meaning that only 1 character of the parameter will be read, so escape sequences won't work.

phrocker:
@bakaid Are you good with merging?

bakaid:
@phrocker I think the serious issues preventing merge has been fixed, but if I gather correctly @msharee9 is still working on this PR.

PR #596, opened 2019-06-24
MINIFICPP-925 - Fix TailFile hang on long lines

bakaid:

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/include/core/ProcessSession.h
```diff
@@ -109,8 +109,8 @@ class ProcessSession : public ReferenceContainer {
    */
   void importFrom(io::DataStream &stream, const std::shared_ptr<core::FlowFile> &flow);
   // import from the data source.
-  void import(std::string source, const std::shared_ptr<core::FlowFile> &flow, bool keepSource = true, uint64_t offset = 0);
-  void import(std::string source, std::vector<std::shared_ptr<FlowFileRecord>> &flows, bool keepSource, uint64_t offset, char inputDelimiter);
```
phrocker: You're changing the contract. We will need to keep the (...bool,uint64_t,char) variant.

phrocker: adding variants is not an issue, though.

libminifi/src/core/ProcessSession.cpp
```diff
@@ -537,63 +534,82 @@ void ProcessSession::import(std::string source, const std::shared_ptr<core::Flow
...
+  std::vector<uint8_t> buffer(getpagesize());
   try {
-    // Open the input file and seek to the appropriate location.
-    std::ifstream input;
-    logger_->log_debug("Opening %s", source);
-    input.open(source.c_str(), std::fstream::in | std::fstream::binary);
-    if (input.is_open() && input.good()) {
-      if (offset != 0) {
+    try {
+      std::ifstream input;
```
phrocker: is this a change in the spacing? Did you apply the linter?

libminifi/src/utils/StringUtils.cpp
```diff
@@ -0,0 +1,162 @@
+/**
```
phrocker: -1 on this change. Had issues linking on arm devices with changes like these. It's like a cmake issue, but let's not make any structural changes, especially in PRs.

extensions/standard-processors/processors/LogAttribute.cpp
```diff
@@ -137,33 +137,32 @@ void LogAttribute::onTrigger(const std::shared_ptr<core::ProcessContext> &contex
       message << "\n" << "Payload:" << "\n";
       ReadCallback callback(flow->getSize());
```
phrocker: this doesn't seem related to this PR.

extensions/standard-processors/processors/HashContent.h
```diff
@@ -72,7 +64,7 @@ namespace {
     if (ret_val.second > 0) {
       unsigned char digest[MD5_DIGEST_LENGTH];
       MD5_Final(digest, &context);
-      ret_val.first = digestToString(digest, MD5_DIGEST_LENGTH);
+      ret_val.first = utils::StringUtils::to_hex(digest, MD5_DIGEST_LENGTH, true /*uppercase*/);
```
phrocker: ditto here, too. If you wish submit the to_hex pr changing variants let's get that as a separate PR so we can accurately prioritize its review. Thanks!

libminifi/src/utils/StringUtils.cpp
phrocker: Actually, this looks a little different and may not have the same linking issues but I'm -1 on this in this PR. Can you submit this into a separate PR so it can be prioritized whether we should include this into this release?

libminifi/src/core/ProcessSession.cpp
```diff
@@ -606,33 +622,28 @@ void ProcessSession::import(std::string source, std::vector<std::shared_ptr<Flow
...
-      throw Exception(FILE_OPERATION_EXCEPTION, "File Import Error");
+    } catch (std::exception &exception) {
+      logger_->log_debug("Caught Exception %s", exception.what());
+      throw;
+    } catch (...) {
+      logger_->log_debug("Caught Exception during process session write");
+      throw;
     }
-  } catch (std::exception &exception) {
-    if (flowFile && flowFile->getResourceClaim() == claim) {
```
phrocker: What was the purpose for removing these? We typically want to decrease ownership count in the face of an exception.

libminifi/include/utils/StringUtils.h
```diff
@@ -234,30 +138,139 @@ class StringUtils {
...
-      return a.first > b.first;
-    });
+  /**
+   * Hexencodes bytes and writes the result to hex
+   * @param hex the output buffer where the hexencoded string will be written. Must be at least length * 2 bytes long.
+   * @param data the bytes to be hexencoded
+   * @param length the length of data. Must not be larger than std::numeric_limits<size_t>::max()
+   * @param uppercase whether the hexencoded string should be upper case
+   */
+  inline static void to_hex(char* hex, const uint8_t* data, size_t length, bool uppercase) {
```
phrocker: I like this piece of code, but what was the impetus for this in this particular PR? I don't see it used in process session or tail file, so I must have missed where it was needed for this PR. This seems like self effacing code. What benefit does this provide over stringstream?

phrocker: When I evaluate necessity it's mostly to know if this is a piece of code that's absolutely needed for 0.7.0 or if it's something that can wait...

extensions/standard-processors/processors/HashContent.h
```diff
@@ -38,21 +38,13 @@
...
 #include "core/Resource.h"
 #include "io/BaseStream.h"
+#include "utils/StringUtils.h"
 
 using HashReturnType = std::pair<std::string, int64_t>;
 
 namespace {
 #define HASH_BUFFER_SIZE 16384
 
-  std::string digestToString(const unsigned char * const digest, size_t size) {
```
phrocker: is there validation to ensure this produces the exact same expected output?

extensions/standard-processors/processors/LogAttribute.cpp
```diff
@@ -137,33 +137,32 @@ void LogAttribute::onTrigger(const std::shared_ptr<core::ProcessContext> &contex
...
+    } else {
+      message << "\n";
     }
-    message << "\n" << dashLine;
+    message << dashLine;
     std::string output = message.str();
 
     switch (level) {
       case LogAttrLevelInfo:
-        logger_->log_info("%s", output);
```
phrocker: let's revert these.

extensions/standard-processors/processors/LogAttribute.h
```diff
@@ -90,26 +90,19 @@ class LogAttribute : public core::Processor {
...
-      ret = stream->read(buffer_, buffer_size_);
-      if (!stream)
-        read_size_ = stream->getSize();
-      else
-        read_size_ = buffer_size_;
-      return ret;
+      if (buffer_.size() == 0U) {
+        return 0U;
+      }
+      int ret = stream->read(buffer_.data(), buffer_.size());
```
phrocker: Am I misunderstanding something? Less data is read from the stream therefore we throw an error? The contract for the streams, I believe, returns the size read and the input size isn't reflective of an error condition.

extensions/standard-processors/tests/integration/TailFileCronTest.cpp
```diff
@@ -73,7 +73,7 @@ class TailFileTestHarness : public IntegrationBase {
   virtual void runAssertions() {
     assert(LogTestController::getInstance().contains("5 flowfiles were received from TailFile input") == true);
     assert(LogTestController::getInstance().contains("Looking for delimiter 0xA") == true);
-    assert(LogTestController::getInstance().contains("li\\ne5") == true);
+    assert(LogTestController::getInstance().contains(utils::StringUtils::to_hex("li\\ne5")) == true);
```
phrocker: Was that test assertion looking for a non-delimited line?

extensions/standard-processors/tests/unit/TailFileTests.cpp
```diff
@@ -105,12 +111,127 @@ TEST_CASE("TailFileWithOutDelimiter", "[tailfiletest2]") {
...
+  // Delete the test and state file.
+  remove(TMP_FILE);
+  remove(STATE_FILE);
+}
+
+TEST_CASE("TailFileLongWithDelimiter", "[tailfiletest2]") {
+  std::string line1("foo");
+  std::string line2(8050, 0);
+  std::mt19937 gen(std::random_device { }());
+  std::generate_n(line2.begin(), line2.size(), [&]() -> char {
```
phrocker: can we not do something like this: https://stackoverflow.com/a/440240 

and add it to the test library negating our need for hex conversion? I may be missing something here.

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -66,7 +66,8 @@ core::Property TailFile::StateFile("State File", "Specifies the file that should
                                    " what data has been ingested so that upon restart NiFi can resume from where it left off",
                                    "TailFileState");
 core::Property TailFile::Delimiter("Input Delimiter", "Specifies the character that should be used for delimiting the data being tailed"
```
phrocker: If a delimiter isn't specified we should consider this an error if we're using a delimiter mode. If size is being used we can have that specified and make this property an error condition. we should separate those ideologies entirely.

extensions/standard-processors/processors/LogAttribute.cpp
```diff
@@ -137,33 +137,32 @@ void LogAttribute::onTrigger(const std::shared_ptr<core::ProcessContext> &contex
...
-      for (unsigned int i = 0, j = 0; i < callback.read_size_; i++) {
-        message << std::hex << callback.buffer_[i];
-        j++;
-        if (j == 80) {
-          message << '\n';
-          j = 0;
-        }
+
+      auto payload_hex = utils::StringUtils::to_hex(callback.buffer_.data(), callback.buffer_.size());
+      for (size_t i = 0; i < payload_hex.size(); i += 80) {
```
phrocker: Isn't this potentially missing a partial line?

extensions/standard-processors/processors/LogAttribute.h
```diff
@@ -90,26 +90,19 @@ class LogAttribute : public core::Processor {
...
+      if (ret != buffer_.size()) {
+        throw std::exception();
+      }
+      return buffer_.size();
     }
-    uint8_t *buffer_;
-    uint64_t buffer_size_;
-    uint64_t read_size_;
+    std::vector<uint8_t> buffer_;
```
phrocker: Thank you for making this change...

libminifi/src/core/ProcessSession.cpp
bakaid: GitHub's diff is really hard to read properly, but this is not a spacing change, it is a new try-catch block.
Along with the change in the catch blocks, this is to eliminate a code duplication. (We log the errors separately for std::exception and other exceptions, then rethrow the exception and decrement the claim refcounter).

libminifi/src/utils/StringUtils.cpp
bakaid: A cpp file was created because constexpr static members were added to the StringUtils class which necessitated it.
Because it was created I made the trivial change of moving the functions I could to the cpp file, but that is not necessary.

libminifi/src/core/ProcessSession.cpp
bakaid: I did not remove it, I have deduplicated it. See the comment above.

libminifi/src/core/ProcessSession.cpp
phrocker: Ah, so the behavior of decrementing ( I think I have a comment elsewhere ) is maintained? if so you can ignore that comment. Thanks!

extensions/standard-processors/processors/HashContent.h
bakaid: Yes, both in exizsting known answer tests in https://github.com/apache/nifi-minifi-cpp/blob/master/extensions/standard-processors/tests/unit/HashContentTest.cpp#L50 and there are unit tests for uppercase hexencoding in this PR.

extensions/standard-processors/tests/unit/TailFileTests.cpp
bakaid: The result of my code is very similar to the code you have linked, as it generates a string of printable characters for ASCII 32 to 126, therefore this particular instance does not necessitate hex conversion.
I will write a global comment explaining the process and rational behind the changes in this PR, hopefully that will clear things up.

libminifi/include/utils/StringUtils.h
```diff
@@ -108,108 +101,19 @@ class StringUtils {
...
-        break;
-      }
-      auto next = std::find_if(curr, end, is_func);
-      result.push_back(std::string(curr, next));
-      curr = next;
-    }
-
-    return result;
-  }
+  static std::vector<std::string> split(const std::string &str, const std::string &delimiter);
```
phrocker: Going back through this I love some of these changes and would prefer they be split into 3 PRs so that we can isolate their risk to separate commits and deal with their priority independently if you don't disagree.

extensions/standard-processors/processors/LogAttribute.cpp
bakaid: Why would it? std::string::substr will take as much as it can, even if the specified length is larger than what remains.
If the string is 40 bytes long, it will run one iteration and create a line of 40 bytes.
If the string is 80 bytes long, it will run one iteration and create a line of 80 bytes.
If the string is 120 bytes long, it will run two iterations, creating a line of 80 bytes, and a line of 40 bytes from offset 80.

phrocker: Sorry I was referring to both interpretations -- but misread the conditional in full view about the max size of the FF and thought it would only log the first MB, but it looks like it only logs flow files < 1 MB  -- which while not clear from the options is at least a sensical line in the sand.

libminifi/src/utils/StringUtils.cpp
phrocker: Ah yeah, totally makes sense, thanks for explaining that!

libminifi/test/pcap-tests/CMakeLists.txt
```diff
@@ -29,17 +29,17 @@ FOREACH(testfile ${PCAP_TESTS})
   	target_include_directories(${testfilename} PRIVATE BEFORE "${CMAKE_BINARY_DIR}/extensions/pcap/pcap++/Dist/header/")
     createTests("${testfilename}")
     if(APPLE)    
-    	target_link_libraries ("${testfilename}" -Wl,-all_load minifi-pcap )
```
phrocker: Good catch! Thanks for making this change!

libminifi/test/unit/StringUtilsTests.cpp
```diff
@@ -93,3 +93,49 @@ TEST_CASE("TestStringUtils::testEnv5", "[test split classname]") {
...
+  REQUIRE(std::string({0x00, 0x01, 0x02, 0x03,
+                       0x04, 0x05, 0x06, 0x07,
+                       0x08, 0x09, 0x0a, 0x0b,
+                       0x0c, 0x0d, 0x0e, 0x0f}) == StringUtils::from_hex("000102030405060708090a0b0c0d0e0f"));
+  REQUIRE(std::string({0x00, 0x01, 0x02, 0x03,
+                       0x04, 0x05, 0x06, 0x07,
+                       0x08, 0x09, 0x0a, 0x0b,
+                       0x0c, 0x0d, 0x0e, 0x0f}) == StringUtils::from_hex("000102030405060708090A0B0C0D0E0F"));
+  try {
+    StringUtils::from_hex("666f6f62617");
```
phrocker: great tests!

libminifi/src/utils/StringUtils.cpp
```diff
@@ -0,0 +1,162 @@
...
+    source_string.replace(lastFound, from_string.size(), to_string);
+    loc = lastFound + to_string.size();
+  }
+  return source_string;
+}
+
+std::string StringUtils::replaceMap(std::string source_string, const std::map<std::string, std::string> &replace_map) {
+  auto result_string = source_string;
+
+  std::vector<std::pair<size_t, std::pair<size_t, std::string>>> replacements;
```
phrocker: this can be shortened and improved. I know you didn't do it ( for all I know I could have done it ), but seeing it here highlights that so thanks!

libminifi/src/utils/StringUtils.cpp
```diff
@@ -0,0 +1,162 @@
...
+        exit(1);
+      case EXCEPT:
+        throw ofr;
+    }
+  }
+
+  return true;
+}
+
+std::string StringUtils::replaceEnvironmentVariables(std::string& original_string) {
```
phrocker: I've always wondered why this is in StringUtils. I probably put it here...but I'm not sure it fits. Making these changes helps highlight this, Thanks!

extensions/standard-processors/tests/unit/TailFileTests.cpp
phrocker: Thank you!

extensions/standard-processors/processors/HashContent.h
phrocker: Thank you!

bakaid:
@phrocker thanks for reviewing this PR.
First of all, I would be more than happy to split this into multiple commits (or PRs), and I am open to postponing merging some of them after 0.7.0, if it is necessary or safer.
Let me explain the rationale behind this PR, and why are these changes included.
First, I created a fix for the buggy behaviour of the delimiting mode of TailFile. As I started to write tests for the fixed behaviour, I used the existing tests as a base, that build a TailFile - LogAttribute pipeline, with LogAttribute configured to log the payload, and using the logged payload to check the delimited content.
Unfortunately, new tests that used long lines (required to test the fixed behaviour of lines longer than 4095 bytes) did not work.
I found that was because LogAttribute was buggy.
First of all, even though it self-imposed a 1 MB limit for FlowFiles to log, in reality it could not log nearly as much, because our formatted logging uses a 1024 byte long buffer to format the log message. Therefore, the content logging is truncated. Hence this fix using the non-formatted logging path of a single c_str, which does not have such limits (and could potentially be a bad idea because of the possibility of a later change which would interpret even standalone strings as printf format strings in the log pipeline, opening the possibility of a printf format string injection, but I will follow up on that later).
The other issue I found was that while LogAttribute tried to hex encode the payload, it failed to do so.
It tried to use
message << std::hex << callback.buffer_[i];
which is ineffective, because uint8_t will be promoted to a char, and will be printed as a char, instead of interpreted as an integer to be printed in its hexadecimal form.
The proper way to hex encode a string using ostreams is
os << std::hex << std::setfill('0') << std::setw(2) << (int) b;
Which can be found in multiple instances of the code, but unfortunately not here.
But I think that the original intent is the correct one. FlowFiles are potentially non-textual, therefore they should be encoded before logging them to avoid logging non-printable characters. (Or it could be a configuration option, or we could detect whether a FlowFile is textual, but as a generic case, hex encoding them is the most straightforward way).
As an added bonus the payload is encoded, therefore there is no possibility of a format string injection attack.
As to why I did not simply duplicate the usual hex encoding code to this instance: I thought about how we could avoid such mistakes in the future, and I thought the best way would be to a create a single, tested hex encoder and decoder utility and use it everywhere. As it was a quick thing to do, I thought it would be worth to do it.
If we do not want to include this in 0.7.0, that is OK, because it is not necessary: I can rewrite the tests to use PutFile and avoid using LogAttribute entirely. However, I do not think it is superfluous: it is fixing a bug, and making it possible to avoid other bugs in the future.

libminifi/src/core/ProcessSession.cpp
phrocker: The view is very difficult to see how things are impacted, thank you for taking the time to explain it above.

phrocker:
@bakaid There are multiple issues there. The preference for larger log statements is to use the LOG* variants 
`logging::LOG_DEBUG(logger_) <<  "Setting " << std::hex <<  max_work_queue_size << " as the max queue size for " << name;
          logging::LOG_DEBUG(logger_) <<  "Setting " << max_work_queue_size << " as the max queue size for " << name;

`
Btw the 1 MB limitation is a NiFi holdover.

extensions/standard-processors/processors/LogAttribute.cpp
bakaid: As I wrote in the global comment, this is a fix for the truncation issue. An another way to solve that would be to either enlarge the log format buffer size, or use a dynamic reallocation approach in it, but this seemed a less disruptive way to fix the bug.

phrocker: okay I'll take a look at that global comment a little closer later. Thanks for taking the time to write that!

extensions/standard-processors/processors/TailFile.cpp
bakaid: I agree, I just did not want to include that in this PR, because that would modify the behaviour of the processor in a way that could not be made backward-compatible with default values.
Currently empty Delimiter means size-based without a size limit, and a non-empty Delimiter means delimiter mode.
If we add a mode property, and make the default of that property the delimiter mode, we would be backward-compatible with the current delimiter mode, but break the non-delimiter mode.
If the default of that property is size-based mode, and the default size is 0 (unlimited), we keep compatibility with the non-delimiter mode, but break compatibility with the delimiter mode.
I think it is a necessary move to make, but I do not know how such a change (that would require configuration changes on update) can be made.

libminifi/include/core/ProcessSession.h
bakaid: You are right, as this change would break API, and should not be made.
However I would prefer to eventually move to this variant, as removing the file in this case does not make much sense (we are reading "lines" and if there is a partial line at the end of the file, it is not ingested, and removing the file would cause data loss). By the way, we do not use this any other way than keeping the file.

bakaid:
@phrocker I didn't know that LOG_DEBUG existed, it is obviously the better solution, I will rewrite the PR to use that.
I will rebase, solve conflicts, clean up the code, test and then submit it.
Thank you!

libminifi/test/pcap-tests/CMakeLists.txt
bakaid: Thanks. The pcap tests no compile, but still not succeed my macOS, but I will open a ticket for that.

libminifi/test/unit/StringUtilsTests.cpp
bakaid: Thank you!

libminifi/test/pcap-tests/CMakeLists.txt
phrocker: Did you run them as root?

bakaid: Yes, I did, it fails on the `assert(LogTestController::getInstance().contains("Accepting ") == true && LogTestController::getInstance().contains("because it matches .*") );` assertion.

phrocker: Not surprised, that processor is mostly abandoned and will hopefully be removed in upcoming releases.

extensions/standard-processors/processors/LogAttribute.h
bakaid: @phrocker To be honest, I am not sure what the contract really is in these cases. I could not find relevant comments or documentation, so I am basing my understanding on the usages I have seen and the implementation of ContentRepositories, ResourceClaim and BaseStream.
What this implementation did before I changed it was the following: it requested the size of the FlowFile and created a ReadCallback with a buffer size equal to the FlowFile's size.
This ReadCallback is then passed to ProcessSession::read which gets a BaseStream from the ContentRepository for the ResourceClaim owned by the FlowFile. It then passes this BaseStream to the ReadCallback's process function where we are currently.
The old implementation then read at most buffer_size_ (so the FlowFile's size) bytes from this stream.
**It then checked whether the stream (which is a shared_ptr) was null, right after it dereferenced it and called a function on it!**
If this code path could have ever been reached (which it couldn't have been, because we would have gotten a SIGSEGV before that), it would have set the read_size_ to the size of the stream, ignoring the return value of the read operation on the stream.
Otherwise (so in every case), it just sets the read_size_ to buffer_size_ (again, ignoring the size returned by read), which is, in fact, the FlowFile's size.
So this ReadCallback could only ever have worked, if we could indeed read all data in one read operation from the stream provided by the ContentRepository. And this is what I think the contract is, and this is how I have seen it used throughout the code. (We assume that in this case read will return all data, and not just partial data).
My implementation just cleans up this logic adds an extra check to see whether this contract is really held true.

bakaid:
@phrocker I have updated the PR as discussed. I would like to run some manual checks before merging though.
Please also see my new comment at LogAttribute::ReadCallback

extensions/standard-processors/processors/LogAttribute.h
phrocker: I don't think that's a correct interpretation of the contract. The null dereference is bad, but my focus is continuation of the same positive path behavior. My opinion is that this should be chunking and checking return size. The fact that the constructor accepts the size doesn't make a lot of sense. We should choose a smaller size and chunk it ( which the contract does support even if it isn't evident ) or be safe and continue reading until -1 or 0 are met.

-1 returning rolls back, but when that exception is caught the same will be done so the change isn't great. The only change is that a partial read ( which the contract doesn't prohibit ). I'll have to think about this general change and the impact and realistic possibility it may occur.

phrocker: Let's leave my comment at this since this is what I mostly care about: I don't mind the change in behavior ( exhibited or theoretical ) -- but can you add a log statement to show how many bytes were read vs requested? 

This processor is used chiefly as a debug tool. Notwithstanding the issues you resolved, if the number of bytes read != stream, I think the user would want to know about that and it may help us triage a potential problem in a repo or on disk. That would mean the flow file size does not match the content size -- and I don't know if that would come across very well even with the code that exists in master.

bakaid: I agree with you that chucking is generally a better solution for this, if for nothing else than for the lower memory requirement. In this case, because we are only dealing with streams no larger than 1 MB, this benefit is not that pronounced.

I think adding the log you mentioned is a good idea, I don't know why I didn't do it, I will do now.

Just to make sure that I have correct understanding: we expect that we will be able to read exactly as many bytes from the stream as flow_file->getSize() returns (let it be in one piece or in chunks). And if that is not the case, it means there is a bug somewhere.

bakaid:
@phrocker I have added new review fixes and ran some manual tests.
Tailing long lines now works, but the processor does not work after restored from its state, but that is a problem on master as well: https://issues.apache.org/jira/browse/MINIFICPP-937

libminifi/include/utils/StringUtils.h
```diff
@@ -347,6 +227,139 @@ class StringUtils {
...
+        if (found_first_nibble) {
+          data[(*data_length)++] = n1 << 4 | n;
+          found_first_nibble = false;
+        } else {
+          n1 = n;
+          found_first_nibble = true;
+        }
+      }
+    }
+    if (found_first_nibble) {
```
arpadboda: Simply return this? :)

libminifi/include/utils/StringUtils.h
```diff
@@ -347,6 +227,139 @@ class StringUtils {
...
+  }
+
+  /**
+   * Hexdecodes a string
+   * @param hex the hexencoded string
+   * @param hex_length the length of hex
+   * @return the vector containing the hexdecoded bytes
+   */
+  inline static std::vector<uint8_t> from_hex(const char* hex, size_t hex_length) {
+    std::vector<uint8_t> decoded(hex_length / 2);
```
arpadboda: Sure?
If hex_length is odd, this buffer becomes too small by the division. 
I know that the scenario is odd, too.

bakaid: If the hex_length is odd (and does not contain any filler characters), then the encoding is invalid (there will be an incomplete byte). In this case the above from_hex function will not assemble and write this byte to the buffer and returns with false instead, so there is no buffer overflow.
There is also a test case specifically for this.

extensions/standard-processors/processors/LogAttribute.h
```diff
@@ -89,27 +89,23 @@ class LogAttribute : public core::Processor {
...
-      else
-        read_size_ = buffer_size_;
-      return ret;
+      if (buffer_.size() == 0U) {
+        return 0U;
+      }
+      int ret = stream->read(buffer_.data(), buffer_.size());
+      if (ret != buffer_.size()) {
+        logger_->log_error("%zu bytes were requested from the stream but %d bytes were read. Rolling back.", buffer_.size(), ret);
+        throw std::exception();
```
arpadboda: Is this the most appropriate exception we can throw?

libminifi/include/core/ProcessSession.h
```diff
@@ -111,6 +111,7 @@ class ProcessSession : public ReferenceContainer {
   // import from the data source.
   void import(std::string source, const std::shared_ptr<core::FlowFile> &flow, bool keepSource = true, uint64_t offset = 0);
   void import(std::string source, std::vector<std::shared_ptr<FlowFileRecord>> &flows, bool keepSource, uint64_t offset, char inputDelimiter);
```
arpadboda: If I remember well this is only used by tailfile.

Removing would be a breaking change, but I guess deprecation is ok as it becomes unreferenced.

libminifi/include/utils/StringUtils.h
arpadboda: Okay, thanks!

libminifi/include/utils/StringUtils.h
bakaid: I think the current version it is more readable than `return !found_first_nibble;`
This way it checks for an error (there is a nibble at the end without a pair), and at the end of function returns with true.
The consolidated return statement would be harder to understand at the first glance.

libminifi/include/core/ProcessSession.h
bakaid: I agree that it should deprecated, but I am not sure what you specifically mean by that. Do we have a process for marking functions deprecated?

extensions/standard-processors/processors/LogAttribute.h
bakaid: Fair point, I have added a better exception.

libminifi/include/core/ProcessSession.h
bakaid: I have added the DEPRECATED macro to MiNiFi.

arpadboda:
Going to merge this later, after CI executions complete and pass.

PR #606, opened 2019-07-04
MINIFICPP-828: Regex Util

nghiaxlee:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/utils/RegexUtils.h
+++ b/libminifi/include/utils/RegexUtils.h
@@ -1,0 +1,84 @@
+/**
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef LIBMINIFI_INCLUDE_IO_REGEXUTILS_H_
+#define LIBMINIFI_INCLUDE_IO_REGEXUTILS_H_
+
+#include <vector>
+#include <regex>
+
+#if (__cplusplus > 201103L) || defined(_WIN32)
+#define NO_MORE_REGFREEE
+#endif
+
+#ifndef NO_MORE_REGFREEE
+#include <regex.h>
+#endif
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace utils {
+
+class Regex {
+public:
+  enum class Mode { ICASE };
+
+  Regex();
+  explicit Regex(const std::string &value);
+  explicit Regex(const std::string &value,
+                const std::vector<Mode> &mode);
+  Regex(const Regex &) = delete;
+  Regex& operator=(const Regex &) = delete;
+  Regex(Regex&& other);
+  Regex& operator=(Regex&& other);
+  ~Regex();
+  bool match(const std::string &pattern);
+  const std::vector<std::string>& getResult() const;
+  const std::string& getSuffix() const;
+
+ private:
+  std::string pat_;
+  std::string suffix_;
+  std::string regexStr_;
+  std::vector<std::string> results_;
+  bool valid_;
+
+#ifdef NO_MORE_REGFREEE
+
+  std::regex compiledRegex_;
+  std::regex_constants::syntax_option_type regex_mode_;
+  std::smatch matches_;
+
+#else
+
+  regex_t compiledRegex_;
+  int regex_mode_;
+  std::vector<regmatch_t> matches_;
+
+#endif
+};
+
+} /* namespace utils */
+} /* namespace minifi */
+} /* namespace nifi */
+} /* namespace apache */
+} /* namespace org */
+
+#endif /* LIBMINIFI_INCLUDE_IO_REGEXUTILS_H_ */
--- a/libminifi/src/utils/RegexUtils.cpp
+++ b/libminifi/src/utils/RegexUtils.cpp
@@ -1,0 +1,165 @@
+/**
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "utils/RegexUtils.h"
+#include "Exception.h"
+#include <iostream>
+#include <vector>
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace utils {
+
+Regex::Regex() : Regex::Regex("") {}
+
+Regex::Regex(const std::string &value) : Regex::Regex(value, {}) {}
+
+Regex::Regex(const std::string &value,
+                           const std::vector<Regex::Mode> &mode)
+    : regexStr_(value),
+      valid_(false) {
+  if (regexStr_.empty())
+    return;
+
+  // Create regex mode
+#ifdef NO_MORE_REGFREEE
+  regex_mode_ = std::regex_constants::ECMAScript;
+#else
+  regex_mode_ = REG_EXTENDED;
+#endif
+  for (const auto m : mode) {
+    switch (m) {
+      case Mode::ICASE:
+#ifdef NO_MORE_REGFREEE
+        regex_mode_ |= std::regex_constants::icase;
+#else
+        regex_mode_ |= REG_ICASE;
+#endif
+        break;
+    }
+  }
+  // Compile
+#ifdef NO_MORE_REGFREEE
+  try {
+    compiledRegex_ = std::regex(regexStr_, regex_mode_);
+    valid_ = true;
+  } catch (const std::regex_error &e) {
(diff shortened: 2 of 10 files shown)


libminifi/include/utils/RegexUtils.h
```diff
@@ -0,0 +1,75 @@
...
+namespace utils {
+
+class RegexUtils {
+public:
+  enum class Mode { EXTENDED, ICASE };
+
+    RegexUtils(const std::string &value,
+                const std::vector<Mode> &mode);
+    ~RegexUtils();
+  bool canMatch(const std::string &pattern);
```
bakaid: I would prefer match() or operator() here, as this functions actually does the matching.

libminifi/include/utils/RegexUtils.h
```diff
@@ -0,0 +1,75 @@
...
+  std::vector<std::string> getResult() const;
+  std::string getSuffix();
```
bakaid: This can be const as well.

libminifi/src/utils/RegexUtils.cpp
```diff
@@ -0,0 +1,128 @@
...
+#include <vector>
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace utils {
+
+RegexUtils::RegexUtils(const std::string &value,
+                           const std::vector<RegexUtils::Mode> &mode)
```
bakaid: I think a `std::set<RegexUtils::Mode>` would be a better choice for mode, as we want this to be a bitmask kind of thing, but vector is fine as well, your call.

libminifi/src/utils/RegexUtils.cpp
```diff
@@ -0,0 +1,128 @@
...
+#endif
+        break;
+    }
+  }
+  // Compile
+#if (__cplusplus > 201103L) || defined(_WIN32)
+  try {
+    compiledRegex = std::regex(regexStr, regex_mode);
+  } catch (const std::regex_error &e) {
+    throw e.what();
```
bakaid: Please consider using our Exception class here (and in the other part of the ifdef as well).

libminifi/src/utils/RegexUtils.cpp
```diff
@@ -0,0 +1,128 @@
...
+    switch (m) {
+      case Mode::EXTENDED:
+#if (__cplusplus > 201103L) || defined(_WIN32)
+        regex_mode |= std::regex_constants::extended;
+#else
+        regex_mode |= REG_EXTENDED;
+#endif
+        break;
+      case Mode::ICASE:
+#if (__cplusplus > 201103L) || defined(_WIN32)
```
arpadboda: Can we avoid copy-pasting this?

Something like
#if (__cplusplus > 201103L) || defined(_WIN32)
#define WE_ARE_SO_HAPPY_TO_USE_STD_REGEX_HAHA
#endif 

and later just rely on this defined value.

libminifi/src/utils/RegexUtils.cpp
```diff
@@ -0,0 +1,128 @@
...
+  try {
+    compiledRegex = std::regex(regexStr, regex_mode);
+  } catch (const std::regex_error &e) {
+    throw e.what();
+  }
+#else
+  int err_code = regcomp(&compiledRegex, regexStr.c_str(), regex_mode);
+  if (err_code) {
+    char msg[100];
```
bakaid: Are you sure that 99 characters are usually enough? You can call regerror with nullptr and 0, and it will return the required buffer size so you can be sure you will be able to get the entire message.

libminifi/src/utils/RegexUtils.cpp
```diff
@@ -0,0 +1,128 @@
...
+    regerror(err_code, &compiledRegex, msg, sizeof(msg));
+    throw std::string(msg);
```
arpadboda: I would prefer dedicated exception type(s) for regex issues.
For the cherry on top function declarations in the header files should indicate the exceptions they might throw.

libminifi/src/utils/RegexUtils.cpp
```diff
@@ -0,0 +1,128 @@
...
+  }
+  int maxGroups = std::count(regexStr.begin(), regexStr.end(), '(') + 1;
+  matches.resize(maxGroups);
+#endif
+}
+
+RegexUtils::~RegexUtils() {
+#if (__cplusplus > 201103L) || defined(_WIN32)
```
bakaid: Please consider rewriting the condition to be able to remove the empty block.

libminifi/src/utils/RegexUtils.cpp
```diff
@@ -0,0 +1,128 @@
...
+    } else {
+      suffix = pattern.substr(matches[0].rm_eo + 1);
+    }
+    return true;
+  }
+  return false;
+#endif
+}
+
+std::vector<std::string> RegexUtils::getResult() const { return results; }
```
arpadboda: Please return const &.
The user copies if he wants to modify or store the result.

libminifi/src/utils/RegexUtils.cpp
```diff
@@ -0,0 +1,128 @@
...
+#if (__cplusplus > 201103L) || defined(_WIN32)
+
+#else
+  regfree(&compiledRegex);
+#endif
+}
+
+bool RegexUtils::canMatch(const std::string &pattern) {
+  results.clear();
+  if (pat == pattern) {
```
bakaid: I think it is a possible and reasonable usage of this class to call it with the same string multiple times. In this case it can be a good optimization not to rematch the string, but then we should return with the result of the previous match. With the current implementation subsequent matches of the same string will be unsuccessful.

libminifi/src/utils/RegexUtils.cpp
```diff
@@ -0,0 +1,128 @@
...
+    }
+    return true;
+  }
+  return false;
+#endif
+}
+
+std::vector<std::string> RegexUtils::getResult() const { return results; }
+
+std::string RegexUtils::getSuffix() { return suffix; }
```
arpadboda: In this case, too!

libminifi/src/utils/RegexUtils.cpp
bakaid: I think this can return with a const std::vector<std::string>& to prevent always copying the results.

libminifi/test/unit/RegexUtilsTests.cpp
```diff
@@ -0,0 +1,59 @@
...
+using org::apache::nifi::minifi::utils::RegexUtils;
+
+TEST_CASE("TestRegexUtils::single_match", "[regex1]") {
+    std::string pat = "Speed limit 130 | Speed limit 80";
+    std::string rgx1 = "Speed limit ([0-9]+)";
+    std::vector<RegexUtils::Mode> mode = {RegexUtils::Mode::EXTENDED, RegexUtils::Mode::ICASE};
+    RegexUtils r1(rgx1, mode);
+    REQUIRE(r1.canMatch(pat));
+    auto ret = r1.getResult();
+    std::vector<std::string> ans;
```
arpadboda: In C++11 you can init this as
```
std::vector<std::string> ans = {"Speed limit 130", "130"};
```

libminifi/include/utils/RegexUtils.h
```diff
@@ -0,0 +1,75 @@
...
+namespace nifi {
+namespace minifi {
+namespace utils {
+
+class RegexUtils {
+public:
+  enum class Mode { EXTENDED, ICASE };
+
+    RegexUtils(const std::string &value,
+                const std::vector<Mode> &mode);
```
arpadboda: Can we add a a default value for mode? If I remember well both underlying regex utils has a default mode.

libminifi/include/utils/RegexUtils.h
arpadboda: const, please!

libminifi/test/unit/RegexUtilsTests.cpp
```diff
@@ -0,0 +1,59 @@
...
+    RegexUtils r2(rgx2, mode);
+    std::terminate();
+  } catch (const std::string &msg) {
+#if (__cplusplus > 201103L) || defined(_WIN32)
+    REQUIRE("The expression contained mismatched [ and ]." == msg);
+#else
+    REQUIRE("brackets ([ ]) not balanced" == msg);
+#endif
+  }
+}
```
arpadboda: This tests are cool, however I would prefer to have some tests on edgecases, like one that verifies icase mode, check empty inputs (either the string or the regex itself).

libminifi/src/utils/RegexUtils.cpp
nghiaxlee: I agree that set would be a better choice, though I will leave it like this.

libminifi/src/utils/RegexUtils.cpp
nghiaxlee: Now the hard part of this ticket is to come up with a good name :((

libminifi/test/unit/RegexUtilsTests.cpp
nghiaxlee: lol, I don't know why I didn't use that albeit I already used for initializing regex mode vector

libminifi/include/utils/RegexUtils.h
nghiaxlee: I will leave extended mode as default.

libminifi/test/unit/RegexUtilsTests.cpp
```diff
@@ -0,0 +1,80 @@
...
+TEST_CASE("TestRegexUtils::empty_input", "[regex3]") {
+  std::string pat = "";
+  std::string rgx1 = "Speed limit ([0-9]+)";
+  std::string rgx2 = "";
+  std::vector<RegexUtils::Mode> mode = {RegexUtils::Mode::ICASE};
+  RegexUtils r1(rgx1, mode);
+  REQUIRE(!r1.match(pat));
+  try {
+    RegexUtils r2(rgx2, mode);
+    std::terminate();
```
arpadboda: REQUIRE_THROWS might be a bit more elegant to handle this :)

libminifi/include/utils/RegexUtils.h
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+#ifndef LIBMINIFI_INCLUDE_IO_REGEXUTILS_H_
+#define LIBMINIFI_INCLUDE_IO_REGEXUTILS_H_
+
+#include <vector>
+#include <regex>
+
+#if (__cplusplus > 201103L) || defined(_WIN32)
+#define NO_MORE_REGFREEE
```
arpadboda: 👍 🤣

libminifi/src/utils/RegexUtils.cpp
```diff
@@ -0,0 +1,168 @@
...
+  regexStr = std::move(other.regexStr);
+  results = std::move(other.results);
+  valid = other.valid;
+  other.valid = false;
+#ifdef NO_MORE_REGFREEE
+  compiledRegex = std::move(other.compiledRegex);
+  regex_mode = other.regex_mode;
+  matches = std::move(other.matches);
+#else
+  if (valid)
```
bakaid: This should depend on the original valid value, but it depends on other's valid value.

PR #12, opened 2016-09-19
Minifi 109

benqiu2016:

PR #610, opened 2019-07-12
MINIFICPP-814 - Fixed ListenHTTP and HTTPClient bugs, created tests f…

bakaid:
…or ListenHTTP

This fixes the following ListenHTTP issues:
 - HTTPS did not work
 - SSL Minimum Version enforcement caused hanging client and leaked fds in the server
 - HEAD requests were not served properly
 - random port did not work with HTTPS

This fixes the following HTTPClient issues:
 - PUT did not work
 - HEAD did not work
 - POST using setPostFields was a potential use-after-free
 - cURL SSL backend was erroneously deduced
 - header parsing was faulty

This adds the following features:
 - logging for CivetWeb
 - logging for cURL
 - tests for ListenHTTP

### For all changes:
     in the commit message?

### For code changes:

### Note:


bootstrap.sh
```diff
@@ -468,13 +468,6 @@ build_cmake_command(){
 
   add_os_flags
 
-  curl -V | grep OpenSSL &> /dev/null
```
bakaid: This was a fundamentally flawed way to try to determine the SSL backend used by curl and make decisions based on this information for the following reasons:
* What SSL backend the command line curl tool on the build machine is built with, bears practically no relevance to what the curl library that ends up invoked in the code is built with.
  * First, while unlikely, the command line tool on the system might not use the libcurl on the system that we will link with.
  * Second, it is not at all guaranteed that we will link with the system libcurl - we might build our own with USE_SYSTEM_CURL=OFF.
  * Thirdly, even if we do dynamically link the system libcurl, the libcurl on the target system, where the depencency will be loaded might be built with an another SSL backend.

Therefore, if we want to determine what SSL backend the libcurl we end up using has, it must be done at runtime.
While this would be possible - even if not elegant - to do by invoking `curl_version_info`, I propose that we should not even try to do this at runtime.

The only purpose we used USE_CURL_NSS for, was to determine whether we should set the SSL information through curl settings or do it directly by interacting with the SSL_CTX.
Even if we somehow determined that our curl used OpenSSL, so that we can use CURLOPT_SSL_CTX_FUNCTION, it is not a good idea to do it.
It is not at all guaranteed that the OpenSSL we are linked with is the same one curl is built with. And if that's the case, then curl will create the SSL_CTX with its own OpenSSL version that we will try to manipulate with an another OpenSSL version, possibly crashing the whole thing.

The only configuration this could work with, is if we are linked with the system libcurl, system OpenSSL, and the system libcurl is linked with the same system OpenSSL. While this is a common configuration, it only represents one configuration out of the many we can be built with.

Additionally, because I see no benefit (in functionality or ease of use), or certainly not one that would outweigh the risks, to use the SSLContextService's `configure_ssl_context` to set up curl's SSL instead of extracting the information from the SSLContextService and passing it properly to curl, and there are many ways in which we can fail if we try to do it, I have removed that path from the code.

extensions/civetweb/processors/ListenHTTP.cpp
```diff
@@ -390,7 +394,25 @@ bool ListenHTTP::Handler::handleGet(CivetServer *server, struct mg_connection *c
   return true;
 }
 
-void ListenHTTP::Handler::write_body(mg_connection *conn, const mg_request_info *req_info) {
+bool ListenHTTP::Handler::handleHead(CivetServer *server, struct mg_connection *conn) {
```
bakaid: Our processor documentation says we serve HEAD requests, and NiFi does it at well, so we should too.

extensions/http-curl/client/HTTPClient.cpp
```diff
@@ -120,8 +120,21 @@ void HTTPClient::forceClose() {
 
 }
 
+int HTTPClient::debug_callback(CURL *handle, curl_infotype type, char *data, size_t size, void *userptr) {
```
bakaid: I think its much better to log this properly when setVerbose is used, instead of it just going to stderr, but if there is a reason for that I would be willing to discuss this.

extensions/http-curl/client/HTTPClient.cpp
```diff
@@ -196,7 +255,7 @@ std::string HTTPClient::escape(std::string string_to_escape) {
 
 void HTTPClient::setPostFields(std::string input) {
   curl_easy_setopt(http_session_, CURLOPT_POSTFIELDSIZE, input.length());
-  curl_easy_setopt(http_session_, CURLOPT_POSTFIELDS, input.c_str());
+  curl_easy_setopt(http_session_, CURLOPT_COPYPOSTFIELDS, input.c_str());
```
bakaid: CURLOPT_POSTFIELDS does NOT copy the data (https://curl.haxx.se/libcurl/c/CURLOPT_POSTFIELDS.html), so using a temporary std::string's c_str here is a use-after free which might be mitigated in some cases by optimization, but has the potential to leak memory to remote servers.

extensions/http-curl/client/HTTPClient.cpp
```diff
@@ -327,7 +356,9 @@ void HTTPClient::set_request_method(const std::string method) {
   if (my_method == "POST") {
     curl_easy_setopt(http_session_, CURLOPT_POST, 1L);
   } else if (my_method == "PUT") {
-    curl_easy_setopt(http_session_, CURLOPT_PUT, 0L);
+    curl_easy_setopt(http_session_, CURLOPT_UPLOAD, 1L);
```
bakaid: CURLOPT_PUT was set to 0 in this case, which means do not use PUT, but CURLOPT_PUT has been deprecated a long time ago anyway.

extensions/http-curl/client/HTTPClient.cpp
```diff
@@ -327,7 +356,9 @@ void HTTPClient::set_request_method(const std::string method) {
...
+  } else if (my_method == "HEAD") {
+    curl_easy_setopt(http_session_, CURLOPT_NOBODY, 1L);
```
bakaid: Just passing "HEAD" as the CURLOPT_CUSTOMREQUEST will not work, because curl will expect a response body. CURLOPT_NOBODY is the proper way to do it.

extensions/http-curl/client/HTTPClient.cpp
```diff
@@ -346,38 +377,38 @@ bool HTTPClient::matches(const std::string &value, const std::string &sregex) {
...
-  curl_easy_setopt(http_session, CURLOPT_SSL_CTX_DATA, static_cast<void*>(ssl_context_service_.get()));
-  curl_easy_setopt(http_session, CURLOPT_CAINFO, 0);
-  curl_easy_setopt(http_session, CURLOPT_CAPATH, 0);
-#endif
+  curl_easy_setopt(http_session, CURLOPT_CAPATH, nullptr);
 }
 
 bool HTTPClient::isSecure(const std::string &url) {
-  if (url.find("https") != std::string::npos) {
+  if (url.find("https") == 0U) {
```
bakaid: http://https.totallysafe.ru is not a secure URL.

extensions/http-curl/client/HTTPClient.cpp
```diff
@@ -265,36 +324,6 @@ bool HTTPClient::submit() {
   }
 
   logger_->log_debug("Finished with %s", url_);
-  std::string key = "";
```
bakaid: This logic was partial, standard non-compliant and buggy. (The concrete issue was that it included extra \r\n's in the last header value, but there are more problems with it). See the reimplementation below.

thirdparty/civetweb-1.10/src/civetweb.c
```diff
@@ -16237,10 +16237,13 @@ worker_thread_run(struct worker_thread_args *thread_args)
 					mg_free(conn->request_info.client_cert);
 					conn->request_info.client_cert = 0;
 				}
-			}
+			} else {
```
bakaid: Fixed in https://github.com/civetweb/civetweb/pull/630 and released in civetweb 1.11, but that is a rather large rewrite and after all I think just having this patch carries less risk than updating the whole thing.

bakaid:
I have a few suspicions about chunked encoding, I will take a look at that.
Also, I ran a manual Windows build, but obviously with a different configuration, because it failed in AppVeyor. I will take a look at that as well.

libminifi/test/resources/TestHTTPGetSecure.yml
```diff
@@ -77,11 +77,11 @@ Controller Services:
       class: SSLContextService
       Properties:
         Client Certificate:
-            - value: cn.crt.pem
+            - value: cn.p12
```
bakaid: This fixes the test failure that happened on some mac versions, so the test could be reenabled.

extensions/civetweb/processors/ListenHTTP.h
```diff
@@ -163,10 +166,47 @@ class ListenHTTP : public core::Processor {
...
+      CivetServer* server = static_cast<CivetServer*>(mg_get_user_data(ctx));
+      if (server == nullptr) {
+        return 0;
+      }
+      std::shared_ptr<logging::Logger>* logger = static_cast<std::shared_ptr<logging::Logger>*>(const_cast<void*>(server->getUserContext()));
+      if (logger == nullptr) {
+        return 0;
+      }
+      logging::LOG_ERROR((*logger)) << "CivetWeb error: " << message;
+    } catch (...) {
```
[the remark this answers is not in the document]
bakaid: This is just best practice for C callbacks. Even if the logger currently does not throw, it might do so in the future, and then we have a nice crash on our hands. I think it is better to be defensive here.

extensions/civetweb/tests/ListenHTTPTests.cpp
```diff
@@ -0,0 +1,514 @@
...
+
+    auto raw_ptr = dynamic_cast<org::apache::nifi::minifi::processors::ListenHTTP*>(listen_http.get());
+    std::string protocol = std::string("http") + (raw_ptr->isSecure() ? "s" : "");
+    std::string portstr = raw_ptr->getPort();
+    REQUIRE(LogTestController::getInstance().contains("Listening on port " + portstr));
+
+    url = protocol + "://localhost:" + portstr + "/contentListener/" + endpoint;
+  }
+
+  void test_connect(bool should_succeed = true, int64_t response_code = 200) {
```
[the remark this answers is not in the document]
bakaid: The naming might be misleading, but should_succeed means that we get an HTTP response (and not for example, die in the TLS connection phase because of a bad cert) and the response_code means the error code we get. Because our processor specification includes specific error codes for specific cases (403 on AuthorizedDNPattern mismatch, 405 on unsupported HTTP methods), I want to specifically test for those.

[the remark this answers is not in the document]
arpadboda: Okay, makes sense, thanks!

thirdparty/civetweb-1.10/src/civetweb.c
bakaid: AgreedL https://issues.apache.org/jira/browse/MINIFICPP-966

extensions/civetweb/tests/resources/generate.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
bakaid: The script is included for reproducibility and for easier understanding of how the certs were created (it could be deduced from looking at the PEM files, but that would take quite some time).
I agree that at least parts of this code (the functions) could be made into a generic tool (which this script can then include), but the certs generated here are specific to this test set.
However, I am not sure where we should place such code and whether we would really use it.
I think this might be worth revisiting if the need for test certs arise in the future.

extensions/civetweb/tests/ListenHTTPTests.cpp
```diff
@@ -0,0 +1,514 @@
...
+#include "processors/ListenHTTP.h"
+#include "client/HTTPClient.h"
+#include "controllers/SSLContextService.h"
+#include "properties/Configure.h"
+
+class ListenHTTPTestsFixture {
+ public:
+  ListenHTTPTestsFixture()
+   : tmp_dir(strdup("/tmp/gt.XXXXXX")) {
+    LogTestController::getInstance().setTrace<TestPlan>();
```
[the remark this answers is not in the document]
bakaid: No, I think this is just a remnant of some debugging, I will remove it.

libminifi/include/utils/HTTPClient.h
```diff
@@ -66,37 +69,89 @@ struct HTTPReadCallback {
...
  public:
-
   HTTPHeaderResponse(int max)
-      : max_tokens_(max) {
+   : max_tokens_(max)
+   , parsed(false) {
   }
 
-  void append(const std::string &header) {
+  DEPRECATED void append(const std::string &header) {
```
phrocker: Can  you add a comment on when this was deprecated so that we can do it in a timely manner since we are coming to the close of 0.x we should probably remove this after 1.0.

extensions/http-curl/client/HTTPClient.cpp
phrocker: I don't think HEAD was ever completed in the client -- this is not a fully complete client and I was hoping we'd switch to https://github.com/jpbarrette/curlpp, where we simply wrapped their code in our HTTPClient. As per the comment, above, do you think this is advantageous? My hope was that we'd do that for 1.0, but if there is consensus maybe we should use their client sooner, especially since cycles to fix bugs ( known and not known ) in our http client seem wasteful. Wrapping curlpp may be ultimately the best solutions.

extensions/http-curl/client/HTTPClient.cpp
phrocker: The more I think about it, I created https://issues.apache.org/jira/browse/MINIFICPP-981. Would love to hear your thoughts on that. curl bugs because we moved fast and loose with a limited number of devs means we're going to have more bugs. Even if people come around and locate them we still have a lot of risk because CURL has a large and constantly evolving API. curlpp has dedicated their project to an http client and I'd love to stop the maintenance of that and instead only wrap it.

bootstrap.sh
bakaid: If we build both libcurl and libressl ourselves, then the risks I outlined would be mitigated, because we can be sure of the SSL backend libcurl uses and we can be sure that we use the exact same SSL library curl does, so we will be able to use configure_ssl_context.
However, by doing this we can not avoid changes entirely.
This current method of determining whether we should use configure_ssl_context (USE_CURL_NSS) would still have to be changed - if we limit ourselves to a statically linked libressl and libcurl, then the answer is we should always use that and we don't really need the curl setopt way of doing it.
If the main risk of change we want to avoid is switching to the curl setopt method of configuring SSL where formerly a configure_ssl_context has been used, then yes, doing this would mitigate that risk of change as well.
While I think this would be a good solution for now, in the long term I would actually like us to use the curl setopt method - it is more versatile and does not limit us to this configuration. But I do understand that to mitigate the risk of doing that we would have to do extensive testing.

extensions/http-curl/client/HTTPClient.cpp
bakaid: I understand that this is not a complete client, and that HEAD requests have not been previously used in the code. However, to test the HEAD functionality of ListenHTTP it has to be added now.
I think it would be best to discuss curlpp in the issue you created. I have not used it, and in general just adding another layer of API which we can be use either properly or badly is not a solution in itself - I will have to review the project to decide for myself whether it would have significant advantages over just using the curl API.
What I am sure I don't want to do is introduce a new, untested thirdparty dependency in 0.6.1, and as this is a very trivial fix in the client, I don't think it warrants that.

libminifi/include/utils/HTTPClient.h
bakaid: Sure, will do.

extensions/http-curl/client/HTTPClient.cpp
bakaid: @phrocker Actually we can't assume that std::strings are COW (indeed they explicitly can't be anymore since C++11). This means that if SSO is not in effect, then the heap containing the string will be copied and then freed when the temporary copy of the string setPostFields is called with is destructed, immediately resulting in a heap use-after-free.
In the specific case if the string is short enough, meaning SSO is in effect, and the calling function calls both setPostFields and then submit, meaning that even though destructed, the temporary copy will still be on stack, then we avert a user-after free. If the HTTPClient object is stored beyond that, which is a usual case, then even this case will result in stack-use-after-return or stack-buffer-overflow.

On the issue of re-copying on every submission: I don't see evidence of that. Once we set the CURLOPT_COPYPOSTFIELDS, the easy handle itself will store that string until we set a new one, or call curl_easy_reset or curl_easy_cleanup the handle. Multiple curl_easy_performs (so multiple submits) do no change that. So our internal copy of the post string is already there: it is in the easy handle.

This function could be optimized if our parameter was const& (we would avoid one copy): I am not sure if changing that counts as an API change, if not, I would be happy to do it.

arpadboda: I think changing the API from string to const& is a "compatible" change, given it just provides further guarantee for the caller, but doesn't break anything that worked before.

extensions/http-curl/client/HTTPClient.cpp
bakaid: @phrocker That is a good idea, but for maintaining API-compatibility we will have to choose a default value for that flag. I prefer the default value to be the logger mode. What do you think?

extensions/http-curl/client/HTTPClient.cpp
```diff
@@ -148,6 +161,52 @@ void HTTPClient::setDisableHostVerification() {
   curl_easy_setopt(http_session_, CURLOPT_SSL_VERIFYHOST, 0L);
 }
 
+bool HTTPClient::setSpecificSSLVersion(SSLVersion specific_version) {
+#if CURL_AT_LEAST_VERSION(7, 54, 0)
+  CURLcode ret = CURLE_UNKNOWN_OPTION;
+  switch (specific_version) {
+    case SSLVersion::SSLv2:
+      ret = curl_easy_setopt(http_session_, CURLOPT_SSLVERSION, CURL_SSLVERSION_SSLv2);
```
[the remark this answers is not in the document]
bakaid: This function (setting the only SSL/TLS version we are willing to support) is really only for testing the ListenHTTP processor's "SSL Minimum Version" function. I would be more than happy not to support SSLv2 and SSLv3 at all, and even the test does not need that, because I can do it with different TLS versions.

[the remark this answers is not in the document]
thenatog: Correct. NiFi currently only supports TLS v1.2 for incoming connections and has backward compatibility to use TLS v1.0 and v1.1 for outgoing connections to support legacy systems. We recommend taking this approach.

This page notes the changes and when they occurred:
https://cwiki.apache.org/confluence/display/NIFI/Migration+Guidance

phrocker:
@bakaid I made some changes in a prior PR that collided with this. one of them, a test fails on OSX. I think you noted that using the P12s worked better on travis? You don't have to fix those tests. I'll fix them as a follow on once this is merged, but since it's only really a dev platform ( and passes locally ) I think we can wait until this is merged. Thanks!

bakaid:
@phrocker Some macOS versions' built-in curl uses SecureTransport as an SSL backend, and some versions of that is only willing to accept private keys in PKCS #12 format with a non-empty password. So switching to that format everywhere fixes the tests.

extensions/http-curl/client/HTTPClient.cpp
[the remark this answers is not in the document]
bakaid: I am changing this to exclude SSLv2 and SSLv3, so only outgoing connections with TLSv1.0, TLSv1.1 and TLSv1.2 will be supported.
However, ListenHTTP's SSLMinimumVersion property (and the underlying implementation) supports every version from SSLv2 to TLSv1.2. While I completely agree with the approach of only allowing TLSv1.2 for incoming connections, this would mean an API change and a deprecation, which I am quite sure can not be made in 0.6.1 and not sure how it can be made in 0.7.0.
@phrocker What are your thoughts on this?

bakaid:
I have added https://issues.apache.org/jira/browse/MINIFICPP-1005 as a blocker follow-up task about TLS versions. I think it's better not to include that in this PR, so that it will be easier to backport to 0.6.0 for 0.6.1.

phrocker:
@bakaid  will take another look and merge after some other test prs are merged. thanks!

extensions/http-curl/client/HTTPClient.cpp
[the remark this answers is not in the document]
thenatog: @alopresto, what are your opinions?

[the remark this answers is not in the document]
bakaid: @alopresto @thenatog Since then we talked about this with @phrocker and agreed that the full modifications recommended could be part of 0.7.0. I have already created a follow-up ticket for that: https://issues.apache.org/jira/browse/MINIFICPP-1005

bakaid:
@phrocker No problem, will rebase and test tomorrow.

PR #612, opened 2019-07-16
MINIFICPP-854: Capture RTSP Frame

nghiaxlee:

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/test/opencv-tests/CMakeLists.txt
```diff
@@ -24,14 +24,14 @@ SET(OPENCV_TEST_COUNT 0)
...
 	add_executable("${testfilename}" "${testfile}")
-	createTests("${testfilename}")
 	target_include_directories(${testfilename} PRIVATE BEFORE "${CMAKE_SOURCE_DIR}/extensions/opencv/")
-
+	target_include_directories(${testfilename} PRIVATE BEFORE "${CMAKE_SOURCE_DIR}/extensions/standard-processors")
 	if (APPLE)
-		target_link_libraries (${testfilename} -Wl,-all_load minifi-opencv)
+		target_link_libraries (${testfilename} -Wl,-all_load minifi-opencv minifi-standard-processors)
 	else ()
-		target_link_libraries (${testfilename} -Wl,--whole-archive minifi-opencv -Wl,--no-whole-archive)
```
arpadboda: Could you to this the "standard" way we do this? (Just setting variables in conditions to match platform and have only one linker command.

libminifi/test/opencv-tests/CMakeLists.txt
```diff
@@ -24,14 +24,14 @@ SET(OPENCV_TEST_COUNT 0)
 FOREACH(testfile ${OPENCV_TESTS})
```
arpadboda: I think opencv tests should be moved to the opencv extension as opencv tests hardly make sense without the processor.

extensions/opencv/CaptureRTSPFrame.cpp
```diff
@@ -116,17 +118,27 @@ void CaptureRTSPFrame::onSchedule(core::ProcessContext *context, core::ProcessSe
...
-  cv::Mat frame;
+  try {
+    video_capture_.open(rtsp_url_);
+    video_backend_driver_ = video_capture_.getBackendName();
+  } catch (const cv::Exception &e) {
+    logger_->log_error("Unable to open RTSP stream: %s", e.what());
+    session->transfer(flow_file, Failure);
+  } catch (...) {
+    logger_->log_error("Unable to open RTSP stream: unhandled exception");
+    session->transfer(flow_file, Failure);
```
arpadboda: I think we should return in failure cases, capture would fail anyway.

extensions/opencv/CaptureRTSPFrame.cpp
```diff
@@ -73,6 +73,8 @@ void CaptureRTSPFrame::initialize() {
   relationships.insert(Success);
   relationships.insert(Failure);
   setSupportedRelationships(std::move(relationships));
+
+  setenv("OPENCV_FFMPEG_CAPTURE_OPTIONS", "rtsp_transport;udp", 1);
```
arpadboda: Could you place a comment here to indicate why setting an env var is necessary?

libminifi/test/opencv-tests/CaptureRTSPFrameTest.cpp
```diff
@@ -16,27 +16,29 @@
...
+#include <fstream>
 #include <utility>
 #include <string>
 #include <set>
-#include "FlowController.h"
-#include "../TestBase.h"
-#include "core/Core.h"
+#include <uuid/uuid.h>
+#include <iostream>
+#include <opencv2/opencv.hpp>
```
arpadboda: Why do we need the test to directly include opencv headers?

nghiaxlee: I am thinking about testing flowfile content later. Does it make sense?

arpadboda: Yes, but we can include it that time. 

I think testing the content only makes sense in case we are the ones producing the content (so we mock the streaming soruce), which is a bigger task to do, keeping one line here doesn't help too much in that. :)

extensions/opencv/CaptureRTSPFrame.cpp
```diff
@@ -73,6 +73,11 @@ void CaptureRTSPFrame::initialize() {
   relationships.insert(Success);
   relationships.insert(Failure);
   setSupportedRelationships(std::move(relationships));
+  // By default in OpenCV, ffmpeg capture is hardcoded to use TCP and this is a workaround
+  // also if UDP timeout, ffmpeg will retry with TCP
+  // Note:
+  // 1. OpenCV community are trying to find a better approach than setenv.
+  setenv("OPENCV_FFMPEG_CAPTURE_OPTIONS", "rtsp_transport;udp", 1);
```
bakaid: setenv is unfortunately not thread-safe. Not only setting the same environment variable from two threads is unsafe, but so is setting anything from two threads at once.
That being said, I know no good solution to this other than a framework for the global synchronization of setenvs and getenvs, and even that does not account from potential setenvs/getenvs called by third parties.
I don't think that there is a high probability of this causing issues, but if we see unexplainable crashes or issues around environment variables, then we should think of this as a cause.
I would like this comment to be extended with this caveat.

extensions/opencv/CaptureRTSPFrame.cpp
```diff
@@ -155,12 +172,12 @@ void CaptureRTSPFrame::onTrigger(const std::shared_ptr<core::ProcessContext> &co
   }
 
   frame.release();
-
 }
 
 void CaptureRTSPFrame::notifyStop() {
   // Release the Capture reference and free up resources.
   video_capture_.release();
+  unsetenv("OPENCV_FFMPEG_CAPTURE_OPTIONS");
```
bakaid: The setenv is in initialize(), which is not the logical pair for notifyStop. When the processor is restarted with onSchedule, the env will not be set again. Either we should not unset this, or move the setenv to onSchedule.

extensions/opencv/tests/CaptureRTSPFrameTest.cpp
```diff
@@ -45,28 +46,39 @@ TEST_CASE("CaptureRTSPFrame", "[opencvtest1]") {
...
 
-    plan->reset();
-    testController.runSession(plan, false);
+  std::shared_ptr<TestPlan> plan = testController.createPlan();
+  std::shared_ptr<core::Processor> captureRTSP = plan->addProcessor("CaptureRTSPFrame", "CaptureRTSPFrame");
 
-    records = plan->getProvenanceRecords();
-    record = plan->getCurrentFlowFile();
-    testController.runSession(plan, false);
+  plan->setProperty(captureRTSP, minifi::processors::CaptureRTSPFrame::RTSPHostname.getName(), "170.93.143.139");
```
bakaid: Could you add a comment here with an example command/description how to start an RTSP server that could be used with this test?

extensions/opencv/CaptureRTSPFrame.cpp
bakaid: @nghiaxlee Now that I think of this, we should not unset this at all. If there are two instances of this processor running, and one is stopped through C2, it will unset it for the other processor as well which could cause issues.

extensions/opencv/CaptureRTSPFrame.cpp
```diff
@@ -116,17 +121,28 @@ void CaptureRTSPFrame::onSchedule(core::ProcessContext *context, core::ProcessSe
...
                                  const std::shared_ptr<core::ProcessSession> &session) {
   auto flow_file = session->create();
 
-  cv::Mat frame;
+  try {
+    video_capture_.open(rtsp_url_);
+    video_backend_driver_ = video_capture_.getBackendName();
+  } catch (const cv::Exception &e) {
+    logger_->log_error("Unable to open RTSP stream: %s", e.what());
+    session->transfer(flow_file, Failure);
```
phrocker: This creates an empty FF, so that means if we can't open the URL we could potentially create many flow files that are empty. The typical approach is to yield in the context. You can still create a failure flow file ( though I would suggest exploring if this is the correct route ), but you should probably yield here.

extensions/opencv/CaptureRTSPFrame.cpp
phrocker: Since this is an env option why not leave it in the operating system environment ( esp since this isn't cross platform compatible ). Can add something to minifi.sh . I have a feeling at some point we will have OPTS set via the conf dir that minifi.sh pulls and adds to a running instance. 

@bakaid this may be a case for using the Initializer construct you added if the shell script doesn't make sense

bakaid: @phrocker Yes, we have ended up with adding an ObjectFactoryInitializer for the OpenCV component with @nghiaxlee, but in person, so we didn't follow up here. That should work without problems for now.
Adding it to minifi.sh could be a good solution if there is a framework for it like you mentioned, but I am not sure everyone uses minifi.sh at the moment (I usually just start the binary).

phrocker: Instructions guide toward minifi.sh so that is the officially supported solution. I'll take a look at the factory initializer a little later. If there is a reason it can't be done in the shell scripts then I'm +1 to that, otherwise I'm a -1 to it.

bakaid: @phrocker I see your point, but I am a little afraid of adding functionality to the minifi.sh. When debugging, for example, you do not start to debug the shell running the minifi, but start debugging the minifi binary itself. If you do that, and the minifi.sh does not set the environment variables, it will have a different behaviour. I know that this is avoidable, but I am sure that more than one people would make this mistake.
Setting (and getting) environment variables from inside the process is perfectly supported, but not thread safe. Having the setenv in an ObjectFactoryInitializer, when the application is single threaded and the ObjectFactoryInitializers are called in sequence should not be a problem.
I do not think either solution is bad, but I feel that the ObjectFactoryInitializer would be better.

arpadboda:
@phrocker : honestly I'm against placing the env var to the minifi.sh script. First I don't think we highlight that the script is the only supported way to start the binary, but Nanofi is a more serious concern at the moment: you can configure MiNiFi processors to the flowchain, which would break in such case. It's an edgecase, but still, valid at the moment.

phrocker:
@arpadboda  https://github.com/apache/nifi-minifi-cpp#running  Running it in other ways would constitute a response that we advise running it in that way. Happy to see a PR if you think that should be clarified.

I don't want to preclude others from running it a way they see fit, but the processor is a specialized case. The documentation that @nghiaxlee writes here should make it clear we expect that environment variable set. It is not the onus of the main readme, but rather in his processor documentation. That also highlights the point that the processors.md should be updated, @nghiaxlee since you have an exceptional case that warrants commenting.

The point about NaNoFi is a good point. https://issues.apache.org/jira/browse/MINIFICPP-742 should clarify some of this. NaNoFi is a library not a set of processors, so "configuring MiNiFi processors" has the same issue that it must be clearly documented. If an implementation arises that uses OpenCV, then that implementation is free to set the environment variable; however, my last comment was to explore other options.

I've presented an option : using the shell script. Happy to hear other options to address that concern, but I currently use OpenCV in such a way with an http rtsp_transport on my local network network. My scripts ( which do a lot more with TF ) use an exported environment variable. 

The alternative is to check and not override. That may be safe -- but I think that is up to @nghiaxlee to evaluate and do a risk assessment. OpenCV is commonly used with python processors. Will this initializer also ensure that anything set for those instances isn't also impacting/overriding this in a thread safe way?

extensions/opencv/CaptureRTSPFrame.cpp
phrocker: @bakaid brought up a good point. We should update https://github.com/apache/nifi-minifi-cpp/blob/master/PROCESSORS.md as there are exceptional conditions, @nghiaxlee . Since you didn't create the processor will you need help generating this? Happy to help.

arpadboda:
@phrocker : according to the documentation we recommend (as we don't provide any other options) the script to start MiNiFi, but don't highlight that doing any other way might end up in unexpected behaviour. Which is fine as it shoudn't in my opinion. 

Let's just mention our own testing, which also starts the binary directly, bypassing the script. Going ahead with this approach, all of opencv tests are going to set (and hopefully unset) env vars, even in case of assertions (failures)? 

The above and the NaNoFi case I think are not issues we need to resolve, just consequences of violating some best practices and principals: what belongs together should be handled as a unit, being responsible to work independently from the environment. 

Don't get me wrong, I agree that the concerns you and @bakaid raised on being thread-safe when setting the env var is valid and should be addressed. What I would like to say that this is something the processor is responsible for, so it should be handled by that, otherwise we inject a hidden dependencies like the ones we listed above.

extensions/opencv/CaptureRTSPFrame.cpp
nghiaxlee: @phrocker : Thank you, that would be great.

nghiaxlee:
Since I already implemented the Initializer, I will put it here, so we can decide whether it is a good approach or not.
@phrocker : Can you give me more detail about the shell script approach? Like why is it good and maybe an example.

phrocker:
@nghiaxlee Sorry i missed the notifications. Allow me to answer the questions in reverse order.  I presented my chief concern, which is the sanctity of environment variables that are set externally. 

Before I begin can I ask that you agree with the following:

1) Environment variables can be set via the command line that are different than what you are setting in the initializer?
2) Python processors can be written a posteriori that set these env variables ( they shouldn't but *may* happen )?

With these agreed upon, my chief concern is dealing with that. Are you and @arpadboda  suggesting that is not a concern? 

"About the python processor, if I set the variable in the Initializer construct, then it will be called at the very beginning, so it should not affect the processors."

We cannot guarantee the order of initializer execution amongst various extensions. Even if that were to become true you can't guarantee the user hasn't set something before your initializers are called. Putting the onus externally improves configuration management. Why should we require that users re-compile a binary to achieve change in an environment variable. 

"Can you give me more detail about the shell script approach? Like why is it good"
I'm sorry if I relayed this as a "good" approach. I never meant to suggest that, but that it is the lesser of two evils. We put the onus on the user OR at the very least we provide the consumer with the ability to make changes. My comment about check and set is an option, but my only goal is to avoid these conflicting settings. 

Are these not a concern? Happy to hear a disagreement on the tenets I've laid out here and previously.

arpadboda:
@phrocker : 

1) env vars are inherited from the parent process, but can be overwritten.
The overwrite we do doesn't effect the parent process (which is most probably a shell).
How could the parent process (command line) overwrite the env var while MiNiFi is running? To be able to edit env vars while MiNiFi is running, the latter needs to be forked, which means copying env vars. 
2) this concern I find valid, too, and in my opinion neither of the solutions we talked about (script, setting in code) protects against this. Moreover, I'm afraid we can't. What we can do is to check the env var in ontrigger logic to ensure it's still the same. In case it has been set to something else, we can log error and yield. Which might be better than doing something stupid, but still far from nice.

phrocker:
@arpadboda I don't think that's a valid interpretation of what I wrote. I would appreciate @nghiaxlee 's interpretation of my comments.

phrocker:
"honestly I'm against placing the env var to the minifi.sh script. First I don't think we highlight that the script is the only supported way to start the binary, but Nanofi is a more serious concern at the moment: you can configure MiNiFi processors to the flowchain, which would break in such case. It's an edgecase, but still, valid at the moment."

@arpadboda I did  not see this response. Allow me to address this. I am not against codifying this, but we need to be aware that the environment have have already set this and if so have a plan that's reasonable ( i.e. not overriding it ). This transport is not the transport that all streams use. Most common, but not always the case. Therefore a user may and do start processes with that environment variable set. 

NanoFi is a library, no? So this will be implementation specific -- as in binaries produced have the ability to control the entry point of that process. In this case, with initializers, we don't control the entry point. Therefore we have to resolve that. I'm completely okay with alternatives. I've presented two, but have heard nothing to address the concerns or alternatives themselves.

arpadboda:
Merging this, handling env vars is a complex topic, to be followed up in scope of https://issues.apache.org/jira/browse/MINIFICPP-991

arpadboda:
Added further follow-ups: 
https://issues.apache.org/jira/browse/MINIFICPP-994
https://issues.apache.org/jira/browse/MINIFICPP-995

PR #613, opened 2019-07-16
Minificpp 927 Nanofi tailfile delimited processor

msharee9:

### For all changes:
     in the commit message?

### For code changes:

### Note:


msharee9:
Rebased MINIFICPP-926 on this branch. So you will see changes from that branch as well in this PR. However, the only changes are in nanofi/src/api/ecu.c file and nanofi/ecu/tailfile_delimited.c

nanofi/src/api/ecu.c
```diff
@@ -0,0 +1,271 @@
...
+                if (flow_file_exists && ffr_delim) {
+                    add_flow_file_record(&ff_list, ffr_delim);
+                    flow_file_exists = 0;
+                }
+                curr_offset++;
+            }
+            begin = (end + 1);
+        }
+
+        if (!end && begin && *begin != '\0' && bytes_read == MAX_BYTES_READ) {
```
bakaid: I do not think `begin` can be `NULL`, so that check is unnecessary.
I see some inconsistency in this conditional.
We only enter it if we could read a full `MAX_BYTES_READ` from the file. But if in this buffer there has been complete lines, we will still create a partial flow_file from the remainder of the buffer.
I do not think this is a bad behaviour, but if we already to that, why can't we do the same if there has been less than `MAX_BYTES_READ`?
I do not think this causes erroneus behaviour, I am just trying to understand the rationale behind it.

nanofi/src/core/file_utils.c
```diff
@@ -0,0 +1,138 @@
...
+    if (stat(path, &dir_stat) < 0) {
+        return 0;
+    }
+    return S_ISDIR(dir_stat.st_mode);
+}
+
+char * get_separator(int force_posix)
+{
+    char * sep = malloc(2 * sizeof(char));
+    strncpy(sep, "/", 1);
```
bakaid: Using simply `strcpy(sep, "/")` this would still be completely safe and the terminating '\0' would not have to be appended manually.

nanofi/src/core/file_utils.c
```diff
@@ -0,0 +1,138 @@
...
+    sep[1] = '\0';
+#ifdef WIN32
+    if (force_posix) {
+        return sep;
+    } else {
+        char * sep = malloc(3 * sizeof(char));
```
bakaid: A C string of a single backslash needs 2 bytes, not 3.

nanofi/src/core/file_utils.c
```diff
@@ -0,0 +1,138 @@
...
+        strncpy(sep, "\\", 2);
```
bakaid: This is a single character.

nanofi/src/core/file_utils.c
bakaid: This also shadows the previous sep, so that one will be leaked.

nanofi/src/core/file_utils.c
```diff
@@ -0,0 +1,138 @@
...
+    }
+#else
+    return sep;
+#endif
+}
+
+char * concat_path(const char * parent, const char * child) {
+    char * path = (char *)malloc((strlen(parent) + strlen(child) + 2) * sizeof(char));
+    strcpy(path, parent);
+    strcat(path, get_separator(1));
```
bakaid: This leaks the separator string.

nanofi/src/core/file_utils.c
```diff
@@ -0,0 +1,138 @@
...
+
+int is_directory(const char * path) {
+    struct stat dir_stat;
+    if (stat(path, &dir_stat) < 0) {
+        return 0;
+    }
+    return S_ISDIR(dir_stat.st_mode);
+}
+
+char * get_separator(int force_posix)
```
bakaid: This function has multiple issues, but could be replaced by using C string literals:
```
const char* get_separator(int force_posix) {
#ifdef WIN32
    if (!force_posix) {
        return "\\";
    }
#endif
    return "/";
}
```

nanofi/include/api/ecu.h
```diff
@@ -0,0 +1,57 @@
...
+#include "api/nanofi.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+extern nifi_instance * instance;
+extern standalone_processor * proc;
+extern flow_file_list * ff_list;
+extern uint64_t curr_offset;
```
bakaid: I think storing the state of a specific processor in a bunch of global variables, especially in the file supposedly used by all ECUs is not a good practice.
This means not only that we can't have multiple instances of the same processor, but even single instances of different processors that use this variable.
For a PoC this is acceptable, but I don't think this is the way to go forward with ECUs.

msharee9: Even though these are global variables, each instance of the ecu will have its own variable. Why do you think that these variables will cause any problem? I agree that having global variables like these is not a good practice and I will try to correct it but I think global variables will not cause any problem with one or multiple instances of same or different ecus.

nanofi/src/core/file_utils.c
```diff
@@ -0,0 +1,138 @@
...
+{
+    char * sep = malloc(2 * sizeof(char));
+    strcpy(sep, "/");
+#ifdef WIN32
+    if (force_posix) {
+        return sep;
+    } else {
+        char * sep = malloc(2 * sizeof(char));
```
arpadboda: I doesn't really make sense for me to reallocate const strings every time this is called, not to mention the fact that it shadows the previous declaration, which results in a leak.

nanofi/src/core/file_utils.c
```diff
@@ -0,0 +1,138 @@
...
+
+    struct stat fstat;
+    if (stat(path, &fstat) < 0) {
+        return -1;
+    }
+    return fstat.st_size;
+}
+
+int make_dir(const char * path) {
+    if (!path) return 0;
```
arpadboda: This doesn't seem to be a happy path...

nanofi/src/core/file_utils.c
```diff
@@ -0,0 +1,138 @@
...
+        if (!found) {
+            return -1;
+        }
+        int len = found - path;
+        char * dir = calloc(len + 1, sizeof(char));
+        strncpy(dir, path, len);
+        dir[len] = '\0';
+        int res = make_dir(dir);
+        if (res < 0) {
+            free(dir);
```
arpadboda: This can be done after the make_dir call (just to avoid duplicating)

nanofi/src/core/flowfiles.c
```diff
@@ -0,0 +1,99 @@
...
+    flow_file_list * el = NULL;
+    LL_FOREACH(*ff_list, el) {
+        if (el) {
+            free_flowfile(el->ff_record);
+        }
+    }
+}
+
+flow_file_record * write_to_flow_file(const char * buff, size_t count, nifi_instance * instance, standalone_processor * proc) {
+    if (!buff || !instance || !proc) {
```
arpadboda: The null check macro can still be used in such cases

nanofi/src/core/string_utils.c
```diff
@@ -20,67 +20,145 @@
...
 #include <stdlib.h>
+#include <stdio.h>
 
-tokens tokenize_string(const char * str, char delim, tokenizer_mode_t mode) {
-    tokens tks;
-    tks.num_strings = 0;
-    tks.total_bytes = 0;
+int validate_list(struct token_list * tk_list) {
+    if (tk_list && tk_list->head && tk_list->tail && tk_list->size > 0) {
+        return 1;
```
arpadboda: -1? Just to be consistent with ourselves.

nanofi/src/core/file_utils.c
bakaid: This is still not fixed. The shadowing and memory leak still occurs in the windows ifdef.
Also, the alternative solution I provided here would not need to be free'd and would be much less error prone to use. I do not see any reason why a dynamically allocated string would be better in this case.

nanofi/include/api/ecu.h
bakaid: Yes, that is my concern too. It would be a valid use case to use more than one ECU in the same application, and the current implementation does not allow that.

nanofi/src/api/ecu.c
```diff
@@ -0,0 +1,266 @@
...
+        struct token_list tokens = tokenize_string_tailfile(buff, delim);
+        if (tokens.total_bytes > 0) {
+            curr_offset += tokens.total_bytes;
+            fseek(fp, curr_offset, SEEK_SET);
+        }
+
+        token_node * head;
+        for (head = tokens.head; head && head->data; head = head->next) {
+            flow_file_record * ffr = write_to_flow_file(head->data, strlen(head->data), instance, proc);
+            char offset[21] = {0};
```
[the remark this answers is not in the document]
bakaid: @arpadboda uint64_t in its decimal representation requires at most 20 characters, plus the \0.

nanofi/CMakeLists.txt
```diff
@@ -33,9 +33,11 @@ else()
 include_directories(../libminifi/opsys/posix)
 endif()
 
-file(GLOB NANOFI_SOURCES  "src/api/*.cpp" "src/core/*.c*" "src/cxx/*.cpp" "src/sitetosite/*.c*")
+file(GLOB NANOFI_SOURCES  "src/api/*.cpp" "src/api/*.c" "src/core/*.c*" "src/cxx/*.cpp" "src/sitetosite/*.c*")
```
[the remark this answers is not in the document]
phrocker: @msharee9  Hey, I know we discussed this ticket 
https://issues.apache.org/jira/browse/MINIFICPP-666, but I never set the version. If we play our cards right in 0.7.0 and 0.8.0 I think we can very quickly decouple these projects. There are other tickets in this release that go toward that ( removing instance ) but @arpadboda's comment is a key reminder that we need to do that.

nanofi/ecu/tailfile_chunk.c
```diff
@@ -0,0 +1,105 @@
...
+#include <unistd.h>
+#include <errno.h>
+#include <limits.h>
+#include <signal.h>
+#include <sys/stat.h>
+#include <stdlib.h>
+#include <string.h>
+#include <stdio.h>
+
+int main(int argc, char** argv) {
```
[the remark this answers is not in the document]
msharee9: Good point, Thought about it earlier but forgot to come back to it. Thanks for pointing out

nanofi/include/api/ecu.h
msharee9: Currently these ECUs are used as a binary file run from command line supplying arguments as an individual unit. Still no connectivity like a flow from one ecu to other. That is the vision of upcoming ECU EFM integration if I am not wrong. I take your point that we should avoid the globals in order to setup ourselves to integrate EFM later on easily.

nanofi/include/core/cstructs.h
```diff
@@ -155,9 +155,33 @@ typedef struct cstream {
...
 
-typedef struct tokens {
-    char ** str_list;
-    uint64_t num_strings;
+typedef struct token_node {
+    char * data;
+    struct token_node * next;
+} token_node;
+
+typedef struct token_list {
```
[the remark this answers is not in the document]
msharee9: yes, I remember that and I mentioned in that PR that this token_list has some additional members that was important for the log_aggregator ecu. If I use utlist I will have to have some other structure outside of the list, to keep track of the offset in the file. Currently the token_list encapsulates that so that token nodes and the offset stay together and it is better to keep it that way.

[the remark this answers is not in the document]
msharee9: Yes, I remember that and I mentioned in that PR that if I use utlist, I will have to use an external structure to keep track of extra variables. Currently it is encapsulated in token_list staying together with the tokens. Using utlist would make these variables to be unnecessarily copied into each linked list node.

msharee9:
> I tried the examples manually, but could not get it to work (it did not transmit flow files to my NiFi instance). After a little bit of debugging what I can see is that is it trying to use the wrapped MiNiFi implementation for sending flow files to the RPG, but that is missing a content repository and many other things.
> I think it should use the same native implementation that is used in the transmit_payload example. @arpadboda might be able to help with it, since he wrote that.
> 
> However, it is a -1 from me until it actually works and I can try it.
> 
> Please also see my inline comments, especially about resolved issues which I think are not resolved.

@bakaid .. please review this again. I have fixed the transmission of flow files to nifi using the transmit_payload as an example.

nanofi/src/api/nanofi.cpp
```diff
@@ -124,25 +124,42 @@ nifi_instance *create_instance(const char *url, nifi_port *port) {
...
+
+char * get_proc_uuid_from_processor(standalone_processor * proc) {
+    const std::string uuid = proc->getUUIDStr();
+    uint64_t len = uuid.size();
+    char * uuid_str = (char *)malloc((len+1) * sizeof(char));
+    strcpy(uuid_str, uuid.c_str());
+    return uuid_str;
+}
+
+char * get_proc_uuid_from_context(processor_context * ctx) {
```
arpadboda: const char * here as well, morover the input context could be const, too.

nanofi/src/core/flowfiles.c
```diff
@@ -0,0 +1,169 @@
...
+                printf("Failed to send payload, flow file %s\n", ffl->ff_record->contentLocation);
+            }
+        }
+        free(buff);
+    }
+    fclose(fp);
+}
+
+void transmit_payload(struct CRawSiteToSiteClient * client, struct flow_file_list * ff_list, int complete) {
+    if (!client || !ff_list) {
```
arpadboda: The null check macro could be used here, too.

nanofi/src/api/ecu.c
```diff
@@ -0,0 +1,494 @@
...
+    }
+}
+
+uint64_t get_current_offset(char * uuid) {
+    struct processor_params * pp = NULL;
+    HASH_FIND_STR(procparams, uuid, pp);
+    if (pp) {
+        return pp->curr_offset;
+    }
+    return 0;
```
arpadboda: Isn't 0 a vlaid offset?

nanofi/src/api/ecu.c
```diff
@@ -0,0 +1,494 @@
...
+}
+
+void update_proc_params(char * uuid, uint64_t value, flow_file_list * ffl) {
+    struct processor_params * pp = get_proc_params(uuid);
+    if (!pp) {
+        pp = (struct processor_params *)malloc(sizeof(struct processor_params));
+        memset(pp, 0, sizeof(struct processor_params));
+        pp->ff_list = ffl;
+        pp->curr_offset = value;
+        char * uuid_str = (char *)malloc((strlen(uuid) + 1) * sizeof(char));
```
arpadboda: As uuids are fixed length, I think we should handle them as a struct. In that case they can be members without using pointers and we save a lot of malloc/free calls.

nanofi/src/api/ecu.c
msharee9: Yes it is a valid offset. 0 is returned the very first time the processor is invoked and we do not find the processor in the hashtable.

msharee9: In subsequent calls, the processor is found in the hashtable.

nanofi/src/api/ecu.c
msharee9: Updated code with static length uuid char array

msharee9:
> It's getting better, I appreciate that it can now transmit flow files to NiFi.
> 
> I have still have some concerns to address:
> -The mixed mode seems to introduce a lot of complexity but not sure it worth the effort, do you really think we need it? @phrocker what do you think?
> -There is no logging even though Nanofi supports it. Could you take a look at site2site and logging part of nanofi for some inspiration?
> -Given the number of allocations I'm a bit afraid of introducing leaks, could you try running valgrind or any other dynamic analytics tool that may help us reduce this risk?

I want to keep the mixed mode. There is no harm here I guess. It is not as complicated to use or to understand from a users perspective. It is an additional tool in hand if chunk and delimited do not suffice the needs of the user.
I had also updated code to rectify valgrind leaks. There are some leaks already present in the code mostly due to the use of C++ code from minifi mostly due to cyclic references I believe. I took care of any additional leaks injected by the nanofi tailfile implementation.
I will look into logging once we start the EFM ECU integration implementation. But for now I think this PR is pretty good to be merged.

nanofi/ecu/log_aggregator.c
```diff
@@ -0,0 +1,78 @@
...
+#include <string.h>
+#include <errno.h>
+#include <limits.h>
+#include <signal.h>
+#include <sys/stat.h>
+
+int main(int argc, char** argv) {
+
+    if (argc < 7) {
+        printf("Error: must run ./log_aggregator <file> <interval> <delimiter> <nifi instance url> <tcp port number> <nifi port uuid>\n");
```
bakaid: The `nifi instance url` is misleading, since we need a hostname now.

nanofi/include/api/ecu.h
```diff
@@ -0,0 +1,89 @@
...
+
+
+typedef struct processor_params {
+    char uuid_str[37]; //key
+    struct flow_file_list * ff_list;
+    uint64_t curr_offset;
+    UT_hash_handle hh;
+} processor_params;
+
+extern processor_params * procparams;
```
bakaid: Having this is a global variable is
 - NOT thread-safe (you can't do anything that would write the hash table with anything else in parallel), meaning you have to be very careful when using ECUs in multiple threads
 - I am not sure how this would scale

These are acceptable for now, but should be documented.

nanofi/src/api/ecu.c
```diff
@@ -0,0 +1,493 @@
...
+            uint64_t len = end - begin;
+            if (len > 0) {
+                if (!ffl_node || ffl_node->complete) {
+                    ffl_node = add_flow_file_to_proc_params(uuid_str, generate_flow(ctx));
+                }
+                write_flow_file(ffl_node->ff_record, begin, len);
+                update_curr_offset(uuid_str, (len + 1));
+            }
+            else {
+                update_curr_offset(uuid_str, 1);
```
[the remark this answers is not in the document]
msharee9: Fixed it. Thanks for pointing it out.

nanofi/src/api/ecu.c
```diff
@@ -0,0 +1,493 @@
...
+    initialize_content_repo(ctx);
+    const char * uuid = get_proc_uuid_from_context(ctx);
+    char uuid_str[37];
+    strcpy(uuid_str, uuid);
+
+    FILE * fp = fopen(props.file_path, "rb");
+
+    if (!fp) {
+        printf("Unable to open file. {file: %s, reason: %s}\n", props.file_path, strerror(errno));
+        return;
```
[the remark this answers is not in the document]
msharee9: Fixed it.

nanofi/src/api/ecu.c
```diff
@@ -0,0 +1,493 @@
...
+    props.file_path = (char *)malloc((len + 1) * sizeof(char));
+    strncpy(props.file_path, file_path, len);
+    props.file_path[len] = '\0';
+    props.delimiter = delim;
+    return props;
+}
+
+void on_trigger_logaggregator(processor_session * ps, processor_context * ctx) {
+
+    struct properties props = get_properties(ctx);
```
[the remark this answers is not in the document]
msharee9: Refactored it so we don't parse for properties every time.

msharee9:
> I can't get any of the ecus actually send flow_files to my NiFi instance.
> After a bit of debugging it seems that the uuid the flow_files are written to in the hash table in ecu.c (that is acquired by calling get_proc_uuid_from_context) is not the same uuid the main function tries to read the flow_files from (which is acquired by get_proc_uuid_from_processor).
> 
> Please also see my inline comments - some are just suggestions, but there are some leaks and crashes.

Tested all ECUs. I am able to send flow file payload to nifi now. It was a bug with returning .c_str() from a temporary string. Pretty bad I would say.

nanofi/include/api/ecu.h
msharee9: The only other way to deal with this kind of issue to not have globals is to modify the internal minifi C++ API we are using for nanofi. Since we will anyhow move away from using that, I did not want to deal with modifying that and throw away that work later. It wouldn't be a problem to have mutex protection when we deal with multiple threads.

Can you please elaborate a use case where we would have to deal with multiple threads in an ECU given we offload EFM communication to the orchestrator agent ?

nanofi/src/core/file_utils.c
```diff
@@ -0,0 +1,146 @@
...
+        return -1;
+    }
+}
+
+char * get_current_working_directory() {
+    char * cwd = (char *)malloc(PATH_MAX * sizeof(char));
+    #ifdef WIN32
+        _getcwd(cwd,PATH_MAX);
+    #else
+        getcwd(cwd, PATH_MAX);
```
bakaid: This can fail (if the current path is too long, for example), returning NULL, and leaving the buffer in an undefined state. In this case we would return uninitialized memory as a C-string which would be a security issue.

nanofi/src/core/flowfiles.c
```diff
@@ -0,0 +1,171 @@
...
+        return;
+    }
+    flow_file_list * el = NULL;
+    LL_FOREACH(ff_list, el) {
+        char * payload = NULL;
```
arpadboda: This seems to be unreferenced

nanofi/src/core/file_utils.c
```diff
@@ -0,0 +1,146 @@
...
+        return "\\";
+    }
+#endif
+    return "/";
+}
+
+char * concat_path(const char * parent, const char * child) {
+    char * path = (char *)malloc((strlen(parent) + strlen(child) + 2) * sizeof(char));
+    strcpy(path, parent);
+    const char * sep = get_separator(1);
```
arpadboda: Why forcing posix here? 
Either the this argument should be part of the signature of concat_path as well, or don't force it.

nanofi/src/core/flowfiles.c
msharee9: I understand that NULL macro can be used but this is more readable and straight forward.

nanofi/ecu/log_aggregator.c
```diff
@@ -0,0 +1,109 @@
...
+    // Check for file existence
+    if (S_ISDIR(stats.st_mode)){
+        printf("Error: %s is a directory!\n", file);
+        exit(1);
+    }
+
+    errno = 0;
+    unsigned long intrvl = strtol(interval, NULL, 10);
+
+    if (errno == ERANGE || intrvl == LONG_MAX || intrvl == LONG_MIN) {
```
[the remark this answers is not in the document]
msharee9: Corrected. Please see updated code.

nanofi/src/core/file_utils.c
msharee9: Refactored. Please see updated code

nanofi/src/core/file_utils.c
msharee9: Updated.

nanofi/src/core/file_utils.c
msharee9: Ok

nanofi/src/core/flowfiles.c
msharee9: I think this is more readable than using the macro.

nanofi/src/core/string_utils.c
msharee9: Updated.

nanofi/ecu/log_aggregator.c
```diff
@@ -0,0 +1,79 @@
...
+#include <stdio.h>
+#include <string.h>
+#include <errno.h>
+#include <limits.h>
+#include <signal.h>
+#include <sys/stat.h>
+
+int main(int argc, char** argv) {
+
+    if (argc < 7) {
```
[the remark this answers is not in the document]
msharee9: The rest will be ignored if greater than or equal to 7.

nanofi/src/api/nanofi.cpp
```diff
@@ -245,28 +262,55 @@ flow_file_record* create_ff_object_nc() {
...
     ffr->ffp = static_cast<void*>(new std::shared_ptr<core::FlowFile>(plan->getCurrentFlowFile()));
     ffr->keepContent = 1;
     auto ff_content_repo_ptr = (static_cast<std::shared_ptr<minifi::core::ContentRepository>*>(ffr->crp));
     auto claim = std::make_shared<minifi::ResourceClaim>(*ff_content_repo_ptr);
-    const char * full_path = claim->getContentFullPath().c_str();
-    int len = strlen(full_path);
-    ffr->contentLocation = (char *) malloc(sizeof(char) * (len + 1));
-    snprintf(ffr->contentLocation, len + 1, "%s", full_path);
+
+    int len = strlen(claim->getContentFullPath().c_str());
```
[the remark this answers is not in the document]
msharee9: Updated.

nanofi/src/api/nanofi.cpp
```diff
@@ -124,25 +124,42 @@ nifi_instance *create_instance(const char *url, nifi_port *port) {
...
   plan->addProcessor(ptr, name);
   ExecutionPlan::addProcessorWithPlan(ptr->getUUIDStr(), plan);
   return static_cast<standalone_processor*>(ptr.get());
 }
 
+void initialize_content_repo(processor_context * ctx) {
+    ctx->initializeContentRepository(get_current_working_directory());
+}
+
+char * get_proc_uuid_from_processor(standalone_processor * proc) {
```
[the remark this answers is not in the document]
msharee9: Updated.

nanofi/src/api/nanofi.cpp
msharee9: Updated.

nanofi/src/core/flowfiles.c
msharee9: Removed it.

nanofi/src/core/flowfiles.c
```diff
@@ -0,0 +1,171 @@
...
+    update_attribute(ffr, "content location", content_location, strlen(content_location));
+    update_attribute(ffr, "tailfile path", (char*)file_path, strlen(file_path));
+}
+
+void transmit_flow_files(nifi_instance * instance, flow_file_list * ff_list, int complete) {
+    if (!instance || !ff_list) {
+        return;
+    }
+    flow_file_list * el = NULL;
+    LL_FOREACH(ff_list, el) {
```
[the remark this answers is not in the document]
msharee9: Refactored.

nanofi/src/core/file_utils.c
```diff
@@ -0,0 +1,146 @@
...
+    rmdir(dir_path);
+    closedir(d);
+}
+
+uint64_t file_size(const char * path) {
+    if (!path) return -1;
+
+    struct stat fstat;
+    if (stat(path, &fstat) < 0) {
+        return -1;
```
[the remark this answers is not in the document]
msharee9: Refactored.

nanofi/src/core/file_utils.c
msharee9: Refactored.

nanofi/ecu/tailfile_delimited.c
```diff
@@ -0,0 +1,76 @@
...
+    char uuid_str[37];
+    get_proc_uuid_from_processor(params.processor, uuid_str);
+
+    while (!stopped) {
+        flow_file_record * new_ff = invoke(params.processor);
+        struct processor_params * pp = NULL;
+        HASH_FIND_STR(procparams, uuid_str, pp);
+        if (pp) {
+            transmit_payload(client, pp->ff_list, 1);
+            delete_completed_flow_files_from_proc(uuid_str);
```
arpadboda: According to valgrind we leak really a **lot** here:

```
==19410== 26,335 (2,640 direct, 23,695 indirect) bytes in 55 blocks are definitely lost in loss record 114 of 114
==19410==    at 0x4C3017F: operator new(unsigned long) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==19410==    by 0x2BE628: create_ff_object_nc (nanofi.cpp:272)
==19410==    by 0x2C3543: generate_flow (nanofi.cpp:277)
==19410==    by 0x2BD825: on_trigger_tailfiledelimited (ecu.c:503)
==19410==    by 0x2EBAC1: operator() (std_function.h:706)
==19410==    by 0x2EBAC1: org::apache::nifi::minifi::processors::CallbackProcessor::onTrigger(org::apache::nifi::minifi::core::ProcessContext*, org::apache::nifi::minifi::core::ProcessSession*) (CallbackProcessor.cpp:44)
==19410==    by 0x2F21CE: ExecutionPlan::runNextProcessor(std::function<void (std::shared_ptr<org::apache::nifi::minifi::core::ProcessContext>, std::shared_ptr<org::apache::nifi::minifi::core::ProcessSession>)>, std::shared_ptr<flowfile_input_params>) (Plan.cpp:183)
==19410==    by 0x2C22E9: invoke_ff (nanofi.cpp:705)
==19410==    by 0x2A72F4: main (tailfile_delimited.c:56)
```

Could you investigate this before mering?

msharee9: I have investigated this, infact, the function to delete_completed_flow_files_from_proc does free all flow files. However, the leak was during creating a flow file. Fixed it. Thanks!

PR #614, opened 2019-07-17
MINIFICPP-732 - Add property to expose librdkafka "debug" values for …

nghiaxlee:
…PublishKafka

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/librdkafka/PublishKafka.cpp
+++ b/extensions/librdkafka/PublishKafka.cpp
@@ -79,6 +79,7 @@ core::Property PublishKafka::KerberosKeytabPath("Kerberos Keytab Path",
 core::Property PublishKafka::MessageKeyField("Message Key Field", "The name of a field in the Input Records that should be used as the Key for the Kafka message.\n"
                                              "Supports Expression Language: true (will be evaluated using flow file attributes)",
                                              "");
+core::Property PublishKafka::DebugContexts("Debug contexts", "A comma-separated list of debug contexts to enable. Including: generic, broker, topic, metadata, feature, queue, msg, protocol, cgrp, security, fetch, interceptor, plugin, consumer, admin, eos, all", "");
 core::Relationship PublishKafka::Success("success", "Any FlowFile that is successfully sent to Kafka will be routed to this Relationship");
 core::Relationship PublishKafka::Failure("failure", "Any FlowFile that cannot be sent to Kafka will be routed to this Relationship");
 
@@ -107,6 +108,7 @@ void PublishKafka::initialize() {
   properties.insert(KerberosPrincipal);
   properties.insert(KerberosKeytabPath);
   properties.insert(MessageKeyField);
+  properties.insert(DebugContexts);
   setSupportedProperties(properties);
   // Set the supported relationships
   std::set<core::Relationship> relationships;
@@ -129,9 +131,16 @@ bool PublishKafka::configureNewConnection(const std::shared_ptr<KafkaConnection>
 
   auto key = conn->getKey();
 
+  if (context->getProperty(DebugContexts.getName(), value) && !value.empty()) {
+    rd_kafka_conf_set(conf_, "debug", value.c_str(), errstr, sizeof(errstr));
+    logger_->log_debug("PublishKafka: debug properties [%s]", value);
+    if (result != RD_KAFKA_CONF_OK)
+      logger_->log_error("PublishKafka: configure debug properties error result [%s]", errstr);
+  }
+
   if (!key->brokers_.empty()) {
     result = rd_kafka_conf_set(conf_, "bootstrap.servers", key->brokers_.c_str(), errstr, sizeof(errstr));
-    logger_->log_debug("PublishKafka: bootstrap.servers [%s]", value);
+    logger_->log_debug("PublishKafka: bootstrap.servers [%s]", key->brokers_.c_str());
     if (result != RD_KAFKA_CONF_OK)
       logger_->log_error("PublishKafka: configure error result [%s]", errstr);
   } else {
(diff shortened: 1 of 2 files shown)


extensions/librdkafka/PublishKafka.cpp
```diff
@@ -141,7 +150,7 @@ bool PublishKafka::configureNewConnection(const std::shared_ptr<KafkaConnection>
 
   if (!key->client_id_.empty()) {
     rd_kafka_conf_set(conf_, "client.id", key->client_id_.c_str(), errstr, sizeof(errstr));
-    logger_->log_debug("PublishKafka: client.id [%s]", value);
+    logger_->log_debug("PublishKafka: client.id [%s]", key->client_id_.c_str());
```
phrocker: you can simply use key->client_id_ as we perform this conversion for you.

extensions/librdkafka/PublishKafka.cpp
```diff
@@ -79,6 +79,7 @@ core::Property PublishKafka::KerberosKeytabPath("Kerberos Keytab Path",
 core::Property PublishKafka::MessageKeyField("Message Key Field", "The name of a field in the Input Records that should be used as the Key for the Kafka message.\n"
                                              "Supports Expression Language: true (will be evaluated using flow file attributes)",
                                              "");
+core::Property PublishKafka::DebugContexts("Debug contexts", "A comma-separated list of debug contexts to enable. Including: generic, broker, topic, metadata, feature, queue, msg, protocol, cgrp, security, fetch, interceptor, plugin, consumer, admin, eos, all", "");
```
phrocker: Is this a defined list that can be controlled via allowable values?

nghiaxlee: yes, according to [https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md](this)

extensions/librdkafka/PublishKafka.cpp
nghiaxlee: thanks, I did not notice that

extensions/librdkafka/PublishKafka.cpp
phrocker: So then can you use our allowable values construct?

extensions/librdkafka/PublishKafka.cpp
```diff
@@ -129,9 +131,16 @@ bool PublishKafka::configureNewConnection(const std::shared_ptr<KafkaConnection>
 
   auto key = conn->getKey();
 
+  if (context->getProperty(DebugContexts.getName(), value) && !value.empty()) {
+    rd_kafka_conf_set(conf_, "debug", value.c_str(), errstr, sizeof(errstr));
```
phrocker: When will we receive feedback regarding this setting?

extensions/librdkafka/PublishKafka.cpp
nghiaxlee: I'm afraid not, the possible value can be any combination from the list above, not just a single element.

extensions/librdkafka/PublishKafka.cpp
nghiaxlee: Can you explain it? I think I don't know what does "feedback about the setting" mean.

extensions/librdkafka/PublishKafka.cpp
phrocker: Allowable values don't preclude having more than one option. Is there something that gave you that impression in the code?

extensions/librdkafka/PublishKafka.cpp
phrocker: If we can'd ultimately do allow able values ( i commented above ) -- then that means we need to rely on this CSV. Very little validation is performed here -- so I presume librdkafka does that validation. In those cases do you get an immediate response or must we wait until data is supplied and sent through Kafka?

extensions/librdkafka/PublishKafka.cpp
nghiaxlee: I assume we are talking about 'withAllowableValues' in PropertyBuilder. In 'LogAttribute.cpp', there is a set of allowable values ("debug", "info", etc), though only one of them can be set at a time. In this case, "msg,topic,broker",  "topic,msg" or "broker,msg" are all valid values. Thus, I don't understand how can allowable values can check that. If I misunderstand anything, I would love to hear your correction.

extensions/librdkafka/PublishKafka.cpp
nghiaxlee: From what I saw on the logs, we can only know after sending the data, and librdkafka will terminate before the first invalid setting ("topic,broker,mewrong,msg" then only "topic,broker" are applied). For validation, I know that I can split the string by "," then check whether they are valid or not. If you have better idea, please let me know.

extensions/librdkafka/PublishKafka.cpp
arpadboda: @nghiaxlee what you say makes perfect sense, NiFI provides ListValidator for the given purpose:
https://github.com/apache/nifi/blob/master/nifi-commons/nifi-utils/src/main/java/org/apache/nifi/processor/util/StandardValidators.java#569

Although we don't have this functionality in MiNiFi at the moment, will create a follow-up Jira soon. I think the scope of this ticket is unquestionably small to cover this change.

phrocker: In this case you would not require a comma separated list and would simply create that comma separated list further down.

phrocker: I don't think allowable values limits you to a single resulting value. Can you not build a flowfile with the YAML multiple entries?

phrocker: @nghiaxlee I think I'm the long pole here so since my time is so fragmented. I created https://issues.apache.org/jira/browse/MINIFICPP-988 as a follow on with no release since there seems to be a disagreement about the approach. You are free to make a decision in that ticket, but I ask that you evaluate all options holistically.

arpadboda:
Merging this as it works and adds feature, property validation will be done in scope of the linked follow-up.

PR #620, opened 2019-08-05
MINIFICPP-996 - Output port keeps printing info message implying tran…

arpadboda:
…smission

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/http-curl/sitetosite/HTTPProtocol.cpp
+++ b/extensions/http-curl/sitetosite/HTTPProtocol.cpp
@@ -54,7 +54,7 @@ std::shared_ptr<Transaction> HttpSiteToSiteClient::createTransaction(std::string
   std::stringstream uri;
   uri << getBaseURI() << "data-transfer/" << dir_str << "/" << getPortId() << "/transactions";
   auto client = create_http_client(uri.str(), "POST");
-  client->appendHeader(PROTOCOL_VERSION_HEADER, "1");
+  client->appendHeader(PROTOCOL_VERSION_HEADER, SUPPORTED_HTTTP_S2S_VERSION);
   client->setConnectionTimeout(5);
   client->setContentType("application/json");
   client->appendHeader("Accept: application/json");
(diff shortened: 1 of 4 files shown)


extensions/http-curl/sitetosite/HTTPProtocol.cpp
```diff
@@ -63,7 +63,16 @@ std::shared_ptr<Transaction> HttpSiteToSiteClient::createTransaction(std::string
   client->submit();
   if (peer_->getStream() != nullptr)
     logger_->log_debug("Closing %s", ((io::HttpStream*) peer_->getStream())->getClientRef()->getURL());
-  if (client->getResponseCode() == 201) {
+  uint64_t  client_response = client->getResponseCode();
+  if (direction == RECEIVE && client_response == 200) {
```
arpadboda: This case cannot occur at the moment as NiFi handles this improperly.
When fixed and tested in scope of https://github.com/apache/nifi/pull/3647 I found that we throw exception and execute rollback in case 200 is received, even though it's a valid case, 200 is not an error.
Transaction being null is for error cases, so this function now creates a dummy transaction, which then gets removed.

extensions/http-curl/sitetosite/HTTPProtocol.cpp
```diff
@@ -63,7 +63,16 @@ std::shared_ptr<Transaction> HttpSiteToSiteClient::createTransaction(std::string
...
+    // Received 200 - remote port has no flowfiles to send
```
phrocker: I feel like this should be a 204 to clarify that there is no content.

extensions/http-curl/sitetosite/HTTPProtocol.cpp
```diff
@@ -63,7 +63,16 @@ std::shared_ptr<Transaction> HttpSiteToSiteClient::createTransaction(std::string
   client->submit();
   if (peer_->getStream() != nullptr)
     logger_->log_debug("Closing %s", ((io::HttpStream*) peer_->getStream())->getClientRef()->getURL());
-  if (client->getResponseCode() == 201) {
+  uint64_t  client_response = client->getResponseCode();
```
bakaid: There is a double space here.

arpadboda: Fixed

PR #621, opened 2019-08-07
MINIFICPP-959: Review librdkafka thread safety

nghiaxlee:

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/librdkafka/PublishKafka.h
```diff
@@ -291,10 +99,9 @@ class KafkaPool {
  private:
```
nghiaxlee: I did not init connection here because the init function has flowfile as parameter (albeit currently the function does not use any flowfile attribute)

extensions/librdkafka/KafkaConnection.h
```diff
@@ -0,0 +1,197 @@
...
+  rd_kafka_conf_t *getConf() {
+    return conf_;
+  }
+
+  rd_kafka_t *getConnection() {
+    return kafka_connection_;
+  }
+
+  bool hasTopic(const std::string &topic) {
+    return topics_.find(topic) != std::end(topics_);
```
arpadboda: Map has count, which returns 0 or 1 (ya, someone in c++ committee definitely had two more beers than he should had). That's just nitpicking, but this func should be const, too.

extensions/librdkafka/KafkaConnection.h
```diff
@@ -0,0 +1,197 @@
...
+    }
+    return nullptr;
+  }
+
+  KafkaConnectionKey const * const getKey() const {
+    return &key_;
+  }
+
+  void putTopic(const std::string &topicName, const std::shared_ptr<KafkaTopic> topic) {
+    topics_.insert(std::make_pair(topicName, topic));
```
arpadboda: Why not simply topics_[topicName] = topic ?
On the other side, your the const shared ptr hardly makes sense, either const& (which doesn't copy, so the counter are not incremented), or simply non-const instance.

extensions/librdkafka/KafkaConnection.h
```diff
@@ -0,0 +1,197 @@
...
+      kafka_connection_ = nullptr;
+    }
+    if (conf_) {
+      rd_kafka_conf_destroy(conf_);
+      conf_ = nullptr;
+    }
+    initialized_ = false;
+  }
+
+  bool initialized() {
```
arpadboda: As it's an extension, we can go fully c++11 at least, can you add override where it makes sense?

extensions/librdkafka/PublishKafka.cpp
```diff
@@ -327,19 +328,21 @@ void PublishKafka::onTrigger(const std::shared_ptr<core::ProcessContext> &contex
...
+    KafkaLease lease(conn);
 
     if (!conn->initialized()) {
       logger_->log_trace("Connection not initialized to %s, %s, %s", client_id, brokers, topic);
-      // get the ownership so we can configure this connection
-      KafkaLease lease = conn->obtainOwnership();
-
       if (!configureNewConnection(conn, context, flowFile)) {
         logger_->log_error("Could not configure Kafka Connection");
-        session->transfer(flowFile, Failure);
```
arpadboda: Why not?

extensions/librdkafka/PublishKafka.h
```diff
@@ -275,9 +80,12 @@ class KafkaPool {
     std::lock_guard<std::mutex> lock(mutex_);
     auto connection = map_.find(key);
     if (connection != map_.end()) {
-      return connection->second;
+      auto conn = connection->second;
+      return conn->inUse() ? nullptr : conn;
```
bakaid: @nghiaxlee  Unfortunately this is still not thread-safe.
If two threads request a connection with the same key, and the connection is unused, both will get the KafkaConnection, and both will try to create a KafkaLease for it.
KafkaLease does not check whether the connection is in use, it just sets it as used, which can be done by both threads, meaning that both threads will access the KafkaConnection. (We have a mutex in this function, but that can be acquired by both threads and a connection return before either thread created the KafkaLease, thereby setting the inUse. The original problem of this code, that there is no synchronization between this mutex and the atomic<bool> still exists.)
For this to work properly you have to
 - make sure KafkaLease checks whether the connection is in use (with atomic_flag_test_and_set or just use a mutex), and if not, fails to construct
 - preferably make getOrCreateConnection itself return the lease and not delegate the responsibility of getting a lease to the user
 - make the user handle the failure of a KafkaLease creation (either wait or yield, as we discussed yielding here is the better option)

extensions/librdkafka/PublishKafka.cpp
```diff
@@ -319,27 +320,29 @@ void PublishKafka::onTrigger(const std::shared_ptr<core::ProcessContext> &contex
 
   std::string client_id, brokers, topic;
 
+  std::unique_ptr<KafkaLease> lease = nullptr;
```
bakaid: This is purely stylistic, but the default constructor of `std::unique_ptr<T>` already initializes the unique_ptr to own nothing, so assigning a `nullptr` is superfluous.

extensions/librdkafka/KafkaConnection.h
```diff
@@ -0,0 +1,217 @@
...
+#include "KafkaTopic.h"
+#include "KafkaPool.h"
+
+class KafkaConnectionKey {
+public:
+    std::string client_id_;
+    std::string brokers_;
+};
+
+struct KafkaConnectionComparator {
```
bakaid: I know this has been in the original code, but do you see any reason having a Comparator functor instead of KafkaConnectionKey having an `operator<`?

extensions/librdkafka/KafkaPool.h
```diff
@@ -0,0 +1,67 @@
...
+      : max_(max) {
+  }
+
+  bool removeConnection(const KafkaConnectionKey &key) {
+    std::lock_guard<std::mutex> lock(mutex_);
+    return map_.erase(key) == 1;
+  }
+
+  std::unique_ptr<KafkaLease> getOrCreateConnection(const KafkaConnectionKey &key) {
+    std::lock_guard<std::mutex> lock(mutex_);
```
bakaid: This logic has some duplication in it. I think it could be more elegant - and less error-prone - if you would unify the connection creation logic and the lease acquiring logic.
So first check whether a connection exists with this key, if not, create one, then you can have a common logic that uses `tryUse` on this connection. This way you could also avoid having to set the `lease_` to `true` in the KafkaConnection's constructor, which is hard to understand at first, I think.

nghiaxlee:
Yes, I did check it with a docker broker

extensions/librdkafka/KafkaConnection.cpp
```diff
@@ -0,0 +1,145 @@
...
+  }
+  initialized_ = false;
+}
+
+bool KafkaConnection::initialized() const {
+  return initialized_;
+}
+
+void KafkaConnection::setConnection(rd_kafka_t *producer, rd_kafka_conf_t *conf) {
+  std::lock_guard<std::mutex> lock(mutex_);
```
arpadboda: I think this is needless here.
The mutex is only responsible in this class for protecting lease_.

To make this a bit more clear, please either rename the mutex or add some comment at the declaration of the mutex to indicate the fact that this mutex is *not* responsible to make the usage of the class thread-fase.

extensions/librdkafka/KafkaConnection.h
```diff
@@ -0,0 +1,130 @@
...
+
+ private:
+
+  std::shared_ptr<logging::Logger> logger_;
+
+  std::mutex mutex_;
+
+  bool lease_;
+
+  std::atomic<bool> initialized_;
```
arpadboda: This seems to be a leftover. As the class is not thread-safe, having such a member is misleading, please keep it as a simple bool. Atomic include can go away as well.

extensions/librdkafka/KafkaConnection.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+rd_kafka_t *KafkaConnection::getConnection() {
+  return kafka_connection_;
+}
+
+bool KafkaConnection::hasTopic(const std::string &topic) const {
+  return topics_.count(topic);
+}
+
+std::shared_ptr<KafkaTopic> KafkaConnection::getTopic(const std::string &topic) {
```
arpadboda: This seems to be a const member function to me.

PR #632, opened 2019-08-20
MINIFICPP-1012 Docker symlink for current to point to versioned working directory

apiri:
*NOTE* This should follow #631 

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/docker/ContainerBuild.sh
+++ b/docker/ContainerBuild.sh
@@ -60,6 +60,6 @@ DOCKER_COMMAND="docker build --build-arg UID=$UID_ARG \
 echo "Docker Command: '$DOCKER_COMMAND'"
 ${DOCKER_COMMAND}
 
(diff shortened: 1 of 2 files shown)


phrocker:
Thanks @apiri . I went back and forth on the naming of this on whether we should call it nifi-minifi-agent or just keep it was it previously was. I did this to help address some factors regarding the windows service. Can address those without a name change, but thought these artifacts were better suited named something a little bit different. Do you think minifi-agent is fine or do you have another option? Name changes are pretty easy, so would love to hear some feedback before we released.

apiri:
@phrocker I don't have any overly strong preferences.  I am okay with them as named, although I did remember wondering if we were potentially gumming things up for folks that have tooling setup to deploy artifacts from a given mirror.  If we stay the course with the items as named, I will suggest we make a note in our release and/or provide migration notices about this.

apiri:
I do find myself thinking that we should keep the cpp in there.  It's a bit verbose but seems necessary until the Java impl is incorporated into NiFi as per MINIFI-422

PR #634, opened 2019-08-26
MINIFICPP-1017 - Improve logging of docker tests

arpadboda:

### For all changes:
     in the commit message?

### For code changes:

### Note:


docker/test/integration/minifi/test/__init__.py
```diff
@@ -118,27 +118,24 @@ def log_nifi_output(self):
...
-                    logging.info('MiNiFi app logs for container \'%s\':\n%s', container.name, minifi_app_logs)
-
-                nifi_app_logs = container.exec_run('/bin/sh -c \'test -f ' + self.nifi_root + '/logs/nifi-app.log '
-                                                                                            '&& cat ' +
-                                                   self.nifi_root + '/logs/nifi-app.log\'')
-                if len(nifi_app_logs) > 0:
-                    logging.info('NiFi app logs for container \'%s\':\n%s', container.name, nifi_app_logs)
+                for app in [("MiNiFi", self.minifi_root + '/logs/minifi-app.log'), ("NiFi", self.nifi_root + '/logs/nifi-app.log')]:
+                    app_log_status, app_log = container.exec_run('/bin/sh -c \'cat ' + app[1] + '\'')
+                    if app_log_status == 0:
```
bakaid: Could you display an error here if app_log_status is not 0? It seems like we silently ignore the error here.

arpadboda: When looking for container logs, we don't know whether the given container was running MiNiFi or NiFi, so we just look for the logs of both. Because of this having non-zero return code here is not really an error. 

Although at least one should be fine, so I added an "else" branch to the for loop to indicate and log error.

bakaid: Ah, thanks for the explanation!

PR #635, opened 2019-08-28
MINIFICPP-819 - OPC Unified Architecture Support

arpadboda:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/opc/include/fetchopc.h
+++ b/extensions/opc/include/fetchopc.h
@@ -59,7 +59,7 @@ public:
   static core::Relationship Failure;
 
   FetchOPCProcessor(std::string name, utils::Identifier uuid = utils::Identifier())
-  : BaseOPCProcessor(name, uuid) {
+  : BaseOPCProcessor(name, uuid), nameSpaceIdx_(0), nodesFound_(0), variablesFound_(0), maxDepth_(0) {
     logger_ = logging::LoggerFactory<FetchOPCProcessor>::getLogger();
   }
 
--- a/extensions/opc/include/fetchopc.h
+++ b/extensions/opc/include/fetchopc.h
@@ -1,0 +1,106 @@
+/**
+ * FetchOPC class declaration
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef NIFI_MINIFI_CPP_FetchOPCProcessor_H
+#define NIFI_MINIFI_CPP_FetchOPCProcessor_H
+
+#include <memory>
+#include <string>
+#include <list>
+#include <map>
+#include <mutex>
+#include <thread>
+
+#include "opc.h"
+#include "opcbase.h"
+#include "utils/ByteArrayCallback.h"
+#include "FlowFileRecord.h"
+#include "core/Processor.h"
+#include "core/ProcessSession.h"
+#include "core/Core.h"
+#include "core/Property.h"
+#include "core/Resource.h"
+#include "controllers/SSLContextService.h"
+#include "core/logging/LoggerConfiguration.h"
+#include "utils/Id.h"
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace processors {
+
+class FetchOPCProcessor : public BaseOPCProcessor {
+public:
+  static constexpr char const* ProcessorName = "FetchOPC";
+  // Supported Properties
+  static core::Property NodeIDType;
+  static core::Property NodeID;
+  static core::Property NameSpaceIndex;
+
+  // Supported Relationships
+  static core::Relationship Success;
+  static core::Relationship Failure;
+
+  FetchOPCProcessor(std::string name, utils::Identifier uuid = utils::Identifier())
+  : BaseOPCProcessor(logging::LoggerFactory<FetchOPCProcessor>::getLogger(), name, uuid) {
+  }
+
+  virtual void onSchedule(const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSessionFactory> &factory) override;
+
+  virtual void onTrigger(const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSession> &session) override;
+
+  virtual void initialize(void) override;
+
+protected:
+  bool nodeFoundCallBack(opc::ClientPtr& clientPtr, const UA_ReferenceDescription *ref, const std::string& path,
+                         const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSession> &session);
+
+  void OPCData2FlowFile(const opc::nodeData& opcnode, const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSession> &session);
+
+  class WriteCallback : public OutputStreamCallback {
+    std::string data_;
+   public:
+    WriteCallback(const std::string& data)
+        : data_(data) {
+    }
+    int64_t process(std::shared_ptr<io::BaseStream> stream) {
+      return stream->write(reinterpret_cast<uint8_t*>(const_cast<char*>(data_.c_str())), data_.size());
+    }
+  };
+  std::string nodeID_;
+  int32_t nameSpaceIdx_;
+  opc::OPCNodeIDType idType_;
+  uint32_t nodesFound_;
+  uint32_t variablesFound_;
+
+private:
+  std::mutex onTriggerMutex_;
+  std::vector<UA_NodeId> translatedNodeIDs_;  // Only used when user provides path, path->nodeid translation is only done once
+
+};
+
+REGISTER_RESOURCE(FetchOPCProcessor, "Fetches OPC-UA node");
+
+} /* namespace processors */
+} /* namespace minifi */
+} /* namespace nifi */
+} /* namespace apache */
+} /* namespace org */
+
+#endif //NIFI_MINIFI_CPP_FetchOPCProcessor_H
--- a/extensions/opc/include/fetchopc.h
+++ b/extensions/opc/include/fetchopc.h
@@ -52,13 +52,16 @@ public:
   static core::Property NodeIDType;
   static core::Property NodeID;
   static core::Property NameSpaceIndex;
+  static core::Property MaxDepth;
 
   // Supported Relationships
   static core::Relationship Success;
   static core::Relationship Failure;
 
   FetchOPCProcessor(std::string name, utils::Identifier uuid = utils::Identifier())
-  : BaseOPCProcessor(logging::LoggerFactory<FetchOPCProcessor>::getLogger(), name, uuid) {
+  : BaseOPCProcessor(name, uuid) {
+    logger_ = logging::LoggerFactory<FetchOPCProcessor>::getLogger();
+    connection_ = opc::ClientPtr(nullptr, std::bind(opc::disconnect, std::placeholders::_1, logger_));
   }
 
   virtual void onSchedule(const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSessionFactory> &factory) override;
@@ -88,6 +91,7 @@ protected:
   opc::OPCNodeIDType idType_;
   uint32_t nodesFound_;
   uint32_t variablesFound_;
+  uint64_t maxDepth_;
 
 private:
   std::mutex onTriggerMutex_;
--- a/extensions/opc/include/fetchopc.h
+++ b/extensions/opc/include/fetchopc.h
@@ -76,8 +76,8 @@ protected:
   class WriteCallback : public OutputStreamCallback {
     std::string data_;
    public:
-    WriteCallback(const std::string& data)
-        : data_(data) {
+    WriteCallback(std::string&& data)
+      : data_(data) {
     }
     int64_t process(std::shared_ptr<io::BaseStream> stream) {
       return stream->write(reinterpret_cast<uint8_t*>(const_cast<char*>(data_.c_str())), data_.size());
--- a/extensions/opc/include/fetchopc.h
+++ b/extensions/opc/include/fetchopc.h
@@ -71,7 +71,7 @@ protected:
   bool nodeFoundCallBack(opc::ClientPtr& clientPtr, const UA_ReferenceDescription *ref, const std::string& path,
                          const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSession> &session);
 
-  void OPCData2FlowFile(const opc::nodeData& opcnode, const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSession> &session);
+  void OPCData2FlowFile(const opc::NodeData& opcnode, const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSession> &session);
 
   class WriteCallback : public OutputStreamCallback {
     std::string data_;
--- a/extensions/opc/include/fetchopc.h
+++ b/extensions/opc/include/fetchopc.h
@@ -61,7 +61,6 @@ public:
   FetchOPCProcessor(std::string name, utils::Identifier uuid = utils::Identifier())
   : BaseOPCProcessor(name, uuid) {
     logger_ = logging::LoggerFactory<FetchOPCProcessor>::getLogger();
-    connection_ = opc::ClientPtr(nullptr, std::bind(opc::disconnect, std::placeholders::_1, logger_));
   }
 
   virtual void onSchedule(const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSessionFactory> &factory) override;
@@ -71,7 +70,7 @@ public:
   virtual void initialize(void) override;
 
 protected:
-  bool nodeFoundCallBack(opc::ClientPtr& clientPtr, const UA_ReferenceDescription *ref, const std::string& path,
+  bool nodeFoundCallBack(opc::Client& client, const UA_ReferenceDescription *ref, const std::string& path,
                          const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSession> &session);
 
   void OPCData2FlowFile(const opc::NodeData& opcnode, const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSession> &session);
--- a/extensions/opc/include/opc.h
+++ b/extensions/opc/include/opc.h
@@ -33,13 +33,45 @@ namespace nifi {
 namespace minifi {
 namespace opc {
 
-void disconnect(UA_Client *client, std::shared_ptr<core::logging::Logger> logger);
-
-using ClientPtr = std::unique_ptr<UA_Client, std::function<void(UA_Client *client)>>;
-
 enum class OPCNodeIDType{ Path, Int, String };
 
 enum class OPCNodeDataType{ Int64, UInt64, Int32, UInt32, Boolean, Float, Double, String };
+
+struct NodeData;
+
+class Client;
+
+using nodeFoundCallBackFunc = bool(Client& client, const UA_ReferenceDescription*, const std::string&);
+
+class Client {
+ public:
+  bool isConnected();
+  UA_StatusCode connect(const std::string& url, const std::string& username = "", const std::string& password = "");
+  ~Client();
+  NodeData getNodeData(const UA_ReferenceDescription *ref, const std::string& basePath = "");
+  UA_ReferenceDescription * getNodeReference(UA_NodeId nodeId);
+  void traverse(UA_NodeId nodeId, std::function<nodeFoundCallBackFunc> cb, const std::string& basePath = "", uint32_t maxDepth = 0, bool fetchRoot = true);
+  bool exists(UA_NodeId nodeId);
+  UA_StatusCode translateBrowsePathsToNodeIdsRequest(const std::string& path, std::vector<UA_NodeId>& foundNodeIDs, const std::shared_ptr<core::logging::Logger>& logger);
+
+  template<typename T>
+  UA_StatusCode update_node(const UA_NodeId nodeId, T value);
+
+  template<typename T>
+  UA_StatusCode add_node(const UA_NodeId parentNodeId, const UA_NodeId targetNodeId, std::string browseName, T value, OPCNodeDataType dt, UA_NodeId *receivedNodeId);
+
+ private:
+  Client (std::shared_ptr<core::logging::Logger> logger, const std::vector<char>& certBuffer, const std::vector<char>& keyBuffer, const std::vector<std::vector<char>>& trustBuffers);
+  UA_Client *client_;
+  std::shared_ptr<core::logging::Logger> logger_;
+
+  friend std::unique_ptr<Client> createClient(std::shared_ptr<core::logging::Logger> logger,
+                                              const std::vector<char>& certBuffer,
+                                              const std::vector<char>& keyBuffer,
+                                              const std::vector<std::vector<char>>& trustBuffers);
+};
+
+using ClientPtr = std::unique_ptr<Client>;
 
 struct NodeData {
   std::vector<uint8_t> data;
@@ -63,7 +95,7 @@ struct NodeData {
     rhs.var_ = nullptr;
   }
 
-private:
+ private:
   UA_Variant* var_;
 
   NodeData(UA_Variant * var = nullptr) {
@@ -76,7 +108,7 @@ private:
     var_ = var;
   }
 
-  friend NodeData getNodeData(opc::ClientPtr&, const UA_ReferenceDescription*, const std::string&);
+  friend class Client;
   friend std::string nodeValue2String(const NodeData&);
 };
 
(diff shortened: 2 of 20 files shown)


arpadboda:
Two things left to be done here:
-Finish tests (and add them to the PR)
-In the meanwhile newer version of open62541 was released, supporting client security. The code is expected to work with that, but the dependency needs to be upgraded.

libminifi/include/core/Property.h
```diff
@@ -174,6 +174,11 @@ class Property {
   void addAllowedValue(const PropertyValue &value) {
     allowed_values_.push_back(value);
   }
+
+  void clearAllowedValues() {
```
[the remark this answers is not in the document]
arpadboda: So the reason I did this is allowed_values_ being stored in a vector, so calling init twice would add the values twice. 
The nice solution would be storing them in set in my opinion. What do you think?

extensions/opc/src/opc.cpp
```diff
@@ -0,0 +1,461 @@
...
+  for(size_t i = 0; i < bResp.resultsSize; ++i) {
+    for(size_t j = 0; j < bResp.results[i].referencesSize; ++j) {
+      UA_ReferenceDescription *ref = &(bResp.results[i].references[j]);
+      if (cb(clientPtr, ref, basePath)) {
+        if (ref->nodeClass == UA_NODECLASS_VARIABLE || ref->nodeClass == UA_NODECLASS_OBJECT) {
+          std::string browsename((char *) ref->browseName.name.data, ref->browseName.name.length);
+          traverse(clientPtr, ref->nodeId.nodeId, cb, basePath + browsename);
+        }
+      } else {
+        UA_BrowseResponse_deleteMembers(&bResp);
```
bakaid: As this is repeated a few lines down, using a util::ScopeGuard would be less error-prone.

extensions/opc/src/opc.cpp
```diff
@@ -0,0 +1,461 @@
...
+  UA_TranslateBrowsePathsToNodeIdsRequest request;
+  UA_TranslateBrowsePathsToNodeIdsRequest_init(&request);
+  request.browsePaths = &browsePath;
+  request.browsePathsSize = 1;
+
+  UA_TranslateBrowsePathsToNodeIdsResponse response = UA_Client_Service_translateBrowsePathsToNodeIds(client, request);
+
+  if(response.resultsSize < 1) {
+    logger->log_warn("No node id in response for %s", path.c_str());
+    return UA_STATUSCODE_BADNODATAAVAILABLE;
```
bakaid: `UA_BrowsePath_deleteMembers(&browsePath);` is not called on this path, causing a leak on error. I suggest a utils::ScopeGuard here.

extensions/opc/src/opc.cpp
```diff
@@ -0,0 +1,461 @@
...
+    if(UA_Client_readValueAttribute(clientPtr.get(), ref->nodeId.nodeId, var) == UA_STATUSCODE_GOOD && var->type != NULL && var->data != NULL) {
+      nodedata.dataTypeID = var->type->typeIndex;
+      nodedata.addVariant(var);
+      if(var->type->typeName) {
+        nodedata.attributes["Typename"] = std::string(var->type->typeName);
+      }
+      if(var->type->memSize) {
+        nodedata.attributes["Datasize"] = std::to_string(var->type->memSize);
+        nodedata.data = std::vector<uint8_t>(var->type->memSize);
+        memcpy(&nodedata.data[0], var->data, var->type->memSize);
```
bakaid: Please use simply `nodedata.data()` here, it's a vector, not a string. If `var->type->memSize` is 0, this is undefined behaviour, also MSVC's debug iterator might complain.

extensions/opc/src/opc.cpp
```diff
@@ -0,0 +1,461 @@
...
+      nodedata.attributes["NodeID"] = nodeidstr;
+      nodedata.attributes["NodeID type"] = "bytestring";
+    } else if (ref->nodeId.nodeId.identifierType == UA_NODEIDTYPE_NUMERIC) {
+      nodedata.attributes["NodeID"] = std::to_string(ref->nodeId.nodeId.identifier.numeric);
+      nodedata.attributes["NodeID type"] = "numeric";
+    }
+    nodedata.attributes["Browsename"] = browsename;
+    nodedata.attributes["Full path"] = basePath + "/" + browsename;
+    nodedata.dataTypeID = UA_TYPES_COUNT;
+    UA_Variant* var = UA_Variant_new();
```
bakaid: This will leak if `UA_Client_readValueAttribute` fails.

extensions/opc/src/opc.cpp
```diff
@@ -0,0 +1,461 @@
...
+      throw Exception(OPC_EXCEPTION, "Data type is not supported ");
+  }
+  return ret_val;
+}
+
+std::string OPCDateTime2String(UA_DateTime raw_date) {
+  UA_DateTimeStruct dts = UA_DateTime_toStruct(raw_date);
+  std::array<char, 100> charBuf;
+
+  snprintf(&charBuf[0], charBuf.size(), "%u-%u-%u %u:%u:%u.%03u", dts.day, dts.month, dts.year, dts.hour, dts.min, dts.sec, dts.milliSec);
```
bakaid: UA_DateTimeStruct's member variables are uint16_t-s, so the proper format string for them is %hu.
Also, in the original UA_DateTime_toString values were zero-padded, which is easier to parse, so I have to agree with that.
I suggest the format string `"%02hu-%02hu-%02hu %02hu:%02hu:%02hu.%03hu"`

bakaid: Also, please consider using `charBuf.data()` instead of `&charBuf[0]`, it's nicer.

extensions/opc/src/opc.cpp
```diff
@@ -0,0 +1,461 @@
...
+
+  return std::string(charBuf.data(), charBuf.size());
+}
+
+} /* namespace opc */
+} /* namespace minifi */
+} /* namespace nifi */
+} /* namespace apache */
+} /* namespace org */
```
bakaid: This needs a newline here.

extensions/opc/include/opcbase.h
```diff
@@ -0,0 +1,80 @@
+/**
+ * FetchOPC class declaration
```
bakaid: This is OPCBase.

extensions/opc/src/fetchopc.cpp
```diff
@@ -0,0 +1,239 @@
...
+      core::PropertyBuilder::createProperty("Namespace index")
+      ->withDescription("The index of the namespace. Used only if node ID type is not path.")
+      ->withDefaultValue<int32_t>(0)->build());
+
+  core::Relationship FetchOPCProcessor::Success("success", "Successfully retrieved OPC-UA nodes");
+  core::Relationship FetchOPCProcessor::Failure("failure", "Retrieved OPC-UA nodes where value cannot be extracted (only if enabled)");
+
+
+  void FetchOPCProcessor::initialize() {
+    //BaseOPCProcessor::initialize();
```
bakaid: I think this is no longer needed.

extensions/opc/src/fetchopc.cpp
```diff
@@ -0,0 +1,239 @@
...
+    } else if (value == "Path") {
+      idType_ = opc::OPCNodeIDType::Path;
+    } else {
+      // Where have our validators gone?
+      logger_->log_error("%s is not a valid node ID type!", value.c_str());
+    }
+
+    if(idType_ == opc::OPCNodeIDType::Int) {
+      try {
+        int t = std::stoi(nodeID_);
```
bakaid: This doesn't perfectly validate whether nodeID_ is an integer, as std::stoi stops at the first character it couldn't convert, so to actually make sure, you have to use its `std::size_t* pos` out parameter to find out whether it consumed the entire string.
Not a big deal, though, I think it is a fair check as it is.

extensions/opc/src/fetchopc.cpp
```diff
@@ -0,0 +1,239 @@
...
+    std::unique_lock<std::mutex> lock(onTriggerMutex_, std::try_to_lock);
+    if(!lock.owns_lock()){
+      logger_->log_warn("processor was triggered before previous listing finished, configuration should be revised!");
+      return;
+    }
+
+    nodesFound_ = 0;
+    variablesFound_ = 0;
+
+    if(!opc::isConnected(connection_)) {
```
bakaid: The connection code is common between FetchOpc and PutOpc. I suggest moving this to the base class as a function, as `connection_` is already there, and we could avoid the mistake of changing one implementation and forgetting about the other.

extensions/opc/src/fetchopc.cpp
```diff
@@ -0,0 +1,239 @@
...
+    std::shared_ptr<FlowFileRecord> flowFile = std::static_pointer_cast<FlowFileRecord>(session->create());
+    if (flowFile == nullptr) {
+      logger_->log_error("Failed to create flowfile!");
+      return;
+    }
+    for(const auto& attr: opcnode.attributes) {
+      flowFile->setAttribute(attr.first, attr.second);
+    }
+    if(opcnode.data.size() > 0) {
+      //if (extractValue_) {
```
bakaid: I assume this is no longer needed.

extensions/opc/src/fetchopc.cpp
```diff
@@ -0,0 +1,239 @@
...
+      try {
+        std::string value = opc::nodeValue2String(opcnode);
+        FetchOPCProcessor::WriteCallback callback(value);
```
bakaid: Can't we avoid this copy here? WriteCallback could store a const std::string& instead of copying it.

extensions/opc/src/putopc.cpp
```diff
@@ -0,0 +1,478 @@
+/**
+ * FetchOPC class definition
```
bakaid: This is PutOPC.

extensions/opc/src/fetchopc.cpp
arpadboda: That's a fair point, but I would prefer to leave that as a follow-up in a generic way: create a wrapper around stoi in our utils and use that instead of stoi in such places. There are many I guess.

extensions/opc/src/fetchopc.cpp
arpadboda: Removed.

extensions/opc/src/opc.cpp
arpadboda: Good idea, added!

extensions/opc/src/opc.cpp
arpadboda: Good catch, added scopeguard.

extensions/opc/src/opc.cpp
arpadboda: 0 can't happen, that's the condition above.
Changed to data()

extensions/opc/src/opc.cpp
arpadboda: Added a delete call there.

extensions/opc/src/opc.cpp
arpadboda: Agreed, that looks better, changed.

extensions/opc/src/opc.cpp
arpadboda: Added.

libminifi/include/Exception.h
```diff
@@ -41,12 +41,13 @@ enum ExceptionType {
   SITE2SITE_EXCEPTION,
   GENERAL_EXCEPTION,
   REGEX_EXCEPTION,
+  OPC_EXCEPTION,
```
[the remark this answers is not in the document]
arpadboda: If you look at the usages of this exception (data type not supported, failed to read value of an OPC node, node type is not supported), it's used to handle OPC protocol related issues mostly, and used in opc.cpp, which doesn't contain processor-related code, so I think OPC exception is a better choice for that.

libminifi/include/Exception.h
```diff
@@ -41,12 +41,13 @@ enum ExceptionType {
...
   MAX_EXCEPTION
 };
 
 // Exception String 
 static const char *ExceptionStr[MAX_EXCEPTION] = { "File Operation", "Flow File Operation", "Processor Operation", "Process Session Operation", "Process Schedule Operation", "Site2Site Protocol",
-    "General Operation", "Regex Operation" };
+    "General Operation", "Regex Operation", "OPC_EXCEPTION" };
```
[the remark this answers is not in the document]
arpadboda: Fixed.

extensions/opc/src/putopc.cpp
arpadboda: Fixed

extensions/opc/src/putopc.cpp
```diff
@@ -0,0 +1,480 @@
...
+            sc = opc::add_node(connection_, parentNodeID_, targetnode, browsename, value, nodeDataType_, &resultnode);
+            break;
+          }
+          case opc::OPCNodeDataType::Double: {
+            double value = std::stod(contentstr);
+            sc = opc::add_node(connection_, parentNodeID_, targetnode, browsename, value, nodeDataType_, &resultnode);
+            break;
+          }
+          case opc::OPCNodeDataType::String: {
+            float value = std::stof(contentstr);
```
bakaid: I don't this is needed here.

extensions/opc/src/putopc.cpp
```diff
@@ -0,0 +1,480 @@
...
+    logger_->log_trace("Read %llu bytes from flowfile content to buffer", stream->getSize());
+
+    return size;
+  }
+
+} /* namespace processors */
+} /* namespace minifi */
+} /* namespace nifi */
+} /* namespace apache */
+} /* namespace org */
```
bakaid: Needs a newline.

extensions/opc/src/fetchopc.cpp
arpadboda: Just changed it to move the value to the callback, adding it as an attribute was just a leftover anyway.

extensions/opc/include/putopc.h
```diff
@@ -0,0 +1,133 @@
...
+  static core::Property ParentNodeID;
+  static core::Property ParentNameSpaceIndex;
+  static core::Property ValueType;
+
+  static core::Property TargetNodeIDType;
+  static core::Property TargetNodeID;
+  static core::Property TargetNodeBrowseName;
+  static core::Property TargetNodeNameSpaceIndex;
+
+  /*static core::Property Username;
```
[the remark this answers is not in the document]
arpadboda: Done

extensions/opc/include/putopc.h
```diff
@@ -0,0 +1,133 @@
...
+
+REGISTER_RESOURCE(PutOPCProcessor, "Creates/updates  OPC nodes");
+
+} /* namespace processors */
+} /* namespace minifi */
+} /* namespace nifi */
+} /* namespace apache */
+} /* namespace org */
+
+#endif  // NIFI_MINIFI_CPP_PUTOPC_H
```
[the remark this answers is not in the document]
arpadboda: Added

extensions/opc/src/putopc.cpp
```diff
@@ -0,0 +1,480 @@
...
+            sc = opc::update_node(connection_, targetnode, value);
+            break;
+          }
+          case opc::OPCNodeDataType::Double: {
+            double value = std::stod(contentstr);
+            sc = opc::update_node(connection_, targetnode, value);
+            break;
+          }
+          case opc::OPCNodeDataType::String: {
+            float value = std::stof(contentstr);
```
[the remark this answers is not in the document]
arpadboda: Removed.

extensions/opc/src/putopc.cpp
arpadboda: Removed

extensions/opc/src/putopc.cpp
arpadboda: Added

extensions/opc/src/opc.cpp
```diff
@@ -0,0 +1,480 @@
...
+      throw Exception(OPC_EXCEPTION, "Data type is not supported ");
+  }
+  return ret_val;
+}
+
+std::string OPCDateTime2String(UA_DateTime raw_date) {
+  UA_DateTimeStruct dts = UA_DateTime_toStruct(raw_date);
+  std::array<char, 100> charBuf;
+
+  snprintf(charBuf.data(), charBuf.size(), "%02hu-%02hu-%02hu %02hu:%02hu:%02hu.%03hu", dts.day, dts.month, dts.year, dts.hour, dts.min, dts.sec, dts.milliSec);
```
bakaid: And I messed this up, the year should be %04hu, which is not an issue, except for dates before 1000 CE, but still.

extensions/opc/src/opcbase.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+    context->getProperty(OPCServerEndPoint.getName(), endPointURL_);
+
+    if (context->getProperty(Username.getName(), username_) != context->getProperty(Password.getName(), password_)) {
+      logger_->log_error("Both or neither of username and password should be provided!");
+      return;
+    }
+
+    if (context->getProperty(CertificatePath.getName(), certpath_) !=
+        context->getProperty(Password.getName(), keypath_)) {
```
bakaid: This should be `KeyPath`.

extensions/opc/src/fetchopc.cpp
arpadboda: Added, thanks!

extensions/opc/src/opc.cpp
arpadboda: Fixed

extensions/opc/src/opcbase.cpp
arpadboda: Thanks, fixed!

libminifi/include/Exception.h
```diff
@@ -41,12 +41,13 @@ enum ExceptionType {
...
   GENERAL_EXCEPTION,
   REGEX_EXCEPTION,
+  OPC_EXCEPTION,
   MAX_EXCEPTION
 };
 
 // Exception String 
 static const char *ExceptionStr[MAX_EXCEPTION] = { "File Operation", "Flow File Operation", "Processor Operation", "Process Session Operation", "Process Schedule Operation", "Site2Site Protocol",
-    "General Operation", "Regex Operation" };
+    "General Operation", "Regex Operation", "OPC Operation" };
```
phrocker: Opc exception should be moved into ooc extension, and there should be no pollution in core

extensions/opc/include/opc.h
```diff
@@ -0,0 +1,123 @@
...
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace opc {
+
+void disconnect(UA_Client *client);
+
+using ClientPtr = std::unique_ptr<UA_Client, decltype(&opc::disconnect)>;
```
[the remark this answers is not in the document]
arpadboda: Done

extensions/opc/src/opc.cpp
```diff
@@ -0,0 +1,461 @@
...
+
+void setCertificates(ClientPtr& clientPtr, const std::vector<char>& certBuffer, const std::vector<char>& keyBuffer) {
+  UA_ClientConfig *cc = UA_Client_getConfig(clientPtr.get());
+  cc->securityMode = UA_MESSAGESECURITYMODE_SIGNANDENCRYPT;
+
+  UA_ByteString certByteString = UA_BYTESTRING_ALLOC(certBuffer.data());
+  UA_ByteString keyByteString = UA_BYTESTRING_ALLOC(keyBuffer.data());
+
+  UA_ClientConfig_setDefaultEncryption(cc, certByteString, keyByteString,
+                                       nullptr, 0,
```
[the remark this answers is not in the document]
arpadboda: Done

libminifi/include/Exception.h
arpadboda: I agree, there should be no pollution in the core. 

I moved the exception to the extension, but it's not so nice as Exception class in MiNiFi was not designed for being extended in any extension. (virtual functions? protected members?)

Because of this I find your comment a bit belated and misguided. Should had been made by the time such a class was designed. 

This topic definitely worth a follow-up ticket,  I'm open to discussions, suggestions.

arpadboda:
> Big Issue for me is pollution of core. Other than that we should wrap the mallocs in scope guards if possible.

Pollution was removed.
Scope guards were already introduced in scope of review-related commits.

extensions/opc/include/fetchopc.h
```diff
@@ -0,0 +1,109 @@
...
+  static core::Property NodeIDType;
+  static core::Property NodeID;
+  static core::Property NameSpaceIndex;
+  static core::Property MaxDepth;
+
+  // Supported Relationships
+  static core::Relationship Success;
+  static core::Relationship Failure;
+
+  FetchOPCProcessor(std::string name, utils::Identifier uuid = utils::Identifier())
```
bakaid: Please initialize the integral type members to 0 in the constructor. I know that we fill them in onSchedule, but a mistake in a future rewrite could change that, so I would like to be defensive here.

extensions/opc/src/fetchopc.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+    std::unique_lock<std::mutex> lock(onTriggerMutex_, std::try_to_lock);
+    if(!lock.owns_lock()){
+      logger_->log_warn("processor was triggered before previous listing finished, configuration should be revised!");
+      return;
+    }
+
+    if (!reconnect()) {
+      yield();
+      return;;
```
bakaid: Needs less semicolons 😄

extensions/opc/src/fetchopc.cpp
```diff
@@ -0,0 +1,235 @@
...
+      UA_NodeId myID;
+      myID.namespaceIndex = nameSpaceIdx_;
+      if(idType_ == opc::OPCNodeIDType::Int) {
+        myID.identifierType = UA_NODEIDTYPE_NUMERIC;
+        myID.identifier.numeric = std::stoi(nodeID_);
+      } else if (idType_ == opc::OPCNodeIDType::String) {
+        myID.identifierType = UA_NODEIDTYPE_STRING;
+        myID.identifier.string = UA_STRING_ALLOC(nodeID_.c_str());
+      }
+      connection_->traverse(myID, f, "", maxDepth_);
```
bakaid: traverse's maxDepth parameter is an uint32_t but we are supplying an uint64_t here.

extensions/opc/include/putopc.h
```diff
@@ -0,0 +1,111 @@
...
+  static core::Property TargetNodeIDType;
+  static core::Property TargetNodeID;
+  static core::Property TargetNodeBrowseName;
+  static core::Property TargetNodeNameSpaceIndex;
+
+  // Supported Relationships
+  static core::Relationship Success;
+  static core::Relationship Failure;
+
+  PutOPCProcessor(std::string name, utils::Identifier uuid = utils::Identifier())
```
bakaid: Please initialize integral typed members here too.

extensions/opc/src/putopc.cpp
```diff
@@ -0,0 +1,466 @@
...
+      core::PropertyBuilder::createProperty("Parent node namespace index")
+          ->withDescription("The index of the namespace. Used only if node ID type is not path.")
+          ->withDefaultValue<int32_t>(0)->build());
+
+  core::Property PutOPCProcessor::ValueType(
+      core::PropertyBuilder::createProperty("Value type")
+          ->withDescription("Set the OPC value type of the created nodes")
+          ->isRequired(true)->build());
+
+  core::Property PutOPCProcessor::TargetNodeIDType(
```
bakaid: This should have the same validator as ParentNodeIDType.

extensions/opc/src/opc.cpp
```diff
@@ -0,0 +1,567 @@
...
+  char local[6] = "en-US";
+  attr.displayName = UA_LOCALIZEDTEXT(local, const_cast<char*>(browseName.c_str()));
+  UA_StatusCode sc = UA_Client_addVariableNode(client_,
+                                               targetNodeId,
+                                               parentNodeId,
+                                               UA_NODEID_NUMERIC(0, OPCNodeDataTypeToTypeID(dt)),
+                                               UA_QUALIFIEDNAME(1, const_cast<char*>(browseName.c_str())),
+                                               UA_NODEID_NULL,
+                                               attr, receivedNodeId);
+  UA_Variant_delete(&attr.value);
```
bakaid: This causes a segfault because `attr.value` is on stack, and `UA_Variant_delete` tries to free it as well as its content. `UA_Variant_clear` seems to work fine here.

extensions/opc/src/putopc.cpp
arpadboda: Can't as it's flowfile-dependent. :(

extensions/opc/include/opc.h
```diff
@@ -0,0 +1,152 @@
...
+
+ private:
+  Client (std::shared_ptr<core::logging::Logger> logger, const std::string& applicationURI,
+      const std::vector<char>& certBuffer, const std::vector<char>& keyBuffer,
+      const std::vector<std::vector<char>>& trustBuffers);
+
+  UA_Client *client_;
+  std::shared_ptr<core::logging::Logger> logger_;
+
+  friend std::unique_ptr<Client> createClient(std::shared_ptr<core::logging::Logger> logger,
```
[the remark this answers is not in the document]
arpadboda: Good idea, moved there, thanks!

extensions/opc/include/opcbase.h
```diff
@@ -0,0 +1,86 @@
...
+  static core::Property ApplicationURI;
+  static core::Property Username;
+  static core::Property Password;
+  static core::Property CertificatePath;
+  static core::Property KeyPath;
+  static core::Property TrustedPath;
+
+  BaseOPCProcessor(std::string name, utils::Identifier uuid = utils::Identifier())
+  : Processor(name, uuid) {
+    connection_ = nullptr;
```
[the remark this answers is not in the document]
arpadboda: Done, removed.

extensions/opc/include/fetchopc.h
arpadboda: Done

extensions/opc/src/fetchopc.cpp
arpadboda: Wiped out every second one without mercy! :)

extensions/opc/src/fetchopc.cpp
arpadboda: Good spot! Changed the signature of traverse.
It happened because of the lack of uint32_t getValue specialisation, which made me change the type of maxDepth.

extensions/opc/include/putopc.h
arpadboda: Done.

extensions/opc/src/opc.cpp
arpadboda: Fixed, thanks!

extensions/opc/src/putopc.cpp
bakaid: 😿

arpadboda:
Tested both secure and insecure behaviour on U18, passed.
Merging with approval of @bakaid

phrocker:
@arpadboda did you address the changes I requested? Usually if we have PMC/committer with changes requested that's generally something on which we don't merge. If you did not address those issues can you create a blocker for 0.7.0 to revert those changes?

PR #636, opened 2019-08-29
Minificpp-780 - Change GenerateFlowFile to allow 0b content FlowFiles

arpadboda:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/standard-processors/processors/GenerateFlowFile.cpp
+++ b/extensions/standard-processors/processors/GenerateFlowFile.cpp
@@ -28,10 +28,7 @@
 #include <string>
 #include <set>
 #include <random>
-#ifdef WIN32
-#define srandom srand
-#define random rand
-#endif
+
 #include "utils/StringUtils.h"
 #include "core/ProcessContext.h"
 #include "core/ProcessSession.h"
@@ -43,7 +40,6 @@ namespace apache {
 namespace nifi {
 namespace minifi {
 namespace processors {
-const char *GenerateFlowFile::DATA_FORMAT_BINARY = "Binary";
 const char *GenerateFlowFile::DATA_FORMAT_TEXT = "Text";
 core::Property GenerateFlowFile::FileSize(
     core::PropertyBuilder::createProperty("File Size")->withDescription("The size of the file that will be used")->isRequired(false)->withDefaultValue<core::DataSizeValue>("1 kB")->build());
@@ -60,8 +56,8 @@ core::Property GenerateFlowFile::UniqueFlowFiles(
         ->isRequired(false)->withDefaultValue<bool>(true)->build());
 
 core::Relationship GenerateFlowFile::Success("success", "success operational on the flow record");
-const unsigned int TEXT_LEN = 90;
-static const char TEXT_CHARS[TEXT_LEN + 1] = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890!@#$%^&*()-_=+/?.,';:\"?<>\n\t ";
+
+static const char * TEXT_CHARS = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890!@#$%^&*()-_=+/?.,';:\"?<>\n\t ";
 
 void GenerateFlowFile::initialize() {
   // Set the supported properties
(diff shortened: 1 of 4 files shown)


arpadboda:
WARNING: this PR depends on #637 - that one should be merged first, otherwise tests included in this are going to fail.

arpadboda:
While checking this processor I found the following issues:
-In case batch count is greater than one, the flowfiles generated in one batch are not unique
-Handling of 0B size was error-prone: allocating zero size arrays, passing null pointers to callbacks, etc
-A potential memory segmentation violation in case the file size is not multiple of 4.

extensions/standard-processors/processors/GenerateFlowFile.cpp
```diff
@@ -77,7 +73,28 @@ void GenerateFlowFile::initialize() {
...
+    std::uniform_int_distribution<> distr(std::numeric_limits<char>::min(), std::numeric_limits<char>::max());
+    auto rand = std::bind(distr, eng);
+    std::generate_n(data.begin(), data.size(), rand);
+  }
+}
+
 void GenerateFlowFile::onTrigger(core::ProcessContext *context, core::ProcessSession *session) {
+  std::unique_lock<std::mutex> lock(onTriggerMutex_, std::try_to_lock);
+  if (!lock.owns_lock()) {
+    logger_->log_warn("processor was triggered before previous listing finished, configuration should be revised!");
```
phrocker: This error message seems like it applies elsewhere since we are generating data, not listing.

extensions/standard-processors/processors/GenerateFlowFile.cpp
```diff
@@ -98,67 +115,32 @@ void GenerateFlowFile::onTrigger(core::ProcessContext *context, core::ProcessSes
...
     }
-    for (uint64_t i = 0; i < batchSize; i++) {
-      // For each batch
-      std::shared_ptr<FlowFileRecord> flowFile = std::static_pointer_cast<FlowFileRecord>(session->create());
-      if (!flowFile)
-        return;
-      if (fileSize > 0)
+    if (fileSize > 0) {
+      if (uniqueFlowFile) {
+        std::vector<char> data(fileSize);
```
phrocker: If we don't have a unique flow file then we should follow what NiFi does, which is generate the data in onSchedule -- that would obviate the need for locking here and simplify the logic altogether. 

Take a look at 
https://github.com/apache/nifi/blob/master/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/GenerateFlowFile.java#L167

extensions/standard-processors/processors/GenerateFlowFile.h
```diff
@@ -40,55 +40,47 @@ class GenerateFlowFile : public core::Processor {
...
  public:
   // OnTrigger method, implemented by NiFi GenerateFlowFile
-  virtual void onTrigger(core::ProcessContext *context, core::ProcessSession *session);
+  virtual void onTrigger(core::ProcessContext *context, core::ProcessSession *session) override;
   // Initialize, over write by NiFi GenerateFlowFile
-  virtual void initialize(void);
+  virtual void initialize(void) override;
 
  protected:
+  std::vector<char> _data;
```
phrocker: If you are changing this class you should enforce the coding constraints and rename to data_.

extensions/standard-processors/processors/GenerateFlowFile.cpp
arpadboda: Removed completely as it became needless.

extensions/standard-processors/processors/GenerateFlowFile.cpp
arpadboda: Very nice idea, I love it, done!

extensions/standard-processors/processors/GenerateFlowFile.h
arpadboda: Good spot, done, thanks!

arpadboda:
@phrocker : comments addressed here, could you revisit?

extensions/standard-processors/processors/GenerateFlowFile.cpp
```diff
@@ -77,88 +73,67 @@ void GenerateFlowFile::initialize() {
...
-void GenerateFlowFile::onTrigger(core::ProcessContext *context, core::ProcessSession *session) {
-  uint64_t batchSize = 1;
-  bool uniqueFlowFile = true;
-  uint64_t fileSize = 1024;
-  bool textData = false;
+void generateData(std::vector<char>& data, bool textData = false) {  // NOLINT
+  std::random_device rd;
+  std::mt19937 eng(rd());
+  if (textData) {
+    std::uniform_int_distribution<> distr(0, TEXT_LEN);
```
bakaid: I know this is legacy, but it took me a while to figure out why this is right. The `TEXT_CHARS` array contains 91 elements, so `TEXT_LEN` in reality is not the length of the array, but the maximum index. I think the previous version, which used % TEXT_LEN was erroneus, and now that we generate indexes from [0, TEXT_LEN] we are good, but it is not easy to understand.
I'm OK with merging this, but it should be replaced with something more straightforward eventually.

phrocker: integer module should be 0 to n-1 inclusive, so erroneous because it's a max index and not a length? I don't necessarily care what the hold code did but this block does need some comments to clarify the change why it is how it is and if the names are misleading a clarification to the reason why it was wrong previously.

extensions/standard-processors/processors/GenerateFlowFile.h
```diff
@@ -40,55 +40,56 @@ class GenerateFlowFile : public core::Processor {
...
   static const char *DATA_FORMAT_TEXT;
   // Supported Relationships
   static core::Relationship Success;
   // Nest Callback Class for write stream
   class WriteCallback : public OutputStreamCallback {
    public:
-    WriteCallback(char *data, uint64_t size)
-        : _data(data),
-          _dataSize(size) {
+    WriteCallback(std::vector<char> && data) : data_(data) {
```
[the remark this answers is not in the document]
arpadboda: Added, thanks

extensions/standard-processors/processors/GenerateFlowFile.cpp
arpadboda: I replaced the array with a C string, so we don't need the length any more.

PR #643, opened 2019-09-06
MINIFICPP-14: Prune class names. Add option to config

phrocker:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/utils/ClassUtils.h
+++ b/libminifi/include/utils/ClassUtils.h
@@ -1,0 +1,43 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef LIBMINIFI_INCLUDE_UTILS_CLASSUTILS_H_
+#define LIBMINIFI_INCLUDE_UTILS_CLASSUTILS_H_
+
+#include <string>
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace utils {
+namespace ClassUtils {
+
+/**
+ * Shortens class names via the canonical representation
+ * @param class_name input class name
+ * @param out output class name that is shortened.
+ */
+extern void shortenClassName(const std::string &class_name, std::string &out);
+
+} /* namespace ClassUtils */
+} /* namespace utils */
+} /* namespace minifi */
+} /* namespace nifi */
+} /* namespace apache */
+} /* namespace org */
+
+#endif /* LIBMINIFI_INCLUDE_UTILS_CLASSUTILS_H_ */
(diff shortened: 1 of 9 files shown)


phrocker:
This is an old PR I created a while ago. I rebased, but should probably go through it myself to clean it up. 

Here is the example output

`
[2019-09-06 16:45:52.582] [o::a::n::m::FlowController] [info] Started Flow Controller
[2019-09-06 16:45:52.582] [o::a::n::m::c::r::FlowFileRepository] [trace] Could not open checkpoint as object doesn't exist. Likely not needed or file system error.
[2019-09-06 16:45:52.582] [o::a::n::m::p::GetFile] [debug] Is listing empty 1
[2019-09-06 16:45:52.582] [o::a::n::m::c::ConfigurableComponent] [debug] Component Input Directory property name Input Directory value ***
[2019-09-06 16:45:52.582] [o::a::n::m::c::ProcessContextExpr] [debug] Compiling expression for PutFile/Input Directory: *****
[2019-09-06 16:45:52.582] [main] [info] MiNiFi started

`

libminifi/src/utils/ClassUtils.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include "utils/StringUtils.h"
+#include <iostream>
+#include <string>
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace utils {
+
+void ClassUtils::shortenClassName(const std::string &class_name, std::string &out) {
```
arpadboda: In case it's void, what's the motivation behind an output parameter, why not just return it?

libminifi/include/utils/ClassUtils.h
```diff
@@ -0,0 +1,43 @@
...
+namespace minifi {
+namespace utils {
+namespace ClassUtils {
+
+/**
+ * Shortens class names via the canonical representation
+ * @param class_name input class name
+ * @param out output class name that is shortened.
+ */
+extern void shortenClassName(const std::string &class_name, std::string &out);
```
arpadboda: Why do we need extern here? Libminifi build should be wholearchive anyway.

libminifi/include/utils/ClassUtils.h
```diff
@@ -0,0 +1,43 @@
...
+#define LIBMINIFI_INCLUDE_UTILS_CLASSUTILS_H_
+
+#include <string>
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace utils {
+namespace ClassUtils {
```
arpadboda: This is cool to have, I like it!
I also think some of the tempalte functions we already have (```getClassName<T>``` and ```getClassNames<T>```) could also be moved to here from classloader.

libminifi/include/utils/ClassUtils.h
phrocker: Thanks for pointing this out! It's not needed as is superfluous in header files.

 It was copied from the original location ( the source file ), in which case the use of wholarchive didn't help locate during the linking phase -- but when rebasing this PR I moved this to the header, at which point it's actually already externalized across translation units.

libminifi/src/utils/ClassUtils.cpp
phrocker: There are several pros and cons but following suit for some of the other API calls I should have returned true/false if shortening occurred. I like the idea of using output variables and since I'm already following other API utils classes I'll make that change. Thanks!

libminifi/src/utils/ClassUtils.cpp
```diff
@@ -26,13 +26,19 @@ namespace nifi {
...
+      class_delim = ".";
+      class_split = utils::StringUtils::split(class_name, class_delim);
+    } else {
+      // if no update can be performed, return false to let the developer know
+      // this. Out will have no updates
+      return false;
+    }
   }
   for (auto &elem : class_split) {
     if (&elem != &class_split.back() && elem.size() > 1) {
```
arpadboda: Tricky :)

libminifi/include/utils/ClassUtils.h
phrocker: Good idea. made a follow on.

PR #645, opened 2019-09-10
MINIFICPP-1028 - http url parsing without port fails

arpadboda:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/utils/HTTPClient.h
+++ b/libminifi/include/utils/HTTPClient.h
@@ -356,8 +356,8 @@ public:
 
 extern std::string get_token(utils::BaseHTTPClient *client, std::string username, std::string password);
 
-extern void parse_url(std::string *url, std::string *host, int *port, std::string *protocol);
-extern void parse_url(std::string *url, std::string *host, int *port, std::string *protocol, std::string *path, std::string *query);
+extern void parse_url(const std::string *url, std::string *host, int *port, std::string *protocol);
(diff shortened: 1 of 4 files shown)


bakaid:
> Thanks! We have this in a few places throughout the project, but those may be for testing. We've had this issue for other cases that tripped us up -- makes me wonder if we could just use civetweb for something like this.

cURL has a dedicated URL API, that can parse URLs: https://github.com/curl/curl/blob/master/include/curl/urlapi.h
However, that would require everything that would want to use this to depend on cURL, and while I am all for not using homecooked parsers, I am not sure whether that's something we want.

phrocker:
> > Thanks! We have this in a few places throughout the project, but those may be for testing. We've had this issue for other cases that tripped us up -- makes me wonder if we could just use civetweb for something like this.
> 
> cURL has a dedicated URL API, that can parse URLs: https://github.com/curl/curl/blob/master/include/curl/urlapi.h
> However, that would require everything that would want to use this to depend on cURL, and while I am all for not using homecooked parsers, I am not sure whether that's something we want.

Yeah a few of us previously discussed breaking that out -- then someone suggested using civet's parsers. Civet, at the time, was not an extension; however, now that we have a 'classloader' like capability we could specialize the parser based on extensions and have URLParser use a variant that relies on curl if it is built. cURL support will be quite prevalent, so we would rely on their more tested parser. That would mean we must still maintain ours, but have clear documentation that using our parser may have risk versus the one in curl-land.

PR #646, opened 2019-09-17
MINIFICPP-1031 - RemoteProcessorGroupPort::setURL fails in case port …

arpadboda:
…is not specified in URL

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/RemoteProcessorGroupPort.h
+++ b/libminifi/include/RemoteProcessorGroupPort.h
@@ -146,7 +146,7 @@ class RemoteProcessorGroupPort : public core::Processor {
     for (auto url : urls) {
       logger_->log_trace("Parsing %s", url);
       std::string host, protocol;
-      int port;
+      int port = -1;
       url = utils::StringUtils::trim(url);
       utils::parse_url(&url, &host, &port, &protocol);
       logger_->log_trace("Parsed -%s- %s %s, %d", url, protocol, host, port);


libminifi/include/RemoteProcessorGroupPort.h
```diff
@@ -146,7 +146,7 @@ class RemoteProcessorGroupPort : public core::Processor {
     for (auto url : urls) {
       logger_->log_trace("Parsing %s", url);
       std::string host, protocol;
-      int port;
+      int port = -1;
```
bakaid: Nice catch!

I wonder what happens if the input is really messed up and the protocol is not http or https. We would add an RPG to nifi_instances_ with a port of -1. Is that handled somewhere else?

arpadboda: In case the protocol is nor http neither https (but it's specified) parse_util fails, both host and protocol remains empty. 
Not nice, I would also prefer an exception to be thrown, but that would break this API, so I didn't change it.

bakaid: Thanks for the explanation, I agree that this was the better choice in this case. Do you think you could create a follow-up task to refactor this part when we get the chance?

arpadboda: Sure, created:
https://issues.apache.org/jira/browse/MINIFICPP-1032

PR #648, opened 2019-09-18
WIP: Minificpp 1025 -- add metadata

phrocker:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/windows-event-log/ConsumeWindowsEventLog.cpp
+++ b/extensions/windows-event-log/ConsumeWindowsEventLog.cpp
@@ -123,8 +123,8 @@ core::Property ConsumeWindowsEventLog::EventHeaderDelimiter(
 core::Property ConsumeWindowsEventLog::EventHeader(
 	core::PropertyBuilder::createProperty("Event Header")->
 	isRequired(false)->
-	withDefaultValue("LOG_NAME=Log Name, SOURCE = Source, TIME_CREATED = Date,EVENT_RECORDID=Record ID,EVENTID = Event ID,TASK_CATEGORY = Task Category,LEVEL = Level,KEYWORDS = Keywords,USER = User,COMPUTER = Computer")->
-	withDescription("Comma seperated list of key/value pairs with the following keys LOG_NAME, SOURCE, TIME_CREATED,EVENT_RECORDID,EVENTID,TASK_CATEGORY,LEVEL,KEYWORDS,USER,and COMPUTER. Eliminating fields will remove them from the header.")->
+	withDefaultValue("LOG_NAME=Log Name, SOURCE = Source, TIME_CREATED = Date,EVENT_RECORDID=Record ID,EVENTID = Event ID,TASK_CATEGORY = Task Category,LEVEL = Level,KEYWORDS = Keywords,USER = User,COMPUTER = Computer, EVENT_TYPE = EventType")->
+	withDescription("Comma seperated list of key/value pairs with the following keys LOG_NAME, SOURCE, TIME_CREATED,EVENT_RECORDID,EVENTID,TASK_CATEGORY,LEVEL,KEYWORDS,USER,COMPUTER, and EVENT_TYPE. Eliminating fields will remove them from the header.")->
 	build());
 
 core::Relationship ConsumeWindowsEventLog::Success("success", "Relationship for successfully consumed events.");
@@ -157,7 +157,7 @@ bool ConsumeWindowsEventLog::insertHeaderName(wel::METADATA_NAMES &header, const
 	wel::METADATA name = wel::WindowsEventLogMetadata::getMetadataFromString(key);
 		
 	if (name != wel::METADATA::UNKNOWN) {
-		header.insert(std::make_pair(name, value));
+		header.emplace_back(std::make_pair(name, value));
 		return true;
 	}
 	else {
@@ -181,6 +181,12 @@ void ConsumeWindowsEventLog::onSchedule(const std::shared_ptr<core::ProcessConte
 			auto key = utils::StringUtils::trim(splitKeyAndValue.at(0));
 			auto value = utils::StringUtils::trim(splitKeyAndValue.at(1));
 			if (!insertHeaderName(header_names_, key, value)) {
+				logger_->log_debug("%s is an invalid key for the header map", key);
+			}
+		}
+		else if (splitKeyAndValue.size() == 1) {
+			auto key = utils::StringUtils::trim(splitKeyAndValue.at(0));
+			if (!insertHeaderName(header_names_, key, "")) {
 				logger_->log_debug("%s is an invalid key for the header map", key);
 			}
 		}
@@ -325,6 +331,7 @@ bool ConsumeWindowsEventLog::subscribe(const std::shared_ptr<core::ProcessContex
 					log_header.setDelimiter(pConsumeWindowsEventLog->header_delimiter_);
 					// render the header.
 					renderedData.rendered_text_ = log_header.getEventHeader(&walker);
+					renderedData.rendered_text_ += "Message:\n";
 					renderedData.rendered_text_ += message;
 				}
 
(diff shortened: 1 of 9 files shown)


phrocker:
Merged based on @am-c-p-p's +1

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -329,7 +404,14 @@ int ConsumeWindowsEventLog::processQueue(const std::shared_ptr<core::ProcessSess
...
     session->getProvenanceReporter()->receive(flowFile, provenanceUri_, getUUIDStr(), "Consume windows event logs", 0);
     session->transfer(flowFile, Success);
-    session->commit();
+
+	flowFile = session->create();
+
+	session->write(flowFile, &WriteCallback(evt.rendered_text_));
+	session->putAttribute(flowFile, FlowAttributeKey(MIME_TYPE), "text/plain");
+	session->getProvenanceReporter()->receive(flowFile, provenanceUri_, getUUIDStr(), "Consume windows event logs", 0);
+	session->transfer(flowFile, Success);
```
arpadboda: I think producing two different flowfiles and transferring them to the same relationship is far from ideal. 
First it requires an additional processor to separate them, but also produces two flowfiles from one event. As the user most probably needs one or the other, this results in a lot of needless IO ops (writing flowfiles to content repo) and filling the disk without any value added. 
To resolve I would either:

- create two different relationships (text and xml) and route them accordingly. Not the best as it solves only one part of the issue. 
- introduce a processor property called "mode" or sg alike, which can be set to "text", "xml" or "both". This would support the current behaviour as well, and it case the user selects this, his is free to split them using routeonattribute, but at least we don't force that. 

As this processor is not released yet, we don't break anything. I would really appreciate to put more effort in this now and avoid being limited by backward compatibility later. 
@bakaid @am-c-p-p @phrocker @apiri - please don't hesitate to share your opinion!

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -329,7 +404,14 @@ int ConsumeWindowsEventLog::processQueue(const std::shared_ptr<core::ProcessSess
...
+	session->commit();
 
     flowFileCount++;
```
arpadboda: This is incorrect as we produce twice the amount, but that doesn't hurt as we ignore the return value where this function is called. So the question is: do we need it at all?

extensions/windows-event-log/wel/MetadataWalker.cpp
```diff
@@ -17,21 +19,62 @@ bool MetadataWalker::for_each(pugi::xml_node &node) {
...
 			if (std::regex_match(attr.name(), regex_)) {
 				std::function<std::string(const std::string &)> idUpdate = [&](const std::string &input) -> std::string {
-					return resolve_ ? utils::OsUtils::userIdToUsername(input) : input;
+					if (resolve_) {
+						auto resolved = utils::OsUtils::userIdToUsername(input);
+						replaced_identifiers_[input] = resolved;
+						return resolved;
+					}
+					else {
+						replaced_identifiers_[input] = input;
```
arpadboda: This else branch has no return value. Which surely triggers a warning, but I'm afraid it's an undefined behaviour.

extensions/windows-event-log/wel/MetadataWalker.cpp
```diff
@@ -17,21 +19,62 @@ bool MetadataWalker::for_each(pugi::xml_node &node) {
...
+					}
 				};
 				updateText(node, attr.name(), std::move(idUpdate));
 
 			}
+
 			if (std::regex_match(attr.value(), regex_)) {
 				std::function<std::string(const std::string &)> idUpdate = [&](const std::string &input) -> std::string {
```
arpadboda: We have exactly the same wrong function here. 
We should avoid moving and only create it once.

extensions/windows-event-log/wel/WindowsEventLog.cpp
```diff
@@ -0,0 +1,200 @@
...
+		}
+		else {
+			return returnValue;
+		}
+	}
+	else if (ERROR_EVT_MESSAGE_NOT_FOUND == status || ERROR_EVT_MESSAGE_ID_NOT_FOUND == status) {
+		return returnValue;
+	}
+	else {
+		return returnValue;
```
arpadboda: I see only one successful path here , where actual value is returned, all the rest are just returning empty string. 
This function should be simplified a bit but I think it would also benefit from some logging introduced.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -127,23 +146,61 @@ ConsumeWindowsEventLog::~ConsumeWindowsEventLog() {
...
+				logger_->log_debug("%s is an invalid key for the header map", key);
+			}
+		}
+		else if (splitKeyAndValue.size() == 1) {
+			auto key = utils::StringUtils::trim(splitKeyAndValue.at(0));
+			if (!insertHeaderName(header_names_, key, "")) {
+				logger_->log_debug("%s is an invalid key for the header map", key);
+			}
+		}
+
```
arpadboda: Are we sure this can't happen? Either a comment here or some error handling that at least logs malformed headers would be nice.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -108,6 +112,21 @@ core::Property ConsumeWindowsEventLog::ResolveAsAttributes(
 	withDescription("If true, any metadata that is resolved ( such as IDs or keyword metadata ) will be placed into attributes, otherwise it will be replaced in the XML or text output")->
 	build());
 
+
+core::Property ConsumeWindowsEventLog::EventHeaderDelimiter(
+	core::PropertyBuilder::createProperty("Event Header Delimiter")->
```
arpadboda: Guess we don't run linter for this module. This would surely break it.

phrocker: no linter in windows.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
phrocker: yep. @am-c-p-p can you do this?

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
phrocker: I don't agree and think that we want to keep in line with the success output which is what we have in java cwel. that was a pain point for those. i don't think what you propose solves your issues.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
phrocker: i think that's vestigial from what @am-c-p-p originally produced. should be a good follow on task Alex.

extensions/windows-event-log/wel/MetadataWalker.cpp
phrocker: @am-c-p-p can you take a look in this to correct?

extensions/windows-event-log/wel/MetadataWalker.cpp
phrocker: @am-c-p-p this too. thanks @arpadboda

extensions/windows-event-log/wel/WindowsEventLog.cpp
phrocker: don't disagree @am-c-p-p can you take a look at this comment too?

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
phrocker: I will add that I think @am-c-p-p is taking the reigns on this and he can disagree, but we should be cognizant of creating confusion between java and cpp variants. this I think yields the lowest confusion with allowing downstream processors to split or take care of this. that is why roa exists.

arpadboda: @phrocker : if you check the second bulletpoint above, that's about keeping success, but making the output configurable.
Don't get me wrong, I have no issue with keeping the current behavior as an option, maybe even as a default option, but I don't feel it right to force the user to use ROA in case he only needs one and not the other.
That's why I would introduce a property with 3 options:

- Both (default, current behaviour)
- Text only
- XML only

To keep aligned with NiFi, I'm happy to add this feature to the Java codebase as well.

PR #650, opened 2019-09-19
MINIFICPP-1034 - MQTT processors doesnt work

arpadboda:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/mqtt/processors/AbstractMQTTProcessor.cpp
+++ b/extensions/mqtt/processors/AbstractMQTTProcessor.cpp
@@ -46,32 +46,15 @@ core::Property AbstractMQTTProcessor::SecurityCA("Security CA", "File or directo
 core::Property AbstractMQTTProcessor::SecurityCert("Security Cert", "Path to client's public key (PEM) used for authentication", "");
 core::Property AbstractMQTTProcessor::SecurityPrivateKey("Security Private Key", "Path to client's private key (PEM) used for authentication", "");
 core::Property AbstractMQTTProcessor::SecurityPrivateKeyPassWord("Security Pass Phrase", "Private key passphrase", "");
-core::Relationship AbstractMQTTProcessor::Success("success", "FlowFiles that are sent successfully to the destination are transferred to this relationship");
-core::Relationship AbstractMQTTProcessor::Failure("failure", "FlowFiles that failed to send to the destination are transferred to this relationship");
 
-void AbstractMQTTProcessor::initialize() {
-  // Set the supported properties
-  std::set<core::Property> properties;
-  properties.insert(BrokerURL);
-  properties.insert(CleanSession);
-  properties.insert(ClientID);
-  properties.insert(UserName);
-  properties.insert(PassWord);
-  properties.insert(KeepLiveInterval);
-  properties.insert(ConnectionTimeOut);
-  properties.insert(QOS);
-  properties.insert(Topic);
-  setSupportedProperties(properties);
-  // Set the supported relationships
-  std::set<core::Relationship> relationships;
-  relationships.insert(Success);
-  relationships.insert(Failure);
-  setSupportedRelationships(relationships);
-  MQTTClient_SSLOptions sslopts_ = MQTTClient_SSLOptions_initializer;
-  sslEnabled_ = false;
+const std::set<core::Property> AbstractMQTTProcessor::getSupportedProperties() {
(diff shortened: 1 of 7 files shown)


extensions/mqtt/processors/AbstractMQTTProcessor.cpp
```diff
@@ -46,32 +46,15 @@ core::Property AbstractMQTTProcessor::SecurityCA("Security CA", "File or directo
...
-  properties.insert(ConnectionTimeOut);
-  properties.insert(QOS);
-  properties.insert(Topic);
-  setSupportedProperties(properties);
-  // Set the supported relationships
-  std::set<core::Relationship> relationships;
-  relationships.insert(Success);
-  relationships.insert(Failure);
-  setSupportedRelationships(relationships);
-  MQTTClient_SSLOptions sslopts_ = MQTTClient_SSLOptions_initializer;
```
arpadboda: This was just hiding instead of initializing the member. So it had no chance to work, but it didn't pop up given the security properties are not exported, so it's never referenced.

extensions/mqtt/processors/AbstractMQTTProcessor.cpp
```diff
@@ -46,32 +46,15 @@ core::Property AbstractMQTTProcessor::SecurityCA("Security CA", "File or directo
...
-  sslEnabled_ = false;
```
arpadboda: Too bad it was only initialized here, but none of the baseclasses called this function.

extensions/mqtt/processors/AbstractMQTTProcessor.cpp
```diff
@@ -46,32 +46,15 @@ core::Property AbstractMQTTProcessor::SecurityCA("Security CA", "File or directo
 core::Property AbstractMQTTProcessor::SecurityCert("Security Cert", "Path to client's public key (PEM) used for authentication", "");
 core::Property AbstractMQTTProcessor::SecurityPrivateKey("Security Private Key", "Path to client's private key (PEM) used for authentication", "");
 core::Property AbstractMQTTProcessor::SecurityPrivateKeyPassWord("Security Pass Phrase", "Private key passphrase", "");
```
arpadboda: These are not exported, but clearly never been tested as well, so created https://issues.apache.org/jira/browse/MINIFICPP-1035 to address this.

extensions/mqtt/processors/AbstractMQTTProcessor.h
```diff
@@ -28,16 +28,16 @@
...
-#define MQTT_QOS_1 "1"
-#define MQTT_QOS_2 "2"
-
 namespace org {
 namespace apache {
 namespace nifi {
 namespace minifi {
 namespace processors {
 
+#define MQTT_QOS_0 "0"
```
bakaid: I agree that these would be better at a namespace/class scope, but defines are not influenced by namespaces, so this change has no effect. Perhaps a constexpr?

extensions/mqtt/processors/PublishMQTT.cpp
```diff
@@ -76,15 +67,15 @@ void PublishMQTT::onSchedule(core::ProcessContext *context, core::ProcessSession
 }
 
 void PublishMQTT::onTrigger(const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSession> &session) {
-  std::shared_ptr<core::FlowFile> flowFile = session->get();
-
-  if (!flowFile) {
+  if (!reconnect()) {
```
bakaid: This way we reconnect even if there is no FlowFile to publish. I am not saying that it is necessarily a problem, but it changes the previous behaviour. Is this intentional?

arpadboda: Reconnect doesn't do anything unless the connection is interrupted. 
If we have an existing connection (expected case), it just returns true.

extensions/mqtt/processors/AbstractMQTTProcessor.h
arpadboda: Just wanted to put the definitions together, but constexpr is a fair point.
Changed them.

extensions/mqtt/processors/PublishMQTT.cpp
bakaid: Yeah, that's the conclusion I reached as well, should be fine, thanks.

PR #651, opened 2019-09-20
MINIFICPP-1033 - PublishKafka sends success flow even though kafka se…

nghiaxlee:
…rver is not running

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/librdkafka/PublishKafka.cpp
+++ b/extensions/librdkafka/PublishKafka.cpp
@@ -345,6 +345,21 @@ void PublishKafka::onTrigger(const std::shared_ptr<core::ProcessContext> &contex
       }
     }
 
+    // Check if brokers are down, if yes then yield.
+    const struct rd_kafka_metadata *metadata;
+    /* Fetch metadata */
+    // TODO: What is the time complexity of this??
+    rd_kafka_resp_err_t err = rd_kafka_metadata(conn->getConnection(), 0, nullptr,
+                            &metadata, 500);
+    if (err != RD_KAFKA_RESP_ERR_NO_ERROR) {
+      logger_->log_error("Failed to acquire metadata: %s\n", rd_kafka_err2str(err));
+      session->transfer(flowFile, Failure);
+      return;
+    } else {
+      logger_->log_debug("There are %d brokers", metadata->broker_cnt);
+      rd_kafka_metadata_destroy(metadata);
+    }
+
     if (!conn->hasTopic(topic)) {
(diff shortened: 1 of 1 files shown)


extensions/librdkafka/PublishKafka.cpp
```diff
@@ -345,6 +345,21 @@ void PublishKafka::onTrigger(const std::shared_ptr<core::ProcessContext> &contex
...
 
+    // Check if brokers are down, if yes then yield.
+    const struct rd_kafka_metadata *metadata;
+    /* Fetch metadata */
+    // TODO: What is the time complexity of this??
+    rd_kafka_resp_err_t err = rd_kafka_metadata(conn->getConnection(), 0, nullptr,
+                            &metadata, 500);
+    if (err != RD_KAFKA_RESP_ERR_NO_ERROR) {
+      logger_->log_error("Failed to acquire metadata: %s\n", rd_kafka_err2str(err));
+      session->transfer(flowFile, Failure);
```
arpadboda: Honestly I'm not sure what's the right thing to do here.

By default I would prefer to yield and not to anything in case the operation cannot be performed, although broker and topic are flowfile-dependent values here, so we have to handle invalid values nicely. 
Maybe instead of routing to failure I would consider penalizing the flowfile. 

With penalization we give 2nd chance to transmit in case of connection errors or any other issues at the remote side, although we are going to keep flowfiles with invalid attributes in our queue forever. 

Optionally we could add a map<uuid, uint> to count the penalization for flowfiles are route to failure after a threshold is exceeded.

@bakaid what do you think?

bakaid: @aboda `rd_kafka_metadata` is not a very good way to the determine whether a particular message actually got delivered to kafka. It might offer some protection in the negative case - if we don't have any brokers, we don't even try to send the message -, but in the positive case, even if we pass this check, librdkafka is asynchronous and there is a network call after that, meaning that the fact that we have brokers at this point does not at all mean that the message will be delivered successfully. If this is not a too expensive call, it might offer some benefits (in the negative case), but then it is definitely a yield case in my view.
@nghiaxlee is working on a `rd_kafka_conf_set_dr_msg_cb` based solution, which provides per-message delivery info. If it works the way we figured it works, it will add some complexity to the code, but it's not too bad, and it is the way to go - as I said, I think giving no information is better than giving bad information resulting in data loss or duplication.

PR #653, opened 2019-09-25
MINIFICPP-1033 - PublishKafka fixes

bakaid:
Fixes:
 - We now use delivery completion callbacks to properly determine successful delivery
 - The 'Delivery Guarantee' property had an invalid default value
 - Most of the configuration errors from librdkafka were silently ignored
 - rd_kafka_conf_t and rd_kafka_topic_conf_t objects were improperly destructed
   after their ownership has been taken by rd_kafka_t and rd_kafka_topic_t, respectively

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/librdkafka/PublishKafka.cpp
+++ b/extensions/librdkafka/PublishKafka.cpp
@@ -184,7 +184,7 @@ bool PublishKafka::configureNewConnection(const std::shared_ptr<KafkaConnection>
   result = rd_kafka_conf_set(conf_, "bootstrap.servers", key->brokers_.c_str(), errstr.data(), errstr.size());
   logger_->log_debug("PublishKafka: bootstrap.servers [%s]", key->brokers_);
   if (result != RD_KAFKA_CONF_OK) {
-    logger_->log_error("PublishKafka: configure error result [%s]", errstr);
+    logger_->log_error("PublishKafka: configure error result [%s]", errstr.data());
     return false;
   }
 
@@ -195,7 +195,7 @@ bool PublishKafka::configureNewConnection(const std::shared_ptr<KafkaConnection>
   result = rd_kafka_conf_set(conf_, "client.id", key->client_id_.c_str(), errstr.data(), errstr.size());
   logger_->log_debug("PublishKafka: client.id [%s]", key->client_id_);
   if (result != RD_KAFKA_CONF_OK) {
-    logger_->log_error("PublishKafka: configure error result [%s]", errstr);
+    logger_->log_error("PublishKafka: configure error result [%s]", errstr.data());
     return false;
   }
 
@@ -204,7 +204,7 @@ bool PublishKafka::configureNewConnection(const std::shared_ptr<KafkaConnection>
     result = rd_kafka_conf_set(conf_, "debug", value.c_str(), errstr.data(), errstr.size());
     logger_->log_debug("PublishKafka: debug [%s]", value);
     if (result != RD_KAFKA_CONF_OK) {
-      logger_->log_error("PublishKafka: configure debug error result [%s]", errstr);
+      logger_->log_error("PublishKafka: configure debug error result [%s]", errstr.data());
       return false;
     }
   }
@@ -213,7 +213,7 @@ bool PublishKafka::configureNewConnection(const std::shared_ptr<KafkaConnection>
     result = rd_kafka_conf_set(conf_, "sasl.kerberos.service.name", value.c_str(), errstr.data(), errstr.size());
     logger_->log_debug("PublishKafka: sasl.kerberos.service.name [%s]", value);
     if (result != RD_KAFKA_CONF_OK) {
-      logger_->log_error("PublishKafka: configure error result [%s]", errstr);
+      logger_->log_error("PublishKafka: configure error result [%s]", errstr.data());
       return false;
     }
   }
@@ -222,7 +222,7 @@ bool PublishKafka::configureNewConnection(const std::shared_ptr<KafkaConnection>
     result = rd_kafka_conf_set(conf_, "sasl.kerberos.principal", value.c_str(), errstr.data(), errstr.size());
     logger_->log_debug("PublishKafka: sasl.kerberos.principal [%s]", value);
     if (result != RD_KAFKA_CONF_OK) {
-      logger_->log_error("PublishKafka: configure error result [%s]", errstr);
+      logger_->log_error("PublishKafka: configure error result [%s]", errstr.data());
       return false;
     }
   }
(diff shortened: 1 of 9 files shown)


bakaid:
Only the most basic positive and negative tests have been run on this, further extensive testing and some investigations are required, hence this being a draft.
@arpadboda I am submitting this so that review can start.

extensions/librdkafka/PublishKafka.cpp
```diff
@@ -43,14 +44,18 @@ core::Property PublishKafka::Topic(core::PropertyBuilder::createProperty("Topic
...
 core::Property PublishKafka::MaxMessageSize(core::PropertyBuilder::createProperty("Max Request Size")->withDescription("Maximum Kafka protocol request message size")->isRequired(false)->build());
 
 core::Property PublishKafka::RequestTimeOut(
-    core::PropertyBuilder::createProperty("Request Timeout")->withDescription("The ack timeout of the producer request in milliseconds")->isRequired(false)->withDefaultValue<core::TimePeriodValue>(
+    core::PropertyBuilder::createProperty("Request Timeout")->withDescription("The ack timeout of the producer request")->isRequired(false)->withDefaultValue<core::TimePeriodValue>(
         "10 sec")->supportsExpressionLanguage(true)->build());
 
+core::Property PublishKafka::MessageTimeOut(
+    core::PropertyBuilder::createProperty("Message Timeout")->withDescription("The total time sending a message could take")->isRequired(false)->withDefaultValue<core::TimePeriodValue>(
+        "30 sec")->supportsExpressionLanguage(true)->build());
```
arpadboda: Please brake this line at "->" operators

extensions/librdkafka/PublishKafka.cpp
```diff
@@ -289,122 +356,309 @@ bool PublishKafka::configureNewConnection(const std::shared_ptr<KafkaConnection>
...
 
-  if (!flowFile) {
+  // Try to get a KafkaConnection
+  std::string client_id, brokers;
+  if (!context->getProperty(ClientName.getName(), client_id)) {
+    logger_->log_error("Client Name property missing or invalid");
+    context->yield();
+    return;
+  }
+  if (!context->getProperty(SeedBrokers.getName(), brokers)) {
```
arpadboda: These should be moved to onschedule in scope of a follow-up ticket.

extensions/librdkafka/PublishKafka.cpp
```diff
@@ -289,122 +356,309 @@ bool PublishKafka::configureNewConnection(const std::shared_ptr<KafkaConnection>
...
+  } else {
+    target_batch_payload_size = 512 * 1024U;
+  }
 
-      conn->putTopic(topic, kafkaTopicref);
-    }
+  // Max Flow Segment Size
+  uint64_t max_flow_seg_size;
+  value = "";
+  if (context->getProperty(MaxFlowSegSize.getName(), value) && !value.empty() && core::Property::StringToInt(value, max_flow_seg_size)) {
```
arpadboda: There is a function to do the check and conversion for you:
```
  template<typename T>
  bool getProperty(const std::string &name, T &value) const {
    return getPropertyImp<typename std::common_type<T>::type>(name, value);
  }
```

It's also fine in scope of a follow-up.

extensions/librdkafka/PublishKafka.cpp
```diff
@@ -289,122 +356,309 @@ bool PublishKafka::configureNewConnection(const std::shared_ptr<KafkaConnection>
...
+      break;
+    }
+    actual_bytes += flowFile->getSize();
+    flowFiles.emplace_back(std::move(flowFile));
+    if (target_batch_payload_size != 0U && actual_bytes >= target_batch_payload_size) {
+      break;
+    }
+  }
+  if (flowFiles.empty()) {
     return;
```
arpadboda: This seems to be a yield scenario.

extensions/librdkafka/PublishKafka.cpp
```diff
@@ -289,122 +356,309 @@ bool PublishKafka::configureNewConnection(const std::shared_ptr<KafkaConnection>
...
+  uint64_t actual_bytes = 0U;
+  std::vector<std::shared_ptr<core::FlowFile>> flowFiles;
+  for (uint32_t i = 0; i < batch_size; i++) {
+    std::shared_ptr<core::FlowFile> flowFile = session->get();
+    if (flowFile == nullptr) {
+      break;
+    }
+    actual_bytes += flowFile->getSize();
+    flowFiles.emplace_back(std::move(flowFile));
+    if (target_batch_payload_size != 0U && actual_bytes >= target_batch_payload_size) {
```
arpadboda: As far as I understand we can exceed the batch size here. 
I don't think there is a session API call to provide the size of the next flowfile without removing that from session. 
I think we should address it in a follow-up as it's ugly to exceed the configured limit, although it's not that urgent.

extensions/librdkafka/PublishKafka.cpp
bakaid: I agree, but I would refer the honourable member to the previous line and ask him to direct his complaints to the original author. (Will fix.)

extensions/librdkafka/PublishKafka.cpp
bakaid: Agreed.

extensions/librdkafka/PublishKafka.cpp
bakaid: Agreed, but that's why I sneakily called the property *Target* Batch Payload Size and not *Max* Batch Payload Size.

extensions/librdkafka/PublishKafka.cpp
bakaid: Agreed.

extensions/librdkafka/PublishKafka.h
```diff
@@ -100,89 +107,207 @@ class PublishKafka : public core::Processor {
...
+
+    void interrupt() {
+      std::unique_lock<std::mutex> lock(mutex);
+      interrupted = true;
+      cv.notify_all();
+    }
+
+    bool wasInterrupted() {
+      std::lock_guard<std::mutex> lock(mutex);
+      return interrupted;
```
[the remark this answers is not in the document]
bakaid: I most probably could, but I need interrupted to be under a mutex for the condition variable in interrupt() anyway, and to be honest, I do not know the exact requirements between condition variables and atomics for correctly publishing modifications to the notified thread from the top of my head, so I would prefer a defensive approach here. We have enough mutex locks anyway for this single call per onTrigger not to matter.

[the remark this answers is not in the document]
arpadboda: Okay, I can live with that, don't think the cost of the mutex would matter here anyway.

extensions/librdkafka/PublishKafka.cpp
bakaid: I agree it would be better, but in this Processor this pattern is used, and I don't want to change everything now. Will create a follow-up.

extensions/librdkafka/PublishKafka.h
```diff
@@ -100,89 +107,207 @@ class PublishKafka : public core::Processor {
   static core::Relationship Failure;
   static core::Relationship Success;
 
+  // Message
+   struct MessageResult {
+    bool completed;
```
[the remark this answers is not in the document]
bakaid: Done.

PR #655, opened 2019-09-30
MINIFICPP-1049 - Fix make docker

bakaid:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/docker/DockerBuild.sh
+++ b/docker/DockerBuild.sh
@@ -31,12 +31,13 @@ echo "CMake Source Directory: $CMAKE_SOURCE_DIR"
 echo "MiNiFi Package: $MINIFI_SOURCE_CODE"
 
 # Copy the MiNiFi source tree to the Docker working directory before building
+rm -rf $CMAKE_SOURCE_DIR/docker/minificppsource
 mkdir -p $CMAKE_SOURCE_DIR/docker/minificppsource
 rsync -avr \
       --exclude '/*build*' \
       --exclude '/*_repository*' \
       --exclude '/logs' \
-      --exclude '/cmake' \
+      --exclude '/cmake-build-*' \
       --exclude '/docker' \
       --exclude '.git' \
       --exclude '/extensions/expression-language/Parser.cpp' \
@@ -57,3 +58,5 @@ DOCKER_COMMAND="docker build --build-arg UID=$UID_ARG \
                              apacheminificpp:$MINIFI_VERSION ."
 echo "Docker Command: '$DOCKER_COMMAND'"
 ${DOCKER_COMMAND}
+
+rm -rf $CMAKE_SOURCE_DIR/docker/minificppsource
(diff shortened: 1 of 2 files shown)


docker/DockerBuild.sh
```diff
@@ -31,12 +31,13 @@ echo "CMake Source Directory: $CMAKE_SOURCE_DIR"
 echo "MiNiFi Package: $MINIFI_SOURCE_CODE"
 
 # Copy the MiNiFi source tree to the Docker working directory before building
+rm -rf $CMAKE_SOURCE_DIR/docker/minificppsource
```
bakaid: Excluding a path will not delete it in the target, even though we have the --delete flag said (this is the proper behaviour I think). However, this masked the error, because if you have already had the cmake directory copied, it left it as it was, just didn't update it, and the build succeeded. I had to clean up my clone to reproduce the issue.

docker/DockerBuild.sh
```diff
@@ -31,12 +31,13 @@ echo "CMake Source Directory: $CMAKE_SOURCE_DIR"
...
 mkdir -p $CMAKE_SOURCE_DIR/docker/minificppsource
 rsync -avr \
       --exclude '/*build*' \
       --exclude '/*_repository*' \
       --exclude '/logs' \
-      --exclude '/cmake' \
+      --exclude '/cmake-build-*' \
```
bakaid: @arpadboda  This was added in https://github.com/apache/nifi-minifi-cpp/pull/634 - I am not sure what was the original intent here, but this excludes the whole cmake directory, which messes up the whole build process. I changed it to this, based on the .gitignore, but I want to make sure there was no other consideration behind it that I would break with this.

docker/Dockerfile
```diff
@@ -76,7 +76,7 @@ ENV MINIFI_HOME $MINIFI_BASE_DIR/nifi-minifi-cpp-${MINIFI_VERSION}
 RUN cd ${MINIFI_BASE_DIR} \
 	&& mkdir build \
 	&& cd build \
-	&& cmake -DOPENSSL_FORCE_SHARED=true -DDISABLE_JEMALLOC=ON -DSTATIC_BUILD= -DSKIP_TESTS=true -DENABLE_JNI=ON .. \
+	&& cmake -DDISABLE_JEMALLOC=ON -DSTATIC_BUILD= -DSKIP_TESTS=true -DENABLE_JNI=ON .. \
```
bakaid: `OPENSSL_FORCE_SHARED` no longer makes sense (doesn't hurt either, but it was confusing).

docker/DockerBuild.sh
arpadboda: I wanted to avoid uploading CMakeCache and cmake-build-debug to the container, thanks for fixing.

docker/DockerBuild.sh
```diff
@@ -57,3 +58,5 @@ DOCKER_COMMAND="docker build --build-arg UID=$UID_ARG \
                              apacheminificpp:$MINIFI_VERSION ."
 echo "Docker Command: '$DOCKER_COMMAND'"
 ${DOCKER_COMMAND}
+
+rm -rf $CMAKE_SOURCE_DIR/docker/minificppsource
```
arpadboda: What's the intent of doing it twice? 
I'm fine with either doing cleanup at the start of the build or at end end, but don't get the point of doing both.

arpadboda:
Executed properly in a new folder on my U18, merging.

PR #671, opened 2019-10-28
MINIFICPP-1067 Add recommendations for antivirus exclusions to README

andrewmlim:


andrewmlim:
@phrocker Thanks for reviewing! Could you provide an example of what you were thinking? I guess I thought it was obvious to users that this was a configurable option, but an example might clear up my misunderstanding.

arpadboda:
@andrewmlim : as far as I understand @phrocker would like to highlight the fact that the repo paths are configurable in _minifi.properties_ file, so the documentation should say something like "exclude the repository folders, which are configurable, default are the following: .... "

phrocker:
@andrewmlim -- @arpadboda is correct; however, I approved it. I can't merge it until I get my computer back from repair today and have my signing keys. At that point I will merge this  -- my comment was simply a comment not one that I felt should block this PR.

andrewmlim:
@arpadboda @phrocker Thanks to you both for the clarification and update!

PR #672, opened 2019-10-29
MINIFICPP-1043 - FetchOPC should support reporting changes only

nghiaxlee:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/opc/include/fetchopc.h
+++ b/extensions/opc/include/fetchopc.h
@@ -22,7 +22,7 @@
 #include <memory>
 #include <string>
 #include <list>
-#include <map>
+#include <unordered_map>
 #include <mutex>
 #include <thread>
 
@@ -97,7 +97,7 @@ protected:
 private:
   std::mutex onTriggerMutex_;
   std::vector<UA_NodeId> translatedNodeIDs_;  // Only used when user provides path, path->nodeid translation is only done once
-  std::map<std::string, std::string> node_timestamp_; // Key = Full path, Value = Timestamp
+  std::unordered_map<std::string, std::string> node_timestamp_; // Key = Full path, Value = Timestamp
 
 };
 
--- a/extensions/opc/include/fetchopc.h
+++ b/extensions/opc/include/fetchopc.h
@@ -53,6 +53,7 @@ public:
   static core::Property NodeID;
   static core::Property NameSpaceIndex;
   static core::Property MaxDepth;
+  static core::Property Lazy;
 
   // Supported Relationships
   static core::Relationship Success;
@@ -91,10 +92,12 @@ protected:
   uint32_t nodesFound_;
   uint32_t variablesFound_;
   uint64_t maxDepth_;
+  bool lazy_mode_;
 
 private:
   std::mutex onTriggerMutex_;
   std::vector<UA_NodeId> translatedNodeIDs_;  // Only used when user provides path, path->nodeid translation is only done once
+  std::map<std::string, std::string> node_timestamp_; // Key = Full path, Value = Timestamp
 
 };
 
--- a/extensions/opc/src/fetchopc.cpp
+++ b/extensions/opc/src/fetchopc.cpp
@@ -65,13 +65,20 @@ namespace processors {
       ->withDescription("Specifiec the max depth of browsing. 0 means unlimited.")
       ->withDefaultValue<uint64_t>(0)->build());
 
+  core::Property FetchOPCProcessor::Lazy(
+      core::PropertyBuilder::createProperty("Lazy mode")
+      ->withDescription("Specifiec the max depth of browsing. 0 means unlimited.")
+      ->isRequired(true)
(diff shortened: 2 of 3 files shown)


extensions/opc/src/fetchopc.cpp
```diff
@@ -65,13 +65,20 @@ namespace processors {
       ->withDescription("Specifiec the max depth of browsing. 0 means unlimited.")
       ->withDefaultValue<uint64_t>(0)->build());
 
+  core::Property FetchOPCProcessor::Lazy(
```
bakaid: I think it would be better to have a default (default "Off") here, this way we would not break previous configurations and make it easier to configure new ones.

extensions/opc/src/fetchopc.cpp
```diff
@@ -194,11 +204,25 @@ namespace processors {
...
-        variablesFound_++;
+        bool write = true;
+        if (lazy_mode_) {
+          write = false;
+          std::string nodeid = nodedata.attributes["Full path"];
+          std::string cur_timestamp = node_timestamp_[nodeid];
+          std::string new_timestamp = nodedata.attributes["Sourcetimestamp"];
+          if (cur_timestamp != new_timestamp) {
+            node_timestamp_[nodeid] = new_timestamp;
+            logger_->log_warn("Node %s has new source timestamp %s", nodeid, new_timestamp);
```
bakaid: I think a warning level log is excessive here, as this is a normal code path. I would leave it at max on info, but if it spams too much even debug.

extensions/opc/include/fetchopc.h
```diff
@@ -91,10 +92,12 @@ class FetchOPCProcessor : public BaseOPCProcessor {
   uint32_t nodesFound_;
   uint32_t variablesFound_;
   uint64_t maxDepth_;
+  bool lazy_mode_;
 
 private:
   std::mutex onTriggerMutex_;
   std::vector<UA_NodeId> translatedNodeIDs_;  // Only used when user provides path, path->nodeid translation is only done once
+  std::map<std::string, std::string> node_timestamp_; // Key = Full path, Value = Timestamp
```
bakaid: I think this would be better as an unordered_map - we don't need ordering and it is potentially very large.

extensions/opc/src/opc.cpp
```diff
@@ -548,9 +569,9 @@ std::string OPCDateTime2String(UA_DateTime raw_date) {
   UA_DateTimeStruct dts = UA_DateTime_toStruct(raw_date);
   std::array<char, 100> charBuf;
 
-  snprintf(charBuf.data(), charBuf.size(), "%02hu-%02hu-%04hu %02hu:%02hu:%02hu.%03hu", dts.day, dts.month, dts.year, dts.hour, dts.min, dts.sec, dts.milliSec);
+  int sz = snprintf(charBuf.data(), charBuf.size(), "%02hu-%02hu-%04hu %02hu:%02hu:%02hu.%03hu", dts.day, dts.month, dts.year, dts.hour, dts.min, dts.sec, dts.milliSec);
 
-  return std::string(charBuf.data(), charBuf.size());
+  return std::string(charBuf.data(), sz);
```
bakaid: Nice catch!

extensions/opc/src/opc.cpp
```diff
@@ -220,10 +220,31 @@ NodeData Client::getNodeData(const UA_ReferenceDescription *ref, const std::stri
       nodedata.attributes["NodeID type"] = "numeric";
     }
     nodedata.attributes["Browsename"] = browsename;
+    // TODO: Base path is always ""
     nodedata.attributes["Full path"] = basePath + "/" + browsename;
     nodedata.dataTypeID = UA_TYPES_COUNT;
     UA_Variant* var = UA_Variant_new();
     if(UA_Client_readValueAttribute(client_, ref->nodeId.nodeId, var) == UA_STATUSCODE_GOOD && var->type != NULL && var->data != NULL) {
+      // Because the timestamps are eliminated in readValueAttribute for simplification
```
bakaid: If possible, we should only do this if we are in lazy mode, otherwise this is unnecessary.

extensions/opc/src/fetchopc.cpp
arpadboda: Agreed, I vote for debug here.

extensions/opc/src/opc.cpp
arpadboda: Non-layz does the same thing in the background, the 3rd party provides a simplified API function to hide this.
I don't mind having a bit more complex but single code path in case it has no performance drawbacks.

nghiaxlee: Yes, this code part is a trimmed down version of `UA_readValueAttribute` with a slight modification. Despite checking lazy mode before running might be better, I then have to pass the lazy flag from fetchopc processor to this function, which I am not quite into.

PR #678, opened 2019-11-11
MINIFICPP-1080 - Unicode conversion fixes and speedup

bakaid:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/windows-event-log/wel/UnicodeConversion.h
+++ b/extensions/windows-event-log/wel/UnicodeConversion.h
@@ -1,0 +1,41 @@
+/**
+ * @file UnicodeConversion.h
+ * Unicode conversion functions
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


extensions/windows-event-log/wel/UnicodeConversion.h
```diff
@@ -0,0 +1,41 @@
...
+#include <atlconv.h>
+
+namespace org {
+  namespace apache {
+    namespace nifi {
+      namespace minifi {
+        namespace wel {
+          inline std::string to_string(const wchar_t* pChar) {
+            ATL::CW2A aString(pChar, CP_UTF8);
+            return std::string(aString);
```
arpadboda: The function looks good, although not sure I understand the purpose of this file. 

In case it's only going to be used in WEL, it can be a func in an anonymus namespace as well. 
Otherwise (in case we plan to use it elsewhere) I should be moved to stringutils.

bakaid: Eventually I want this to be part of StringUtils, but while I have manually tested this with this particular processor, and it seems to work OK, I am not sure that
 - using ATL will not incur any issues
 - this works with various types Unicode strings (and malformatted UTF-16, incomplete surrogate pairs, etc.)
Therefore I would be more comfortable keeping it here until it proves it works fine and only include it in our API after that - and then including util tests. I will create a follow-up issue for this.
Until then this is a helper for this extension, which are included in the wel namespace. Once we transition to StringUtils, this file will be easily removable.

arpadboda: Okay, sounds reasonable, thanks!

PR #681, opened 2019-11-14
MINIFICPP-1085

am-c-p-p:

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -258,92 +389,65 @@ bool ConsumeWindowsEventLog::subscribe(const std::shared_ptr<core::ProcessContex
...
 
+  if (!pBookmark_) {
+    logger_->log_error("!pBookmark_");
+    return false;
+  }
+
+  auto channel = std::wstring(channel_.begin(), channel_.end());
+  auto query = std::wstring(query_.begin(), query_.end());
+
+  [&channel, &query, this]() {
```
bakaid: What is the purpose of this lambda if it is unnamed and called right away?

extensions/windows-event-log/Bookmark.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+  auto ret = file_.is_open();
+  if (!ret) {
+    logger_->log_error("Cannot open %s", filePath_.c_str());
+  }
+
+  return ret;
+}
+
+// Creates directory "processor_repository\ConsumeWindowsEventLog\uuid\{uuid}" under "bin" directory.
```
bakaid: This is bad practice, we never place or access anything relative to the executable's path, only relative to MINIFI_HOME or the values configured in properties or the config.yml.
Please take a look at how TailFile's state storage is implemented (it has a Property to set it).
By the way (for testing reasons) getting the executable's path and dir are already implemented in `FileUtils::get_executable_path` and `FileUtils::get_executable_dir`.

extensions/windows-event-log/Bookmark.cpp
```diff
@@ -0,0 +1,217 @@
...
+        return false;
+      }
+    }
+  }
+
+  return true;
+}
+
+std::string Bookmark::filePath(const std::string& uuid) {
+  static auto s_filePath = [&uuid, this]() {
```
bakaid: This function is unused (and very weird, I could not decipher the original intent).

extensions/windows-event-log/Bookmark.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+bool Bookmark::getBookmarkXmlFromFile() {
+  bookmarkXml_.clear();
+
+  std::wifstream file(filePath_);
+  if (!file.is_open()) {
+    return createEmptyBookmarkXmlFile();
+  }
+
+  wchar_t c;
```
bakaid: This looks extremely ineffective, with single characters read and appended to the string.

extensions/windows-event-log/Bookmark.cpp
```diff
@@ -0,0 +1,217 @@
...
+    if (!file) {
+      break;
+    }
+
+    bookmarkXml_ += c;
+  } while (true);
+
+  file.close();
+
+  file_.open(filePath_);
```
bakaid: What's the purpose of always keeping open a file handle? Is this some kind of locking? We use UUIDs in the path of these files, who are we afraid would access them concurrently?

extensions/windows-event-log/Bookmark.cpp
```diff
@@ -0,0 +1,217 @@
...
+      std::vector<wchar_t> buf(bufferSize / 2 + 1);
+
+      if (!EvtRender(0, hBookmark_, EvtRenderBookmark, bufferSize, &buf[0], &bufferUsed, &propertyCount)) {
+        logger_->log_error("!EvtRender error: %d.", GetLastError());
+        return false;
+      }
+
+      file_.seekp(std::ios::beg);
+
+      // Write new bookmark over old and in the end write '!'. Then new bookmark is read until '!'.
```
bakaid: Why do we need this? Why can't we just truncate the file and write the new bookmark?

extensions/windows-event-log/Bookmark.cpp
am-c-p-p: Yes, it is unused, left after refactoring, will remove.

extensions/windows-event-log/Bookmark.cpp
am-c-p-p: Yes, is extremely ineffective for large data. 
But bookmark is small XML (~100 bytes) contains just channel name and RecordId number. 
After first read all data is in file io buffer, and bookmarkXml_ memory after several `+` will have enough memory (string doubles memory if it is not enough) to hold XML.

extensions/windows-event-log/Bookmark.cpp
am-c-p-p: I'll fix it.

extensions/windows-event-log/Bookmark.cpp
am-c-p-p: I'll fix it to put under MINIFI_HOME.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
am-c-p-p: I'll refactor.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -135,14 +152,22 @@ ConsumeWindowsEventLog::ConsumeWindowsEventLog(const std::string& name, utils::I
   } else {
     LogWindowsError();
   }
+
+  writeXML_ = false;
+  writePlainText_ = false;
+
+  pBookmark_ = std::make_unique<Bookmark>(getUUIDStr(), logger_);
+  if (!*pBookmark_) {
+    pBookmark_.release();
```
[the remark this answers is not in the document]
am-c-p-p: I'll fix.

extensions/windows-event-log/Bookmark.cpp
am-c-p-p: The file is opened because bookmark is written to the file for every event and it could be many of events per second.

bakaid: We should really review and merge the rocksdb state storage PR for these exact cases.
Until then I suggest two things: make bookmarking configurable via a Property to make sure it can be easily switched off if it slows things down too much.
If measurements show that this really slows things down, you could also make it that it only saves the bookmark every N event (or on shutdown).

extensions/windows-event-log/Bookmark.cpp
```diff
@@ -0,0 +1,217 @@
...
+namespace processors {
+
+Bookmark::Bookmark(const std::string& uuid, std::shared_ptr<logging::Logger> logger)
+  :logger_(logger) {
+  if (createUUIDDir(uuid, filePath_)) {
+    filePath_ += "Bookmark.txt";
+  }
+
+  if (!filePath_.empty()) {
+    if (!getBookmarkXmlFromFile()) {
```
arpadboda: These two conditions an be merged.

extensions/windows-event-log/Bookmark.cpp
```diff
@@ -0,0 +1,217 @@
...
+    logger_->log_error("Cannot open %s", filePath_.c_str());
+  }
+
+  return ret;
+}
+
+// Creates directory "processor_repository\ConsumeWindowsEventLog\uuid\{uuid}" under "bin" directory.
+bool Bookmark::createUUIDDir(const std::string& uuid, std::string& dir)
+{
+  auto getRootDirectory = [](std::string& rootDir, std::shared_ptr<logging::Logger> logger)
```
arpadboda: Do we use it anywhere else? 
What's the point in creating the lambda?

extensions/windows-event-log/Bookmark.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+  dir.clear();
+
+  if (!getRootDirectory(dir, logger_))
+    return false;
+
+  for (const auto& curDir : std::vector<std::string>{ "processor_repository", "ConsumeWindowsEventLog", "uuid", uuid }) {
+    dir += curDir + '\\';
+
+    if (GetFileAttributes(dir.c_str()) == INVALID_FILE_ATTRIBUTES) {
```
arpadboda: in FileUtils, there is a function called "create_dir".
It was just fixed for windows and returns true in case the dir could be created or it already exists. 
I think that's exactly what you need here.

extensions/windows-event-log/Bookmark.h
```diff
@@ -0,0 +1,54 @@
...
+namespace minifi {
+namespace processors {
+
+class Bookmark
+{
+public:
+  Bookmark(const std::string& uuid, std::shared_ptr<logging::Logger> logger);
+  ~Bookmark();
+
+  operator bool();
```
arpadboda: const pls for this and all the getters.

extensions/windows-event-log/Bookmark.cpp
am-c-p-p: It is similar to breaking a large function in small functions even though they are not used anywhere other than in the large function.

am-c-p-p:
I resolved all comments in Fixed in https://github.com/bakaid/nifi-minifi-cpp/pull/1.

extensions/windows-event-log/Bookmark.cpp
am-c-p-p: Fixed.

extensions/windows-event-log/Bookmark.h
am-c-p-p: Fixed.

extensions/windows-event-log/Bookmark.cpp
am-c-p-p: Fixed.

extensions/windows-event-log/Bookmark.cpp
am-c-p-p: I tested writing to file wcahr_t buf[200]; file.seekp(std::ios::beg); file << &buf[0] << L'!'; file.flush();

bakaid:
Merged in #685

PR #670, opened 2019-10-28
MINIFICPP-1074 - TailFileTests was not working on WIN32 due do CLRF

asdaraujo:
Several points in the test cases implementation were not handling
the WIN32 CRLF enf-of-line correctly.

Thank you for submitting a contribution to Apache NiFi - MiNiFi C++.

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [x] Is there a JIRA ticket associated with this PR? Is it referenced
     in the commit message?

- [x] Does your PR title start with MINIFICPP-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.

- [x] Has your PR been rebased against the latest commit within the target branch (typically master)?

- [x] Is your initial contribution a single, squashed commit?

### For code changes:
- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?
- [ ] If applicable, have you updated the LICENSE file?
- [ ] If applicable, have you updated the NOTICE file?

### For documentation related changes:
- [ ] Have you ensured that format looks appropriate for the output in which it is rendered?

### Note:
Please ensure that once the PR is submitted, you check travis-ci for build issues and submit an update to your PR as soon as possible.

PR #683, opened 2019-11-15
MINIFICPP-1087 - Proper handling of errors during onSchedule calls

arpadboda:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/CPPLINT.cfg
+++ b/CPPLINT.cfg
@@ -1,3 +1,3 @@
 set noparent
-filter=-runtime/reference
+filter=-runtime/reference,-runtime/string
 
--- a/libminifi/src/core/ProcessGroup.cpp
+++ b/libminifi/src/core/ProcessGroup.cpp
@@ -149,7 +149,7 @@ void ProcessGroup::removeProcessGroup(ProcessGroup *child) {
 
 void ProcessGroup::startProcessingProcessors(const std::shared_ptr<TimerDrivenSchedulingAgent> timeScheduler,
     const std::shared_ptr<EventDrivenSchedulingAgent> &eventScheduler, const std::shared_ptr<CronDrivenSchedulingAgent> &cronScheduler) {
-  std::lock_guard<std::recursive_mutex> lock(mutex_);
+  std::unique_lock<std::recursive_mutex> lock(mutex_);
 
   std::set<std::shared_ptr<Processor> > failed_processors;
 
@@ -189,17 +189,16 @@ void ProcessGroup::startProcessingProcessors(const std::shared_ptr<TimerDrivenSc
 
   if (!onScheduleTimer_ && !failed_processors_.empty() && onschedule_retry_msec_ > 0) {
     logger_->log_info("Retrying failed processors in %lld msec", onschedule_retry_msec_.load());
-    onScheduleTimer_.reset(new utils::CallBackTimer(std::chrono::milliseconds(onschedule_retry_msec_)));
     auto func = [this, eventScheduler, cronScheduler, timeScheduler]() {
       this->startProcessingProcessors(timeScheduler, eventScheduler, cronScheduler);
     };
-    onScheduleTimer_->start(func);
-  } else if (failed_processors_.empty() && onScheduleTimer_) {
+    onScheduleTimer_.reset(new utils::CallBackTimer(std::chrono::milliseconds(onschedule_retry_msec_), func));
+    onScheduleTimer_->start();
+  } else if (failed_processors_.empty() && onScheduleTimer_ && onScheduleTimer_->is_running()) {
+    lock.unlock();
     onScheduleTimer_->stop();
-    onScheduleTimer_.reset();
   }
 }
-
 
 void ProcessGroup::startProcessing(const std::shared_ptr<TimerDrivenSchedulingAgent> timeScheduler, const std::shared_ptr<EventDrivenSchedulingAgent> &eventScheduler,
                                    const std::shared_ptr<CronDrivenSchedulingAgent> &cronScheduler) {
@@ -226,8 +225,7 @@ void ProcessGroup::startProcessing(const std::shared_ptr<TimerDrivenSchedulingAg
 
 void ProcessGroup::stopProcessing(const std::shared_ptr<TimerDrivenSchedulingAgent> timeScheduler, const std::shared_ptr<EventDrivenSchedulingAgent> &eventScheduler,
                                   const std::shared_ptr<CronDrivenSchedulingAgent> &cronScheduler) {
-
-  if (onScheduleTimer_) {
+  if (onScheduleTimer_ && onScheduleTimer_->is_running()) {
     onScheduleTimer_->stop();
   }
 
--- a/libminifi/src/core/ProcessGroup.cpp
+++ b/libminifi/src/core/ProcessGroup.cpp
@@ -189,7 +189,7 @@ void ProcessGroup::startProcessingProcessors(const std::shared_ptr<TimerDrivenSc
 
   if (!onScheduleTimer_ && !failed_processors_.empty() && onschedule_retry_msec_ > 0) {
     logger_->log_info("Retrying failed processors in %lld msec", onschedule_retry_msec_.load());
-    onScheduleTimer_.reset(new utils::CallBackTimer(onschedule_retry_msec_));
+    onScheduleTimer_.reset(new utils::CallBackTimer(std::chrono::milliseconds(onschedule_retry_msec_)));
     auto func = [this, eventScheduler, cronScheduler, timeScheduler]() {
       this->startProcessingProcessors(timeScheduler, eventScheduler, cronScheduler);
     };
--- a/libminifi/src/core/ProcessGroup.cpp
+++ b/libminifi/src/core/ProcessGroup.cpp
@@ -59,6 +59,12 @@ ProcessGroup::ProcessGroup(ProcessGroupType type, std::string name, utils::Ident
   }
 
   yield_period_msec_ = 0;
+
+  if (parent_process_group_ != 0) {
+    onschedule_retry_msec_ = parent_process_group_->getOnScheduleRetryPeriod();
+  } else {
+    onschedule_retry_msec_ = ONSCHEDULE_RETRY_INTERVAL;
+  }
   transmitting_ = false;
   transport_protocol_ = "RAW";
 
@@ -74,6 +80,7 @@ ProcessGroup::ProcessGroup(ProcessGroupType type, std::string name)
   id_generator_->generate(uuid_);
 
   yield_period_msec_ = 0;
+  onschedule_retry_msec_ = ONSCHEDULE_RETRY_INTERVAL;
   transmitting_ = false;
   transport_protocol_ = "RAW";
 
@@ -81,6 +88,10 @@ ProcessGroup::ProcessGroup(ProcessGroupType type, std::string name)
 }
 
 ProcessGroup::~ProcessGroup() {
+  if (onScheduleTimer_) {
+    onScheduleTimer_->stop();
+  }
+
   for (auto &&connection : connections_) {
     connection->drain();
   }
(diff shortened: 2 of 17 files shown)


libminifi/src/core/ProcessGroup.cpp
```diff
@@ -156,6 +168,49 @@ void ProcessGroup::startProcessing(const std::shared_ptr<TimerDrivenSchedulingAg
...
+    catch (...) {
+      logger_->log_error("Failed to start processor %s (%s)", processor->getUUIDStr(), processor->getName());
+      failed_processors.insert(processor);
+    }
+  }
+  failed_processors_ = std::move(failed_processors);
+
+  for (auto& processor : failed_processors_) {
+    try {
+      processor->onUnSchedule();
```
phrocker: why not call notifystop here?

libminifi/src/core/yaml/YamlConfiguration.cpp
```diff
@@ -61,6 +61,21 @@ core::ProcessGroup *YamlConfiguration::parseRootProcessGroupYaml(YAML::Node root
 
   this->name_ = flowName;
 
+  if (rootFlowNode["onschedule retry interval"]) {
```
phrocker: I guess the logic is that this is something that is optional? This makes a lot of sense if requested via the flow configuration but it does not always make sense in the guise of distributed containers/systems.

libminifi/src/core/ProcessGroup.cpp
```diff
@@ -136,13 +147,14 @@ void ProcessGroup::removeProcessGroup(ProcessGroup *child) {
   }
 }
 
-void ProcessGroup::startProcessing(const std::shared_ptr<TimerDrivenSchedulingAgent> timeScheduler, const std::shared_ptr<EventDrivenSchedulingAgent> &eventScheduler,
-                                   const std::shared_ptr<CronDrivenSchedulingAgent> &cronScheduler) {
+void ProcessGroup::startProcessingProcessors(const std::shared_ptr<TimerDrivenSchedulingAgent> timeScheduler,
```
phrocker: line lengths are 200 fyi.

CPPLINT.cfg
```diff
@@ -1,3 +1,3 @@
 set noparent
-filter=-runtime/reference
+filter=-runtime/reference,-runtime/string
```
phrocker: a quick glance did not make the necessity of this clear. I assume this is for your test?

libminifi/src/core/ProcessGroup.cpp
arpadboda: Default implementation of onUnSchedule calls notifyStop, but the you are free to override that in your processor.
I think this matches the way it's done in NiFi.

libminifi/src/core/yaml/YamlConfiguration.cpp
arpadboda: Something configurable. You are free to set 0, which is going to turn it off. 
Although to match NiFi behaviour retry should be the default way to do things. In case a processor can't connect (for eg. the destination is temporary down), we shouldn't give up entirely.

libminifi/src/core/ProcessGroup.cpp
arpadboda: Could you be a bit more detailed?
Linter is ok with this.

CPPLINT.cfg
arpadboda: Yes, it's for the
```
static const std::string
```
members in KamikazeProcessor, but it's not the only place we would like to have it, see further references:
 https://github.com/apache/nifi-minifi-cpp/pull/670/files#diff-dfeecc634b3aea462f1f8dc52c2abaceR50

This rule forces people to use raw pointers instead of string, which is simply error-prone, but doesn't add value.

libminifi/src/core/ProcessGroup.cpp
```diff
@@ -136,13 +147,14 @@ void ProcessGroup::removeProcessGroup(ProcessGroup *child) {
...
+    const std::shared_ptr<EventDrivenSchedulingAgent> &eventScheduler, const std::shared_ptr<CronDrivenSchedulingAgent> &cronScheduler) {
   std::lock_guard<std::recursive_mutex> lock(mutex_);
 
-  try {
-    // Start all the processor node, input and output ports
-    for (const auto &processor : processors_) {
+  std::set<std::shared_ptr<Processor> > failed_processors;
+
+  for (const auto &processor : failed_processors_) {
+    try {
```
bakaid: If I understand correctly, this means that processors that can be successfully scheduled will remain scheduled even if other processors fails to schedule.
This will cause processors before the failed ones to fill the connections until backpressure stops them, and processors after the failed ones to starve.
I am not saying this is a bad thing, I am asking if this is the behaviour we want/NiFi follows.

libminifi/src/core/ProcessGroup.cpp
bakaid: I agree that by default this should be notifyStop, but I like that it is overrideable by the implementation, because it leaves space for smarter failure-handling logics, so I am OK with this as it is.

libminifi/src/core/yaml/YamlConfiguration.cpp
```diff
@@ -61,6 +61,21 @@ core::ProcessGroup *YamlConfiguration::parseRootProcessGroupYaml(YAML::Node root
...
+    int64_t onScheduleRetryPeriodValue = -1;
+    std::string onScheduleRetryPeriod = rootFlowNode["onschedule retry interval"].as<std::string>();
+    logger_->log_debug("parseRootProcessGroup: onschedule retry period => [%s]", onScheduleRetryPeriod);
+
+    core::TimeUnit unit;
+
+    if (core::Property::StringToTime(onScheduleRetryPeriod, onScheduleRetryPeriodValue, unit)
+        && core::Property::ConvertTimeUnitToMS(onScheduleRetryPeriodValue, unit, onScheduleRetryPeriodValue)
+        && group) {
+      logger_->log_debug("parseRootProcessGroup: onschedule retry => [%ll] ms", onScheduleRetryPeriodValue);
```
bakaid: `%ll` is a length field without a type field, so it won't consume the int64_t and print `parseRootProcessGroup: onschedule retry => [%] ms`.
It should be `%lld`.

libminifi/src/utils/CallBackTimer.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace utils {
+
+CallBackTimer::CallBackTimer(int interval) : execute_(false), interval_(interval) {
+  if (interval_ <= 0) {
+    throw std::invalid_argument("Interval cannot be negative!");
```
bakaid: Why is the constructor parameter a signed int then? Making it an unsigned one would mean that the constructor could not be used erroneously and we wouldn't need to throw exceptions.

libminifi/src/utils/CallBackTimer.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+void CallBackTimer::stop() {
+  execute_.store(false, std::memory_order_release);
+  if (thd_.joinable()) {
+    thd_.join();
+  }
+}
+
+void CallBackTimer::start(std::function<void(void)> func) {
+  if (execute_.load(std::memory_order_acquire)) {
```
bakaid: I think without an atomic exchange the thread creation and assignment to `thd_` is not protected properly if two `start`s are called from two threads.
I think using a std::mutex instead of atomics would be fine here, as I don't think there would be too much contention in our use case.

arpadboda: Agreed, added a mutex.

libminifi/src/utils/CallBackTimer.cpp
arpadboda: Fair point. Simply changed it take chrono::milliseconds - I think that should be talkative enough. Removed the exception.

libminifi/src/core/yaml/YamlConfiguration.cpp
arpadboda: Good spot, thanks!

libminifi/src/core/ProcessGroup.cpp
arpadboda: In my opinion, yes, because of: 
-We can assume that in real production systems onSchedule errors are mostly transient (network, etc), so a retry is going to fix them. 
-What could we do to avoid such scenarios? We could do a BFS initiated from the failed processor and unschedule all the sources, but in my opinion that's even worse: it's quite complex and we just achieve immediately loosing data instead of buffering it to our connections.

libminifi/src/utils/CallBackTimer.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+void CallBackTimer::stop() {
+  std::lock_guard<std::recursive_mutex> guard(mtx_);
+  execute_ = false;
+  if (thd_.joinable()) {
+    thd_.join();
+  }
+}
+
+void CallBackTimer::start(std::function<void(void)> func) {
```
bakaid: This way we unnecessarily copy `func` twice, the parameter could be const& and we can just copy it into the lambda as it is now.

libminifi/src/utils/CallBackTimer.cpp
```diff
@@ -0,0 +1,71 @@
...
+  std::lock_guard<std::recursive_mutex> guard(mtx_);
+  if (execute_) {
+    stop();
+  }
+
+  execute_ = true;
+  thd_ = std::thread([this, func]() {
+                       std::this_thread::sleep_for(interval_);
```
bakaid: This could rewritten to use a condition_variable and wait_for to make the thread exit immediately on stop. This way it has to wait for the whole sleep.

libminifi/test/integration/OnScheduleErrorHandlingTests.cpp
```diff
@@ -0,0 +1,70 @@
...
+    unsigned int occurances = 0;
+    do {
+      pos = logs.find(minifi::processors::KamikazeProcessor::OnScheduleExceptionStr, pos);
+      if (pos != std::string::npos) {
+        pos = logs.find(minifi::processors::KamikazeProcessor::OnUnScheduleLogStr, pos);
+        if (pos != std::string::npos) {
+          occurances++;
+        }
+      }
+    } while (pos != std::string::npos && pos < logs.length());
```
bakaid: If `std::string::find` returned `pos >= logs.length()` without it being `std::string::npos` I would be scared out of my mind.

libminifi/src/utils/CallBackTimer.cpp
arpadboda: Done, tested with high timeout (100 sec), MiNiFi stops immediately.

libminifi/src/utils/CallBackTimer.cpp
arpadboda: Good spot, done.

libminifi/test/resources/TestOnScheduleRetry.yml
```diff
@@ -0,0 +1,40 @@
...
+      class: org.apache.nifi.processors.KamikazeProcessor
+      max concurrent tasks: 1
+      scheduling strategy: TIMER_DRIVEN
+      scheduling period: 100 sec
+      penalization period: 30 sec
+      yield period: 1 sec
+      run duration nanos: 0
+      auto-terminated relationships list:
+      Properties:
+        Throw in onSchedule: true
```
bakaid: This is a good test, but I would like to have other tests that test:
  - recovery (onSchedule throwing, then not throwing on retry, meaning onSchedule is not called anymore, onTrigger is called on the other hand)
 - onTrigger throwing and proper logs are created about the rollback (now that this sets it to `warn` instead of `debug`

libminifi/test/integration/OnScheduleErrorHandlingTests.cpp
arpadboda: Fair point, it's just a leftover of a previous version was pos was amended as well.

libminifi/test/resources/TestOnScheduleRetry.yml
arpadboda: Done, the integration test was extended with recovery and onTrigger calls as well.

arpadboda:
@phrocker : replied/addressed comments here, do you still have concerns?

PR #690, opened 2019-12-05
MINIFICPP-1026 - Added base64 encoder-decoder to StringUtils

bakaid:

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/include/utils/StringUtils.h
```diff
@@ -330,6 +336,194 @@ class StringUtils {
...
+      case 1:
+        return false;
+      case 2:
+        digits[2] = 0x00;
+      case 3: {
+        digits[3] = 0x00;
+
+        uint8_t bytes_temp[3];
+        base64_digits_to_bytes(digits, bytes_temp);
+        const size_t num_bytes = digit_counter == 2 ? 1 : 2;
```
arpadboda: As 2 and 3 are the only possible cases, it simply looks like digit_counter -1 :)

libminifi/include/utils/StringUtils.h
```diff
@@ -330,6 +336,194 @@ class StringUtils {
...
+
+  /**
+   * Decodes the Base64 encoded string into data
+   * @param data the output buffer where the decoded bytes will be written. Must be at least (base64_length / 4 + 1) * 3 bytes long.
+   * @param data_length pointer to the length of data the data buffer. It will be filled with the length of the decoded bytes.
+   * @param base64 the Base64 encoded string
+   * @param base64_length the length of base64
+   * @return true on success
+   */
+  inline static bool from_base64(uint8_t* data, size_t* data_length, const char* base64, size_t base64_length) {
```
arpadboda: I think the impl of these can be moved to the correspondig cpp file.
The already existing functions were left here to reduce the impact, but I think everything newly added should have the impl in cpp file unless it's a one liner to really be inlined.

phrocker: Is this your own impl or a copy and paste?

libminifi/include/utils/StringUtils.h
```diff
@@ -330,6 +336,194 @@ class StringUtils {
...
+   * Base64 decodes a string
+   * @param base64 the Base64 encoded string
+   * @param base64_length the length of base64
+   * @return the vector containing the decoded bytes
+   */
+  inline static std::vector<uint8_t> from_base64(const char* base64, size_t base64_length) {
+    std::vector<uint8_t> decoded((base64_length / 4 + 1) * 3);
+    size_t data_length = decoded.size();
+    if (!from_base64(decoded.data(), &data_length, base64, base64_length)) {
+      throw std::runtime_error("Base64 encoded string is malformatted");
```
phrocker: I always felt that runtime errors "are due to events beyond the scope of the program and can not be easily predicted." so when those are used versus internal or named exceptions they have different connotations. We catch them but this error seems recoverable from a programmatic perspective. I don't have a super strong opinion on this but it does have a bit of code smell worth mentioning

bakaid:
> It seems that we're leaning toward writing too much library code here. We build curl in most cases and in the cases where we can't we likely couldn't run anyway -- why not just rely on that which is already available versus writing our own.

 - cURL does not have a public base64 encode-decode API. We have been using a header file adapted from the base64 implementation of an older cURL version, and used it independently of whether we actually built cURL.
 - that implementation was a particularly bad one and has since been rewritten in cURL:
   - it contained a buffer overread with the comment `This final decode may actually read slightly past the end of the buffer if the input string is missing pad bytes.  This will almost always be harmless.`
   - it couldn't handle invalid characters in base64-encoded strings (didn't signal an error, just resulted in garbage)
   - it couldn't handle newlines, which are potentially valid in base64-encoded strings (didn't signal an error, just resulted in garbage)
   - it couldn't handle url-safe base64
 - the newer cURL implementation depends on other cURL-specific code. It could have been rewritten to be used independently of cURL, but I decided to create an own, well-tested implementation instead.
 - keep in mind that even if we used some base64 implementation from a third party we would still need
   - all the StringUtils functions that wrap it, except the encoding/decoding function
   - all the tests
 - "We build curl in most cases and in the cases where we can't we likely couldn't run anyway" is not a very convincing argument for anything. We should either make cURL build mandatory or fix the non-cURL build, if it indeed couldn't run.

libminifi/include/utils/StringUtils.h
bakaid: Own implementation, reasons explained in the main comment.

libminifi/include/utils/StringUtils.h
bakaid: That's fair, I'll think of something else.

libminifi/include/utils/StringUtils.h
bakaid: Po-tay-to po-tah-to, but if you really like `digit_counter -1` better, I can change it.

libminifi/include/utils/StringUtils.h
bakaid: I want to keep a few inline, but the largest ones can be moved to the cpp, you are right.

libminifi/include/utils/StringUtils.h
arpadboda: Just took me a while to understand what it does and why, I'm fine with either ways

bakaid: Changed it.

libminifi/include/utils/StringUtils.h
bakaid: Rewrote them to `std::invalid_argument`.

arpadboda:
LGTM, I'm happy to proceed with this if CIs pass.

phrocker:
>     * keep in mind that even if we used some base64 implementation from a third party we would still need
>       
>       * all the StringUtils functions that wrap it, except the encoding/decoding function
>       * all the tests
> 


This is debatable but a tested third party impl if one exists can be trusted. This is a lot of code with the potential for bugs. In your case you've written a lot of code, so I don't want to demean it but the code for an external third party with StringUtils seems much less risky with much lower tests -- unless we simply can't find a third party impl. What you say about cURL is a valid point. Did you attempt to find a third party example? If not I  can take a look -- there may be other string related libraries that provide this and some of what we do with StringUtils..

libminifi/test/unit/StringUtilsTests.cpp
```diff
@@ -155,28 +155,133 @@ TEST_CASE("TestStringUtils::testHexDecode", "[test hex decode]") {
...
+  REQUIRE_THROWS_WITH(StringUtils::from_base64("aa==?"), "Base64 encoded string is malformatted");
+  REQUIRE_THROWS_WITH(StringUtils::from_base64("aa==a"), "Base64 encoded string is malformatted");
+  REQUIRE_THROWS_WITH(StringUtils::from_base64("aa==="), "Base64 encoded string is malformatted");
+  REQUIRE_THROWS_WITH(StringUtils::from_base64("?"), "Base64 encoded string is malformatted");
+  REQUIRE_THROWS_WITH(StringUtils::from_base64("aaaa?"), "Base64 encoded string is malformatted");
+}
+
+TEST_CASE("TestStringUtils::testBase64EncodeDecode", "[test base64 encode decode]") {
+  std::mt19937 gen(std::random_device { }());
+  for (size_t i = 0U; i < 1024U; i++) {
```
phrocker: should add a test or scenario for arbitrarily large strings.

arpadboda:
> > ```
> > * keep in mind that even if we used some base64 implementation from a third party we would still need
> >   
> >   * all the StringUtils functions that wrap it, except the encoding/decoding function
> >   * all the tests
> > ```
> 
> This is debatable but a tested third party impl if one exists can be trusted. This is a lot of code with the potential for bugs. In your case you've written a lot of code, so I don't want to demean it but the code for an external third party with StringUtils seems much less risky with much lower tests -- unless we simply can't find a third party impl. What you say about cURL is a valid point. Did you attempt to find a third party example? If not I can take a look -- there may be other string related libraries that provide this and some of what we do with StringUtils..

Partly agree, partly disagree. 
For eg. boost does a lot of thing. But it is a huge dependency, heavily increases build time, etc. 
For most of the 3rd parties, in case we don't want to flood our own repo, we need to maintain cmake files, sometimes update dependencies, maybe edit bootstrap. These are also costs and they apply time to time when something needs to be updated. There are cases when these costs are definitely justified by the functionality the library introduces (just as opencv, curl, kafka, mqtt, etc...), but in case of some quite simple string util functionality the longer terms costs of having the dependency might show exceed the costs of doing it right, once.

bakaid:
> > ```
> > * keep in mind that even if we used some base64 implementation from a third party we would still need
> >   
> >   * all the StringUtils functions that wrap it, except the encoding/decoding function
> >   * all the tests
> > ```
> 
> This is debatable but a tested third party impl if one exists can be trusted. This is a lot of code with the potential for bugs. In your case you've written a lot of code, so I don't want to demean it but the code for an external third party with StringUtils seems much less risky with much lower tests -- unless we simply can't find a third party impl. What you say about cURL is a valid point. Did you attempt to find a third party example? If not I can take a look -- there may be other string related libraries that provide this and some of what we do with StringUtils..

@phrocker There are a myriad of base64 implementations in C, the two I meet the most often are https://web.mit.edu/freebsd/head/contrib/wpa/src/utils/base64.c and https://opensource.apple.com/source/QuickTimeStreamingServer/QuickTimeStreamingServer-452/CommonUtilitiesLib/base64.c
I have yet to find one that is at the same time
 - standalone
 - can pass the test suite I have written without modification
 - well tested in itself

The core part - that we would be able to substitute with a third-party implementation, albeit only after modifications - is composed of two look-up tables, a 30-line and a 65-line function.
I think if this is reviewed properly, given our test suite, the risk is pretty contained.

As for replacing the majority of StringUtils with some third party utility library: I don't know one that would fit our needs. There might well be, and in that I case I wouldn't be against it, but I don't want to do that integration work in the scope of this issue.

libminifi/test/unit/StringUtilsTests.cpp
bakaid: I thought about this, but after a few 4-base64-digit groups it doesn't involve different code paths, it would just test how large your memory is. What did you have in mind for size and what advantage would it have?

bakaid:
@phrocker Are you -1 on this? If so, please elaborate your technical justification for it, otherwise I would like to go forward with this PR.

arpadboda:
Merging this as no better alternative was proposed in a timely manner and it's clearly an improvement compared to the current (flawed) implementation we have. 

Feel free to open follow-ups in case any related issue or a promising out-of-the-box alternative is found.

phrocker:
> -1

Yep. I was 1. My gut suggestion was to use boost's base 64 since it's relatively isolated. If you look at the boost variant they mention portions come from this repo ( https://github.com/ReneNyffenegger/cpp-base64)

phrocker:
> Merging this as no better alternative was proposed in a timely manner and it's clearly an improvement compared to the current (flawed) implementation we have.
> 
> Feel free to open follow-ups in case any related issue or a promising out-of-the-box alternative is found.

The subjective precedent here should be clarified and defined clearly here : https://github.com/apache/nifi-minifi-cpp/blob/master/CONTRIB.md .

arpadboda:
> > Merging this as no better alternative was proposed in a timely manner and it's clearly an improvement compared to the current (flawed) implementation we have.
> > Feel free to open follow-ups in case any related issue or a promising out-of-the-box alternative is found.
> 
> The subjective precedent here should be clarified and defined clearly here : https://github.com/apache/nifi-minifi-cpp/blob/master/CONTRIB.md .

Please clarify!

PR #692, opened 2019-12-10
MINIFICPP-1099 - Fix Windows MSI build

bakaid:

### For all changes:
     in the commit message?

### For code changes:

### Note:


bakaid:
Still testing different configurations - @amarmer if you could try it with your on environment, that would be nice.

CMakeLists.txt
```diff
@@ -661,27 +661,54 @@ endif()
...
+
+	file(COPY "${CMAKE_CURRENT_SOURCE_DIR}/conf/" DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/conf/")
+	file(COPY "${CMAKE_CURRENT_SOURCE_DIR}/LICENSE" DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
+	file(COPY "${CMAKE_CURRENT_SOURCE_DIR}/README.md" DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
+	file(COPY "${CMAKE_CURRENT_SOURCE_DIR}/NOTICE" DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
+
+	# Determine the path of the VC Redistributable Merge Modules
+	if (DEFINED ENV{VCToolsRedistDir})
+		# Just get the redist dir that has been set by the build environment
+		set(VCRUNTIME_REDIST_DIR $ENV{VCToolsRedistDir})
```
bakaid: This seems like a pretty reliable method with 2017, it is both defined when building directly from Visual Studio, and when using a terminal vcvarsall.bat, but I included the backup method because I am not sure how reliable it is across VS versions.

win_build_vs.bat
```diff
@@ -23,6 +23,7 @@ set builddir=%1
 set skiptests=OFF
 set cmake_build_type=Release
 set build_type=Release
+set build_platform=Win32
```
bakaid: When using with vcvarsall.bat, it sets an environment variable which makes msbuild default to x86 as the default platform, not Win32, so we use this to make sure we use the proper Win32 platform.

CMakeSettings.json
```diff
@@ -3,7 +3,7 @@
   "configurations": [
     {
       "name": "x64-Release",
-      "generator": "Visual Studio 15 2017",
+      "generator": "Visual Studio 15 2017 Win64",
```
bakaid: This is a 64-bit build, it should use the 64-bit generator.

CMakeLists.txt
```diff
@@ -661,27 +661,54 @@ endif()
...
+	else()
+		# Try to fall back to a redist dir relative to the MSVC compiler
+		string(REGEX REPLACE "/VC/Tools/.*" "/VC/Redist/MSVC/" VCRUNTIME_REDIST_BASE_DIR ${CMAKE_C_COMPILER})
+		file(GLOB VCRUNTIME_REDIST_VERSIONS "${VCRUNTIME_REDIST_BASE_DIR}/*")
+		if (NOT VCRUNTIME_REDIST_VERSIONS)
+			message(FATAL_ERROR "Could not find the VC Redistributable Merge Modules. Please set VCRUNTIME_X86_MERGEMODULE_PATH and VCRUNTIME_X64_MERGEMODULE_PATH manually!")
+		endif()
+		list(SORT VCRUNTIME_REDIST_VERSIONS)
+		list(REVERSE VCRUNTIME_REDIST_VERSIONS)
+		list(GET VCRUNTIME_REDIST_VERSIONS 0 VCRUNTIME_REDIST_DIR)
```
arpadboda: I think a comment here would be nice. Guess it finds the latest one, which happens to be the last element after sorting the list.

CMakeLists.txt
```diff
@@ -661,27 +661,54 @@ endif()
...
+	else()
+		list(GET VCRUNTIME_X86_MERGEMODULES 0 VCRUNTIME_X86_MERGEMODULE_PATH)
+		list(GET VCRUNTIME_X64_MERGEMODULES 0 VCRUNTIME_X64_MERGEMODULE_PATH)
+	endif()
 
+	if(CMAKE_SIZEOF_VOID_P EQUAL 8)
+		message("Using ${VCRUNTIME_X64_MERGEMODULE_PATH} VC Redistributable Merge Module")
+		configure_file("msi/x64.wsi" "msi/x64.wsi" @ONLY)
+		set(CPACK_WIX_EXTRA_SOURCES "${CMAKE_CURRENT_BINARY_DIR}/msi/x64.wsi")
+	elseif(CMAKE_SIZEOF_VOID_P EQUAL 4)
```
arpadboda: Nitpicking, but either this should be a simple "else", or there should be a fallback branch to exit with error msg in case we can't determine.

bakaid: This isn't new code, but I agree. Even though void* not being 8 or 4 would surprise me quite a lot.

CMakeLists.txt
bakaid: Yep, hopefully it does that. Will add a comment.

am-c-p-p:
I tested. LGTM!

bakaid:
> I tested. LGTM!
> It would be good to test minifi.exe (without using installer) on machine with old redistributables, (maybe on Windows 7) and verify that it doesn't work and then install it and check that it works.

@amarmer Thank you for trying it.
I have tested it on a Windows 7 instance, and I verified that if I had a preinstalled VC 2015 Runtime, our msi really did update the dlls to a newer version.
I couldn't try it from a not working state, as the older DLLs are still binary compatible, even if it is unsupported to use an older version than our code was built with - this is the reason we have been able to run on Windows at all so far.

arpadboda:
Tested the build On Win10 X64 with VS2017 community, builds as expected, installer works, MiNiFi starts after installation.

PR #707, opened 2020-01-09
MINIFICPP-1117 - minifi::Exception is now nothrow copyable

szaszm:
Implemented in terms of std::runtime_error

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/Exception.h
+++ b/libminifi/include/Exception.h
@@ -31,7 +31,6 @@ namespace apache {
 namespace nifi {
 namespace minifi {
 
-// ExceptionType 
 enum ExceptionType {
   FILE_OPERATION_EXCEPTION = 0,
   FLOW_EXCEPTION,
@@ -44,11 +43,9 @@ enum ExceptionType {
   MAX_EXCEPTION
 };
 
-// Exception String 
 static const char *ExceptionStr[MAX_EXCEPTION] = { "File Operation", "Flow File Operation", "Processor Operation", "Process Session Operation", "Process Schedule Operation", "Site2Site Protocol",
     "General Operation", "Regex Operation" };
 
-// Exception Type to String 
 inline const char *ExceptionTypeToString(ExceptionType type) {
   if (type < MAX_EXCEPTION)
     return ExceptionStr[type];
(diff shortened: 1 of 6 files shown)


szaszm:
I've added some boilerplate for string joining. I can remove it at some performance cost (extra `std::string` allocations) or additional dependency (`boost::algorithm::join`)

libminifi/include/Exception.h
```diff
@@ -44,49 +43,56 @@ enum ExceptionType {
...
 
- private:
-  // Exception type
-  ExceptionType _type;
-  // Exception detailed information
-  std::string _errorMsg;
-  // Hold the what result
-  mutable std::string _whatStr;
+template<typename... Strs>
+std::string StringJoin(Strs&&... strs) {
```
nghiaxlee: I think this should be moved to our StringUtils if we haven't had this feature yet.

arpadboda: Good point, we do have a join there.

libminifi/include/Exception.h
```diff
@@ -44,49 +43,56 @@ enum ExceptionType {
...
-  }
+namespace detail {
+inline size_t StringLength(const char* str) { return strlen(str); }
 
-  // Destructor
-  virtual ~Exception() noexcept {
-  }
-  virtual const char * what() const noexcept {
+template<size_t L>
+constexpr size_t StringLength(const char (&str)[L]) { return L; }
```
szaszm: bug here: L is the size of the array, not the string inside. I'll change it to `max(L-1, 0)` and leave the case of early null termination unhandled if the functionality is not already present in StringUtils.

libminifi/include/Exception.h
szaszm: moved to StringUtils in dbc0033

szaszm:
update:
- fix `ProcessorTests`
- move string utilities to `StringUtils` and generalize them
- add unit test to cover the new `StringUtils::join_pack`
Moving the utilities from a private namespace to `StringUtils` means that they need to be more general IMO. I've extended them to work with any character type and do SFINAE checking on the string types. 
This means quite some template code, which may not be welcome in this codebase. Suggestions are welcome.

PR #708, opened 2020-01-13
MINIFICPP-1118 - MiNiFi C++ on Windows stops running in a secure env …

arpadboda:
…when NiFi becomes unreachable

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/src/io/tls/TLSSocket.cpp
+++ b/libminifi/src/io/tls/TLSSocket.cpp
@@ -24,6 +24,7 @@
 #include "io/tls/TLSSocket.h"
 #include "io/tls/TLSUtils.h"
 #include "properties/Configure.h"
+#include "utils/ScopeGuard.h"
 #include "utils/StringUtils.h"
 #include "core/Property.h"
 #include "core/logging/LoggerConfiguration.h"
@@ -39,8 +40,8 @@ std::mutex OpenSSLInitializer::context_mutex;
 
 TLSContext::TLSContext(const std::shared_ptr<Configure> &configure, const std::shared_ptr<minifi::controllers::SSLContextService> &ssl_service)
     : SocketContext(configure),
-      error_value(0),
-      ctx(0),
+      error_value(TLS_GOOD),
+      ctx(nullptr),
       logger_(logging::LoggerFactory<TLSContext>::getLogger()),
       configure_(configure),
       ssl_service_(ssl_service) {
@@ -73,6 +74,12 @@ int16_t TLSContext::initialize(bool server_method) {
     error_value = TLS_ERROR_CONTEXT;
     return error_value;
   }
+
+  utils::ScopeGuard ctxGuard([this]() {
+    SSL_CTX_free(ctx);
+    ctx = nullptr;
+  });
+
   if (needClientCert) {
     std::string certificate;
     std::string privatekey;
@@ -84,6 +91,8 @@ int16_t TLSContext::initialize(bool server_method) {
         error_value = TLS_ERROR_CERT_ERROR;
         return error_value;
       }
+      ctxGuard.disable();
+      error_value = TLS_GOOD;
       return 0;
     }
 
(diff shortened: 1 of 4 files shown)


libminifi/src/io/tls/TLSSocket.cpp
```diff
@@ -182,7 +182,15 @@ int16_t TLSSocket::initialize(bool blocking) {
     setNonBlocking();
   logger_->log_trace("Initializing TLSSocket %d", is_server);
   int16_t ret = context_->initialize(is_server);
-  Socket::initialize();
+
+  if (ret == 0) {
```
bakaid: This control flow is not very readable this way, with `ret` controlling what parts of the code run, sometimes with `== 0`, sometimes with `!= 0` and sometimes with `!ret`. Is there a reason we can't just log and `return -1` each time something fails? I think it would be more readable and would require less nested ifs.

libminifi/src/io/tls/TLSSocket.cpp
```diff
@@ -200,17 +208,12 @@ int16_t TLSSocket::initialize(bool blocking) {
         logger_->log_trace("want read");
         return 0;
       } else {
+        logger_->log_error("SSL socket connect failed to %s %d", requested_hostname_, port_);
+        SSL_free(ssl_);
+        ssl_ = NULL;
+        Socket::closeStream();
```
szaszm: These lines for cleanup are present in multiple places in the file. Did you consider extracting to a function or utilizing RAII for cleanup?

libminifi/src/io/tls/TLSSocket.cpp
arpadboda: Fair point, simplified a bit.

libminifi/src/io/tls/TLSSocket.cpp
arpadboda: Dtor cleans up anyway, but I was not brave enough to assume that socket objects are never reused. :) 

Simplified cleanup logic anyway.

libminifi/src/io/tls/TLSSocket.cpp
```diff
@@ -182,9 +187,19 @@ int16_t TLSSocket::initialize(bool blocking) {
...
-  if (!ret && listeners_ == 0) {
+  if (ret != 0) {
+    logger_->log_warn("Failed to initialize SSL context!");
+    return -1;
+  }
+
+  ret = Socket::initialize();
+  if (ret != 0) {
+    logger_->log_warn("Failed to initialise basic socket for TLS socket");
+    return -1;
```
szaszm: `context_` is left initialized even if socket initialization has failed. This is not changed behavior, but is it intended?

Speaking about suspicious behavior, `TLSSocket::logger_` will be `nullptr` if the object is initialized via the move ctor.

szaszm: I was wrong about the `logger_` part, since both `Socket` and `TLSSocket` have their own separate private `logger_`s, so move will work.

arpadboda: Unified loggers to only have one.
Made context init reentrant.

szaszm:
@bakaid `make`, `make linter` and `make test` resulted in no errors on my linux installation.

bakaid:
@szaszm I'm glad, but I am interested in whether it s2s actually works with a NiFi instance over SSL, and I don't believe we have unit/integration tests for that.

arpadboda:
> @arpadboda review fixes look good. Did you test this on Linux as well?

We have a test called "SecureSocketGetTCPTest" (in standard processors extension), which verifies communication via TLS sockets, so together with the Win verification I trust it. 
What's your opinion?

bakaid:
> > @arpadboda review fixes look good. Did you test this on Linux as well?
> 
> We have a test called "SecureSocketGetTCPTest" (in standard processors extension), which verifies communication via TLS sockets, so together with the Win verification I trust it.
> What's your opinion?

Yep, for the time being it should be OK, but I think it warrants a follow-up task to:
 - test the name resolution failure path in our integration test
 - create a docker test to actually test at least the positive (and preferable negative) cases with a real NiFi instance

PR #710, opened 2020-01-15
MINIFICPP - 1110,1111 - PublishKafka, OPC processors should config and

nghiaxlee:
throw in onSchedule

* Finished OPC processors
* Working on Kafka processor

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/opc/src/fetchopc.cpp
```diff
@@ -93,14 +93,12 @@ namespace processors {
...
 
-    BaseOPCProcessor::onSchedule(context, factory);
-
-    if(!configOK_) {
-      return;
+    try {
+      BaseOPCProcessor::onSchedule(context, factory);
+    } catch (const Exception &e) {
+      throw e;
     }
```
szaszm: What's the intention behind these lines? We catch `minifi::Exception`s then rethrow them without doing anything.

extensions/librdkafka/PublishKafka.cpp
```diff
@@ -187,81 +239,77 @@ bool PublishKafka::configureNewConnection(const std::shared_ptr<KafkaConnection>
...
-    logger_->log_error("There are no brokers");
-    return false;
+    throw Exception(PROCESS_SCHEDULE_EXCEPTION, "There are no brokers");
   }
   result = rd_kafka_conf_set(conf_, "bootstrap.servers", key->brokers_.c_str(), errstr.data(), errstr.size());
   logger_->log_debug("PublishKafka: bootstrap.servers [%s]", key->brokers_);
   if (result != RD_KAFKA_CONF_OK) {
-    logger_->log_error("PublishKafka: configure error result [%s]", errstr.data());
-    return false;
+    auto error_msg = utils::StringUtils::join_pack("PublishKafka: configure error result [%s]", errstr.data());
```
arpadboda: I don't think it's gonna work as you expect. It just joins them, doesn't work as a format string. 

What you mean is more like:
```
auto error_msg = utils::StringUtils::join_pack("PublishKafka: configure error result [", errstr.data(), "]");
```

extensions/librdkafka/PublishKafka.cpp
```diff
@@ -187,81 +239,77 @@ bool PublishKafka::configureNewConnection(const std::shared_ptr<KafkaConnection>
...
-    logger_->log_error("Client id is empty");
-    return false;
+    throw Exception(PROCESS_SCHEDULE_EXCEPTION, "Client id is empty");
   }
   result = rd_kafka_conf_set(conf_, "client.id", key->client_id_.c_str(), errstr.data(), errstr.size());
   logger_->log_debug("PublishKafka: client.id [%s]", key->client_id_);
   if (result != RD_KAFKA_CONF_OK) {
-    logger_->log_error("PublishKafka: configure error result [%s]", errstr.data());
-    return false;
+    auto error_msg = utils::StringUtils::join_pack("PublishKafka: configure error result ", errstr.data());
```
arpadboda: Nitpicking: a ':' would be nice to indicate where the error string begins

The same applies for all these cases.

Altough it's used in many places, so I think it would be better to prepar the error string and just concat the actual error when throwing the exception to avoid multiplication of this code snippet.

extensions/librdkafka/tests/PublishKafkaOnScheduleTests.cpp
```diff
@@ -0,0 +1,84 @@
...
+            last_pos = pos;
+            occurances++;
+          }
+        }
+      } while (pos != std::string::npos);
+
+      assert(occurances > 1);  // Verify retry of onSchedule and onUnSchedule calls
+
+      // Make sure onSchedule succeeded after property was set
+      assert(logs.find("Successfully configured PublishKafka", last_pos) != std::string::npos);
```
arpadboda: I like this test, nice work! (Even though some parts make me feel deja vu :) )

extensions/librdkafka/tests/PublishKafkaOnScheduleTests.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+class PublishKafkaOnScheduleTests : public IntegrationBase {
+public:
+    virtual void runAssertions() {
+      std::string logs = LogTestController::getInstance().log_output.str();
+      size_t pos = 0;
+      size_t last_pos = 0;
+      unsigned int occurances = 0;
+      do {
+        pos = logs.find(" value 1 is outside allowed range 1000..1000000000", pos);
```
arpadboda: As this logic is not so nice ( I admit to be the guilty one responsible for this :) ), I would prefer not to copy-paste it, but create a function like "countMessagePairsInLog" and make it to be part of test framework. 
If you have a better idea for naming this, I'm happy to see that.

extensions/librdkafka/tests/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+
+SET(KAFKA_TEST_COUNT 0)
+
+FOREACH(testfile ${KAFKA_TESTS})
+    get_filename_component(testfilename "${testfile}" NAME_WE)
+    add_executable("${testfilename}" "${testfile}")
+    target_include_directories(${testfilename} PRIVATE BEFORE "${CMAKE_SOURCE_DIR}/extensions/librdkafka/")
+    target_include_directories(${testfilename} PRIVATE BEFORE "${CMAKE_SOURCE_DIR}/extensions/standard-processors")
+    target_wholearchive_library(${testfilename} minifi-rdkafka-extensions)
+    target_wholearchive_library(${testfilename} minifi-standard-processors)
```
arpadboda: Why do we need standard processors to be included and linked to this test?

extensions/librdkafka/tests/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+    createTests("${testfilename}")
+    MATH(EXPR KAFKA_TEST_COUNT "${KAFKA_TEST_COUNT}+1")
+    if (${testfilename} MATCHES "(.)*OnSchedule(.)*")
```
arpadboda: What's the point this? Did you try to prepare this test file to have different kind of Kafka tests?

extensions/standard-processors/processors/LogAttribute.cpp
```diff
@@ -95,7 +95,7 @@ void LogAttribute::onSchedule(const std::shared_ptr<core::ProcessContext> &conte
     // in configuration. In future releases we can add that exception handling there.
     if (!flowsToLog.getValue().validate("Validating FlowFilesToLog").valid())
       throw Exception(PROCESS_SCHEDULE_EXCEPTION, "Invalid value for flowfiles to log: " + flowsToLog.getValue().to_string());
-    flowfiles_to_log_ = flowsToLog.getValue();
+    flowfiles_to_log_ = static_cast<uint64_t>(flowsToLog.getValue());
```
arpadboda: There is a getProperty template to get typed value, so I think the following does the job:
```
getProperty(FlowFilesToLog.getName(), flowfiles_to_log_);
```

Didn't try though

libminifi/src/core/PropertyValidation.cpp
```diff
@@ -27,6 +27,7 @@ std::shared_ptr<PropertyValidator> StandardValidators::VALID = std::make_shared<
 StandardValidators::StandardValidators() {
   INVALID = std::make_shared<AlwaysValid>(false, "INVALID");
   INTEGER_VALIDATOR = std::make_shared<IntegerValidator>("INTEGER_VALIDATOR");
+  UNSIGNED_INT_VALIDATOR = std::make_shared<UnsignedIntValidator>("UNSIGNED_INT_VALIDATOR");
```
arpadboda: To ensure compatibility with MiNiFi java via C2 (having matching manifest in case of validators) please make sure about UInt32 validator exists in NiFi and it has the same name. ( I can help with this)

nghiaxlee: https://github.com/apache/nifi/blob/master/nifi-commons/nifi-utils/src/main/java/org/apache/nifi/processor/util/StandardValidators.java#L292 :sweat:

extensions/librdkafka/PublishKafka.cpp
szaszm: Maybe we could use the formatting library used by spdlog to create these strings? I've mentioned directly depending on the formatting library in [MINIFICPP-1121](https://issues.apache.org/jira/projects/MINIFICPP/issues/MINIFICPP-1121).

nghiaxlee: I actually did not expect I forgot this one :), @arpadboda what do you think about adding formatting library?

extensions/librdkafka/tests/CMakeLists.txt
nghiaxlee: Just a workaround for myself, I have some unit tests which require local Kafka server but never commit them. Should I add them in or modify this file for this test only? (will rewrite anyway because that line looks weird)

libminifi/src/core/PropertyValidation.cpp
arpadboda: Fair, but not sure whether that's 32 or 64 bits.

extensions/librdkafka/PublishKafka.cpp
arpadboda: I'm fine with that, but please make sure you do that in a separate commit and in case we start using that directly we should add the cmake dependency directly as well. 
https://github.com/apache/nifi-minifi-cpp/blob/master/ThirdParties.md

libminifi/src/core/PropertyValidation.cpp
szaszm: Java `int` is 32 bit signed, but the validator filters all the negatives so it's effectively 31 bit.

nghiaxlee: Side note: IIRC, Java 8 provides some Unsinged methods for Integer class so that it can work in range 0..2^32-1.

extensions/librdkafka/PublishKafka.cpp
```diff
@@ -141,6 +141,59 @@ void PublishKafka::initialize() {
...
+  } else {
+    target_batch_payload_size_ = 512 * 1024U;
+  }
+
+  // Max Flow Segment Size
+  value = "";
+  if (context->getProperty(MaxFlowSegSize.getName(), max_flow_seg_size_)) {
+    logger_->log_debug("PublishKafka: Max Flow Segment Size [%llu]", max_flow_seg_size_);
+  } else {
+    max_flow_seg_size_ = 0U;
```
arpadboda: Instead of doing this, why don't we simply add default values to these properties, so getProperty always succeeds?

extensions/standard-processors/processors/LogAttribute.cpp
```diff
@@ -88,23 +88,14 @@ void LogAttribute::initialize() {
...
-  }
 
-  std::string value;
-  if (context->getProperty(HexencodePayload.getName(), value)) {
-    utils::StringUtils::StringToBool(value, hexencode_);
-  }
-  if (context->getProperty(MaxPayloadLineLength.getName(), value)) {
-    core::Property::StringToInt(value, max_line_length_);
-  }
+  context->getProperty(FlowFilesToLog.getName(), flowfiles_to_log_);
```
nghiaxlee: This line, if I wrote like: 
```
if (!context->getProperty(FlowFilesToLog.getName(), flowfiles_to_log_)) {
      throw ...
}
```
Then the GetTCPTest (only this) will fail due to exception, so I guess that test's configuration has issues, will take a look later.

extensions/standard-processors/processors/LogAttribute.cpp
```diff
@@ -88,23 +88,14 @@ void LogAttribute::initialize() {
...
 void LogAttribute::onSchedule(const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSessionFactory> &factory) {
-  core::Property flowsToLog = FlowFilesToLog;
-
-  if (getProperty(FlowFilesToLog.getName(), flowsToLog)) {
-    // we are going this route since to avoid breaking backwards compatibility the get property function doesn't perform validation ( That's done
-    // in configuration. In future releases we can add that exception handling there.
-    if (!flowsToLog.getValue().validate("Validating FlowFilesToLog").valid())
-      throw Exception(PROCESS_SCHEDULE_EXCEPTION, "Invalid value for flowfiles to log: " + flowsToLog.getValue().to_string());
-    flowfiles_to_log_ = flowsToLog.getValue();
-  }
```
nghiaxlee: This line, if I wrote like: 
```
if (!context->getProperty(FlowFilesToLog.getName(), flowfiles_to_log_)) {
      throw ...
}
```
Then the GetTCPTest (only this) will fail due to exception, so I guess that test's configuration has issues, will take a look later.

extensions/librdkafka/PublishKafka.cpp
nghiaxlee: Fixed, we already had default value, maybe we were overcareful

extensions/librdkafka/tests/PublishKafkaOnScheduleTests.cpp
nghiaxlee: Wrapped the logic to a new function

extensions/standard-processors/processors/LogAttribute.cpp
arpadboda: You are right. 

Initialise() is not called in the testcases, which would set the default value. 
Added that call to every testcase fixes the issue.

extensions/librdkafka/PublishKafka.cpp
nghiaxlee: okay, then I will do a follow-up refactoring later

libminifi/include/core/state/Value.h
```diff
@@ -188,6 +264,18 @@ class BoolValue : public Value {
     }
   }
 
+  virtual bool getValue(uint32_t &ref) {
```
arpadboda: Please, no!

Yes, it was already horrible, but that doesn't justify copy-pasting the same! :)

```
template<typename Integral, typename std::enable_if<
    (sizeof(Integral) >= 4) && std::is_integral<Integral>::value
    ,Integral>::type* = nullptr>
virtual bool getValue(Integral &ref) {
  if(ref != 0 && ref != 1) {
   return false;
  }
 ref = (ref == 1);
 return true;
}
```

libminifi/src/core/PropertyValidation.cpp
```diff
@@ -27,6 +27,7 @@ std::shared_ptr<PropertyValidator> StandardValidators::VALID = std::make_shared<
 StandardValidators::StandardValidators() {
   INVALID = std::make_shared<AlwaysValid>(false, "INVALID");
   INTEGER_VALIDATOR = std::make_shared<IntegerValidator>("INTEGER_VALIDATOR");
+  UNSIGNED_INT_VALIDATOR = std::make_shared<UnsignedIntValidator>("NON_NEGATIVE_INTEGER_VALIDATOR");
```
arpadboda: 👍

libminifi/test/integration/OnScheduleErrorHandlingTests.cpp
```diff
@@ -1,6 +1,4 @@
 /**
- * @file GenerateFlowFile.h
```
arpadboda: Thanks :)

libminifi/include/core/state/Value.h
```diff
@@ -87,6 +88,15 @@ class Value {
     type_id = std::type_index(typeid(T));
   }
 
+  virtual bool getValue(uint32_t &ref) {
+    const auto negative = string_value.find_first_of('-') != std::string::npos;
```
arpadboda: You don't really need the variable here, simply put this in the "if"

libminifi/include/core/state/Value.h
```diff
@@ -115,6 +125,64 @@ class Value {
...
+
+  explicit UInt32Value(const std::string &strvalue)
+      : Value(strvalue),
+        value(std::stoul(strvalue)) {
+    /**
+     * This is a fundamental change in that we would be changing where this error occurs.
+     * We should be prudent about breaking backwards compatibility, but since Uint32Value
+     * is only created with a validator and type, we **should** be okay.
+     */
+    const auto negative = strvalue.find_first_of('-') != std::string::npos;
```
arpadboda: Same here

libminifi/test/integration/IntegrationBase.h
```diff
@@ -42,6 +42,17 @@ class IntegrationBase {
     configureSecurity();
   }
 
+  // Return the last position and number of occurrences.
+  std::pair<size_t, uint32_t> countPatInStr(const std::string &str, const std::string &pattern) {
+    size_t last_pos = 0;
+    unsigned int occurrences = 0;
```
arpadboda: Please be consistent with types, it's returned as uint32_t

libminifi/test/integration/OnScheduleErrorHandlingTests.cpp
```diff
@@ -29,32 +27,22 @@ class OnScheduleErrorHandlingTests : public IntegrationBase {
...
-          last_pos = pos;
-          occurances++;
-        }
-      }
-    } while (pos != std::string::npos);
 
-    assert(occurances > 1);  // Verify retry of onSchedule and onUnSchedule calls
+    auto result = countPatInStr(logs, minifi::processors::KamikazeProcessor::OnScheduleExceptionStr);
+    size_t last_pos = result.first;
+    unsigned int occurances = result.second;
```
arpadboda: Will need to change this to uint32_t as well

libminifi/test/integration/IntegrationBase.h
szaszm: Indeed, it should be consistent. This is minor, but I'll write it down anyway.
I don't think the count type should be `uint32_t`, since there's nothing that mandates 32 bits of storage for our integer.
The position is fine as `size_t`.

Some candidates:
- `int`: plain and simple, we return an integer
- `unsigned int`: as above, but a count can not be smaller than zero and we can signal this with the type
- `size_t`: The max. possible count of matches, in the case of a theoretical string spanning the full addressable memory and a search string matching on every character.
- `ptrdiff_t`: The max. possible count of matches is the size of the string, that is the difference between the last and first element pointers. Note: signed integer
- `typename iterator_traits<std::string::iterator>::difference_type`: Same rationale as for `ptrdiff_t`, but for last and first iterators. This is what `std::count` returns.

My suggestion would be a simple `int`, `unsigned int` or a correct `typename iterator_traits<std::string::iterator>::difference_type`, in the order of my preference.

libminifi/test/integration/OnScheduleErrorHandlingTests.cpp
szaszm: typo: it's spelled occurence

libminifi/include/core/state/Value.h
szaszm: @arpadboda: The name helps readability and since it's `const` it doesn't increase complexity. (by potentially introducing extra states)

libminifi/include/core/state/Value.h
arpadboda: The implementation is also bad smell, I agree, didn't check that. 

In case the functions are virtual we should introduce a templated private helper and call that from the virtuals. Still better than 4 times duplicated code.

libminifi/test/integration/IntegrationBase.h
arpadboda: I'm fine with size_t as well, I don't really mind which one we choose, just make sure it's consistent.

libminifi/include/core/state/Value.h
arpadboda: In my opinion the variable just decreases readability here, but this is really minor, I'm fine with the current way as well. 

Although it's simply wrong. 
It's going to fail in case the '-' char is after the digits (like "3-"), which std::stoull handles. 
Either let it go or go strict and apply isdigit check for all.

I will create a Jira for this, should be done in scope of a separate PR.

libminifi/include/core/state/Value.h
```diff
@@ -115,6 +125,64 @@ class Value {
...
+  }
+ protected:
+
+  virtual bool getValue(uint32_t &ref) {
+    ref = value;
+    return true;
+  }
+
+  virtual bool getValue(int &ref) {
+    if (value <= (std::numeric_limits<int>::max)()) {
```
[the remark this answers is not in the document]
nghiaxlee: Could you elaborate it?
This is derived from previous impl of uint64, but I think it makes sense (in logic, not sure that's actually what we want).
From my understanding, this check whether `value`(0..2^32-1) is smaller than 2^31-1 or not, if it is smaller then we assign to `ref`

libminifi/test/integration/IntegrationBase.h
nghiaxlee: Done, @szaszm thanks for the detailed explanation. ^^

libminifi/include/core/state/Value.h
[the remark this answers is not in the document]
szaszm: My bad, I confused `value` with `ref`

libminifi/include/core/state/Value.h
nghiaxlee: Will keep the current impl for now

nghiaxlee:
Rebased and resolved conflicts

libminifi/include/core/state/Value.h
```diff
@@ -218,6 +274,16 @@ class BoolValue : public Value {
   }
 
   bool value;
+
+ private:
+  template<typename T>
+  bool PreventSwearingInFutureRefactor(T &ref) {
```
arpadboda: 😄

PR #712, opened 2020-01-20
MINIFICPP-1047 Add property "Drop empty flow files" to PublishKafka

szaszm:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/librdkafka/PublishKafka.cpp
+++ b/extensions/librdkafka/PublishKafka.cpp
@@ -99,14 +99,13 @@ core::Property PublishKafka::MessageKeyField("Message Key Field", "The name of a
                                              "");
 core::Property PublishKafka::DebugContexts("Debug contexts", "A comma-separated list of debug contexts to enable."
                                            "Including: generic, broker, topic, metadata, feature, queue, msg, protocol, cgrp, security, fetch, interceptor, plugin, consumer, admin, eos, all", "");
-const core::Property PublishKafka::DropEmptyFlowFiles(
-    core::PropertyBuilder::createProperty("Drop empty flow files")
-        ->withDescription("Keep backwards compatibility with <=0.7.0 bug which caused flow files with empty content to not be published to Kafka. The old behavior is deprecated. Use connections "
-                          "to drop empty flow files!")
+const core::Property PublishKafka::FailEmptyFlowFiles(
+    core::PropertyBuilder::createProperty("Fail empty flow files")
+        ->withDescription("Keep backwards compatibility with <=0.7.0 bug which caused flow files with empty content to not be published to Kafka and forwarded to failure. The old behavior is "
+                          "deprecated. Use connections to drop empty flow files!")
         ->isRequired(false)
         ->withDefaultValue<bool>(true)
-        ->build()
-);
+        ->build());
 
 core::Relationship PublishKafka::Success("success", "Any FlowFile that is successfully sent to Kafka will be routed to this Relationship");
 core::Relationship PublishKafka::Failure("failure", "Any FlowFile that cannot be sent to Kafka will be routed to this Relationship");
@@ -139,7 +138,7 @@ void PublishKafka::initialize() {
   properties.insert(KerberosKeytabPath);
   properties.insert(MessageKeyField);
   properties.insert(DebugContexts);
-  properties.insert(DropEmptyFlowFiles);
+  properties.insert(FailEmptyFlowFiles);
   setSupportedProperties(properties);
   // Set the supported relationships
   std::set<core::Relationship> relationships;
@@ -644,18 +643,23 @@ void PublishKafka::onTrigger(const std::shared_ptr<core::ProcessContext> &contex
       continue;
     }
 
+    bool failEmptyFlowFiles = true;
+    context->getProperty(FailEmptyFlowFiles.getName(), failEmptyFlowFiles);
+
     PublishKafka::ReadCallback callback(max_flow_seg_size, kafkaKey, thisTopic->getTopic(), conn->getConnection(), flowFile,
-                                        attributeNameRegex, messages, flow_file_index);
+                                        attributeNameRegex, messages, flow_file_index, failEmptyFlowFiles);
     session->read(flowFile, &callback);
 
-    bool dropEmptyFlowFiles;
-    if (!callback.called_ && context->getProperty(DropEmptyFlowFiles.getName(), dropEmptyFlowFiles) && !dropEmptyFlowFiles) {
-      // workaround: call callback since ProcessSession doesn't do so for empty flow files
-      logger_->log_debug("ReadCallback workaround on empty flow file, because DropEmptyFlowFiles is false, uuid: %s", flowFile->getUUIDStr());
+    if (!callback.called_) {
+      // workaround: call callback since ProcessSession doesn't do so for empty flow files without resource claims
       callback.process(nullptr);
     }
-    if (!callback.called_ && dropEmptyFlowFiles) {
-      logger_->log_info("Deprecated behavior, use connections to drop empty flow files! Dropped empty flow file with uuid: %s", flowFile->getUUIDStr());
+
+    if (flowFile->getSize() == 0 && failEmptyFlowFiles) {
+      logger_->log_debug("Deprecated behavior, use connections to drop empty flow files! Failing empty flow file with uuid: %s", flowFile->getUUIDStr());
+      messages->modifyResult(flow_file_index, [](FlowFileResult& flow_file_result) {
+        flow_file_result.flow_file_error = true;
+      });
     }
 
     if (callback.status_ < 0) {
--- a/extensions/librdkafka/PublishKafka.cpp
+++ b/extensions/librdkafka/PublishKafka.cpp
@@ -646,7 +646,7 @@ void PublishKafka::onTrigger(const std::shared_ptr<core::ProcessContext> &contex
     bool failEmptyFlowFiles = true;
     context->getProperty(FailEmptyFlowFiles.getName(), failEmptyFlowFiles);
 
-    PublishKafka::ReadCallback callback(max_flow_seg_size, kafkaKey, thisTopic->getTopic(), conn->getConnection(), flowFile,
+    PublishKafka::ReadCallback callback(max_flow_seg_size, kafkaKey, thisTopic->getTopic(), conn->getConnection(), *flowFile,
                                         attributeNameRegex, messages, flow_file_index, failEmptyFlowFiles);
     session->read(flowFile, &callback);
 
--- a/extensions/librdkafka/PublishKafka.cpp
+++ b/extensions/librdkafka/PublishKafka.cpp
@@ -18,7 +18,7 @@
  * limitations under the License.
  */
 #include "PublishKafka.h"
-#include <stdio.h>
+#include <cstdio>
 #include <algorithm>
 #include <memory>
 #include <string>
@@ -99,6 +99,14 @@ core::Property PublishKafka::MessageKeyField("Message Key Field", "The name of a
                                              "");
 core::Property PublishKafka::DebugContexts("Debug contexts", "A comma-separated list of debug contexts to enable."
                                            "Including: generic, broker, topic, metadata, feature, queue, msg, protocol, cgrp, security, fetch, interceptor, plugin, consumer, admin, eos, all", "");
+const core::Property PublishKafka::DropEmptyFlowFiles(
+    core::PropertyBuilder::createProperty("Drop empty flow files")
+        ->withDescription("Keep backwards compatibility with <=0.7.0 bug which caused flow files with empty content to not be published to Kafka. The old behavior is deprecated. Use connections "
+                          "to drop empty flow files!")
+        ->isRequired(false)
+        ->withDefaultValue<bool>(true)
+        ->build()
+);
 
 core::Relationship PublishKafka::Success("success", "Any FlowFile that is successfully sent to Kafka will be routed to this Relationship");
 core::Relationship PublishKafka::Failure("failure", "Any FlowFile that cannot be sent to Kafka will be routed to this Relationship");
@@ -131,6 +139,7 @@ void PublishKafka::initialize() {
   properties.insert(KerberosKeytabPath);
   properties.insert(MessageKeyField);
   properties.insert(DebugContexts);
+  properties.insert(DropEmptyFlowFiles);
   setSupportedProperties(properties);
   // Set the supported relationships
   std::set<core::Relationship> relationships;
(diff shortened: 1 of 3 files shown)


szaszm:
What is your opinion on using `const` by default?

Pros:
* extra static analysis -> bugs caught earlier
* intent stated more explicitly -> more readable

Cons:
* more typing
* more clutter -> less readable
* not consistent with the current state of the codebase
  * but can be viewed as gradual evolution if we're in favor of this style

extensions/librdkafka/PublishKafka.cpp
```diff
@@ -639,6 +647,17 @@ void PublishKafka::onTrigger(const std::shared_ptr<core::ProcessContext> &contex
...
     session->read(flowFile, &callback);
+
+    bool dropEmptyFlowFiles;
+    if (!callback.called_ && context->getProperty(DropEmptyFlowFiles.getName(), dropEmptyFlowFiles) && !dropEmptyFlowFiles) {
+      // workaround: call callback since ProcessSession doesn't do so for empty flow files
+      logger_->log_debug("ReadCallback workaround on empty flow file, because DropEmptyFlowFiles is false, uuid: %s", flowFile->getUUIDStr());
+      callback.process(nullptr);
+    }
+    if (!callback.called_ && dropEmptyFlowFiles) {
+      logger_->log_info("Deprecated behavior, use connections to drop empty flow files! Dropped empty flow file with uuid: %s", flowFile->getUUIDStr());
```
arpadboda: I like this log message, but I would prefer moving this to connection to avoid duplicating to multiple processors. 
Please also change the log level to debug (it's for developers, not operations)

extensions/librdkafka/PublishKafka.cpp
```diff
@@ -99,6 +99,14 @@ core::Property PublishKafka::MessageKeyField("Message Key Field", "The name of a
...
 core::Property PublishKafka::DebugContexts("Debug contexts", "A comma-separated list of debug contexts to enable."
                                            "Including: generic, broker, topic, metadata, feature, queue, msg, protocol, cgrp, security, fetch, interceptor, plugin, consumer, admin, eos, all", "");
+const core::Property PublishKafka::DropEmptyFlowFiles(
+    core::PropertyBuilder::createProperty("Drop empty flow files")
+        ->withDescription("Keep backwards compatibility with <=0.7.0 bug which caused flow files with empty content to not be published to Kafka. The old behavior is deprecated. Use connections "
+                          "to drop empty flow files!")
+        ->isRequired(false)
+        ->withDefaultValue<bool>(true)
+        ->build()
+);
```
arpadboda: Just nitpicking, but it's usually at the end of the last call (usually "build")

extensions/librdkafka/PublishKafka.h
```diff
@@ -193,121 +186,150 @@ class PublishKafka : public core::Processor {
...
-            hdrs = rd_kafka_headers_new(8);
-          }
-          err = rd_kafka_header_add(hdrs, kv.first.c_str(), kv.first.size(), kv.second.c_str(), kv.second.size());
-        }
-      }
+      called_ = true;
 
       size_t segment_num = 0U;
+
+      assert(flow_size_ == 0 || max_seg_size_ != 0 && "at this point, max_seg_size_ is only zero if flow_size_ is zero");
```
arpadboda: Assert doesn't really compile to anything (as we mainly compile release with debug symbols), so this doesn't really add value.
A logging error here might help more.

extensions/librdkafka/PublishKafka.cpp
```diff
@@ -639,6 +647,17 @@ void PublishKafka::onTrigger(const std::shared_ptr<core::ProcessContext> &contex
     PublishKafka::ReadCallback callback(max_flow_seg_size, kafkaKey, thisTopic->getTopic(), conn->getConnection(), flowFile,
                                         attributeNameRegex, messages, flow_file_index);
     session->read(flowFile, &callback);
+
+    bool dropEmptyFlowFiles;
+    if (!callback.called_ && context->getProperty(DropEmptyFlowFiles.getName(), dropEmptyFlowFiles) && !dropEmptyFlowFiles) {
```
arpadboda: How does it solve the original issue?
In case session->read fails due to missing content claim, it swallows exception, so the execution continues but the callback is never called? 

What's the motivation of moving the kafka related logic to the callback? 

I also have a feeling that this solution won't work in the there IS a content claim, but the content is actually empty. 
Because of this I don't think that moving the logic to the callback is a good idea.

extensions/librdkafka/PublishKafka.cpp
szaszm: We can not move this to Connection, since the data loss happens in the processors, and no Connections have any way of knowing whether the empty flow file was processed or not.

extensions/librdkafka/PublishKafka.cpp
szaszm: What is at the end of the last call? Do you mean the closing paren and semicolon?

extensions/librdkafka/PublishKafka.h
szaszm: It's meant as a comment to the developer, not for condition checking. The logging line would be dead code if I changed to that.

extensions/librdkafka/PublishKafka.cpp
arpadboda: Yep

extensions/librdkafka/PublishKafka.cpp
arpadboda: Fair point :)

extensions/librdkafka/PublishKafka.h
arpadboda: Same still applies: I hardly compile code in debug mode, so I don't think this assertion would even warn me even if I touch the processor code.

extensions/librdkafka/PublishKafka.cpp
szaszm: Producing the kafka message was part of the callback even before my changes. The issue is that when there is no `ResourceClaim`, `ProcessSession` doesn't call the callback and therefore we have no way of sending the message.

I have extended the callback to handle empty flow files and made sure it's called with the above lines. The callback will be called exactly when it was not called and the used opted in for the correct behavior by setting `DropEmptyFlowFiles` to false.

szaszm: I think the callback is the right place for producing the kafka message, since this is normally done in response to a successful read operation, as the next step of the pipeline.
I'm not sure that the workaround belongs to the callback, but since there's a lot of shared code, I think it's fine.

szaszm: Indeed, the behavior is inconsistent. When DropEmptyFlowFiles is true and there is a content claim, we don't drop the empty flow file. Will fix this.

extensions/librdkafka/PublishKafka.cpp
szaszm: fixed in c872f3d

extensions/librdkafka/PublishKafka.h
szaszm: fixed in c872f3d

extensions/librdkafka/PublishKafka.cpp
szaszm: fixed in c872f3d

extensions/librdkafka/PublishKafka.cpp
```diff
@@ -99,6 +99,13 @@ core::Property PublishKafka::MessageKeyField("Message Key Field", "The name of a
                                              "");
 core::Property PublishKafka::DebugContexts("Debug contexts", "A comma-separated list of debug contexts to enable."
                                            "Including: generic, broker, topic, metadata, feature, queue, msg, protocol, cgrp, security, fetch, interceptor, plugin, consumer, admin, eos, all", "");
+const core::Property PublishKafka::FailEmptyFlowFiles(
+    core::PropertyBuilder::createProperty("Fail empty flow files")
+        ->withDescription("Keep backwards compatibility with <=0.7.0 bug which caused flow files with empty content to not be published to Kafka and forwarded to failure. The old behavior is "
+                          "deprecated. Use connections to drop empty flow files!")
+        ->isRequired(false)
```
arpadboda: I think it should be required.

extensions/librdkafka/PublishKafka.h
```diff
@@ -193,121 +186,146 @@ class PublishKafka : public core::Processor {
...
-          err = rd_kafka_header_add(hdrs, kv.first.c_str(), kv.first.size(), kv.second.c_str(), kv.second.size());
-        }
+      assert(max_seg_size_ != 0 || flow_size_ == 0 && "max_seg_size_ == 0 implies flow_size_ == 0");
+      // ^^ therefore checking max_seg_size_ == 0 handles both division by zero and flow_size_ == 0 cases
+      const size_t reserved_msg_capacity = max_seg_size_ == 0 ? 1 : utils::intdiv_ceil(flow_size_, max_seg_size_);
+      messages_->modifyResult(flow_file_index_, [reserved_msg_capacity](FlowFileResult& flow_file) {
+        flow_file.messages.reserve(reserved_msg_capacity);
+      });
+
+      // If the flow file is empty, we still want to send the message, unless the user wants to fail_empty_flow_files_
```
arpadboda: The default value is the property is now true, which means in this case we drop the files. 
This is not a backward compatible behaviour.

szaszm: It is backwards compatible, but not desired. That behavior is why the issue exists.

extensions/librdkafka/PublishKafka.cpp
szaszm: If we make it required, old flows missing the property will fail, so we will not be backwards compatible.

extensions/librdkafka/PublishKafka.h
szaszm: see these 3 lines for proof: https://github.com/apache/nifi-minifi-cpp/pull/712/files#diff-9917deef91cc1503aff82b472310b7eaL662

arpadboda: Ah, okay, thanks!

extensions/librdkafka/PublishKafka.cpp
arpadboda: Fair point.
Having a default value while the property is not required is always a bad smell for me. In this case it's justified.

arpadboda:
LGTM, thanks!

extensions/librdkafka/PublishKafka.cpp
szaszm: Does a default value satisfy the requirement for the property? If so, then I can set `isRequired(true)` and it would still keep old flows working

arpadboda: In our perspective, it does, as whenever we get the value, there is a valid value. 

Required would force the user to specify something, in this case it doesn't apply, so it ensures backward compatibility as you highlighted.

libminifi/include/utils/GeneralUtils.h
```diff
@@ -0,0 +1,50 @@
...
+  return std::unique_ptr<T>{ new T{ std::forward<Args>(args)... } };
+}
+
+template<typename T, typename std::enable_if<std::is_integral<T>::value>::type* = nullptr>
+T intdiv_ceil(T numerator, T denominator) {
+  // note: division and remainder is 1 instruction on x86
+  return numerator / denominator + (numerator % denominator > 0);
+}
+
+template <typename T, typename std::enable_if<std::is_pointer<T>::value>::type* = nullptr>
```
bakaid: This construct is good for marking that the owner of a raw pointer has the responsibility of deallocating it, but I would argue, that aside from interfacing with C code, this is the only case a raw pointer should be used in C++ anyway.
I think in the case of `rd_kafka_headers_s` creating a RAII wrapper for it would be a better solution.

extensions/librdkafka/PublishKafka.h
```diff
@@ -193,121 +186,146 @@ class PublishKafka : public core::Processor {
...
+          read_size_(0),
+          attributeNameRegex_(attributeNameRegex),
+          fail_empty_flow_files_(fail_empty_flow_files)
+    { }
 
     ~ReadCallback() {
-      if (hdrs) {
-        rd_kafka_headers_destroy(hdrs);
-      }
+      rd_kafka_headers_destroy(hdrs);
```
bakaid: I am not a big fan of unnecessarily checking for nullptr in cases where it really does not make sense, like before `free`.
In this case, however `rd_kafka_headers_destroy` will segfault if it gets a nullptr: https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_header.c#L36, and generically, I like more defensiveness around third party APIs.
Even if at the current state of the code we can be sure that hdrs won't be nullptr, a later refactor might broke that implicit assumption and cause an issue here.

libminifi/include/utils/GeneralUtils.h
szaszm: `std::unique_ptr<rd_kafka_headers_t, decltype([](rd_kafka_headers_t* p) { rd_kafka_headers_destroy(p); })>` both acts as an RAII wrapper and does the null checking.

p.s.: can't wait for C++20

szaszm:
> Will run tests.

@bakaid Did you run the tests?

bakaid:
@szaszm Unfortunately I was always preempted by something, but I am building your final version right now.

bakaid:
@szaszm moving the make_headers call to the ReadCallback constructor body solves the issue.

extensions/librdkafka/PublishKafka.h
```diff
@@ -194,120 +187,142 @@ class PublishKafka : public core::Processor {
...
+          flow_file.messages.resize(segment_num + 1);
+        }
+      });
+    }
+
+    static rd_kafka_headers_unique_ptr make_headers(const core::FlowFile& flow_file, utils::Regex& attribute_name_regex) {
+      const utils::owner<rd_kafka_headers_t*> result{ rd_kafka_headers_new(8) };
+      if (!result) { throw std::bad_alloc{}; }
+
+      for (const auto& kv : flow_file.getAttributes()) {
```
bakaid: Compilation fails with
```
/Users/danielbakai/nifi-minifi-cpp/extensions/librdkafka/PublishKafka.h:212:29: error: 'this' argument to
      member function 'getAttributes' has type 'const core::FlowFile', but function is not marked const
      for (const auto& kv : flow_file.getAttributes()) {
```

szaszm: rebased on top of dfbe225, That should solve the issue

arpadboda:
Merged as 455d4b35bb6bc70373372d66ceaf9d35a4d32a49
Forgot to add close with commit, closing manually.

szaszm:
Both appveyor test failures are unrelated:
* ProcessorTests: Fails because it tests that a file with a name starting with a dot is treated as a hidden file and ignored by GetFile
* TailFileTests: Fails because test files are written with CRLF line ending but TailFile expects LF.
Both are fixed in MINIFICPP-1096 (PR coming soon)

PR #715, opened 2020-01-27
MINIFICPP-1126 - Reduce sawtooth in memory usage of rocksdb flowfile …

arpadboda:
…repo

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/rocksdb-repos/FlowFileRepository.cpp
+++ b/extensions/rocksdb-repos/FlowFileRepository.cpp
@@ -15,17 +15,20 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#include "FlowFileRecord.h"
 #include "FlowFileRepository.h"
+#include "utils/ScopeGuard.h"
+
 #include "rocksdb/options.h"
 #include "rocksdb/write_batch.h"
 #include "rocksdb/slice.h"
+
+#include <chrono>
 #include <memory>
 #include <string>
 #include <utility>
 #include <vector>
 #include <list>
-#include "FlowFileRecord.h"
-#include "utils/ScopeGuard.h"
 
 namespace org {
 namespace apache {
@@ -103,13 +106,18 @@ void FlowFileRepository::printStats() {
 }
 
 void FlowFileRepository::run() {
+  auto last = std::chrono::system_clock::now();
   if (running_) {
     prune_stored_flowfiles();
   }
   while (running_) {
     std::this_thread::sleep_for(std::chrono::milliseconds(purge_period_));
     flush();
-    printStats();
+    auto now = std::chrono::system_clock::now();
+    if ((now-last) > std::chrono::seconds(30)) {
+      printStats();
+      last = now;
+    }
   }
 }
 
--- a/extensions/rocksdb-repos/FlowFileRepository.cpp
+++ b/extensions/rocksdb-repos/FlowFileRepository.cpp
@@ -16,6 +16,7 @@
  * limitations under the License.
  */
 #include "FlowFileRepository.h"
+#include "rocksdb/options.h"
 #include "rocksdb/write_batch.h"
 #include "rocksdb/slice.h"
 #include <memory>
@@ -35,7 +36,6 @@ namespace repository {
 
 void FlowFileRepository::flush() {
   rocksdb::WriteBatch batch;
-  uint64_t decrement_total = 0;
   rocksdb::ReadOptions options;
 
   std::vector<std::shared_ptr<FlowFileRecord>> purgeList;
@@ -61,7 +61,6 @@ void FlowFileRepository::flush() {
       continue;
     }
 
-    decrement_total += values[i].size();
     std::shared_ptr<FlowFileRecord> eventRead = std::make_shared<FlowFileRecord>(shared_from_this(), content_repo_);
     if (eventRead->DeSerialize(reinterpret_cast<const uint8_t *>(values[i].data()), values[i].size())) {
       purgeList.push_back(eventRead);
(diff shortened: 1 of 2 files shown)


extensions/rocksdb-repos/FlowFileRepository.h
```diff
@@ -103,6 +105,9 @@ class FlowFileRepository : public core::Repository, public std::enable_shared_fr
     options.create_if_missing = true;
     options.use_direct_io_for_flush_and_compaction = true;
     options.use_direct_reads = true;
+    options.write_buffer_size = 8 << 20;
```
arpadboda: This is the important part of this PR.

When operations are done in rocksdb, it stores them in an unsorted list of events. The buffer getting full means that these events should be merged and serialized, so records are written to the disk in the regular structure of rocksdb. 

In our case it means two things:
-During our regular usecase the content of the buffer in continuously growing as creation and later deletion of the same element results in two events added to the buffer (log). 
-When the buffer is full and events are merged, there is a CPU spike and the result (nothing or nearly nothing in our case) is written to the underlying storage. 

As this buffer only contains flowfile metadata (attributes and content location), it's filled quite slowly. It takes hours in case 10 flowfiles are generated / sec. 

To avoid memory usage going that high (FF repo can consume more than the rest of whole MiNiFI when the buffer is close to full) and have smaller CPU peaks the buffer is now emptied more frequently. This still means minutes with a decent load. 

For more information check this: https://github.com/facebook/rocksdb/wiki/Memory-usage-in-RocksDB#memtable
(Note: write amplification doesn't matter in our case as we usually persist negligible amount of data as flowfiles fade out of the system in a very short time)

The logging added in this PR logs the current write buffer usage as well. It helps monitoring the peaks.

extensions/rocksdb-repos/FlowFileRepository.cpp
```diff
@@ -69,14 +68,8 @@ void FlowFileRepository::flush() {
...
-
-  if (db_->Write(rocksdb::WriteOptions(), &batch).ok()) {
-    logger_->log_trace("Decrementing %u from a repo size of %u", decrement_total, repo_size_.load());
-    if (decrement_total > repo_size_.load()) {
-      repo_size_ = 0;
-    } else {
-      repo_size_ -= decrement_total;
-    }
+  if (!db_->Write(rocksdb::WriteOptions(), &batch).ok()) {
+    logger_->log_warn("Failed to execute batch operation when flushing FlowFileRepository");
```
arpadboda: I removed size handling as:
- It was done in a inconsistent way.
- Even if the repo_full flag was set, nothing happened, new elements were inserted. 
- Handling repo size out of the repo (without taking the size of keys into account) is most probably a bad design.

extensions/rocksdb-repos/FlowFileRepository.cpp
```diff
@@ -89,23 +82,28 @@ void FlowFileRepository::flush() {
   }
 }
 
-void FlowFileRepository::run() {
-  // threshold for purge
+void FlowFileRepository::printStats() {
```
bakaid: If I understand correctly, this is ran every purge_period, which is, by default, 2.5 s.
4 log lines every 2.5 seconds is way too much noise, affecting everyone, as virtually everyone uses this FlowFileRepo and the default log level is info. This would account for the majority of logs in many cases.
I think this should definitely be at debug level.

extensions/rocksdb-repos/FlowFileRepository.h
bakaid: I would like to see more in-code comments about the rationale behind these values.
Why 8 MB? Why do we set max_write_buffer_number to 4? As far as I understand that means it can keep 4 of these memtables in memory, and the default is 2.

extensions/rocksdb-repos/FlowFileRepository.cpp
bakaid: Agreed, but I think it is something we would need, done properly. Is there a follow-up issues for it?

extensions/rocksdb-repos/FlowFileRepository.cpp
arpadboda: Fair point.

I would keep it at info level, but call less frequently.

extensions/rocksdb-repos/FlowFileRepository.cpp
arpadboda: https://issues.apache.org/jira/browse/MINIFICPP-1129

extensions/rocksdb-repos/FlowFileRepository.h
```diff
@@ -105,8 +105,15 @@ class FlowFileRepository : public core::Repository, public std::enable_shared_fr
     options.create_if_missing = true;
     options.use_direct_io_for_flush_and_compaction = true;
     options.use_direct_reads = true;
+
+    // Write buffers are used as db oepration logs. When they get filled the events are merged and serialized.
```
bakaid: typo: oepration -> operation

extensions/rocksdb-repos/FlowFileRepository.h
```diff
@@ -105,8 +105,15 @@ class FlowFileRepository : public core::Repository, public std::enable_shared_fr
...
+    // The default size is 64MB.
+    // In our case it's usually too much, causing sawtooth in memory consumption. (Consumes more than the whole MiniFi)
+    // To avoid DB write issues during heavy load it's recommended to have high number of buffer.
+    // Rocksdb's stall featur can also trigger in case the number of buffers is >= 3.
```
bakaid: typo: featur -> feature

extensions/rocksdb-repos/FlowFileRepository.cpp
szaszm: I'd also suggest logging the stats in a single line. It reduces log spam while keeping usefulness.

arpadboda: Now it's single line.

extensions/rocksdb-repos/FlowFileRepository.h
arpadboda: Added a long comment to explain the decision behind.

arpadboda:
> Did anything significant change since my last approve?

Just rebased to master to resolve conflict.

extensions/rocksdb-repos/FlowFileRepository.cpp
```diff
@@ -97,23 +91,33 @@ void FlowFileRepository::flush() {
...
+
+  std::string all_memtables;
+  db_->GetProperty("rocksdb.cur-size-all-mem-tables", &all_memtables);
+
+  logger_->log_info("Repository stats: key count: %zu, table readers size: %zu, all memory tables size: %zu",
+      key_count, table_readers, all_memtables);
+}
 
+void FlowFileRepository::run() {
+  auto last = std::chrono::system_clock::now();
```
bakaid: Please use `std::chrono::steady_clock`, system_clock is not monotonic and not suitable for measuring intervals.

arpadboda: Done.

PR #718, opened 2020-01-28
MINIFICPP-1132 - CWEL should stop committing further flowfiles in cas…

arpadboda:
…e outgoing connection is full

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/standard-processors/tests/unit/ProcessorTests.cpp
+++ b/extensions/standard-processors/tests/unit/ProcessorTests.cpp
@@ -21,6 +21,7 @@
 #include <vector>
 #include <set>
 #include <fstream>
+#include <GenerateFlowFile.h>
 
 #include "TestBase.h"
 #include "LogAttribute.h"
@@ -129,6 +130,7 @@ TEST_CASE("Test GetFile Ignore", "[getfileCreate3]") {
   LogTestController::getInstance().setDebug<minifi::processors::GetFile>();
   std::shared_ptr<core::ContentRepository> content_repo = std::make_shared<core::repository::VolatileContentRepository>();
   std::shared_ptr<core::Processor> processor = std::make_shared<org::apache::nifi::minifi::processors::GetFile>("getfileCreate2");
+  processor->initialize();
 
   std::shared_ptr<core::Repository> test_repo = std::make_shared<TestRepository>();
   std::shared_ptr<TestRepository> repo = std::static_pointer_cast<TestRepository>(test_repo);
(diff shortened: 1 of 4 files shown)


extensions/standard-processors/tests/unit/ProcessorTests.cpp
```diff
@@ -207,6 +209,60 @@ TEST_CASE("Test GetFile Ignore", "[getfileCreate3]") {
   prev++;
 }
 
+TEST_CASE("TestConnectionFull", "[ConnectionFull]") {
+  TestController testController;
+  LogTestController::getInstance().setDebug<minifi::processors::GenerateFlowFile>();
+  std::shared_ptr<core::ContentRepository> content_repo = std::make_shared<core::repository::VolatileContentRepository>();
+  std::shared_ptr<core::Processor> processor = std::make_shared<org::apache::nifi::minifi::processors::GenerateFlowFile>("GFF");
```
szaszm: Use `auto` to reduce type name clutter when it's obvious from the context, like in the case of `make_shared`, `make_unique` and iterators.

libminifi/src/core/ProcessSession.cpp
```diff
@@ -965,6 +965,18 @@ std::shared_ptr<core::FlowFile> ProcessSession::get() {
...
+bool ProcessSession::outgoingConnectionsFull(const std::string& relationship) {
+  std::set<std::shared_ptr<Connectable>> connections = process_context_->getProcessorNode()->getOutGoingConnections(relationship);
+  Connection * connection = nullptr;
+  for (const auto& conn : connections) {
+    connection = dynamic_cast<Connection*>(conn.get());
+    if (connection && connection->isFull()) {
+      return true;
+    }
+  }
+  return false;
```
szaszm: I prefer to use STL algorithms when possible.
```
return std::find_if(std::cbegin(connections), std::cend(connections), [](const auto& conn) {
    auto* connection = dynamic_cast<Connection*>(conn.get());
    return connection && connection->isFull();
}) != std::cend(connections);
```

arpadboda: I also like std algorithms, but it in this I don't think it makes the code any more readable.

extensions/standard-processors/tests/unit/ProcessorTests.cpp
arpadboda: As you can see both shared ptrs are base class type, so using auto wouldn't be equivalent.

szaszm: True, nevermind then

libminifi/src/core/ProcessSession.cpp
szaszm: Maybe `std::any_of` would be better, I've just learned about it. I still prefer anything higher level than a raw loop.
Also see: C++ seasoning at ~4:30 https://www.youtube.com/watch?v=qH6sSOr-yk8

I've already approved the PR since this is a minor detail, so feel free to merge it either way.

libminifi/src/core/ProcessSession.cpp
```diff
@@ -965,6 +965,18 @@ std::shared_ptr<core::FlowFile> ProcessSession::get() {
   return NULL;
 }
 
+bool ProcessSession::outgoingConnectionsFull(const std::string& relationship) {
+  std::set<std::shared_ptr<Connectable>> connections = process_context_->getProcessorNode()->getOutGoingConnections(relationship);
+  Connection * connection = nullptr;
+  for (const auto& conn : connections) {
+    connection = dynamic_cast<Connection*>(conn.get());
```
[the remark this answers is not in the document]
arpadboda: Processors for eg.

PR #719, opened 2020-01-29
MINIFICPP-1133 - Rocksdb operation errors should cause exceptions ins…

arpadboda:
…tead of being silently ignored

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/rocksdb-repos/FlowFileRepository.cpp
+++ b/extensions/rocksdb-repos/FlowFileRepository.cpp
@@ -83,8 +83,8 @@ void FlowFileRepository::flush() {
   } else {
     for (const auto& key: keystrings) {
       keys_to_delete.enqueue(key);  // Push back the values that we could get but couldn't delete
-      return;  // Stop here - don't delete from content repo while we have records in FF repo
     }
+    return;  // Stop here - don't delete from content repo while we have records in FF repo
   }
 
   if (nullptr != content_repo_) {
--- a/extensions/rocksdb-repos/FlowFileRepository.cpp
+++ b/extensions/rocksdb-repos/FlowFileRepository.cpp
@@ -57,6 +57,7 @@ void FlowFileRepository::flush() {
   for(size_t i=0; i<keys.size() && i<values.size() && i<multistatus.size(); ++i) {
     if(!multistatus[i].ok()) {
       logger_->log_error("Failed to read key from rocksdb: %s! DB is most probably in an inconsistent state!", keys[i].data());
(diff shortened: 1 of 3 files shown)


extensions/rocksdb-repos/FlowFileRepository.cpp
```diff
@@ -158,7 +166,21 @@ void FlowFileRepository::prune_stored_flowfiles() {
...
+  int waitTime = FLOWFILE_REPOSITORY_RETRY_INTERVAL;
+  for (int i=0; i<3; ++i) {
+    auto status = operation();
+    if (status.ok()) {
+      logger_->log_trace("Rocksdb operation executed successfully");
+      return true;
+    }
+    logger_->log_error("Rocksdb operation failed: %s", status.ToString());
+    std::this_thread::sleep_for(std::chrono::milliseconds(waitTime));
+    waitTime += FLOWFILE_REPOSITORY_RETRY_INTERVAL;
```
szaszm: The suffix `RETRY_INTERVAL` suggests that we retry every so often but here we increase the delay between retries in each iteration.

extensions/rocksdb-repos/FlowFileRepository.h
```diff
@@ -168,6 +172,8 @@ class FlowFileRepository : public core::Repository, public std::enable_shared_fr
 
  private:
 
+  virtual bool ExecuteWithRetry(std::function<rocksdb::Status()> operation);
```
szaszm: This is just a local helper function, it doesn't need to be `virtual`. Am I missing something?

arpadboda: Fair, as it's private, it add no value.

extensions/rocksdb-repos/FlowFileRepository.cpp
arpadboda: Agreed, I'm open for suggestion :)

szaszm: Either change the behavior to always wait `FLOWFILE_REPOSITORY_RETRY_INTERVAL` milliseconds instead of increasing the wait time, or change the name to something like `FLOWFILE_REPOSITORY_RETRY_INTERVAL_INCREMENTS`. I don't like the name I've suggested, but a bad name is better than a misleading one.

arpadboda: Done.

PR #721, opened 2020-01-29
MINIFICPP-1135 - Add a watchdog to schedulingAgent to warn in case of…

arpadboda:
… long-lasting onTrigger calls

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/src/SchedulingAgent.cpp
```diff
@@ -125,6 +138,17 @@ bool SchedulingAgent::onTrigger(const std::shared_ptr<core::Processor> &processo
   return false;
 }
 
+void SchedulingAgent::watchDogFunc() {
+  std::lock_guard<std::mutex> lock(watchdog_mtx_);
+  auto now = std::chrono::system_clock::now();
```
bakaid: std::chrono::system_clock is not guaranteed to be monotonic, and it is indeed not in DST situations, manually setting the time or when using NTP. This can cause false positives and false negatives.
I think std::chrono::steady_clock should be used here.

libminifi/src/SchedulingAgent.cpp
```diff
@@ -125,6 +138,17 @@ bool SchedulingAgent::onTrigger(const std::shared_ptr<core::Processor> &processo
...
+  for (const auto& info : scheduled_processors_) {
+    auto elapsed = std::chrono::duration_cast<std::chrono::milliseconds>(now - info.start_time_).count();
+    if (elapsed > SCHEDULING_WATCHDOG_ALERT_PERIOD) {
+      logger_->log_warn("%s::onTrigger is running for %zu ms in %s", info.name_, elapsed, info.uuid_);
```
bakaid: `std::chrono::duration::count()` has a return type of `Rep`, which in the case of `std::chrono::milliseconds` is `/*signed integer type of at least 45 bits*/` (https://en.cppreference.com/w/cpp/chrono/duration).
So we can safely assume it's an `int64_t`, but surely not a `size_t`, that the `%zu` would indicate.
I suggest making `elapsed` an `int64_t` and using `%lld` in the format string (`PRId64` would be the completely correct solution, but we don't use these anywhere, that's the scope of a complete refactor).

libminifi/include/SchedulingAgent.h
```diff
@@ -190,8 +201,39 @@ class SchedulingAgent {
...
+    std::string name_;
+    std::string uuid_;
+
+    SchedulingInfo(const std::shared_ptr<core::Processor> &processor) {
+     start_time_ = std::chrono::system_clock::now();
+     name_ = processor->getName();
+     uuid_ = processor->getUUIDStr();
+    }
+
+    bool operator <(const SchedulingInfo& o) const {
```
bakaid:
```
bool operator <(const SchedulingInfo& o) const {
    return std::tie(start_time_, name_, uuid_) < std::tie(o.start_time_, o.name_, o.uuid_);
}
```

arpadboda: Nice cut!

libminifi/src/SchedulingAgent.cpp
arpadboda: I don't think that any ontrigger call could last more than pow(2,31) millisecs, (that would be at about 23 days), although I agree with the approach :)

libminifi/src/SchedulingAgent.cpp
arpadboda: Done

libminifi/src/SchedulingAgent.cpp
bakaid: It's not mainly about representability, but type correctness. With printf it's usually not such a big issue, with scanf it is the source of many security issues, so I like to keep format strings correct.
Also, because the original code cast a signed int to an unsigned one, and because of its usage of non-monotonic clock which could really result in the difference being negative, we would have logged pretty interesting values.

libminifi/include/SchedulingAgent.h
```diff
@@ -190,8 +201,28 @@ class SchedulingAgent {
   std::shared_ptr<core::controller::ControllerServiceProvider> controller_service_provider_;
 
  private:
+  struct SchedulingInfo {
+    std::chrono::time_point<std::chrono::steady_clock> start_time_;
+    std::string name_;
+    std::string uuid_;
+
+    SchedulingInfo(const std::shared_ptr<core::Processor> &processor) {
```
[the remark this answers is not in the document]
arpadboda: Explicit is fine as emplace calls explicits as well, so change it.

libminifi/include/SchedulingAgent.h
```diff
@@ -190,8 +201,28 @@ class SchedulingAgent {
...
+     start_time_ = std::chrono::steady_clock::now();
+     name_ = processor->getName();
+     uuid_ = processor->getUUIDStr();
```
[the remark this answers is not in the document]
arpadboda: Done

libminifi/src/SchedulingAgent.cpp
```diff
@@ -125,6 +138,17 @@ bool SchedulingAgent::onTrigger(const std::shared_ptr<core::Processor> &processo
   return false;
 }
 
+void SchedulingAgent::watchDogFunc() {
+  std::lock_guard<std::mutex> lock(watchdog_mtx_);
+  auto now = std::chrono::steady_clock::now();
+  for (const auto& info : scheduled_processors_) {
+    int64_t elapsed = std::chrono::duration_cast<std::chrono::milliseconds>(now - info.start_time_).count();
+    if (elapsed > SCHEDULING_WATCHDOG_ALERT_PERIOD) {
+      logger_->log_warn("%s::onTrigger is running for %lld ms in %s", info.name_, elapsed, info.uuid_);
```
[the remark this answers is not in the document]
arpadboda: Done

libminifi/include/SchedulingAgent.h
```diff
@@ -190,8 +201,28 @@ class SchedulingAgent {
...
+
+    bool operator <(const SchedulingInfo& o) const {
+      return std::tie(start_time_, name_, uuid_) < std::tie(o.start_time_, o.name_, o.uuid_);
+    }
+  };
+
   // Logger
   std::shared_ptr<logging::Logger> logger_;
+  std::mutex watchdog_mtx_;  // used to protect the vector below
+  std::set<SchedulingInfo> scheduled_processors_;
```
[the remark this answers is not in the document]
arpadboda: Fair point, added a comment.

libminifi/include/SchedulingAgent.h
```diff
@@ -131,11 +136,16 @@ class SchedulingAgent {
...
     thread_pool_.start();
+
+    std::function<void(void)> f = std::bind(&SchedulingAgent::watchDogFunc, this);
+    watchDogTimer_.reset(new utils::CallBackTimer(std::chrono::milliseconds(SCHEDULING_WATCHDOG_CHECK_PERIOD), f));
+    watchDogTimer_->start();
   }
   // Destructor
   virtual ~SchedulingAgent() {
-
+    watchDogTimer_->stop();
```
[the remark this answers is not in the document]
arpadboda: Removed the whole  dtor.

bakaid:
@arpadboda Ran tests, functionality seems OK.
I have concerns about the hardcoded value for the warning time.
It may be perfectly valid to have onTriggers running for more than 5 seconds, especially if there are periodic commits. The general case is that it is probably not good to have an onTrigger running for more than 5 seconds, so it is a good thing to warn developers and flow creators about it, but in case it is an accepted thing, I definitely don't want a warn level log about it every second.
I propose adding a configuration property to the effect of `nifi.scheduler.ontrigger.warning.time` with a default value of `5 sec` and using that instead of the hardcoded value.

libminifi/include/SchedulingAgent.h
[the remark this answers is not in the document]
szaszm: By removing the dtor, you've made it non-virtual, which is also problematic because there are derived classes from `SchedulingAgent`.
I'd suggest `virtual ~SchedulingAgent() = default;`, possibly `noexcept` if all the member and derived class dtors are also `noexcept`.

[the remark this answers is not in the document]
arpadboda: I had to put the stop call back to the dtor, added a comment there to make sure.

libminifi/include/SchedulingAgent.h
```diff
@@ -190,11 +207,27 @@ class SchedulingAgent {
   std::shared_ptr<core::controller::ControllerServiceProvider> controller_service_provider_;
 
  private:
+  struct SchedulingInfo {
+    std::chrono::time_point<std::chrono::steady_clock> start_time_ = std::chrono::steady_clock::now();
+    mutable std::chrono::time_point<std::chrono::steady_clock> last_alert_time_ = std::chrono::steady_clock::now();
```
szaszm: Don't lie on the interface about `const`-ness, unless dealing with mutexes, caches or similar.

libminifi/include/SchedulingAgent.h
```diff
@@ -190,11 +207,27 @@ class SchedulingAgent {
...
+
   // Logger
   std::shared_ptr<logging::Logger> logger_;
-  // Prevent default copy constructor and assignment operation
-  // Only support pass by reference or pointer
-
+  mutable std::mutex watchdog_mtx_;  // used to protect the set below
+  std::set<SchedulingInfo> scheduled_processors_;  // set was chosen to avoid iterator invalidation
+  std::unique_ptr<utils::CallBackTimer> watchDogTimer_;
+  int64_t alert_time_;  // msec
```
szaszm: unit in the type > unit in the name > unit in comment > no indication of unit in code
I suggest changing the type to `std::chrono::milliseconds` and only converting to integral when we're passing the value through an incompatible API. (Like the logging in `watchDogFunc`)

libminifi/src/SchedulingAgent.cpp
```diff
@@ -125,6 +138,19 @@ bool SchedulingAgent::onTrigger(const std::shared_ptr<core::Processor> &processo
...
+void SchedulingAgent::watchDogFunc() const {
+  std::lock_guard<std::mutex> lock(watchdog_mtx_);
+  auto now = std::chrono::steady_clock::now();
+  for (auto& info : scheduled_processors_) {
+    int64_t elapsed = std::chrono::duration_cast<std::chrono::milliseconds>(now - info.last_alert_time_).count();
+    if (elapsed > alert_time_) {
+      logger_->log_warn("%s::onTrigger has been running for %lld ms in %s", info.name_,
+          std::chrono::duration_cast<std::chrono::milliseconds>(now - info.start_time_).count(), info.uuid_);
+      info.last_alert_time_ = now;
+    }
```
szaszm: To keep the unit in the type for as long as possible, I suggest this:
```
auto elapsed = now - info.last_alert_time;
if (elapsed > alert_time_) {
  int64_t elapsed_ms{ std::chrono::duration_cast<std::chrono::milliseconds>(now - info.start_time_).count() };
  logger_->log_warn("%s::onTrigger has been running for %lld ms in %s", info.name_, elapsed_ms, info.uuid_);
  info.last_alert_time_ = now;
}
```
Using braced initializer for `elapsed` prevents narrowing implicit conversion.

libminifi/include/SchedulingAgent.h
```diff
@@ -39,6 +41,9 @@
 #include "core/controller/ControllerServiceProvider.h"
 #include "core/controller/ControllerServiceNode.h"
 
+#define SCHEDULING_WATCHDOG_CHECK_PERIOD 1000  // msec
+#define SCHEDULING_WATCHDOG_ALERT_PERIOD 5000  // msec
```
szaszm: `SCHEDULING_WATCHDOG_ALERT_PERIOD` is now redundant with the introduction of `nifi.flow.engine.alert.time`

libminifi/include/SchedulingAgent.h
arpadboda: I don't think it's a lie. 
It think it depends on preference, in my opinion marking some helper members mutable and maintaining const correctness is better than having a lot of non-const functions that look like they should have been.

libminifi/src/SchedulingAgent.cpp
arpadboda: This doesn't compile:
```
/Users/aboda/work/shadow/minifi/nifi-minifi-cpp/libminifi/src/SchedulingAgent.cpp:146:17: error: invalid operands to binary expression ('std::__1::chrono::duration<long long, std::__1::ratio<1, 1000000000> >' and 'int64_t'
      (aka 'long long'))
    if (elapsed > alert_time_) {
```

Yep, crono suxx

libminifi/include/SchedulingAgent.h
arpadboda: Done

libminifi/src/SchedulingAgent.cpp
arpadboda: Done

libminifi/include/SchedulingAgent.h
arpadboda: That should be the default value to avoid hardcoded numbers in the code.
Fixed, renamed.

libminifi/include/SchedulingAgent.h
szaszm: Before doing some additional research, my approach was: I should have a very good reason to use `mutable`. The ability to mark `watchDogFunc` as `const` is not a good enough reason IMO, but this requires subjective judgement.

Now I came across a great approach on stackoverflow:
> [mutable] allows the differentiation of bitwise const and logical const. Logical const is when an object doesn't change in a way that is visible through the public interface, like [mutex].
https://stackoverflow.com/a/105061/3997716

The change of `last_alert_time_` is observable from the public interface which makes marking it `mutable` objectively wrong. It can even be changed through the public interface.

arpadboda: Tried to change it, but it cannot be changed as the object is used in a set and only const iteration can be performed on that. Added a comment to indicate the reason of being mutable.

szaszm: ok, thanks

libminifi/include/SchedulingAgent.h
```diff
@@ -128,14 +133,25 @@ class SchedulingAgent {
      * to be able to debug why an agent doesn't work and still allow a restart via updates in these cases.
      */
     auto csThreads = configure_->getInt(Configure::nifi_flow_engine_threads, 2);
+    alert_time_ = std::chrono::milliseconds(configure_->getInt(Configure::nifi_flow_engine_alert_time, SCHEDULING_WATCHDOG_ALERT_DEFAULT_PERIOD));
```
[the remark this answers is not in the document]
arpadboda: Ok, moved.

libminifi/include/SchedulingAgent.h
```diff
@@ -190,11 +207,27 @@ class SchedulingAgent {
...
+
   // Logger
   std::shared_ptr<logging::Logger> logger_;
-  // Prevent default copy constructor and assignment operation
-  // Only support pass by reference or pointer
-
+  mutable std::mutex watchdog_mtx_;  // used to protect the set below
+  std::set<SchedulingInfo> scheduled_processors_;  // set was chosen to avoid iterator invalidation
+  std::unique_ptr<utils::CallBackTimer> watchDogTimer_;
+  std::chrono::milliseconds alert_time_;  // msec
```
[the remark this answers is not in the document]
arpadboda: Renamed both.

libminifi/src/Configure.cpp
```diff
@@ -28,6 +28,7 @@ const char *Configure::nifi_flow_configuration_file = "nifi.flow.configuration.f
 const char *Configure::nifi_flow_configuration_file_exit_failure = "nifi.flow.configuration.file.exit.onfailure";
 const char *Configure::nifi_flow_configuration_file_backup_update = "nifi.flow.configuration.backup.on.update";
 const char *Configure::nifi_flow_engine_threads = "nifi.flow.engine.threads";
+const char *Configure::nifi_flow_engine_alert_time = "nifi.flow.engine.alert.time";
```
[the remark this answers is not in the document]
arpadboda: We could, but none of our current settings contain unit in the name, so I would go for being consistent with what we have.

In the longer term I would like to have a documentation that lists all the config params we have with units, description, etc.

libminifi/include/SchedulingAgent.h
```diff
@@ -39,6 +41,9 @@
 #include "core/controller/ControllerServiceProvider.h"
 #include "core/controller/ControllerServiceNode.h"
 
+#define SCHEDULING_WATCHDOG_CHECK_PERIOD 1000  // msec
+#define SCHEDULING_WATCHDOG_ALERT_DEFAULT_PERIOD 5000  // msec
```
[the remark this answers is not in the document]
arpadboda: Done.

PR #728, opened 2020-02-06
MINIFICPP-1147 Implemented.

am-c-p-p:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/windows-event-log/ConsumeWindowsEventLog.cpp
+++ b/extensions/windows-event-log/ConsumeWindowsEventLog.cpp
@@ -164,6 +164,9 @@ ConsumeWindowsEventLog::ConsumeWindowsEventLog(const std::string& name, utils::I
 }
 
 ConsumeWindowsEventLog::~ConsumeWindowsEventLog() {
+  if (hMsobjsDll_) {
+    FreeLibrary(hMsobjsDll_);
+  }
 }
 
 void ConsumeWindowsEventLog::initialize() {
@@ -226,6 +229,25 @@ void ConsumeWindowsEventLog::onSchedule(const std::shared_ptr<core::ProcessConte
     }
   }
 
+  std::string mode;
+  context->getProperty(OutputFormat.getName(), mode);
+
+  writeXML_ = (mode == Both || mode == XML);
+
+  writePlainText_ = (mode == Both || mode == Plaintext);
+
+  if (writeXML_) {
+    char systemDir[MAX_PATH];
+    if (GetSystemDirectory(systemDir, sizeof(systemDir))) {
+      hMsobjsDll_ = LoadLibrary((systemDir + std::string("\\msobjs.dll")).c_str());
+      if (!hMsobjsDll_) {
+        logger_->log_error("!LoadLibrary error %x", GetLastError());
+      }
+    } else {
+      logger_->log_error("!GetSystemDirectory error %x", GetLastError());
+    }
+  }
+
   if (subscriptionHandle_) {
     logger_->log_error("Processor already subscribed to Event Log, expected cleanup to unsubscribe.");
   } else {
@@ -233,13 +255,6 @@ void ConsumeWindowsEventLog::onSchedule(const std::shared_ptr<core::ProcessConte
 
     subscribe(context);
   }
-
-  std::string mode;
-  context->getProperty(OutputFormat.getName(), mode);
-
-  writeXML_ = (mode == Both || mode == XML);
-
-  writePlainText_ = (mode == Both || mode == Plaintext);
 }
 
 void ConsumeWindowsEventLog::onTrigger(const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSession> &session) {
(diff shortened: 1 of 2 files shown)


extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -286,6 +301,103 @@ wel::WindowsEventLogHandler ConsumeWindowsEventLog::getEventLogHandler(const std
...
+
+  struct TreeWalker: public pugi::xml_tree_walker {
+    TreeWalker(HMODULE hMsobjsDll, std::unordered_map<std::string, std::string>& xmlPercentageItemsResolutions, std::shared_ptr<logging::Logger> logger)
+      : hMsobjsDll_(hMsobjsDll), xmlPercentageItemsResolutions_(xmlPercentageItemsResolutions), logger_(logger) {
+    }
+
+    bool for_each(pugi::xml_node& node) override {
+      const std::string& nodeText = node.text().get();
+
+      auto beginNumberPos = nodeText.find("%%", 0);
```
arpadboda: Multiple problems here:

1. This is true for anything that contains "%%" somewhere in the string
2. the "isdigit(c)" only breaks, but doesn't result in error

Becase of the two issues above the following string: "abc %%1 efg" is something we are going to consider as an error code and try to look it up. 

I would suggest fixing this way:
```
if (nodeText.rfind("%%", 0) != 0) {
  return true;
} 
std::string numText = nodeText.substr(2);
if (!std::all_of(numText.begin(), numText.end(), ::isdigit) {
  return true;
}
auto n = std::stoul(numText);
```

This guarantees that "%%" can occur only at the beginning of the text and rest of the characters should be digits.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -286,6 +301,103 @@ wel::WindowsEventLogHandler ConsumeWindowsEventLog::getEventLogHandler(const std
...
+            n,
+            MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
+            (LPTSTR)&pBuffer,
+            1024,
+            0
+          )) {
+            value = pBuffer;
+            LocalFree(pBuffer);
+
+            // Remove trailing white spaces.
```
arpadboda:
```
value  = StringUtils::trimRight(value);
```

( https://github.com/apache/nifi-minifi-cpp/blob/master/libminifi/include/utils/StringUtils.h#L114 )

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -226,20 +229,32 @@ void ConsumeWindowsEventLog::onSchedule(const std::shared_ptr<core::ProcessConte
...
+
+  writeXML_ = (mode == Both || mode == XML);
+
+  writePlainText_ = (mode == Both || mode == Plaintext);
+
+  if (writeXML_) {
+    char systemDir[MAX_PATH];
+    if (GetSystemDirectory(systemDir, sizeof(systemDir))) {
+      hMsobjsDll_ = LoadLibrary((systemDir + std::string("\\msobjs.dll")).c_str());
+      if (!hMsobjsDll_) {
```
arpadboda: It's nice to have error handling here, although we should make sure not to call "FormatMessage" later in case this dll open failed.

am-c-p-p: It is checked in the beginning of substituteXMLPercentageItems.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
am-c-p-p: %% might be anywhere in the text, for instance https://docs.microsoft.com/en-us/windows/security/threat-protection/auditing/event-4662 `Properties` in Event XML.

arpadboda: It also begins with that, but ok. 
You can still use stoul to convert, that drops he rest (from the first non-digit char) of the string.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -226,20 +229,32 @@ void ConsumeWindowsEventLog::onSchedule(const std::shared_ptr<core::ProcessConte
...
+  context->getProperty(OutputFormat.getName(), mode);
+
+  writeXML_ = (mode == Both || mode == XML);
+
+  writePlainText_ = (mode == Both || mode == Plaintext);
+
+  if (writeXML_) {
+    char systemDir[MAX_PATH];
+    if (GetSystemDirectory(systemDir, sizeof(systemDir))) {
+      hMsobjsDll_ = LoadLibrary((systemDir + std::string("\\msobjs.dll")).c_str());
```
bakaid: Having this in onSchedule and `FreeLibrary` in the destructor means that if onSchedule is called multiple times (configuration reload, for example), the reference counter for this DLL will be incremented multiple times and it won't be released when the object is finally destructed.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -226,20 +229,32 @@ void ConsumeWindowsEventLog::onSchedule(const std::shared_ptr<core::ProcessConte
...
+  std::string mode;
+  context->getProperty(OutputFormat.getName(), mode);
+
+  writeXML_ = (mode == Both || mode == XML);
+
+  writePlainText_ = (mode == Both || mode == Plaintext);
+
+  if (writeXML_) {
+    char systemDir[MAX_PATH];
+    if (GetSystemDirectory(systemDir, sizeof(systemDir))) {
```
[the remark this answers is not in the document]
am-c-p-p: It is not necessary since minificpp uses Multi-Byte Character Set and macro GetSystemDirectory resolved as GetSystemDirectoryA.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp

extensions/windows-event-log/ConsumeWindowsEventLog.cpp

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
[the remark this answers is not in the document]
am-c-p-p: This is a common way to use any Windows API without specifying 'A' or 'W', for instance GetSystemDirectory usage in chromium:
https://chromium.googlesource.com/chromium/src/base/+/refs/heads/master/base_paths_win.cc
We should use Windows API this way it as well.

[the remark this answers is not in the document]
bakaid: If it consistently gives a compilation error when you try to use LPSTR in place of LPWSTR or vica versa, I can live with this, but I still don't see the benefit of using the macro version of the functions. We are specifically using either the A or the W versions, we don't have code where we would consistently use LPTSTRs, for example, so that it would work both with Unicode or ANSI, so that the macro version would be warranted.

PR #731, opened 2020-02-10
MINIFICPP-1096 fix BackTrace, OOB indexing, tests, appveyor reporting

szaszm:
Squashed commit of the following:

commit b0764d9ad83dee400a0a757578f92eb0d1f37e7e
Author: Szász Márton <szaszm@users.noreply.github.com>
Date:   Mon Feb 3 11:08:44 2020 +0100

    Check unit test failure report on appveyor

commit 8fc8f2a3336111947ab6228c6f05240ad743c91d
Author: Marton Szasz <szaszm01@gmail.com>
Date:   Fri Jan 31 16:15:15 2020 +0000

    MINIFICPP-1096 Fix TailFileTests on windows

    TailFile only supports single character (line) delimiter, but Windows
    uses CRLF ("\r\n") line endings. Changed the tests to write the test
    file in binary mode to avoid LF -> CRLF conversion that breaks TailFile.

    We should consider adapting TailFile to support CRLF as it's standard on
    Windows.

commit c4a1ee6714da534f8fdea96c960a86b2f8dbdcf2
Author: Marton Szasz <szaszm01@gmail.com>
Date:   Wed Jan 29 14:48:40 2020 +0100

    MINIFICPP-1096 fix BackTrace and OOB indexing issues in streams

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/http-curl/client/HTTPStream.cpp
+++ b/extensions/http-curl/client/HTTPStream.cpp
@@ -59,7 +59,7 @@ int HttpStream::writeData(std::vector<uint8_t> &buf, int buflen) {
     throw minifi::Exception{ExceptionType::GENERAL_EXCEPTION, "negative buflen"};
   }
 
-  if (buf.size() < buflen) {
+  if (buf.size() < static_cast<size_t>(buflen)) {
     return -1;
   }
   return writeData(buf.data(), buflen);
--- a/extensions/http-curl/client/HTTPStream.cpp
+++ b/extensions/http-curl/client/HTTPStream.cpp
@@ -22,6 +22,7 @@
 #include <vector>
 #include <memory>
 #include <string>
+#include <Exception.h>
 
 #include "HTTPCallback.h"
 #include "io/validation.h"
@@ -54,10 +55,14 @@ void HttpStream::seek(uint64_t offset) {
 }
 
 int HttpStream::writeData(std::vector<uint8_t> &buf, int buflen) {
-  if ((int) buf.capacity() < buflen) {
+  if (buflen < 0) {
+    throw minifi::Exception{ExceptionType::GENERAL_EXCEPTION, "negative buflen"};
+  }
+
+  if (buf.size() < buflen) {
     return -1;
   }
-  return writeData(reinterpret_cast<uint8_t *>(&buf[0]), buflen);
+  return writeData(buf.data(), buflen);
 }
 
 // data stream overrides
@@ -90,13 +95,17 @@ inline std::vector<uint8_t> HttpStream::readBuffer(const T& t) {
 }
 
 int HttpStream::readData(std::vector<uint8_t> &buf, int buflen) {
-  if ((int) buf.capacity() < buflen) {
+  if (buflen < 0) {
+    throw minifi::Exception{ExceptionType::GENERAL_EXCEPTION, "negative buflen"};
+  }
+
+  if (buf.size() < static_cast<size_t>(buflen)) {
     buf.resize(buflen);
   }
-  int ret = readData(reinterpret_cast<uint8_t*>(&buf[0]), buflen);
+  int ret = readData(buf.data(), buflen);
 
   if (ret < buflen) {
-    buf.resize(ret);
+    buf.resize((std::max)(ret, 0));
   }
   return ret;
 }
--- a/extensions/standard-processors/processors/ExtractText.cpp
+++ b/extensions/standard-processors/processors/ExtractText.cpp
@@ -25,6 +25,7 @@
 
 #include <iostream>
 #include <sstream>
+#include <utility>
 
 #include "ExtractText.h"
 #include "core/ProcessContext.h"
@@ -121,7 +122,7 @@ int64_t ExtractText::ReadCallback::process(std::shared_ptr<io::BaseStream> strea
   ctx_->getProperty(SizeLimit.getName(), sizeLimitStr);
   ctx_->getProperty(RegexMode.getName(), regex_mode);
 
-  if (sizeLimitStr == "")
+  if (sizeLimitStr.empty())
     size_limit = DEFAULT_SIZE_LIMIT;
   else if (sizeLimitStr != "0")
     size_limit = std::stoi(sizeLimitStr);
@@ -130,7 +131,7 @@ int64_t ExtractText::ReadCallback::process(std::shared_ptr<io::BaseStream> strea
 
   while (read_size < size_limit) {
     // Don't read more than config limit or the size of the buffer
-    ret = stream->readData(buffer_, std::min<uint64_t>((size_limit - read_size), buffer_.capacity()));
+    ret = stream->readData(buffer_, std::min<uint64_t>(size_limit - read_size, buffer_.size()));
 
     if (ret < 0) {
       return -1;  // Stream error
@@ -138,7 +139,7 @@ int64_t ExtractText::ReadCallback::process(std::shared_ptr<io::BaseStream> strea
       break;  // End of stream, no more data
     }
 
-    contentStream.write(reinterpret_cast<const char*>(&buffer_[0]), ret);
+    contentStream.write(reinterpret_cast<const char*>(buffer_.data()), ret);
     read_size += ret;
     if (contentStream.fail()) {
       return -1;
@@ -212,10 +213,10 @@ int64_t ExtractText::ReadCallback::process(std::shared_ptr<io::BaseStream> strea
 }
 
 ExtractText::ReadCallback::ReadCallback(std::shared_ptr<core::FlowFile> flowFile, core::ProcessContext *ctx,  std::shared_ptr<logging::Logger> lgr)
-    : flowFile_(flowFile),
+    : flowFile_(std::move(flowFile)),
       ctx_(ctx),
-      logger_(lgr) {
-  buffer_.reserve(std::min<uint64_t>(flowFile->getSize(), MAX_BUFFER_SIZE));
+      logger_(std::move(lgr)) {
+  buffer_.resize(std::min<uint64_t>(flowFile_->getSize(), MAX_BUFFER_SIZE));
 }
 
 } /* namespace processors */
--- a/extensions/standard-processors/tests/unit/ProcessorTests.cpp
+++ b/extensions/standard-processors/tests/unit/ProcessorTests.cpp
@@ -31,6 +31,7 @@
 #include "LogAttribute.h"
 #include "GetFile.h"
 #include "unit/ProvenanceTestHelper.h"
+#include "utils/file/FileUtils.h"
 #include "core/Core.h"
 #include "core/FlowFile.h"
 #include "core/Processor.h"
@@ -199,7 +200,7 @@ TEST_CASE("Test GetFile Ignore", "[getfileCreate3]") {
 #ifdef WIN32
   {
     // hide file on windows, because a . prefix in the filename doesn't imply a hidden file
-    const auto hide_file_error = FileUtils::hide_file(hidden_file_name);
+    const auto hide_file_error = utils::file::FileUtils::hide_file(hidden_file_name.c_str());
     REQUIRE(!hide_file_error);
   }
 #endif /* WIN32 */
--- a/extensions/standard-processors/tests/unit/ProcessorTests.cpp
+++ b/extensions/standard-processors/tests/unit/ProcessorTests.cpp
@@ -199,11 +199,8 @@ TEST_CASE("Test GetFile Ignore", "[getfileCreate3]") {
 #ifdef WIN32
   {
     // hide file on windows, because a . prefix in the filename doesn't imply a hidden file
-    const bool result = SetFileAttributesA(hidden_file_name.c_str(), FILE_ATTRIBUTE_HIDDEN);
-    if (!result) {
-      throw std::runtime_error{ std::system_category().message(GetLastError()) };
-    }
-    REQUIRE(result);
+    const auto hide_file_error = FileUtils::hide_file(hidden_file_name);
+    REQUIRE(!hide_file_error);
   }
 #endif /* WIN32 */
 
--- a/extensions/standard-processors/tests/unit/ProcessorTests.cpp
+++ b/extensions/standard-processors/tests/unit/ProcessorTests.cpp
@@ -22,6 +22,10 @@
 #include <set>
 #include <fstream>
 #include <GenerateFlowFile.h>
+#ifdef WIN32
+#include <fileapi.h>
+#include <system_error>
+#endif /* WIN32 */
 
 #include "TestBase.h"
 #include "LogAttribute.h"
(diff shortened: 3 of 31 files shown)


szaszm:
backwards compatibility notes:
The OOB indexing issue resolution means a backwards compatibility breakage because previously we incorrectly expected only the read buffer `capacity()` to be sufficient (and proceeded to write to the reserved but not allocated memory) but with this PR the new requirement is sufficient `size()`

libminifi/src/io/DataStream.cpp
```diff
@@ -17,13 +17,13 @@
...
-#include <iostream>
 #include <cstdint>
-#include <cstdio>
-#include <cstring>
 #include <string>
 #include <algorithm>
 #include <iterator>
+#include <Exception.h>
+#include <cassert>
+#include <iterator>
```
szaszm: valid linter report on travis: duplicate include

szaszm:
WIP: appveyor is still not red on unit test failure + the above duplicate include

The rest is ready to review if someone's interested

szaszm:
The next appveyor build should be green. The issue was early `ERRORLEVEL` variable substitution, proof that appveyor is red on unit test failure with the fix: https://ci.appveyor.com/project/szaszm/nifi-minifi-cpp/builds/30810385

szaszm:
WIP: there seems to be a SEGFAULT triggered by `ManipulateArchiveTests`, I'm looking into it.

libminifi/include/utils/file/FileUtils.h
```diff
@@ -112,31 +112,33 @@ class FileUtils {
 #endif
   }
 
-  static std::string create_temp_directory(char *format) {
+  static std::string create_temp_directory(const char * const format) {
```
szaszm: This breaks backwards compatibility on POSIX because the buffer pointed-to by the argument is no longer modified. It was never modified on Windows, so the new behavior is consistent but different.
Is this OK to change?

arpadboda: As it's test code and I think nobody has ever made anything that depended on the modified buffer, I'm fine with this.

szaszm: I bailed on the idea since too many processor tests rely on the buffer being modified and they continue to compile but fail at runtime.

szaszm:
GenerateFlowFileTests is a flicker (didn't happen last time), I'm investigating.

libminifi/include/utils/BackTrace.h
```diff
@@ -41,20 +42,16 @@ class TraceResolver;
  */
 class BackTrace {
  public:
-  BackTrace() {
-  }
-  BackTrace(const std::string &name)
-      : name_(name) {
+  BackTrace() = default;
+
+  explicit BackTrace(std::string name)
```
bakaid: `explicit` breaks API.

libminifi/include/utils/file/FileUtils.h
```diff
@@ -112,31 +112,28 @@ class FileUtils {
 #endif
   }
 
-  static std::string create_temp_directory(char *format) {
+  static std::string create_temp_directory(char * const format) {
```
bakaid: What is the purpose of using a const pointer (not pointer to const) here as a function argument? It gives no extra guarantees to the caller.

libminifi/include/utils/file/FileUtils.h
```diff
@@ -112,31 +112,28 @@ class FileUtils {
...
-		  tempDirectory += id.to_string();
-		  create_dir(tempDirectory);
-	  }
-	  return tempDirectory;
+    std::string tempDirectory;
+    char tempBuffer[MAX_PATH];
+    auto ret = GetTempPath(MAX_PATH, tempBuffer);
+    if (ret <= MAX_PATH && ret != 0)
+    {
+      static std::shared_ptr<minifi::utils::IdGenerator> generator;
```
bakaid: I know this is old code, but it is horrifically unsafe. Since it is currently only used from tests I don't mind it so much, but:
 - `minifi::utils::IdGenerator::getIdGenerator` is already a singleton getter so this does function static variable does not make much sense aside from saving a shared_ptr copy
 - only the creation (and initialization to nullptr) of `generator` is protected, the check for nullptr and assignment from `minifi::utils::IdGenerator::getIdGenerator` is not, so it is a race condition
 - `generator->initialize` would not be protected even if `generator` would be initially initialized to `minifi::utils::IdGenerator::getIdGenerator()`, that's why `generator->initialize` should only be called at the beginning of a main executable, when there is only a single thread (see `main/MiNiFiMain.cpp` and `libminifi/test/TestBase.h`)

This whole part should just be rewritten to `utils::IdGenerator::getIdGenerator()->generate(id)`, and if it breaks something, that's the real problem.

libminifi/include/utils/file/FileUtils.h
```diff
@@ -640,10 +637,10 @@ class FileUtils {
     std::vector<char> buf(1024U);
     while (true) {
       ssize_t ret = readlink("/proc/self/exe", buf.data(), buf.size());
-      if (ret == -1) {
+      if (ret < 0) {
```
bakaid: `readlink`'s contract is pretty clear from the POSIX specification (https://pubs.opengroup.org/onlinepubs/9699919799/): it will return -1 on error, not negative on error. This is fine the way it is.

extensions/http-curl/client/HTTPStream.cpp
```diff
@@ -90,13 +95,17 @@ inline std::vector<uint8_t> HttpStream::readBuffer(const T& t) {
 }
 
 int HttpStream::readData(std::vector<uint8_t> &buf, int buflen) {
-  if ((int) buf.capacity() < buflen) {
+  if (buflen < 0) {
+    throw minifi::Exception{ExceptionType::GENERAL_EXCEPTION, "negative buflen"};
+  }
+
+  if (buf.size() < static_cast<size_t>(buflen)) {
```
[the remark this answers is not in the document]
szaszm: Yes, the purpose is to silence the warnings about comparison between signed and unsigned. I think I've applied this universally, but I'll double check.

extensions/standard-processors/processors/ExtractText.cpp
```diff
@@ -121,7 +122,7 @@ int64_t ExtractText::ReadCallback::process(std::shared_ptr<io::BaseStream> strea
   ctx_->getProperty(SizeLimit.getName(), sizeLimitStr);
   ctx_->getProperty(RegexMode.getName(), regex_mode);
 
-  if (sizeLimitStr == "")
+  if (sizeLimitStr.empty())
```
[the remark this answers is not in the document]
szaszm: Normally when I touch a file, I go through the clang-tidy warnings and apply the suggested fixes. This is one of them. Not strictly necessary, but it takes close to 0 effort and looks nicer.

libminifi/include/utils/file/FileUtils.h
szaszm: It belongs to the definition, not the declaration. It doesn't matter for the caller but provides extra checks for the definition. I tend to use `const` whenever it's possible and I don't know a reason not to.

ref1: this slide: https://youtu.be/iz5Qx18H6lg?t=293
ref2: http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#con1-by-default-make-objects-immutable

libminifi/include/utils/file/FileUtils.h
szaszm: I know, but I feel safer with a `< 0` check before making a narrowing cast (to unsigned) below. (To avoid unsigned vs signed comparison warnings.)

extensions/standard-processors/tests/unit/ProcessorTests.cpp
```diff
@@ -180,19 +184,37 @@ TEST_CASE("Test GetFile Ignore", "[getfileCreate3]") {
...
+    ss << dir << utils::file::FileUtils::get_separator() << ".filewithoutanext";
+    return ss.str();
+  }();
+  {
+    std::ofstream file{ hidden_file_name };
+    file << "tempFile";
+  }
+
+#ifdef WIN32
+  {
```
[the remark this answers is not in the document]
szaszm: This is inherently specific to Windows, since on windows "hidden" is a file system attribute while on unix it's a filename prefix. Because of the wildly different semantics, I would only add it inside an `#ifdef WIN32`/`#endif /* WIN32 */` there as well.

Would that be good in your opinion?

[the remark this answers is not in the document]
bakaid: Yeah, that's fine.

libminifi/include/utils/file/FileUtils.h
bakaid: It doesn't hurt, I just didn't understand the purpose, but I can live with it.

libminifi/include/utils/file/FileUtils.h
bakaid: I understand your intent, but it can be confusing for users when looking at the function signature. I agree with making as many things const as possible, but this, in my opinion, reduces readability of the API without having much benefit.
None of your references specifically address the case of having const pointers in function signatures, they are just speaking generically about immutable objects - and as I said, in the generic case, I agree.
Neither do I know of any major projects that would use const pointers on their API.

extensions/standard-processors/processors/ExtractText.cpp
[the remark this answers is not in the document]
bakaid: I don't agree with it looking nicer, but I can live with it.

libminifi/include/utils/file/FileUtils.h
bakaid: @szaszm my bad, your second reference actually addresses the issue, and spells out the reasons I give not to use it in this case:
```Exception Function arguments are rarely mutated, but also rarely declared const. To avoid confusion and lots of false positives, don’t enforce this rule for function arguments.```

Comment (szaszm) [reply libminifi/include/utils/file/FileUtils.h]: Pointers are objects. My interpretation of the exception is that the issue is too minor to be enforced by static analyzers, and is commonly not followed, hence the "false positives". In other words, my interpretation of the exception doesn't suggest that function argument objects shouldn't be declared `const`.

I can see the point behind the readability argument, since `const` is present in the argument list, yet, the string behind the pointer is mutated. I think the risk of misunderstanding is minor given that we expect people who view the source to know the language well enough to know what `const` applies to. With this assumption, IMO the value gained by extra static analysis is greater than the value of lost readability.

I couldn't find any clear guidelines on the `const`-ness of argument pointers. I'm willing to remove `const` if you insist on it.

File: libminifi/include/utils/BackTrace.h [hunk 10]
```diff
@@ -41,20 +42,16 @@ class TraceResolver;
...
+      : name_(std::move(name)) {
   }
-  BackTrace(BackTrace &&) = default;
-  BackTrace(BackTrace &) = delete;
```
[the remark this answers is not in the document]
szaszm: Making it copyable in addition to being movable doesn't break the previous promise. We can keep it non-copyable if desired, but I didn't see the point and since both declarations were wrong I decided to rely on the implicit copy and move ctors.

Given the above, do you still want to disable copy or change the description? (I'd just remove the "movable" word.)

libminifi/include/utils/file/FileUtils.h
bakaid: Our core code is expected to be written by people very familiar with the language. We strive to create an ecosystem where extensions can be written (our API can be used) safely even by people without great familiarity with the language.
This change introduces a risk of misunderstanding by someone less familiar with a language (or someone familiar with it, just running low on caffeine), without bringing comparable benefits.
Please revert it.

libminifi/include/utils/BackTrace.h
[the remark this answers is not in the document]
bakaid: I am OK with it being copyable, but then the description should be changed.

extensions/standard-processors/tests/unit/ProcessorTests.cpp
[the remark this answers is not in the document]
arpadboda: I also used this Win API call in GetFileTests.cpp. In case you create an util func, please make that depend on that, too!

extensions/http-curl/client/HTTPStream.cpp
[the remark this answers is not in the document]
szaszm: I've missed this and `HTTPStream::writeData`, but I've found no more occurences of missing cast.

libminifi/include/utils/BackTrace.h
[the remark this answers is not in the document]
szaszm: changed description

libminifi/include/utils/BackTrace.h
szaszm: removed `explicit`

libminifi/include/utils/file/FileUtils.h
szaszm: reverted

libminifi/include/utils/file/FileUtils.h
szaszm: rewritten definition and fixed a possible data race in `IdGenerator`

szaszm:
WIP: debugging C2VerifyHeartbeatAndStop on windows (flicker)

bakaid:
@arpadboda The branch is currently not mergeable, so after @szaszm rebased it to the current master and I have been able to successfully compile it on Windows and run some tests I would be okay with merging it with the follow-up issue.

szaszm:
rebased to master, compiled and ran the test suite on linux and windows, removed WIP tag given the separate handling of the remaining flicker.

bakaid:
@szaszm Thanks, will start the compilation and tests.

bakaid:
@arpadboda @szaszm Built and ran tests on macOS and Windows. `C2VerifyHeartbeatAndStopSecure` and `C2VerifyHeartbeatAndStop` fail pretty frequently for me, both on macOS and Windows.
If these - as I suspect - are part of the follow-up https://issues.apache.org/jira/browse/MINIFICPP-1172 issue, then it's OK, but then those issues should be addressed sooner than later, as they prevent successful CI runs - both Travis and AppVeyor failed on the latest version of this PR- and finally having successful CI runs would have been one of the goals for this PR.

szaszm:
The issue is that the C2 heartbeat is sometimes missing agentInfo in those test cases. Since it's likely to be because of the referenced bug, I'm ok with handling it there. I'd prefer not to waste effort on debugging it in context of this PR until the referenced issue is fixed.

I'm indifferent on whether we should merge this PR on not before MINIFICPP-1172 is resolved, but I'm against closing the Jira ticket until the CI is stable. If MINIFICPP-1172 doesn't solve the issue, a followup PR/commits are the right way to fix the C2 heartbeat flickers IMO.

szaszm:
Additional info: I could observe the assertion failure in the C2 tests occuring on a civet worker thread while the main thread tries to delete the web server. I only observed this in a Debug build, not in a RelWithDebInfo, but I assume it always happens there and the different observation with RelWithDebInfo builds were because of optimization messing with debuggability.

PR #734, opened 2020-02-11
MINIFICPP-1157 Implement lightweight C2 heartbeat.

msharee9:
Remove agent manifest from regular heartbeat messages.
  Send agent manifest in response to DESCRIBE manifest request

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/conf/minifi.properties
+++ b/conf/minifi.properties
@@ -51,7 +51,9 @@ nifi.database.content.repository.directory.default=${MINIFI_HOME}/content_reposi
 #nifi.c2.flow.base.url=
 #nifi.c2.rest.url=
 #nifi.c2.rest.url.ack=
-nifi.c2.root.classes=DeviceInfoNode,AgentInformationWithoutManifest,FlowInformation
+nifi.c2.root.classes=DeviceInfoNode,AgentInformation,FlowInformation
+## Minimize heartbeat payload size by excluding agent manifest from the heartbeat
+#nifi.c2.root.classes=DeviceInfoNode,AgentInformationWithoutManifest,FlowInformation
 ## heartbeat 4 times a second
 #nifi.c2.agent.heartbeat.period=250
 ## define parameters about your agent 
--- a/conf/minifi.properties
+++ b/conf/minifi.properties
@@ -51,7 +51,7 @@ nifi.database.content.repository.directory.default=${MINIFI_HOME}/content_reposi
 #nifi.c2.flow.base.url=
 #nifi.c2.rest.url=
 #nifi.c2.rest.url.ack=
-nifi.c2.root.classes=DeviceInfoNode,AgentInformation,FlowInformation
+nifi.c2.root.classes=DeviceInfoNode,AgentInformationWithoutManifest,FlowInformation
 ## heartbeat 4 times a second
 #nifi.c2.agent.heartbeat.period=250
 ## define parameters about your agent 
--- a/extensions/coap/tests/CoapIntegrationBase.h
+++ b/extensions/coap/tests/CoapIntegrationBase.h
@@ -41,13 +41,13 @@ class CoapIntegrationBase : public IntegrationBase {
 
   void setUrl(std::string url, CivetHandler *handler);
 
-  virtual ~CoapIntegrationBase();
+  virtual ~CoapIntegrationBase() = default;
 
   void shutdownBeforeFlowController() override {
     stop_webserver(server);
   }
 
-  virtual void run(std::string test_file_location) override {
+  virtual void run(std::string test_file_location) {
     testSetup();
 
     std::shared_ptr<core::Repository> test_repo = std::make_shared<TestRepository>();
@@ -67,6 +67,8 @@ class CoapIntegrationBase : public IntegrationBase {
     std::shared_ptr<core::ProcessGroup> pg = std::shared_ptr<core::ProcessGroup>(ptr.get());
 
     queryRootProcessGroup(pg);
+
+    configureC2RootClasses();
 
     ptr.release();
 
@@ -89,10 +91,6 @@ class CoapIntegrationBase : public IntegrationBase {
  protected:
   CivetServer *server;
 };
-
-CoapIntegrationBase::~CoapIntegrationBase() {
-
-}
 
 void CoapIntegrationBase::setUrl(std::string url, CivetHandler *handler) {
 
--- a/extensions/coap/tests/CoapIntegrationBase.h
+++ b/extensions/coap/tests/CoapIntegrationBase.h
@@ -47,7 +47,7 @@ class CoapIntegrationBase : public IntegrationBase {
     stop_webserver(server);
   }
 
-  virtual void run(std::string test_file_location) {
+  virtual void run(std::string test_file_location) override {
     testSetup();
 
     std::shared_ptr<core::Repository> test_repo = std::make_shared<TestRepository>();
--- a/extensions/http-curl/tests/C2JstackTest.cpp
+++ b/extensions/http-curl/tests/C2JstackTest.cpp
@@ -45,123 +45,114 @@
 #include "CivetServer.h"
 #include <cstring>
 #include "protocols/RESTSender.h"
+#include "HTTPIntegrationBase.h"
+#include "HTTPHandlers.h"
 
-void waitToVerifyProcessor() {
-  std::this_thread::sleep_for(std::chrono::seconds(10));
-}
-
-
-class ConfigHandler : public CivetHandler {
+class VerifyC2Describe : public VerifyC2Base {
  public:
-  ConfigHandler() {
-    calls_ = 0;
-  }
-  bool handlePost(CivetServer *server, struct mg_connection *conn) {
-    calls_++;
-    std::string heartbeat_response = "{\"operation\" : \"heartbeat\",\"requested_operations\": [  {"
-          "\"operation\" : \"describe\", "
-          "\"operationid\" : \"8675309\", "
-          "\"name\": \"jstack\""
-          "}]}";
-      mg_printf(conn, "HTTP/1.1 200 OK\r\nContent-Type: "
-                "text/plain\r\nContent-Length: %lu\r\nConnection: close\r\n\r\n",
-                heartbeat_response.length());
-      mg_printf(conn, "%s", heartbeat_response.c_str());
-
-
-    return true;
+  explicit VerifyC2Describe(bool isSecure)
+      : VerifyC2Base(isSecure) {
   }
 
-  bool handleGet(CivetServer *server, struct mg_connection *conn) {
-    std::ifstream myfile(test_file_location_.c_str());
+  virtual ~VerifyC2Describe() = default;
 
-    if (myfile.is_open()) {
-      std::stringstream buffer;
-      buffer << myfile.rdbuf();
-      std::string str = buffer.str();
-      myfile.close();
-      mg_printf(conn, "HTTP/1.1 200 OK\r\nContent-Type: "
-                "text/plain\r\nContent-Length: %lu\r\nConnection: close\r\n\r\n",
-                str.length());
-      mg_printf(conn, "%s", str.c_str());
-    } else {
-      mg_printf(conn, "HTTP/1.1 500 Internal Server Error\r\n");
-    }
-
-    return true;
+  void testSetup() {
+    LogTestController::getInstance().setTrace<minifi::c2::C2Agent>();
+    LogTestController::getInstance().setDebug<minifi::c2::RESTSender>();
+    LogTestController::getInstance().setInfo<minifi::FlowController>();
+    VerifyC2Base::testSetup();
   }
-  std::string test_file_location_;
-  std::atomic<size_t> calls_;
+
+  void configureC2RootClasses() {
+    configuration->set("nifi.c2.root.classes", "DeviceInfoNode,AgentInformationWithoutManifest,FlowInformation");
+  }
+};
+
+class VerifyC2DescribeJstack : public VerifyC2Describe {
+ public:
+  explicit VerifyC2DescribeJstack(bool isSecure)
+      : VerifyC2Describe(isSecure) {
+  }
+
+  virtual ~VerifyC2DescribeJstack() = default;
+
+  virtual void runAssertions() {
+    assert(LogTestController::getInstance().contains("SchedulingAgent") == true);
+  }
+
+};
+
+class DescribeManifestHandler: public HeartbeatHandler {
+public:
+
+  explicit DescribeManifestHandler(bool isSecure)
+      : HeartbeatHandler(isSecure) {
+  }
+
+  virtual ~DescribeManifestHandler() =  default;
+
+  virtual void handleHeartbeat(const rapidjson::Document& root, struct mg_connection * conn) {
+    sendHeartbeatResponse("DESCRIBE", "manifest", "889345", conn);
+  }
+
+  virtual void handleAcknowledge(const rapidjson::Document& root) {
+    verifyJsonHasAgentManifest(root);
+  }
+};
+
+class DescribeJstackHandler : public HeartbeatHandler {
+ public:
+  explicit DescribeJstackHandler(bool isSecure)
+     : HeartbeatHandler(isSecure) {
+  }
+
+  virtual ~DescribeJstackHandler() = default;
+
+  virtual void handleHeartbeat(const rapidjson::Document& root, struct mg_connection * conn) {
+    sendHeartbeatResponse("DESCRIBE", "jstack", "889398", conn);
+  }
+
+  virtual void handleAcknowledge(const rapidjson::Document& root) {
+    assert(root.HasMember("SchedulingAgent #0") == true);
+  }
+
 };
 
 int main(int argc, char **argv) {
-  mg_init_library(0);
-  LogTestController::getInstance().setInfo<minifi::FlowController>();
-  LogTestController::getInstance().setDebug<minifi::utils::HTTPClient>();
-  LogTestController::getInstance().setDebug<minifi::c2::RESTSender>();
-  LogTestController::getInstance().setTrace<minifi::c2::C2Agent>();
-
-  const char *options[] = { "document_root", ".", "listening_ports", "0", 0 };
-  std::vector<std::string> cpp_options;
-  for (int i = 0; i < (sizeof(options) / sizeof(options[0]) - 1); i++) {
-    cpp_options.push_back(options[i]);
-  }
-
-  CivetServer server(cpp_options);
-
-  std::string port_str = std::to_string(server.getListeningPorts()[0]);
-
-  ConfigHandler h_ex;
-  server.addHandler("/update", h_ex);
-  std::string key_dir, test_file_location;
+  std::string key_dir, test_file_location, url;
+  url = "http://localhost:0/api/heartbeat";
   if (argc > 1) {
-    h_ex.test_file_location_ = test_file_location = argv[1];
-    key_dir = argv[2];
+    test_file_location = argv[1];
+    if (argc > 2) {
+      url = "https://localhost:0/api/heartbeat";
+      key_dir = argv[2];
+    }
   }
 
+  bool isSecure = false;
+  if (url.find("https") != std::string::npos) {
+    isSecure = true;
+  }
+  {
+    VerifyC2Describe harness(isSecure);
 
-  std::shared_ptr<minifi::Configure> configuration = std::make_shared<minifi::Configure>();
+    harness.setKeyDir(key_dir);
 
-  std::string c2_rest_url = "http://localhost:" + port_str + "/update";
+    DescribeManifestHandler responder(isSecure);
 
-  configuration->set("c2.rest.url", c2_rest_url);
-  configuration->set("c2.agent.heartbeat.period", "1000");
+    harness.setUrl(url, &responder);
 
-  std::shared_ptr<core::Repository> test_repo = std::make_shared<TestRepository>();
-  std::shared_ptr<core::Repository> test_flow_repo = std::make_shared<TestFlowRepository>();
+    harness.run(test_file_location);
+  }
 
-  configuration->set(minifi::Configure::nifi_flow_configuration_file, test_file_location);
+  VerifyC2Describe harness(isSecure);
 
-  std::shared_ptr<minifi::io::StreamFactory> stream_factory = minifi::io::StreamFactory::getInstance(configuration);
-  std::shared_ptr<core::ContentRepository> content_repo = std::make_shared<core::repository::VolatileContentRepository>();
-  std::unique_ptr<core::FlowConfiguration> yaml_ptr = std::unique_ptr<core::YamlConfiguration>(
-      new core::YamlConfiguration(test_repo, test_repo, content_repo, stream_factory, configuration, test_file_location));
-  std::shared_ptr<TestRepository> repo = std::static_pointer_cast<TestRepository>(test_repo);
+  harness.setKeyDir(key_dir);
 
-  std::shared_ptr<minifi::FlowController> controller = std::make_shared<minifi::FlowController>(test_repo, test_flow_repo, configuration, std::move(yaml_ptr), content_repo, DEFAULT_ROOT_GROUP_NAME,
-  true);
+  DescribeJstackHandler responder(isSecure);
 
-  core::YamlConfiguration yaml_config(test_repo, test_repo, content_repo, stream_factory, configuration, test_file_location);
+  harness.setUrl(url, &responder);
 
-  std::unique_ptr<core::ProcessGroup> ptr = yaml_config.getRoot(test_file_location);
-  std::shared_ptr<core::ProcessGroup> pg = std::shared_ptr<core::ProcessGroup>(ptr.get());
-  ptr.release();
-  auto start = std::chrono::system_clock::now();
+  harness.run(test_file_location);
 
-  controller->load();
-  controller->start();
-  waitToVerifyProcessor();
-
-  controller->waitUnload(60000);
-  auto then = std::chrono::system_clock::now();
-
-  auto milliseconds = std::chrono::duration_cast<std::chrono::milliseconds>(then - start).count();
-  std::string logs = LogTestController::getInstance().log_output.str();
-  #ifndef WIN32
-  assert(logs.find("SchedulingAgent") != std::string::npos);
-  #endif
-  LogTestController::getInstance().reset();
-  assert(h_ex.calls_ <= (milliseconds / 1000) + 1);
-
-  return 0;
 }
--- a/extensions/http-curl/tests/HTTPHandlers.h
+++ b/extensions/http-curl/tests/HTTPHandlers.h
@@ -351,8 +351,6 @@ class HeartbeatHandler : public CivetHandler {
       : isSecure(isSecure) {
   }
 
-  virtual ~HeartbeatHandler() = default;
-
   std::string readPost(struct mg_connection *conn) {
     std::string response;
     int blockSize = 1024 * sizeof(char), readBytes;
@@ -431,6 +429,8 @@ class HeartbeatHandler : public CivetHandler {
         handleHeartbeat(root, conn);
       } else if (operation == "acknowledge") {
         handleAcknowledge(root);
+      } else {
+        throw std::runtime_error("operation not supported " + operation);
       }
     }
   }
--- a/extensions/http-curl/tests/HTTPHandlers.h
+++ b/extensions/http-curl/tests/HTTPHandlers.h
@@ -20,6 +20,8 @@
 #include "concurrentqueue.h"
 #include "CivetStream.h"
 #include "io/CRCStream.h"
+#include "rapidjson/document.h"
+
 #ifndef LIBMINIFI_TEST_CURL_TESTS_SITETOSITEHTTP_HTTPHANDLERS_H_
 #define LIBMINIFI_TEST_CURL_TESTS_SITETOSITEHTTP_HTTPHANDLERS_H_
 static std::atomic<int> transaction_id;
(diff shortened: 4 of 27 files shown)


libminifi/include/core/state/nodes/AgentInformation.h
```diff
@@ -623,18 +632,70 @@ class AgentManifest : public DeviceInformation {
...
-class AgentInformation : public DeviceInformation, public AgentMonitor, public AgentIdentifier {
+class AgentInformationWithoutManifest : public DeviceInformation, public AgentMonitor, public AgentIdentifier {
+public:
+
+    AgentInformationWithoutManifest(std::string name, utils::Identifier & uuid)
+    : DeviceInformation(name, uuid) {
+      setArray(false);
+ }
+
+    AgentInformationWithoutManifest(const std::string &name)
```
szaszm: Single parameter constructors should either be marked `explicit` or have a comment explaining why implicit conversion from the parameter type is desired.

Same with `AgentInformationWithManifest`.

libminifi/include/core/state/nodes/AgentInformation.h
```diff
@@ -623,18 +632,70 @@ class AgentManifest : public DeviceInformation {
...
- * Purpose and Justification: Prints classes along with their properties for the current agent.
+ * This class is used for regular heartbeat without manifest
+ * A light weight heartbeat
  */
-class AgentInformation : public DeviceInformation, public AgentMonitor, public AgentIdentifier {
+class AgentInformationWithoutManifest : public DeviceInformation, public AgentMonitor, public AgentIdentifier {
+public:
+
+    AgentInformationWithoutManifest(std::string name, utils::Identifier & uuid)
+    : DeviceInformation(name, uuid) {
```
szaszm: Constructor initializer list should be indented +4 spaces to the right.

Use `std::move(name)` to avoid unnecessary copying of `name`, since `DeviceInformation` takes it by-value, too.

libminifi/include/core/state/nodes/AgentInformation.h
```diff
@@ -623,18 +632,70 @@ class AgentManifest : public DeviceInformation {
 };
 
 /**
- * Purpose and Justification: Prints classes along with their properties for the current agent.
+ * This class is used for regular heartbeat without manifest
+ * A light weight heartbeat
  */
-class AgentInformation : public DeviceInformation, public AgentMonitor, public AgentIdentifier {
+class AgentInformationWithoutManifest : public DeviceInformation, public AgentMonitor, public AgentIdentifier {
```
szaszm: Public inheritance means is-a relationship. By looking at the names, this class fits none of its base classes.
Use composition (preferred) or private inheritance for implemented-in-terms-of relationship.

Couldn't this class become just a `serialize_agent_information_without_manifest(...)` free function? OOP is a toolbox, not an obligation.

libminifi/include/core/state/nodes/AgentInformation.h
```diff
@@ -643,16 +704,7 @@ class AgentInformation : public DeviceInformation, public AgentMonitor, public A
...
-
-    SerializedResponseNode ident;
-
-    ident.name = "identifier";
-    ident.value = identifier_;
-
-    SerializedResponseNode agentClass;
-    agentClass.name = "agentClass";
-    agentClass.value = agent_class_;
+    std::vector<SerializedResponseNode> serialized(std::move(AgentInformationWithoutManifest::serialize()));
```
szaszm: `std::move` is redundant, since the result of a function call expression is already an rvalue. It will prevent copy/move elision, thus making the end result even slower than without `std::move`.

libminifi/src/c2/C2Agent.cpp
```diff
@@ -315,16 +317,35 @@ void C2Agent::performHeartBeat() {
...
+    //Send agent manifest in first heartbeat
+    if (!manifest_sent_
+        && (reporter = std::dynamic_pointer_cast<state::response::NodeReporter>(update_sink_))
+        && (agentInfoManifest = reporter->getAgentInformationWithManifest())
+        && metric.first == agentInfoManifest->getName()) {
+
+        metricName = agentInfoManifest->getName();
+        isArray = agentInfoManifest->isArray();
+        metrics = std::move(agentInfoManifest->serialize());
+        manifest_sent_ = true;
```
szaszm: These lines have one extra level of indentation

libminifi/include/core/state/nodes/AgentInformation.h
```diff
@@ -643,16 +704,7 @@ class AgentInformation : public DeviceInformation, public AgentMonitor, public A
   }
 
   std::vector<SerializedResponseNode> serialize() {
```
szaszm: Can the `serialize()` member functions be marked as `const`? They don't seem to mutate the object.

phrocker:
@msharee9 taking a look, thanks.

phrocker:
@msharee9 Why not have a different endpoint that is used for a full posting and then another endpoint for a lightweight HB. The period for those could be different or the full posting could be sent only once ( with success validation ).

Another alternative is to have the C2 server request a full posting and then the registration occur in that case and otherwise always be a light weight HB. That may alleviate some of the complexity per @arpadboda

libminifi/include/core/state/nodes/AgentInformation.h
msharee9: Did not understand why it does not fit is-a relationship to any of its base classes. Can you give example?

No this cannot become a free function unless we change the way we register and load classes. 
This instantiation of this class and other class (AgentInformation) depends on configuration of root node classes in "nifi.c2.root.classes" property that are loaded during C2 initialization in FlowController. I want to keep it that way and be consistent with the existing code base.

libminifi/src/c2/C2Agent.cpp
```diff
@@ -553,6 +577,20 @@ void C2Agent::handle_describe(const C2ContentResponse &resp) {
...
 
+    auto reporter = std::dynamic_pointer_cast<state::response::NodeReporter>(update_sink_);
+    if (reporter != nullptr) {
+        std::vector<std::shared_ptr<state::response::ResponseNode>> metrics_vec;
+
+        C2Payload agentInfo(Operation::ACKNOWLEDGE, resp.ident, false, true);
+        agentInfo.setLabel("agentInfo");
+
+        reporter->getManifestNodes(metrics_vec, 0);
+        for (auto metric : metrics_vec) {
```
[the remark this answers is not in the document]
msharee9: fair enough

msharee9:
> This seems to be a bit complicated and introduce some code duplication.
> 
> Why can't we simply extend agentinformation (for eg. with a ctor parameter) to include/exclude manifest?

To give some background first:
The existing minifi property nifi.c2.root.classes has "AgentInformation" as one of the root classes. And this class has all the information including (agent status, agent id, class and the manifest).
For backward compatibility we want to keep that class as it is.

Going forward, you have a choice to specify if you need a heartbeat with or without manifest.
AgentInformation = (With manifest as it was before this change)
AgentInformationWithoutManifest = (Everything as before except manifest)

Because of the above reason (configurable class name), we cannot introduce a parameter to include/exclude manifest unless we modify the class loader code 
```
template <class T>
std::shared_ptr<T> ClassLoader::instantiate(const std::string &class_name, const std::string &name);
```
with some if/else and I don't think that is a good choice.

I agree that there is some code duplication and I will make some improvement there.

msharee9:
> @msharee9 Could you have a different endpoint that is used for a full posting and then another endpoint for a lightweight HB. The period for those could be different or the full posting could be sent only once ( with success validation ).
> 
> Another alternative is to have the C2 server request a full posting and then the registration occur in that case and otherwise always be a light weight HB. That may alleviate some of the complexity per @arpadboda

The full posting is done only once in this code change and that is during C2 initialization and after that it is lightweight both going to the same endpoint though. I don't think it would work best from C2 server standpoint if we want to send to two different endpoints.

Sending the heartbeat during initialization is an optimization (thanks to @szaszm suggestion), otherwise, the original idea was to only send manifest when C2 server requests for it in DESCRIBE and always send lightweight heartbeats always.

In case the first heavyweight heartbeat is lost due to some reason, the C2 server will request manifest anyhow upon first successful reception of lightweight heartbeat.

libminifi/include/core/state/nodes/AgentInformation.h
msharee9: fair point. Thanks for catching it.

libminifi/src/c2/C2Agent.cpp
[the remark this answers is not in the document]
msharee9: If C2 server does not receive the manifest in the first heartbeat, it requests manifest in DESCRIBE command when it next receives lightweight heartbeat. Sending the manifest during the first heartbeat is a slight optimization.

libminifi/src/FlowController.cpp
```diff
@@ -931,6 +942,35 @@ int16_t FlowController::getMetricsNodes(std::vector<std::shared_ptr<state::respo
   return 0;
 }
 
+int16_t FlowController::getManifestNodes(std::vector<std::shared_ptr<state::response::ResponseNode>>& manifest_vector, uint16_t metricsClass) {
+    std::lock_guard<std::mutex> lock(metrics_mutex_);
+    (void)metricsClass;
+    for (auto metric : agent_information_) {
```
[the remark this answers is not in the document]
msharee9: ok

libminifi/include/core/state/nodes/AgentInformation.h
msharee9: good point. Will remove it

libminifi/src/FlowController.cpp
```diff
@@ -931,6 +942,35 @@ int16_t FlowController::getMetricsNodes(std::vector<std::shared_ptr<state::respo
   return 0;
 }
 
+int16_t FlowController::getManifestNodes(std::vector<std::shared_ptr<state::response::ResponseNode>>& manifest_vector, uint16_t metricsClass) {
+    std::lock_guard<std::mutex> lock(metrics_mutex_);
+    (void)metricsClass;
```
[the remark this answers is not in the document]
msharee9: This is a cache of manifest object (not the actual manifest data). Needed in response to DESCRIBE manifest.

libminifi/src/c2/C2Agent.cpp
```diff
@@ -521,9 +545,9 @@ void C2Agent::handle_describe(const C2ContentResponse &resp) {
     std::vector<std::string> keys;
     std::copy_if(unsanitized_keys.begin(), unsanitized_keys.end(), std::back_inserter(keys), [](std::string key) {return key.find("pass") == std::string::npos;});
     C2Payload response(Operation::ACKNOWLEDGE, resp.ident, false, true);
-    response.setLabel("configuration_options");
+    response.setLabel("configurationOptions");
```
bakaid: Why was this changed? Doesn't this break backward compatibility? (I guess if DESCRIBE has not been formalized yet it doesn't, but I want to make sure.)

libminifi/src/c2/C2Agent.cpp
```diff
@@ -506,13 +527,16 @@ void C2Agent::handle_describe(const C2ContentResponse &resp) {
       }
 
       std::vector<std::shared_ptr<state::response::ResponseNode>> metrics_vec;
-
-      reporter->getResponseNodes(metrics_vec, metric_class_id);
       C2Payload response(Operation::ACKNOWLEDGE, resp.ident, false, true);
       response.setLabel("metrics");
+
+      C2Payload metrics(Operation::ACKNOWLEDGE, resp.ident, false, true);
```
bakaid: Why are two layers of `C2Payload` needed with the same label? "jstack" has only one to which the children containing the real data is added.
"configuration" is rewritten here to have a "configurationOptions" node to which an other "configurationOptions" node is added to which the children are added.
"manifest" is written with the same extra layer.

libminifi/src/c2/C2Agent.cpp
```diff
@@ -539,9 +563,9 @@ void C2Agent::handle_describe(const C2ContentResponse &resp) {
   } else if (resp.name == "manifest") {
     auto keys = configuration_->getConfiguredKeys();
     C2Payload response(Operation::ACKNOWLEDGE, resp.ident, false, true);
-    response.setLabel("configuration_options");
+    response.setLabel("configurationOptions");
```
bakaid: Why does this have the same label as the response for "configuration"?
It would also have the same label as the response for "jstack", but "jstack" hasn't been changed to "configurationOptions".
This does not look right.

libminifi/src/c2/C2Agent.cpp
```diff
@@ -539,9 +563,9 @@ void C2Agent::handle_describe(const C2ContentResponse &resp) {
   } else if (resp.name == "manifest") {
     auto keys = configuration_->getConfiguredKeys();
```
bakaid: This, unlike the "configuration" branch does not sanitize the keys. Isn't this an issue?
(Also, if this is this way because of a mistake, removing the code duplication between the two should prevent such things happening.)

libminifi/src/c2/C2Agent.cpp
```diff
@@ -553,6 +577,20 @@ void C2Agent::handle_describe(const C2ContentResponse &resp) {
...
+      agentInfo.setLabel("agentInfo");
+
+      reporter->getManifestNodes(metrics_vec);
+      for (auto& metric : metrics_vec) {
+          serializeMetrics(agentInfo, metric->getName(), metric->serialize());
+      }
+      response.addPayload(std::move(agentInfo));
+    }
+
     if (device_information_.size() > 0) {
```
bakaid: I know this is not new code, but I don't see any place where `C2Agent`'s `device_information_` would get filled, making this dead code (along with its counterpart in `performHeartBeat`).
Device information seems to be sourced and transmitted from the `DeviceInfoNode` as a "deviceInfo" root node anyway, and not as "AgentInformation".

phrocker:
> > @msharee9 Could you have a different endpoint that is used for a full posting and then another endpoint for a lightweight HB. The period for those could be different or the full posting could be sent only once ( with success validation ).
> > Another alternative is to have the C2 server request a full posting and then the registration occur in that case and otherwise always be a light weight HB. That may alleviate some of the complexity per @arpadboda
> 
> The full posting is done only once in this code change and that is during C2 initialization and after that it is lightweight both going to the same endpoint though. I don't think it would work best from C2 server standpoint if we want to send to two different endpoints.
> 
> Sending the full heartbeat with manifest during initialization is an optimization (thanks to @szaszm suggestion), otherwise, the original idea was to only send manifest when C2 server requests for it in DESCRIBE.
> 
> In case the first heavyweight heartbeat is lost due to some reason, the C2 server will request manifest anyhow upon first successful reception of lightweight heartbeat.

I'm very confused, the manifest should already be part of the full heartbeat. 

In regards to multiple endpoints, CoAP should does this already, which is analogized best with knocking on someone's door -- it is natural that this spec evolve too ( with backwards compatibility in mind)

Nonetheless, my only real concern is that prior functionality is also maintained, even if it's not the default to specifically address high latency environment concerns. If

libminifi/src/c2/C2Agent.cpp
phrocker: Since a release was performed it's technically possible someone used this ( semver on an open source Apache project ) -- so we should maintain it; however, it's not defined in https://cwiki.apache.org/confluence/display/MINIFI/C2+Design specifically as configuration_options. Previously the goal was to fill that in when some of this was finalized.

libminifi/include/core/state/nodes/AgentInformation.h
szaszm: "agent information (without manifest) is a device information"
"agent information (without manifest) is an agent monitor"
"agent information (without manifest) in an agent identifier"
None of the above statements sound true, although I'm not sure about the first one.

In the meantime I realized that this relationship was not introduced now but taken over from old `AgentInformation`, so I'm fine with not changing the design.

Thanks for the explanation why it cannot be a free function. Feel free to mark this resolved.

phrocker: I just happened to see this E-mail and I think this is spot on. This evolved poorly and AgentInformation became more of a compositional relationship with these other artifacts as they changed in name and purpose. 

This should be improved at some point for maintainability.

libminifi/src/c2/C2Agent.cpp
```diff
@@ -315,16 +317,35 @@ void C2Agent::performHeartBeat() {
...
-      payload.addPayload(std::move(child_metric_payload));
+  for (auto metric : root_response_nodes_) {
+    C2Payload child_metric_payload(Operation::HEARTBEAT);
+    bool isArray{false};
+    std::string metricName;
+    std::vector<state::response::SerializedResponseNode> metrics;
+    std::shared_ptr<state::response::NodeReporter> reporter;
+    std::shared_ptr<state::response::ResponseNode> agentInfoManifest;
+
+    //Send agent manifest in first heartbeat
```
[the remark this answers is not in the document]
msharee9: When AgentInformation class is configured, this code uses AgentInformation class, not AgentInformationWithManifest.

AgentInformation class has been modified structurally but, when you instantiate and serialize it, it gives you the full heartbeat including manifest. This class is kept here to maintain backwards compatibility.

May be the function "getAgentInformationWithManifest" name is confusing to you. If you look at this function, it instantiates "AgentInformation" class. Will change the name to "getAgentInformation" instead.

libminifi/src/c2/C2Agent.cpp
msharee9: Agree. The wiki doc is very much out of date and we cannot tell from that if someone might be using it. Will revert the label name.

msharee9:
> > > @msharee9 Could you have a different endpoint that is used for a full posting and then another endpoint for a lightweight HB. The period for those could be different or the full posting could be sent only once ( with success validation ).
> > > Another alternative is to have the C2 server request a full posting and then the registration occur in that case and otherwise always be a light weight HB. That may alleviate some of the complexity per @arpadboda
> > 
> > 
> > The full posting is done only once in this code change and that is during C2 initialization and after that it is lightweight both going to the same endpoint though. I don't think it would work best from C2 server standpoint if we want to send to two different endpoints.
> > Sending the full heartbeat with manifest during initialization is an optimization (thanks to @szaszm suggestion), otherwise, the original idea was to only send manifest when C2 server requests for it in DESCRIBE.
> > In case the first heavyweight heartbeat is lost due to some reason, the C2 server will request manifest anyhow upon first successful reception of lightweight heartbeat.
> 
> I'm very confused, the manifest should already be part of the full heartbeat.
> 
> In regards to multiple endpoints, CoAP should does this already, which is analogized best with knocking on someone's door -- it is natural that this spec evolve too ( with backwards compatibility in mind)
> 
> Nonetheless, my only real concern is that prior functionality is also maintained, even if it's not the default to specifically address high latency environment concerns. If

@phrocker the manifest would still be part of every heartbeat when "AgentInformation" class is configured in "nifi.c2.root.classes" property. This maintains backward compatibility so the prior functionality is maintained.
In the next release, the default class in "nifi.c2.root.classes" will be "AgentInformationWithoutManifest".

libminifi/src/c2/C2Agent.cpp
msharee9: That is because of the way C2Payload and C2ContentResponse classes are structured and the way we serializeJson payload. This needs some cleanup and might require some larger effort. Will take a note and open a Jira ticket to simplify this design.

libminifi/src/c2/C2Agent.cpp
msharee9: good catch. Will make the necessary corrections here.

libminifi/src/c2/C2Agent.cpp
msharee9: For jstack "configuration_options" does not appear in the response payload. The inner payload label matters while serializing json. That is what I found out while testing DESCRIBE jstack. 
On the side note, the DESCRIBE operand should be pstack I believe for a better name for C++. Looks like jstack was taken from Minifi Java.
Nonetheless, removed that label for jstack response.

libminifi/src/c2/C2Agent.cpp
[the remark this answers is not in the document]
bakaid: > May be the function "getAgentInformationWithManifest" name is confusing to you. If you look at this function, it instantiates "AgentInformation" class. Will change the name to "getAgentInformation" instead.
Yep, it did confuse me, thanks for the clarification.

Let me unpack the other parts of why this does not feel right (while describing my understanding of the processes behind our C2 implementation).

The classes we want to use for C2 responses are defined in the `nifi.c2.root.classes` property.
These are then instantiated using our classloader architecture by `FlowController::initializeC2`, static casted into `state::response::ResponseNode`s and stored in root_response_nodes_ .
3 checks are made on these objects using dynamic casts:
 - if they implement `state::response::AgentIdentifier`, the agent identifier and class are set on them
 - if they implement `state::response::AgentMonitor`, the provenance and flowfile repos and the `FlowController` itself are made known to them
 - if they implement `state::response::FlowMonitor` flow-related information and the `FlowController` itself is made known to them

We don't assume anything about these classes other than based on what interfaces they implement and we only interact with them through these interfaces.

These response nodes are exposed through `FlowController::getResponseNodes` (which overrides `StateManager::getResponseNodes`), and get fed into the C2Agent through a convoluted scheme involving registering the C2Agent as an `UpdateController` via `registerUpdateListener` and a TreeUpdateListener through which the `FlowController::getResponseNodes()` is looped back into `FlowController::setResponseNodes` (which in reality is `StateManager::setResponseNodes`, because `FlowController` does not override that function), which will call `setResponseNodes` on all registered UpdateControllers, thereby calling `C2Agent::setResponseNodes`, completing a section of the Rube Goldberg machine that we call our C2 implementation.

The only thing `C2Agent` did so far was to call `serializeMetrics` on these when needed by `performHeartBeat`. `serializeMetrics` depends only on the interfaces implemented by the defined response classes: all behaviour is defined by them.

What your change does, is if the CoreComponent name of the one of these classes match `agentInfo` (which is the name of both `AgentInformationWithoutManifest` and `AgentInformation`) overrides the behaviour of an class defined by the user in `nifi.c2.root.classes` and instantiated by `FlowController` by creating a new instance of `AgentInformation` via `FlowController::getAgentInformation` and using that for the first heartbeat irrespective of what the configured class is.

This, instead of depending on interfaces and letting the behaviour stay at level of the response classes, changes behaviour at the level of the `C2Agent`, based on the name of the response node instance, something which we did not depend on so far. It also makes us expose a hardcoded object factory on `FlowController` for the `AgentInformation` class (`FlowController::getAgentInformation()`); something that was not needed so far, as everything was done dynamically through our classloader.

I do not have a perfect solution for this, and I do not know if a perfect solution exists in the current architecture.

We could change`AgentInformationWithoutManifest` to include the manifest on its first `serialize` call, but that would not be a solution: we cannot be sure and we should not depend on whether that call was done by the C2Agent and whether the resulting heartbeat was successfully delivered, thereby really negating the need to include the manifest in further serializations.

I see two main root causes for this:
 - C2Agent sometimes needs a different set of information, but the information is only accessible through a composite object (`AgentInformation` or `AgentInformationWithoutManifest`) which contains a hardcoded set of information
 - we have no method of communicating the needs of the `C2Agent` to `AgentInformation`, therefore we can't request a particular set of information (in this case `C2Agent` would specify the set of information needed), nor do we have the ability to give feedback about the need for including a particular set of information (in this case `AgentInformation` would be able to decide what information to include based on the state of `C2Agent`, namely whether the first heartbeat has been successfully sent)

 If we can find no better solution I can live with this until we change the architecture, but we will have to consider how to refactor the C2 architecture to better fit our needs.

libminifi/src/c2/C2Agent.cpp
bakaid: Sounds fair, I am OK with this with a follow-up Jira, thank you.

libminifi/src/c2/C2Agent.cpp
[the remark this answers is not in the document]
msharee9: Replying to your first part of the comment, I agree with you that the registration of C2Agent and the code involving setting and getting response nodes is not very clean and difficult to understand. From my understanding, FlowController is NodeReporter (metrics source) and C2Agent is ResponseNodeSink (metrics sink). The TreeUpdateListener runs a loop in a thread I believe which constantly updates the sink cache with metrics and I think this is unnecessary optimization.

Up until now, the FlowController was responsible for instantiating classes that were defined in the nifi.c2.root.classes property, exposed via getResponseNodes and retrieved by the C2Agent when needed. The change I made also exposes the class via FlowController, this time without it being defined in the properties but that is out of necessity for optimization. (I am talking about first full heartbeat). I agree that the behavior wise this is a little hacky that we are now making a behavioral change in C2Agent rather than making that decision in the metrics classes themselves but this is because there is no way of knowing about the C2Agent state in those classes as you mentioned.

We definitely need a cleaner design especially surrounding the structure of metrics classes and the source and sink concepts. I believe may be a visitor approach could make this design much simpler.

Having said that, we can still remove the first heartbeat optimization in lieu of putting back the behavior in the metrics classes whichever is a better tradeoff. What do you think?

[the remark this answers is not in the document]
bakaid: I like having the manifest in the first heartbeat, I think it makes sense.
I dislike how we had to implement it, but if we create follow-up tickets for this to figure out how to refactor the architecture, I am OK with it for the time being.

libminifi/src/FlowController.cpp
```diff
@@ -931,6 +942,35 @@ int16_t FlowController::getMetricsNodes(std::vector<std::shared_ptr<state::respo
   return 0;
 }
 
+int16_t FlowController::getManifestNodes(std::vector<std::shared_ptr<state::response::ResponseNode>>& manifest_vector, uint16_t metricsClass) {
```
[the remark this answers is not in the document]
msharee9: Makes sense. Made it const

libminifi/src/c2/C2Agent.cpp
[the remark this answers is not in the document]
msharee9: C2 refactor design doc work on its way. Will create tickets after we review the doc.

msharee9:
> @msharee9 The new functionality should be tested.
> Something like `C2JstackTest` for requesting a DESCRIBE manifest can be added (although some fixes are required, because `C2JstackTest` does nothing at the moment, because C2 is misconfigured in it, and it asserts nothing, so it passes anyway... the C2 configuration and assertion approach can be used from `C2VerifyHeartbeatAndStop`).
> The lightweight heartbeat can be tested by extending/parameterizing `C2VerifyHeartbeatAndStop`.

@bakaid Refactored existing C2Jstack test and added new tests for light weight heartbeat and describe manifest

extensions/http-curl/tests/C2JstackTest.cpp
```diff
@@ -45,123 +45,114 @@
...
-  LogTestController::getInstance().setInfo<minifi::FlowController>();
-  LogTestController::getInstance().setDebug<minifi::utils::HTTPClient>();
-  LogTestController::getInstance().setDebug<minifi::c2::RESTSender>();
-  LogTestController::getInstance().setTrace<minifi::c2::C2Agent>();
-
-  const char *options[] = { "document_root", ".", "listening_ports", "0", 0 };
-  std::vector<std::string> cpp_options;
-  for (int i = 0; i < (sizeof(options) / sizeof(options[0]) - 1); i++) {
-    cpp_options.push_back(options[i]);
+class DescribeManifestHandler: public HeartbeatHandler {
```
bakaid: It is misleading to have both the DESCRIBE manifest and DESCRIBE jstack tests in a test named C2JstackTest. I would prefer if these were in two separate files (and therefore two separate tests).

conf/minifi.properties
```diff
@@ -51,7 +51,7 @@ nifi.database.content.repository.directory.default=${MINIFI_HOME}/content_reposi
 #nifi.c2.flow.base.url=
 #nifi.c2.rest.url=
 #nifi.c2.rest.url.ack=
-nifi.c2.root.classes=DeviceInfoNode,AgentInformation,FlowInformation
+nifi.c2.root.classes=DeviceInfoNode,AgentInformationWithoutManifest,FlowInformation
```
bakaid: I think changing this by default could cause problems.
If someone wants to use an older C2 server with newer agents without adapting the C2 server to send DESCRIBE manifest requests when needed would have their workflow broken.
I think the possibility of changing to lightweight heartbeats should be documented in `C2.md`, and even a commented out line here with a `nifi.c2.root.classes` containing `AgentInformationWithoutManifest` (and an explanation of what it does) would be fine, but changing it by default does not seem safe.

extensions/http-curl/tests/C2JstackTest.cpp
```diff
@@ -45,123 +45,114 @@
...
+public:
+
+  explicit DescribeManifestHandler(bool isSecure)
+      : HeartbeatHandler(isSecure) {
   }
 
-  CivetServer server(cpp_options);
+  virtual ~DescribeManifestHandler() =  default;
```
arpadboda: The dtor is virtual anyway as it's inherited, I you only have to write it in case there is really something to do there.

extensions/http-curl/tests/HTTPHandlers.h
```diff
@@ -343,4 +345,104 @@ class DeleteTransactionResponder : public CivetHandler {
...
+    std::cerr << post_data << std::endl;
+    if (!IsNullOrEmpty(post_data)) {
+      rapidjson::Document root;
+      rapidjson::ParseResult ok = root.Parse(post_data.data(), post_data.size());
+      std::string operation = root["operation"].GetString();
+      if (operation == "heartbeat") {
+        handleHeartbeat(root, conn);
+      } else if (operation == "acknowledge") {
+        handleAcknowledge(root);
+      }
```
arpadboda: I think think an else branch to throw an exception (to make sure we fail the test in case something unexpected is received) would make sense here.

libminifi/include/core/state/nodes/MetricsBase.h
```diff
@@ -228,6 +228,20 @@ class NodeReporter {
...
+   * @return 0 on Success, -1 on failure
+   */
+  virtual int16_t getManifestNodes(std::vector<std::shared_ptr<state::response::ResponseNode>>& manifest_vector) const = 0;
+
+  /**
+   * Returns a response node containing all agent information with manifest and agent status
+   * @return a shared pointer to agent information
+   */
+  virtual std::shared_ptr<state::response::ResponseNode> getAgentInformation() const {
+      return nullptr;
```
arpadboda: This looks a bit dangerous.
I don't really like functions returning nullptr by default.
Either:
-Make it pure virtual
-Return empty node to make sure the caller won't crash at least
-Throw an exception here in case this should never be called

conf/minifi.properties
msharee9: Ok

extensions/coap/tests/CoapIntegrationBase.h
```diff
@@ -41,13 +41,13 @@ class CoapIntegrationBase : public IntegrationBase {
...
   void setUrl(std::string url, CivetHandler *handler);
 
-  virtual ~CoapIntegrationBase();
+  virtual ~CoapIntegrationBase() = default;
 
   void shutdownBeforeFlowController() override {
     stop_webserver(server);
   }
 
-  virtual void run(std::string test_file_location) override {
```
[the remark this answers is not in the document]
msharee9: May have removed it accidentally. Will put it back.

extensions/http-curl/tests/C2JstackTest.cpp
msharee9: Valid point.

extensions/http-curl/tests/HTTPHandlers.h
msharee9: Ok. Makes sense.

libminifi/include/core/state/nodes/MetricsBase.h
msharee9: Made it pure virtual.

extensions/http-curl/tests/C2JstackTest.cpp
msharee9: Created another test file for DESCRIBE manifest

libminifi/include/core/state/nodes/AgentInformation.h
```diff
@@ -80,7 +80,7 @@ class ComponentManifest : public DeviceInformation {
     return CoreComponent::getName();
   }
 
-  virtual std::vector<SerializedResponseNode> serialize() {
+  virtual std::vector<SerializedResponseNode> serialize() const {
```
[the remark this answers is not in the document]
msharee9: ok. reverted.

extensions/http-curl/tests/HTTPIntegrationBase.h
```diff
@@ -91,4 +87,74 @@ void CoapIntegrationBase::setUrl(std::string url, CivetHandler *handler) {
...
+      : isSecure(isSecure) {
+    char format[] = "/tmp/ssth.XXXXXX";
+    dir = testController.createTempDirectory(format);
+  }
+
+  virtual void testSetup() {
+    LogTestController::getInstance().setDebug<utils::HTTPClient>();
+    LogTestController::getInstance().setDebug<LogTestController>();
+    std::fstream file;
+    ss << dir << "/" << "tstFile.ext";
```
arpadboda: For C2 verification I think we don't need test files. Or do I miss something?

msharee9: yes. We don't need them. Was a remnant from previous code. Removed it.

msharee9:
Closed this PR in lieu of https://github.com/apache/nifi-minifi-cpp/pull/743

PR #741, opened 2020-02-24
MINIFICPP-1139 Implemented.

am-c-p-p:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/windows-event-log/Bookmark.cpp
+++ b/extensions/windows-event-log/Bookmark.cpp
@@ -18,17 +18,19 @@ Bookmark::Bookmark(const std::wstring& channel, const std::wstring& query, const
 
   filePath_ += "Bookmark.txt";
 
-  std::wstring bookmarkXml;
-  if (!getBookmarkXmlFromFile(bookmarkXml)) {
+  if (!getBookmarkXmlFromFile(bookmarkXml_)) {
     return;
   }
 
-  if (!bookmarkXml.empty()) {
-    if (hBookmark_ = EvtCreateBookmark(bookmarkXml.c_str())) {
+  if (!bookmarkXml_.empty()) {
+    if (hBookmark_ = EvtCreateBookmark(bookmarkXml_.c_str())) {
       ok_ = true;
       return;
     }
 
+    logger_->log_error("!EvtCreateBookmark error: %d.", GetLastError());
+
+    bookmarkXml_.clear();
     if (!createEmptyBookmarkXmlFile()) {
       return;
     }
@@ -39,7 +41,7 @@ Bookmark::Bookmark(const std::wstring& channel, const std::wstring& query, const
     return;
   }
 
-  auto hEventResults = EvtQuery(0, channel.c_str(), query.c_str(), EvtQueryChannelPath);
+  const auto hEventResults = EvtQuery(0, channel.c_str(), query.c_str(), EvtQueryChannelPath);
   if (!hEventResults) {
     logger_->log_error("!EvtQuery error: %d.", GetLastError());
     return;
@@ -71,7 +73,18 @@ Bookmark::operator bool() const {
   return ok_;
 }
   
-EVT_HANDLE Bookmark::bookmarkHandle() const {
+EVT_HANDLE Bookmark::getBookmarkHandleFromXML() {
+  if (hBookmark_) {
+    EvtClose(hBookmark_);
+    hBookmark_ = 0;
+  }
+
+  hBookmark_ = EvtCreateBookmark(bookmarkXml_.c_str());
+  if (!(hBookmark_ = EvtCreateBookmark(bookmarkXml_.c_str()))) {
+    logger_->log_error("!EvtCreateBookmark error: %d.", GetLastError());
+    return 0;
+  }
+
   return hBookmark_;
 }
 
@@ -112,7 +125,7 @@ bool Bookmark::getNewBookmarkXml(EVT_HANDLE hEvent, std::wstring& bookmarkXml) {
       bookmarkXml = &buf[0];
 
       return true;
-    } 
+    }
     if (ERROR_SUCCESS != (status = GetLastError())) {
       logger_->log_error("!EvtRender error: %d.", GetLastError());
       return false;
@@ -123,6 +136,8 @@ bool Bookmark::getNewBookmarkXml(EVT_HANDLE hEvent, std::wstring& bookmarkXml) {
 }
 
 void Bookmark::saveBookmarkXml(const std::wstring& bookmarkXml) {
+  bookmarkXml_ = bookmarkXml;
+
   // Write new bookmark over old and in the end write '!'. Then new bookmark is read until '!'. This is faster than truncate.
   file_.seekp(std::ios::beg);
 
@@ -130,7 +145,6 @@ void Bookmark::saveBookmarkXml(const std::wstring& bookmarkXml) {
 
   file_.flush();
 }
-
 
 bool Bookmark::createEmptyBookmarkXmlFile() {
   if (file_.is_open()) {
@@ -206,7 +220,7 @@ bool Bookmark::getBookmarkXmlFromFile(std::wstring& bookmarkXml) {
   // '!' should be at the end of bookmark.
   auto pos = bookmarkXml.find(L'!');
   if (std::wstring::npos == pos) {
-    logger_->log_error("No '!' in bookmarXml '%ws'", bookmarkXml.c_str());
+    logger_->log_error("No '!' in bookmarXml '%ls'", bookmarkXml.c_str());
     bookmarkXml.clear();
     return createEmptyBookmarkXmlFile();
   }
--- a/extensions/windows-event-log/Bookmark.cpp
+++ b/extensions/windows-event-log/Bookmark.cpp
@@ -28,7 +28,7 @@ Bookmark::Bookmark(const std::wstring& channel, const std::wstring& query, const
       return;
     }
 
-    logger_->log_error("!EvtCreateBookmark error: %d.", GetLastError());
+    LOG_LAST_ERROR(EvtCreateBookmark);
 
     bookmarkXml_.clear();
     if (!createEmptyBookmarkXmlFile()) {
@@ -37,26 +37,26 @@ Bookmark::Bookmark(const std::wstring& channel, const std::wstring& query, const
   }
 
   if (!(hBookmark_ = EvtCreateBookmark(0))) {
-    logger_->log_error("!EvtCreateBookmark error: %d.", GetLastError());
+    LOG_LAST_ERROR(EvtCreateBookmark);
     return;
   }
 
   const auto hEventResults = EvtQuery(0, channel.c_str(), query.c_str(), EvtQueryChannelPath);
   if (!hEventResults) {
-    logger_->log_error("!EvtQuery error: %d.", GetLastError());
+    LOG_LAST_ERROR(EvtQuery);
     return;
   }
   const utils::ScopeGuard guard_hEventResults([hEventResults]() { EvtClose(hEventResults); });
 
   if (!EvtSeek(hEventResults, 0, 0, 0, EvtSeekRelativeToLast)) {
-    logger_->log_error("!EvtSeek error: %d.", GetLastError());
+    LOG_LAST_ERROR(EvtSeek);
     return;
   }
 
   DWORD dwReturned{};
   EVT_HANDLE hEvent{};
   if (!EvtNext(hEventResults, 1, &hEvent, INFINITE, 0, &dwReturned)) {
-    logger_->log_error("!EvtNext error: %d.", GetLastError());
+    LOG_LAST_ERROR(EvtNext);
     return;
   }
 
@@ -81,7 +81,7 @@ EVT_HANDLE Bookmark::getBookmarkHandleFromXML() {
 
   hBookmark_ = EvtCreateBookmark(bookmarkXml_.c_str());
   if (!(hBookmark_ = EvtCreateBookmark(bookmarkXml_.c_str()))) {
-    logger_->log_error("!EvtCreateBookmark error: %d.", GetLastError());
+    LOG_LAST_ERROR(EvtCreateBookmark);
     return 0;
   }
 
@@ -102,7 +102,7 @@ bool Bookmark::saveBookmark(EVT_HANDLE hEvent)
 
 bool Bookmark::getNewBookmarkXml(EVT_HANDLE hEvent, std::wstring& bookmarkXml) {
   if (!EvtUpdateBookmark(hBookmark_, hEvent)) {
-    logger_->log_error("!EvtUpdateBookmark error: %d.", GetLastError());
+    LOG_LAST_ERROR(EvtUpdateBookmark);
     return false;
   }
 
@@ -118,7 +118,7 @@ bool Bookmark::getNewBookmarkXml(EVT_HANDLE hEvent, std::wstring& bookmarkXml) {
       std::vector<wchar_t> buf(bufferSize / 2 + 1);
 
       if (!EvtRender(0, hBookmark_, EvtRenderBookmark, bufferSize, &buf[0], &bufferUsed, &propertyCount)) {
-        logger_->log_error("!EvtRender error: %d.", GetLastError());
+        LOG_LAST_ERROR(EvtRender);
         return false;
       }
 
@@ -127,7 +127,7 @@ bool Bookmark::getNewBookmarkXml(EVT_HANDLE hEvent, std::wstring& bookmarkXml) {
       return true;
     }
     if (ERROR_SUCCESS != (status = GetLastError())) {
-      logger_->log_error("!EvtRender error: %d.", GetLastError());
+      LOG_LAST_ERROR(EvtRender);
       return false;
     }
   }
--- a/extensions/windows-event-log/Bookmark.cpp
+++ b/extensions/windows-event-log/Bookmark.cpp
@@ -3,6 +3,7 @@
 #include <direct.h>
 
 #include "utils/file/FileUtils.h"
+#include "utils/ScopeGuard.h"
 
 namespace org {
 namespace apache {
@@ -10,7 +11,7 @@ namespace nifi {
 namespace minifi {
 namespace processors {
 
-Bookmark::Bookmark(const std::string& bookmarkRootDir, const std::string& uuid, std::shared_ptr<logging::Logger> logger)
+Bookmark::Bookmark(const std::wstring& channel, const std::wstring& query, const std::string& bookmarkRootDir, const std::string& uuid, std::shared_ptr<logging::Logger> logger)
   :logger_(logger) {
   if (!createUUIDDir(bookmarkRootDir, uuid, filePath_))
     return;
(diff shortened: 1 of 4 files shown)


extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -248,21 +239,119 @@ void ConsumeWindowsEventLog::onSchedule(const std::shared_ptr<core::ProcessConte
...
+            commitAndSaveBookmark = false;
+          }
+        }
+      }
+
+      hasEvent = true;
+    }
+
+    if (commitAndSaveBookmark) {
+      session.commit();
```
szaszm: Can we postpone `commit()` to outside the outermost loop, running it only once every `onTrigger()` call, unless specified otherwise via `batch_commit_size_`? For performance reasons.

extensions/windows-event-log/ConsumeWindowsEventLog.h
```diff
@@ -101,14 +100,12 @@ class ConsumeWindowsEventLog : public core::Processor
...
-  void unsubscribe();
-  int processQueue(const std::shared_ptr<core::ProcessSession> &session);
+  void processEventsAfterBookmark(core::ProcessSession& session);
+  void processEventRender(const EventRender& renderedData, core::ProcessSession& session);
   wel::WindowsEventLogHandler getEventLogHandler(const std::string & name);
   bool insertHeaderName(wel::METADATA_NAMES &header, const std::string &key, const std::string &value);
   void LogWindowsError();
-  void processEvent(EVT_HANDLE eventHandle);
-  bool processEventsAfterBookmark(EVT_HANDLE hEventResults, const std::wstring& channel, const std::wstring& query);
+  bool processEvent(EVT_HANDLE eventHandle, EventRender& renderedData);
```
szaszm: This member function deserves a Doxygen-style comment describing the behavior and the meaning of the return value, since it's not obvious by looking at the signature.

am-c-p-p:
@szaszm Correct, this is the purpose of the changes.

extensions/windows-event-log/Bookmark.cpp
```diff
@@ -205,7 +206,7 @@ bool Bookmark::getBookmarkXmlFromFile(std::wstring& bookmarkXml) {
   // '!' should be at the end of bookmark.
   auto pos = bookmarkXml.find(L'!');
   if (std::wstring::npos == pos) {
-    logger_->log_error("No '!' in bookmarXml '%s'", bookmarkXml.c_str());
+    logger_->log_error("No '!' in bookmarXml '%ws'", bookmarkXml.c_str());
```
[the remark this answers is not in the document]
am-c-p-p: According to https://devblogs.microsoft.com/oldnewthing/20190830-00/?p=102823 it works in Windows. I'll change to comply with `C standard`.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -248,21 +239,119 @@ void ConsumeWindowsEventLog::onSchedule(const std::shared_ptr<core::ProcessConte
...
+    if (!hEventResults) {
+      logger_->log_error("!EvtQuery error: %d.", GetLastError());
       return;
     }
+    const utils::ScopeGuard guard_hEventResults([hEventResults]() { EvtClose(hEventResults); });
+
+    if (!EvtSeek(hEventResults, 1, pBookmark_->bookmarkHandle(), 0, EvtSeekRelativeToBookmark)) {
+      logger_->log_error("!EvtSeek error: %d.", GetLastError());
+      return;
+    }
```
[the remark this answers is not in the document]
am-c-p-p: There is only common code `EvtQuery(0, channel.c_str(), query.c_str(), EvtQueryChannelPath)`.

extensions/windows-event-log/Bookmark.cpp
```diff
@@ -111,8 +112,8 @@ bool Bookmark::getNewBookmarkXml(EVT_HANDLE hEvent, std::wstring& bookmarkXml) {
       bookmarkXml = &buf[0];
 
       return true;
-    }
-    else if (ERROR_SUCCESS != (status = GetLastError())) {
+    }
```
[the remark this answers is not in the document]
am-c-p-p: Fixed.

extensions/windows-event-log/Bookmark.cpp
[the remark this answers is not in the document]
am-c-p-p: Fixed.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -248,21 +239,119 @@ void ConsumeWindowsEventLog::onSchedule(const std::shared_ptr<core::ProcessConte
...
+          logger_->log_error("!EvtNext error %d.", status);
+        }
+        break;
+      }
+      const utils::ScopeGuard guard_hEvent([hEvent]() { EvtClose(hEvent); });
+
+      commitAndSaveBookmark = false;
+      EventRender renderedData;
+      if (processEvent(hEvent, renderedData)) {
+        if (pBookmark_->getNewBookmarkXml(hEvent, bookmarkXml)) {
```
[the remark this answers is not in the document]

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -248,21 +239,119 @@ void ConsumeWindowsEventLog::onSchedule(const std::shared_ptr<core::ProcessConte
...
+                               std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::high_resolution_clock::now() - before_commit).count());
+
+            pBookmark_->saveBookmarkXml(bookmarkXml);
+
+            if (session.outgoingConnectionsFull("success")) {
+              return;
+            }
+
+            commitAndSaveBookmark = false;
+          }
```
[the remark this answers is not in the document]

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
am-c-p-p: Removed outer loop since fixed a bug (should be used handle from last saved xml) `pBookmark_->getBookmarkHandleFromXML())`.

extensions/windows-event-log/ConsumeWindowsEventLog.h
am-c-p-p: Fixed.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -391,322 +467,122 @@ void ConsumeWindowsEventLog::substituteXMLPercentageItems(pugi::xml_document& do
...
-}
-
-int ConsumeWindowsEventLog::processQueue(const std::shared_ptr<core::ProcessSession> &session)
+void ConsumeWindowsEventLog::processEventRender(const EventRender& renderedData, core::ProcessSession& session)
 {
-  struct WriteCallback: public OutputStreamCallback {
+  struct WriteCallback : public OutputStreamCallback {
     WriteCallback(const std::string& str)
-      : data_(str.c_str()), size_(str.length()) {
+      : str_(str.c_str()) {
```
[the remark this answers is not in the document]

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -248,21 +239,119 @@ void ConsumeWindowsEventLog::onSchedule(const std::shared_ptr<core::ProcessConte
...
+      return;
+    }
+
+    size_t eventCount = 0;
+    auto before_time = std::chrono::high_resolution_clock::now();
+    utils::ScopeGuard timeGuard([&]() {
+      logger_->log_debug("processed %d Events in %llu ms",
+                         eventCount,
+                         std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::high_resolution_clock::now() - before_time).count());
+    });
```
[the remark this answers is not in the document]

extensions/windows-event-log/ConsumeWindowsEventLog.h
```diff
@@ -101,14 +100,11 @@ class ConsumeWindowsEventLog : public core::Processor
...
-  bool subscribe(const std::shared_ptr<core::ProcessContext> &context);
-  void unsubscribe();
-  int processQueue(const std::shared_ptr<core::ProcessSession> &session);
+  void putEventRenderFlowFileToSession(const EventRender& eventRender, core::ProcessSession& session);
   wel::WindowsEventLogHandler getEventLogHandler(const std::string & name);
   bool insertHeaderName(wel::METADATA_NAMES &header, const std::string &key, const std::string &value);
   void LogWindowsError();
-  void processEvent(EVT_HANDLE eventHandle);
-  bool processEventsAfterBookmark(EVT_HANDLE hEventResults, const std::wstring& channel, const std::wstring& query);
+  bool createEventRender(EVT_HANDLE eventHandle, EventRender& eventRender);
```
szaszm: A small suggestion: "Throw an exception to signal that a function can’t perform its assigned task" (instead of returning success/failure `bool`)
This way `eventRender` can become the return value, making the function more intuitive to use.

http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#e2-throw-an-exception-to-signal-that-a-function-cant-perform-its-assigned-task

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -271,19 +279,84 @@ void ConsumeWindowsEventLog::onTrigger(const std::shared_ptr<core::ProcessContex
...
+    return true;
+  };
 
-  const auto now = GetTickCount64();
+  size_t eventCount = 0;
+  const auto before_time = std::chrono::high_resolution_clock::now();
+  utils::ScopeGuard timeGuard([&]() {
+    logger_->log_debug("processed %d Events in %llu ms",
+                       eventCount,
+                       std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - before_time).count());
```
bakaid: @am-c-p-p you have to changed this single instance from `high_resolution_clock` to `steady_clock`. Subtracting a time point created with one type clock from a time point created by an another is a very bad idea.
Please change *all* instances of `high_resolution_clock` to `steady_clock`.

bakaid: @am-c-p-p you have changed this single instance from `high_resolution_clock` to `steady_clock`. Subtracting a time point created with one type clock from a time point created by an another is a very bad idea.
Please change *all* instances of `high_resolution_clock` to `steady_clock`.

bakaid:
Ran some backpressure tests with a CWEL -> S2S to NiFi flow, seems to handle it without significant memory increase and can continue event collection after backpressure subsides.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -391,322 +463,122 @@ void ConsumeWindowsEventLog::substituteXMLPercentageItems(pugi::xml_document& do
...
+    logger_->log_error("!EvtRender error %d.", GetLastError());
+    return false;
   }
-}
 
-bool ConsumeWindowsEventLog::processEventsAfterBookmark(EVT_HANDLE hEventResults, const std::wstring& channel, const std::wstring& query) {
-  if (!EvtSeek(hEventResults, 1, pBookmark_->bookmarkHandle(), 0, EvtSeekRelativeToBookmark)) {
-    logger_->log_error("!EvtSeek error %d.", GetLastError());
+  if (used > maxBufferSize_) {
+    logger_->log_error("Dropping event %x because it couldn't be rendered within %ll bytes.", hEvent, maxBufferSize_);
```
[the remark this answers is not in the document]
szaszm: For `uint64_t`, the `PRIu64` macro should be used from `<cinttypes>`, since nothing guarantees that it's a typedef to `unsigned long long int`. On my machine, it's a typedef to `unsigned long int` and I get warnings for those format strings.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
am-c-p-p: This is not an `idea`, it is omission :-)

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -191,23 +192,14 @@ bool ConsumeWindowsEventLog::insertHeaderName(wel::METADATA_NAMES &header, const
 }
 
 void ConsumeWindowsEventLog::onSchedule(const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSessionFactory> &sessionFactory) {
+  sessionFactory_ = sessionFactory;
```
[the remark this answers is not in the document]

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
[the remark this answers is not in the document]

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -248,21 +239,38 @@ void ConsumeWindowsEventLog::onSchedule(const std::shared_ptr<core::ProcessConte
...
+    }
+    pBookmark_ = std::make_unique<Bookmark>(wstrChannel_, wstrQuery_, bookmarkDir, getUUIDStr(), logger_);
+    if (!*pBookmark_) {
+      pBookmark_.reset();
+      return;
+    }
   }
+
+  context->getProperty(MaxBufferSize.getName(), maxBufferSize_);
+  logger_->log_debug("ConsumeWindowsEventLog: maxBufferSize_ %lld", maxBufferSize_);
```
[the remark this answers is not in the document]

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -271,19 +279,84 @@ void ConsumeWindowsEventLog::onTrigger(const std::shared_ptr<core::ProcessContex
...
-  const auto flowFileCount = processQueue(session);
+  const auto commitAndSaveBookmark = [&] (const std::wstring& bookmarkXml){
+    const auto before_commit = std::chrono::high_resolution_clock::now();
+    session->commit();
+    logger_->log_debug("processQueue commit took %llu ms",
+      std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::high_resolution_clock::now() - before_commit).count());
+
+    pBookmark_->saveBookmarkXml(bookmarkXml);
+
+    if (session->outgoingConnectionsFull("success")) {
```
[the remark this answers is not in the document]

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -391,322 +463,122 @@ void ConsumeWindowsEventLog::substituteXMLPercentageItems(pugi::xml_document& do
...
-      break;
-    }
-
-    processEvent(hEvent);
-
-    EvtClose(hEvent);
+  size = used;
+  std::vector<wchar_t> buf(size / 2 + 1);
+  if (!EvtRender(NULL, hEvent, EvtRenderEventXml, size, &buf[0], &used, &propertyCount)) {
+    logger_->log_error("!EvtRender error: %d.", GetLastError());
```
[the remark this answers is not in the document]
am-c-p-p: Fixed by using %x.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -271,19 +279,84 @@ void ConsumeWindowsEventLog::onTrigger(const std::shared_ptr<core::ProcessContex
...
+    }
+
+    return true;
+  };
 
-  const auto now = GetTickCount64();
+  size_t eventCount = 0;
+  const auto before_time = std::chrono::high_resolution_clock::now();
+  utils::ScopeGuard timeGuard([&]() {
+    logger_->log_debug("processed %d Events in %llu ms",
```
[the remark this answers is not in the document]

extensions/windows-event-log/ConsumeWindowsEventLog.cpp

extensions/windows-event-log/ConsumeWindowsEventLog.h
am-c-p-p: Agree, but since it is minor will leave it as it is.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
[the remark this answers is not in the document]
bakaid: @szaszm I agree that this is the completely correct solution, but as I said before on an another PR, I don't see much point in selectively using inttypes macros in random parts of the code, when most of it does not use it. We should refactor the code to change every instance to the proper macro and then stick to using them consistently.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -271,18 +280,83 @@ void ConsumeWindowsEventLog::onTrigger(const std::shared_ptr<core::ProcessContex
     return;
   }
 
-  const auto flowFileCount = processQueue(session);
+  const auto commitAndSaveBookmark = [&] (const std::wstring& bookmarkXml){
+    const auto before_commit = std::chrono::steady_clock::now();
+    session->commit();
+    logger_->log_debug("processQueue commit took %llu ms",
+                       std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - before_commit).count());
```
szaszm: There's already a thread about `%llu`, but this case is somewhat different. Here, the standard doesn't specify the returned type, but it specifies the minimum number of bits.
I suggest assigning the result of `milliseconds::count` to a `uint64_t` (where it will definitely fit), then printing it as a normal `uint64_t`, i.e. with the `"%" PRIu64` format specifier

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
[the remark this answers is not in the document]
szaszm: I prefer to keep new code correct even if this means inconsistency with old code. I'll do the  wider scale replacement in [MINIFICPP-1151](https://issues.apache.org/jira/projects/MINIFICPP/issues/MINIFICPP-1151) as it results in compiler warnings.

bakaid:
@am-c-p-p please let us know in a comment when you are ready with the review fixes.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp

am-c-p-p:
All review comments are addressed.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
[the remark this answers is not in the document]
bakaid: @amarmer the `%x` is still not replaced, please see my previous comment here about why it is necessary.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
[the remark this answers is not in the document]
bakaid: @amarmer `Bookmark.cpp` is still full of it.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -271,19 +280,88 @@ void ConsumeWindowsEventLog::onTrigger(const std::shared_ptr<core::ProcessContex
     return;
   }
 
-  const auto flowFileCount = processQueue(session);
+  struct TimeDiff {
+    auto operator()() const {
+      return std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - time_).count();
+    }
+    const decltype(std::chrono::steady_clock::now()) time_ = std::chrono::steady_clock::now();
```
szaszm: Use `auto` instead of `decltype(std::chrono::steady_clock::now())` for simplicity. They are equivalent in this case, since we work with values, not references.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -271,19 +280,88 @@ void ConsumeWindowsEventLog::onTrigger(const std::shared_ptr<core::ProcessContex
     return;
   }
 
-  const auto flowFileCount = processQueue(session);
+  struct TimeDiff {
+    auto operator()() const {
```
szaszm: Since the scope is local and we want to be about to reliably print the duration as milliseconds, I suggest using `uint64_t` instead of `auto` for the return type and avoid casting at the call sites.

I can also live with wrapping the call sites in `uint64_t{ timeDiff() }`, but I prefer an explicit return type for the sake of easy logging.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -271,19 +280,88 @@ void ConsumeWindowsEventLog::onTrigger(const std::shared_ptr<core::ProcessContex
...
+          return;
+        }
+
+        commitAndSaveBookmarkCount = eventCount;
+      }
     }
   }
+
+  if (eventCount > commitAndSaveBookmarkCount) {
+    commitAndSaveBookmark(bookmarkXml);
```
szaszm: Would it be possible to commit only once every `onTrigger()`, for performance reasons? I had a thread open about this but it was closed with a seemingly unrelated note.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -271,19 +280,88 @@ void ConsumeWindowsEventLog::onTrigger(const std::shared_ptr<core::ProcessContex
...
+    auto operator()() const {
+      return std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - time_).count();
+    }
+    const decltype(std::chrono::steady_clock::now()) time_ = std::chrono::steady_clock::now();
+  };
+
+  const auto commitAndSaveBookmark = [&] (const std::wstring& bookmarkXml) {
+    const TimeDiff timeDiff;
+    session->commit();
+    logger_->log_debug("processQueue commit took %" PRIu64 " ms", (uint64_t)timeDiff());
```
szaszm: Please don't use C-style casts! My suggestion about the explicit return value would make it redundant in this case.

http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es48-avoid-casts
http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es49-if-you-must-use-a-cast-use-a-named-cast (note: here, we don't need a cast)

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
bakaid: @szaszm If `batch_commit_size_` is not 0, we must commit once every batch size, that's the point of the feature, and it is an important performance consideration (the uncommitted flow files would otherwise just take up more and more memory and the flow would be blocked, as further processors can't work on uncommitted flow files).
This particular code is to make sure we safely commit the rest of the flow files (or all of the flow files, if `batch_commit_size_ == 0U`) before saving the bookmark. This is required, because committing can throw an exception, causing us to roll back, but since we would already have saved the bookmark, we would skip these events, causing data loss.
An another commit will automatically occur when we return from onTrigger, but since there will be no flow files to commit, that will not have much performance impact.

We can change to framework in the future to make it possible to disable autocommit, but until then the tradeoff between having an empty commit and not losing data necessitates this.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
am-c-p-p: What is there a difference between `(uint64_t)timeDiff())` and `uint64_t{ timeDiff() }`?

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
am-c-p-p: `auto` for class members allowed for `static` only.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
am-c-p-p: We need cast because of `narrowing conversion` error.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
szaszm: This suggests that the underlying type of `milliseconds` is a signed type and a signed -> unsigned conversion is narrowing. Narrowing conversions are not allowed with direct/copy list-initialization.

I was wrong when suggesting `uint64_t`, because the underlying type is a signed integer type. Try `int64_t`, that will most likely work. The standard only specifies "signed integer type of at least 45 bits". https://en.cppreference.com/w/cpp/chrono/duration
Please keep the list initialization to prevent lossy conversion in the future if the underlying type changed to a wider one, like `int128_t`.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
szaszm: Shouldn't be the case since you've changed the function to return `int64_t`

am-c-p-p:
New review comments are addressed.

PR #743, opened 2020-03-03
Minificpp 1169 - Simplify C2 metrics collection and reporting

msharee9:

### For all changes:
     in the commit message?

### For code changes:

### Note:


C2.md
```diff
@@ -49,6 +49,11 @@ Release 0.6.0: Please note that all c2 properties now exist as `nifi.c2.*`. If y
 files contain the former naming convention of `c2.*`, we will continue to support that as
 an alternate key, but you are encouraged to switch your configuration options as soon as possible.
```
msharee9: This will not be true anymore. Will take care of this

conf/minifi.properties
```diff
@@ -52,6 +52,8 @@ nifi.database.content.repository.directory.default=${MINIFI_HOME}/content_reposi
 #nifi.c2.rest.url=
 #nifi.c2.rest.url.ack=
 nifi.c2.root.classes=DeviceInfoNode,AgentInformation,FlowInformation
+## Minimize heartbeat payload size by excluding agent manifest from the heartbeat
```
msharee9: This will not be true anymore

libminifi/src/FlowController.cpp
```diff
@@ -449,13 +451,9 @@ void FlowController::initializeC2() {
 
   if (!c2_initialized_) {
     configuration_->setAgentIdentifier(identifier_str);
-    state::StateManager::initialize();
-    std::shared_ptr<c2::C2Agent> agent = std::make_shared<c2::C2Agent>(std::dynamic_pointer_cast<FlowController>(shared_from_this()), std::dynamic_pointer_cast<FlowController>(shared_from_this()),
+    c2_agent_ = std::make_shared<c2::C2Agent>(std::dynamic_pointer_cast<FlowController>(shared_from_this()), std::dynamic_pointer_cast<FlowController>(shared_from_this()),
```
bakaid: This was and still is a `shared_ptr` cycle between `C2Agent` and `FlowController`, which is not a bad practice and the reason for many of our memory leaks.
Since we are refactoring this we should solve this as well.

libminifi/include/utils/ThreadPool.h
```diff
@@ -412,6 +414,22 @@ class ThreadPool {
   void run_tasks(std::shared_ptr<WorkerThread> thread);
 };
 
+template<typename T>
+bool ThreadPool<T>::execute(Worker<T> &&task) {
```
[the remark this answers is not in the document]
msharee9: Reverted this change.

libminifi/src/FlowController.cpp
msharee9: As discussed, we will take this in another PR as it requires some major changes that might break interfaces.

szaszm: The base class changes of `FlowController` also break API backwards compatibility. If we care about bw compat in this PR, we should revert that as well. If we don't, then we should probably go ahead with the above as well.

I was not part of the cited discussion, so I might be missing something.

extensions/http-curl/tests/HTTPHandlers.h
```diff
@@ -343,4 +345,104 @@ class DeleteTransactionResponder : public CivetHandler {
...
 
+class HeartbeatHandler : public CivetHandler {
+ public:
+  explicit HeartbeatHandler(bool isSecure)
+      : isSecure(isSecure) {
+  }
+
+  std::string readPost(struct mg_connection *conn) {
+    std::string response;
+    int blockSize = 1024 * sizeof(char), readBytes;
```
arpadboda: :suspect: 

Why do we read size of blocksize in case the buffer size is 1024?
My gut feeling says that this should be std::min or someting like that.

extensions/http-curl/tests/HTTPIntegrationBase.h
```diff
@@ -91,4 +87,64 @@ void CoapIntegrationBase::setUrl(std::string url, CivetHandler *handler) {
...
+  explicit VerifyC2Base(bool isSecure)
+      : isSecure(isSecure) {
+  }
+
+  virtual void testSetup() {
+    LogTestController::getInstance().setDebug<utils::HTTPClient>();
+    LogTestController::getInstance().setDebug<LogTestController>();
+  }
+
+  void runAssertions() {
```
arpadboda: This should be defined by any derived class (to make sure they test something), so I think you can delete it.
That has the advantage of making this baseclass abstract.

extensions/http-curl/tests/HTTPIntegrationBase.h
```diff
@@ -91,4 +87,64 @@ void CoapIntegrationBase::setUrl(std::string url, CivetHandler *handler) {
...
+  }
+
+  virtual void queryRootProcessGroup(std::shared_ptr<core::ProcessGroup> pg) {
+    std::shared_ptr<core::Processor> proc = pg->findProcessor("invoke");
+    assert(proc != nullptr);
+
+    std::shared_ptr<minifi::processors::InvokeHTTP> inv = std::dynamic_pointer_cast<minifi::processors::InvokeHTTP>(proc);
```
arpadboda: I don't think this part and the above has any value here.

libminifi/include/FlowController.h
```diff
@@ -304,23 +305,23 @@ class FlowController : public core::controller::ControllerServiceProvider, publi
...
    */
-  virtual int16_t getResponseNodes(std::vector<std::shared_ptr<state::response::ResponseNode>> &metric_vector, uint16_t metricsClass);
+  virtual std::shared_ptr<state::response::ResponseNode> getMetricsNode() const;
+
+  /**
+   * Retrieves root nodes configured to be included in heartbeat
+   * @param includeManifest -- determines if manifest is to be included
+   * @return a list of response nodes
+   */
+  virtual std::vector<std::shared_ptr<state::response::ResponseNode>> getHeartbeatNodes(bool includeManifest) const;
```
arpadboda: Do these really need to be shared?
Do we ever copy the metrics, heartbeat nodes and the manifest?

Please take it as a question, so in case the answer is yes, I'm totally fine with this.
If we don't copy, just generate and send, I would prefer using unique ptrs to have a clear indication of ownership.

libminifi/include/FlowController.h
```diff
@@ -406,22 +409,27 @@ class FlowController : public core::controller::ControllerServiceProvider, publi
 
   std::chrono::steady_clock::time_point start_time_;
 
-  std::mutex metrics_mutex_;
+  mutable std::mutex metrics_mutex_;
```
arpadboda: 👍 for this and making those functions const.

libminifi/include/FlowController.h
```diff
@@ -406,22 +409,27 @@ class FlowController : public core::controller::ControllerServiceProvider, publi
...
   std::map<uint8_t, std::vector<std::shared_ptr<state::response::ResponseNode>>> component_metrics_by_id_;
+
   // metrics last run
   std::chrono::steady_clock::time_point last_metrics_capture_;
 
  private:
   std::shared_ptr<logging::Logger> logger_;
   std::string serial_number_;
+
+  std::shared_ptr<state::UpdateController> c2_agent_;
```
arpadboda: As far as I see the code, the c2 agent is created within this class and no reference (shared pointer instance) is passed anywhere, so this could be a simple unique ptr to indicate that flowcontroller owns the agent.
Feel free to fix me if I missed something in this aspect.

libminifi/include/c2/C2Agent.h
```diff
@@ -171,13 +159,20 @@ class C2Agent : public state::UpdateController, public state::response::Response
...
   std::timed_mutex metrics_mutex_;
   std::map<std::string, std::shared_ptr<state::response::ResponseNode>> metrics_map_;
 
   /**
-     * Device information stored in the metrics format
-     */
-    std::map<std::string, std::shared_ptr<state::response::ResponseNode>> root_response_nodes_;
+   * Device information stored in the metrics format
+   */
+  std::map<std::string, std::shared_ptr<state::response::ResponseNode>> root_response_nodes_;
```
arpadboda: Wonder if we could store unique ptrs in this as well? I know this is old code, so in case it has a bigger impact, I'm ok with a follow-up as well.

libminifi/include/c2/C2Agent.h
```diff
@@ -244,6 +239,10 @@ class C2Agent : public state::UpdateController, public state::response::Response
   std::string bin_location_;
 
   std::shared_ptr<logging::Logger> logger_;
+
+  utils::ThreadPool<state::Update> heartbeat_thread_pool_;
```
arpadboda: I'm afraid it's going to conflict with my threadpool PR, which is about having one single threadpool.
In case that gets merged before this, let's have a talk to fit this in that architecture.

libminifi/include/core/state/nodes/MetricsBase.h
```diff
@@ -208,26 +208,28 @@ class NodeReporter {
   virtual ~NodeReporter() {
   }
 
+  //This function will be removed when we remove TreeUpdateListener class
```
arpadboda: 👍

libminifi/include/FlowController.h
szaszm: I think `root_response_nodes_` owns the objects and this function shares this ownership with the callers. Whether this shared ownership is needed or not is not clear to me.
We can definitely not use `unique_ptr`s for the returned pointers here, as ownership belongs to `root_response_nodes_`, but we could maybe use `state::response::ResponseNode*` (i.e. observer ptr) if the lifetime of the nodes don't necessitate shared ownership.

Additionally, `root_response_nodes_` is protected, which means it's part of our API and contains `shared_ptr`s, so we're not going to get `unique_ptr`s there.

msharee9: We can definitely have raw pointers in root_response_nodes_ but is there a reason not to use shared pointers?
I understand we cannot use unique_ptrs here as we want to have the FlowController maintain ownership of the nodes. But, given the choices of smart pointers we have shared_ptr is the best one we can use in this case. Having said that, I am not a strong adherent of using shared_ptr universally. In this case I think it is more of a coding guidelines/style/consistency in the code base.

libminifi/include/FlowController.h
msharee9: You have a valid point. Made it a unique_ptr

extensions/http-curl/tests/HTTPHandlers.h
```diff
@@ -343,4 +345,104 @@ class DeleteTransactionResponder : public CivetHandler {
...
+    char buffer[1024];
+    while ((readBytes = mg_read(conn, buffer, blockSize)) > 0) {
+      response.append(buffer, 0, (readBytes / sizeof(char)));
+    }
+    return response;
+  }
+
+  void sendStopOperation(struct mg_connection *conn) {
+    std::string resp = "{\"operation\" : \"heartbeat\", \"requested_operations\" : [{ \"operationid\" : 41, \"operation\" : \"stop\", \"operand\" : \"invoke\"  }, "
+        "{ \"operationid\" : 42, \"operation\" : \"stop\", \"operand\" : \"FlowController\"  } ]}";
```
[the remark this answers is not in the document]
msharee9: I like the neatness of using string literals however, I would rather build a json with some json library and serialize it. But I would keep it as it is for the test code. If we ever want to reuse this response string elsewhere, I would be happy to refactor this.

extensions/http-curl/tests/HTTPHandlers.h
```diff
@@ -343,4 +345,104 @@ class DeleteTransactionResponder : public CivetHandler {
...
+          found = true;
+        }
+
+      }
+    }
+    assert(found == true);
+  }
+
+  virtual void handleHeartbeat(const rapidjson::Document& root, struct mg_connection * conn) {
+    (void)conn;
```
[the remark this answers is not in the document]
msharee9: Good point. Will remove the variable name from the argument list.

extensions/http-curl/tests/HTTPHandlers.h
msharee9: Not sure why, but makes sense to simplify this. We are just reading 1024 bytes from the connection in each call. The variable blockSize is extraneous.

nanofi/include/cxx/C2CallbackAgent.h
```diff
@@ -47,7 +47,10 @@ class C2CallbackAgent : public c2::C2Agent {
 
  public:
 
-  explicit C2CallbackAgent(const std::shared_ptr<core::controller::ControllerServiceProvider> &controller, const std::shared_ptr<state::StateMonitor> &updateSink, const std::shared_ptr<Configure> &configure);
+  explicit C2CallbackAgent(const std::shared_ptr<core::controller::ControllerServiceProvider> &controller,
```
msharee9: This class is not used anywhere in the code today. It will be removed later when nanofi c2 work gets completed.

libminifi/src/FlowController.cpp
```diff
@@ -173,9 +173,9 @@ void FlowController::initializePaths(const std::string &adjustedFilename) {
 
 FlowController::~FlowController() {
   stop(true);
+  c2_agent_->stop();
```
arpadboda: Are you sure we can  avoid the nullcheck here?
I don't think we have C2 agent in case c2 is not configured in the minifi.properties file

libminifi/src/c2/C2Agent.cpp
```diff
@@ -342,6 +342,31 @@ void C2Agent::serializeMetrics(C2Payload &metric_payload, const std::string &nam
   }
 }
 
+/*
+void C2Agent::serializeMetrics(C2Payload &metric_payload, const std::string &name, const std::vector<state::response::SerializedResponseNode> &metrics, bool is_container, bool is_collapsible) {
```
arpadboda: What about this, why comment?

libminifi/src/FlowController.cpp
msharee9: No we cannot avoid null check here. Fixed it in latest commit

libminifi/src/c2/C2Agent.cpp
msharee9: Committed that accidentally. Removed the comments in latest commit.

libminifi/src/FlowController.cpp
```diff
@@ -172,6 +173,9 @@ void FlowController::initializePaths(const std::string &adjustedFilename) {
 
 FlowController::~FlowController() {
   stop(true);
+  if (c2_agent_)
+    c2_agent_->stop();
+  thread_pool_.shutdown();
```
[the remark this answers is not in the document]
msharee9: We do stop the schedulers before stopping threadpool here. Look at line # 175.

libminifi/src/FlowController.cpp
```diff
@@ -907,29 +908,44 @@ int16_t FlowController::clearConnection(const std::string &connection) {
...
+  if (!metricsClass.empty()) {
+    const auto citer = component_metrics_.find(metricsClass);
+    if (citer != component_metrics_.end()) {
+      return citer->second;
+    }
+  } else {
+    const auto iter = root_response_nodes_.find("metrics");
+    if (iter != root_response_nodes_.end()) {
+      return iter->second;
+    }
```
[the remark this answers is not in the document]
msharee9: There is no code duplication as far as I see. we are looking up different maps in if and else branches.

libminifi/src/FlowController.cpp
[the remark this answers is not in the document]
szaszm: Ah, the github folding confused me. Thanks for the clarification.

libminifi/src/FlowController.cpp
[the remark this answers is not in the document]
szaszm: Indeed, I missed that.

extensions/http-curl/tests/C2JstackTest.cpp
```diff
@@ -16,152 +16,63 @@
...
-  std::string logs = LogTestController::getInstance().log_output.str();
-  #ifndef WIN32
-  assert(logs.find("SchedulingAgent") != std::string::npos);
-  #endif
-  LogTestController::getInstance().reset();
-  assert(h_ex.calls_ <= (milliseconds / 1000) + 1);
+  harness.run(test_file_location);
 
-  return 0;
 }
```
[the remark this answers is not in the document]
arpadboda: I guess that's a timing issue: in case the c2 component is initialized before the static id generator, it can crash.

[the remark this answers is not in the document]
msharee9: That is interesting. Never saw the crash when I ran tests. However, I see Arpads point, it might be an issue of initialization order. I think instead of having a static pointer pointing to a singleton in class CoreComponent, we could just call out the static function that returns singleton. This way the singleton will be initialized whenever it is used.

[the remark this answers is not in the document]
bakaid: This is very weird. You do not have to initialize the IdGenerator.
That is, `utils::IdGenerator::getIdGenerator()` is a singleton factory function: the first time you call it, it creates a function static IdGenerator and returns it. This thread safe, protected by the function static initialization.

After this, if you want (and in normal usage, you want), you can call `initialize` on the IdGenerator, supplying it properties, so that it can configure what method should be used for ID generation. This is not a thread safe function, and is called in both `MiNiFiMain.cpp` and in `TestBase` as one of the first things, before other threads are started.

However, this is not a required step: if you don't call `initialize` it will just default to the time-based UUID implementation.

`uuid_impl_`, which on non-Windows platforms is a handle of the third party UUID implementation we use (OSSP UUID), is created in the `IdGenerator` constructor, so there should be no way for it to be nullptr (unless of course its constructor throws, or we get a bad_alloc, but we should have seen an exception in this case).

`CoreComponent` uses the IdGenerator properly, in `libminifi/include/core/Core.h:166` it calls `utils::IdGenerator::getIdGenerator()` as it should, it does not store a static pointer, as far as I can see, @msharee9

I could not reproduce this on macOS, so I am going to try to reproduce it on Linux.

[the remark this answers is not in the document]
szaszm: The above happened on Ubuntu 18.04 with GCC 7.5 (system-level, from `build-essential`). It hangs when being run through `ctest` and segfaults when run with `gdb`.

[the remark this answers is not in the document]
msharee9: @bakaid The CoreComponent before this change had a static pointer to utils::IdGenerator.
And thats where the problem was. The call to generate(uuid_) could well happen before the idgenerator_ was initialized.

msharee9:
> `C2FailedUpdateTest` fails for me consistently on macOS, and also seems to fail on Travis.
> Also, the C2JstackTest never seems to return on Travis, that's why jobs are timing out (I could not reproduce this locally).
> I am going to review the review fixes next, but these test failures must be addressed.

It fails because we have an unnecessary and out of context assert to test the number of calls which somehow depends on the number of times the client POSTs to the CivetServer.
assert(h_ex.calls_ <= (some time duration) + 1);
I think that line came as a copy/paste from some other tests. It is certainly not needed here and also does not test anything related to the C2 failed update itself.

The number of calls to handlePost should not be tested against some arbitrary amount of time duration. It does not make sense at all.

Also I created a follow up ticket to cleanup the C2 Update tests. There is a lot of code duplication between some C2 update tests.

https://issues.apache.org/jira/browse/MINIFICPP-1183

extensions/http-curl/tests/C2JstackTest.cpp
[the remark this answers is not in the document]
bakaid: @msharee9 you are right, the proper calls to the IdGenerator was added in this PR. Even with the previous version though, `CoreComponent::id_generator_` was statically initialized (see the original `Core.cpp`, so an issue with it not yet being initialized could only have occured if some other static initialization tried to use. Even then, it should have failed much sooner, not at the underlying uuid_impl_ point: the `id_generator_` shared_ptr itself would have been nullptr.

Anyway, unfortunately I was able to reproduce the issue on Linux with your latest version. The IdGenerator (or the underlying uuid_impl_ ) is not nullptr, rather, a substructure of the uuid object, the md5 calculator seems to be changed/overwritten at some time:
```
Thread 58 "C2JstackTest" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff4f07700 (LWP 39351)]
MD5Update (context=0x21, input=0x7ffff4f06040 "\354\004u^", inputLen=32) at uuid_md5.c:176
176	    idx = (unsigned int)((context->count[0] >> 3) & 0x3F);
(gdb) bt
#0  MD5Update (context=0x21, input=0x7ffff4f06040 "\354\004u^", inputLen=32) at uuid_md5.c:176
#1  0x0000555555952abe in uuid_md5_update (md5=<optimized out>, data_ptr=data_ptr@entry=0x7ffff4f06040, data_len=data_len@entry=32) at uuid_md5.c:398
#2  0x00005555559537a3 in uuid_prng_data (prng=0x555555df8c80, data_ptr=data_ptr@entry=0x7ffff4f060c8, data_len=data_len@entry=2) at uuid_prng.c:164
#3  0x0000555555951e12 in uuid_make_v1 (ap=<optimized out>, mode=<optimized out>, uuid=<optimized out>) at uuid.c:952
#4  uuid_make (uuid=0x555555df8030, mode=1) at uuid.c:1181
#5  0x000055555595037a in uuid::make(unsigned int, ...) ()
#6  0x00005555557dd0f1 in org::apache::nifi::minifi::utils::IdGenerator::generateWithUuidImpl (this=0x555555df5e70, mode=<optimized out>, output=0x7ffff4f06300 "\006") at /home/bakaid/nifi-minifi-cpp/libminifi/src/utils/Id.cpp:286
#7  0x00005555557dd255 in org::apache::nifi::minifi::utils::IdGenerator::generate (this=<optimized out>, ident=...) at /home/bakaid/nifi-minifi-cpp/libminifi/src/utils/Id.cpp:329
#8  0x0000555555770084 in org::apache::nifi::minifi::core::CoreComponent::CoreComponent (name=..., this=0x7ffff4f06430) at /home/bakaid/nifi-minifi-cpp/libminifi/include/core/Core.h:166
#9  org::apache::nifi::minifi::core::Connectable::Connectable (this=0x7ffff4f06430, name=...) at /home/bakaid/nifi-minifi-cpp/libminifi/src/core/Connectable.cpp:43
#10 0x0000555555816de8 in org::apache::nifi::minifi::state::response::ResponseNode::ResponseNode (name="status", this=0x7ffff4f06430) at /home/bakaid/nifi-minifi-cpp/libminifi/include/core/state/nodes/../nodes/MetricsBase.h:48
#11 org::apache::nifi::minifi::state::response::DeviceInformation::DeviceInformation (name="status", this=0x7ffff4f06430) at /home/bakaid/nifi-minifi-cpp/libminifi/include/core/state/nodes/../nodes/MetricsBase.h:97
#12 org::apache::nifi::minifi::state::response::StateMonitorNode::StateMonitorNode (name="status", this=0x7ffff4f06430) at /home/bakaid/nifi-minifi-cpp/libminifi/include/core/state/nodes/../nodes/StateMonitor.h:53
#13 org::apache::nifi::minifi::state::response::AgentStatus::AgentStatus (name="status", this=0x7ffff4f06430) at /home/bakaid/nifi-minifi-cpp/libminifi/include/core/state/nodes/AgentInformation.h:418
#14 org::apache::nifi::minifi::state::response::AgentNode::getAgentStatus (this=this@entry=0x555556019640) at /home/bakaid/nifi-minifi-cpp/libminifi/include/core/state/nodes/AgentInformation.h:687
#15 0x000055555581d5bf in org::apache::nifi::minifi::state::response::AgentInformation::serialize (this=0x555556019640) at /home/bakaid/nifi-minifi-cpp/libminifi/include/core/state/nodes/AgentInformation.h:738
#16 0x000055555573ed5e in org::apache::nifi::minifi::c2::C2Agent::performHeartBeat (this=0x5555560199c0) at /home/bakaid/nifi-minifi-cpp/libminifi/src/c2/C2Agent.cpp:314
#17 0x000055555573f14f in org::apache::nifi::minifi::c2::C2Agent::<lambda()>::operator()(void) const (__closure=0x7ffff4f06d10) at /home/bakaid/nifi-minifi-cpp/libminifi/src/c2/C2Agent.cpp:100
#18 0x000055555573f951 in std::function<org::apache::nifi::minifi::utils::TaskRescheduleInfo ()>::operator()() const (this=0x7ffff4f06d10) at /usr/include/c++/7/bits/std_function.h:706
#19 org::apache::nifi::minifi::utils::Worker<org::apache::nifi::minifi::utils::TaskRescheduleInfo>::run (this=0x7ffff4f06ce0) at /home/bakaid/nifi-minifi-cpp/libminifi/include/utils/ThreadPool.h:97
#20 0x00005555557f2ec9 in org::apache::nifi::minifi::utils::ThreadPool<org::apache::nifi::minifi::utils::TaskRescheduleInfo>::run_tasks (this=0x555556012aa8, thread=...)
    at /home/bakaid/nifi-minifi-cpp/libminifi/src/utils/ThreadPool.cpp:49
#21 0x00005555557e3d49 in std::__invoke_impl<void, void (org::apache::nifi::minifi::utils::ThreadPool<org::apache::nifi::minifi::utils::TaskRescheduleInfo>::*&)(std::shared_ptr<org::apache::nifi::minifi::utils::WorkerThread>), org::apache::nifi::minifi::utils::ThreadPool<org::apache::nifi::minifi::utils::TaskRescheduleInfo>*&, std::shared_ptr<org::apache::nifi::minifi::utils::WorkerThread>&> (__t=<optimized out>, __f=<optimized out>)
    at /usr/include/c++/7/bits/invoke.h:73
#22 std::__invoke<void (org::apache::nifi::minifi::utils::ThreadPool<org::apache::nifi::minifi::utils::TaskRescheduleInfo>::*&)(std::shared_ptr<org::apache::nifi::minifi::utils::WorkerThread>), org::apache::nifi::minifi::utils::ThreadPool<org::apache::nifi::minifi::utils::TaskRescheduleInfo>*&, std::shared_ptr<org::apache::nifi::minifi::utils::WorkerThread>&> (__fn=<optimized out>) at /usr/include/c++/7/bits/invoke.h:95
#23 std::_Bind<void (org::apache::nifi::minifi::utils::ThreadPool<org::apache::nifi::minifi::utils::TaskRescheduleInfo>::*(org::apache::nifi::minifi::utils::ThreadPool<org::apache::nifi::minifi::utils::TaskRescheduleInfo>*, std::shared_ptr<org::apache::nifi::minifi::utils::WorkerThread>))(std::shared_ptr<org::apache::nifi::minifi::utils::WorkerThread>)>::__call<void, , 0ul, 1ul>(std::tuple<>&&, std::_Index_tuple<0ul, 1ul>) (__args=..., this=<optimized out>)
    at /usr/include/c++/7/functional:467
#24 std::_Bind<void (org::apache::nifi::minifi::utils::ThreadPool<org::apache::nifi::minifi::utils::TaskRescheduleInfo>::*(org::apache::nifi::minifi::utils::ThreadPool<org::apache::nifi::minifi::utils::TaskRescheduleInfo>*, std::shared_ptr<org::apache::nifi::minifi::utils::WorkerThread>))(std::shared_ptr<org::apache::nifi::minifi::utils::WorkerThread>)>::operator()<, void>() (this=<optimized out>) at /usr/include/c++/7/functional:551
#25 std::_Function_handler<void (), std::_Bind<void (org::apache::nifi::minifi::utils::ThreadPool<org::apache::nifi::minifi::utils::TaskRescheduleInfo>::*(org::apache::nifi::minifi::utils::ThreadPool<org::apache::nifi::minifi::utils::TaskRescheduleInfo>*, std::shared_ptr<org::apache::nifi::minifi::utils::WorkerThread>))(std::shared_ptr<org::apache::nifi::minifi::utils::WorkerThread>)> >::_M_invoke(std::_Any_data const&) (__functor=...)
    at /usr/include/c++/7/bits/std_function.h:316
#26 0x00007ffff74c96ef in execute_native_thread_routine () at ../../../../../src/libstdc++-v3/src/c++11/thread.cc:80
#27 0x00007ffff7bbd6db in start_thread (arg=0x7ffff4f07700) at pthread_create.c:463
#28 0x00007ffff6f2488f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
(gdb) f 6
#6  0x00005555557dd0f1 in org::apache::nifi::minifi::utils::IdGenerator::generateWithUuidImpl (this=0x555555df5e70, mode=<optimized out>, output=0x7ffff4f06300 "\006") at /home/bakaid/nifi-minifi-cpp/libminifi/src/utils/Id.cpp:286
286	    uuid_impl_->make(mode);
(gdb) p *uuid_impl_.get()->ctx
$13 = {obj = {time_low = 3412410872, time_mid = 27348, time_hi_and_version = 490, clock_seq_hi_and_reserved = 33 '!', clock_seq_low = 0 '\000', node = "\000\000\000\000\000"}, prng = 0x555555df8c80, md5 = 0x555555df7e70, sha1 = 0x0,
  mac = "\241\000\000\000\000", time_last = {tv_sec = 93825001311728, tv_usec = 93825001663376}, time_seq = 0}
(gdb) p *uuid_impl_.get()->ctx->prng
$14 = {dev = 1330905088, md5 = 0x21, cnt = 93825001295265}
(gdb) p *uuid_impl_.get()->ctx->prng->md5
Cannot access memory at address 0x21
```
It is nondeterministic, and there are also random crashes and aborts at other points.
valgrind also sometimes finds errors:
```
==38778== HEAP SUMMARY:
==38778==     in use at exit: 1,887,773 bytes in 17,209 blocks
==38778==   total heap usage: 98,643 allocs, 81,434 frees, 16,173,074 bytes allocated
==38778==
==38778== Searching for pointers to 17,209 not-freed blocks
==38778== Checked 60,375,896 bytes
==38778==
==38778== LEAK SUMMARY:
==38778==    definitely lost: 1,088 bytes in 1 blocks
==38778==    indirectly lost: 19,129 bytes in 126 blocks
==38778==      possibly lost: 2,128 bytes in 7 blocks
==38778==    still reachable: 1,865,428 bytes in 17,075 blocks
==38778==                       of which reachable via heuristic:
==38778==                         multipleinheritance: 744 bytes in 1 blocks
==38778==         suppressed: 0 bytes in 0 blocks
==38778== Rerun with --leak-check=full to see details of leaked memory
==38778==
==38778== ERROR SUMMARY: 425 errors from 1 contexts (suppressed: 0 from 0)
==38778==
==38778== 425 errors in context 1 of 1:
==38778== Conditional jump or move depends on uninitialised value(s)
==38778==    at 0x4C32D08: strlen (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==38778==    by 0x2ACC57: length (char_traits.h:320)
==38778==    by 0x2ACC57: std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::basic_string(char const*, std::allocator<char> const&) [clone .isra.565] (basic_string.h:511)
==38778==    by 0x2D5D2B: readPost (HTTPHandlers.h:360)
==38778==    by 0x2D5D2B: HeartbeatHandler::verify(mg_connection*) (HTTPHandlers.h:421)
==38778==    by 0x2D6640: HeartbeatHandler::handlePost(CivetServer*, mg_connection*) (HTTPHandlers.h:438)
==38778==    by 0x4AB8C8: CivetServer::requestHandler(mg_connection*, void*) (CivetServer.cpp:148)
==38778==    by 0x4B9E4C: handle_request (civetweb.c:12704)
==38778==    by 0x4BB4DC: process_new_connection (civetweb.c:15918)
==38778==    by 0x4BB82F: worker_thread_run (civetweb.c:16225)
==38778==    by 0x4BB82F: worker_thread (civetweb.c:16290)
==38778==    by 0x4E436DA: start_thread (pthread_create.c:463)
==38778==    by 0x5B3E88E: clone (clone.S:95)
==38778==  Uninitialised value was created by a stack allocation
==38778==    at 0x2D5CB0: HeartbeatHandler::verify(mg_connection*) (HTTPHandlers.h:420)
==38778==
==38778== ERROR SUMMARY: 425 errors from 1 contexts (suppressed: 0 from 0)
```

The whole thing seems to me like a memory corruption/stack corruption issue. I am investigating it further, but I do not at all find it unlikely that it could be caused by the things we do in the SIGUSR2 signal handlers to get BackTraces.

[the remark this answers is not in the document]
szaszm: This is a different but similar issue, @bakaid . I've verified that `uuid_impl_` was null by inserting an assert above the failing line, after seeing it in gdb with the first reproductions. My posted backtrace comes from code with this modification, that's why I skipped the top frames (as it's about assert and abort).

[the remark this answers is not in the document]
bakaid: Thank you, sounds fair, I have occasionally seen it manifest in other ways, so it is reasonable to assume that it occurs in different ways for us based on our environment. And it reinforces my memory corruption hunch.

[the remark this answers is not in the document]
msharee9: @bakaid Your analysis of the crash is spot on. I could also reproduce the C2JstackTest crash on Linux and MacOSX.
The reason behind the crash stems from the memory leak due to shared_ptr cycles. When the test completes the FlowController is never destroyed and that means the thread pool is not shutdown. Therefore, the C2Agent thread or any other thread that accesses the static memory that was released by the main thread causes a segfault.

Solutions: There are different approaches to solve this problem.
1. Do not share the threadpool with C2Agent. Run the C2Agent operations in a separate thread and when we stop FlowController we can optionally stop the C2Agent as well.

2. Optionally stop the threadpool in FlowController::stop method. We accept boolean parameter to determine shutting down the threadpool.
 
3. Expose a public member function FlowController::stopThreadPool that will just shutdown the threadpool.

I personally like option 1, but I would like to create a separate Jira to do that as it will be more extensive work that well surpasses the scope of this PR.

Option 2 is also acceptable but there is a caveat while trying to stop components other than FlowController. It requires more investigation finding its side effects.

Option 1 seems to be a cleaner approach and so effort and time spent for option 2 makes little sense.
In the scope of this PR, I will go with option 3 and here is a follow up Jira that will implement Option 1 and get rid of the extra method we have to add here.

Follow up: https://issues.apache.org/jira/browse/MINIFICPP-1186

libminifi/src/c2/C2Agent.cpp
```diff
@@ -32,24 +32,34 @@
...
+
 namespace org {
 namespace apache {
 namespace nifi {
 namespace minifi {
 namespace c2 {
 
-C2Agent::C2Agent(const std::shared_ptr<core::controller::ControllerServiceProvider> &controller, const std::shared_ptr<state::StateMonitor> &updateSink,
-                 const std::shared_ptr<Configure> &configuration)
+std::shared_ptr<utils::IdGenerator> C2Agent::id_generator_ = utils::IdGenerator::getIdGenerator();
```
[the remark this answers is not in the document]
msharee9: Yep, that remained to be fixed. I forgot to make correction here. There is one more in FlowController.cpp.

extensions/http-curl/tests/C2JstackTest.cpp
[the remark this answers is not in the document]
bakaid: @msharee9 Currently on master C2 has its own thread pool (StateManager's `listener_thread_pool_`), which is separate from the processor thread pool.
As I wrote in an another review comment, using the same thread pool for both C2 and processors is not a good idea.
Merging this PR with the unified thread pool would actually make things worse compared to master, so **I am -1 on this PR** until this is changed.
Therefore, because the thread pool separation should be done anyway, please solve this issue with `Option 1`.

nanofi/include/cxx/C2CallbackAgent.h
```diff
@@ -70,6 +69,7 @@ class C2CallbackAgent : public c2::C2Agent {
     c2_ag_stop_callback *stop;
 
  private:
+    utils::ThreadPool<utils::TaskRescheduleInfo> thread_pool_;
```
szaszm: Since this class inherits from C2Agent, it should also inherit its `ThreadPool`, so there should be no need for 2 `ThreadPool`s in `C2CallbackAgent`.

Do you think it would make sense to revert C2Agent to take the thread pool as a constructor argument instead of creating it, in the spirit of the "[Tell, Don't Ask](https://www.martinfowler.com/bliki/TellDontAsk.html)" principle, the "[Single-responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle)" (object creation is a reponsibility) and "[Inversion of control](https://en.wikipedia.org/wiki/Inversion_of_control)"? This approach would solve the issue by making it possible to reuse the same thread pool for both classes, while leaving option to use separate thread pools if needed.

msharee9: Oh well, the thread_pool_ shouldn't be here at all. This file requires no changes. I forgot to remove it from the previous commit.

libminifi/src/utils/BackTrace.cpp
```diff
@@ -80,25 +80,21 @@ BackTrace TraceResolver::getBackTrace(std::string thread_name, std::thread::nati
...
     emplace_handler();
-    if (pthread_kill(thread_handle_, SIGUSR2) != 0) {
+    std::unique_lock<std::mutex> ulock(trace_mutex_);
+    if (pthread_kill(thread_handle, SIGUSR2) != 0) {
       return std::move(trace_);
     }
-    sigset_t mask;
-    sigfillset(&mask);
-    sigdelset(&mask, SIGUSR2);
-    sigsuspend(&mask);
```
msharee9, line 101: There was a race condition here which lead to a deadlock.
Caller thread (Thread that is interested in pulling backtrace of other thread) acquired lock on a mutex (worker_queue_mutex_), that is outside this function and sent a SIGUSR2 signal to another thread (thread_handle_) but before the caller thread goes into a suspended state or may be even before it removes the SIGUSR2 from its blocking signal set (sigdelset) , the callee while executing its thread handler, sent a SIGUSR2 signal back to the caller and at this time the caller thread will run through its signal handler and after returning from signal handler it goes back to suspended state. From this state there is no way of waking it up.
The callee was waiting on a condition variable tied to the mutex caller was holding thus creating a deadlock.

bakaid: Nice catch.

libminifi/include/utils/BackTrace.h
```diff
@@ -81,13 +83,32 @@ void pull_trace(uint8_t frames_to_skip = 1);
  */
 void emplace_handler();
 
+class Lock {
```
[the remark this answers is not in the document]
msharee9: std::unique_lock could be used here. Removed the custom Lock class. my bad.

libminifi/src/c2/C2Agent.cpp
```diff
@@ -89,44 +93,68 @@ C2Agent::C2Agent(const std::shared_ptr<core::controller::ControllerServiceProvid
...
   c2_consumer_ = [&]() {
     auto now = std::chrono::steady_clock::now();
     if ( queue_mutex.try_lock_until(now + std::chrono::seconds(1)) ) {
-      if (responses.size() > 0) {
-        const C2Payload payload(std::move(responses.back()));
-        responses.pop_back();
-        extractPayload(std::move(payload));
+      if (responses.empty()) {
+        queue_mutex.unlock();
+        return utils::TaskRescheduleInfo::RetryImmediately();
```
arpadboda, line 117: I guess in this case we constantly keep acquiring and releasing a lock in a threadpool worker, ending in a busy wait.
Greta dislikes heating using cpu. :)

msharee9: > @msharee9 Tried lightweight and normal heartbeat and different flows, they seem to work fine.
> 
> However, even when running with no flow, but C2 enabled, minifi consumes 150% CPU both on macOS and Linux.
> 
> This is very likely caused by the issue pointed out by Arpad, that is, we are busy looping because of an incorrect usage of the Thread Pool.
> 
> This callgrind profiling of minifi also seems to reinforce that suspicion:
> Screenshot 2020-04-14 at 17 21 42

Changed it to Retry in 100 ms. I guess then the RetryImmediately() cannot be used anywhere. We should either remove it altogether or have some delay before we call continue at this line.
https://github.com/apache/nifi-minifi-cpp/blob/3116eb817a7182486bf3e50b87c3032ac2074031/libminifi/src/utils/ThreadPool.cpp#L53

arpadboda: RetryImmediately was introduced for a reason: when something takes quite long (an ontrigger call of a processor for eg), it might happen that the next schedule is already in the past. Although we cannot continue executing it as that would result in starvation. 
In this case it's fair to reschedule immediately. 
This means that the given task is put back to the end of the queue, so in case there are tasks already waiting for CPU, those are executed first, otherwise this one can be continued immediately without unnecessary wait.

extensions/http-curl/tests/HTTPIntegrationBase.h
```diff
@@ -91,4 +87,55 @@ void CoapIntegrationBase::setUrl(std::string url, CivetHandler *handler) {
...
+  explicit VerifyC2Base(bool isSecure)
+      : isSecure(isSecure) {
+  }
+
+  virtual void testSetup() {
+    LogTestController::getInstance().setDebug<utils::HTTPClient>();
+    LogTestController::getInstance().setDebug<LogTestController>();
+  }
+
+  virtual void queryRootProcessGroup(std::shared_ptr<core::ProcessGroup> pg) {
```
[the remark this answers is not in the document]
arpadboda: And missing override as well.

libminifi/src/c2/C2Agent.cpp
```diff
@@ -89,44 +93,68 @@ C2Agent::C2Agent(const std::shared_ptr<core::controller::ControllerServiceProvid
...
   c2_consumer_ = [&]() {
     auto now = std::chrono::steady_clock::now();
     if ( queue_mutex.try_lock_until(now + std::chrono::seconds(1)) ) {
-      if (responses.size() > 0) {
-        const C2Payload payload(std::move(responses.back()));
-        responses.pop_back();
-        extractPayload(std::move(payload));
+      if (responses.empty()) {
+        queue_mutex.unlock();
+        return utils::TaskRescheduleInfo::RetryIn(std::chrono::milliseconds(100));
```
arpadboda, line 117: Even if it's not configurable, please define this as a const variable somewhere instead of duplicating the magic number.

libminifi/src/c2/C2Agent.cpp
```diff
@@ -89,44 +93,68 @@ C2Agent::C2Agent(const std::shared_ptr<core::controller::ControllerServiceProvid
...
   functions_.push_back(c2_producer_);
 
   c2_consumer_ = [&]() {
     auto now = std::chrono::steady_clock::now();
     if ( queue_mutex.try_lock_until(now + std::chrono::seconds(1)) ) {
-      if (responses.size() > 0) {
-        const C2Payload payload(std::move(responses.back()));
-        responses.pop_back();
-        extractPayload(std::move(payload));
+      if (responses.empty()) {
```
arpadboda, line 115: I don't get why the structure of this function was changed, but now we need to unlock the mutex in two different places, which is definitely error-prone.

libminifi/include/c2/C2Agent.h
```diff
@@ -52,39 +55,28 @@ namespace c2 {
...
  */
-class C2Agent : public state::UpdateController, public state::response::ResponseNodeSink, public std::enable_shared_from_this<C2Agent> {
+class C2Agent : public state::UpdateController {
  public:
 
-  C2Agent(const std::shared_ptr<core::controller::ControllerServiceProvider> &controller, const std::shared_ptr<state::StateMonitor> &updateSink, const std::shared_ptr<Configure> &configure);
+  C2Agent(const std::shared_ptr<core::controller::ControllerServiceProvider> &controller,
+          const std::shared_ptr<state::StateMonitor> &updateSink,
+          const std::shared_ptr<Configure> &configure,
+          utils::ThreadPool<utils::TaskRescheduleInfo> &pool);
```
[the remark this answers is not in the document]
msharee9: Used a separate threadpool

libminifi/src/c2/C2Agent.cpp
msharee9: Defined a const.

libminifi/src/c2/C2Agent.cpp
msharee9: It had to change because of a potential deadlock situation there.
The deadlock happens as follows:
Thread performing heartbeat acquires request_mutex and while holding lock on this mutex acquires lock on queue_mutex in enqueue_c2_server_response
Thread handling C2 tries to acquire both locks in the reverse order creating a deadlock.

PR #153, opened 2017-10-25
MINIFICPP-263: Move merge content to an extension. Also remove circul…

phrocker:
…ar dependencies

MINIFICPP-263: Remove boost deps by updating yaml-cpp

MINIFICPP-263: Resolve issues on u14

Thank you for submitting a contribution to Apache NiFi - MiNiFi C++.

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [ ] Is there a JIRA ticket associated with this PR? Is it referenced
     in the commit message?

- [ ] Does your PR title start with MINIFI-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.

- [ ] Has your PR been rebased against the latest commit within the target branch (typically master)?

- [ ] Is your initial contribution a single, squashed commit?

### For code changes:
- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?
- [ ] If applicable, have you updated the LICENSE file?
- [ ] If applicable, have you updated the NOTICE file?

### For documentation related changes:
- [ ] Have you ensured that format looks appropriate for the output in which it is rendered?

### Note:
Please ensure that once the PR is submitted, you check travis-ci for build issues and submit an update to your PR as soon as possible.

PR #746, opened 2020-03-23
MINIFICPP-1185 - Remove moodycamel::concurrentqueue from threadpool

arpadboda:

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/include/utils/ConcurrentQueue.h
```diff
@@ -0,0 +1,155 @@
...
+  
+  template <typename... Args>
+  void enqueue(Args&&... args) {
+    ConcurrentQueue<T>::enqueue(std::forward<Args>(args)...);
+    if (running_) {
+      cv_.notify_one();
+    }
+  }
+  
+  bool tryDequeue(T& out) override {
```
bakaid: `try_deque` in `moodycamel` meant that it won't block if there are no elements available, it will return false.
`tryDequeue` in `ConcurrentQueue` means that it won't block if there are no elements available, it will return false.
Yet, in `ConditionConcurrentQueue` `tryDequeue` means that it will block if there are no elements available. This is pretty confusing.
Also, I don't see much value of being able to access the `ConditionConcurrentQueue` through the `ConcurrentQueue` interface.
This function should be `dequeue`, and either it should not be the descendant of `ConcurrentQueue`, or it is also perfectly okay to have a `tryDequeue` as well, which is not overridden.

libminifi/include/utils/ConcurrentQueue.h
```diff
@@ -0,0 +1,155 @@
...
+    std::unique_lock<std::mutex> lck(this->mtx_);
+    if (running_ && this->queue_.empty()) {
```
bakaid: This `if` is unnecessary, `cv.wait` will first check the predicate, and if it is true, it will return immediately without waiting.

libminifi/include/utils/ConcurrentQueue.h
```diff
@@ -0,0 +1,155 @@
...
+    running_ = false;
+    cv_.notify_all();
+  }
+
+  void start() {
+    std::lock_guard<std::mutex> guard(this->mtx_);
+    if (!running_) {
+      running_ = true;
+      if (!this->queue_.empty()) {
+	cv_.notify_all();
```
bakaid: Indentation is bad.

libminifi/include/utils/ConcurrentQueue.h
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+ private:
+   ConcurrentQueue(ConcurrentQueue&& other, std::lock_guard<std::mutex>&)
+    : queue_( std::move(other.queue_) ) {}
+
+ protected:
+  bool tryDequeue(std::unique_lock<std::mutex>& lck, T& out) {
+    if (!lck.owns_lock()) {
+      return false;
```
bakaid: This can only happen if we messed up our code, or someone wrote a derived class and messed up the code.
This is not a normal "we do not have elements" condition, this is an assertion error. I would be more happy to see an exception here to make sure it does not go unnoticed if it occurs.

libminifi/src/utils/ThreadPool.cpp
```diff
@@ -64,8 +64,10 @@ void ThreadPool<T>::run_tasks(std::shared_ptr<WorkerThread> thread) {
         }
       }
     } else {
-      std::unique_lock<std::mutex> lock(worker_queue_mutex_);
-      tasks_available_.wait(lock);
+      // This means that the threadpool is running, but the ConcurrentQueue is stopped -> shouldn't happen
+      if (running_.load()) {
+        std::this_thread::sleep_for(std::chrono::milliseconds(0));
```
bakaid: If this shouldn't happen, how sleeping for 0 milliseconds will help?

libminifi/include/utils/ThreadPool.h
```diff
@@ -330,20 +331,18 @@ class ThreadPool {
...
-  moodycamel::ConcurrentQueue<std::shared_ptr<WorkerThread>> deceased_thread_queue_;
+  ConcurrentQueue<std::shared_ptr<WorkerThread>> deceased_thread_queue_;
 // worker queue of worker objects
-  moodycamel::ConcurrentQueue<Worker<T>> worker_queue_;
+  ConditionConcurrentQueue<Worker<T>> worker_queue_;
   std::priority_queue<Worker<T>, std::vector<Worker<T>>, DelayedTaskComparator<T>> delayed_worker_queue_;
-// notification for available work
-  std::condition_variable tasks_available_;
+// mutex to  protect task status and delayed queue   
+  std::mutex worker_queue_mutex_;
```
szaszm: Is this mutex still used?

libminifi/include/utils/ConcurrentQueue.h
```diff
@@ -0,0 +1,155 @@
...
+    return running_ && ConcurrentQueue<T>::tryDequeue(lck, out);
+  }
+  
+  void stop() {
+    std::lock_guard<std::mutex> guard(this->mtx_);
+    running_ = false;
+    cv_.notify_all();
+  }
+
+  void start() {
```
szaszm: What does `start` mean? Why do I have to start my queue after creating it? Same for `isRunning`. If it serves some purpose, we need a comment explaining it.

libminifi/include/utils/ConcurrentQueue.h
szaszm: If we are in an unknown state, crash and restart by the service manager is better than throwing and/or logging something and then continuing somewhere somehow, IMHO.

libminifi/include/utils/ThreadPool.h
```diff
@@ -303,8 +303,9 @@ class ThreadPool {
    * Drain will notify tasks to stop following notification
    */
   void drain() {
+    worker_queue_.stop();
     while (current_workers_ > 0) {
-      tasks_available_.notify_one();
+      std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
szaszm: What is the purpose of this sleep?

bakaid: In my understanding, it is to wait for all worker thread functions to return. They are either in a task or waiting for one. If they are waiting for one, worker_queue_.stop() will make the tryDeqeue return with false and they will end. If they are in a task, they will finish executing it, call tryDeqeue, which will return false, and they will end.

My initial reaction at first was "why are we not using the ConcurrentQueue's cv instead of "polling" but I realized that is more complex than that.

libminifi/include/utils/ConcurrentQueue.h
arpadboda: Renamed to "dequeue", tryDequeue is also available as a non-blocking alternative.

libminifi/include/utils/ConcurrentQueue.h
arpadboda: Thanks, fixed

libminifi/include/utils/ConcurrentQueue.h
arpadboda: Agreed, logic error is thrown here.

libminifi/include/utils/ThreadPool.h
arpadboda: Yes, still used to protect the delayed queue and the status of current tasks.

libminifi/include/utils/ConcurrentQueue.h
arpadboda: You can construct in started state.

Running means consumers are allowed to wait for notifications (blocking calls). 
Stopping prevents it and wakes them up to make sure no threads hang when no further data is expected to arrive. 

Added a comment to the bool member to explain the purpose.

libminifi/include/utils/ThreadPool.h
arpadboda: @bakaid is right. 
Stopping the queue wakes up all the worker threads actually doing nothing (waiting for work), but there can be worker threads doing some useful work (ontrigger calls, c2 heartbeats, whatever). These should end in a timely manner, but I found no better option to wait for that.

libminifi/src/utils/ThreadPool.cpp
arpadboda: Meant to sleep for 1 msec, corrected that. 

In case this somehow occurs in a transient state (during startup or shutdown), this prevent the worker threads to busy wait (dequeue fails immediately and the workers would retry without sleep).

libminifi/include/utils/ConcurrentQueue.h
arpadboda: Unfortunately not. 
```wait causes the current thread to block until the condition variable is notified or a spurious wakeup occurs, optionally looping until some predicate is satisfied.```
I think this means the predicate can only be used to wait *again* for further notification in case it's not fulfilled (in case of spurious wakeup for eg), but cannot avoid waiting. 

I gave it a try, but removing the if significantly changed the behaviour, calls block even if there is are tasks in the queue.

File: libminifi/include/utils/ConcurrentQueue.h [hunk 9]
```diff
@@ -0,0 +1,155 @@
...
+    out = std::move(queue_.front());
+    queue_.pop_front();
+    return true;
+  }
+  std::deque<T> queue_;
+  mutable std::mutex mtx_;
+};
+
+template <typename T>
+class ConditionConcurrentQueue : public ConcurrentQueue<T> {
```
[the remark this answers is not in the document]
arpadboda: When I was reading the private inheritance idea, my first though was "I love it". 

Although I realised that private inheritance would made me rewrap or copy/paste all the functions (size(), clear()) I would like to inherit from the base class without any modification on them. 

In case it can somehow be avoided (I found no way, but naturally open to suggestions), I absolutely favour the private inheritance. Otherwise - to avoid code duplication or needless wrappers - I would stay with the current implementation. 

Virtual functions are removed in the meanwhile to address that part of your concerns.

libminifi/include/utils/ConcurrentQueue.h
bakaid: As stated in https://en.cppreference.com/w/cpp/thread/condition_variable/wait the wait with predicate it is equivalent to
```
while (!pred()) {
    wait(lock);
}
```

And it is really implemented that way: https://github.com/gcc-mirror/gcc/blob/b7c9bd36eaacac42631b882dc67a6f0db94de21c/libstdc%2B%2B-v3/include/std/condition_variable#L98

If removing the if causes an issue, then there is a real issue in the code. I will review.

arpadboda: I removed first and tests started to fail, so I reverted it. 

Now I added some debug prints there to see what's going on, removed the ifs and everything worked. 
It seems to work now properly, not sure what happened with my build.

arpadboda:
> Compiled and tested the branch on macOS.
> Also rebased MINIFICPP-550 to this branch and tried whether it fixes the C2 task starvation problem that made me find this issue - it does.
> @arpadboda What I would really like to see are at least some minimal positive tests for the new `ConcurrentQueue` and `ConditionConcurrentQueue`. We are replacing a known (bad for our purpose, but still tested) implementation with an untested one. There are some minimal ThreadPool tests and indirectly some tests depend on this, but that does not make finding errors easy, so I would really prefer some unit tests here.

Absuletly agree, already working on some. 
Wanted to make sure it works well on all the platforms first (and CI passes).

libminifi/include/utils/ConcurrentQueue.h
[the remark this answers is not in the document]
szaszm: In the case of any "implemented in terms of" relationship, the new class would need to provide all of the required functionality and delegate to the inner class. This means boilerplate, as you pointed out, regardless of whether it's private inheritance or composition.

One way to reduce boilerplate in the case of private inheritance and unmodified member functions is via using-declarations. It still requires iterating the member functions that are to be exposed without modification, but they are shorter than providing wrapper implementations.

In my opinion, cleaner dependencies (in this case) add more value than avoided boilerplate, but this is subjective.

libminifi/include/utils/ThreadPool.h
szaszm: Could you add a code comment explaining this, including the fact that this is a best-effort solution? It will be helpful to future readers of the code.

libminifi/include/utils/ConcurrentQueue.h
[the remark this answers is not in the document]
arpadboda: Changed to private inheritance

libminifi/include/utils/ThreadPool.h
arpadboda: Added

arpadboda:
Added tests

libminifi/test/unit/MinifiConcurrentQueueTests.cpp
```diff
@@ -0,0 +1,158 @@
...
+#include <string>
+#include <thread>
+#include <vector>
+#include <set>
+
+#include "../TestBase.h"
+#include "utils/MinifiConcurrentQueue.h"
+#include "utils/StringUtils.h"
+
+using namespace org::apache::nifi::minifi::utils;
```
szaszm: You can either remove the `utils::` prefix from the symbol references from this namespace in this file or change this line to `namespace utils = org::apache::nifi::minifi::utils;`. I personally have no preference, but abseil (google) recommends no using-directives at all.

Abseil: "Tip of the Week #153: Don't use using-directives" https://abseil.io/tips/153

The C++ Core Guidelines allow this usage of using-directives:
http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-using

libminifi/test/unit/MinifiConcurrentQueueTests.cpp
```diff
@@ -0,0 +1,158 @@
...
+  utils::ConcurrentQueue<std::string> queue;
+  std::set<std::string> results;
+
+  std::thread producer([&queue]() {
+      queue.enqueue("ba");
+      std::this_thread::sleep_for(std::chrono::milliseconds(3));
+      queue.enqueue("dum");
+      std::this_thread::sleep_for(std::chrono::milliseconds(3));
+      queue.enqueue("tss");
+    });
```
szaszm: there seems to be one extra level of indentation

libminifi/include/utils/ThreadPool.h
arpadboda: Removed

libminifi/include/utils/MinifiConcurrentQueue.h
```diff
@@ -0,0 +1,160 @@
...
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace utils {
+
+template <typename T>
+class ConcurrentQueue {
+ public:    
+  ConcurrentQueue() = default;
+  virtual ~ConcurrentQueue() = default;
```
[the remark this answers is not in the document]
arpadboda: Removed

libminifi/include/utils/MinifiConcurrentQueue.h
```diff
@@ -0,0 +1,160 @@
...
+    return true;
+  }
+  std::deque<T> queue_;
+  mutable std::mutex mtx_;
+};
+
+template <typename T>
+class ConditionConcurrentQueue : private ConcurrentQueue<T> {
+ public:
+  ConditionConcurrentQueue(bool start = false) : ConcurrentQueue<T>(), running_{start} {};
```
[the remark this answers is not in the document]
arpadboda: Fixed them

libminifi/include/utils/MinifiConcurrentQueue.h
```diff
@@ -0,0 +1,160 @@
...
+    }
+    if (queue_.empty()) {
+      return false;
+    }
+    out = std::move(queue_.front());
+    queue_.pop_front();
+    return true;
+  }
+  std::deque<T> queue_;
+  mutable std::mutex mtx_;
```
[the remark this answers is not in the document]
arpadboda: The container is now private, the mutex isn't as you have to acquire that in derived classes as well.

libminifi/include/utils/MinifiConcurrentQueue.h
```diff
@@ -0,0 +1,160 @@
...
+
+  template <typename... Args>
+  void enqueue(Args&&... args) {
+    ConcurrentQueue<T>::enqueue(std::forward<Args>(args)...);
+    if (running_) {
+      cv_.notify_one();
+    }
+  }
+  
+  bool dequeue(T& out) {
```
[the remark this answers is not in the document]
arpadboda: Added wait_for, so now it support:

- tryDequeue
- DequeueWait
- DequeueWaitFor

libminifi/include/utils/MinifiConcurrentQueue.h
```diff
@@ -0,0 +1,160 @@
...
+    out = std::move(queue_.front());
+    queue_.pop_front();
+    return true;
+  }
+  std::deque<T> queue_;
+  mutable std::mutex mtx_;
+};
+
+template <typename T>
+class ConditionConcurrentQueue : private ConcurrentQueue<T> {
```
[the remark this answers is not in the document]
arpadboda: Added comments to explain

libminifi/test/unit/MinifiConcurrentQueueTests.cpp
```diff
@@ -0,0 +1,158 @@
...
+      }
+    }
+  });
+
+  producer.join();
+
+  // Give some time for the consumer to loop over the queue
+  std::this_thread::sleep_for(std::chrono::milliseconds(10));
+
+  consumer.join();
```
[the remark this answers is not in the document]
arpadboda: Good point, removed.

libminifi/src/utils/ThreadPool.cpp
```diff
@@ -64,8 +64,10 @@ void ThreadPool<T>::run_tasks(std::shared_ptr<WorkerThread> thread) {
         }
       }
     } else {
-      std::unique_lock<std::mutex> lock(worker_queue_mutex_);
-      tasks_available_.wait(lock);
+      // This means that the threadpool is running, but the ConcurrentQueue is stopped -> shouldn't happen
+      if (running_.load()) {
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
+      }
```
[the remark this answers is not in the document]
arpadboda: There is no logger here. 
It's not a critical error that worth an exception/dump and it might happen during startup/shutdown for a very short time.

libminifi/test/unit/MinifiConcurrentQueueTests.cpp
arpadboda: Did the 2nd

libminifi/test/unit/MinifiConcurrentQueueTests.cpp
arpadboda: Fixed

libminifi/test/unit/MinifiConcurrentQueueTests.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+using namespace org::apache::nifi::minifi::utils;
+
+TEST_CASE("TestConqurrentQueue::testQueue", "[TestQueue]") {
+  utils::ConcurrentQueue<std::string> queue;
+  std::vector<std::string> results;
+
+  std::thread producer([&queue]() {
+      queue.enqueue("ba");
+      std::this_thread::sleep_for(std::chrono::milliseconds(3));
```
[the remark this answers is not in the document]
arpadboda: In case  there would be no sleeps here, producer most probably just inserts everything before the consumer is even started. 
Which would also make the last testcases useless, where the goal is to prove that elements inserted later (when some are already looped in the queue) also get to a consumer.

libminifi/include/utils/MinifiConcurrentQueue.h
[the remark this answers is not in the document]
szaszm: It's fine and expected that we use it in libminifi. What I'd like is that we only support dependency on public members for libminifi API users and treat `protected`/`private` as "internal". The language doesn't provide tools for that (like package-private in Java), so that's an API documentation issue. We could also use pimpl, but that would mean a lot of boilerplate, an extra level of indirection and no inlining, so I'm against that.

libminifi/include/utils/MinifiConcurrentQueue.h
```diff
@@ -29,11 +29,12 @@ namespace nifi {
...
 namespace utils {
 
+
+// Provides a queue API and guarantees no race conditions in case of multiple producers and consumers.
 template <typename T>
 class ConcurrentQueue {
  public:    
-  ConcurrentQueue() = default;
-  virtual ~ConcurrentQueue() = default;
+  explicit ConcurrentQueue() = default;
```
szaszm: I don't see the need to make this explicit, although it doesn't hurt much. It prevents code like this:
`ConcurrentQueue make_queue() { return {}; }`

libminifi/include/utils/MinifiConcurrentQueue.h
```diff
@@ -29,11 +29,12 @@ namespace nifi {
 namespace minifi {
 namespace utils {
 
+
+// Provides a queue API and guarantees no race conditions in case of multiple producers and consumers.
```
szaszm: I would add:
- Guarantees that the elements are dequeued in the order of insertion

libminifi/include/utils/MinifiConcurrentQueue.h
```diff
@@ -81,25 +82,38 @@ class ConcurrentQueue {
...
+
+  bool tryDequeueImpl(std::unique_lock<std::mutex>& lck, T& out) {
+    checkLock(lck);
     if (queue_.empty()) {
       return false;
     }
     out = std::move(queue_.front());
     queue_.pop_front();
     return true;
   }
```
szaszm: I like that you used protected functions taking `std::unique_lock` to make it an error to use the non-locking functions without a lock.

libminifi/include/utils/MinifiConcurrentQueue.h
```diff
@@ -119,10 +133,22 @@ class ConditionConcurrentQueue : private ConcurrentQueue<T> {
...
   
-  bool dequeue(T& out) {
+  bool dequeueWait(T& out) {
+    std::unique_lock<std::mutex> lck(this->mtx_);
+    cv_.wait(lck, [this, &lck]{ return !running_ || !this->emptyImpl(lck); });  // Only wake up if there is something to return or stopped 
+    return running_ && ConcurrentQueue<T>::tryDequeueImpl(lck, out);
+  }
+
+  template< class Rep, class Period >
+  bool dequeueWaitFor(T& out, const std::chrono::duration<Rep, Period>& time) {
```
szaszm: I like that you used a template here, not just a concrete duration type like `milliseconds`, which is more common in the code base.

libminifi/include/utils/MinifiConcurrentQueue.h
```diff
@@ -0,0 +1,186 @@
...
+    return queue_.empty();
+  }
+
+  mutable std::mutex mtx_;
+ private:
+  std::deque<T> queue_;
+};
+
+
+// A ConcurrentQueue extended with a condition variable to be able to block and wait for incoming data
```
szaszm: I would like to add:
- `stop` interrupts all consumers without a chance to consume remaining elements in the queue
- started means queue elements can be consumed/dequeued. 
- It's possible to enqueue elements regardless of the running state.

libminifi/include/utils/MinifiConcurrentQueue.h
```diff
@@ -0,0 +1,186 @@
...
+    cv_.notify_all();
+  }
+
+  void start() {
+    std::unique_lock<std::mutex> lck(this->mtx_);
+    if (!running_) {
+      running_ = true;
+      if (!this->emptyImpl(lck)) {
+        cv_.notify_all();
+      }
```
szaszm: Since we are transitioning from stopped to started in a critical section, and it's not possible to have consumers in a stopped state, this notification shouldn't be necessary.

libminifi/test/unit/MinifiConcurrentQueueTests.cpp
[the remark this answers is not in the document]
szaszm: I see your point, but we should cover data races as well. Can we have at least a handful of test cases with rapid insertions, i.e. with no sleep, to cover both problems?

libminifi/src/utils/ThreadPool.cpp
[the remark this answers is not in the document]
szaszm: thx for the explanation. Feel free to close the thread.

libminifi/include/utils/MinifiConcurrentQueue.h
arpadboda: Removed

libminifi/include/utils/MinifiConcurrentQueue.h
arpadboda: Added

libminifi/include/utils/MinifiConcurrentQueue.h
arpadboda: Added

libminifi/test/unit/MinifiConcurrentQueueTests.cpp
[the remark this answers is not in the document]
arpadboda: Done, added a new testcases that deals with a lot of elements without sleeps.

libminifi/include/utils/MinifiConcurrentQueue.h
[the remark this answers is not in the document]
arpadboda: I think that topic is way more general than this PR.

I would be happy to a see a doc about the interfaces we consider public API, but that doesn't exist yet.

PR #747, opened 2020-04-01
MINIFICPP-1036 - libarchive extension improvements

bakaid:
This commit
 - transfers the libarchive compilation to an ExternalProject
 - adds bundled liblzma and bzip2
 - makes the libarchive extension compile on Windows
 - adds the "Encapsulate in TAR" Property to CompressContent
 - adds raw GZIP compression support to CompressContent

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/src/io/ZlibStream.cpp
```diff
@@ -0,0 +1,202 @@
...
+      state_ = ZlibStreamState::ERRORED;
+      return -1;
+    }
+  } while (strm_.avail_out == 0);
+
+  return size;
+}
+
+void ZlibCompressStream::closeStream() {
+  if (state_ == ZlibStreamState::INITIALIZED) {
```
arpadboda, line 121: In case error happened during write and close is called, what happens? Or we don't expect it to happen at all?

libminifi/src/io/ZlibStream.cpp
```diff
@@ -0,0 +1,202 @@
...
+  do {
+    logger_->log_trace("writeData has %u B of input data left", strm_.avail_in);
+
+    strm_.next_out = outputBuffer_.data();
+    strm_.avail_out = outputBuffer_.size();
+
+    ret = inflate(&strm_, Z_NO_FLUSH);
+    if (ret == Z_STREAM_ERROR ||
+        ret == Z_NEED_DICT ||
+        ret == Z_DATA_ERROR ||
```
arpadboda, line 176: Bleh, is this API designed so badly?

I would expect a success case or something, but handling some errors that look like a specific set of errors is weird.

libminifi/src/io/ZlibStream.cpp
```diff
@@ -0,0 +1,202 @@
...
+  if (ret != Z_OK) {
+    logger_->log_error("Failed to initialize z_stream with inflateInit2, error code: %d", ret);
+    throw Exception(ExceptionType::GENERAL_EXCEPTION, "zlib inflateInit2 failed");
+  }
+
+  state_ = ZlibStreamState::INITIALIZED;
+}
+
+ZlibDecompressStream::~ZlibDecompressStream() {
+  if (state_ != ZlibStreamState::UNINITIALIZED) {
```
arpadboda, line 147: The same here, uninited state can't happen, although I wonder if inflateEnd should be called in case decompress failed and we have an error state.

libminifi/test/archive-tests/CompressContentTests.cpp
```diff
@@ -932,3 +938,107 @@ TEST_CASE("DecompressFileXYLZMA", "[compressfiletest8]") {
...
+  std::vector<uint8_t> compressed_content((std::istreambuf_iterator<char>(compressed)), std::istreambuf_iterator<char>());
+  REQUIRE(2 < compressed_content.size());
+  // gzip magic number
+  REQUIRE(0x1f == compressed_content[0]);
+  REQUIRE(0x8b == compressed_content[1]);
+
+  // Check decompressed file
+  std::ifstream decompressed(decompressed_file, std::ios::in | std::ios::binary);
+  std::string decompressed_content((std::istreambuf_iterator<char>(decompressed)), std::istreambuf_iterator<char>());
+  REQUIRE(content == decompressed_content);
```
arpadboda, line 1041: Nice test!

libminifi/test/unit/ZlibStreamTests.cpp
```diff
@@ -0,0 +1,105 @@
...
+  SECTION("Empty") {
+  }
+  SECTION("Simple content in one write") {
+    REQUIRE(-1 != compressStream.write(reinterpret_cast<uint8_t*>(const_cast<char*>("foobar")), strlen("foobar")));
+    original += "foobar";
+  }
+  SECTION("Simple content in two writes") {
+    REQUIRE(-1 != compressStream.write(reinterpret_cast<uint8_t*>(const_cast<char*>("foo")), strlen("foo")));
+    original += "foo";
+    REQUIRE(-1 != compressStream.write(reinterpret_cast<uint8_t*>(const_cast<char*>("bar")), strlen("bar")));
```
arpadboda, line 42: I think the return values should match the length of the string. 
Or in case they return the number of compressed bytes written, let's check for grearter than zero.

thirdparty/bzip2/bzip2.patch
```diff
@@ -0,0 +1,31 @@
+diff -rupN orig/CMakeLists.txt patched/CMakeLists.txt
```
arpadboda, line 1: A justification for this file (in a comment or sg) would be nice.

thirdparty/liblzma/liblzma.patch
```diff
@@ -0,0 +1,118 @@
+diff -rupN orig/CMakeLists.txt patched/CMakeLists.txt
+--- orig/CMakeLists.txt	1970-01-01 01:00:00.000000000 +0100
++++ patched/CMakeLists.txt	2020-03-31 21:52:39.000000000 +0200
+@@ -0,0 +1,114 @@
++cmake_minimum_required(VERSION 3.7)
++
++project(liblzma)
++
++set(SOURCES src/common/tuklib_cpucores.c
```
arpadboda, line 9: Do we patch to only compile the subset of files/features we need?

bakaid: We can't run the configure script on Windows, and the only thing lzma provides for Windows are msbuild files which do not integrate well into our CMake build system, so I wrote a windows-specific CMakeLists.txt for the project.

thirdparty/bzip2/bzip2.patch
bakaid: bzip2 only provides a Makefile which is not very usable on Windows, and it is a pretty simple project so I created a CMake file to compile it on all platforms.

libminifi/src/io/ZlibStream.cpp
bakaid: This is how it is designed, see the example usage: https://zlib.net/zlib_how.html

libminifi/src/io/ZlibStream.cpp
```diff
@@ -0,0 +1,202 @@
...
+  if (ret != Z_OK) {
+    logger_->log_error("Failed to initialize z_stream with deflateInit2, error code: %d", ret);
+    throw Exception(ExceptionType::GENERAL_EXCEPTION, "zlib deflateInit2 failed");
+  }
+
+  state_ = ZlibStreamState::INITIALIZED;
+}
+
+ZlibCompressStream::~ZlibCompressStream() {
+  if (state_ != ZlibStreamState::UNINITIALIZED) {
```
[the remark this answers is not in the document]
bakaid: You are right, but I think it is better readable this way.

libminifi/src/io/ZlibStream.cpp
bakaid: It's perfectly okay to call `closeStream` if an error happens, and we will do nothing: we will keep the error state as it was.

libminifi/src/io/ZlibStream.cpp
bakaid: inflateEnd frees the resources allocated by zlib, so if we called inflateInit and it succeeded we must call inflateEnd, even if there as an error.

arpadboda: Okay, thanks, that makes sense.

extensions/libarchive/CompressContent.cpp
```diff
@@ -39,6 +39,11 @@ core::Property CompressContent::CompressLevel("Compression Level", "The compress
 core::Property CompressContent::CompressMode("Mode", "Indicates whether the processor should compress content or decompress content.", MODE_COMPRESS);
 core::Property CompressContent::CompressFormat("Compression Format", "The compression format to use.", COMPRESSION_FORMAT_ATTRIBUTE);
 core::Property CompressContent::UpdateFileName("Update Filename", "Determines if filename extension need to be updated", "false");
+core::Property CompressContent::EncapsulateInTar("Encapsulate in TAR", "If true, on compression the FlowFile is added to a TAR archive and then compressed, "
```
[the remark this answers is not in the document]
bakaid: I completely agree, but because all other Properties used the older API, I didn't want this to stand out.
However, after using this processor through C2, I am convinced that converting all Properties to use validators in this processor will worth the (not significant) effort, so I'll do that.

extensions/libarchive/CompressContent.h
```diff
@@ -341,6 +346,73 @@ class CompressContent: public core::Processor {
...
+            } else if (ret == 0) {
+              break;
+            } else {
+              if (outputStream_->writeData(buffer.data(), ret) != ret) {
+                return -1;
+              }
+              read_size += ret;
+            }
+          }
+          outputStream_->closeStream();
```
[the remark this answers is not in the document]
bakaid: The generic answer is that our Stream API is not very highly (or properly) defined and whatever `closeStream` does, or when it should be called is implementation-defined.
In this particular case we have to call `closeStream` in the successful case, because that's how we signal the end of the input data and force zlib to flush its buffers and write out the end of the compressed stream.

libminifi/include/io/ZlibStream.h
```diff
@@ -0,0 +1,105 @@
...
+ protected:
+  ZlibStreamState state_{ZlibStreamState::UNINITIALIZED};
+  z_stream strm_{};
+  std::vector<uint8_t> outputBuffer_;
+};
+
+class ZlibCompressStream : public ZlibBaseStream {
+ public:
+  ZlibCompressStream(bool gzip = true, int level = Z_DEFAULT_COMPRESSION);
+  ZlibCompressStream(DataStream* other, bool gzip = true, int level = Z_DEFAULT_COMPRESSION);
```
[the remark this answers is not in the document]
bakaid: Fair point. I'll replace the `bool gzip` with an enum class to the effect of `ZlibFormat format`. A pointer is not converted implicitly to an enum class, and it will also improve readability.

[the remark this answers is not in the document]
szaszm: Could you also mark both ctors as `explicit`? Since this is new code, it doesn't break API compatibility.

libminifi/src/io/ZlibStream.cpp
szaszm: Would it make sense to make `state_ != UNINITIALIZED` a postcondition of the ctor and remove the UNINITIALIZED state? It seems to be already the case, so we could remove all possibility of related dead code by removing the enum value. We don't need to prepare for any usage of an object, including destruction, during object creation.

bakaid: Not if you ever want to make this object movable: in that case you would have to set the moved-from object to UNINITIALIZED to prevent it from freeing the z_stream.

libminifi/include/io/ZlibStream.h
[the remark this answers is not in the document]
bakaid: I could, although I don't really see the point in this case.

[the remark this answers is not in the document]
szaszm: To prevent implicit conversion from `bool`/`ZlibFormat` and `DataStream*`, as suggested in [C.46: By default, declare single-argument constructors explicit](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-explicit). (The guideline doesn't, but the reason covers all constructors that can be called with a single argument, i.e. including ctors with more parameters where all N>1-th parameter has a default argument.)

Or if you intended the implicit conversion to be part of the API, I'm also fine with an explanation and possibly an example of intended usage added in a code comment.

libminifi/test/unit/ZlibStreamTests.cpp
bakaid: It must return the number of bytes consumed, will change to an exact check.

libminifi/test/archive-tests/CompressContentTests.cpp
bakaid: Thanks!

extensions/libarchive/CompressContent.h
```diff
@@ -341,6 +346,73 @@ class CompressContent: public core::Processor {
...
+    }
+
+    ~GzipWriteCallback() override = default;
+
+    std::shared_ptr<logging::Logger> logger_;
+    std::string compress_mode_;
+    int64_t compress_level_;
+    std::shared_ptr<core::FlowFile> flow_;
+    std::shared_ptr<core::ProcessSession> session_;
+    bool success_{false};
```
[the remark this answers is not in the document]
bakaid: Changed.

extensions/libarchive/CompressContent.h
```diff
@@ -341,6 +346,73 @@ class CompressContent: public core::Processor {
...
+    GzipWriteCallback(std::string compress_mode, int64_t compress_level, std::shared_ptr<core::FlowFile> flow, std::shared_ptr<core::ProcessSession> session)
+      : logger_(logging::LoggerFactory<CompressContent>::getLogger())
+      , compress_mode_(std::move(compress_mode))
+      , compress_level_(compress_level)
+      , flow_(std::move(flow))
+      , session_(std::move(session))
+    {
+    }
+
+    ~GzipWriteCallback() override = default;
```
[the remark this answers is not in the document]
bakaid: Force of habit, removed.

libminifi/include/io/ZlibStream.h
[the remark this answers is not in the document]
bakaid: Fair, will mark them explicit.

extensions/libarchive/CompressContent.cpp
```diff
@@ -74,7 +80,13 @@ void CompressContent::onSchedule(core::ProcessContext *context, core::ProcessSes
   if (context->getProperty(UpdateFileName.getName(), value) && !value.empty()) {
     org::apache::nifi::minifi::utils::StringUtils::StringToBool(value, updateFileName_);
   }
-  logger_->log_info("Compress Content: Mode [%s] Format [%s] Level [%d] UpdateFileName [%d]", compressMode_, compressFormat_, compressLevel_, updateFileName_);
+  value = "";
+  if (context->getProperty(EncapsulateInTar.getName(), value) && !value.empty()) {
```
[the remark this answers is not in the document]
bakaid: Done.

extensions/libarchive/ArchiveCommon.h
```diff
@@ -0,0 +1,29 @@
...
+ */
+#ifndef EXTENSIONS_LIBARCHIVE_ARCHIVECOMMON_H_
+#define EXTENSIONS_LIBARCHIVE_ARCHIVECOMMON_H_
+
+#include <stdint.h>
+
+#ifdef WIN32
+#define mode_t uint16_t
+#define uid_t int64_t
+#define gid_t int64_t
```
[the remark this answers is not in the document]
bakaid: Not sure, I think I was distracted by some third party header mess when doing this, typedef is the nice solution here, you are right. Changed.

bakaid:
@arpadboda @szaszm Addressed review comments, it also builds for me on Travis now (on Xenial for some unfathomable reason some constexpr string literals in the sftp extension started to get ODR-used and therefore lacked storage and build failed with a link error - even though I haven't even touched the extension. Anyway, added storage for them in the last commit and it builds properly now.)

libminifi/test/archive-tests/CompressContentTests.cpp
```diff
@@ -932,3 +938,107 @@ TEST_CASE("DecompressFileXYLZMA", "[compressfiletest8]") {
...
+    for (size_t i = 0U; i < 1024 * 1024U; i++) {
+      content_ss << "foobar";
+    }
+    content = content_ss.str();
+  }
+
+  std::fstream file;
+  file.open(src_file, std::ios::out);
+  file << content;
+  file.close();
```
szaszm, line 1025: Could be a oneliner: `std::ofstream{src_file} << content;`. Just an idea, I don't mind if it stays like this.

extensions/libarchive/CompressContent.cpp
```diff
@@ -35,10 +35,32 @@ namespace nifi {
...
+    core::PropertyBuilder::createProperty("Mode")->withDescription("Indicates whether the processor should compress content or decompress content.")
+        ->isRequired(false)->withAllowableValues<std::string>({MODE_COMPRESS, MODE_DECOMPRESS})->withDefaultValue(MODE_COMPRESS)->build());
+core::Property CompressContent::CompressFormat(
+    core::PropertyBuilder::createProperty("Compression Format")->withDescription("The compression format to use.")
+        ->isRequired(false)
+        ->withAllowableValues<std::string>({
+          COMPRESSION_FORMAT_ATTRIBUTE,
+          COMPRESSION_FORMAT_GZIP,
+          COMPRESSION_FORMAT_BZIP2,
+          COMPRESSION_FORMAT_XZ_LZMA2,
```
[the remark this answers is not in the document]
bakaid: Could you elaborate? Any YAML config that didn't use values from this enum was erroneous and resulted in an error on onTrigger. How this change would break those configurations further?

libminifi/test/archive-tests/CompressContentTests.cpp
bakaid: This is a test, I copied this from somewhere, I could not really be bothered whether its one line or four.

libminifi/include/io/ZlibStream.h
```diff
@@ -0,0 +1,110 @@
...
+ public:
+  ZlibBaseStream();
+  ZlibBaseStream(DataStream* other);
+
+  ZlibBaseStream(const ZlibBaseStream&) = delete;
+  ZlibBaseStream& operator=(const ZlibBaseStream&) = delete;
+  ZlibBaseStream(ZlibBaseStream&& other) = delete;
+  ZlibBaseStream& operator=(ZlibBaseStream&& other) = delete;
+
+  ~ZlibBaseStream() override = default;
```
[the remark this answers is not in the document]
bakaid: Done.

libminifi/include/io/ZlibStream.h
```diff
@@ -0,0 +1,110 @@
...
+
+class ZlibBaseStream : public BaseStream {
+ public:
+  ZlibBaseStream();
+  ZlibBaseStream(DataStream* other);
+
+  ZlibBaseStream(const ZlibBaseStream&) = delete;
+  ZlibBaseStream& operator=(const ZlibBaseStream&) = delete;
+  ZlibBaseStream(ZlibBaseStream&& other) = delete;
+  ZlibBaseStream& operator=(ZlibBaseStream&& other) = delete;
```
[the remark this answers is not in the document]
bakaid: Done.

bakaid:
Addressed new review comments.
I would appreciate if we could move this PR along.

extensions/libarchive/CompressContent.cpp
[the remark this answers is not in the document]
szaszm: In that case, I was wrong.

PR #748, opened 2020-04-01
MINIFICPP-1187 Implemented.

am-c-p-p:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/windows-event-log/Bookmark.cpp
+++ b/extensions/windows-event-log/Bookmark.cpp
@@ -11,7 +11,7 @@ namespace nifi {
 namespace minifi {
 namespace processors {
 
-Bookmark::Bookmark(const std::wstring& channel, const std::wstring& query, const std::string& bookmarkRootDir, const std::string& uuid, std::shared_ptr<logging::Logger> logger)
+Bookmark::Bookmark(const std::wstring& channel, const std::wstring& query, const std::string& bookmarkRootDir, const std::string& uuid, bool processOldEvents, std::shared_ptr<logging::Logger> logger)
   :logger_(logger) {
   if (!createUUIDDir(bookmarkRootDir, uuid, filePath_))
     return;
@@ -48,7 +48,7 @@ Bookmark::Bookmark(const std::wstring& channel, const std::wstring& query, const
   }
   const utils::ScopeGuard guard_hEventResults([hEventResults]() { EvtClose(hEventResults); });
 
-  if (!EvtSeek(hEventResults, 0, 0, 0, EvtSeekRelativeToLast)) {
+  if (!EvtSeek(hEventResults, 0, 0, 0, processOldEvents? EvtSeekRelativeToFirst : EvtSeekRelativeToLast)) {
     LOG_LAST_ERROR(EvtSeek);
     return;
   }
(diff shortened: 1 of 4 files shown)


extensions/windows-event-log/Bookmark.cpp
```diff
@@ -11,7 +11,7 @@ namespace nifi {
 namespace minifi {
 namespace processors {
 
-Bookmark::Bookmark(const std::wstring& channel, const std::wstring& query, const std::string& bookmarkRootDir, const std::string& uuid, std::shared_ptr<logging::Logger> logger)
+Bookmark::Bookmark(const std::wstring& channel, const std::wstring& query, const std::string& bookmarkRootDir, const std::string& uuid, bool processOldEvents, std::shared_ptr<logging::Logger> logger)
```
arpadboda, line 14: This is an incompatible API change.
Why don't you put the bool to the end with a default value?

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -149,6 +149,14 @@ core::Property ConsumeWindowsEventLog::BookmarkRootDirectory(
   withDescription("Directory which contains processor state data.")->
   build());
 
+core::Property ConsumeWindowsEventLog::ProcessOldEvents(
+  core::PropertyBuilder::createProperty("Process Old Events")->
+  isRequired(true)->
+  withAllowableValues<std::string>({ "true", "false" })->
```
arpadboda, line 155: In case you add a bool validator, you don't need this, but it becomes more consistent with C2.

extensions/windows-event-log/Bookmark.cpp
```diff
@@ -48,7 +48,7 @@ Bookmark::Bookmark(const std::wstring& channel, const std::wstring& query, const
   }
   const utils::ScopeGuard guard_hEventResults([hEventResults]() { EvtClose(hEventResults); });
 
-  if (!EvtSeek(hEventResults, 0, 0, 0, EvtSeekRelativeToLast)) {
+  if (!EvtSeek(hEventResults, 0, 0, 0, processOldEvents? EvtSeekRelativeToFirst : EvtSeekRelativeToLast)) {
```
arpadboda, line 51: What happens in case the property is true and the server restarted? (It processed some events before, so not all event should be reprocessed)

am-c-p-p: Since already processed events are bookmarked, only events which are not processed from  bookmark point will be processed, it could be not yet processed old events or new ones.

extensions/windows-event-log/Bookmark.cpp
am-c-p-p: The change in Bookmark is internal to CWEL, it doesn't affect other code, it is not necessary to maintain API signature.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp

szaszm:
Shouldn't this be a configuration setting instead of a processor property? Since it's effective only once, on the first agent run, there's no point in ever changing it.

PR #770, opened 2020-04-24
MINIFICPP-1203 - upgrade linter version to 1.4.4 and fix relevant linter errors

hunyadi-dev:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/src/io/StreamFactory.cpp
+++ b/libminifi/src/io/StreamFactory.cpp
@@ -16,12 +16,13 @@
  * limitations under the License.
  */
 #include "io/StreamFactory.h"
+
 #include <atomic>
 #include <memory>
 #include <mutex>
 #include <string>
-#include <utils/GeneralUtils.h>
 
+#include <utils/GeneralUtils.h> // NOLINT
 #ifdef OPENSSL_SUPPORT
 #include "io/tls/TLSSocket.h"
 #endif
--- a/libminifi/src/utils/OsUtils.cpp
+++ b/libminifi/src/utils/OsUtils.cpp
@@ -24,16 +24,18 @@
 #ifndef WIN32_LEAN_AND_MEAN
 #define WIN32_LEAN_AND_MEAN
 #endif
-#include <Windows.h>
 #include <sddl.h>
+#include <Windows.h>
 #pragma comment(lib, "Ws2_32.lib")
 #else
-#include <mutex>
-#include <pwd.h>
-#include <netdb.h>
 #include <arpa/inet.h>
+#include <netdb.h>
+#include <pwd.h>
 #include <sys/types.h>
+
+#include <mutex>
 #endif
+
 namespace org {
 namespace apache {
 namespace nifi {
--- a/libminifi/src/utils/OsUtils.cpp
+++ b/libminifi/src/utils/OsUtils.cpp
@@ -24,8 +24,8 @@
 #ifndef WIN32_LEAN_AND_MEAN
 #define WIN32_LEAN_AND_MEAN
 #endif
-#include <sddl.h>
 #include <Windows.h>
+#include <sddl.h>
 #pragma comment(lib, "Ws2_32.lib")
 #else
 #include <arpa/inet.h>
--- a/libminifi/test/archive-tests/FocusArchiveTests.cpp
+++ b/libminifi/test/archive-tests/FocusArchiveTests.cpp
@@ -16,24 +16,23 @@
  * limitations under the License.
  */
 
+#include <algorithm>
 #include <iostream>
 #include <map>
+#include <memory>
 #include <set>
 #include <string>
-#include <algorithm>
-#include <memory>
 #include <utility>
 
-#include "../TestBase.h"
+#include <archive.h> // NOLINT
+#include <archive_entry.h> // NOLINT
 #include "ArchiveTests.h"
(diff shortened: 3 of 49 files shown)


hunyadi-dev:
:exclamation: This PR depends on #769: ensuring that the `make linter` builds are clean with both versions of the linter.

Originally the linter was sourced from the google style-guide fork which is not very well maintained, this is now replaced with the script shipped by the cpplint/cpplint repository maintainers

libminifi/src/io/StreamFactory.cpp
```diff
@@ -16,12 +16,13 @@
...
  */
 #include "io/StreamFactory.h"
+
 #include <atomic>
 #include <memory>
 #include <mutex>
 #include <string>
-#include <utils/GeneralUtils.h>
 
+#include <utils/GeneralUtils.h> // NOLINT
```
szaszm, line 25: why do we need NOLINT here?
Also, end of line comments should have 2 preceding spaces.

libminifi/test/archive-tests/FocusArchiveTests.cpp
```diff
@@ -16,24 +16,23 @@
...
+#include <memory>
 #include <set>
 #include <string>
-#include <algorithm>
-#include <memory>
 #include <utility>
 
-#include "../TestBase.h"
+#include <archive.h> // NOLINT
+#include <archive_entry.h> // NOLINT
```
szaszm, line 28: why do we need NOLINT here?

End of line comments should have 2 spaces before and 1 after the `//`.

libminifi/src/io/StreamFactory.cpp
hunyadi-dev: This is incorrectly recognized as a C header. Also `NOLINT` comments do not need to be preceded by 2 spaces.

libminifi/test/archive-tests/FocusArchiveTests.cpp
hunyadi-dev: This is incorrectly recognized as a C header. Also `NOLINT` comments do not need to be preceded by 2 spaces.

hunyadi-dev:
> I find the Google include ordering rules strange, but that's in CONTRIB.md, so I guess we're better off enforcing them.
> 
> Thanks for the upgrade!

Agreed on the rules being strange, I would much rather prefer the llvm standard too. I could patch the linter for changing the order, but that would mean that we would need to apply the same patch every time we want to upgrade the linter version (and that everyone developping a code with linter assistance would need to apply the same patch to their own linter). I don't think the change is worth the effort.

libminifi/src/io/StreamFactory.cpp
szaszm: https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

I think if we replaced the angle brackets with quotation marks, the warning would go away. Could you try this? If it doesn't work, them I'm fine with leaving the NOLINT comments of such cases in place.

libminifi/src/utils/OsUtils.cpp
```diff
@@ -24,16 +24,18 @@
 #ifndef WIN32_LEAN_AND_MEAN
 #define WIN32_LEAN_AND_MEAN
 #endif
-#include <Windows.h>
 #include <sddl.h>
+#include <Windows.h>
```
szaszm, line 28: Judging from the appveyor output, the order of these includes matter. Not sure of the exact reason, but in general windows headers are a mess, so I'm not surprised. :(

See the end of the log here: https://ci.appveyor.com/project/ApacheSoftwareFoundation/nifi-minifi-cpp/builds/33004452

libminifi/test/tensorflow-tests/TensorFlowTests.cpp
```diff
@@ -19,14 +19,14 @@
...
-#include <tensorflow/cc/framework/scope.h>
-#include <tensorflow/cc/ops/standard_ops.h>
-#include <processors/PutFile.h>
-#include <processors/GetFile.h>
-#include <processors/LogAttribute.h>
+#include <processors/GetFile.h> // NOLINT
+#include <processors/LogAttribute.h> // NOLINT
+#include <processors/PutFile.h> // NOLINT
+#include <tensorflow/cc/framework/scope.h> // NOLINT
+#include <tensorflow/cc/ops/standard_ops.h> // NOLINT
```
szaszm, line 26: Regarding my previous suggestion: I suggest trying to convert our headers to use quotation marks and leave 3rd parties as is, even if it means that we need to use NOLINT.

hunyadi-dev: Will do that in a separate PR, right? I don't want to rebase my linter PR again.

libminifi/src/utils/OsUtils.cpp
hunyadi-dev: Corrected.

libminifi/test/tensorflow-tests/TensorFlowTests.cpp
szaszm: My initial idea was to do it here, but you're right, it's better not to mix in this.

arpadboda:
Merged in 29460fa1a035096b251a7f2a323ddffcec2c4210

PR #773, opened 2020-04-27
MINIFICPP-1202 - Extend interface and add new tests for MinifiConcurrentQueue

hunyadi-dev:

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/test/unit/MinifiConcurrentQueueTests.cpp
```diff
@@ -123,37 +234,79 @@ TEST_CASE("TestConqurrentQueue::testQueueWithReAdd", "[TestQueueWithReAdd]") {
...
+
+TEST_CASE("TestConditionConqurrentQueue::testConditionQueueDequeueApplyWaitForNoSignal", "[testConditionQueueDequeueApplyWaitForNoSignal]") {
+  using namespace MinifiConcurrentQueueTestProducersConsumers;
+  utils::ConditionConcurrentQueue<std::string> queue(true);
+  std::set<std::string> results;
+  std::mutex mutex;
+  std::unique_lock<std::mutex> lock(mutex);
+
+  std::thread producer{ getBlockedProducerThread(queue, mutex) };
+  std::thread consumer{ getDequeueWaitForConsumerThread(queue, results) };
```
szaszm, line 302: There are some copy-paste errors with the wrong functions being called. Could you double-check them?

This test is identical to the above but its name suggests that it shouldn't be.

libminifi/include/utils/MinifiConcurrentQueue.h
```diff
@@ -99,6 +106,19 @@ class ConcurrentQueue {
...
 
+  template<typename Functor>
+  bool dequeueApplyImpl(std::unique_lock<std::mutex>& lck, Functor&& fun) {
+    checkLock(lck);
+    if (queue_.empty()) {
+      return false;
+    }
+    T elem = std::move(queue_.front());
+    queue_.pop_front();
+    lck.unlock();
```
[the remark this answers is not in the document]
arpadboda: It seems like a timed bomb for me. 
You pass a lock (which you locked) as a reference and the called function may(!!!) unlock it, but only if it's not empty. 
On the caller side it's a nightmare to handle. 

The only way I can imagine this to work is moving the lock (this function actually takes ownership of that), so the caller is aware that it no longer holds it under any circumstances. 
Still not nice, but at least it protects against the undefined ownership issue.

libminifi/include/utils/MinifiConcurrentQueue.h
```diff
@@ -127,33 +147,58 @@ class ConditionConcurrentQueue : private ConcurrentQueue<T> {
...
 
   template< class Rep, class Period >
   bool dequeueWaitFor(T& out, const std::chrono::duration<Rep, Period>& time) {
+    if (!running_) {
+      return false;
+    }
     std::unique_lock<std::mutex> lck(this->mtx_);
     cv_.wait_for(lck, time, [this, &lck]{ return !running_ || !this->emptyImpl(lck); });  // Wake up with timeout or in case there is something to do
-    return running_ && ConcurrentQueue<T>::tryDequeueImpl(lck, out);
+    return ConcurrentQueue<T>::tryDequeueImpl(lck, out);
```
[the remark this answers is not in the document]
arpadboda: As first look this wouldn't hurt, but I would prefer checking the running flag in scoped of the locked mutex to make sure.

libminifi/include/utils/MinifiConcurrentQueue.h
```diff
@@ -127,33 +147,58 @@ class ConditionConcurrentQueue : private ConcurrentQueue<T> {
...
+  }
+
+  template<typename Functor, class Rep, class Period>
+  bool dequeueApplyWaitFor(Functor&& fun, const std::chrono::duration<Rep, Period>& time) {
+    if (!running_) {
```
arpadboda, line 189: Here as well.

libminifi/include/utils/MinifiConcurrentQueue.h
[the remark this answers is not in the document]
hunyadi-dev: My reasoning for doing this was that `stop` is always changing the `running_` boolean under a lock, so if we don't have it flipped before we grab the lock, then there is no reason to have it after. I would move this back under the loop, because optimizing for this rare case is not worth the risk of reading a non-atomic boolean outside locking.

libminifi/test/unit/MinifiConcurrentQueueTests.cpp
hunyadi-dev: Ah, I see, that is a very good catch, this should be calling the consumer using `apply`. Also, just realized, that the strings `TestConcurrentqueue`/`TestConditionConcurrentQueue` are consistently (but not consequently) mistyped, will correct at least that as well.

libminifi/include/utils/MinifiConcurrentQueue.h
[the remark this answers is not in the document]
hunyadi-dev: Good catch, now the `unique_lock` can only be moved to the function.

libminifi/include/utils/MinifiConcurrentQueue.h
hunyadi-dev: Updated by moving the check under the lock.

libminifi/include/utils/MinifiConcurrentQueue.h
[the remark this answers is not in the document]
hunyadi-dev: Updated by moving the check under the lock.

libminifi/include/utils/MinifiConcurrentQueue.h
```diff
@@ -32,7 +33,7 @@ namespace utils {
 
 // Provides a queue API and guarantees no race conditions in case of multiple producers and consumers.
 // Guarantees elements to be dequeued in order of insertion
-template <typename T>
+template <typename T, typename = typename std::enable_if<std::is_nothrow_move_constructible<T>::value>::type>
```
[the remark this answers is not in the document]
hunyadi-dev: This is now replaced with static assertions.

libminifi/include/utils/MinifiConcurrentQueue.h
```diff
@@ -57,6 +58,12 @@ class ConcurrentQueue {
     return tryDequeueImpl(lck, out);
   }
 
+  template<typename Functor>
+  bool dequeueApply(Functor&& fun) {
```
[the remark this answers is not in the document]
hunyadi-dev: Renamed to `consume`.

libminifi/test/unit/MinifiConcurrentQueueTests.cpp
hunyadi-dev: Replaced duplicate tests and fixed typos.

libminifi/include/utils/MinifiConcurrentQueue.h
```diff
@@ -99,6 +106,19 @@ class ConcurrentQueue {
...
+  template<typename Functor>
+  bool dequeueApplyImpl(std::unique_lock<std::mutex>& lck, Functor&& fun) {
+    checkLock(lck);
+    if (queue_.empty()) {
+      return false;
+    }
+    T elem = std::move(queue_.front());
+    queue_.pop_front();
+    lck.unlock();
+    fun(elem);
```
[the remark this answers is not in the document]
hunyadi-dev: Replaced with forwading on both fun and elem.

libminifi/include/utils/MinifiConcurrentQueue.h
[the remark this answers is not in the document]
szaszm: What loop?

I think the problem Arpad refers to is the check without the mutex lock at the beginning. Unsynchronized accesses are undefined behavior [according to the standard](http://eel.is/c++draft/intro.races#21), even if it feels like it "should work".

libminifi/include/utils/MinifiConcurrentQueue.h
```diff
@@ -21,14 +21,32 @@
...
+using void_t = void;
+
+template<typename /* FunType */, typename T, typename = void>
+struct TryMoveCall {
+    template<typename Fun>
+    static void call(Fun&& fun, T& elem) { std::forward<Fun>(fun)(elem); }
+};
+
+template<typename FunType, typename T>
+struct TryMoveCall<FunType, T, void_t<decltype(std::declval<FunType>()(std::declval<T>()))>> {
```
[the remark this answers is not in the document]
hunyadi-dev: Agreed.

libminifi/include/utils/MinifiConcurrentQueue.h
```diff
@@ -175,47 +175,50 @@ class ConditionConcurrentQueue : private ConcurrentQueue<T> {
...
     return ConcurrentQueue<T>::consumeImpl(std::move(lck), std::forward<Functor>(fun));
   }
 
   bool tryDequeue(T& out) {
     std::unique_lock<std::mutex> lck(this->mtx_);
-    return running_ && ConcurrentQueue<T>::tryDequeueImpl(lck, out);
+    if (!running_) {
+      return false;
+    }
+    return ConcurrentQueue<T>::tryDequeueImpl(lck, out);
```
szaszm, line 221: This is the same as the old version, but longer.

libminifi/include/utils/MinifiConcurrentQueue.h
```diff
@@ -99,6 +124,20 @@ class ConcurrentQueue {
     return true;
   }
 
+  template<typename Functor>
+  bool consumeImpl(std::unique_lock<std::mutex>&& lck, Functor&& fun) {
+    static_assert(std::is_nothrow_move_assignable<T>::value, "T has to be nothrow move assignable.");
```
szaszm, line 129: This is the one only doing move construction (once for initializing `elem` and potentially a second time when moving elem to the `fun` argument) and the other one is only doing move assignment (when assigning the element to `out`).

libminifi/include/utils/MinifiConcurrentQueue.h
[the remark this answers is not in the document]
szaszm: In the new version, the user passes an rvalue reference to their lock guard, but no move will actually happen that would prevent the surprising behavior. I suggest passing the `std::unique_lock` by value.

libminifi/include/utils/MinifiConcurrentQueue.h
hunyadi-dev: I wanted to make consistent over concise.

libminifi/include/utils/MinifiConcurrentQueue.h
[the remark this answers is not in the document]
szaszm: Typo in 4.: s/cosidered/considered/
2.: makes no assumptions about the parameter type of a `FunType` object. If it takes an lvalue, then passing 1. will be ill-formed.

Rewritten in my words, but significantly longer and more complex. Take what you like:
1. `std::declval` looks similar to this: `template<typename T> T&& declval();`. Not defined, therefore it's only usable in unevaluated context. No requirements regarding `T`, therefore makes it possible to create hypothetical objects without requiring e.g. a default constructor and a destructor, like the `T{}` expression does.
2. `std::declval<FunType>()` resolves to an object of type `FunType`. If `FunType` is an lvalue reference, then this will also result in an lvalue reference due to reference collapsing.
3. `std::declval<FunType>()(std::declval<T>())` resolves to an object of the result type of a call on a function object of type `FunType` with an rvalue argument of type `T`. It is ill-formed if the function object expect an lvalue reference.
    - Example: `FunType` is a pointer to a `bool(int)` and `T` is `int`. This expression will result in a `bool` object.
    - Example: `FunType` is a function object modeling `bool(int&)` and `T` is `int`. This expression will be ill-formed because it's illegal to bind an `int` rvalue to an `int&`.
4. `void_t<decltype(*3*)>` checks for the well-formedness of 3., then discards it. If 3. is ill-formed, then this specialization is ignored through SFINAE. If well-formed, then it's considered more specialized than the other and takes precedence.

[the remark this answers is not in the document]
hunyadi-dev: Will perform a capture-by-copy-paste on it, this is as good as a documentation can get :)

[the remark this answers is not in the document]
hunyadi-dev: Comment updated in the commit mentioning your name :)

libminifi/include/utils/MinifiConcurrentQueue.h
hunyadi-dev: Restored to original state.

libminifi/include/utils/MinifiConcurrentQueue.h
hunyadi-dev: Corrected.

libminifi/include/utils/MinifiConcurrentQueue.h
```diff
@@ -21,14 +21,49 @@
...
 #include <stdexcept>
+#include <type_traits>
 
 namespace org {
 namespace apache {
 namespace nifi {
 namespace minifi {
 namespace utils {
 
+namespace detail {
```
arpadboda, line 34: This is good stuff, but I think it should be in GeneralUtils.h as this is not a concurrentQueue-specific functionality.

libminifi/include/utils/MinifiConcurrentQueue.h
```diff
@@ -127,33 +184,58 @@ class ConditionConcurrentQueue : private ConcurrentQueue<T> {
...
+      return false;
+    }
+    cv_.wait(lck, [this, &lck]{ return !running_ || !this->emptyImpl(lck); });  // Only wake up if there is something to return or stopped
+    return ConcurrentQueue<T>::tryDequeueImpl(lck, out);
+  }
+
+  template<typename Functor>
+  bool consumeWait(Functor&& fun) {
+    std::unique_lock<std::mutex> lck(this->mtx_);
+    if (!running_) {
```
arpadboda, line 207: I wonder why this is any better than the original solutions.
It's 3 lines longer, but for what?

The original code didn't start waiting for the CV as the predicate was true at the beginning, so immediately returned false, as we this does. 

Could you explain the rational behind this change?

libminifi/test/unit/MinifiConcurrentQueueTests.cpp
```diff
@@ -29,132 +29,290 @@
...
+    }
+    SECTION("using consumeWaitFor") {
+      consumer = getConsumeWaitForConsumerThread(queue, results);
+    }
+    consumer.join();
+    lock.unlock();
+    producer.join();
 
-  REQUIRE(utils::StringUtils::join("-", results) == "ba-dum-tss");
+    REQUIRE(0 == results.size());
```
szaszm, line 311: clang-tidy: The `empty` method should be used to check for emptiness instead of `size`

libminifi/test/unit/MinifiConcurrentQueueTests.cpp
```diff
@@ -29,132 +29,290 @@
...
-
-  std::thread producer([&queue]() {
-    queue.enqueue("ba");
-    std::this_thread::sleep_for(std::chrono::milliseconds(3));
-    queue.enqueue("dum");
-    std::this_thread::sleep_for(std::chrono::milliseconds(3));
-    queue.enqueue("tss");
-  });
+      SECTION("tryDequeue on empty queue returns false") {
+        REQUIRE(false == queue.tryDequeue(s));
```
szaszm, line 179: clion inspection: "Expression can be simplified to `! queue.tryDequeue(s)`"

libminifi/test/unit/MinifiConcurrentQueueTests.cpp
```diff
@@ -29,132 +29,290 @@
...
-    while (results.size() < 3) {
-      std::string s;
-      if (queue.tryDequeue(s)) {
-        results.insert(s);
-        queue.enqueue(std::move(s));
-      } else {
-        std::this_thread::sleep_for(std::chrono::milliseconds(1));
+      SECTION("consume on empty queue returns false") {
+        bool ret = queue.consume([&s] (const std::string& elem) { s = elem; });
+        REQUIRE(false == ret);
```
szaszm, line 184: clion inspection: "Expression can be simplified to `!ret`"

libminifi/include/utils/MinifiConcurrentQueue.h
szaszm: +1 for extracting the `void_t` implementation but -1 for `TryMoveCall` (at least until we need it elsewhere).
If `void_t` is extracted, please adapt the partial detection idiom impl above `StringUtils::join_pack` to use the extracted one.

libminifi/test/unit/MinifiConcurrentQueueTests.cpp
```diff
@@ -29,132 +29,290 @@
...
+        queue.enqueue(std::move(s));
+      }
+    });
+  }
 
-  producer.join();
+  std::thread getDequeueWaitForConsumerThread(utils::ConditionConcurrentQueue<std::string>& queue, std::vector<std::string>& results) {
+    return std::thread([&queue, &results] {
+      const std::size_t max_read_attempts = 6;
+      std::size_t attemt_num = 0;
```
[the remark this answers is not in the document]
hunyadi-dev: Thanks for linking the dictionary :)

libminifi/include/utils/MinifiConcurrentQueue.h
hunyadi-dev: I felt like refraining from waiting on the lock if the queue is stopped. This is redundant though and will be removed.

libminifi/test/unit/MinifiConcurrentQueueTests.cpp
hunyadi-dev: Sure, I'll replace this, I am just used to writing tests in `expected, actual` notation.

libminifi/include/utils/MinifiConcurrentQueue.h
hunyadi-dev: Extracted both functions to their according header, updated `StringUtils` to include `void_t` from the external definition.

libminifi/include/utils/MinifiConcurrentQueue.h
hunyadi-dev: Removed.

libminifi/test/unit/MinifiConcurrentQueueTests.cpp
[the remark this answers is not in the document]
hunyadi-dev: Corrected to follow the spelling of line 138 :)

libminifi/test/unit/MinifiConcurrentQueueTests.cpp
hunyadi-dev: Updated as requested.

libminifi/test/unit/MinifiConcurrentQueueTests.cpp
```diff
@@ -29,132 +29,290 @@
...
+      while (results.size() < 3 && attemt_num < max_read_attempts) {
```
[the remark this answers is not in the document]
hunyadi-dev: Updated as requested.

libminifi/test/unit/MinifiConcurrentQueueTests.cpp
hunyadi-dev: Updated as requested.

libminifi/test/unit/MinifiConcurrentQueueTests.cpp
hunyadi-dev: Updated as requested.

libminifi/test/unit/MinifiConcurrentQueueTests.cpp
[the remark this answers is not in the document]
szaszm: I didn't notice the correct one on the previous line. Also, I meant no offense by linking the dictionary. Thanks for fixing it. :)

libminifi/test/unit/MinifiConcurrentQueueTests.cpp
szaszm: Many of my code review comments point out things that can be done in a shorter/simpler way if there are no reasons for doing otherwise. Your preference for the "expected, actual" notation is a good enough reason for me to ignore this inspection. Feel free to question my suggestions with valid points.

szaszm:
Thanks for your effort towards fixing all of the issues we pointed out. The consume family of member functions will provide a significant improvement in the usability of the class. I also like that the unit tests now utilize sections.

As I mentioned [here](https://github.com/apache/nifi-minifi-cpp/pull/773#discussion_r419476317), feel free to question my code review comments anytime, as oftentimes they are just mere ideas to keep things simple/short, and may be missing some important point.

I'll do another review of the latest changes soon.

libminifi/include/utils/Void_t.h
```diff
@@ -0,0 +1,17 @@
+#pragma once
```
szaszm, line 17: I feel it a bit of overkill to have a separate file for such a tiny utility. I would just put `void_t` in GeneralUtils.h next to the other polyfills.

libminifi/include/utils/Void_t.h
```diff
@@ -0,0 +1,33 @@
...
+ */
+#pragma once
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace utils {
+
+// ppartial detection idiom impl
```
szaszm, line 25: `void_t` is used by, but not really part of the detection idiom. Could you move the comment line back to StringUtils.h?

I forgot to add a reference to cppreference.com in the original. When moving it back, could you please append `, from cppreference.com` to the comment and fix the typo (`s/ppartial/partial/`)?

So my ask summarized:
- remove this comment line
- Add the line below to where it was removed from:
```
// partial detection idiom impl, from cppreference.com
```

libminifi/include/utils/TryMoveCall.h
```diff
@@ -0,0 +1,61 @@
...
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace utils {
+
+// TryMoveCall calls an
+//  - unary function of a lvalue reference-type argument by passing a ref
+//  - unary function of any other argument type by moving into it
+template<typename /* FunType */, typename T, typename = void>
+struct TryMoveCall {
```
szaszm, line 33: If we decouple this from the queue, we should consider making it handle return values (other than `void`).

libminifi/include/utils/MinifiConcurrentQueue.h
```diff
@@ -21,20 +21,23 @@
 #include <deque>
 #include <mutex>
 #include <condition_variable>
+#include <utility>
 #include <stdexcept>
+#include <type_traits>
```
szaszm, line 26: `<type_traits>` is no longer used since `TryMoveCall` and `void_t` were extracted. Please fix headers to follow `include-what-you-use`

rationale:
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf11-header-files-should-be-self-contained
If `utils/TryMoveCall.h` were not self-contained, following iwyu would more likely expose the issue in the form of compilation caure.

hunyadi-dev:
Built and tested on windows as well:
Screenshot 2020-05-05 at 12 33 49

libminifi/include/utils/Void_t.h
hunyadi-dev: Moved to `GeneralUtils.h`

libminifi/include/utils/MinifiConcurrentQueue.h
hunyadi-dev: Removed inclusion of `type_traits.h`.

libminifi/include/utils/TryMoveCall.h
hunyadi-dev: Added support for non-void return types.

libminifi/include/utils/Void_t.h
hunyadi-dev: Changed as requested.

libminifi/include/utils/GeneralUtils.h
```diff
@@ -56,6 +56,11 @@ T exchange(T& obj, U&& new_value) {
   obj = std::forward<U>(new_value);
   return old_value;
 }
+
+// partial detection idiom impl, from cppreference.com
```
szaszm, line 60: this comment should go back to StringUtils.h

libminifi/include/utils/Void_t.h
szaszm: the comment belongs to stringutils.h

libminifi/include/utils/GeneralUtils.h
hunyadi-dev: Done.

PR #775, opened 2020-04-29
MINIFICPP-965 Add other supported relationships to InvokeHTTP processor.

msharee9:
Added response, retry, no retry, failure relationships.
Transfer request and response flow files appropriately.
Overloaded virtual functions to set connection timeout and read timeout
using std chrono
Deprecated the old set timeout functions that use int64_t

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/http-curl/client/HTTPClient.cpp
+++ b/extensions/http-curl/client/HTTPClient.cpp
@@ -36,8 +36,6 @@ HTTPClient::HTTPClient(const std::string &url, const std::shared_ptr<minifi::con
     : core::Connectable("HTTPClient"),
       ssl_context_service_(ssl_context_service),
       url_(url),
-      connect_timeout_(0),
-      read_timeout_(0),
       content_type_str_(nullptr),
       headers_(nullptr),
       callback(nullptr),
@@ -46,8 +44,6 @@ HTTPClient::HTTPClient(const std::string &url, const std::shared_ptr<minifi::con
       read_callback_(INT_MAX),
       header_response_(-1),
       res(CURLE_OK),
-      keep_alive_probe_(-1),
-      keep_alive_idle_(-1),
       logger_(logging::LoggerFactory<HTTPClient>::getLogger()) {
   http_session_ = curl_easy_init();
 }
@@ -56,8 +52,6 @@ HTTPClient::HTTPClient(std::string name, utils::Identifier uuid)
     : core::Connectable(name, uuid),
       ssl_context_service_(nullptr),
       url_(),
-      connect_timeout_(0),
-      read_timeout_(0),
       content_type_str_(nullptr),
       headers_(nullptr),
       callback(nullptr),
@@ -66,8 +60,6 @@ HTTPClient::HTTPClient(std::string name, utils::Identifier uuid)
       read_callback_(INT_MAX),
       header_response_(-1),
       res(CURLE_OK),
-      keep_alive_probe_(-1),
-      keep_alive_idle_(-1),
       logger_(logging::LoggerFactory<HTTPClient>::getLogger()) {
   http_session_ = curl_easy_init();
 }
@@ -76,8 +68,6 @@ HTTPClient::HTTPClient()
     : core::Connectable("HTTPClient"),
       ssl_context_service_(nullptr),
       url_(),
-      connect_timeout_(0),
-      read_timeout_(0),
       content_type_str_(nullptr),
       headers_(nullptr),
       callback(nullptr),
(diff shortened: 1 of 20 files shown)


extensions/http-curl/tests/HTTPHandlers.h
```diff
@@ -445,4 +445,49 @@ class HeartbeatHandler : public CivetHandler {
   bool isSecure;
 };
 
+class InvokeHTTPResponseOKHandler : public CivetHandler {
+public:
+  bool handlePost(CivetServer *, struct mg_connection *conn) {
+    std::stringstream headers;
+    headers << "HTTP/1.1 201 OK\r\nContent-Type: text/plain\r\nContent-Length: 0\r\nConnection: close\r\n\r\n";
```
arpadboda, line 452: I don't think stringstream makes sense in case we only put one string to it. 
Can we simply pass the hardcoded cstr to the mg_printf?

extensions/http-curl/tests/HTTPHandlers.h
```diff
@@ -445,4 +445,49 @@ class HeartbeatHandler : public CivetHandler {
...
+  }
+};
+
+class InvokeHTTPResponseTimeoutHandler : public CivetHandler {
+public:
+    InvokeHTTPResponseTimeoutHandler(std::chrono::milliseconds wait_ms)
+        : wait_(wait_ms) {
+    }
+  bool handlePost(CivetServer *, struct mg_connection *conn) {
+    std::this_thread::sleep_for(wait_);
```
arpadboda, line 484: Nice solution!

libminifi/test/resources/TestInvokeHTTPPost.yml
```diff
@@ -0,0 +1,161 @@
...
+  auto-terminated relationships list: []
+  Properties:
+    Always Output Response: 'false'
+    Connection Timeout: 3 s
+    Content-type: application/octet-stream
+    Disable Peer Verification: 'false'
+    HTTP Method: POST
+    Include Date Header: 'true'
+    Read Timeout: 4 s
+    Remote URL: http://localhost:10033/minifi
```
arpadboda, line 79: I don't really like using hardcoded magic ports and in my opinion we can avoid this.

Civet server supports starting the webserver on random port (I think we already have some examples for that usage in our tests), and in queryRootProcessGroup we could override this property.
The civet server should be started before this function executes, but I think that's already done in proper order.

extensions/http-curl/tests/HTTPHandlers.h
msharee9: Absolutely. stringstream is unnecessary here.

extensions/http-curl/client/HTTPClient.h
```diff
@@ -124,11 +135,11 @@ class HTTPClient : public BaseHTTPClient, public core::Connectable {
 
   bool setMinimumSSLVersion(SSLVersion minimum_version) override;
 
-  void setKeepAliveProbe(long probe){
+  void setKeepAliveProbe(std::chrono::milliseconds probe){
     keep_alive_probe_ = probe;
   }
 
-  void setKeepAliveIdle(long idle){
```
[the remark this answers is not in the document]
msharee9: Agree. especially when they are public members.

extensions/http-curl/client/HTTPClient.cpp
```diff
@@ -296,12 +296,13 @@ bool HTTPClient::submit() {
   }
   curl_easy_setopt(http_session_, CURLOPT_HEADERFUNCTION, &utils::HTTPHeaderResponse::receive_headers);
   curl_easy_setopt(http_session_, CURLOPT_HEADERDATA, static_cast<void*>(&header_response_));
-  if (keep_alive_probe_ > 0){
-    logger_->log_debug("Setting keep alive to %d",keep_alive_probe_);
+  if (keep_alive_probe_.count() > 0) {
+    const auto keepAlive = std::chrono::duration_cast<std::chrono::duration<uint64_t>>(keep_alive_probe_);
```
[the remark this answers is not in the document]
msharee9: Agree that std::chrono::seconds is better here. However, on a side note, chrono::seconds is std::duration<int64_t>
https://gcc.gnu.org/onlinedocs/gcc-4.6.3/libstdc++/api/a00798_source.html#l00508

It does not matter here if we use int64_t or uint64_t because we are checking for negative value.

libminifi/test/resources/TestInvokeHTTPPost.yml
msharee9: Agree with you on this. Changed the behavior to use generated port numbers.

arpadboda:
@msharee9 : the code looks good to me, but I see an exception thrown in the Windows CI job, could you take a look?

msharee9:
> @msharee9 : the code looks good to me, but I see an exception thrown in the Windows CI job, could you take a look?

@arpadboda Fixed it. Should be good to go.

PR #776, opened 2020-05-05
MINIFICPP-1202 - Handle C2 requests/responses using MinifiConcurrentQueue

hunyadi-dev:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/c2/C2Agent.h
+++ b/libminifi/include/c2/C2Agent.h
@@ -181,10 +181,10 @@ class C2Agent : public state::UpdateController {
   std::map<std::string, std::shared_ptr<state::response::ResponseNode>> device_information_;
 
   // responses for the the C2 agent.
-  utils::ConditionConcurrentQueue<C2Payload> responses;
+  utils::ConcurrentQueue<C2Payload> responses;
 
   // requests that originate from the C2 server.
-  utils::ConditionConcurrentQueue<C2Payload> requests;
+  utils::ConcurrentQueue<C2Payload> requests;
 
   // heart beat period.
   int64_t heart_beat_period_;
--- a/libminifi/include/c2/C2Agent.h
+++ b/libminifi/include/c2/C2Agent.h
@@ -34,21 +34,16 @@
 #include "C2Protocol.h"
 #include "io/validation.h"
 #include "HeartBeatReporter.h"
-#include "utils/ThreadPool.h"
 #include "utils/Id.h"
+#include "utils/MinifiConcurrentQueue.h"
+#include "utils/ThreadPool.h"
 
 namespace org {
 namespace apache {
 namespace nifi {
 namespace minifi {
 
-namespace utils {
-template <typename T>
-class ConditionConcurrentQueue;
-}
-
 namespace c2 {
-
 
 #define C2_AGENT_UPDATE_NAME "C2UpdatePolicy"
 /**
--- a/libminifi/include/c2/C2Agent.h
+++ b/libminifi/include/c2/C2Agent.h
@@ -41,7 +41,14 @@ namespace org {
 namespace apache {
 namespace nifi {
 namespace minifi {
+
+namespace utils {
+template <typename T>
+class ConditionConcurrentQueue;
+}
+
 namespace c2 {
+
 
 #define C2_AGENT_UPDATE_NAME "C2UpdatePolicy"
 /**
@@ -128,16 +135,14 @@ class C2Agent : public state::UpdateController {
    * Enqueues a C2 server response for us to evaluate and parse.
    */
   void enqueue_c2_server_response(C2Payload &&resp) {
-    std::lock_guard<std::timed_mutex> lock(queue_mutex);
-    responses.push_back(std::move(resp));
+    responses.enqueue(std::move(resp));
   }
 
   /**
    * Enqueues a c2 payload for a response to the C2 server.
    */
   void enqueue_c2_response(C2Payload &&resp) {
-    std::lock_guard<std::timed_mutex> lock(request_mutex);
-    requests.push_back(std::move(resp));
+    requests.enqueue(std::move(resp));
   }
 
   /**
@@ -179,17 +184,12 @@ class C2Agent : public state::UpdateController {
    * Device information stored in the metrics format
    */
   std::map<std::string, std::shared_ptr<state::response::ResponseNode>> device_information_;
-  // queue mutex
-  std::timed_mutex queue_mutex;
-
-  // queue mutex
-  std::timed_mutex request_mutex;
 
   // responses for the the C2 agent.
-  std::vector<C2Payload> responses;
+  utils::ConditionConcurrentQueue<C2Payload> responses;
 
   // requests that originate from the C2 server.
-  std::vector<C2Payload> requests;
+  utils::ConditionConcurrentQueue<C2Payload> requests;
 
   // heart beat period.
   int64_t heart_beat_period_;
--- a/libminifi/src/c2/C2Agent.cpp
+++ b/libminifi/src/c2/C2Agent.cpp
@@ -17,6 +17,7 @@
  */
 
 #include "c2/C2Agent.h"
+
 #include <csignal>
 #include <utility>
 #include <limits>
@@ -24,6 +25,7 @@
 #include <map>
 #include <string>
 #include <memory>
+
 #include "c2/ControllerSocketProtocol.h"
 #include "core/ProcessContext.h"
 #include "core/CoreComponentState.h"
@@ -33,8 +35,9 @@
 #include "utils/file/DiffUtils.h"
 #include "utils/file/FileUtils.h"
 #include "utils/file/FileManager.h"
-#include "utils/HTTPClient.h"
 #include "utils/GeneralUtils.h"
+#include "utils/HTTPClient.h"
+#include "utils/MinifiConcurrentQueue.h"
 #include "utils/Monitors.h"
 
 namespace org {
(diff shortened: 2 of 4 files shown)


hunyadi-dev:
This PR is dependent on #773. Locking until that one is merged.

libminifi/src/c2/C2Agent.cpp
```diff
@@ -75,54 +78,54 @@ C2Agent::C2Agent(const std::shared_ptr<core::controller::ControllerServiceProvid
...
-            catch(...) {
-              logger_->log_error("Unknonwn exception occurred while consuming payload.");
-            }
-          }while(!requests.empty() && ++count < max_c2_responses);
+    if (protocol_.load() != nullptr) {
+      std::vector<C2Payload> payload_batch;
+      payload_batch.reserve(max_c2_responses);
+      auto getRequestPayload = [&payload_batch] (C2Payload&& payload) { payload_batch.emplace_back(std::move(payload)); };
+      for (std::size_t attempt_num = 0; attempt_num < max_c2_responses; ++attempt_num) {
+        if (!requests.consumeWaitFor(getRequestPayload, std::chrono::seconds(1))) {
```
arpadboda, line 86: I don't think this behaviour is identical. 
The original waited at most 1 sec to acquire the mutex and grabbed as many events as it could.

Now we grab as many events as we can, waiting at most 1s for each.

libminifi/src/c2/C2Agent.cpp
```diff
@@ -75,54 +78,54 @@ C2Agent::C2Agent(const std::shared_ptr<core::controller::ControllerServiceProvid
...
-      {
-        std::lock_guard<std::timed_mutex> lock(queue_mutex, std::adopt_lock);
-        if (responses.empty()) {
-          return utils::TaskRescheduleInfo::RetryIn(std::chrono::milliseconds(C2RESPONSE_POLL_MS));
-        }
-        payload = std::move(responses.back());
-        responses.pop_back();
+  c2_consumer_ = [&] {
+    if (responses.size()) {
+      if (!responses.consumeWaitFor([this](C2Payload&& e) { extractPayload(std::move(e)); }, std::chrono::seconds(1))) {
```
arpadboda, line 126: Here as well.

libminifi/src/c2/C2Agent.cpp
hunyadi-dev: Will add a `wait_until` method to the queue and max out on 1 sec total time here.

libminifi/src/c2/C2Agent.cpp
hunyadi-dev: This is the same as before, right?

libminifi/src/c2/C2Agent.cpp
hunyadi-dev: Added `wait_until`.

libminifi/src/c2/C2Agent.cpp
```diff
@@ -75,54 +78,55 @@ C2Agent::C2Agent(const std::shared_ptr<core::controller::ControllerServiceProvid
...
-      checkTriggers();
+        try {
+          performHeartBeat();
+        }
+        catch (const std::exception &e) {
+          logger_->log_error("Exception occurred while performing heartbeat. error: %s", e.what());
+        }
+        catch (...) {
+          logger_->log_error("Unknonwn exception occurred while performing heartbeat.");
+        }
```
szaszm, line 115: incorrect/misleading indentation

libminifi/src/c2/C2Agent.cpp
```diff
@@ -75,54 +78,55 @@ C2Agent::C2Agent(const std::shared_ptr<core::controller::ControllerServiceProvid
...
-              logger_->log_error("Unknonwn exception occurred while consuming payload.");
-            }
-          }while(!requests.empty() && ++count < max_c2_responses);
+    if (protocol_.load() != nullptr) {
+      std::vector<C2Payload> payload_batch;
+      payload_batch.reserve(max_c2_responses);
+      auto getRequestPayload = [&payload_batch] (C2Payload&& payload) { payload_batch.emplace_back(std::move(payload)); };
+      const std::chrono::system_clock::time_point timeout_point = std::chrono::system_clock::now() + std::chrono::milliseconds(1);
+      for (std::size_t attempt_num = 0; attempt_num < max_c2_responses; ++attempt_num) {
+        if (!requests.consumeWaitUntil(getRequestPayload, timeout_point)) {
```
szaszm, line 87: Previously we took the immediately available payloads only. Now we wait 1ms. Why is this change?

If we reverted to the old way, it would also no longer be a problem to reunify the first two loops and eliminate the need for the temporary buffer (`payload_batch`).

hunyadi-dev: Good catch, we want to wait 1 second, not 1 millisecond on locking.

libminifi/src/c2/C2Agent.cpp
hunyadi-dev: That is correct, that is one-too-many indents due to the `std::for_each` actually taking up two identation levels.

libminifi/src/c2/C2Agent.cpp
```diff
@@ -75,54 +78,55 @@ C2Agent::C2Agent(const std::shared_ptr<core::controller::ControllerServiceProvid
...
-      {
-        std::lock_guard<std::timed_mutex> lock(queue_mutex, std::adopt_lock);
-        if (responses.empty()) {
-          return utils::TaskRescheduleInfo::RetryIn(std::chrono::milliseconds(C2RESPONSE_POLL_MS));
-        }
-        payload = std::move(responses.back());
-        responses.pop_back();
+  c2_consumer_ = [&] {
+    if (responses.size()) {
+      if (!responses.consumeWaitFor([this](C2Payload&& e) { extractPayload(std::move(e)); }, std::chrono::seconds(1))) {
```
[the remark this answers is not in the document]
hunyadi-dev: As far as I can see, this is the same behaviour:

- If no elements were present in the queue, we did not call `extractPayload()` at all.
- If there were elements, but by the time we wanted to grab the lock the queue became busy, we waited 1 second and tried fetching an element
  - If by this time we succeeded, we called `extractPayload()` with this dequeued data
  - Otherwise we used the default constructed payload: `C2Payload{ Operation::HEARTBEAT }` instead.

To me `!empty()` feels like something that is easily mistaken for `empty()`, `size()` is guaranteed to be constant time for `std::deque` and `empty()` is probably implemented as `size() == 0` anyway, right?

libminifi/include/c2/C2Agent.h
```diff
@@ -41,8 +41,15 @@ namespace org {
 namespace apache {
 namespace nifi {
 namespace minifi {
+
+namespace utils {
+template <typename T>
+class ConditionConcurrentQueue;
+}
```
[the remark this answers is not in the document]
hunyadi-dev: You are completely right, at first I thought we would only need the declaration and left this here. Will replace it with a proper include.

libminifi/src/c2/C2Agent.cpp
szaszm: The mutex inside the queue is not `timed_mutex` so we wait indefinitely for locking and only wait definitely for content.

libminifi/src/c2/C2Agent.cpp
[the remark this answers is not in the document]
szaszm: 1. My preference is `!empty()`, because it reads as "not empty", which is the intention here. On the other hand I'm ok with using `size() > 0`, as that's only one logical step away from the intention, but not plain `size()` because that implicit int -> bool conversion is surprising to the reader, like every implicit conversion that converts to something that's not logically "the same thing". I think the readers mind goes like this:
    1. if the size of responses. wtf, that doesn't make sense
    2. Ah, implicit int -> bool conversion, so "if the size of the responses is not zero".
    3. That's "if the responses are not empty", i.e. "if there are responses"

    Sorry for the long comment, I hope this makes sense.

2. Sorry for the accusation, you're right about the second point. I must have been salty because of something when writing this. :(

    However, this raises another problem, which is that I, as the reader, misunderstood the code, so it's probably too complex. I suggest extracting the lambda and maybe even the consume call, so that the identifiers can guide the reader.

3. \-

```suggestion
    if (responses.size() > 0) {  // or (!responses.empty()), or (not responses.empty())
      const auto on_dequeue = [this](C2Payload&& payload) { extractPayload(std::move(payload)); };
      const auto dequeue_success = responses.consumeWaitFor(on_dequeue, std::chrono::seconds{1});
      if (!dequeue_success) {
```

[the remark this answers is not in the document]
hunyadi-dev: 1.
> Call empty instead of checking size() against zero. 

Of course I would do that, my problem is with the hidden bangs when reading code :) Unfortunately, the current linter setting warns on `! empty()` which would make it at least stand out. Maybe it is just personal because I have committed incorrect code due to not noticing a `!`, but always felt reading `if(vector.size())` as `if vector has elements`.

tldr; will change it

2.
Will extract the logic out.

3.
I dont want to commit that comment on, can I somehow edit the suggestion? :)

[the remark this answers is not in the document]
szaszm: I don't think you can edit my suggestion (in my previous comment), but you can manually apply the desired changes in your working directory, or I can edit my comment for you if you specify what changes do you want.
Do you want to keep the current suggestion except the comment, or change something else, too?

libminifi/src/c2/C2Agent.cpp
szaszm: The second set of try/catch blocks are outside of the `for_each` lambda

libminifi/include/c2/C2Agent.h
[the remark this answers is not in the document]
hunyadi-dev: Replaced with include.

libminifi/src/c2/C2Agent.cpp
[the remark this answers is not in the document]
hunyadi-dev: Corrected as requested.

libminifi/src/c2/C2Agent.cpp
hunyadi-dev: What is the recommendation here?

szaszm: I understand that the indefinite waiting on the mutex is because of the use of `MinifiConcurrentQueue`. But why did we introduce a waiting on the queue condition variable for elements?

I just want an explanation, to see that the change was intentional and why.

hunyadi-dev: I just thought that if one second waiting time was reasonable here, than I should keep this if-it-works-in-one-second-it-is-good behaviour. Maybe there should be a timeout on the mutex itself, but I wonder what solution sounds reasonable here?

szaszm: In the old ideal (and likely common) case before, we got the mutex instantly, took whatever was in the queue and continued execution.

In the new common case, we lock the mutex, take whatever is in the queue, wait 1 second for more to arrive and continue execution.

The common case execution time is increased by almost 1 second, which is probably orders of magnitude more than how much time is spent executing business logic. I'd like to get rid of that waiting if it's not needed. I suggest using `consume` from `ConcurrentQueue` without waiting on the condition variable. This way if we assume little waiting for the mutex, the execution time stays similar.

szaszm:
https://github.com/apache/nifi-minifi-cpp/pull/776#discussion_r431809781

As described here, the average producer execution time has increased to 1 second and I think also the heartbeat interval is effectively restricted to at most once a second in absence of other C2 requests, but I didn't test this observation. 

Until this is fixed (or explained why my observation is wrong) I can't +1 this PR.

arpadboda:
> [#776 (comment)](https://github.com/apache/nifi-minifi-cpp/pull/776#discussion_r431809781)
> 
> As described here, the average producer execution time has increased to 1 second and I think also the heartbeat interval is effectively restricted to at most once a second in absence of other C2 requests, but I didn't test this observation.
> 
> Until this is fixed (or explained why my observation is wrong) I can't +1 this PR.

Good point, I agree that we should simply process whatever is there without waiting. 

The wait_until interface of the concurrentqueue is still useful, so I would keep that, just don't use in the particular case.

PR #777, opened 2020-05-06
MINIFICPP-1216 - Controller Services Integration test is unstable

arpadboda:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/src/FlowController.cpp
+++ b/libminifi/src/FlowController.cpp
@@ -242,15 +242,19 @@ int16_t FlowController::stop(bool force, uint64_t timeToWait) {
     logger_->log_info("Stop Flow Controller");
     if (this->root_)
       this->root_->stopProcessing(timer_scheduler_, event_scheduler_, cron_scheduler_);
-    this->flow_file_repo_->stop();
-    this->provenance_repo_->stop();
     // stop after we've attempted to stop the processors.
     this->timer_scheduler_->stop();
     this->event_scheduler_->stop();
     this->cron_scheduler_->stop();
+    thread_pool_.shutdown();
+    /* STOP! Before you change it, consider the following:
+     * -Stopping the schedulers doesn't actually quit the onTrigger functions of processors
+     * -They only guarantee that the processors are not scheduled any more
+     * -After the threadpool is stopped we can make sure that processors don't need repos and controllers anymore */
+    this->flow_file_repo_->stop();
+    this->provenance_repo_->stop();
     // stop the ControllerServices
     this->controller_service_provider_->disableAllControllerServices();
-    thread_pool_.shutdown();
     running_ = false;
   }
   return 0;


libminifi/src/FlowController.cpp
```diff
@@ -248,9 +248,9 @@ int16_t FlowController::stop(bool force, uint64_t timeToWait) {
     this->timer_scheduler_->stop();
     this->event_scheduler_->stop();
     this->cron_scheduler_->stop();
+    thread_pool_.shutdown();
```
arpadboda, line 251: So the reason behind:
When the schedulers above  are stopped, they don't actually stop the processors they schedule, only remove all waiting tasks of those from the threadpool. 
This means that ontrigger calls can actually be in process by the time we start disabling all controller services. Which seems to be a potential risk to me. I guess this introduced the test case flicker related to mocked controller services, for eg in this CI execution:
https://travis-ci.org/github/apache/nifi-minifi-cpp/jobs/683790379

In my opinion controller services have no dependency on the threadpool of flowcontroller (they don't even get reference to flowcontroller, just the configuration as far as I see), so shutting down the threadpool and disabling controller services only at the end seems to be the safety choice.

@bakaid  as you authored the original change to stop controller services here, what do you think?

bakaid: @arpadboda During development of the new state storage at one point I remember shutting down the thread pool before disabling the controller services. While it was my gut feeling that there should be no requirement for the thread pool to be running (nor did I found one), nevertheless it caused the shutdown to fail and the whole agent to freeze on C2 reload for example.

I was operating under the assumption that `this->root_->stopProcessing` will stop the processors. If that's not the case, `this->flow_file_repo_->stop();` and `this->provenance_repo_->stop();` that also currently come before the processors' running onTriggers could end would also be problematic (in the same manner as disabling the controller services: it's not a nullptr exception but the last onTrigger will not be able to use them).

This forced `disableAllControllerServices` was done because we leak resources on flow reload and we can't have two RocksDbPersistableKeyValueStoreServices running at the same time, because the first locks the db: the new flow's one will fail. This is just a workaround, not a proper solution: a proper solution would be to refactor and fix flow reload.

If `disableAllControllerServices` can run fine without the thread pool (again, it failed once, but it might have been just one state inbetween thread pool and C2 refactors that came and went while I was doing work on the state PR), then I agree that it would be the safer option to so, but this change has to be tested thoroughly on C2 reload and shutdown conditions.

Also, in this case looking into the repo shutdowns would also make sense.

arpadboda: Your point related to the repos absolutely makes sense, I moved those after shutdown of threadpool.

C2 tasks now have a dedicated threadpool, so I think such issues shoudn't occur. 
I did some tests to reconfigure the agent via C2 and flowcontroller restarts always succeeded, so I hope this works at least as a temporary solution.

bakaid: The litmus test is having the same stateful processor configured both in the old and the new flow - if on C2 reload it picks up the old state properly, it should likely be OK.

szaszm: Given the above discussion, the changes look reasonable to me. However, the discussion revealed that this code has numerous hidden interconnections: processors may run after initiating stop => both repositories and controller services need to be stopped after the thread pool, controller services need to be stopped because of db locking, affected C2 which I don't understand in detail.

Could you add a code comment mentioning the existence of these to the future reader/developer? While not all of it is strictly in the scope of this fix, I think the value/effort ratio would be outstanding here, making it worth doing IMO.

arpadboda: @szaszm : added a comment to describe what's going to
@bakaid : created a getfile->putfile flow, put one file to the source dir, it was written to the dest dir. Deleted it from the dest dir, updated the flow, created a new source file (differentt name) and only the new was written to the dest dir, so persistency was properly saved and loaded during flow upgrade.

bakaid: @arpadboda GetFile is not a stateful processor, it just deletes the files it has already read so it is not a proper test for this.

arpadboda: @bakaid : fair point. 
Verified with tailfile, added a new line a non-empty source file after publishing new flow. It was picked up properly, so tailing continued where it had to.

PR #778, opened 2020-05-11
MINIFICPP-1218 remove unimplemented declaration

szaszm:
Trivial change at line 338. The C2-related PRs introduced this, so stable APIs are not affected. Tests unaffected on clang/linux. (Success except for [MINIFICPP-1172](https://issues.apache.org/jira/browse/MINIFICPP-1172)-related.)

Also changed the declarations in the file to adhere to [C.128: Virtual functions should specify exactly one of `virtual`, `override`, or `final`](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c128-virtual-functions-should-specify-exactly-one-of-virtual-override-or-final) from the C++ Core Guidelines and a few other similar changes. Let me know if I should exclude these.

-----

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/FlowController.h
+++ b/libminifi/include/FlowController.h
@@ -21,6 +21,7 @@
 #define __FLOW_CONTROLLER_H__
 
 #include <stdio.h>
+#include <utility>
 #include <vector>
 #include <queue>
 #include <map>
@@ -68,21 +69,22 @@ class FlowController : public core::controller::ControllerServiceProvider, publi
    * Flow controller constructor
    */
   explicit FlowController(std::shared_ptr<core::Repository> provenance_repo, std::shared_ptr<core::Repository> flow_file_repo, std::shared_ptr<Configure> configure,
-                          std::unique_ptr<core::FlowConfiguration> flow_configuration, std::shared_ptr<core::ContentRepository> content_repo, const std::string name, bool headless_mode);
+                          std::unique_ptr<core::FlowConfiguration> flow_configuration, std::shared_ptr<core::ContentRepository> content_repo, std::string name, bool headless_mode);
 
   explicit FlowController(std::shared_ptr<core::Repository> provenance_repo, std::shared_ptr<core::Repository> flow_file_repo, std::shared_ptr<Configure> configure,
                           std::unique_ptr<core::FlowConfiguration> flow_configuration, std::shared_ptr<core::ContentRepository> content_repo)
-      : FlowController(provenance_repo, flow_file_repo, configure, std::move(flow_configuration), content_repo, DEFAULT_ROOT_GROUP_NAME, false) {
+      : FlowController(std::move(provenance_repo), std::move(flow_file_repo), std::move(configure), std::move(flow_configuration), std::move(content_repo), DEFAULT_ROOT_GROUP_NAME, false) {
   }
 
   explicit FlowController(std::shared_ptr<core::Repository> provenance_repo, std::shared_ptr<core::Repository> flow_file_repo, std::shared_ptr<Configure> configure,
                           std::unique_ptr<core::FlowConfiguration> flow_configuration)
-      : FlowController(provenance_repo, flow_file_repo, configure, std::move(flow_configuration), std::make_shared<core::repository::FileSystemRepository>(), DEFAULT_ROOT_GROUP_NAME, false) {
-    content_repo_->initialize(configure);
+      : FlowController(std::move(provenance_repo), std::move(flow_file_repo), std::move(configure), std::move(flow_configuration),
+          std::make_shared<core::repository::FileSystemRepository>(), DEFAULT_ROOT_GROUP_NAME, false) {
+    content_repo_->initialize(configuration_);
   }
 
   // Destructor
-  virtual ~FlowController();
+  ~FlowController() override;
 
   // Get the provenance repository
   virtual std::shared_ptr<core::Repository> getProvenanceRepository() {
@@ -98,7 +100,7 @@ class FlowController : public core::controller::ControllerServiceProvider, publi
   virtual void load(const std::shared_ptr<core::ProcessGroup> &root = nullptr, bool reload = false);
 
   // Whether the Flow Controller is start running
-  virtual bool isRunning() {
+  bool isRunning() override {
     return running_.load() || updating_.load();
   }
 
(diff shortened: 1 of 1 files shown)


libminifi/include/FlowController.h
```diff
@@ -68,21 +69,22 @@ class FlowController : public core::controller::ControllerServiceProvider, publi
    * Flow controller constructor
    */
   explicit FlowController(std::shared_ptr<core::Repository> provenance_repo, std::shared_ptr<core::Repository> flow_file_repo, std::shared_ptr<Configure> configure,
-                          std::unique_ptr<core::FlowConfiguration> flow_configuration, std::shared_ptr<core::ContentRepository> content_repo, const std::string name, bool headless_mode);
+                          std::unique_ptr<core::FlowConfiguration> flow_configuration, std::shared_ptr<core::ContentRepository> content_repo, std::string name, bool headless_mode);
```
fgerlits, line 72: this should be updated in the cpp file, too

szaszm: No, this was intentional. Top-level `const` in the parameter list is ignored in the declaration, so I prefer no-`const` in declaration and `const` in the definition (when possible).

szaszm:
> personally, I would prefer the 3 logical units (remove unimplemented method, virtual -> override, add moves) of this change to be in 3 separate commits inside the PR -- that would help when looking at the history later

I think the rationale is that normally a PR should be one logical unit, even if larger. This is part of the apache workflow. https://cwiki.apache.org/confluence/display/NIFI/Contributor+Guide#ContributorGuide-OpeningaPullRequest(PR)totheNiFiproject

I realize that this PR consists of 3 logical units, but they are trivial IMO and it was easier to submit them this way. Because of this, the rationale no longer applies and I was just blindly following the rule. I can change the PR is this is a problem, but the changes seem too trivial to me be worth the effort.

libminifi/include/FlowController.h
```diff
@@ -107,42 +109,39 @@ class FlowController : public core::controller::ControllerServiceProvider, publi
...
+  virtual void waitUnload(uint64_t timeToWaitMs);
   // Unload the current flow xml, clean the root process group and all its children
   virtual void unload();
   // Load new xml
   virtual void reload(std::string yamlFile);
   // update property value
   void updatePropertyValue(std::string processorName, std::string propertyName, std::string propertyValue) {
     if (root_ != nullptr)
-      root_->updatePropertyValue(processorName, propertyName, propertyValue);
+      root_->updatePropertyValue(std::move(processorName), std::move(propertyName), std::move(propertyValue));
```
adamdebreceni, line 139: (i) I would defer the string copy as long as possible, especially when they are passed on unmodified

adamdebreceni, line 139: (i) I would defer the string copy as long as possible, especially when they are passed on unmodified

adamdebreceni: so rather than move, const ref parameters

fgerlits:
Yeah, not important for this small PR, but for future PRs we should relax or change the single commit recommendation.  I'm thinking of something like "Does your PR implement a single change in a small number of commits? If it has multiple logical parts, please consider submitting them in separate PRs."

libminifi/include/FlowController.h
```diff
@@ -198,125 +197,125 @@ class FlowController : public core::controller::ControllerServiceProvider, publi
...
    */
-  virtual std::vector<std::shared_ptr<core::controller::ControllerServiceNode>> disableReferencingServices(std::shared_ptr<core::controller::ControllerServiceNode> &serviceNode);
+  std::vector<std::shared_ptr<core::controller::ControllerServiceNode>> disableReferencingServices(std::shared_ptr<core::controller::ControllerServiceNode> &serviceNode) override;
 
   /**
    * Verify can enable referencing components
    * @param serviceNode service node whose referenced components will be scheduled.
    */
-  virtual void verifyCanEnableReferencingServices(std::shared_ptr<core::controller::ControllerServiceNode> &serviceNode);
+  void verifyCanEnableReferencingServices(std::shared_ptr<core::controller::ControllerServiceNode>&) override;
```
hunyadi-dev, line 262: As far as I understand, this function doesn't do anything.

libminifi/include/FlowController.h
```diff
@@ -198,125 +197,125 @@ class FlowController : public core::controller::ControllerServiceProvider, publi
...
    */
-  virtual std::vector<std::shared_ptr<core::controller::ControllerServiceNode>> unscheduleReferencingComponents(std::shared_ptr<core::controller::ControllerServiceNode> &serviceNode);
+  std::vector<std::shared_ptr<core::controller::ControllerServiceNode>> unscheduleReferencingComponents(std::shared_ptr<core::controller::ControllerServiceNode> &serviceNode) override;
 
   /**
    * Verify can disable referencing components
    * @param serviceNode service node whose referenced components will be scheduled.
    */
-  virtual void verifyCanDisableReferencingServices(std::shared_ptr<core::controller::ControllerServiceNode> &serviceNode);
+  void verifyCanDisableReferencingServices(std::shared_ptr<core::controller::ControllerServiceNode> &serviceNode) override;
```
hunyadi-dev, line 250: Minor, but could be `const` throughout the inheritance hierarchy.

libminifi/include/FlowController.h
```diff
@@ -98,7 +100,7 @@ class FlowController : public core::controller::ControllerServiceProvider, publi
   virtual void load(const std::shared_ptr<core::ProcessGroup> &root = nullptr, bool reload = false);
 
   // Whether the Flow Controller is start running
-  virtual bool isRunning() {
+  bool isRunning() override {
```
hunyadi-dev, line 103: Minor, but seem logically `const`.

szaszm: I can't fix the const-correctness of overridden virtual functions without changing more classes and potentially breaking API. I wanted to limit my unrelated refactorings to this file.

libminifi/include/FlowController.h
szaszm: I wanted to limit the scope of my extra changes to the originally affected file. Since this is an overridden virtual function, I didn't reason about possible changes here.

From what I can see, the class has a `ControllerServiceProvider` base and a member, and the overridden function delegates to the same function of the member instead of the base.

libminifi/include/FlowController.h
szaszm: I don't want to extend the scope of this PR to anything outside of this file. Currently `ProcessGroup::updatePropertyValue` takes the strings by value.

It would be nice if they went through the whole call graph as a const ref, but they are taken by value here and I don't want to change 2-3 more classes to fix this unrelated issue in this PR. A move makes more sense locally IMO.

Given the above, if you still want me to change this function to take its parameters by const ref, in anticipation of further future changes, I'm fine with doing so.

szaszm:
> Yeah, not important for this small PR, but for future PRs we should relax or change the single commit recommendation. I'm thinking of something like "Does your PR implement a single change in a small number of commits? If it has multiple logical parts, please consider submitting them in separate PRs."

I think your wording discourages PRs like this one where a file had a clear issue warranting a contribution, but the contributor happened to read through the file and make tiny refactorings here and there, which, although unrelated, may still provide positive value to the codebase IMO.

I tend to go through some static analyzer warnings while reading the code but generally leave these modifications in my working directory since my feeling is that "making 1-2 files adhere to 3 more guidelines" is not a strong enough reason to create an issue and PR. I submit them when I need to submit the file anyway, like in this PR.

Idea: "In your initial contributions, does each commit correspond to exactly one logical unit of change and vice versa?"
I'm also fine with leaving this as it is.

szaszm:
I've got plenty of reviews on this PR from new contributors, thank you all for that. If you think my reasoning in a resolved issue (comment thread) is not correct, missing a point, or any other reason you're not convinced about the resolved status of the issue, feel free to comment again on those threads.

I welcome clarifications and constructive debate.

PR #780, opened 2020-05-12
MINIFICPP-1220 fix memleak and clarify ownership semantics in CWEL

szaszm:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/windows-event-log/Bookmark.cpp
+++ b/extensions/windows-event-log/Bookmark.cpp
@@ -38,7 +38,7 @@ Bookmark::Bookmark(const std::wstring& channel, const std::wstring& query, const
   }
 
   if (!bookmarkXml_.empty()) {
-    if (hBookmark_ = EvtCreateBookmark(bookmarkXml_.c_str())) {
+    if (hBookmark_ = unique_evt_handle{ EvtCreateBookmark(bookmarkXml_.c_str()) }) {
       ok_ = true;
       return;
     }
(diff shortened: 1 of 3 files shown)


extensions/windows-event-log/Bookmark.cpp
```diff
@@ -50,56 +50,52 @@ Bookmark::Bookmark(const std::wstring& channel, const std::wstring& query, const
...
   }
 
-  DWORD dwReturned{};
-  EVT_HANDLE hEvent{};
-  if (!EvtNext(hEventResults, 1, &hEvent, INFINITE, 0, &dwReturned)) {
-    LOG_LAST_ERROR(EvtNext);
-    return;
-  }
+  const unique_evt_handle hEvent = [this,&hEventResults] {
+    DWORD dwReturned{};
```
arpadboda, line 70: Why do we create a lambda for only getting something once?

extensions/windows-event-log/Bookmark.cpp
```diff
@@ -50,56 +50,52 @@ Bookmark::Bookmark(const std::wstring& channel, const std::wstring& query, const
...
+    EVT_HANDLE hEvent{ nullptr };
+    if (!EvtNext(hEventResults.get(), 1, &hEvent, INFINITE, 0, &dwReturned)) {
+      LOG_LAST_ERROR(EvtNext);
+    }
+    return unique_evt_handle{ hEvent };
+  }();
+  if (!hEvent) { return; }
```
arpadboda, line 77: This feels needless to me. 
```
ok_ = hEvent && saveBookmark(hEvent.get()); 
```
Seems to do the job.

extensions/windows-event-log/Bookmark.h
```diff
@@ -21,20 +22,26 @@ class Bookmark
 public:
   Bookmark(const std::wstring& channel, const std::wstring& query, const std::string& bookmarkRootDir, const std::string& uuid, bool processOldEvents, std::shared_ptr<core::CoreComponentStateManager> state_manager, std::shared_ptr<logging::Logger> logger);
   ~Bookmark();
-  operator bool() const;
-  EVT_HANDLE getBookmarkHandleFromXML();
+  explicit operator bool() const noexcept;
+
+  /* non-owning */ EVT_HANDLE getBookmarkHandleFromXML();
```
arpadboda, line 27: Can it return const to indicate the lack of ownership?

extensions/windows-event-log/Bookmark.cpp
szaszm: To be able to make it const and hide the complexity of out params and the returned array of length 1.

https://www.bfilipek.com/2016/11/iife-for-complex-initialization.html

extensions/windows-event-log/Bookmark.h
```diff
@@ -21,20 +22,26 @@ class Bookmark
...
   bool getNewBookmarkXml(EVT_HANDLE hEvent, std::wstring& bookmarkXml);
   bool saveBookmarkXml(const std::wstring& bookmarkXml);
 private:
   bool saveBookmark(EVT_HANDLE hEvent);
   bool getBookmarkXmlFromFile(std::wstring& bookmarkXml);
 
+  struct evt_deleter {
+    void operator()(EVT_HANDLE) const noexcept;
+  };
+  using unique_evt_handle = std::unique_ptr<std::remove_pointer<EVT_HANDLE>::type, evt_deleter>;
```
fgerlits, line 37: This `unique_evt_handle` could be useful in all Windows-related code; is there some "Windows utilities" header it could be moved to?

extensions/windows-event-log/Bookmark.h
szaszm: How would a const handle indicate lack of ownership? Normally raw pointers should be non-owning and the thought here was that `EVT_HANDLE` is a type alias to `void*`, so I used it indicator of lack of ownership.

This is a bit problematic in our codebase since we can't return owning raw pointers without gsl (that would normally be `gsl::owner<T*>`), but with `unique_ptr` I didn't need it this time.

extensions/windows-event-log/Bookmark.h
szaszm: I'm not aware of any. This extension is the only one that handles windows events, so I think the proper place for this utility is somewhere inside the extension subdirectory.

I think it should be extracted from the scope as soon as we need it elsewhere in the extension, but until then, I think it's better to keep it in a smaller scope.

extensions/windows-event-log/Bookmark.h
arpadboda: Okay, makes sense. 

Another reason to bring GSL... :)

extensions/windows-event-log/Bookmark.h
arpadboda: Agreed. Currently this is the only Wundows-specific codepart we have.

We might get back to create some Windows-untils in case further Win-specific features are needed.

extensions/windows-event-log/Bookmark.cpp
szaszm: good point, resolved in 6e93dc7

arpadboda:
Looks good, will run some tests before merging.

PR #781, opened 2020-05-13
MINIFICPP-1214: Converts H2O Processors to use ALv2 compliant H20-3 library

james94:
Short summary:

I removed the 3 processors that use Driverless AI MOJO Scoring Pipeline and Python Scoring Pipeline. I replaced these 3 processors with 1 processor that uses the H2O-3 MOJO Model. So, this new processor uses the **ALv2 compliant H2O-3 library** resolving JIRA issue: [MINIFICPP-1214](https://issues.apache.org/jira/browse/MINIFICPP-1214).

1 H2O-3 MOJO Model Processor Added:
- h2o/h2o3/mojo/**ExecuteH2oMojoScoring.py**

3 Driverless AI Scoring Pipeline Processors Removed:
- h2o/dai/psp/H2oPspScoreBatches.py
- h2o/dai/psp/H2oPspScoreRealTime.py
- h2o/dai/msp/H2oMojoPwScoring.py

I have created a YouTube video on how to use this new MiNiFi processor **ExecuteH2oMojoScoring.py** within the MiNiFi Data Flow: 

- ML Model Deployment and Scoring on an AWS EC2 node with H2O3 AutoML and MiNiFi Data Flow: https://youtu.be/Wm7OCrF-FAM

Long summary:

**Added py processor to make predictions with h2o3 model**

Replaced previous py processors that use Driverless AI models to make predictions
with a new py processor that uses open source ALv2 h2o3 library to make predictions.
This new py processor uses the h2o3 MOJO model to do batch scoring or real-time
scoring. All the user needs to run this processor is the filepath of h2o3 MOJO model
and pass that filepath into the 'MOJO Model Filetpath' minifi processor property.
The other required property 'Is First Line Header' is set to True by default. There
are 3 optional properties: 'Input Schema', 'Use Output Header', 'Output Schema'.
This processor makes use of the h2o's open source datatable library licensed under
Mozilla Public License 2.0. Datatable is used to load in the flow file content.
Then that datatable frame is converted to a numpy array. Then the numpy array and
input column names are passed to H2OFrame(np, col_names) to create h2oframe. Then
the h2o3 MOJO model is used to make predictions on the h2oframe and the result is
stored into a new h2oframe. That frame is then converted to pandas frame, which is
stored into the outgoing flow file content. Thus, the predictions are stored in
the outgoing flow file content.

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/pythonprocessors/h2o/h2o3/mojo/ExecuteH2oMojoScoring.py
```diff
@@ -0,0 +1,165 @@
...
+    conda config --append channels conda-forge
+    conda install -y -c h2oai h2o
+
+    Option 2: Install H2O-3 with pip
+
+    pip install requests
+    pip install tabulate
+    pip install "colorama>=0.3.8"
+    pip install future
+    pip uninstall h2o
```
phrocker, line 46: Why must h2o be uninstalled? is that an instruction from the directions of H2O-3?

extensions/pythonprocessors/h2o/h2o3/mojo/ExecuteH2oMojoScoring.py
```diff
@@ -0,0 +1,165 @@
...
+    processor.addProperty("Output Schema", "To set Output Schema, 'Use Output Header' must be set to 'True' \
+        If you want more descriptive column names for your predictions, then add an Output Schema. If there \
+        is more than one column name, write a comma separated list of column names. Else, H2O-3 will include \
+        them by default", "", False, False)
+
+def onSchedule(context):
+    """ onSchedule is where you load and read properties
+        this function is called 1 time when the processor is scheduled to run
+    """
+    global mojo_model
```
phrocker, line 94: Is it the case that this applies to all processors instances that may be using this?

phrocker:
@james94 I will take a closer look after some meetings. Thanks for the quick follow up. I will run through your instructions and run the test. Thanks again for the contribution and sorry for the mailing list confusions ( which were likely caused by me being a bit too peanut buttered ) -- community always encourages contributions!

extensions/pythonprocessors/h2o/h2o3/mojo/ExecuteH2oMojoScoring.py
james94: the uninstall line is there in case a user has an older version of h2o, then they uninstall it, and install the latest version of h2o. Yes this instruction is from the directions of H2O-3.

http://docs.h2o.ai/h2o/latest-stable/h2o-docs/downloading.html

extensions/pythonprocessors/h2o/h2o3/mojo/ExecuteH2oMojoScoring.py
james94: I use global for mojo_model, so I can access mojo_model in the onSchdule() function and onTrigger() function. In onSchedule(), I specify global since the mojo_model will change in this function and my intention is to instantiate a mojo_model object 1 time right at the start when the processor is scheduled to run. This case applies to all processor instances. Then in the onTrigger(), we use the mojo_model to make predictions. 

For example, one processor instance could instantiate a classification mojo_model while another processor could instantiate a regression mojo_model. Then we have one processor that is making classification predictions while the other processor is making regression predictions.

Would each processor has its own unique mojo_model object or no since I am using global?

james94:
@phrocker thank you! I am looking forward to your feedback.

You are welcome. I am happy to help and contribute new functionality to MiNiFi CPP. No worries on the mailing list confusions, I did check the mailing list and replied to certain messages, so thank you for the reminder too.

szaszm:
I have a concern regarding the contribution of these processors. Myself and probably most minifi c++ contributors are unfamiliar with H2O, which makes it problematic to maintain the new processors. 

The first problem I ran into while reviewing #784 is that I couldn't easily verify whether those changes would break these new processors.

This problem (and others) could be mostly solved if you could submit integration tests covering these processors. Could you do that?

phrocker:
> I have a concern regarding the contribution of these processors. Myself and probably most minifi c++ contributors are unfamiliar with H2O, which makes it problematic to maintain the new processors.
> 
> The first problem I ran into while reviewing #784 is that I couldn't easily verify whether those changes would break these new processors.
> 
> This problem (and others) could be mostly solved if you could submit integration tests covering these processors. Could you do that?

It took me a while to get around to testing this. My hope on the original merge would be that I would create a ticket and create a test for this, but I'm coming around to the fact that I'm too busy to do such things. I agree with what @szaszm is saying, especially in regards to having regression tests that would make this easier to merge. I think this solves the licensing issue; however, some tests would be super useful if possible. pytest-mock would be a useful testing framework to explore and should make this a pretty low water mark to solve this.

james94:
@phrocker and @szaszm  would writing regression tests for this new python processor be similar approach to writing regression test for nifi processor, but instead of using nifi-mock, we use pytest-mock with mutiple test cases toward this processor? Are integration tests the same as regression tests?

There are two python processors, I would need to write regression tests for:

- **ExecuteH2oMojoScoring.py**
- **ConvertDsToCsv.py**

james94:
@phrocker @szaszm I created a Jira Ticket for creating regression tests for the new H2O Processors using pytest-mock framework: https://issues.apache.org/jira/browse/MINIFICPP-1233

I will start working on regression tests for the following two processors:

- **ExecuteH2oMojoScoring.py**
- **ConvertDsToCsv.py**

How should I approach integration tests for [#784](https://github.com/apache/nifi-minifi-cpp/pull/784) and these two new H2O Processors? It sounds like @phrocker already tested integration?

szaszm:
I may have used the wrong terminology, I'm definitely not an expert when it comes to software QA. I meant that we need some kind of test, be it unit or integration or whatever. We would definitely use it to detect regressions.

I assume Marc tested your processors manually. I'll write the rest of my comment on the Jira ticket.

phrocker:
> 
> 
> @phrocker @szaszm I created a Jira Ticket for creating regression tests for the new H2O Processors using pytest-mock framework: https://issues.apache.org/jira/browse/MINIFICPP-1233
> 
> I will start working on regression tests for the following two processors:
> 
>     * **ExecuteH2oMojoScoring.py**
> 
>     * **ConvertDsToCsv.py**
> 
> 
> How should I approach integration tests for [#784](https://github.com/apache/nifi-minifi-cpp/pull/784) and these two new H2O Processors? It sounds like @phrocker already tested integration?

James I tested the processors manually. In regards to testing these processors I'd focus more on validating your logic and if there are any external services mock those out. I'll take a look at @szaszm's comments in the ticket.

 If you can create some tests we can create a way to run them.

szaszm:
I think we can merge this now if it works, i.e. this shouldn't be blocked by MINIFICPP-1233.
My point is that in order to be able to maintain the code, we will need automated test coverage. 

I referred to existing similar tests in our repo as an example [in a comment on the Jira issue](https://issues.apache.org/jira/browse/MINIFICPP-1233?focusedCommentId=17113180&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17113180), but I don't know how to proceed further with the implementation, so I can't provide much help there. I will try to help by reviewing the corresponding pull request when it's ready or by answering further questions.

phrocker:
Thanks @szaszm  . I don't have my keys so I'll merge with a signed commit tomorrow. I'll take another quick pass through and respond with some ideas on the follow on ticket. thanks to you and @james94  !

james94:
@szaszm and @phrocker I will let you know as soon as I have some test scripts for these h2o processors. I will follow up soon on them.

Thanks for your guidance.

PR #784, opened 2020-05-14
MINIFICPP-1206 - Rework and test ExecutePythonProcessor, add in-place script support

hunyadi-dev:
Rework and test ExecutePythonProcessor, add in-place script support

Some changes will be added in different PRs:
 - https://issues.apache.org/jira/browse/MINIFICPP-1222
 - https://issues.apache.org/jira/browse/MINIFICPP-1223
 - https://issues.apache.org/jira/browse/MINIFICPP-1224

---
### :a: uto generated text, please do not waste your time reading this:

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/test/script-tests/ExecutePythonProcessorTests.cpp
```diff
@@ -0,0 +1,276 @@
...
+      return temp_dir;
+    }
+
+    std::string getFileContent(const std::string& file_name) {
+      std::ifstream file_handle(file_name);
+      REQUIRE(file_handle.is_open());
+      const std::string file_content{ (std::istreambuf_iterator<char>(file_handle)), (std::istreambuf_iterator<char>())};
+      file_handle.close();
+      return file_content;
+    }
```
szaszm, line 98: These are generic enough that they could go to `FileUtils` IMO.

I'd remove the `std::ifstream::close` calls because they are both right before `return`, after which the file is closed anyway. They don't hurt but they're not needed either.

If we handle errors in some way, we shouldn't ignore them IMO. I mean if we check for a good file stream, we should throw if it's not good.

libminifi/test/script-tests/ExecutePythonProcessorTests.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+    std::string getScriptFullPath(const std::string& script_file_name) {
+      return SCRIPT_FILES_DIRECTORY + utils::file::FileUtils::get_separator() + script_file_name;
+    }
+
+    const std::string TEST_FILE_NAME{ "test_file.txt" };
+    const std::string TEST_FILE_CONTENT{ "Test text\n" };
+    const std::string SCRIPT_FILES_DIRECTORY{ "test_scripts" };
```
szaszm, line 106: I think we only need one object of these, not one every instance. They should go outside of the class, become `static`, or possibly both (with `static` referring to linkage in this case).

libminifi/test/script-tests/ExecutePythonProcessorTests.cpp
```diff
@@ -0,0 +1,276 @@
...
+TEST_CASE_METHOD(SimplePythonFlowFileTransferTest, "Simple file passthrough", "[executePythonProcessorSimple]") {
+  // Expectations
+  const auto OUTPUT_FILE_MATCHES_INPUT          = SimplePythonFlowFileTransferTest::Expectation::OUTPUT_FILE_MATCHES_INPUT;
+  const auto RUNTIME_RELATIONSHIP_EXCEPTION    = SimplePythonFlowFileTransferTest::Expectation::RUNTIME_RELATIONSHIP_EXCEPTION;
+  const auto PROCESSOR_INITIALIZATION_EXCEPTION = SimplePythonFlowFileTransferTest::Expectation::PROCESSOR_INITIALIZATION_EXCEPTION;
+  // ExecutePython outbound relationships
+  const core::Relationship SUCCESS {"success", "description"};
+  const core::Relationship FAILURE{"failure", "description"};
+
+  // For the tests "" is threated as none-provided since no optional implementation was ported to the project yet
```
szaszm, line 231: typo: s/threated/treated/

libminifi/test/script-tests/ExecutePythonProcessorTests.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+  ////////////////////////////////////////////////////////////
+  // 0. Neither valid script file nor script body provided  //
+  ////////////////////////////////////////////////////////////
+
+  //                                          TEST EXPECTATION  OUT_REL        USE_AS_SCRIPT_FILE  USE_AS_SCRIPT_BODY
+  testSimpleFilePassthrough(PROCESSOR_INITIALIZATION_EXCEPTION, SUCCESS,                       "", "");  // NOLINT
```
szaszm, line 238: Comment blocks like this are a bad sign IMO. I like the content in this case, just not the formatting, because it distracts attention from the code to the comment.
```suggestion
  // Neither valid script file nor script body provided
  //                                          TEST EXPECTATION  OUT_REL        USE_AS_SCRIPT_FILE  USE_AS_SCRIPT_BODY
  testSimpleFilePassthrough(PROCESSOR_INITIALIZATION_EXCEPTION, SUCCESS,                       "", "");  // NOLINT
```

extensions/script/python/ExecutePythonProcessor.cpp
```diff
@@ -46,144 +50,177 @@ core::Relationship ExecutePythonProcessor::Failure("failure", "Script failures")
...
-    try {
-      engine->evalFile(prop);
-      auto me = shared_from_this();
-      triggerDescribe(engine, me);
-      triggerInitialize(engine, me);
+  valid_init_ = false;
+  appendPathForImportModules();
+  loadScript();
+  try {
+    if ("" != script_to_exec_) {
```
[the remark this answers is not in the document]
hunyadi-dev: I think this is more expressive, as it shows we are handling an `std::string` as opposed to `std::vector<std::string>`. Also even on reading this comment I missed the bang on the first option, and `not` is warned on (unreliably though) by the linter.

extensions/script/python/ExecutePythonProcessor.cpp
```diff
@@ -46,144 +50,177 @@ core::Relationship ExecutePythonProcessor::Failure("failure", "Script failures")
...
-      }
-
-      if (!script_file_.empty()) {
-        engine->evalFile(script_file_);
-      } else {
-        throw std::runtime_error("No Script File is available to execute");
-      }
+    // TODO(hunyadi): When using "Script File" property, we currently re-read the script file content every time the processor is on schedule. This should change to single-read when we release 1.0.0
+    // https://issues.apache.org/jira/browse/MINIFICPP-1223
+    reloadScriptIfUsingScriptFileProperty();
```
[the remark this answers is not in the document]
hunyadi-dev: 1) I am happy to do the changes for this. I will raise this as a question the next time we have a team meeting.

2) Good idea, but this change should have its own Jira.

extensions/script/python/ExecutePythonProcessor.cpp
```diff
@@ -46,144 +50,177 @@ core::Relationship ExecutePythonProcessor::Failure("failure", "Script failures")
...
+    logger_->log_info("Destroying script engine because it is no longer needed");
+  }
+}
+
+void ExecutePythonProcessor::appendPathForImportModules() {
+  // TODO(hunyadi): I have spent some time trying to figure out pybind11, but
+  // could not get this working yet. It is up to be implemented later
+  // https://issues.apache.org/jira/browse/MINIFICPP-1224
+  if ("" != module_directory_) {
+    logger_ -> log_error("Not supported property: Module Directory.");
```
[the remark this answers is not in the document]
hunyadi-dev: Good catch. Unfortunately as much as I try to fit the code I write to context, this is how I instinctively write `->` operators. If you see one I forgot to correct please feel free to point it out :)

extensions/script/python/ExecutePythonProcessor.cpp
```diff
@@ -46,144 +50,177 @@ core::Relationship ExecutePythonProcessor::Failure("failure", "Script failures")
...
+}
+
+void ExecutePythonProcessor::loadScriptFromFile(const std::string& file_path) {
+  std::ifstream file_handle(file_path);
+  if (!file_handle.is_open()) {
+    script_to_exec_ = "";
+    throw std::runtime_error("Failed to read Script File: " + file_path);
+  }
+  script_to_exec_ = std::string{ (std::istreambuf_iterator<char>(file_handle)), (std::istreambuf_iterator<char>())};
+  file_handle.close();
```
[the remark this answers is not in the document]
hunyadi-dev: Will add extra padding and remove `close()` call.

libminifi/test/script-tests/CMakeLists.txt
```diff
@@ -19,6 +19,13 @@
 
 if (NOT DISABLE_PYTHON_SCRIPTING)
 	file(GLOB EXECUTESCRIPT_PYTHON_INTEGRATION_TESTS  "Python*.cpp")
+	file(GLOB EXECUTEPYTHONPROCESSOR_UNIT_TESTS  "ExecutePythonProcessorTests.cpp")
+	file(GLOB PY_SOURCES  "python/*.cpp")
+	find_package(PythonLibs 3.5)
+	if (NOT PYTHONLIBS_FOUND)
+		find_package(PythonLibs 3.0 REQUIRED)
+	endif()
```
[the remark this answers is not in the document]
hunyadi-dev: This behaviour is copied from the script extension cmake recipe at `extensions/script/CMakeLists.txt`. We should either correct both of them or have both of them fall back to `python 3.0`.

libminifi/test/script-tests/ExecutePythonProcessorTests.cpp
```diff
@@ -0,0 +1,276 @@
...
+#include "processors/GetFile.h"
+#include "python/ExecutePythonProcessor.h"
+#include "processors/LogAttribute.h"
+#include "processors/PutFile.h"
+#include "utils/file/FileUtils.h"
+
+namespace {
+
+#include <unistd.h>
+#define GetCurrentDir getcwd
```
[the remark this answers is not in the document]
hunyadi-dev: You are right, this is useless as it is.

This was meant to be
```c++
#ifdef WIN32
#include <direct.h>
#define GetCurrentDir _getcwd
#else
#include <unistd.h>
#define GetCurrentDir getcwd
#endif
```

Background:
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/getcwd?view=vs-2019

libminifi/test/script-tests/ExecutePythonProcessorTests.cpp
hunyadi-dev: 1) Sure, but I would prefer to have them in some kind of `TestUtils.h` due to depending on `REQUIRE` macro. Do you want me to create such a file? I can also add the same definition to `FileUtils.h` with a throwing version if that is what you ask in 3).

2) Those will be removed.

3) I am not sure if I understand, does reply 1) apply here?

libminifi/test/script-tests/ExecutePythonProcessorTests.cpp
hunyadi-dev: I will change this to static members sacrificing some readibility.

libminifi/test/script-tests/ExecutePythonProcessorTests.cpp
```diff
@@ -0,0 +1,276 @@
...
+// +-----------+     +--------------------------+       +-----------+
+// |  Getfile  |   .-+  ExecutePythonProcessor  |     .-+  PutFile  |
+// +-----------+  /  +  -  -  -  -  -  -  -  -  +    /  +-----------+
+// |  success  +-°   |    Attribute: Script     |   /   |  success  +-+ checked
+// +-----------+     +--------------------------+  /    +-----------+
+//                   |         success          +-°
+//                   +--------------------------+
+//                   |         failure          +-X either success or failure is hooked up
+//                   +--------------------------+
+//
```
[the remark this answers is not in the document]
hunyadi-dev: Grr, ASCII art... Who drew this? :D

Will remove it.

libminifi/test/script-tests/ExecutePythonProcessorTests.cpp
hunyadi-dev: Good catch

libminifi/test/script-tests/ExecutePythonProcessorTests.cpp
hunyadi-dev: Will edit this manually, as I don't see it here if he titles are correctly aligned now.

libminifi/test/script-tests/ExecutePythonProcessorTests.cpp
```diff
@@ -0,0 +1,276 @@
...
+  testSimpleFilePassthrough(     OUTPUT_FILE_MATCHES_INPUT, SUCCESS, "", "passthrough_processor_transfering_to_success.py");  // NOLINT 
+  testSimpleFilePassthrough(RUNTIME_RELATIONSHIP_EXCEPTION, FAILURE, "", "passthrough_processor_transfering_to_success.py");  // NOLINT 
+  testSimpleFilePassthrough(RUNTIME_RELATIONSHIP_EXCEPTION, SUCCESS, "", "passthrough_processor_transfering_to_failure.py");  // NOLINT 
+  testSimpleFilePassthrough(     OUTPUT_FILE_MATCHES_INPUT, FAILURE, "", "passthrough_processor_transfering_to_failure.py");  // NOLINT 
+  testSimpleFilePassthrough(RUNTIME_RELATIONSHIP_EXCEPTION, FAILURE, "",                   "non_transferring_processor.py");  // NOLINT
+
+  ////////////////////////////////
+  // 3. Setting both attributes //
+  ////////////////////////////////
+
```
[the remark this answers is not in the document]
hunyadi-dev: Will be corrected.

libminifi/test/script-tests/PythonExecuteScriptTests.cpp
```diff
@@ -29,6 +29,15 @@
...
 #include "processors/PutFile.h"
 
+// ,-------------------------------------------------------------,
+// | ! |                      Disclaimer                     | ! |
+// |---'                                                     '---'
+// |                                                             |
+// | This file contains tests for the "ExecuteScript" processor, |
+// |         not for the "ExecutePython" processor.              |
+// |                                                             |
+// '-------------------------------------------------------------'
```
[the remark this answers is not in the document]
hunyadi-dev: Ah, I liked this thought, it reminded me that I am looking at the wrong file when checking tests :) Will replace this and rename the file to an ugly-long-named-one instead so that it will be harder to confuse this file With `ExecutePythonProcessorTests.cpp` :)

extensions/script/python/ExecutePythonProcessor.cpp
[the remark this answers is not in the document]
hunyadi-dev: Jira added here: https://issues.apache.org/jira/browse/MINIFICPP-1240

libminifi/test/script-tests/ExecutePythonProcessorTests.cpp
szaszm: 1. I'd replace REQUIRE with check + throw, so that they're reusable and the control flow is the same (tests fail on uncaught exceptions).

3. The first function only operates on the file it the open was successful. If not, there is no indication of the error to the caller. Since there is already some kind of check, I recommend throwing an exception in the else branch.

libminifi/test/script-tests/ExecutePythonProcessorTests.cpp
szaszm: How does making them `static` decrease readability?
If you mean the added `static` keyword increasing the line length, I think the added information of shared constant objects balances that. An alternative is extracting them from the class and putting them in an anonymous namespace, keeping the lines short and the scope limited to the translation unit.

hunyadi-dev: I mean the initialization of this static member goes outside the class, so if people check the declaration they won't immediately see these values.

libminifi/test/script-tests/CMakeLists.txt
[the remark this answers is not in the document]
szaszm: Ok, then leave as is for now. Thanks for the clarification.

libminifi/test/script-tests/ExecutePythonProcessorTests.cpp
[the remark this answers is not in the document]
szaszm:
```
#include <unistd.h>
```
shouldn't be inside an anonymous namespace. We shouldn't mess with the linkage of declarations of system headers.

libminifi/test/script-tests/ExecutePythonProcessorTests.cpp
hunyadi-dev: 1. I would argue that the workflow would not be any more reusable, as a having a `TestController` is a dependency either way. Also, I would rather have the REQUIRE assertions inside the test-helper function for two reasons:
 - Someone can easily forget adding EXPECT_THROW on the call side ending up accidentaly satisfying "EXPECT_THROW" calls on an upper call stack.
 - Adding the expectation for throwing would make the code would look like we test the helper functions inside of unit tests, and might not be obvious why is there an expectation for not throwing for a given test.

libminifi/test/script-tests/ExecutePythonProcessorTests.cpp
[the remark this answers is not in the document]
hunyadi-dev: Yes, I already said I have realized that this was there as a bug as well :)

Screenshot 2020-05-28 at 16 10 46

libminifi/test/script-tests/PythonExecuteScriptTests.cpp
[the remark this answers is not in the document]
hunyadi-dev: Renamed the tests.

libminifi/test/script-tests/ExecutePythonProcessorTests.cpp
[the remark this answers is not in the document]
hunyadi-dev: Removed comment decoration.

libminifi/test/script-tests/ExecutePythonProcessorTests.cpp
szaszm: That's right. It's a good idea to keep them near the class implementation, but I would keep them inside the class as they are just constants.

libminifi/test/script-tests/ExecutePythonProcessorTests.cpp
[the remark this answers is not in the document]
szaszm: Sry, I didn't fully understand that part at a first reading.

libminifi/test/script-tests/ExecutePythonProcessorTests.cpp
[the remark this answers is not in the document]
james94: You could re-design your minifi flow diagram in draw.io or a drawing software tool of your choice and save it as a. jpg file and provide it as a reference for the test program. I think it would helpful. Is there a place to put assets?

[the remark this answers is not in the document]
szaszm: We discussed this with Adam and he convinced me that it provides value to the reader, so I'm neutral, but it's removed already. I don't think we have a place to put such assets at the moment, and it would be an overkill to write separate docs for this test only IMO.

So my current opinion: text comment == ASCII art comment > separate file
If you both find it useful, we should probably revert to the ASCII art version.

libminifi/include/utils/TestUtils.h
```diff
@@ -0,0 +1,80 @@
...
+#define GetCurrentDir getcwd
+#endif
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace utils {
+
+std::string getCurrentWorkingDir(void) {
```
arpadboda, line 37: This exists in FileUtils or pathutils

libminifi/include/utils/TestUtils.h
```diff
@@ -0,0 +1,80 @@
...
+  char buff[FILENAME_MAX];
+  GetCurrentDir(buff, FILENAME_MAX);
+  std::string current_working_dir(buff);
+  return current_working_dir;
+}
+
+std::string createTempDir(TestController* testController) {
```
arpadboda, line 44: Why do we need this wrapper? Especially why do we do with asserts.

libminifi/include/utils/TestUtils.h
```diff
@@ -0,0 +1,80 @@
...
+  char dirtemplate[] = "/tmp/gt.XXXXXX";
+  std::string temp_dir = testController->createTempDirectory(dirtemplate);
+  REQUIRE(!temp_dir.empty());
+  struct stat buffer;
+  REQUIRE(-1 != stat(temp_dir.c_str(), &buffer));
```
arpadboda, line 49: Using stat this way is most probably not platform-independent. 
On the other side, is_dir function exists in fileutils to avoid this.

extensions/script/python/ExecutePythonProcessor.cpp
```diff
@@ -35,7 +35,11 @@ namespace python {
 namespace processors {
 
 core::Property ExecutePythonProcessor::ScriptFile("Script File",  // NOLINT
-    R"(Path to script file to execute)", "");
+    R"(Path to script file to execute.
+                                            Only one of Script File or Script Body may be used)", "");
+core::Property ExecutePythonProcessor::ScriptBody("Script Body",  // NOLINT
+    R"(Script to execute.
+                                            Only one of Script File or Script Body may be used)", "");
```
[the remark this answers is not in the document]
hunyadi-dev: I do not know, they were split by the original design. I was not sure if this is a hack used for displaying the definition in the EFM, so I left the split in place.

libminifi/include/utils/TestUtils.h
hunyadi-dev: I wanted a wrapper that does not require a directory template, but asserts on the new directory being generated.

libminifi/include/utils/TestUtils.h
hunyadi-dev: I cannot find it there:

```bash
doxygen -g && perl -pi -e "s/GENERATE_(HTML|LATEX)(\s*)=.*$/GENERATE_\1\2= NO/g; s/GENERATE_MAN(\s*)=.*$/GENERATE_MAN\1= YES/g; s;INPUT(\s*)=.*$;INPUT\1=$(find . -name FileUtils.h);g" Doxyfile && doxygen && man ./man/man3/*
```
Screenshot 2020-06-04 at 17 08 52

extensions/script/python/ExecutePythonProcessor.cpp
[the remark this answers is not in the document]
hunyadi-dev: This is the state I restored from the original documentation. I have no knowledge on why the whitespaces might be useful.

libminifi/include/utils/TestUtils.h
arpadboda: Misguided you a bit, sorry!
It's called:
```
utils::Environment::getCurrentWorkingDirectory()
```

extensions/script/python/ExecutePythonProcessor.cpp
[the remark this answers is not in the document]
arpadboda: I don't think it's helpful at all, no other processors has this weird line breaks in descriptions.

libminifi/include/utils/TestUtils.h
arpadboda: Can it fail anyhow?
I feel that in this case we verify our own testutils, which seems like a bit overprotective.
A lot of tests are going to fail in case we can't create tempdirs.

extensions/script/python/ExecutePythonProcessor.cpp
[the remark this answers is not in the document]
hunyadi-dev: Checked if this pattern is recurring:
```bash
 find . -name '*.cpp' | xargs grep -Hn -A2 "core::Property" | grep -v '\-\-' | perl -p -e "s/^(.*?)[:-]([0-9]+).*$/\1 \2/g" | xargs -n2 sh -c 'perl -ne "print "\""$1:"\"" . \$_ if $. == $2" $1' sh | grep '\"' | less | awk -F \" '!(NF % 2)'
./extensions/script/ExecuteScript.cpp:                                         R"(Path to script file to execute.
./extensions/script/ExecuteScript.cpp:                                            Only one of Script File or Script Body may be used)", "");
./extensions/script/ExecuteScript.cpp:                                         R"(Body of script to execute.
./extensions/script/ExecuteScript.cpp:                                            Only one of Script File or Script Body may be used)", "");
./extensions/script/ExecuteScript.cpp:                                              R"(Comma-separated list of paths to files and/or directories which
./extensions/script/ExecuteScript.cpp:                                                 contain modules required by the script)", "");
./extensions/script/python/ExecutePythonProcessor.cpp:    R"(Path to script file to execute.
./extensions/script/python/ExecutePythonProcessor.cpp:                                            Only one of Script File or Script Body may be used)", "");
./extensions/script/python/ExecutePythonProcessor.cpp:    R"(Script to execute.
./extensions/script/python/ExecutePythonProcessor.cpp:                                            Only one of Script File or Script Body may be used)", "");
./extensions/script/python/ExecutePythonProcessor.cpp:    R"(Comma-separated list of paths to files and/or directories which
./extensions/script/python/ExecutePythonProcessor.cpp:                                                 contain modules required by the script)", "");
```
Seems like the only two files having this pattern is `ExecuteScript` and `ExecutePythonProcessor.cpp`. This is probably due to the boilerplate of `ExecutePythonProcessor` was originally copied from `ExecuteScript.cpp. Should I remove both?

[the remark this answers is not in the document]
arpadboda: Yes, please!

libminifi/include/utils/TestUtils.h
hunyadi-dev: The main issue, is that `createTempDirectory` swallows whatever is returned from `boost::filesystem::create_directory()`
Relevant code:
```c++
create_dir(tempDirectory);  << RETURN VALUE NOT HANDLED
return tempDirectory;
```
From `create_dir()`:
```c++
if(boost::filesystem::create_directory(dir))
{
  return 0;
}
else
{
  return -1;
}
```
Comparison of the interfaces:
| | FileUtils::create_temp_directory() | TestController::createTempDirectory() | utils::createTempDir() |
| ------------- | ------------- | ------------- |  ------------- |
| Needs format? | Yes | Yes | ✅ NO|
| Checks if dir creation successful? | No | No | ✅ Yes|
| Cleans after itself? | No | ✅ Yes | ✅ Yes |

libminifi/include/utils/TestUtils.h
hunyadi-dev: Changed to the FileUtils version.

extensions/script/python/ExecutePythonProcessor.cpp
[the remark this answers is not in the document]
hunyadi-dev: Due to [performance impact](https://stackoverflow.com/questions/62081977/why-is-there-no-optimization-for-checking-for-empty-string-via-comparison) I went with replacing `"" ==` and `"" !=` with `.empty()` and `.size()` calls respectively.

extensions/script/python/ExecutePythonProcessor.cpp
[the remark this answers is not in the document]
hunyadi-dev: Replaced.

libminifi/include/utils/TestUtils.h
hunyadi-dev: Thanks for the suggestion, I removed the redundant function.

extensions/script/python/ExecutePythonProcessor.cpp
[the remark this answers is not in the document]
szaszm: The comment should still be removed from here (but left in place in `onTrigger`). Rereading in `onSchedule` is not a problem at all and we shouldn't ever change the behavior in `onSchedule`.

For clarification: `onSchedule` is called when we need to (re)configure the processor (rarely). `onTrigger` is called on processor execution, i.e. often.

extensions/script/python/ExecutePythonProcessor.cpp
```diff
@@ -35,155 +35,188 @@ namespace python {
...
-      if (engine == nullptr) {
-        throw std::runtime_error("No script engine available");
-      }
-
-      if (!script_file_.empty()) {
-        engine->evalFile(script_file_);
-      } else {
-        throw std::runtime_error("No Script File is available to execute");
-      }
+    // TODO(hunyadi): When using "Script File" property, we currently re-read the script file content every time the processor is on schedule. This should change to single-read when we release 1.0.0
```
szaszm, line 124: `s/on schedule/triggered/`

extensions/script/python/ExecutePythonProcessor.cpp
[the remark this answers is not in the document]
szaszm: Not yet resolved

libminifi/test/script-tests/ExecutePythonProcessorTests.cpp
```diff
@@ -0,0 +1,220 @@
...
+      REQUIRE(TEST_FILE_CONTENT == output_file_content);
+    }
+  }
+  void testsStatefulProcessor() {
+    reInitialize();
+    const std::string output_dir = createTempDir(testController_.get());
+
+    auto executePythonProcessor = plan_->addProcessor("ExecutePythonProcessor", "executePythonProcessor");
+    plan_->setProperty(executePythonProcessor, org::apache::nifi::minifi::python::processors::ExecutePythonProcessor::ScriptFile.getName(), getScriptFullPath("stateful_processor.py"));
+    executePythonProcessor -> initialize();
```
szaszm, line 123: There should be no spaces around `->`

libminifi/test/script-tests/ExecutePythonProcessorTests.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+  std::shared_ptr<core::Processor> addExecutePythonProcessorToPlan(const std::string& used_as_script_file, const std::string& used_as_script_body) {
+    auto executePythonProcessor = plan_->addProcessor("ExecutePythonProcessor", "executePythonProcessor", core::Relationship("success", "description"), true);
+    if ("" != used_as_script_file) {
+      plan_->setProperty(executePythonProcessor, org::apache::nifi::minifi::python::processors::ExecutePythonProcessor::ScriptFile.getName(), getScriptFullPath(used_as_script_file));
+    }
+    if ("" != used_as_script_body) {
+        plan_->setProperty(executePythonProcessor, org::apache::nifi::minifi::python::processors::ExecutePythonProcessor::ScriptBody.getName(), getFileContent(getScriptFullPath(used_as_script_body)));
+    }
+    executePythonProcessor -> initialize();
```
szaszm, line 156: There should be no spaces around `->`

libminifi/test/script-tests/ExecutePythonProcessorTests.cpp
hunyadi-dev: Padding removed.

libminifi/test/script-tests/ExecutePythonProcessorTests.cpp
hunyadi-dev: Padding removed.

extensions/script/python/ExecutePythonProcessor.cpp
hunyadi-dev: Comment corrected.

extensions/script/python/ExecutePythonProcessor.cpp
```diff
@@ -35,155 +35,185 @@ namespace python {
...
+
+}
+
+void ExecutePythonProcessor::loadScriptFromFile(const std::string& file_path) {
+  std::ifstream file_handle(file_path);
+  if (!file_handle.is_open()) {
+    script_to_exec_ = "";
+    throw std::runtime_error("Failed to read Script File: " + file_path);
+  }
+  script_to_exec_ = std::string{ (std::istreambuf_iterator<char>(file_handle)), (std::istreambuf_iterator<char>())};
```
szaszm, line 185: Spacing inside the braces are inconsistent.

hunyadi-dev: Added extra padding.

libminifi/include/utils/TestUtils.h
```diff
@@ -0,0 +1,80 @@
...
+  REQUIRE(S_ISDIR(buffer.st_mode));
+  return temp_dir;
+}
+
+std::string putFileToDir(const std::string& dir_path, const std::string& file_name, const std::string& content) {
+  std::string file_path(dir_path + utils::file::FileUtils::get_separator() + file_name);
+  std::ofstream out_file(file_path);
```
fgerlits, line 41: I would write these files in binary mode.  In text mode (which is the default), the contents of the file will be different in Windows and Linux/Mac because of line endings, which will complicate some tests.

libminifi/include/utils/TestUtils.h
```diff
@@ -0,0 +1,80 @@
...
+  if (out_file.is_open()) {
+    out_file << content;
+  }
+  return file_path;
+}
+
+std::string createTempDirWithFile(TestController* testController, const std::string& file_name, const std::string& content) {
+  std::string temp_dir = createTempDir(testController);
+  putFileToDir(temp_dir, file_name, content);
+  return temp_dir;
```
fgerlits, line 51: I would return the full path (ie. temp_dir + separator + file_name), as I expect that is what most users of this function will need.  I understand you can get the full path from temp_dir more easily than the other way round, but if a user needs the name of the directory, they can use `createTempDir` + `putFileToDir` instead.

hunyadi-dev: I think this was exactly what I needed: the name of directory with a single file in it, in order to have the processor `GetFile` exactly one target to run on. Maybe I should rename the function to indicate that this is giving back the path of a temp dir with a file?

libminifi/include/utils/TestUtils.h
hunyadi-dev: Updated.

libminifi/include/utils/TestUtils.h
fgerlits: No, I think one version is enough; in either case, one can call `createTempDir` + `putFileToDir` and use the return value of either of these.

My reason for asking was that in my TailFile tests, I have a similar local utility function (and in those tests, I need the full path), and I could replace my local function with `createTempDirWithFile` if it returned the full path, too.

I think we don't have enough data to decide which is the more common use case, so I'm fine with leaving it as is; after some time, if we find that the TailFileTests kind of usage is the more common one, we can change it.

extensions/script/python/ExecutePythonProcessor.cpp
```diff
@@ -35,155 +35,185 @@ namespace python {
...
+
+void ExecutePythonProcessor::appendPathForImportModules() {
+  // TODO(hunyadi): I have spent some time trying to figure out pybind11, but
+  // could not get this working yet. It is up to be implemented later
+  // https://issues.apache.org/jira/browse/MINIFICPP-1224
+  if (module_directory_.size()) {
+    logger_->log_error("Not supported property: Module Directory.");
+  }
+
+}
```
szaszm, line 177: Redundant empty line near the end of the function body. The linter would most likely warn about this after your other PR is merged. :smile:

libminifi/include/utils/TestUtils.h
```diff
@@ -0,0 +1,67 @@
...
+std::string createTempDir(TestController* testController) {
+  char dirtemplate[] = "/tmp/gt.XXXXXX";
+  std::string temp_dir = testController->createTempDirectory(dirtemplate);
+  REQUIRE(!temp_dir.empty());
+  REQUIRE(file::FileUtils::is_directory(temp_dir.c_str()));
+  return temp_dir;
+}
+
+std::string putFileToDir(const std::string& dir_path, const std::string& file_name, const std::string& content) {
+  std::string file_path(dir_path + file::FileUtils::get_separator() + file_name);
```
adamdebreceni, line 42: `file::FileUtils::concat_path` might be used here

extensions/script/python/ExecutePythonProcessor.cpp
```diff
@@ -35,155 +35,184 @@ namespace python {
 namespace processors {
 
 core::Property ExecutePythonProcessor::ScriptFile("Script File",  // NOLINT
-    R"(Path to script file to execute)", "");
+    R"(Path to script file to execute. Only one of Script File or Script Body may be used)", "");
+core::Property ExecutePythonProcessor::ScriptBody("Script Body",  // NOLINT
```
arpadboda, line 39: Using property builder (withdescription, blah blah) can help breaking these lines without linter errors. And improve readability as well. 
You can also specify some to be required.

extensions/script/python/ExecutePythonProcessor.cpp
```diff
@@ -35,155 +35,184 @@ namespace python {
...
+      throw std::runtime_error("Only one of Script File or Script Body may be used");
+    }
+    loadScriptFromFile(script_file);
+    return;
+  }
+  if (script_body.size()) {
+    script_to_exec_ = script_body;
+    return;
+  }
+  throw std::runtime_error("Neither Script Body nor Script File is available to execute");
```
arpadboda, line 203: Not a big deal, but I prefer a structure of the function that throws in error branches.

```
if (script_body.size() == 0) {
  throw std::runtime_error("Neither Script Body nor Script File is available to execute");
}
script_to_exec_ = script_body;
```

This way it seems to be shorter and better structured to me.

extensions/script/python/ExecutePythonProcessor.h
```diff
@@ -18,12 +18,16 @@
...
 
-#include <concurrentqueue.h>
-#include <core/Resource.h>
-#include <core/Processor.h>
+#include <memory>
+#include <string>
+#include <vector>
+
+#include <concurrentqueue.h> // NOLINT
+#include <core/Resource.h> // NOLINT
```
arpadboda, line 29: Linter detects these as C headers because  "<" is used instead quotes.

libminifi/test/script-tests/ExecutePythonProcessorTests.cpp
```diff
@@ -0,0 +1,220 @@
...
+    addPutFileProcessorToPlan(core::Relationship("success", "description"), output_dir);
+    plan_->runNextProcessor();  // ExecutePythonProcessor
+    for (std::size_t i = 0; i < 10; ++i) {
+      plan_->runCurrentProcessor();  // ExecutePythonProcessor
+    }
+    plan_->runNextProcessor();  // PutFile
+    for (std::size_t i = 0; i < 10; ++i) {
+      plan_->runCurrentProcessor();  // PutFile
+      const std::string state_name = std::to_string(i);
+      const std::string output_file_path = output_dir + utils::file::FileUtils::get_separator() + state_name;
```
arpadboda, line 134: There is ```concat_path``` in FileUtils.h.

extensions/script/python/ExecutePythonProcessor.h
hunyadi-dev: I replaced the `<>` with "", but for concurrentqueue, one gets a different error:
> 28:1  error  cpplint  Include the directory when naming .h files  [build/include_subdir] [4]

arpadboda: I would simply disable that error, makes no sense to me.

extensions/script/python/ExecutePythonProcessor.cpp
hunyadi-dev: Restructured.

libminifi/include/utils/TestUtils.h
hunyadi-dev: As requested.

extensions/script/python/ExecutePythonProcessor.cpp
hunyadi-dev: Removed.

extensions/script/python/ExecutePythonProcessor.cpp
hunyadi-dev: Replaced.

extensions/script/python/ExecutePythonProcessor.h
hunyadi-dev: Noted :)

libminifi/test/script-tests/ExecutePythonProcessorTests.cpp
hunyadi-dev: Replaced w/ fully qualified call to `concat_path`.

adamdebreceni: for future reference I think we should think about using a c++17 compatible `path` class, with all the goodies, because this fully qualified `concat_path` (although better than direct string concat) is still a mouthful, and IMO `path` and `std::string` shouldn't be implicitly convertible, the type should clearly represent its purpose (which `std::string path;` does not accomplish)

arpadboda: Good shout, I'm open to it. 
Preferably in a follow-up ticket as it's a general util, doesn't really belong to the scope of this PR.

PR #791, opened 2020-05-21
MINIFICPP-1177 Improvements to the TailFile processor

fgerlits:

### Description of PR

Fixed some bugs and implemented missing features in the TailFile processor:
- handle rotation of log files correctly -- instead of the dodgy logic of doing one file at a time in onTrigger and trying to keep track using the persisted state, we follow the same logic as NiFi: only look at rolled files if the input file got truncated, only look at new rotated files matching the pattern, and stream all new content in one onTrigger call
- find and pick up new files in Multiple file mode
- yield when no flow files were generated, and only then (used to fail in multiple and rotation cases)
 - include the delimiter at the end of the flow file, as this is what NiFi does and it would work better on Windows
- make '\n' the default delimiter (currently the default is no delimiter)
- support Recursive lookup in Multiple files mode
- support Lookup frequency in Multiple files mode (we do a lookup in every onTrigger at the moment)
- lots of minor cleanup, unit test fixes, and added more unit tests

Not done as part of this PR, as it was already very large; I have created the Jira MINIFICPP-1244 for this:
- support NiFi's Initial Start Position setting (Beginning of Time | Beginning of File | Current Time); we always use Beginning of File at the moment

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -120,56 +158,42 @@ void TailFile::onSchedule(const std::shared_ptr<core::ProcessContext> &context,
   std::string value;
 
   if (context->getProperty(Delimiter.getName(), value)) {
-    delimiter_ = value;
+    delimiter_ = parseDelimiter(value);
+  }
+
+  if (!context->getProperty(FileName.getName(), file_to_tail_)) {
+    throw minifi::Exception(ExceptionType::PROCESSOR_EXCEPTION, "File to Tail is a required property");
```
arpadboda, line 171: I would suggest using propertybuilder and make the property required, so this can't happen.

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -120,56 +158,42 @@ void TailFile::onSchedule(const std::shared_ptr<core::ProcessContext> &context,
...
-    throw minifi::Exception(ExceptionType::PROCESSOR_EXCEPTION, "File to Tail is a required property");
-  }
   if (mode == "Multiple file") {
-    // file is a regex
-    std::string base_dir;
-    if (!context->getProperty(BaseDirectory.getName(), base_dir)) {
+    tail_mode_ = Mode::MULTIPLE;
+
+    if (!context->getProperty(BaseDirectory.getName(), base_dir_)) {
       throw minifi::Exception(ExceptionType::PROCESSOR_EXCEPTION, "Base directory is required for multiple tail mode.");
```
arpadboda, line 181: This is fair, but I would also add a check here to make sure that the directory exists. 
As far as I remember we have such utility func in fileutils, so not a big change.

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -120,56 +158,42 @@ void TailFile::onSchedule(const std::shared_ptr<core::ProcessContext> &context,
...
+    // in multiple mode, we check for new/removed files in every onTrigger
 
   } else {
+    tail_mode_ = Mode::SINGLE;
+
     std::string fileLocation, fileName;
-    if (utils::file::PathUtils::getFileNameAndPath(file, fileLocation, fileName)) {
-      tail_states_.insert(std::make_pair(fileName, TailState { fileLocation, fileName, 0, 0 }));
+    if (utils::file::PathUtils::getFileNameAndPath(file_to_tail_, fileLocation, fileName)) {
+      tail_states_.emplace(fileName, TailState{fileLocation, fileName, 0, 0, 0, 0});
```
arpadboda, line 191: Are we sure about this? 
As far as I understand the change this means that whenever MiNiFi is restarted, the state is dropped and we start reading the file from the beginning.

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -200,52 +224,65 @@ void TailFile::parseStateFileLine(char *buf) {
...
-  value = trimRight(value);
+  key = utils::StringUtils::trimRight(key);
+  value = utils::StringUtils::trimRight(value);
 
   if (key == "FILENAME") {
     std::string fileLocation, fileName;
     if (utils::file::PathUtils::getFileNameAndPath(value, fileLocation, fileName)) {
       logger_->log_debug("State migration received path %s, file %s", fileLocation, fileName);
-      tail_states_.insert(std::make_pair(fileName, TailState { fileLocation, fileName, 0, 0 }));
+      state.insert(std::make_pair(fileName, TailState{fileLocation, fileName, 0, 0, 0, 0}));
```
arpadboda, line 240: Using emplace we can avoid making pair.

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -403,102 +403,141 @@ void TailFile::onTrigger(const std::shared_ptr<core::ProcessContext> &context, c
...
-    struct stat statbuf;
-
-    logger_->log_debug("Tailing file %s from %llu", fullPath, state.second.currentTailFilePosition_);
-    if (stat(fullPath.c_str(), &statbuf) == 0) {
-      if ((uint64_t) statbuf.st_size <= state.second.currentTailFilePosition_) {
-        logger_->log_trace("Current pos: %llu", state.second.currentTailFilePosition_);
-        logger_->log_trace("%s", "there are no new input for the current tail file");
-        context->yield();
-        return;
+    did_something |= processFile(context, session, state.first, state.second);
```
arpadboda, line 419: Wonder if we could use std::any_of here

extensions/standard-processors/processors/TailFile.cpp
fgerlits: this is old code just moved around, but good point, I'll change it

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -403,102 +403,141 @@ void TailFile::onTrigger(const std::shared_ptr<core::ProcessContext> &context, c
...
+                             const std::string &fileName,
+                             TailState &state) {
+    std::string full_file_name = state.fileNameWithPath();
+
+    bool did_something = false;
+
+    if (utils::file::FileUtils::file_size(full_file_name) < state.position_) {
+      std::vector<TailState> rotated_file_states = findRotatedFiles(state);
+      for (TailState &file_state : rotated_file_states) {
+        did_something |= processSingleFile(context, session, file_state.file_name_, file_state);
```
arpadboda, line 438: std::any_of here as well

extensions/standard-processors/processors/TailFile.cpp
fgerlits: hm, not sure...  do we want to refuse to start if the directory does not exist (yet)?  maybe it will be created later

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -403,102 +403,141 @@ void TailFile::onTrigger(const std::shared_ptr<core::ProcessContext> &context, c
...
-              + extension;
-          flowFile->updateKeyedAttribute(FILENAME, logName);
-          state.second.currentTailFilePosition_ += flowFile->getSize();
-          storeState(context);
-        }
+    did_something |= processSingleFile(context, session, fileName, state);
+    return did_something;
+  }
+
+  bool TailFile::processSingleFile(const std::shared_ptr<core::ProcessContext> &context,
```
arpadboda, line 448: The indendatation is strange here, either this line shouldn't be indented or most of the function should be a level deeper.

extensions/standard-processors/processors/TailFile.cpp
fgerlits: this will be overwritten in `recoverState()` if there is a persisted state, but I agree it looks misleading -- I'll think about how to make it clearer

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -403,102 +403,141 @@ void TailFile::onTrigger(const std::shared_ptr<core::ProcessContext> &context, c
...
+      flowFile->addKeyedAttribute(ABSOLUTE_PATH, full_file_name);
+      uint64_t checksum = state.checksum_;
+      session->import(full_file_name, flowFile, true, state.position_, &checksum);
+      session->transfer(flowFile, Success);
+      logger_->log_info("TailFile %s for %llu bytes", fileName, flowFile->getSize());
+      std::string logName = baseName + "." + std::to_string(state.position_) + "-" +
+                            std::to_string(state.position_ + flowFile->getSize()) + "."
+                            + extension;
+      flowFile->updateKeyedAttribute(FILENAME, logName);
+      state.position_ += flowFile->getSize();
```
arpadboda, line 503: Wonder if this could share some code with the loop above if we introduce a function. 
Just an idea, it might make it even worse :)

extensions/standard-processors/processors/TailFile.cpp
arpadboda: Okay, understood. I'm fine with just having a comment here that tells this.

extensions/standard-processors/processors/TailFile.cpp
arpadboda: OnSchedule throwing an exception results in being called after a while (30s by default as far as I remember).

So that doesn't mean giving it up completely, just trying later and only let ontrigger happen when at least we can list the dir.

extensions/standard-processors/processors/TailFile.cpp
fgerlits: done

extensions/standard-processors/processors/TailFile.cpp
fgerlits: done

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -204,52 +230,65 @@ void TailFile::parseStateFileLine(char *buf) {
...
+  }
 
+  logger_->log_debug("load state succeeded");
 
-bool TailFile::recoverState(const std::shared_ptr<core::ProcessContext>& context) {
-  bool state_load_success = false;
+  tail_states_ = std::move(new_tail_states);
+
+  // Save the state to the state manager
+  storeState(context);
```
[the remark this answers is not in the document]
fgerlits: I have just made some changes here to fix a bug (in Single file mode, the old state overwrote the File to Tail property if the property changed but the state wasn't cleared), so it makes more sense now than before. :)

In any case, `storeState()` is not expensive, as it only updates the StateManager, which has its own schedule for how often to write changes to the DB.

extensions/standard-processors/processors/TailFile.cpp
fgerlits: I have added a comment

extensions/standard-processors/processors/TailFile.cpp
fgerlits: fixed

extensions/standard-processors/processors/TailFile.cpp
fgerlits: `any_of` short-circuits, so I can't use it directly; I could do something like
```c++
  std::vector<bool> did_something(tail_states_.size());
  std::transform(tail_states_.begin(), tail_states_.end(), std::back_inserter(did_something),
      [this, &context, &session](std::map<std::string, TailState>::value_type &state) {
    return processFile(context, session, state.first, state.second);
  });

  if (std::none_of(did_something.begin(), did_something.end(), [](bool x){ return x; })) {
    yield();
  }
```
but I'm not sure that is better.  Would you prefer this?

extensions/standard-processors/processors/TailFile.cpp
fgerlits: I have removed all this `did_something` ugliness, and replaced it with `ProcessSession::existsFlowFileInRelationship(Success)`.

extensions/standard-processors/processors/TailFile.cpp
fgerlits: done

extensions/standard-processors/processors/TailFile.cpp
fgerlits: I think we should get rid of the empty delimiter case (which puts everything in the log file since the last read into a single flow file, and doesn't seem useful, nor is it supported in NiFi) at some point, probably in a major release.

Until then, I would keep it like this, both to remind ourselves that this needs to be removed, and to make the removal easier.

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -58,42 +47,108 @@
...
+        } else {
+          return "\\";
+        }
+      } else {
+        return input.substr(0, 1);
+      }
+    } else {
+      return "";
+    }
+  }
```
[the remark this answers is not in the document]
fgerlits: done

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -204,52 +249,78 @@ void TailFile::parseStateFileLine(char *buf) {
...
+      state.emplace(fileName, TailState{fileLocation, fileName, 0, 0, 0, 0});
     } else {
-      tail_states_.insert(std::make_pair(value, TailState { fileLocation, value, 0, 0 }));
+      state.emplace(value, TailState{fileLocation, value, 0, 0, 0, 0});
     }
   }
   if (key == "POSITION") {
     // for backwards compatibility
-    if (tail_states_.size() != 1) {
+    if (tail_states_.size() != (std::size_t) 1) {
```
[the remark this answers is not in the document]
fgerlits: done

libminifi/src/utils/file/PathUtils.cpp
```diff
@@ -84,6 +108,19 @@ std::string PathUtils::getFullPath(const std::string& path) {
...
+  std::string PathUtils::globToRegex(const std::string &glob) {
+    std::string glob_copy = glob;
+    replaceAll(glob_copy, ".", "\\.");
+    replaceAll(glob_copy, "*", ".*");
+    replaceAll(glob_copy, "?", ".");
+    return glob_copy;
+  }
+
+  std::string PathUtils::replacePlaceholderWithBaseName(const std::string &pattern, const std::string& base_name) {
+    static const std::string PLACEHOLDER = "${filename}";
```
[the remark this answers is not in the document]
fgerlits: I have moved `replaceOne` to `StringUtils` (and discovered that an almost identical `replaceAll` was already there), and inlined this `replacePlaceholderWithBaseName` function into TailFile

libminifi/src/utils/file/PathUtils.cpp
```diff
@@ -84,6 +108,19 @@ std::string PathUtils::getFullPath(const std::string& path) {
 #endif
 }
 
+  std::string PathUtils::globToRegex(const std::string &glob) {
+    std::string glob_copy = glob;
+    replaceAll(glob_copy, ".", "\\.");
+    replaceAll(glob_copy, "*", ".*");
+    replaceAll(glob_copy, "?", ".");
+    return glob_copy;
+  }
```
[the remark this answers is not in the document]
fgerlits: done

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -335,65 +387,45 @@ bool TailFile::storeState(const std::shared_ptr<core::ProcessContext>& context)
...
 
-    if (matchedFiles.size() < 1) {
-      logger_->log_debug("No newer files found in directory!");
-      return;
+  std::vector<TailState> matched_files;
+  auto collect_matching_files = [&](const std::string &path, const std::string &file_name) -> bool {
+    if (file_name != state.file_name_ && utils::Regex::matchesFullInput(pattern, file_name)) {
+      std::string full_file_name = path + utils::file::FileUtils::get_separator() + file_name;
+      uint64_t mtime = utils::file::FileUtils::last_write_time(full_file_name);
+      if (mtime >= state.timestamp_ / 1000) {
```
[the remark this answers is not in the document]
fgerlits: done

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -204,52 +249,78 @@ void TailFile::parseStateFileLine(char *buf) {
...
     const auto file = key.substr(strlen(POSITION_STR));
-    tail_states_[file].currentTailFilePosition_ = std::stoull(value);
+    state[file].position_ = std::stoull(value);
   }
 }
 
+bool TailFile::recoverState(const std::shared_ptr<core::ProcessContext>& context) {
+  std::map<std::string, TailState> new_tail_states;
+  bool state_load_success = getStateFromStateManager(new_tail_states) ||
+                            getStateFromLegacyStateFile(new_tail_states);
```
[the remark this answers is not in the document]
fgerlits: thanks

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -403,102 +435,136 @@ void TailFile::onTrigger(const std::shared_ptr<core::ProcessContext> &context, c
...
+      ffr->addKeyedAttribute(ABSOLUTE_PATH, full_file_name);
+      ffr->updateKeyedAttribute(FILENAME, logName);
+      session->transfer(ffr, Success);
+      state.position_ += ffr->getSize();
+      state.timestamp_ = getTimeMillis();
+      state.checksum_ = checksum;
+      storeState(context);
+    }
+  } else {
+    std::shared_ptr<FlowFileRecord> flowFile = std::static_pointer_cast<FlowFileRecord>(session->create());
```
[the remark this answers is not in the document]
fgerlits: I think it doesn't hurt to explicitly show that this is a `shared_ptr`.  That is implied by `static_pointer_cast`, but being explicit is better.

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -403,102 +435,136 @@ void TailFile::onTrigger(const std::shared_ptr<core::ProcessContext> &context, c
...
+    if (flowFile) {
```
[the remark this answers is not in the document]
fgerlits: I think the logic is clearer this way, and the indentation level is not too bad (yet).  If more conditionals or loops are added later, we can make this change at that time.

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -403,102 +435,136 @@ void TailFile::onTrigger(const std::shared_ptr<core::ProcessContext> &context, c
...
+  if (!delimiter_.empty()) {
+    char delim = delimiter_[0];
+    logger_->log_trace("Looking for delimiter 0x%X", delim);
+
+    std::vector<std::shared_ptr<FlowFileRecord>> flowFiles;
+    uint64_t checksum = state.checksum_;
+    session->import(full_file_name, flowFiles, state.position_, delim, true, &checksum);
+    logger_->log_info("%u flowfiles were received from TailFile input", flowFiles.size());
+
+    for (auto &ffr : flowFiles) {
```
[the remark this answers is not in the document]
fgerlits: I don't know -- if I was reading this code and it had `const auto &`, my mind would go "wait, what, how is this const? we will modify the flow file... ah OK, it's a const pointer to non-const".  Leaving it as `auto &` was an attempt to spare others this mental detour, but I understand it can create other mental detours. :)

I will change it if you want me to, just checking if you really do.

extensions/standard-processors/processors/TailFile.h
```diff
@@ -33,38 +33,40 @@ namespace nifi {
...
-} TailMatchedFileItem;
+  std::string fileNameWithPath() const {
+    return path_ + utils::file::FileUtils::get_separator() + file_name_;
+  }
 
+  std::string path_;
+  std::string file_name_;
+  uint64_t position_;
+  uint64_t timestamp_;
+  uint64_t mtime_;
```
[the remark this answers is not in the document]
fgerlits: I have moved `mtime_` out of `TailState` as it is not really part of the state, and I'm using an ad-hoc `{TailState, mtime}` struct in the one place where it is needed (`findRotatedFiles`).

I have also renamed `timestamp_` to `last_read_time_`.

libminifi/src/core/ProcessSession.cpp
```diff
@@ -469,6 +470,14 @@ void ProcessSession::import(std::string source, const std::shared_ptr<core::Flow
           invalidWrite = true;
         }
       }
+
+      io::BaseStream *stream = baseStream.get();
+      std::unique_ptr<io::CRCStream<io::BaseStream>> crcStream;
+      if (crc) {
+        crcStream = utils::make_unique<io::CRCStream<io::BaseStream>>(stream, *crc);
+        stream = crcStream.get();
+      }
```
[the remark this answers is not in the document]
fgerlits: I agree this is a bit nasty.  Let me think of a better way, too, and then we'll discuss it.

libminifi/src/core/ProcessSession.cpp
```diff
@@ -574,15 +589,16 @@ void ProcessSession::import(const std::string& source, std::vector<std::shared_p
           if (zlen < std::numeric_limits<int>::min() || zlen > std::numeric_limits<int>::max()) {
             logger_->log_error("narrowing conversion failed");
           }
-          const int len = zlen;
+          const bool foundDelimiter = (delimiterPos != end);
+          const int len = (includeDelimiter && foundDelimiter) ? zlen + 1 : zlen;
```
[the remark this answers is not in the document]
fgerlits: done

libminifi/src/utils/file/FileUtils.cpp
```diff
@@ -0,0 +1,58 @@
...
+    constexpr uint64_t BUFFER_SIZE = 4096u;
+    std::array<char, (std::size_t) BUFFER_SIZE> buffer;
+
+    std::ifstream stream{file_name, std::ios::in | std::ios::binary};
+
+    uint64_t checksum = 0;
+    uint64_t remaining_bytes_to_be_read = up_to_position;
+
+    while (stream && remaining_bytes_to_be_read > 0) {
+      using std::min;  // for min(), to work around https://docs.microsoft.com/en-us/windows/win32/multimedia/min
```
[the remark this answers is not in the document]
fgerlits: I've changed it to `(std::min)`.

libminifi/test/unit/FileUtilsTests.cpp
```diff
@@ -178,3 +179,157 @@ TEST_CASE("TestFileUtils::getFullPath", "[TestGetFullPath]") {
...
+  uint64_t timeAfterSecondWrite = getTimeMillis() / 1000;
+
+  uint64_t second_mtime = FileUtils::last_write_time(test_file);
+  REQUIRE(second_mtime >= first_mtime);
+  REQUIRE(second_mtime >= timeAfterFirstWrite);
+  REQUIRE(second_mtime <= timeAfterSecondWrite);
+
+  test_file_stream.close();
+  uint64_t third_mtime = FileUtils::last_write_time(test_file);
+  REQUIRE(third_mtime == second_mtime);
```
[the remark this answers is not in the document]
fgerlits: There is no write to `test_file` between the two calls to `FileUtils::last_write_time` (and the stream was flushed earlier), so the mtimes should be identical, I think.

extensions/standard-processors/tests/unit/TailFileTests.cpp
```diff
@@ -47,7 +47,33 @@ static std::string NEWLINE_FILE = ""  // NOLINT
...
+    std::string full_file_name = directory + utils::file::FileUtils::get_separator() + file_name;
+    std::remove(full_file_name.c_str());
+  }
+
+  void renameTempFile(const std::string &directory, const std::string &old_file_name, const std::string &new_file_name) {
+    std::string old_full_file_name = directory + utils::file::FileUtils::get_separator() + old_file_name;
+    std::string new_full_file_name = directory + utils::file::FileUtils::get_separator() + new_file_name;
+    rename(old_full_file_name.c_str(), new_full_file_name.c_str());
+  }
+}  // namespace
```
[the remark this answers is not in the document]
fgerlits: fixed

extensions/standard-processors/processors/TailFile.cpp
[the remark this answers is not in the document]
fgerlits: In some cases (migration from the old state file to the state manager, or when the File to Tail property changes in Single file mode) the state could have changed, so it is necessary to write it.  I could limit the write to those cases, but this seemed safer, and this part of the code only runs when the processor is (re-)scheduled, so a small performance gain didn't seem worth the extra complexity.

extensions/standard-processors/processors/TailFile.cpp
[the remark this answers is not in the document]
szaszm: The difference is clear to me, but I can only speak for myself. I'm OK with leaving it as is.

libminifi/src/core/ProcessSession.cpp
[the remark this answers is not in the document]
szaszm: I thought a bit about this. There is a function similar to `ProcessSession::import` called `ProcessSession::importFrom` which takes an input stream instead of a file name. It doesn't support FlowFile delimiters, though.

My idea is to extend `ProcessSession::importFrom` to support delimiters and use that with a `CRCStream` over a `FileStream`. If I understand correctly, we include every byte that we processed in the checksum, so the stream composition wouldn't change behavior, but correct me if I'm wrong.

extensions/standard-processors/processors/TailFile.cpp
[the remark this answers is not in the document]
adamdebreceni: I would agree with @fgerlits, I don't know how often we actually modify an `std::shared_ptr<FlowFile>`, but if it is negligible I would go as far to replace all these with a `FlowFilePtr` with more sensible (for me) pointer access logic, i.e. `const T& operator->() const` (and we can also add a `FlowFileConstPtr` when needed)

extensions/standard-processors/processors/TailFile.cpp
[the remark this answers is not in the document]
adamdebreceni: early return gets the worm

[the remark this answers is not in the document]
szaszm: To be clear, this is a question of style, and the goal is consensus. If I caused you to change anything with resentment, then that's a failure on my side. This is true for most of my suggestions, by the way.

My goal is not to avoid "too much" indentation, but to minimize the indentation level, unless a more important goal (e.g. immediately invoked lamba pattern for encapsulated side effects) warrants their use.

I consider every level of indentation a new context pushed onto the mental stack of the reader. The less context info needed to be kept in mind, the better. For this reason, I try to keep all if/for/lambda blocks as short as possible, so that for most code, less mental context needs to be kept in mind and more mental resources can be dedicated to understanding the specific business logic.

By handling the simple cases locally (like unsuccessful flow file creation at the point of error), the reader can avoid thinking about the possibility of the error, because it's "already handled" by the time he is at the business logic. When it's handled later, one needs to remember that the simple case in question (error in the example) was not yet handled.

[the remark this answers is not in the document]
fgerlits: Don't worry, I won't implement any suggestion that I think will make the code worse.
In this case, I think both versions have merit, and with @adebreceni 's +1 (at least I think it was a +1), I have made this tiny change.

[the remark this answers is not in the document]
adamdebreceni: yes I am in favor of early returns, sorry for the confusion 🙂

arpadboda:
@fgerlits : one of the tests failed in CI:
https://travis-ci.org/github/apache/nifi-minifi-cpp/jobs/697214111

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -335,170 +563,210 @@ bool TailFile::storeState(const std::shared_ptr<core::ProcessContext>& context)
...
       }
-      state.second.currentTailFileModificationTime_ = ((uint64_t) (statbuf.st_mtime) * 1000);
-    } else {
-      logger_->log_warn("Unable to stat file %s", fullPath);
     }
+
+    state = state_copy;
+    storeState(context);
+
+    logger_->log_info("%u flowfiles were received from TailFile input", num_flow_files);
```
szaszm, line 702: Please use `zu` format specifier for `size_t`

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -335,170 +563,210 @@ bool TailFile::storeState(const std::shared_ptr<core::ProcessContext>& context)
...
+
+    storeState(context);
+  }
+}
+
+void TailFile::updateFlowFileAttributes(const std::string &full_file_name, const TailState &state,
+                                        const std::string &fileName, const std::string &baseName,
+                                        const std::string &extension,
+                                        std::shared_ptr<FlowFileRecord> &flow_file) const {
+  logger_->log_info("TailFile %s for %u bytes", fileName, flow_file->getSize());
```
szaszm, line 721: Please use `PRIu64` format specifier for `uint64_t`.

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -335,170 +563,210 @@ bool TailFile::storeState(const std::shared_ptr<core::ProcessContext>& context)
...
+                                 const std::shared_ptr<core::ProcessSession> &session,
+                                 const std::string &full_file_name,
+                                 TailState &state) {
+  std::string fileName = state.file_name_;
+
+  if (utils::file::FileUtils::file_size(full_file_name) == 0u) {
+    logger_->log_warn("Unable to read file %s as it does not exist or has size zero", full_file_name);
+    return;
   }
+  logger_->log_debug("Tailing file %s from %llu", full_file_name, state.position_);
```
szaszm, line 669: Please use `PRIu64` format specifier for `uint64_t`

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -335,170 +563,210 @@ bool TailFile::storeState(const std::shared_ptr<core::ProcessContext>& context)
...
+void TailFile::onTrigger(const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSession> &session) {
+  std::lock_guard<std::mutex> tail_lock(tail_file_mutex_);
 
-    // Going ahead in the file rolled over
-    if (file.current_file_name_ != base_file_name) {
-      logger_->log_debug("Resetting posotion since %s != %s", base_file_name, file.current_file_name_);
-      file.currentTailFilePosition_ = 0;
+  if (tail_mode_ == Mode::MULTIPLE) {
+    if (last_multifile_lookup_ + lookup_frequency_ < std::chrono::steady_clock::now()) {
+      logger_->log_debug("Lookup frequency %d ms have elapsed, doing new multifile lookup", lookup_frequency_.count());
```
szaszm, line 618: Printing chrono underlying types for the correct format specifier is tricky. Please see my comment here, a previous similar case: https://github.com/apache/nifi-minifi-cpp/pull/741#discussion_r385725765

Suggestion: wrap the count call in `uint64_t{}` and use `PRIu64` as the format specifier.

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -40,60 +35,283 @@
...
+      found_delimiter = (delimiter_pos != end);
+
+      ptrdiff_t zlen{std::distance(begin, delimiter_pos)};
+      if (found_delimiter) {
+        zlen += 1;
+      }
+      if (zlen < (std::numeric_limits<int>::min)() || zlen > (std::numeric_limits<int>::max)()) {
+        logger_->log_error("narrowing conversion failed");
+      }
+      const int len = zlen;
```
szaszm, line 224: A very similar code fragment was changed recently in my GSL PR. #805
```suggestion
      const int len = gsl::narrow<int>(std::distance(begin, delimiter_pos) + (found_delimiter ? 1 : 0));
```

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -40,60 +35,283 @@
...
+        input_stream_.read(reinterpret_cast<char *>(buffer.data()), buffer.size());
+
+        std::streamsize num_bytes_read = input_stream_.gcount();
+        logger_->log_trace("Read %d bytes of input", num_bytes_read);
+
+        begin = buffer.data();
+        end = begin + num_bytes_read;
+      }
+
+      uint8_t *delimiter_pos = std::find(begin, end, input_delimiter_);
```
szaszm, line 214: Signed/unsigned comparisons are happening on this line. I suggest explicitly converting `input_delimiter_` to `uint8_t` instead.
```suggestion
      uint8_t *delimiter_pos = std::find(begin, end, gsl::narrow_cast<uint8_t>(input_delimiter_));
```

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -40,60 +35,283 @@
...
+  logger->log_debug("Opening %s", file_name);
+  input_stream.open(file_name.c_str(), std::fstream::in | std::fstream::binary);
+  if (!input_stream.is_open() || !input_stream.good()) {
+    input_stream.close();
+    throw Exception(FILE_OPERATION_EXCEPTION, "Could not open file: " + file_name);
+  }
+  if (offset != 0U) {
+    input_stream.seekg(offset, std::ifstream::beg);
+    if (!input_stream.good()) {
+      logger->log_error("Seeking to %lu failed for file %s (does file/filesystem support seeking?)", offset, file_name);
```
szaszm, line 179: Please use `PRIu64` as the format specifier for `uint64_t`.

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -40,60 +35,283 @@
...
+
+    uint64_t num_bytes_written = 0;
+    bool found_delimiter = false;
+
+    while (hasMoreToRead() && !found_delimiter) {
+      if (begin == end) {
+        input_stream_.read(reinterpret_cast<char *>(buffer.data()), buffer.size());
+
+        std::streamsize num_bytes_read = input_stream_.gcount();
+        logger_->log_trace("Read %d bytes of input", num_bytes_read);
```
szaszm, line 208: `std::streamsize` is an unspecified signed integral type used for sizes. I think we can assume that whatever value it holds can be represented by a `std::ptrdiff_t`, which has a defined format specifier.
```suggestion
        const auto num_bytes_read = input_stream_.gcount();
        logger_->log_trace("Read %td bytes of input", std::ptrdiff_t{num_bytes_read});
```

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -40,60 +35,283 @@
...
+
+    io::CRCStream<io::BaseStream> crc_stream{output_stream.get(), checksum_};
+
+    uint64_t num_bytes_written = 0;
+
+    while (input_stream_.good()) {
+      input_stream_.read(reinterpret_cast<char *>(buffer.data()), buffer.size());
+
+      std::streamsize num_bytes_read = input_stream_.gcount();
+      logger_->log_trace("Read %d bytes of input", num_bytes_read);
```
szaszm, line 292: `std::streamsize` is an unspecified signed integral type used for sizes. I think we can assume that whatever value it holds can be represented by a `std::ptrdiff_t`, which has a defined format specifier.
```suggestion
        const auto num_bytes_read = input_stream_.gcount();
        logger_->log_trace("Read %td bytes of input", std::ptrdiff_t{num_bytes_read});
```

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -335,170 +563,210 @@ bool TailFile::storeState(const std::shared_ptr<core::ProcessContext>& context)
...
+  std::size_t last_dot_position = state.file_name_.find_last_of('.');
+  std::string base_name = state.file_name_.substr(0, last_dot_position);
+  std::string pattern = utils::StringUtils::replaceOne(rolling_filename_pattern_, "${filename}", base_name);
+
+  std::vector<TailStateWithMtime> matched_files_with_mtime;
+  auto collect_matching_files = [&](const std::string &path, const std::string &file_name) -> bool {
+    if (file_name != state.file_name_ && utils::Regex::matchesFullInput(pattern, file_name)) {
+      std::string full_file_name = path + utils::file::FileUtils::get_separator() + file_name;
+      TailStateWithMtime::TimePoint mtime{utils::file::FileUtils::last_write_time_point(full_file_name)};
+      logger_->log_debug("File %s with mtime %llu matches rolling filename pattern %s", file_name, mtime.time_since_epoch().count(), pattern);
```
szaszm, line 579: Tricky chrono logging again, see my other comment about this.

```suggestion
      logger_->log_debug("File %s with mtime %" PRIu64 " matches rolling filename pattern %s", file_name, uint64_t{mtime.time_since_epoch().count()}, pattern);
```

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -335,170 +563,210 @@ bool TailFile::storeState(const std::shared_ptr<core::ProcessContext>& context)
...
-        uint64_t candidateModTime = ((uint64_t) (sb.st_mtime) * 1000);
-        if (candidateModTime >= file.currentTailFileModificationTime_) {
-          logging::LOG_TRACE(logger_) << "File " << filename << " (short name " << file.current_file_name_ <<
-          ") disk mod time " << candidateModTime << ", struct mod time " << file.currentTailFileModificationTime_ << ", size on disk " << sb.st_size << ", position " << file.currentTailFilePosition_;
-          if (filename == file.current_file_name_ && candidateModTime == file.currentTailFileModificationTime_ &&
-              sb.st_size == file.currentTailFilePosition_) {
-            return true;  // Skip the current file as a candidate in case it wasn't updated
+std::vector<TailState> TailFile::findRotatedFiles(const TailState &state) const {
+  logger_->log_debug("Searching for files rolled over; last read time is %llu",
+      std::chrono::time_point_cast<std::chrono::seconds>(state.last_read_time_).time_since_epoch().count());
```
szaszm, line 568: Tricky chrono logging again, see my other comment about this.

```suggestion
  logger_->log_debug("Searching for files rolled over; last read time is %" PRIu64,
      uint64_t{std::chrono::time_point_cast<std::chrono::seconds>(state.last_read_time_).time_since_epoch().count()});
```

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -17,19 +17,14 @@
...
  * limitations under the License.
  */
-#include <sys/types.h>
 #include <sys/stat.h>
-#include <time.h>
-#include <stdio.h>
-
-#include <limits.h>
 #ifndef WIN32
 #include <dirent.h>
```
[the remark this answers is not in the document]
fgerlits: I have removed it.

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -40,60 +35,283 @@
...
+
+    return num_bytes_written;
+  }
+
+  uint64_t checksum() const {
+    return checksum_;
+  }
+
+  bool hasMoreToRead() const {
+    return begin != end || input_stream_.good();
```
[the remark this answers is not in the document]
fgerlits: @szaszm is right, but it's true the code is not as clear as it could be.  Would adding some comments help?  eg.
```c++
   return begin != end ||        // there is more to read from the current buffer
          input_stream_.good();  // ... or there is more to read on the disk
```

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -40,60 +35,283 @@
...
+
+    uint64_t num_bytes_written = 0;
+
+    while (input_stream_.good()) {
+      input_stream_.read(reinterpret_cast<char *>(buffer.data()), buffer.size());
+
+      std::streamsize num_bytes_read = input_stream_.gcount();
+      logger_->log_trace("Read %d bytes of input", num_bytes_read);
+
+      if (num_bytes_read < (std::numeric_limits<int>::min)() || num_bytes_read > (std::numeric_limits<int>::max)()) {
```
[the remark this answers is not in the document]
fgerlits: done

libminifi/test/unit/StringUtilsTests.cpp
```diff
@@ -338,3 +338,30 @@ TEST_CASE("TestStringUtils::testJoinPackNegative", "[test join_pack negative]")
...
+  REQUIRE(utils::StringUtils::replaceOne("banana", "an", "AN") == "bANana");
+  REQUIRE(utils::StringUtils::replaceOne("banana", "an", "***") == "b***ana");
+  REQUIRE(utils::StringUtils::replaceOne("banana", "banana", "kiwi") == "kiwi");
+  REQUIRE(utils::StringUtils::replaceOne("banana", "banana", "grapefruit") == "grapefruit");
+}
+
+TEST_CASE("StringUtils::replaceAll works correctly", "[replaceAll]") {
+  auto replaceAll = [](std::string input, const std::string &from, const std::string &to) -> std::string {
+    return utils::StringUtils::replaceAll(input, from, to);
+  };
```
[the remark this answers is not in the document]
fgerlits: The reason this lambda is there is that `replaceAll` takes its first argument as a non-const reference.  I could make this clearer by copying explicity, eg.
```c++
  auto replaceAll = [](const std::string &input, const std::string &from, const std::string &to) -> std::string {
    std::string input_copy{input};
    return utils::StringUtils::replaceAll(input_copy, from, to);
  };
```
Would you prefer that?

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -40,60 +35,283 @@
...
+ private:
+  char input_delimiter_;
+  uint64_t checksum_;
+  std::ifstream input_stream_;
+  std::shared_ptr<logging::Logger> logger_;
+
+  constexpr static std::size_t BUFFER_SIZE = 4096;
+  std::array<uint8_t, BUFFER_SIZE> buffer;
+  uint8_t *begin = buffer.data();
+  uint8_t *end = buffer.data();
```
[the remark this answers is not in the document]
fgerlits: done

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -204,52 +418,78 @@ void TailFile::parseStateFileLine(char *buf) {
...
     }
   }
   if (key == "POSITION") {
     // for backwards compatibility
-    if (tail_states_.size() != 1) {
+    if (tail_states_.size() != std::size_t{1}) {
       throw minifi::Exception(ExceptionType::PROCESSOR_EXCEPTION, "Incompatible state file types");
     }
     const auto position = std::stoull(value);
     logger_->log_debug("Received position %d", position);
```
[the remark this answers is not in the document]
fgerlits: done

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -260,72 +500,60 @@ bool TailFile::recoverState(const std::shared_ptr<core::ProcessContext>& context
...
         }
       } catch (...) {
         continue;
       }
     }
-    state_load_success = true;
-    tail_states_ = std::move(new_tail_states);
     for (const auto& s : tail_states_) {
-      logger_->log_debug("TailState %s: %s, %s, %llu, %llu", s.first, s.second.path_, s.second.current_file_name_, s.second.currentTailFilePosition_, s.second.currentTailFileModificationTime_);
+      logger_->log_debug("TailState %s: %s, %s, %llu, %llu",
```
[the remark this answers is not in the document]
fgerlits: done

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -335,170 +563,210 @@ bool TailFile::storeState(const std::shared_ptr<core::ProcessContext>& context)
...
-  return true;};
-
-    utils::file::FileUtils::list_dir(file.path_, lambda, logger_, false);
 
-    if (matchedFiles.size() < 1) {
-      logger_->log_debug("No newer files found in directory!");
-      return;
-    }
+  std::vector<TailState> matched_files;
+  std::transform(matched_files_with_mtime.begin(), matched_files_with_mtime.end(), std::back_inserter(matched_files),
```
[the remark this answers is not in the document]
fgerlits: done

extensions/standard-processors/processors/TailFile.cpp
fgerlits: done

extensions/standard-processors/processors/TailFile.cpp
fgerlits: done

extensions/standard-processors/processors/TailFile.cpp
fgerlits: done

extensions/standard-processors/processors/TailFile.cpp
fgerlits: The underlying types of `seconds` and `milliseconds` are "signed integer type of at least 35 bits" and "signed integer type of at least 45 bits", respectively, so I changed this (and similar logs) to `int64_t{some_duration.count()}` and `PRId64`.

extensions/standard-processors/processors/TailFile.cpp
fgerlits: I have kept the `if (found_delimiter)` instead of a ternary because I think it's more readable, but I have replaced the range checking with `gsl::narrow`.

extensions/standard-processors/processors/TailFile.cpp
fgerlits: I have changed the type of the buffer from `uint8_t` to `char` so I only need to convert once, when I'm writing to the `CRCStream`.

extensions/standard-processors/processors/TailFile.cpp
fgerlits: done

extensions/standard-processors/processors/TailFile.cpp
fgerlits: For some reason, Visual Studio complained that the conversion from `std::streamsize` to `std::ptrdiff_t` was narrowing, so I ended up using `std::intmax_t` and `%jd`.

extensions/standard-processors/processors/TailFile.cpp
fgerlits: see above

extensions/standard-processors/processors/TailFile.cpp
fgerlits: done

extensions/standard-processors/processors/TailFile.cpp
fgerlits: done

extensions/standard-processors/tests/unit/TailFileTests.cpp
```diff
@@ -652,46 +872,606 @@ TEST_CASE("TailFileWithMultileRolledOverFiles", "[tailfiletest2]") {
...
+
+  std::ofstream test_file_stream_1(test_file, std::ios::binary);
+  test_file_stream_1 << "Pineapple" << DELIM << "Kiwi" << DELIM;
+  test_file_stream_1.close();
+
+  std::string second_rotated_file = dir + utils::file::FileUtils::get_separator() + "fruits.1.log";
+  REQUIRE(rename(test_file.c_str(), second_rotated_file.c_str()) == 0);
+
+  std::ofstream test_file_stream_2(test_file, std::ios::binary);
+  test_file_stream_2 << "Apricont" << DELIM;
```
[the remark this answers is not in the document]
fgerlits: An apricont is a type of apricot often depicted in Russian religious paintings.  Fixed.

libminifi/test/unit/StringUtilsTests.cpp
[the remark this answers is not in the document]
szaszm: I didn't realize that `StringUtils::replaceAll` modifies the source string. My bad.

extensions/standard-processors/processors/TailFile.cpp
szaszm: You're right, thanks.

extensions/standard-processors/processors/TailFile.cpp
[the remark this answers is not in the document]
arpadboda: Sounds good to me, thanks!

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -204,52 +418,78 @@ void TailFile::parseStateFileLine(char *buf) {
...
-  bool state_load_success = false;
+  if (tail_mode_ == Mode::SINGLE) {
+    if (tail_states_.size() == 1) {
+      auto state_it = tail_states_.begin();
+      const auto it = new_tail_states.find(state_it->first);
+      if (it != new_tail_states.end()) {
+        state_it->second = it->second;
+      }
+    } else {
+      throw minifi::Exception(ExceptionType::PROCESSOR_EXCEPTION, "This should never happen: "
```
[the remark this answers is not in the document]
fgerlits: I don't think it can happen, as a configuration change will restart the processor.  A newly created processor in Single file mode starts with a one-element state, and we only update that one item here, we don't add to it.  This `throw` is here only to guard against later code changes which change this behavior.

As to invalidating the state, I'll need to look at how NiFi does that.  I think there is value in keeping the state of existing files after a configuration change: for example, if the selection is broadened in Multi file mode (say from `*.log` to `*.*`), we probably don't want to restart the files we have already tailed from the beginning.  But we should do whatever NiFi does.  Can that be a separate pull request?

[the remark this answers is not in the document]
arpadboda: Yep, it's fine to do as a follow-up.

fgerlits:
> My tests showed incorrect handling of rotation while the minifi c++ agent is stopped. Instead of finding the file that we were last working with, under a potentially different name, I observed an iteration over all rotated files and consuming everything again. Could you please check that?

https://github.com/apache/nifi-minifi-cpp/pull/791/commits/cca8f3775667e5f824056d681bed3963ad43bcb2 fixes this bug according to my tests -- please redo your tests, too

The rollover detection logic is still primitive: it only checks for rollover if the current file length is less than the last read position.  So if
* minifi is stopped after reading 100 bytes from test.log
* 900 more bytes are written to test.log, and it is renamed to test.log.1
* 300 bytes are written to a new test.log file
* and minifi is restarted, then it will happily continue from byte 101 in the new file, and skip the 900 + 100 bytes written in between.

I am planning to fix that in a separate pull request, if that is OK.

PR #805, opened 2020-06-09
MINIFICPP-1252 Add gsl-lite

szaszm:

### For all changes:
     in the commit message?

### For code changes:

### Note:


szaszm:
Since header-only libraries are [recommended to be included in the source tree](https://github.com/apache/nifi-minifi-cpp/blob/master/ThirdParties.md#interface-libraries), I separated the gsl-lite inclusion from the rest of the changes for easier review.

libminifi/include/utils/ScopeGuard.h
```diff
@@ -22,39 +22,19 @@
...
-      } catch (...) {
-      }
-    }
-  }
-
-  ScopeGuard(const ScopeGuard&) = delete;
-  ScopeGuard(ScopeGuard&&) = delete;
-  ScopeGuard& operator=(const ScopeGuard&) = delete;
-  ScopeGuard& operator=(ScopeGuard&&) = delete;
+struct ScopeGuard : ::gsl::final_action<std::function<void()>> {
```
hunyadi-dev, line 33: 👍

libminifi/src/FlowControlProtocol.cpp
```diff
@@ -162,11 +164,12 @@ int FlowControlProtocol::sendRegisterReq() {
     return -1;
 
   // Calculate the total payload msg size
-  uint32_t payloadSize = FlowControlMsgIDEncodingLen(FLOW_SERIAL_NUMBER, 0) + FlowControlMsgIDEncodingLen(FLOW_YML_NAME, this->_controller->getName().size() + 1);
-  uint32_t size = sizeof(FlowControlProtocolHeader) + payloadSize;
+  const auto payloadSize = FlowControlMsgIDEncodingLen(FLOW_SERIAL_NUMBER, 0) + FlowControlMsgIDEncodingLen(FLOW_YML_NAME, gsl::narrow<int>(this->_controller->getName().size() + 1));
```
hunyadi-dev, line 167: I don't see any try catch block around this.

Maybe we should catch and log narrowing exceptions in `FlowControlProtocol::run()`, so that if the thread catches due to them we see the reson.

szaszm: Normally `gsl::narrow<T>` calls `std::terminate()` on failed conversion. I set the config options to change the behavior to throwing an exception in release builds, but we should normally not rely on this.

The motivation here is to fail early at runtime instead of invoking undefined behavior. It should never fail, and if it does, it's a programming error. This should be a hard error, not an exception.

libminifi/include/FlowController.h
```diff
@@ -50,6 +51,7 @@
 #include "utils/Id.h"
 #include "core/state/UpdateController.h"
 #include "core/state/nodes/FlowInformation.h"
+#include "utils/gsl.h"
```
szaszm, line 54: Redundant include

extensions/librdkafka/PublishKafka.cpp
```diff
@@ -31,6 +31,7 @@
 #include "utils/StringUtils.h"
 #include "utils/ScopeGuard.h"
 #include "utils/GeneralUtils.h"
+#include "utils/gsl.h"
```
szaszm, line 34: Redundant include

szaszm: 2259722

libminifi/include/FlowController.h
szaszm: 2259722

szaszm:
Only README.md, LICENSE and include/ are included now from gsl-lite

PR #812, opened 2020-06-12
MINIFICPP-1247 - Enhance logging for CWEL

nghiaxlee:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/windows-event-log/ConsumeWindowsEventLog.cpp
+++ b/extensions/windows-event-log/ConsumeWindowsEventLog.cpp
@@ -174,7 +174,8 @@ ConsumeWindowsEventLog::ConsumeWindowsEventLog(const std::string& name, utils::I
 }
 
 void ConsumeWindowsEventLog::notifyStop() {
-  logger_->log_trace("start notifyStop"); 
+  std::lock_guard<std::mutex> lock(onTriggerMutex_);
+  logger_->log_trace("start notifyStop");
   pBookmark_.reset();
   if (hMsobjsDll_) {
     if (FreeLibrary(hMsobjsDll_)) {
--- a/extensions/windows-event-log/ConsumeWindowsEventLog.cpp
+++ b/extensions/windows-event-log/ConsumeWindowsEventLog.cpp
@@ -178,6 +178,7 @@ void ConsumeWindowsEventLog::notifyStop() {
   pBookmark_.reset();
   if (hMsobjsDll_) {
     FreeLibrary(hMsobjsDll_);
+    hMsobjsDll_ = nullptr;
   }
   logger_->log_trace("finish notifyStop"); 
 }
@@ -345,6 +346,7 @@ void ConsumeWindowsEventLog::onTrigger(const std::shared_ptr<core::ProcessContex
   const auto hEventResults = EvtQuery(0, wstrChannel_.c_str(), wstrQuery_.c_str(), EvtQueryChannelPath);
   if (!hEventResults) {
     LOG_LAST_ERROR(EvtQuery);
+    context->yield();
     return;
   }
   const utils::ScopeGuard guard_hEventResults([hEventResults]() { EvtClose(hEventResults); });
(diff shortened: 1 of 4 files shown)


extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -173,6 +173,15 @@ ConsumeWindowsEventLog::ConsumeWindowsEventLog(const std::string& name, utils::I
   writePlainText_ = false;
 }
 
+void ConsumeWindowsEventLog::notifyStop() {
+  logger_->log_trace("start notifyStop"); 
+  pBookmark_.reset();
+  if (hMsobjsDll_) {
+    FreeLibrary(hMsobjsDll_);
```
szaszm, line 180: Unloading the DLL on every reload is quite dangerous IMO. Why do we do it?

In any case, to avoid double free, we should set `hMsobjsDll_` to `nullptr` after freeing it. Otherwise the dtor would call `FreeLibrary` again, which might be problematic.

nghiaxlee: I have not read the Load/FreeLibrary functions. I just realized that this processor haven't had a notifyStop yet, so I go ahead and create one.

As a lesson from PublishKafka, things allocated in onSchedule should be released in notifyStop, because every time the agent receives a new configuration, it will create new processor objects instead of reusing those previous ones, and those previous instances will just stay inactive but not destruct (due to shared ptr leak?), @arpadboda please correct me if I misunderstand.

@szaszm if you already knew the above issue or if you think that I misunderstand it, please let me know (and if it is feasible, a brief summary about Load/FreeLibrary would be great, but I will surely take a look at the docs as well).

Since I'm unfamiliar with Windows API, I am not sure if the current implementation is intended, I will check the docs. In any case, your second point makes sense, thanks for the suggestion!

szaszm: `LoadLibrary`/`FreeLibrary` are used to dynamically load and unload DLLs, similarly to `dlopen`/`dlclose` on POSIX.

I didn't mean to question the validity of implementing `notifyStop`, thank you for doing that. I didn't know the exact lifetime of processors and assumed reuse of the objects. Not sure about `dlopen`, but in case of `LoadLibrary`, repeated opens of the same DLL increment a reference count instead of loading the library multiple times, so it's good news that we can only leak it once. :slightly_smiling_face: 

The reason I thought `FreeLibrary` on every reload is dangerous was that I had to debug an issue in the past (on a different project) where a callback registered from a DLL was called after the unloading of the DLL which caused the DLL's instance of common types (e.g. `std::string`) to become inaccessible and the program to crash in a hard-to-debug way when trying to call that code.

After looking further into our usage of this particular DLL, I noticed that we only use it in `substituteXMLPercentageItems`, and only to pass the handle to a WINAPI call (`FormatMessage` in particular), so unloading the DLL while the processor is not running shouldn't be a problem here.

This means that the first point is a non-issue. However, as you also recognized, the second point still stands, i.e. we should set the pointer to `nullptr` after releasing the lib.

arpadboda: I think we should grab the mutex here. 

A long-running onTrigger execution can still be running when notifyStop is called from a different thread.

So we might unload the library while it's in use. 

To avoid this, let's lock onTriggerMutex_

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -381,23 +405,27 @@ void ConsumeWindowsEventLog::onTrigger(const std::shared_ptr<core::ProcessContex
     }
   }
 
+  logger_->log_trace("Finish enumerating events.");
```
szaszm, line 408: Shouldn't we also yield here, when there are no more events left?

arpadboda: This can mean that the processor gathered a bunch of events (and those events has been generated since the last ontrigger call), so I don't think waiting  for a longer time helps in this case.
Yield might be  considered in case the ontrigger call generated 0 new events.

PR #816, opened 2020-06-16
MINIFICPP-1261 - Refactor non-trivial usages of ScopeGuard class

hunyadi-dev:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/rocksdb-repos/FlowFileRepository.cpp
+++ b/extensions/rocksdb-repos/FlowFileRepository.cpp
@@ -122,27 +122,26 @@ void FlowFileRepository::run() {
 }
 
 void FlowFileRepository::prune_stored_flowfiles() {
-  rocksdb::DB* stored_database_ = nullptr;
-  utils::ScopeGuard db_guard([&stored_database_]() {
-    delete stored_database_;
-  });
+  std::unique_ptr<rocksdb::DB> stored_database;
+  rocksdb::DB* used_database;
   bool corrupt_checkpoint = false;
   if (nullptr != checkpoint_) {
     rocksdb::Options options;
     options.create_if_missing = true;
     options.use_direct_io_for_flush_and_compaction = true;
     options.use_direct_reads = true;
-    rocksdb::Status status = rocksdb::DB::OpenForReadOnly(options, FLOWFILE_CHECKPOINT_DIRECTORY, &stored_database_);
+    rocksdb::Status status = rocksdb::DB::OpenForReadOnly(options, FLOWFILE_CHECKPOINT_DIRECTORY, &used_database);
+    stored_database.reset(used_database);
     if (!status.ok()) {
-      stored_database_ = db_;
-      db_guard.disable();
+      used_database = db_;
+      stored_database.release();
     }
   } else {
     logger_->log_trace("Could not open checkpoint as object doesn't exist. Likely not needed or file system error.");
     return;
   }
 
-  rocksdb::Iterator* it = stored_database_->NewIterator(rocksdb::ReadOptions());
+  rocksdb::Iterator* it = used_database->NewIterator(rocksdb::ReadOptions());
   for (it->SeekToFirst(); it->Valid(); it->Next()) {
     std::shared_ptr<FlowFileRecord> eventRead = std::make_shared<FlowFileRecord>(shared_from_this(), content_repo_);
     std::string key = it->key().ToString();
--- a/extensions/rocksdb-repos/FlowFileRepository.cpp
+++ b/extensions/rocksdb-repos/FlowFileRepository.cpp
@@ -122,8 +122,8 @@ void FlowFileRepository::run() {
 }
 
 void FlowFileRepository::prune_stored_flowfiles() {
-  std::unique_ptr<rocksdb::DB> stored_database;
   rocksdb::DB* used_database;
+  std::unique_ptr<rocksdb::DB> stored_database;
   bool corrupt_checkpoint = false;
   if (nullptr != checkpoint_) {
     rocksdb::Options options;
@@ -131,10 +131,10 @@ void FlowFileRepository::prune_stored_flowfiles() {
     options.use_direct_io_for_flush_and_compaction = true;
     options.use_direct_reads = true;
     rocksdb::Status status = rocksdb::DB::OpenForReadOnly(options, FLOWFILE_CHECKPOINT_DIRECTORY, &used_database);
-    stored_database.reset(used_database);
-    if (!status.ok()) {
+    if (status.ok()) {
+      stored_database.reset(used_database);
+    } else {
       used_database = db_;
-      stored_database.release();
     }
   } else {
     logger_->log_trace("Could not open checkpoint as object doesn't exist. Likely not needed or file system error.");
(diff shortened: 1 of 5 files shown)


szaszm:
The CI failures look legit and relevant

extensions/rocksdb-repos/FlowFileRepository.cpp
```diff
@@ -122,27 +122,26 @@ void FlowFileRepository::run() {
...
+  rocksdb::DB* used_database;
   bool corrupt_checkpoint = false;
   if (nullptr != checkpoint_) {
     rocksdb::Options options;
     options.create_if_missing = true;
     options.use_direct_io_for_flush_and_compaction = true;
     options.use_direct_reads = true;
-    rocksdb::Status status = rocksdb::DB::OpenForReadOnly(options, FLOWFILE_CHECKPOINT_DIRECTORY, &stored_database_);
+    rocksdb::Status status = rocksdb::DB::OpenForReadOnly(options, FLOWFILE_CHECKPOINT_DIRECTORY, &used_database);
+    stored_database.reset(used_database);
```
arpadboda, line 134: This makes few sense to me to do here in case we release 4 lines later. 
I would suggest doing it only when status is ok. Might be a bit more lines, but increases readability.

hunyadi-dev: Changed as requested.

libminifi/include/utils/ScopeGuard.h
```diff
@@ -33,10 +33,6 @@ namespace utils {
 
 struct ScopeGuard : ::gsl::final_action<std::function<void()>> {
   using ::gsl::final_action<std::function<void()>>::final_action;
-
-  void disable() noexcept {
-    dismiss();
-  }
```
szaszm, line 39: I wouldn't remove this just for the sake of removing it, because external code might depend on it, and it's not a huge maintenance burden to keep it.

libminifi/include/io/tls/TLSSocket.h
```diff
@@ -80,10 +77,13 @@ class TLSContext : public SocketContext {
...
 
  private:
+  static void deleteContext(SSL_CTX* ptr) { SSL_CTX_free(ptr); }
+
   std::shared_ptr<logging::Logger> logger_;
   std::shared_ptr<Configure> configure_;
   std::shared_ptr<minifi::controllers::SSLContextService> ssl_service_;
-  SSL_CTX *ctx;
+  using Context = std::unique_ptr<SSL_CTX, decltype(&deleteContext)>;
+  Context ctx;
```
szaszm, line 86: I would localize the context type alias to the implementation, so that pointer semantics are obvious when looking at the declaration. This is probably very subjective because no guidelines come to my mind to support this.
```suggestion
  std::unique_ptr<SSL_CTX, decltype(&deleteContext)> ctx;
```

libminifi/src/io/tls/TLSSocket.cpp
```diff
@@ -72,30 +72,25 @@ int16_t TLSContext::initialize(bool server_method) {
...
     std::string caCertificate;
 
     if (ssl_service_ != nullptr) {
-      if (!ssl_service_->configure_ssl_context(ctx)) {
+      if (!ssl_service_->configure_ssl_context(local_context.get())) {
         error_value = TLS_ERROR_CERT_ERROR;
         return error_value;
       }
-      ctxGuard.disable();
+      ctx.swap(local_context);
```
szaszm, line 93: I think a move assignment would be more clear and would expose potential lifetime issues earlier. (Unless it's implemented in terms of `swap`.)
```suggestion
      ctx = std::move(local_context);
```

libminifi/src/io/tls/TLSSocket.cpp
```diff
@@ -120,35 +115,35 @@ int16_t TLSContext::initialize(bool server_method) {
...
         error_value = TLS_ERROR_CERT_ERROR;
         return error_value;
       }
     }
 
-    logger_->log_debug("Load/Verify Client Certificate OK. for %X and %X", this, ctx);
+    logger_->log_debug("Load/Verify Client Certificate OK. for %X and %X", this, local_context.get());
   }
-  ctxGuard.disable();
+  ctx.swap(local_context);
```
szaszm, line 146: Same as above, move assignment feels like it would be a more accurate description of intent.
```suggestion
  ctx = std::move(local_context);
```

libminifi/src/controllers/SSLContextService.cpp
```diff
@@ -56,30 +56,30 @@ void SSLContextService::initialize() {
...
       EVP_PKEY* pkey = nullptr;
       X509* cert = nullptr;
       STACK_OF(X509)* ca = nullptr;
-      if (!PKCS12_parse(p12, passphrase_.c_str(), &pkey, &cert, &ca)) {
+      if (!PKCS12_parse(p12.get(), passphrase_.c_str(), &pkey, &cert, &ca)) {
         logging::LOG_ERROR(logger_) << "Failed to parse certificate file " << certificate << " as PKCS#12, " << getLatestOpenSSLErrorString();
         return false;
       }
-      utils::ScopeGuard certs_guard([pkey, cert, ca]() {
+      ::gsl::final_action<std::function<void()>> certs_guard([pkey, cert, ca]() {
```
szaszm, line 82: Please use `auto` and `gsl::finally`. It's not recommended (by gsl-lite) to depend on the global `gsl` namespace, as they want to eventually move to the `gsl_lite` namespace. `ScopeGuard` has this dependency to greatly simplify the implementation, but here it doesn't provide a benefit.
```suggestion
      const auto certs_guard = gsl::finally([pkey, cert, ca]() {
```

libminifi/src/io/tls/TLSSocket.cpp
hunyadi-dev: Adding manually, as I do not want to split the change into two commits. The other suggestion has been applied.

libminifi/include/io/tls/TLSSocket.h
hunyadi-dev: Adding manually, as I do not want to split the change into two commits. The other suggestion has been applied.

libminifi/include/utils/ScopeGuard.h
hunyadi-dev: I do not like this pattern being present in the codebase at all. Restored, but marked for depracation ASAP (1.0).

libminifi/src/io/tls/TLSSocket.cpp
hunyadi-dev: Adding manually, as I do not want to split the change into two commits. The other suggestion has been applied.

libminifi/include/io/tls/TLSSocket.h
szaszm: Not a big deal, just my opinion/preference. I prefer to use convenience aliases only if they bring a large benefit. This is because in the past I was working with a codebase that really overused type aliases and I had no idea what's what while reading the code.

libminifi/include/utils/ScopeGuard.h
szaszm: Ok. I didn't think this was harmful, but if you think so, then deprecation sounds like a good idea.

PR #817, opened 2020-06-16
MINIFICPP-1262 Fix the MinifiConcurrentQueueTests unit tests

fgerlits:
### Description of the changes

* Move the timed and untimed waiting consumer tests into separate test cases, as the timed consumers do not need a sleep in the test (they wait and retry internally) and they do not need to stop the queue, but the untimed consumers do.
* In the untimed consumer case, change the fixed 10 ms wait to a wait until the queue is empty (but not more than 1 second).
* Change possibly infinite loops in two other tests to also wait for the condition with a 1 second timeout.
* Move some other independent tests, which used to be sections within a single test, into separate test cases.
* Minor cleanup

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/test/unit/MinifiConcurrentQueueTests.cpp
+++ b/libminifi/test/unit/MinifiConcurrentQueueTests.cpp
@@ -331,7 +331,7 @@ TEST_CASE("TestConcurrentQueue: test the readding dequeue consumer", "[ProducerC
   utils::ConditionConcurrentQueue<std::string> queue(true);
   std::vector<std::string> results;
 
-  std::atomic_int results_size;
+  std::atomic_int results_size{0};
   std::thread consumer { getReaddingDequeueConsumerThread(queue, results, results_size) };
   std::this_thread::sleep_for(std::chrono::milliseconds(1));
   std::thread producer { getSimpleProducerThread(queue) };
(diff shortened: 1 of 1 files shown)


libminifi/test/unit/MinifiConcurrentQueueTests.cpp
```diff
@@ -31,6 +31,20 @@ namespace utils = org::apache::nifi::minifi::utils;
 
 namespace {
 
+  template <typename Function, typename Duration>
```
arpadboda, line 34: I think relying on chrono::duration is a bit more bulletproof:
```
template<typename Function, class Rep, class Period>
bool becomesTrueWithinTimeout(const Function &conditio, const std::chrono::duration<Rep, Period>& timeout) {
```

fgerlits: I was mainly trying to shave off a few characters, but yeah, it's probably not worth it.  I have changed the argument type to `std::chrono::duration<Rep, Period>` but kept it as pass-by-value, since it is just a number.

arpadboda: Okay, thanks!

PR #818, opened 2020-06-16
MINIFICPP-1263 - Fix for segfault

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/http-curl/tests/AbsoluteTimeoutTest.cpp
+++ b/extensions/http-curl/tests/AbsoluteTimeoutTest.cpp
@@ -35,7 +35,7 @@ int main() {
     std::chrono::milliseconds(500)
   });
 
-  auto server = start_webserver(port, rootURI, &handler);
+  CivetServer* server = start_webserver(port, rootURI, &handler);
 
   auto plan = controller.createPlan();
 
@@ -47,4 +47,6 @@ int main() {
   plan->runNextProcessor();
 
   assert(LogTestController::getInstance().contains("HTTP operation timed out, with absolute timeout 3000ms"));
+
+  delete server;
 }
--- a/extensions/http-curl/tests/AbsoluteTimeoutTest.cpp
+++ b/extensions/http-curl/tests/AbsoluteTimeoutTest.cpp
@@ -35,7 +35,7 @@ int main() {
     std::chrono::milliseconds(500)
   });
 
-  CivetServer* server = start_webserver(port, rootURI, &handler);
+  TestServer server(port, rootURI, &handler);
 
   auto plan = controller.createPlan();
 
@@ -47,6 +47,4 @@ int main() {
   plan->runNextProcessor();
 
   assert(LogTestController::getInstance().contains("HTTP operation timed out, with absolute timeout 3000ms"));
-
-  delete server;
 }
--- a/extensions/http-curl/tests/ServerAwareHandler.h
+++ b/extensions/http-curl/tests/ServerAwareHandler.h
@@ -21,16 +21,25 @@
 
 class ServerAwareHandler: public CivetHandler{
 protected:
-  const std::atomic_bool *is_server_running{nullptr};
-  bool isServerRunning(){
-    assert(is_server_running);
-    return *is_server_running;
+  void sleep_for(std::chrono::milliseconds time) {
+    std::unique_lock<std::mutex> lock(mutex_);
+    stop_signal_.wait_for(lock, time, [&] {return terminate_.load();});
   }
+
+  bool isServerRunning() {
+    return !terminate_.load();
+  }
+
 public:
-  void initServerFlag(std::atomic_bool& is_running){
-    assert(is_server_running == nullptr);
-    is_server_running = &is_running;
+  void stop(){
+    terminate_ = true;
+    stop_signal_.notify_all();
   }
+
+ private:
+  std::mutex mutex_;
+  std::condition_variable stop_signal_;
+  std::atomic_bool terminate_{false};
 };
 
 #endif //NIFI_MINIFI_CPP_SERVERAWAREHANDLER_H
--- a/extensions/http-curl/tests/ServerAwareHandler.h
+++ b/extensions/http-curl/tests/ServerAwareHandler.h
@@ -26,7 +26,7 @@ protected:
     stop_signal_.wait_for(lock, time, [&] {return terminate_.load();});
   }
 
-  bool isServerRunning() {
+  bool isServerRunning() const {
     return !terminate_.load();
   }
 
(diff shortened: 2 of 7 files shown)


extensions/http-curl/tests/AbsoluteTimeoutTest.cpp
```diff
@@ -47,4 +47,6 @@ int main() {
   plan->runNextProcessor();
 
   assert(LogTestController::getInstance().contains("HTTP operation timed out, with absolute timeout 3000ms"));
+
+  delete server;
```
arpadboda, line 51: Could you explain why does leaking it cause segfault?

Btw I think unique_ptr would be a nicer solution.

adamdebreceni: it was just a temporary check, in no way did I intend it to be the solution, but the cause is: if the handler returns after a server shutdown was initiated it will try to consume the next socket and since that signal will never be notified by the server (as it is waiting for the handlers to shutdown 🤦) we have a deadlock

adamdebreceni: the relevant part in `civetweb.c:16178`:

```
	/* Call consume_socket() even when ctx->stop_flag > 0, to let it
	 * signal sq_empty condvar to wake up the master waiting in
	 * produce_socket() */
	while (consume_socket(ctx, &conn->client, conn->thread_index)) {
		conn->conn_birth_time = time(NULL);
```

adamdebreceni: but the master is not waiting in the `produce_socket`

adamdebreceni:
```
static int
consume_socket(struct mg_context *ctx, struct socket *sp, int thread_index)
{
	DEBUG_TRACE("%s", "going idle");
	ctx->client_socks[thread_index].in_use = 0;
	event_wait(ctx->client_wait_events[thread_index]);
	*sp = ctx->client_socks[thread_index];
	DEBUG_TRACE("grabbed socket %d, going busy", sp ? sp->sock : -1);

	return !ctx->stop_flag;
}
```

adamdebreceni: the segfault was (probably) caused by the `CivetServer` creating a load of threads and we just exited `main` like there is no tomorrow

extensions/http-curl/tests/TestServer.h
```diff
@@ -24,65 +24,87 @@
...
+ */
+class TestServer{
+  struct CivetLibrary{
+    CivetLibrary() {
+      if (getCounter()++ == 0) {
+        mg_init_library(0);
+      }
+    }
+    ~CivetLibrary() {
+      if (--getCounter() == 0) {
```
arpadboda, line 42: What's the motivation behind using a getter on our own private member?

extensions/http-curl/tests/TestServer.h
```diff
@@ -24,65 +24,87 @@
...
-    cpp_options.push_back(options[i]);
+    addHandler(rooturi, handler);
   }
-  CivetServer *server = new CivetServer(cpp_options);
-
-  server->addHandler(rooturi, handler);
 
-  return server;
+  void addHandler(const std::string& uri, CivetHandler* handler) {
+    handlers_.emplace_back(handler);
```
arpadboda, line 89: It's only fancy, but I prefer not to use emplace when inserting already existing objects as emplace calls explicit ctors as well and it's very easy to point the gun against ourselves.

extensions/http-curl/tests/TestServer.h
```diff
@@ -24,65 +24,87 @@
...
+    }
 
-  /* Un-initialize the library */
-  mg_exit_library();
-}
+  }
+ private:
+  CivetLibrary lib_;
+  std::unique_ptr<CivetServer> server_;
+  std::vector<CivetHandler*> handlers_;
```
arpadboda, line 107: I think a comment here would be nice to state that the destruction (stopping) of handlers depend on the server, so the order of members is really important here.

extensions/http-curl/tests/ServerAwareHandler.h
```diff
@@ -21,16 +21,25 @@
...
-  const std::atomic_bool *is_server_running{nullptr};
-  bool isServerRunning(){
-    assert(is_server_running);
-    return *is_server_running;
+  void sleep_for(std::chrono::milliseconds time) {
+    std::unique_lock<std::mutex> lock(mutex_);
+    stop_signal_.wait_for(lock, time, [&] {return terminate_.load();});
   }
+
+  bool isServerRunning() {
```
arpadboda, line 29: fancy: const

extensions/http-curl/tests/TestServer.h
```diff
@@ -24,65 +24,87 @@
...
-  server->addHandler(rooturi, handler);
+    server_ = utils::make_unique<CivetServer>(cpp_options, (CivetCallbacks*)callbacks);
 
-  return server;
+    addHandler(rooturi, handler);
+  }
 
-}
+  TestServer(std::string &port, std::string &rooturi, CivetHandler *handler) {
+    const char *options[] = {"document_root", ".", "listening_ports", port.c_str(), 0};
```
arpadboda, line 77: I know this is old code, but as far as I see we could simply pass and std::vector<std::string> to Civetweb's ctor, so this could be something like:
```
std::vector<std::string> options = {"document_root", ".", "listening_ports", port};
server_ = utils::make_unique<CivetServer>(options);
```

The same applies to the above.

extensions/http-curl/tests/TestServer.h
adamdebreceni: will make the `getCounter` static, we do not access a member though this getter, but a static library initialization counter

extensions/http-curl/tests/TestServer.h
adamdebreceni: will add a comment as there is a dependence, between the `lib_` and the `server_` but not between the `handlers_` and others, those are stopped in the `~TestServer` before members are destructed

extensions/http-curl/tests/TestServer.h
adamdebreceni: good advice

extensions/http-curl/tests/TestServer.h
adamdebreceni: done

extensions/http-curl/tests/TestServer.h
adamdebreceni: done

extensions/http-curl/tests/TestServer.h
adamdebreceni: done

extensions/http-curl/tests/ServerAwareHandler.h
adamdebreceni: done

PR #821, opened 2020-06-24
MINIFICPP-1251 - Implement and test RetryFlowFile processor

hunyadi-dev:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/standard-processors/processors/RetryFlowFile.cpp
+++ b/extensions/standard-processors/processors/RetryFlowFile.cpp
@@ -117,12 +117,12 @@ void RetryFlowFile::onTrigger(core::ProcessContext* context, core::ProcessSessio
         session->transfer(flow_file, Failure);
         return;
       }
+      // Assuming reuse_mode_ == WARN_ON_REUSE || reuse_mode_ == RESET_REUSE
+      core::logging::LOG_LEVEL reuse_mode_log_level(core::logging::LOG_LEVEL::debug);
       if (reuse_mode_ == WARN_ON_REUSE) {
-        reuse_mode_log_level_ = core::logging::LOG_LEVEL::warn;
-      } else {  // Assuming reuse_mode_ == RESET_REUSE
-        reuse_mode_log_level_ = core::logging::LOG_LEVEL::debug;
+        reuse_mode_log_level = core::logging::LOG_LEVEL::warn;
       }
-      logger_->log_string(reuse_mode_log_level_, "Reusing retry attribute that belongs to different processor. Resetting value to 0.");
+      logger_->log_string(reuse_mode_log_level, "Reusing retry attribute that belongs to different processor. Resetting value to 0.");
       retry_property_value = 0;
     }
   }
--- a/extensions/standard-processors/processors/RetryFlowFile.cpp
+++ b/extensions/standard-processors/processors/RetryFlowFile.cpp
@@ -99,13 +99,12 @@ void RetryFlowFile::onTrigger(core::ProcessContext* context, core::ProcessSessio
     return;
   }
 
-  bool failure_due_to_non_numerical_retry;
   utils::optional<uint64_t> maybe_retry_property_value = getRetryPropertyValue(flow_file);
   if (!maybe_retry_property_value) {
     session->transfer(flow_file, Failure);
     return;
   }
-  uint64_t retry_property_value = maybe_retry_property_value.value_or(0);
+  uint64_t retry_property_value = maybe_retry_property_value.value();
   const std::string last_retried_by_property_name = retry_attribute_ + ".uuid";
   const std::string current_processor_uuid = getUUIDStr();
   std::string last_retried_by_uuid;
@@ -119,10 +118,11 @@ void RetryFlowFile::onTrigger(core::ProcessContext* context, core::ProcessSessio
         return;
       }
       if (reuse_mode_ == WARN_ON_REUSE) {
-        logger_->log_warn("Reusing retry attribute that belongs to different processor. Resetting value to 0.");
+        reuse_mode_log_level_ = core::logging::LOG_LEVEL::warn;
       } else {  // Assuming reuse_mode_ == RESET_REUSE
-        logger_->log_debug("Reusing retry attribute that belongs to different processor. Resetting value to 0.");
+        reuse_mode_log_level_ = core::logging::LOG_LEVEL::debug;
       }
+      logger_->log_string(reuse_mode_log_level_, "Reusing retry attribute that belongs to different processor. Resetting value to 0.");
       retry_property_value = 0;
     }
   }
--- a/extensions/standard-processors/processors/RetryFlowFile.cpp
+++ b/extensions/standard-processors/processors/RetryFlowFile.cpp
@@ -100,12 +100,12 @@ void RetryFlowFile::onTrigger(core::ProcessContext* context, core::ProcessSessio
   }
 
   bool failure_due_to_non_numerical_retry;
-  uint64_t retry_property_value;
-  std::tie(retry_property_value, failure_due_to_non_numerical_retry) = getRetryPropertyValue(flow_file);
-  if (failure_due_to_non_numerical_retry) {
+  utils::optional<uint64_t> maybe_retry_property_value = getRetryPropertyValue(flow_file);
+  if (!maybe_retry_property_value) {
     session->transfer(flow_file, Failure);
     return;
   }
+  uint64_t retry_property_value = maybe_retry_property_value.value_or(0);
   if (updateUUIDMarkerAndCheckFailOnReuse(flow_file)) {
     session->transfer(flow_file, Failure);
     return;
@@ -135,38 +135,37 @@ void RetryFlowFile::onTrigger(core::ProcessContext* context, core::ProcessSessio
 }
 
 void RetryFlowFile::readDynamicPropertyKeys(core::ProcessContext* context) {
-  exceeded_flowfile_attribute_keys.clear();
+  exceeded_flowfile_attribute_keys_.clear();
   const std::vector<std::string> dynamic_prop_keys = context->getDynamicPropertyKeys();
   logger_->log_info("RetryFlowFile registering %d keys", dynamic_prop_keys.size());
   for (const auto& key : dynamic_prop_keys) {
-    exceeded_flowfile_attribute_keys.emplace_back(core::PropertyBuilder::createProperty(key)->withDescription("auto generated")->supportsExpressionLanguage(true)->build());
+    exceeded_flowfile_attribute_keys_.emplace_back(core::PropertyBuilder::createProperty(key)->withDescription("auto generated")->supportsExpressionLanguage(true)->build());
     logger_->log_info("RetryFlowFile registered attribute '%s'", key);
   }
 }
 
-// Returns (1, true) on non-numerical or out-of-bounds retry value
-std::pair<uint64_t, bool> RetryFlowFile::getRetryPropertyValue(const std::shared_ptr<FlowFileRecord>& flow_file) {
+utils::optional<uint64_t> RetryFlowFile::getRetryPropertyValue(const std::shared_ptr<FlowFileRecord>& flow_file) const {
   std::string value_as_string;
   try {
     if (flow_file->getAttribute(retry_attribute_, value_as_string)) {
-      return std::make_pair(std::stoul(value_as_string), false);
+      return utils::make_optional<uint64_t>(std::stoull(value_as_string));
     }
   }
   catch(const std::invalid_argument&) {
     if (fail_on_non_numerical_overwrite_) {
       logger_->log_info("Non-numerical retry property in RetryFlowFile. Sending flowfile to failure...", value_as_string);
-      return std::make_pair(1, true);
+      return {};
     }
-    logger_->log_info("Non-numerical retry property in RetryFlowFile: overwriting %s with 1.", value_as_string);
+    logger_->log_info("Non-numerical retry property in RetryFlowFile: overwriting %s with 0.", value_as_string);
   }
   catch(const std::out_of_range&) {
     logger_->log_error("Narrowing Exception for %s, treating it as non-numerical value", value_as_string);
   }
-  return std::make_pair(1, false);
+  return utils::make_optional<uint64_t>(0);
 }
 
 // Returns true on fail on reuse scenario
-bool RetryFlowFile::updateUUIDMarkerAndCheckFailOnReuse(const std::shared_ptr<FlowFileRecord>& flow_file) {
+bool RetryFlowFile::updateUUIDMarkerAndCheckFailOnReuse(const std::shared_ptr<FlowFileRecord>& flow_file) const {
   const std::string last_retried_by_property_name = retry_attribute_ + ".uuid";
   const std::string current_processor_uuid = getUUIDStr();
   std::string last_retried_by_uuid;
@@ -189,9 +188,9 @@ bool RetryFlowFile::updateUUIDMarkerAndCheckFailOnReuse(const std::shared_ptr<Fl
   return false;
 }
 
-bool RetryFlowFile::setRetriesExceededAttributesOnFlowFile(core::ProcessContext* context, const std::shared_ptr<FlowFileRecord>& flow_file) {
+bool RetryFlowFile::setRetriesExceededAttributesOnFlowFile(core::ProcessContext* context, const std::shared_ptr<FlowFileRecord>& flow_file) const {
   try {
-    for (const auto& attribute : exceeded_flowfile_attribute_keys) {
+    for (const auto& attribute : exceeded_flowfile_attribute_keys_) {
       std::string value;
       context->getDynamicProperty(attribute, value, flow_file);
       flow_file->setAttribute(attribute.getName(), value);
--- a/extensions/standard-processors/processors/RetryFlowFile.cpp
+++ b/extensions/standard-processors/processors/RetryFlowFile.cpp
@@ -151,21 +151,19 @@ void RetryFlowFile::readDynamicPropertyKeys(core::ProcessContext* context) {
 
 utils::optional<uint64_t> RetryFlowFile::getRetryPropertyValue(const std::shared_ptr<FlowFileRecord>& flow_file) const {
   std::string value_as_string;
+  flow_file->getAttribute(retry_attribute_, value_as_string);
+  uint64_t value;
   try {
-    if (flow_file->getAttribute(retry_attribute_, value_as_string)) {
-      return utils::make_optional<uint64_t>(std::stoull(value_as_string));
-    }
+    utils::internal::ValueParser(value_as_string).parse(value).parseEnd();
+    return utils::make_optional<uint64_t>(value);
   }
-  catch(const std::invalid_argument&) {
+  catch(const utils::internal::ParseException&) {
     if (fail_on_non_numerical_overwrite_) {
-      logger_->log_info("Non-numerical retry property in RetryFlowFile. Sending flowfile to failure...", value_as_string);
+      logger_->log_info("Non-numerical retry property in RetryFlowFile (value: %s). Sending flowfile to failure...", value_as_string);
       return {};
     }
-    logger_->log_info("Non-numerical retry property in RetryFlowFile: overwriting %s with 0.", value_as_string);
   }
-  catch(const std::out_of_range&) {
-    logger_->log_error("Narrowing Exception for %s, treating it as non-numerical value", value_as_string);
-  }
+  logger_->log_info("Non-numerical retry property in RetryFlowFile: overwriting %s with 0.", value_as_string);
   return utils::make_optional<uint64_t>(0);
 }
 
--- a/extensions/standard-processors/processors/RetryFlowFile.cpp
+++ b/extensions/standard-processors/processors/RetryFlowFile.cpp
@@ -90,7 +90,7 @@ void RetryFlowFile::onSchedule(core::ProcessContext* context, core::ProcessSessi
   context->getProperty(PenalizeRetries.getName(), penalize_retries_);
   context->getProperty(FailOnNonNumericalOverwrite.getName(), fail_on_non_numerical_overwrite_);
   context->getProperty(ReuseMode.getName(), reuse_mode_);
-  readDynamicProperties(context);
+  readDynamicPropertyKeys(context);
 }
 
 void RetryFlowFile::onTrigger(core::ProcessContext* context, core::ProcessSession* session) {
@@ -134,7 +134,7 @@ void RetryFlowFile::onTrigger(core::ProcessContext* context, core::ProcessSessio
   session->transfer(flow_file, RetriesExceeded);
 }
 
-void RetryFlowFile::readDynamicProperties(core::ProcessContext* context) {
+void RetryFlowFile::readDynamicPropertyKeys(core::ProcessContext* context) {
   exceeded_flowfile_attribute_keys.clear();
   const std::vector<std::string> dynamic_prop_keys = context->getDynamicPropertyKeys();
   logger_->log_info("RetryFlowFile registering %d keys", dynamic_prop_keys.size());
@@ -144,6 +144,7 @@ void RetryFlowFile::readDynamicProperties(core::ProcessContext* context) {
   }
 }
 
+// Returns (1, true) on non-numerical or out-of-bounds retry value
 std::pair<uint64_t, bool> RetryFlowFile::getRetryPropertyValue(const std::shared_ptr<FlowFileRecord>& flow_file) {
   std::string value_as_string;
   try {
@@ -164,6 +165,7 @@ std::pair<uint64_t, bool> RetryFlowFile::getRetryPropertyValue(const std::shared
   return std::make_pair(1, false);
 }
 
+// Returns true on fail on reuse scenario
 bool RetryFlowFile::updateUUIDMarkerAndCheckFailOnReuse(const std::shared_ptr<FlowFileRecord>& flow_file) {
   const std::string last_retried_by_property_name = retry_attribute_ + ".uuid";
   const std::string current_processor_uuid = getUUIDStr();
--- a/extensions/standard-processors/processors/RetryFlowFile.cpp
+++ b/extensions/standard-processors/processors/RetryFlowFile.cpp
@@ -1,0 +1,210 @@
+/**
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "RetryFlowFile.h"
+
+#include "core/PropertyValidation.h"
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace processors {
+
+core::Property RetryFlowFile::RetryAttribute(core::PropertyBuilder::createProperty("Retry Attribute")
+    ->withDescription(
+        "The name of the attribute that contains the current retry count for the FlowFile."
+        "WARNING: If the name matches an attribute already on the FlowFile that does not contain a numerical value, "
+        "the processor will either overwrite that attribute with '1' or fail based on configuration.")
+    ->withDefaultValue("flowfile.retries")
+    ->supportsExpressionLanguage(true)
+    ->build());
+
+core::Property RetryFlowFile::MaximumRetries(core::PropertyBuilder::createProperty("Maximum Retries")
+    ->withDescription("The maximum number of times a FlowFile can be retried before being passed to the 'retries_exceeded' relationship.")
+    ->withDefaultValue<uint64_t>(3)
+    ->supportsExpressionLanguage(true)
+    ->build());
+
+core::Property RetryFlowFile::PenalizeRetries(core::PropertyBuilder::createProperty("Penalize Retries")
+  ->withDescription("If set to 'true', this Processor will penalize input FlowFiles before passing them to the 'retry' relationship. This does not apply to the 'retries_exceeded' relationship.")
+  ->withDefaultValue<bool>(true)
+  ->build());
+
+core::Property RetryFlowFile::FailOnNonNumericalOverwrite(core::PropertyBuilder::createProperty("Fail on Non-numerical Overwrite")
+    ->withDescription("If the FlowFile already has the attribute defined in 'Retry Attribute' that is *not* a number, fail the FlowFile instead of resetting that value to '1'")
+    ->withDefaultValue<bool>(false)
+    ->build());
+
+core::Property RetryFlowFile::ReuseMode(core::PropertyBuilder::createProperty("Reuse Mode")
+    ->withDescription(
+        "Defines how the Processor behaves if the retry FlowFile has a different retry UUID than "
+        "the instance that received the FlowFile. This generally means that the attribute was "
+        "not reset after being successfully retried by a previous instance of this processor.")
+    ->withAllowableValues<std::string>({FAIL_ON_REUSE, WARN_ON_REUSE, RESET_REUSE})
+    ->withDefaultValue(FAIL_ON_REUSE)
+    ->build());
+
+core::Relationship RetryFlowFile::Retry("retry",
+  "Input FlowFile has not exceeded the configured maximum retry count, pass this relationship back to the input Processor to create a limited feedback loop.");
+core::Relationship RetryFlowFile::RetriesExceeded("retries_exceeded",
+  "Input FlowFile has exceeded the configured maximum retry count, do not pass this relationship back to the input Processor to terminate the limited feedback loop.");
+core::Relationship RetryFlowFile::Failure("failure",
+    "The processor is configured such that a non-numerical value on 'Retry Attribute' results in a failure instead of resetting "
+    "that value to '1'. This will immediately terminate the limited feedback loop. Might also include when 'Maximum Retries' contains "
+    " attribute expression language that does not resolve to an Integer.");
+
+void RetryFlowFile::initialize() {
+  setSupportedProperties({
+    RetryAttribute,
+    MaximumRetries,
+    PenalizeRetries,
+    FailOnNonNumericalOverwrite,
+    ReuseMode,
+  });
+  setSupportedRelationships({
+    Retry,
+    RetriesExceeded,
+    Failure,
+  });
+}
+
+void RetryFlowFile::onSchedule(core::ProcessContext* context, core::ProcessSessionFactory* /* sessionFactory */) {
+  context->getProperty(RetryAttribute.getName(), retry_attribute_);
+  context->getProperty(MaximumRetries.getName(), maximum_retries_);
+  context->getProperty(PenalizeRetries.getName(), penalize_retries_);
+  context->getProperty(FailOnNonNumericalOverwrite.getName(), fail_on_non_numerical_overwrite_);
+  context->getProperty(ReuseMode.getName(), reuse_mode_);
+  readDynamicProperties(context);
+}
+
+void RetryFlowFile::onTrigger(core::ProcessContext* context, core::ProcessSession* session) {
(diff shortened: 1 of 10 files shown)


extensions/standard-processors/tests/unit/RetryFlowFileTests.cpp
```diff
@@ -0,0 +1,221 @@
...
+    logTestController_.setDebug<UpdateAttribute>();
+    logTestController_.setDebug<RetryFlowFile>();
+    logTestController_.setDebug<PutFile>();
+    logTestController_.setDebug<PutFile>();
+    logTestController_.setDebug<LogAttribute>();
+    logTestController_.setDebug<core::ProcessSession>();
+    logTestController_.setDebug<core::Connectable>();
+    logTestController_.setDebug<minifi::Connection>();
+  }
+
```
hunyadi-dev, line 77: Here is an ASCII art diagram for the layout:
```
+--------------------+
|  GenerateFlowFile  |
+--------------------+
|       success      |
+--------------------+
           |
           v                   +-----------+
+-------------------+          |  PutFile  |
|  UpdateAttribute  |          +-----------+
+-------------------+   +----->+  success  +--+
|      success      |   |      +-----------+  |
+-------------------+   |                     |
           |            |      +-----------+  |   +----------------+
           v            | +--->+  PutFile  |  |   |  LogAttribute  |
+--------------------+  | |    +-----------+  |   +----------------+
|   RetryFlowFile    |  | |    |  success  +----->+     success    +-X
+--------------------+  | |    +-----------+  |   +----------------+
|        retry       +--+ |                   |
+--------------------+    |    +-----------+  |
|  retries_exceeded  +----+ +->+  PutFile  |  |
+--------------------+      |  +-----------+  |
|       failure      +------+  |  success  +--+
+--------------------+         +-----------+
```

extensions/standard-processors/processors/RetryFlowFile.cpp
```diff
@@ -0,0 +1,212 @@
...
+    Failure,
+  });
+}
+
+void RetryFlowFile::onSchedule(core::ProcessContext* context, core::ProcessSessionFactory* /* sessionFactory */) {
+  context->getProperty(RetryAttribute.getName(), retry_attribute_);
+  context->getProperty(MaximumRetries.getName(), maximum_retries_);
+  context->getProperty(PenalizeRetries.getName(), penalize_retries_);
+  context->getProperty(FailOnNonNumericalOverwrite.getName(), fail_on_non_numerical_overwrite_);
+  context->getProperty(ReuseMode.getName(), reuse_mode_);
```
adamdebreceni, line 92: although EFM only allows selecting the ones specified in `withAllowableValues`, from the yml we could set it to a value not in this set, this verification should eventually happen before `onSchedule` is called, but for now we have to manually check it

extensions/standard-processors/processors/RetryFlowFile.cpp
```diff
@@ -0,0 +1,212 @@
...
+  try {
+    for (const auto& attribute : exceeded_flowfile_attribute_keys) {
+      std::string value;
+      context->getDynamicProperty(attribute, value, flow_file);
+      flow_file->setAttribute(attribute.getName(), value);
+      logger_->log_info("Set attribute '%s' of flow file '%s' with value '%s'", attribute.getName(), flow_file->getUUIDStr(), value);
+    }
+    return true;
+  }
+  catch (const std::exception& e) {
```
arpadboda, line 202: What do we expect to throw here?

extensions/standard-processors/processors/RetryFlowFile.cpp
```diff
@@ -0,0 +1,212 @@
...
+      } else {  // Assuming reuse_mode_ == RESET_REUSE
+        logger_->log_debug("Reusing retry attribute that belongs to different processor. Resetting value to 1.");
+      }
+    }
+  }
+  flow_file->setAttribute(last_retried_by_property_name, getUUIDStr());
+  return false;
+}
+
+bool RetryFlowFile::setRetriesExceededAttributesOnFlowFile(core::ProcessContext* context, const std::shared_ptr<FlowFileRecord>& flow_file) {
```
arpadboda, line 192: Const!

I don't think this function should return true or false. 

This should be void. In case something throws when trying to set attributes (shouldn't), just let the exception fly and that results in a rollback.

extensions/standard-processors/processors/RetryFlowFile.cpp
```diff
@@ -0,0 +1,212 @@
...
+  context->getProperty(PenalizeRetries.getName(), penalize_retries_);
+  context->getProperty(FailOnNonNumericalOverwrite.getName(), fail_on_non_numerical_overwrite_);
+  context->getProperty(ReuseMode.getName(), reuse_mode_);
+  readDynamicPropertyKeys(context);
+}
+
+void RetryFlowFile::onTrigger(core::ProcessContext* context, core::ProcessSession* session) {
+  std::shared_ptr<FlowFileRecord> flow_file = std::static_pointer_cast<FlowFileRecord> (session->get());
+  if (!flow_file) {
+    return;
```
arpadboda, line 99: yield

extensions/standard-processors/processors/RetryFlowFile.cpp
```diff
@@ -0,0 +1,212 @@
...
+  }
+  if (updateUUIDMarkerAndCheckFailOnReuse(flow_file)) {
+    session->transfer(flow_file, Failure);
+    return;
+  }
+
+  if (retry_property_value < maximum_retries_) {
+    try {
+      flow_file->setAttribute(retry_attribute_, std::to_string(gsl::narrow_cast<uint64_t>(retry_property_value + 1)));
```
adamdebreceni, line 116: the `narrow_cast` has no effect, the `retry_property_value` being `uint64_t` would silently wrap around if it could, but since you check if it is less than `maximum_retries_` (also being `uint64_t`) it is safe to increment

extensions/standard-processors/processors/RetryFlowFile.cpp
```diff
@@ -0,0 +1,212 @@
...
+    logger_->log_info("Non-numerical retry property in RetryFlowFile: overwriting %s with 1.", value_as_string);
+  }
+  catch(const std::out_of_range&) {
+    logger_->log_error("Narrowing Exception for %s, treating it as non-numerical value", value_as_string);
+  }
+  return std::make_pair(1, false);
+}
+
+// Returns true on fail on reuse scenario
+bool RetryFlowFile::updateUUIDMarkerAndCheckFailOnReuse(const std::shared_ptr<FlowFileRecord>& flow_file) {
```
arpadboda, line 169: This looks const as well, doesn't seem to change anything in the processor.

I would prefer to have this separated into two functions: one to check reuse, the other to update.

extensions/standard-processors/processors/RetryFlowFile.cpp
```diff
@@ -0,0 +1,212 @@
...
+  if (updateUUIDMarkerAndCheckFailOnReuse(flow_file)) {
+    session->transfer(flow_file, Failure);
+    return;
+  }
+
+  if (retry_property_value < maximum_retries_) {
+    try {
+      flow_file->setAttribute(retry_attribute_, std::to_string(gsl::narrow_cast<uint64_t>(retry_property_value + 1)));
+    }
+    catch(const gsl::narrowing_error& e) {
```
arpadboda, line 118: How could it occur?

If maximum_retries_ is a valid uint64 and retry_property_value is less, I don't think incrementing it by one can cause any issues anyhow.

extensions/standard-processors/tests/unit/RetryFlowFileTests.cpp
```diff
@@ -0,0 +1,226 @@
...
+#include "utils/OptionalUtils.h"
+#include "utils/RegexUtils.h"
+#include "utils/TestUtils.h"
+
+namespace {
+using org::apache::nifi::minifi::utils::createTempDir;
+using org::apache::nifi::minifi::utils::optional;
+
+std::vector<std::pair<std::string, std::string>> list_dir_all(const std::string& dir, const std::shared_ptr<logging::Logger>& logger, bool recursive = true) {
+  return org::apache::nifi::minifi::utils::file::FileUtils::list_dir_all(dir, logger, recursive);
```
arpadboda, line 42: Why? Using would do the job.

extensions/standard-processors/processors/RetryFlowFile.cpp
```diff
@@ -0,0 +1,212 @@
...
+    logger_->log_info("RetryFlowFile registered attribute '%s'", key);
+  }
+}
+
+// Returns (1, true) on non-numerical or out-of-bounds retry value
+std::pair<uint64_t, bool> RetryFlowFile::getRetryPropertyValue(const std::shared_ptr<FlowFileRecord>& flow_file) {
+  std::string value_as_string;
+  try {
+    if (flow_file->getAttribute(retry_attribute_, value_as_string)) {
+      return std::make_pair(std::stoul(value_as_string), false);
```
adamdebreceni, line 152: watch our as `std::stoul` does many counterintuitive things, e.g. it doesn't try to parse till the end of the string, meaning `std::stoul("1banana")` will happily return `1`, moreover `std::stoul("-1")` happily returns the max `unsigned long`, check out #797 `ValueParser.h` which aims to tackle these problems

libminifi/test/TestBase.h
```diff
@@ -251,6 +251,8 @@ class TestPlan {
   std::shared_ptr<core::Processor> addProcessor(const std::string &processor_name, utils::Identifier& uuid, const std::string &name, const std::initializer_list<core::Relationship>& relationships,
                                                 bool linkToPrevious = false);
 
+  std::shared_ptr<minifi::Connection> addConnection(const std::shared_ptr<core::Processor> source_proc, const core::Relationship& source_relationship, const std::shared_ptr<core::Processor>& destination_proc);
```
arpadboda, line 254: I don't think the relationship being in the middle is the best signature, but this is only my personal opinion, I can live with this in case the majority prefers this.

extensions/standard-processors/processors/RetryFlowFile.cpp
```diff
@@ -0,0 +1,212 @@
...
+  const std::vector<std::string> dynamic_prop_keys = context->getDynamicPropertyKeys();
+  logger_->log_info("RetryFlowFile registering %d keys", dynamic_prop_keys.size());
+  for (const auto& key : dynamic_prop_keys) {
+    exceeded_flowfile_attribute_keys.emplace_back(core::PropertyBuilder::createProperty(key)->withDescription("auto generated")->supportsExpressionLanguage(true)->build());
+    logger_->log_info("RetryFlowFile registered attribute '%s'", key);
+  }
+}
+
+// Returns (1, true) on non-numerical or out-of-bounds retry value
+std::pair<uint64_t, bool> RetryFlowFile::getRetryPropertyValue(const std::shared_ptr<FlowFileRecord>& flow_file) {
```
adamdebreceni, line 148: seems like a job for `optional`

extensions/standard-processors/processors/RetryFlowFile.cpp
adamdebreceni, line 169: I'm a big fan of using enums even for such simple cases e.g. 
```
enum class FlowFileAction{
   CONTINUE,
   ROUTE_TO_FAILURE
}
```
this greatly enhances the information content of the function signature as well

(for `setRetriesExceededAttributesOnFlowFile ` as well)

extensions/standard-processors/processors/RetryFlowFile.cpp
```diff
@@ -0,0 +1,212 @@
...
+  if (flow_file->getAttribute(last_retried_by_property_name, last_retried_by_uuid)) {
+    if (last_retried_by_uuid != current_processor_uuid) {
+      if (reuse_mode_ == FAIL_ON_REUSE) {
+        logger_->log_error("FlowFile %s was previously retried with the same attribute by a different "
+            "processor (uuid: %s, current uuid: %s). Transfering flowfile to 'failure'...",
+            flow_file->getUUIDStr(), last_retried_by_uuid, current_processor_uuid);
+        return true;
+      }
+      if (reuse_mode_ == WARN_ON_REUSE) {
+        logger_->log_warn("Reusing retry attribute that belongs to different processor. Resetting value to 1.");
```
adamdebreceni, line 182: where is it actually reset to `1`?

extensions/standard-processors/processors/RetryFlowFile.cpp
hunyadi-dev: This exact pattern occurs in the following processors:
 1. ApplyTemplate
 1. ListenHTTP
 1. PutSQL
 1. RetryFlowFile
 1. RouteOnAttribute
 1. UpdateAttribute
 1. TFApplyGraph
 1. TFConvertImageToTensor
 1. TFExtractTopLabels

extensions/standard-processors/processors/RetryFlowFile.cpp
```diff
@@ -0,0 +1,212 @@
...
+    if (fail_on_non_numerical_overwrite_) {
+      logger_->log_info("Non-numerical retry property in RetryFlowFile. Sending flowfile to failure...", value_as_string);
+      return std::make_pair(1, true);
+    }
+    logger_->log_info("Non-numerical retry property in RetryFlowFile: overwriting %s with 1.", value_as_string);
+  }
+  catch(const std::out_of_range&) {
+    logger_->log_error("Narrowing Exception for %s, treating it as non-numerical value", value_as_string);
+  }
+  return std::make_pair(1, false);
```
adamdebreceni, line 165: if you return `1` when `maximum_retries_` is `1` then you will never actually `retry` the (previously) malformed flowfile, is this by design?

extensions/standard-processors/processors/RetryFlowFile.h
```diff
@@ -0,0 +1,121 @@
...
+  std::pair<uint64_t, bool> getRetryPropertyValue(const std::shared_ptr<FlowFileRecord>& flow_file);
+  // Returns true on fail on reuse scenario
+  bool updateUUIDMarkerAndCheckFailOnReuse(const std::shared_ptr<FlowFileRecord>& flow_file);
+  bool setRetriesExceededAttributesOnFlowFile(core::ProcessContext* context, const std::shared_ptr<FlowFileRecord>& flow_file);
+
+  std::string retry_attribute_;
+  uint64_t maximum_retries_ = 3;  // The real default value is set by the default on the MaximumRetries property
+  bool penalize_retries_ =  true;  // The real default value is set by the default on the PenalizeRetries property
+  bool fail_on_non_numerical_overwrite_ = false;  // The real default value is set by the default on the FailOnNonNumericalOverwrite property
+  std::string reuse_mode_;
```
adamdebreceni, line 100: could be an enum

libminifi/include/Connection.h
```diff
@@ -59,11 +59,11 @@ class Connection : public core::Connectable, public std::enable_shared_from_this
   }
 
   // Set Source Processor UUID
-  void setSourceUUID(utils::Identifier &uuid) {
+  void setSourceUUID(const utils::Identifier& uuid) {
```
adamdebreceni, line 62: 👍

extensions/standard-processors/processors/RetryFlowFile.cpp
hunyadi-dev: I do not know, probably some exception on trying to set attributes on the flowfile that should not be set, like "UUID". This pattern follows the one set in `UpdateAttribute`. I can potentially extract the functionality to a helper function.

libminifi/test/TestBase.h
adamdebreceni, line 254: missing `&` from `source_proc`

libminifi/test/TestBase.cpp
```diff
@@ -168,6 +168,30 @@ std::shared_ptr<core::Processor> TestPlan::addProcessor(const std::string &proce
   return addProcessor(processor_name, uuid, name, relationships, linkToPrevious);
 }
 
+std::shared_ptr<minifi::Connection> TestPlan::addConnection(const std::shared_ptr<core::Processor> source_proc, const core::Relationship& source_relationship, const std::shared_ptr<core::Processor>& destination_proc) {
+  std::stringstream connection_name;
+  connection_name << source_proc->getUUIDStr() << "-to-" << destination_proc->getUUIDStr();
```
adamdebreceni, line 173: info: EFM names connections as `<source proc name>/<relationship name>/<dest proc name>`

extensions/standard-processors/processors/RetryFlowFile.cpp
hunyadi-dev: Will be marked const.

extensions/standard-processors/processors/RetryFlowFile.cpp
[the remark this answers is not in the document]
hunyadi-dev: Valid point, will be corrected.

extensions/standard-processors/processors/RetryFlowFile.cpp
arpadboda: I simply don't think we should silently swallow such error (even if we log).
Not being able to set an attribute is a serious core framework issue, I would let that exception propagate.

extensions/standard-processors/processors/RetryFlowFile.cpp
arpadboda: Everything done in the given onTrigger execution (adding/modifying attributes, routing, creating flowfiles) is rolled back.

extensions/standard-processors/processors/RetryFlowFile.cpp
arpadboda: Some might miss it, it's not a big issue.

extensions/standard-processors/processors/RetryFlowFile.cpp
hunyadi-dev: Good catch, I should probably set the value to 0 here and have it incremented in the `retry` branch.

extensions/standard-processors/processors/RetryFlowFile.cpp
hunyadi-dev: So do I want to yield here?

extensions/standard-processors/processors/RetryFlowFile.cpp
hunyadi-dev: Would this be solved if I added a custom validator to the default value?

extensions/standard-processors/processors/RetryFlowFile.cpp
hunyadi-dev: I can add this, but it would pollute the class declaration. Do you guys think that even with the comment the usage is not obvious?

extensions/standard-processors/processors/RetryFlowFile.cpp
hunyadi-dev: 😱 it is not. Will add a reset to 0 instead of 1 on the flowfile attribute here as well.

extensions/standard-processors/processors/RetryFlowFile.cpp
hunyadi-dev: Will change this to throw.

extensions/standard-processors/processors/RetryFlowFile.cpp
adamdebreceni: it could but I wouldn't go that way, eventually it should be handled automatically at the `Property` level, creating a `SetValidator` class and whatnot, I don't think it should be done in this story, [MINIFICPP-1238](https://issues.apache.org/jira/browse/MINIFICPP-1238) is better suited

extensions/standard-processors/processors/RetryFlowFile.h
hunyadi-dev: This is definitely not worth the trouble. There is already a list of strings of allowable values, adding an enum and keeping the two sync is just an extra layer of complexity.

libminifi/test/TestBase.cpp
hunyadi-dev: `TestPlan` does it like this. I would like to refactor this class at some point anyway, want me to change the names then?

libminifi/test/TestBase.h
hunyadi-dev: Well, this reflects the timeline of events.

hunyadi-dev: Good catch.

libminifi/test/TestBase.cpp
adamdebreceni: nope, it was just an info, since `TestPlan` does it like this and it is it's method, I'm good

extensions/standard-processors/processors/RetryFlowFile.cpp
hunyadi-dev: Correct, this check will be removed.

extensions/standard-processors/processors/RetryFlowFile.cpp
hunyadi-dev: See other thread above.

extensions/standard-processors/tests/unit/RetryFlowFileTests.cpp
hunyadi-dev: Yeah, this was a stupid idea for what it is worth. `FileUtils` is a class, so its methods cannot be used with `using`, so instead of stepping back I went on with this. I will put an alias to FileUtils here instead :)

extensions/standard-processors/processors/RetryFlowFile.cpp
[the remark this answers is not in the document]
hunyadi-dev: Corrected.

hunyadi-dev: Replaced with optional.

extensions/standard-processors/processors/RetryFlowFile.cpp
hunyadi-dev: Try-catch block removed.

extensions/standard-processors/processors/RetryFlowFile.cpp
hunyadi-dev: Well, this is not merged yet, so using `std::stoull` is the best I can do. I do not think having the issues you mentioned should be a big problem though, I can live with those conversions.

extensions/standard-processors/tests/unit/RetryFlowFileTests.cpp
hunyadi-dev: Replaced.

extensions/standard-processors/processors/RetryFlowFile.cpp
arpadboda: They just do it outside of the function.
I believe in single responsibility, one function is responsible for one thing, one class is responsible for one logical unit of something.

hunyadi-dev: Resolved by moving the function content to `onTrigger` as recommended.

hunyadi-dev: Resolved by moving the function content to `onTrigger` as you recommended.

extensions/standard-processors/tests/unit/RetryFlowFileTests.cpp
szaszm: There are no rules against ASCII art that I know of. I previously expressed that I didn't like them personally, but at the end of the discussion the conclusion was that they provide value to the reader, therefore they are useful.

I am just a contributor to the project and I happened to contribute many reviews lately, because I believe ensuring that no easily preventable issues are submitted, and the code is of the highest quality obtainable with relative ease is usually the most efficient use of my time. I believe it provides more value on the long term by enabling contributors (myself and others) to work with the code more effectively.

Having said that, my opinions and review suggestions are my attempt at being helpful, not rules. I think they become de jure rules when someone proposes it and the community agrees. On the other hand, they may easily become de facto "rules", that is, if I (or anybody) made a suggestion that improved code in some way, then I would expect the same contributors to try to submit an already "improved" version of similar code in their future contributions. I usually consider de facto "rules" conventions rather than rules.

I'm writing this because I don't want you or any contributors to be discouraged by my past reviews.

tl;dr: my comments are not rules

hunyadi-dev: Thanks, sorry for using the word "rules allow it", I should have written "if others think there is added value in moving this to comment in the codebase".

extensions/standard-processors/processors/RetryFlowFile.cpp
hunyadi-dev: I checked the corresponding NiFi implementation, and they also do not yield in this use case. I think there is no point in yielding on event-driven processors when they are triggered without flowfiles.

extensions/standard-processors/processors/RetryFlowFile.cpp
hunyadi-dev: Updated using `ValueParser`.

extensions/standard-processors/processors/RetryFlowFile.cpp
```diff
@@ -0,0 +1,183 @@
...
+    return;
+  }
+
+  bool failure_due_to_non_numerical_retry;
+  utils::optional<uint64_t> maybe_retry_property_value = getRetryPropertyValue(flow_file);
+  if (!maybe_retry_property_value) {
+    session->transfer(flow_file, Failure);
+    return;
+  }
+  uint64_t retry_property_value = maybe_retry_property_value.value_or(0);
```
adamdebreceni, line 108: using `value_or` could indicate that we must be prepared to handle an empty optional even though that cannot be true here

extensions/standard-processors/processors/RetryFlowFile.cpp
```diff
@@ -0,0 +1,183 @@
...
+        logger_->log_error("FlowFile %s was previously retried with the same attribute by a different "
+            "processor (uuid: %s, current uuid: %s). Transfering flowfile to 'failure'...",
+            flow_file->getUUIDStr(), last_retried_by_uuid, current_processor_uuid);
+        session->transfer(flow_file, Failure);
+        return;
+      }
+      if (reuse_mode_ == WARN_ON_REUSE) {
+        logger_->log_warn("Reusing retry attribute that belongs to different processor. Resetting value to 0.");
+      } else {  // Assuming reuse_mode_ == RESET_REUSE
+        logger_->log_debug("Reusing retry attribute that belongs to different processor. Resetting value to 0.");
```
adamdebreceni, line 124: I would prefer an explicit check and throwing an exception in `else`

extensions/standard-processors/processors/RetryFlowFile.cpp
```diff
@@ -0,0 +1,183 @@
...
+namespace nifi {
+namespace minifi {
+namespace processors {
+
+core::Property RetryFlowFile::RetryAttribute(core::PropertyBuilder::createProperty("Retry Attribute")
+    ->withDescription(
+        "The name of the attribute that contains the current retry count for the FlowFile."
+        "WARNING: If the name matches an attribute already on the FlowFile that does not contain a numerical value, "
+        "the processor will either overwrite that attribute with '1' or fail based on configuration.")
+    ->withDefaultValue("flowfile.retries")
```
arpadboda, line 34: Shouldn't this be required?

extensions/standard-processors/processors/RetryFlowFile.cpp
```diff
@@ -0,0 +1,183 @@
...
+    ->supportsExpressionLanguage(true)
+    ->build());
+
+core::Property RetryFlowFile::MaximumRetries(core::PropertyBuilder::createProperty("Maximum Retries")
+    ->withDescription("The maximum number of times a FlowFile can be retried before being passed to the 'retries_exceeded' relationship.")
+    ->withDefaultValue<uint64_t>(3)
```
arpadboda, line 40: This as well

extensions/standard-processors/processors/RetryFlowFile.cpp
```diff
@@ -0,0 +1,183 @@
...
+    ->withDefaultValue<bool>(false)
+    ->build());
+
+core::Property RetryFlowFile::ReuseMode(core::PropertyBuilder::createProperty("Reuse Mode")
+    ->withDescription(
+        "Defines how the Processor behaves if the retry FlowFile has a different retry UUID than "
+        "the instance that received the FlowFile. This generally means that the attribute was "
+        "not reset after being successfully retried by a previous instance of this processor.")
+    ->withAllowableValues<std::string>({FAIL_ON_REUSE, WARN_ON_REUSE, RESET_REUSE})
+    ->withDefaultValue(FAIL_ON_REUSE)
```
arpadboda, line 60: This as well

extensions/standard-processors/processors/RetryFlowFile.cpp
arpadboda: I don't think the exception is fine, that would end up in a rollback. 

Although I would prefer to avoid duplication of the string. 
log_warn is just an alias, log level can be a parameter.

extensions/standard-processors/tests/unit/RetryFlowFileTests.cpp
arpadboda: These tests are nice as they cover all possible scenarios, although I would prefer to have a testcase that simulates the real life usecase: a loop in the flow and the flowfile getting out of that after the retry count is exceeded. 

This one can be an integration one using a predefined yaml in case creating such flow using testplan is too painful.

extensions/standard-processors/processors/RetryFlowFile.cpp
adamdebreceni: the problem is that we currently don't verify anywhere that the `reuse_mode_` indeed contains one of the three allowed values, and not doing anything (i.e. rollback) is better than doing something (possibly unintended) on invalid user input

extensions/standard-processors/processors/RetryFlowFile.cpp
adamdebreceni: it might be beneficial to make all new properties required by default, and if in the future it proves to be a pain point, we could always give a default based on real usage requirements, while if we would want to make something required later we would break previous flows

hunyadi-dev: The NiFi implementation has this default value set:
http://nifi.apache.org/docs/nifi-docs/components/org.apache.nifi/nifi-standard-nar/1.11.4/org.apache.nifi.processors.standard.RetryFlowFile/index.html

extensions/standard-processors/processors/RetryFlowFile.cpp
hunyadi-dev: Same as above.

extensions/standard-processors/processors/RetryFlowFile.cpp
hunyadi-dev: [Same as above.](http://nifi.apache.org/docs/nifi-docs/components/org.apache.nifi/nifi-standard-nar/1.11.4/org.apache.nifi.processors.standard.RetryFlowFile/index.html)

extensions/standard-processors/processors/RetryFlowFile.cpp
hunyadi-dev: Good catch, modified this as we handle failure on non numerical retry inside the retry property reader. Also removed unused variable declared on line 102.

extensions/standard-processors/processors/RetryFlowFile.cpp
hunyadi-dev: We have `withAllowableValues<std::string>({FAIL_ON_REUSE, WARN_ON_REUSE, RESET_REUSE})`. It should be sufficient for validation of the string. I will change the code to use logger aliasing though.

hunyadi-dev: Will need to change the logger interface to access this log_string (was previously protected). Asked @arpadboda on this change and this should be obviously a feature we want people to have access to.

extensions/standard-processors/tests/unit/RetryFlowFileTests.cpp
hunyadi-dev: Added Jira here:
https://issues.apache.org/jira/browse/MINIFICPP-1285

extensions/standard-processors/processors/RetryFlowFile.cpp
arpadboda: Looking at NiFi code is always a good source of hints and what we see there worth consideration, but it's not an unerring guidance or revelation. 
In case we think that our implementation definitely requires a valid value for the given property to work, I think we should make it required.

In case of properties that define attribute names I would also consider something like this: https://www.javadoc.io/doc/org.apache.nifi/nifi-processor-utils/0.7.0/org/apache/nifi/processor/util/StandardValidators.html#NON_BLANK_VALIDATOR
As we request a valid string here that can name an attribute.

extensions/standard-processors/processors/RetryFlowFile.cpp
```diff
@@ -0,0 +1,183 @@
...
+    if (last_retried_by_uuid != current_processor_uuid) {
+      if (reuse_mode_ == FAIL_ON_REUSE) {
+        logger_->log_error("FlowFile %s was previously retried with the same attribute by a different "
+            "processor (uuid: %s, current uuid: %s). Transfering flowfile to 'failure'...",
+            flow_file->getUUIDStr(), last_retried_by_uuid, current_processor_uuid);
+        session->transfer(flow_file, Failure);
+        return;
+      }
+      if (reuse_mode_ == WARN_ON_REUSE) {
+        reuse_mode_log_level_ = core::logging::LOG_LEVEL::warn;
```
adamdebreceni, line 121: `reuse_mode_log_level_` is local to this segment of the code, if it later turns out to be useful throughout the different processor methods we can make it a field, but for now I'd rather it be a local variable of the smallest possible scope

hunyadi-dev: Localized.

extensions/standard-processors/processors/RetryFlowFile.cpp
```diff
@@ -0,0 +1,183 @@
...
+    if (last_retried_by_uuid != current_processor_uuid) {
+      if (reuse_mode_ == FAIL_ON_REUSE) {
+        logger_->log_error("FlowFile %s was previously retried with the same attribute by a different "
+            "processor (uuid: %s, current uuid: %s). Transfering flowfile to 'failure'...",
+            flow_file->getUUIDStr(), last_retried_by_uuid, current_processor_uuid);
+        session->transfer(flow_file, Failure);
+        return;
+      }
+      // Assuming reuse_mode_ == WARN_ON_REUSE || reuse_mode_ == RESET_REUSE
+      core::logging::LOG_LEVEL reuse_mode_log_level(core::logging::LOG_LEVEL::debug);
```
adamdebreceni, line 121: most vexing parse?

hunyadi-dev: We warn on `[-Wvexing-parse]`

extensions/standard-processors/processors/RetryFlowFile.cpp
hunyadi-dev: Added non-blank validator.

libminifi/include/core/PropertyValidation.h
```diff
@@ -259,6 +259,22 @@ class UnsignedLongValidator : public PropertyValidator {
...
 
+class NonBlankValidator : public PropertyValidator {
+ public:
+  explicit NonBlankValidator(const std::string& name)
+      : PropertyValidator(name) {
+  }
+  ~NonBlankValidator() override = default;
+
+  ValidationResult validate(const std::string& subject, const std::shared_ptr<minifi::state::response::Value>& input) const final {
+    return validate(subject, input->getStringValue());
```
arpadboda, line 270: How would it validate not being blank?
I think we should trim both left and right and check if the result is empty.

hunyadi-dev: I thought yaml parsing ignored padding, so
```yaml
name:  A
```
was the same as 
```yaml
name: A
```

arpadboda:
I still don't see any properties being required, otherwise looks good.

arpadboda:
@hunyadi-dev : could you rebase and push? 
The changes look good to me, but I would like to see it build on Win without errors.

libminifi/include/core/PropertyValidation.h
```diff
@@ -259,6 +259,22 @@ class UnsignedLongValidator : public PropertyValidator {
...
+  }
+
+  ValidationResult validate(const std::string& subject, const std::string& input) const final {
+    return ValidationResult::Builder::createBuilder().withSubject(subject).withInput(input).isValid(utils::StringUtils::trimLeft(input).size()).build();
```
adamdebreceni, line 274: I would prefer and explicit `.size() != 0`

hunyadi-dev: Should not be a factor in readibility.

szaszm:
> Looks good to me, thanks!
> 
> @adamdebreceni @szaszm could you get back to this and check if your comments are properly replied/handled?
I only left a single reply comment, no reviews. I got a reply, so don't feel blocked by me. :)

PR #822, opened 2020-06-25
MINIFICPP-1253 - Do not hang on full cycles

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:


arpadboda:
@adamdebreceni some tests failed in CI:
https://travis-ci.org/github/apache/nifi-minifi-cpp/jobs/701990105

libminifi/include/core/Processor.h
```diff
@@ -305,6 +311,19 @@ class Processor : public Connectable, public ConfigurableComponent, public std::
   Processor &operator=(const Processor &parent);
 
  private:
+  static std::mutex& getGraphMutex() {
+    static std::mutex mutex{};
+    return mutex;
+  }
+
+  // must hold the graphMutex
+  void updateReachability(const std::lock_guard<std::mutex>& graph_lock, bool force = false);
```
arpadboda, line 320: Do we want to hide these functions from actual processor implementations?
I can accept that as a design decision, just interested in the motivation.

libminifi/src/core/Processor.cpp
```diff
@@ -101,14 +102,27 @@ void Processor::setScheduledState(ScheduledState state) {
 }
 
 bool Processor::addConnection(std::shared_ptr<Connectable> conn) {
-  bool ret = false;
+  enum class DidSet{
```
arpadboda, line 105: Nitpicking: maybe "SetAs" would be a better name to represent what's going on here.

libminifi/test/flow-tests/CustomProcessors.h
```diff
@@ -0,0 +1,106 @@
...
+#ifndef NIFI_MINIFI_CPP_CUSTOMPROCESSORS_H
+#define NIFI_MINIFI_CPP_CUSTOMPROCESSORS_H
+
+#include <unordered_map>
+#include <string>
+#include <random>
+#include <YamlConfiguration.h>
+#include "core/Processor.h"
+#include "TestBase.h"
+#include "../../extensions/standard-processors/processors/GenerateFlowFile.h"
```
arpadboda, line 28: I think you could add the standard processors to target include directory in the cmake file to avoid such paths.

libminifi/test/flow-tests/CustomProcessors.h
```diff
@@ -0,0 +1,106 @@
...
+      session->transfer(flowFile, Banana);
+      return;
+    }
+    throw std::runtime_error("Couldn't route file");
+  }
+  void onSchedule(const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSessionFactory> &sessionFactory) override {
+    int apple;
+    assert(context->getProperty(AppleProbability.getName(), apple));
+    int banana;
+    assert(context->getProperty(BananaProbability.getName(), banana));
```
arpadboda, line 78: Without nodebug ifdef I'm not sure these assert actually ever do something.

libminifi/test/flow-tests/FlowBuilder.h
```diff
@@ -0,0 +1,73 @@
...
+
+  std::shared_ptr<minifi::FlowController> controller = std::make_shared<minifi::FlowController>(
+      prov_repo, ff_repo, configuration,
+      std::move(flow),
+      content_repo, DEFAULT_ROOT_GROUP_NAME, true);
+
+  return Flow{std::move(controller), std::move(root)};
+}
+
+double calculateCongestion(const std::shared_ptr<minifi::Connection>& conn) {
```
arpadboda, line 64: Is this referenced anywhere?

libminifi/test/flow-tests/LoopTest.cpp
```diff
@@ -0,0 +1,102 @@
...
+  std::ofstream{yamlPath} << flowConfigurationYaml;
+
+  Flow flow = createFlow(yamlPath);
+
+  auto procGenerator = std::static_pointer_cast<org::apache::nifi::minifi::processors::TestFlowFileGenerator>(flow.root_->findProcessor("Generator"));
+  auto procA = std::static_pointer_cast<org::apache::nifi::minifi::processors::TestProcessor>(flow.root_->findProcessor("A"));
+
+  int tryCount = 0;
+  // wait for the procA to get triggered 15 times
+  while (tryCount++ < 10 && !(procA->trigger_count.load() >= 15)) {
```
arpadboda, line 96: I think in case something goes wrong here, this test will end in an infinite loop instead of failing.

libminifi/include/core/Processor.h
adamdebreceni: if it turns out that there is a legit use-case where we would need to expose this method, I'm more than happy to make it non-private but I couldn't come up with a reason to do so

libminifi/src/core/Processor.cpp
adamdebreceni: interesting because I recall that I first called it "SetAs" but the commit history doesn't confirm this, will make the change

libminifi/test/flow-tests/LoopTest.cpp
adamdebreceni: `tryCount` is our friend 🙂

libminifi/test/flow-tests/FlowBuilder.h
adamdebreceni: removed

libminifi/test/flow-tests/CustomProcessors.h
adamdebreceni: done

libminifi/test/flow-tests/CustomProcessors.h
adamdebreceni: done

libminifi/src/core/Processor.cpp
adamdebreceni: done

libminifi/include/core/Processor.h
arpadboda: Okay, I'm fine with this, thanks!

libminifi/src/core/Processor.cpp
```diff
@@ -289,6 +303,109 @@ bool Processor::isWorkAvailable() {
...
+      }
+      auto dest = std::dynamic_pointer_cast<const Processor>(connection->getDestination());
+      if (!dest) {
+        continue;
+      }
+      if (reachable_processors_[connection].insert(dest).second) {
+        didChange = true;
+      }
+      for (auto& reachedIt : dest->reachable_processors_) {
+        for (auto &reached_proc : reachedIt.second) {
```
arpadboda, line 323: O(scary), but good enough in practice. :D

libminifi/test/flow-tests/LoopTest.cpp
arpadboda: Fair point :)

libminifi/src/core/Processor.cpp
adamdebreceni: I'm all for a O(better) solution 🙂

PR #826, opened 2020-06-29
MINIFICPP-1274 - Commit delete operation before shutdown

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/src/FlowFileRecord.cpp
+++ b/libminifi/src/FlowFileRecord.cpp
@@ -366,7 +366,7 @@ bool FlowFileRecord::DeSerialize(const uint8_t *buffer, const int bufferSize) {
     return false;
   }
 
-  if (nullptr == claim_) {
+  if (nullptr == claim_ && content_full_fath_.length() != 0) {
     claim_ = std::make_shared<ResourceClaim>(content_full_fath_, content_repo_, true);
   }
   return true;
--- a/libminifi/src/FlowFileRecord.cpp
+++ b/libminifi/src/FlowFileRecord.cpp
@@ -366,7 +366,7 @@ bool FlowFileRecord::DeSerialize(const uint8_t *buffer, const int bufferSize) {
     return false;
   }
 
-  if (nullptr == claim_ && content_full_fath_.length() != 0) {
+  if (nullptr == claim_) {
     claim_ = std::make_shared<ResourceClaim>(content_full_fath_, content_repo_, true);
   }
   return true;
--- a/libminifi/test/rocksdb-tests/RepoTests.cpp
+++ b/libminifi/test/rocksdb-tests/RepoTests.cpp
@@ -326,3 +326,82 @@ TEST_CASE("Test FlowFile Restore", "[TestFFR6]") {
 
   LogTestController::getInstance().reset();
 }
+
+TEST_CASE("Flush deleted flowfiles before shutdown", "[TestFFR7]") {
+  using ConnectionMap = std::map<std::string, std::shared_ptr<core::Connectable>>;
+
+  class TestFlowFileRepository: public core::repository::FlowFileRepository{
+   public:
+    explicit TestFlowFileRepository(const std::string& name)
+        : core::SerializableComponent(name),
+          FlowFileRepository(name, FLOWFILE_REPOSITORY_DIRECTORY, MAX_FLOWFILE_REPOSITORY_ENTRY_LIFE_TIME,
+                             MAX_FLOWFILE_REPOSITORY_STORAGE_SIZE, 1) {}
+    void flush() override {
+      FlowFileRepository::flush();
+      if (onFlush_) {
+        onFlush_();
+      }
+    }
+    std::function<void()> onFlush_;
+  };
+
+  TestController testController;
+  char format[] = "/tmp/testRepo.XXXXXX";
+  auto dir = testController.createTempDirectory(format);
+
+  auto config = std::make_shared<minifi::Configure>();
+  config->set(minifi::Configure::nifi_flowfile_repository_directory_default, utils::file::FileUtils::concat_path(dir, "flowfile_repository"));
+
+  auto connection = std::make_shared<minifi::Connection>(nullptr, nullptr, "Connection");
+  ConnectionMap connectionMap{{connection->getUUIDStr(), connection}};
+  // initialize repository
+  {
+    std::shared_ptr<TestFlowFileRepository> ff_repository = std::make_shared<TestFlowFileRepository>("flowFileRepository");
+
+    std::atomic<int> flush_counter{0};
+
+    std::atomic<bool> stop{false};
+    std::thread shutdown{[&] {
+      while (!stop.load()) {}
+      ff_repository->stop();
+    }};
+
+    ff_repository->onFlush_ = [&] {
+      if (++flush_counter != 1) {
+        return;
+      }
+      
+      for (int keyIdx = 0; keyIdx < 100; ++keyIdx) {
+        auto file = std::make_shared<minifi::FlowFileRecord>(ff_repository, nullptr);
+        file->setUuidConnection(connection->getUUIDStr());
+        // Serialize is sync
+        file->Serialize();
+        if (keyIdx % 2 == 0) {
+          // delete every second flowFile
+          ff_repository->Delete(file->getUUIDStr());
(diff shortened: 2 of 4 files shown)


libminifi/test/rocksdb-tests/RepoTests.cpp
```diff
@@ -326,3 +326,82 @@ TEST_CASE("Test FlowFile Restore", "[TestFFR6]") {
...
+          ff_repository->Delete(file->getUUIDStr());
+        }
+      }
+      stop = true;
+      // wait for the shutdown thread to start waiting for the worker thread
+      std::this_thread::sleep_for(std::chrono::milliseconds{100});
+    };
+
+    ff_repository->setConnectionMap(connectionMap);
+    ff_repository->initialize(config);
```
szaszm, line 390: Errors should make the test fail early.
```suggestion
    const bool init_success = ff_repository->initialize(config);
    REQUIRE(init_success);
```

libminifi/test/rocksdb-tests/RepoTests.cpp
```diff
@@ -326,3 +326,82 @@ TEST_CASE("Test FlowFile Restore", "[TestFFR6]") {
...
+      FlowFileRepository::flush();
+      if (onFlush_) {
+        onFlush_();
+      }
+    }
+    std::function<void()> onFlush_;
+  };
+
+  TestController testController;
+  char format[] = "/tmp/testRepo.XXXXXX";
```
szaszm, line 349: Repo tests should place temporary repos on `/var/tmp`, to avoid crashing on systems with `/tmp` mounted as `tmpfs`.

https://issues.apache.org/jira/browse/MINIFICPP-1188

```suggestion
  char format[] = "/var/tmp/testRepo.XXXXXX";
```

libminifi/test/rocksdb-tests/RepoTests.cpp
```diff
@@ -326,3 +326,82 @@ TEST_CASE("Test FlowFile Restore", "[TestFFR6]") {
...
+  {
+    std::shared_ptr<TestFlowFileRepository> ff_repository = std::make_shared<TestFlowFileRepository>("flowFileRepository");
+    ff_repository->setConnectionMap(connectionMap);
+    ff_repository->initialize(config);
+    ff_repository->loadComponent(nullptr);
+    ff_repository->start();
+    std::this_thread::sleep_for(std::chrono::milliseconds{100});
+    REQUIRE(connection->getQueueSize() == 50);
+  }
+}
```
szaszm, line 407: All text files (including source) should end with a newline character. Please add one.

libminifi/src/FlowFileRecord.cpp
```diff
@@ -366,7 +366,7 @@ bool FlowFileRecord::DeSerialize(const uint8_t *buffer, const int bufferSize) {
     return false;
   }
 
-  if (nullptr == claim_) {
```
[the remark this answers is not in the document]
arpadboda: Yep, the goal would be to *always* have claim, even if the content is empty.
The reason behind is that we don't have to deal with failing streams and claims when handling flowfiles. A good example of this is PutFile or PublishKafka, where handling both types of "empty" flowfiles is a pain.

libminifi/test/rocksdb-tests/RepoTests.cpp
arpadboda: More like ```REQUIRE(ff_repository->initialize(config))``` imho :)

libminifi/test/rocksdb-tests/RepoTests.cpp
```diff
@@ -326,3 +326,82 @@ TEST_CASE("Test FlowFile Restore", "[TestFFR6]") {
...
+  ConnectionMap connectionMap{{connection->getUUIDStr(), connection}};
+  // initialize repository
+  {
+    std::shared_ptr<TestFlowFileRepository> ff_repository = std::make_shared<TestFlowFileRepository>("flowFileRepository");
+
+    std::atomic<int> flush_counter{0};
+
+    std::atomic<bool> stop{false};
+    std::thread shutdown{[&] {
+      while (!stop.load()) {}
```
arpadboda, line 365: Maybe it's a bit too late, but it seems like a busy wait for me. 
Which naturally works, but I would avoid that on hour already overloaded CI infra.

libminifi/test/rocksdb-tests/RepoTests.cpp
```diff
@@ -326,3 +326,82 @@ TEST_CASE("Test FlowFile Restore", "[TestFFR6]") {
 
   LogTestController::getInstance().reset();
 }
+
+TEST_CASE("Flush deleted flowfiles before shutdown", "[TestFFR7]") {
+  using ConnectionMap = std::map<std::string, std::shared_ptr<core::Connectable>>;
```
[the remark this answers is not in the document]
adamdebreceni: indeed, it is only used once (previously it was used multiple times) will remove

libminifi/test/rocksdb-tests/RepoTests.cpp
adamdebreceni: interesting will rollback
- do we have a list of platforms we plan to support?
- is the `/var/tmp` the second best place? according to [this](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s15.html) they are preserved between reboots

libminifi/test/rocksdb-tests/RepoTests.cpp
adamdebreceni: will add, but why do we do this?

libminifi/test/rocksdb-tests/RepoTests.cpp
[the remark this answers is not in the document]
adamdebreceni: done

libminifi/src/FlowFileRecord.cpp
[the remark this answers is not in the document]
adamdebreceni: done

libminifi/test/rocksdb-tests/RepoTests.cpp
adamdebreceni: done

agree with the fail fast philosophy, what I don't agree is `Repository::initialize` returning `false` on failure, I don't see how is the user expected to handle it, of the two places it is called, one doesn't check the return value and one terminates the application

do we have a list of changes we were holding back on, until a new major release?

libminifi/test/rocksdb-tests/RepoTests.cpp
adamdebreceni: done

libminifi/test/rocksdb-tests/RepoTests.cpp
adamdebreceni: done

libminifi/test/rocksdb-tests/RepoTests.cpp
adamdebreceni: done

libminifi/test/rocksdb-tests/RepoTests.cpp
szaszm: I agree that returning a bool to signal error is not the best idea. I would prefer the use of exceptions for all errors that are not usually part of the program flow (i.e. exceptional).

I'm not aware of a list of changes we are "holding back", but creating a Jira issue with Fix version = "1.0" could be one way of maintaining such a list, because we can search for those later.

libminifi/test/rocksdb-tests/RepoTests.cpp
szaszm: 1. I don't think we have a clear list. I think we just aim for a large GNU/Linux coverage.
2. Not sure if it's the second best place. Normally we clean up temporary directories, but this doesn't happen when tests crash, so there may be leftovers in some rare situations. In a discussion around the time of fixing the referred bug, we considered disabling direct IO, but that would mean we don't test the same behavior that we are running, so going for a different directory seemed to be the better approach.

libminifi/test/rocksdb-tests/RepoTests.cpp
szaszm: https://stackoverflow.com/a/729795/3997716

PR #827, opened 2020-06-30
MINIFICPP-1273 - Drain connections on flow shutdown

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/src/core/ProcessGroup.cpp
```diff
@@ -403,6 +403,16 @@ void ProcessGroup::removeConnection(std::shared_ptr<Connection> connection) {
...
 }
 
+void ProcessGroup::drainConnections() {
+  for (auto &&connection : connections_) {
+    connection->drain();
+  }
+
+  for (std::set<ProcessGroup *>::iterator it = child_process_groups_.begin(); it != child_process_groups_.end(); ++it) {
+    (*it)->drainConnections();
+  }
```
szaszm, line 413: This could become a range-based for loop.

libminifi/test/flow-tests/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+FOREACH(testfile ${FLOW_TESTS})
+    get_filename_component(testfilename "${testfile}" NAME_WE)
+    add_executable("${testfilename}" "${testfile}")
+    createTests("${testfilename}")
+    target_link_libraries(${testfilename} ${CATCH_MAIN_LIB})
+    target_wholearchive_library(${testfilename} minifi-standard-processors)
+    MATH(EXPR FLOW_TEST_COUNT "${FLOW_TEST_COUNT}+1")
+    add_test(NAME "${testfilename}" COMMAND "${testfilename}" WORKING_DIRECTORY ${TEST_DIR})
+ENDFOREACH()
+message("-- Finished building ${FLOW_TEST_COUNT} flow related test file(s)...")
```
szaszm, line 31: The file should end with a newline character

adamdebreceni: done

libminifi/src/core/ProcessGroup.cpp
adamdebreceni: done

libminifi/test/flow-tests/CMakeLists.txt
adamdebreceni: (it seems like the Enter puts too much stress on my pinky)

libminifi/test/flow-tests/FlowControllerTests.cpp
```diff
@@ -0,0 +1,125 @@
...
+#include "core/repository/AtomicRepoEntries.h"
+#include "core/RepositoryFactory.h"
+#include "FlowFileRecord.h"
+#include "provenance/Provenance.h"
+#include "properties/Configure.h"
+#include "../unit/ProvenanceTestHelper.h"
+#include "../TestBase.h"
+#include "YamlConfiguration.h"
+
+const char* yamlConfig =
```
arpadboda, line 36: We have a test/resources folder, where you could place it as a real yaml file and read it. 

The location of the yaml file can be a cmd line argument as well as we do in our integration tests.

libminifi/src/core/ProcessGroup.cpp
```diff
@@ -92,13 +92,12 @@ ProcessGroup::~ProcessGroup() {
     onScheduleTimer_->stop();
   }
 
-  for (auto &&connection : connections_) {
+  for (auto&& connection : connections_) {
```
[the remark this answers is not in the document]
szaszm: That's a forwarding reference, because `auto` uses template deduction rules. In other words, "I don't care what it is, bind a reference to it" reference.
In this case it will be a const lvalue reference, because:
1. `auto&&`
2. `const std::shared_ptr<Connection>& &&` after deducing `auto` (note: `connections_` is a `std::set`, which only has const iterators)
3. `const std::shared_ptr<Connection>&` after reference-collapsing

Normally I point out that `const auto&` or `auto&` would be more explicit and readable, but since this is old code with just a space change, I didn't want to bother @adamdebreceni with this.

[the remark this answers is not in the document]
adamdebreceni: I am torn between always use `auto&&` in range-based for loops and never, it is really only useful when the iteration yields a temporary proxy object like with `std::vector<bool>`, `auto&` could be used, I wouldn't use `const auto&` as it makes me think that the connection is const whereas only the `std::shared_ptr<...>` is

libminifi/test/flow-tests/FlowControllerTests.cpp
adamdebreceni: I saw that a number of tests store their config files there, but the fact that I have to search for the test by name then check the CMakeLists.txt what the command line arguments are going to be, and then manually navigate to the file, makes me think that it is not the best approach

libminifi/test/flow-tests/CMakeLists.txt
adamdebreceni: (some more bikeshedding questions about the POSIX "line")

Screenshot 2020-07-06 at 8 32 31

what does POSIX say about where the cursor would be? it cannot be on "line 2" as there is no "line 2" 

or take the error message:

Screenshot 2020-07-06 at 8 33 49

what is the `2` in `test.cpp:**2**:8`?

szaszm: The tools were written to behave decently with non-text files and consider the line number to be the encountered `\n`s + 1. Historical posix text processing utilities may ignore the contents after the last `\n`, but I haven't seen modern tools do so. [1]

Please note that I'm by no means a POSIX lawyer, and the part about handling is mostly speculation.

[1] https://unix.stackexchange.com/a/18789/366131

adamdebreceni: `text-file` is a file format that has the magic number `\n` at the end of the file 🤯

szaszm: I'm not sure about the above implication (`\n` before EOF -> is text file), but its contrary (no `\n` before EOF -> is not a text file) seems to be true to me.

adamdebreceni: this is a response to a deleted comment about how `text file` is a file format with the magic number `\n` at the end of the file (I don't know how to resurrect the comment)

PR #833, opened 2020-07-06
MINIFICPP-1272 - Graceful shutdown on flow update

adamdebreceni:
Will conflict with #822 as they both use the newly introduced test processors.

----

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/src/FlowController.cpp
```diff
@@ -239,8 +239,35 @@ int16_t FlowController::stop(bool force, uint64_t timeToWait) {
...
+        std::string shutdown_timeout_str;
+        if (configuration_->get(minifi::Configure::nifi_flowcontroller_drain_timeout, shutdown_timeout_str)) {
+          shutdown_timeout = core::TimePeriodValue::fromString(shutdown_timeout_str);
+        }
+        return std::chrono::milliseconds{shutdown_timeout.value_or(default_timeout).getMilliseconds()};
+      };
+      std::size_t count;
+      while (shutdown_timer < shutdown_timeout() && (count = this->root_->getTotalFlowFileCount()) != 0) {
+        std::this_thread::sleep_for(shutdown_check_interval_);
+        shutdown_timer += shutdown_check_interval_;
```
arpadboda, line 266: I would do it by saving the current time when this activity is started and check if the time elapsed exceeds the maximum allowed for this operation.

libminifi/test/flow-tests/FlowControllerTests.cpp
```diff
@@ -119,7 +107,125 @@ TEST_CASE("Flow shutdown drains connections", "[TestFlow1]") {
...
+    // sleep only on the first trigger
+    if (first_onTrigger.compare_exchange_strong(isFirst, false)) {
+      std::this_thread::sleep_for(std::chrono::milliseconds{1500});
+    }
+  };
+
+  testController.startFlow();
+
+  // wait for the source processor to enqueue its flowFiles
+  std::this_thread::sleep_for(std::chrono::milliseconds{50});
```
arpadboda, line 172: Can't we simply wait here while the source processor's trigger count becomes one?

libminifi/src/core/ProcessGroup.cpp
```diff
@@ -412,6 +414,18 @@ void ProcessGroup::drainConnections() {
   }
 }
 
+std::size_t ProcessGroup::getTotalFlowFileCount() const {
+  std::size_t sum = 0;
+  for (auto& conn : connections_) {
```
arpadboda, line 419: Nitpicking: const auto&

libminifi/src/core/ProcessGroup.cpp
```diff
@@ -412,6 +414,18 @@ void ProcessGroup::drainConnections() {
...
+    sum += conn->getQueueSize();
+  }
+
+  for (ProcessGroup* childGroup : child_process_groups_) {
```
arpadboda, line 423: const ptr

libminifi/src/FlowController.cpp
```diff
@@ -239,8 +239,35 @@ int16_t FlowController::stop(bool force, uint64_t timeToWait) {
...
+      });
+      std::chrono::milliseconds shutdown_timer{0};
+      // we enable C2 to progressively increase the timeout
+      // in case it sees that waiting for a little longer could
+      // allow the FlowFiles to be processed
+      auto shutdown_timeout = [&]() -> std::chrono::milliseconds {
+        if (timeToWait != 0) {
+          return std::chrono::milliseconds{timeToWait};
+        }
+        static const core::TimePeriodValue default_timeout{"10 sec"};
```
arpadboda, line 255: I think this shouldn't be defined in the middle of the code, very difficult to search for.
Moreover it can delay shutdown, so I would prefer to leave it 0 by default.

libminifi/src/core/ProcessGroup.cpp
adamdebreceni: that would make me think, that the `Connection` is `const` but only the `std::shared_ptr<...>` would be
if it is clear for everybody else I can make the change

libminifi/src/FlowController.cpp
adamdebreceni: done

libminifi/src/core/ProcessGroup.cpp
adamdebreceni: done

libminifi/test/flow-tests/FlowControllerTests.cpp
adamdebreceni: we are not waiting for the trigger but for the FlowFiles to be enqueued into the Connection, during shutdown we cannot check if a source processor stopped, thus we can only check the connections' contents, should we expose the fact that there are active sessions for a processor?

arpadboda: Just checked the code: there are functions to increment and decrement active tasks of a processor. The decrement happens after the session is commited, which a piece of good news for us: 
```
bool Processor::isRunning() {
  return (state_ == RUNNING && active_tasks_ > 0);
}
```

This should properly be able to tell us whether the execution (including committing and destroying the session) already finished or not.

libminifi/src/FlowController.cpp
adamdebreceni: done

libminifi/test/flow-tests/FlowControllerTests.cpp
adamdebreceni: `active_tasks_` is a strange animal, should be revisited, but waiting for the FlowFiles to be enqueued into the connections through the `ProcessGroup::getTotalFlowFileCount` method is a viable solution

arpadboda: Sounds good to me, thanks!

libminifi/src/core/ProcessGroup.cpp
arpadboda: Fair point, not sure which option is better, so feel free to choose. You can also mark this thread resolved.

libminifi/src/FlowController.cpp
```diff
@@ -239,8 +239,36 @@ int16_t FlowController::stop(bool force, uint64_t timeToWait) {
...
+        std::string shutdown_timeout_str;
+        if (configuration_->get(minifi::Configure::nifi_flowcontroller_drain_timeout, shutdown_timeout_str)) {
+          shutdown_timeout = core::TimePeriodValue::fromString(shutdown_timeout_str);
+        }
+        if (shutdown_timeout) {
+          return std::chrono::milliseconds{shutdown_timeout->getMilliseconds()};
+        }
+        return std::chrono::milliseconds{0};
+      };
+      std::size_t count;
```
hunyadi-dev, line 265: This is never read.

libminifi/src/FlowController.cpp
```diff
@@ -239,8 +239,36 @@ int16_t FlowController::stop(bool force, uint64_t timeToWait) {
...
+      // we enable C2 to progressively increase the timeout
+      // in case it sees that waiting for a little longer could
+      // allow the FlowFiles to be processed
+      auto shutdown_timeout = [&]() -> std::chrono::milliseconds {
+        if (timeToWait != 0) {
+          return std::chrono::milliseconds{timeToWait};
+        }
+        utils::optional<core::TimePeriodValue> shutdown_timeout;
+        std::string shutdown_timeout_str;
+        if (configuration_->get(minifi::Configure::nifi_flowcontroller_drain_timeout, shutdown_timeout_str)) {
```
hunyadi-dev, line 257: Where do we expect to re-read the configuration during shutdown?

PR #836, opened 2020-07-09
MINIFICPP-1248 Create unit tests for the ConsumeWindowsEventLog processor

fgerlits:
Unit tests for the ConsumeWindowsEventLog processor and the related helper classes Bookmark and MetadataWalker.

Small changes in the classes under test to help unit testing:
* move the unique_evt_handle class from Bookmark.h to its own header, so we can reuse it,
* refactor MetadataWalker to make the Windows event dependencies mockable.

Coverage is far from perfect.  The two main gaps are:
* the processor is not tested in `ProcessOldEvents=true` mode, as this would read the event log from the start and there is no way (that I know of) to clear the event log without Administrator rights,
* the Plaintext output format is not tested, only the XML format, as (I think) one needs to publish manifest-based events for `EvtFormatMessage(... flags=EvtFormatMessageEvent ...)` to work, and registering a manifest requires Administrator rights, too.

I have tested these use cases manually, but clearly a unit test would be better.  If you have an idea how to do that, please let me know.

---

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/windows-event-log/tests/BookmarkTests.cpp
```diff
@@ -0,0 +1,245 @@
...
+  }
+  return results;
+}
+
+unique_evt_handle getFirstEventFromResults(const unique_evt_handle& results) {
+  REQUIRE(results);
+  EVT_HANDLE event_raw_handle = 0;
+  DWORD timeout_ms = 100;
+  DWORD num_results_found = 0;
+  bool result = EvtNext(results.get(), 1, &event_raw_handle, timeout_ms, 0, &num_results_found);
```
arpadboda, line 83: Nitpicking: this variable seems to be needless.

extensions/windows-event-log/tests/BookmarkTests.cpp
```diff
@@ -0,0 +1,245 @@
...
+  if (!results) {
+    FAIL("EvtQuery() failed; error code: " << GetLastError());
+  }
+  return results;
+}
+
+unique_evt_handle getFirstEventFromResults(const unique_evt_handle& results) {
+  REQUIRE(results);
+  EVT_HANDLE event_raw_handle = 0;
+  DWORD timeout_ms = 100;
```
arpadboda, line 81: Could you define this at the beginning of the file? 
I prefer to have constants there with capital case.

extensions/windows-event-log/tests/ConsumeWindowsEventLogTests.cpp
```diff
@@ -0,0 +1,393 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "ConsumeWindowsEventLog.h"
+
+#include "core/ConfigurableComponent.h"
+#include "../../../../extensions/standard-processors/processors/LogAttribute.h"
```
arpadboda, line 21: You can avoid this by adding standard processors to the cmake file of the test as "target_include"dir"

extensions/windows-event-log/tests/ConsumeWindowsEventLogTests.cpp
```diff
@@ -0,0 +1,393 @@
...
+
+TEST_CASE("ConsumeWindowsEventLog constructor works", "[create]") {
+  TestController test_controller;
+  std::shared_ptr<TestPlan> test_plan = test_controller.createPlan();
+
+  REQUIRE_NOTHROW(ConsumeWindowsEventLog processor_one("one"));
+
+  REQUIRE_NOTHROW(
+    utils::Identifier uuid = utils::IdGenerator::getIdGenerator()->generate();
+    ConsumeWindowsEventLog processor_two("two", uuid);
```
arpadboda, line 53: What's the point in "manually" generating the uuid here?
As far as I remember, it's done by default in case it's not specified.

extensions/windows-event-log/tests/ConsumeWindowsEventLogTests.cpp
```diff
@@ -0,0 +1,393 @@
...
+    // ConsumeWindowsEventLog::RenderFormatXML,  // FIXME(fgerlits): not defined, does not exist in NiFi either; should be removed
+    ConsumeWindowsEventLog::MaxBufferSize,
+    // ConsumeWindowsEventLog::InactiveDurationToReconnect,  // FIXME(fgerlits): obsolete, see definition; should be removed
+    ConsumeWindowsEventLog::IdentifierMatcher,
+    ConsumeWindowsEventLog::IdentifierFunction,
+    ConsumeWindowsEventLog::ResolveAsAttributes,
+    ConsumeWindowsEventLog::EventHeader,
+    ConsumeWindowsEventLog::OutputFormat,
+    ConsumeWindowsEventLog::BatchCommitSize,
+    ConsumeWindowsEventLog::BookmarkRootDirectory,
```
arpadboda, line 82: This one is obsolete as well. 
This was introduced back in those days when we had no central state storage for processor, but the current implementation of CWEL doesn't rely on this any more as it uses the corresponding controller service to store state.

extensions/windows-event-log/tests/ConsumeWindowsEventLogTests.cpp
```diff
@@ -0,0 +1,393 @@
...
+}  // namespace
+
+TEST_CASE("ConsumeWindowsEventLog output format can be set", "[create][output_format]") {
+  outputFormatSetterTestHelper("XML", 1);
+  outputFormatSetterTestHelper("Plaintext", 1);
+  outputFormatSetterTestHelper("Both", 2);
+
+  // NOTE(fgerlits): this may be a bug, as I would expect this to throw in onSchedule(),
+  // but it starts merrily, just does not write flow files in either format
+  outputFormatSetterTestHelper("InvalidValue", 0);
```
arpadboda, line 291: I think @adebreceni 's latest change with property validation should address this issue. 
Could you guys sync?

extensions/windows-event-log/tests/ConsumeWindowsEventLogTests.cpp
```diff
@@ -0,0 +1,393 @@
...
+    // the ID of the event goes here (a number)
+    REQUIRE(LogTestController::getInstance().contains(R"(</EventRecordID><Channel>Application</Channel><Computer>)"));
+    // the computer name goes here
+    REQUIRE(LogTestController::getInstance().contains(R"(</Computer><Security/></System><EventData><Data>Event one</Data></EventData></Event>)"));
+  }
+}
+
+namespace {
+
+int countOccurrences(const std::string& text, const std::string& substring) {
```
arpadboda, line 340: There is a function called ```countPatInStr``` in IntegrationBase.h doing quite the same.

Although I feel like this logic should be moved to LogTestController, so both unit and integration tests could depend on it.

extensions/windows-event-log/tests/ConsumeWindowsEventLogTests.cpp
```diff
@@ -0,0 +1,393 @@
...
+  outputFormatSetterTestHelper("XML", 1);
+  outputFormatSetterTestHelper("Plaintext", 1);
+  outputFormatSetterTestHelper("Both", 2);
+
+  // NOTE(fgerlits): this may be a bug, as I would expect this to throw in onSchedule(),
+  // but it starts merrily, just does not write flow files in either format
+  outputFormatSetterTestHelper("InvalidValue", 0);
+}
+
+// NOTE(fgerlits): I don't know how to unit test this, as my manually published events all result in an empty string if OutputFormat is Plaintext
```
arpadboda, line 294: Hm, this is weird as we seem to successfully capture event data as we can render it in xml format.
I would definitely be interested in the root cause of this.

extensions/windows-event-log/tests/BookmarkTests.cpp
fgerlits: we check it in the next line; would you prefer to inline the `EvtNext` call into `if (result)`?  (or maybe rename `result` to `success`?)

extensions/windows-event-log/tests/ConsumeWindowsEventLogTests.cpp
fgerlits: It is slightly different, because the default value of `uuid` is `utils::Identifier()`, which is blank/all-0s.  I wanted to exercise both code paths, but it's probably not really necessary.  Do you want me to remove it?

extensions/windows-event-log/tests/ConsumeWindowsEventLogTests.cpp
fgerlits: Proper manifest-based event types have a dll containing string mappings used by `EvtFormatMessage(... flags=EvtFormatMessageEvent ...)`, so XML like
```xml
<EventData>
  <Binary>(some hex-encoded binary data)</Binary>
  <Data>Cloudera Digital Delivery Services</Data>
  <Data>4.0.52.0</Data>
  <Data>1033</Data>
  <Data>Cloudera Inc</Data>
  <Data>0</Data>
</EventData>
```
can get transformed to plain text like
```
Windows Installer reconfigured the product.
Product Name: Cloudera Digital Delivery Services.
Product Version: 4.0.52.0.
Product Language: 1033.
Manufacturer: Cloudera Inc.
Reconfiguration success or error status: 0.
```
I think the reason I get an empty result is that `EvtFormatMessage` is not able to find these mappings for my ad-hoc events.  I will do some more investigation; maybe it is possible to use some predefined event type which has predefined mappings.

extensions/windows-event-log/tests/ConsumeWindowsEventLogTests.cpp
arpadboda: Nope, I was just interested in the added value. This covers another path, it's fine, thanks!

extensions/windows-event-log/tests/BookmarkTests.cpp
arpadboda: I just meant embedding in the if statement, but I can live with the current code as well, not a big deal.

extensions/windows-event-log/tests/ConsumeWindowsEventLogTests.cpp
arpadboda: That makes sense, thanks!

extensions/windows-event-log/tests/ConsumeWindowsEventLogTests.cpp
fgerlits: He has told me his change was for numeric (and I think bool) properties only; `withAllowableValues()` does not throw on a non-allowable value, it only includes the list of allowable values in the heartbeat so clients can do valuation on their side.  Shall I create a Jira to add validation?

extensions/windows-event-log/tests/BookmarkTests.cpp
fgerlits: done

extensions/windows-event-log/tests/BookmarkTests.cpp
fgerlits: done

extensions/windows-event-log/tests/ConsumeWindowsEventLogTests.cpp
fgerlits: done

extensions/windows-event-log/tests/ConsumeWindowsEventLogTests.cpp
fgerlits: I've added a comment about deprecation

extensions/windows-event-log/tests/ConsumeWindowsEventLogTests.cpp
fgerlits: done

extensions/windows-event-log/tests/ConsumeWindowsEventLogTests.cpp
arpadboda: Yep, a follow-up would make sense to me. Not sure we can introduce this without a breaking change, so might be targeted for 1.0, but definitely seems to be a useful improvement to me.

fgerlits: https://issues.apache.org/jira/browse/MINIFICPP-1292

arpadboda:
@fgerlits : 
The stringutils part seems to fail on gcc:
https://travis-ci.org/github/apache/nifi-minifi-cpp/jobs/708347373

PR #840, opened 2020-07-15
MINIFICPP-1293 Fix the failing PropertyTests

fgerlits:
See https://issues.apache.org/jira/browse/MINIFICPP-1293 for the description of the bug.

Description of the change: instead of using `mktime` and adjusting by the time zone offset, use `_mkgmtime` on Windows where it is available, and use a hand-written version of `mkgmtime` elsewhere.

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/utils/TimeUtil.h
+++ b/libminifi/include/utils/TimeUtil.h
@@ -73,7 +73,7 @@ inline time_t mkgmtime(struct tm *date_time) {
   static const int month_lengths[] =      {31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31};
   static const int month_lengths_leap[] = {31, 29, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31};
   static const auto is_leap_year = [](int year) -> bool {
-    return (year - 1968) % 4 == 0 && ((year - 1900) % 100 != 0 || (year - 1600) % 400 == 0);
+    return year % 4 == 0 && (year % 100 != 0 || year % 400 == 0);
   };
   static const auto num_leap_days = [](int year) -> int {
     return (year - 1968) / 4 - (year - 1900) / 100 + (year - 1600) / 400;
--- a/libminifi/include/utils/TimeUtil.h
+++ b/libminifi/include/utils/TimeUtil.h
@@ -66,6 +66,42 @@ inline std::string getTimeStr(uint64_t msec, bool enforce_locale = false) {
   return ret;
 }
 
+inline time_t mkgmtime(struct tm *date_time) {
+#ifdef WIN32
+  return _mkgmtime(date_time);
+#else
+  static const int month_lengths[] =      {31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31};
+  static const int month_lengths_leap[] = {31, 29, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31};
+  static const auto is_leap_year = [](int year) -> bool {
+    return (year - 1968) % 4 == 0 && ((year - 1900) % 100 != 0 || (year - 1600) % 400 == 0);
+  };
+  static const auto num_leap_days = [](int year) -> int {
+    return (year - 1968) / 4 - (year - 1900) / 100 + (year - 1600) / 400;
+  };
+
+  int year = date_time->tm_year + 1900;
+  time_t result = year - 1970;
+  result *= 365;
+  result += num_leap_days(year - 1);
+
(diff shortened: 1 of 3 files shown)


libminifi/test/unit/TimeUtilTests.cpp
```diff
@@ -0,0 +1,87 @@
...
+#include "../TestBase.h"
+
+namespace {
+  constexpr int ONE_HOUR = 60 * 60;
+  constexpr int ONE_DAY = 24 * ONE_HOUR;
+
+  struct tm createTm(int year, int month, int day, int hour, int minute, int second, bool is_dst = false) {
+    struct tm date_time;
+    date_time.tm_year = year - 1900;
+    date_time.tm_mon = month - 1;
```
arpadboda, line 28: Do I understand correctly that months are stored in [0-11] range, but days are stored in [1-31]?

fgerlits: Yes, and years start at 1900 -- the person who designed `struct tm` was probably drunk at the time.

arpadboda: 🍺

libminifi/include/utils/TimeUtil.h
```diff
@@ -108,23 +144,9 @@ inline int64_t parseDateTimeStr(const std::string &str) {
...
-  }
-  int64_t timezone_offset = utc - local;
-
-  /* Convert parsed date */
-  time_t time = mktime(&timeinfo);
-  if (time == -1) {
-    return -1;
-  }
-  return time + timezone_offset;
+  return static_cast<int64_t>(mkgmtime(&timeinfo));
```
szaszm, line 149: Instead of providing `mkgmtime` on posix, I would opt for converting `parseDateTimeStr` to use `strptime` on posix and use the current implementation with a direct call to `_mkgmtime` on windows. This way we can avoid maintaining date/time code, [which is a pain](https://www.youtube.com/watch?v=-5wpm-gesOY).

fgerlits: `strptime` is orthogonal to this change, as `strptime` converts `string`  -> `struct tm`, and the problem is with the `struct tm` -> UTC `time_t` conversion.

Date/time functions are a pain because of time zones, DST, and leap seconds; `mkgmtime` (mercifully) does not need to deal with any of these.  The formula set by Pope Gregory in 1582 is unlikely to change anytime soon.

fgerlits: In C++20, there will (finally) be some usable date-time parsing functions; something like
```c++
stringstream stream{"2020-08-01T01:00:00Z"};
utc_time<seconds> date_time;
stream >> parse("%Y-%m-%dT%H:%M:%SZ", date_time);
return date_time.time_since_epoch().count();
```
but I haven't been able to find a compiler which can compile this, yet.

szaszm: 1. Sorry, I got too carried away with the UTC/localtime issue that I didn't even look at the signature of `strptime`.
2. I believe `mkgmtime` / [`timegm`](https://linux.die.net/man/3/timegm) ([code](https://code.woboq.org/userspace/glibc/time/timegm.c.html), don't use this) have to deal with some of these issues, too, e.g. leap seconds occur in UTC.
3. You might find [`parse`](https://howardhinnant.github.io/date/date.html#parse) interesting from [HowardHinnant/date](https://github.com/HowardHinnant/date), the basis of C++20 date extensions.

fgerlits: I didn't know about `timegm`; we could use that on Linux, but we'd still need the rest of the code for non-Linux non-Windows compiles (I guess that's why you said don't use it).

Leap seconds occur in UTC, but not in Unix time.  This means that arguably `parseDateTimeStr("2016-12-31T23:59:60Z")` should return 1483228799 (ie. the same as `2016-12-31T23:59:59Z`) instead of 1483228800 (ie. the same as `2017-01-01T00:00:00Z`).  But mktime doesn't do this, either, so the old `parseDateTimeStr` before this pull request also returned 1483228799.

I would not bring in a new library just to fix a bug that only affects dates on 1 January 1970.  If you hate this change too much, my preferred second best solution would be to simply leave parseDateTimeStr as it is, and document this limitation ("doesn't work for some times on 1 January 1970 in some time zones on some architectures") in the unit test.

I understand your point, and agree that MiNiFi code should not contain an implementation of the Gregorian calendar, as it is not its responsibility, but I think having it temporarily until we can upgrade to C++20 is not terrible.

PR #837, opened 2020-07-14
MINIFICPP-1121 - Upgrade spdlog to version 1.8.0

hunyadi-dev:
This is a draft PR used for running CI jobs on different platforms.

Thank you for submitting a contribution to Apache NiFi - MiNiFi C++.

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [ ] Is there a JIRA ticket associated with this PR? Is it referenced
     in the commit message?

- [ ] Does your PR title start with MINIFICPP-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.

- [ ] Has your PR been rebased against the latest commit within the target branch (typically master)?

- [ ] Is your initial contribution a single, squashed commit?

### For code changes:
- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?
- [ ] If applicable, have you updated the LICENSE file?
- [ ] If applicable, have you updated the NOTICE file?

### For documentation related changes:
- [ ] Have you ensured that format looks appropriate for the output in which it is rendered?

### Note:
Please ensure that once the PR is submitted, you check travis-ci for build issues and submit an update to your PR as soon as possible.

PR #848, opened 2020-07-23
MINIFICPP-1268 Fix compiler warnings on Windows (VS2017)

fgerlits:
Change the warning level from W1 (severe warnings only) to W3 (production quality warnings).

The largest batch of changes is due to `uint64_t` -> `size_t` conversions, which are narrowing on Windows because we build a 32-bit binary.

Almost all changes should have no effect on behavior, except when a narrowing would occur.  The two exceptions are:
* in `C2Agent::configure()` (C2Agent.cpp:256) the value of the property used to be read and discarded; now it is assigned to a field (the correct one, I hope);
* in `WriteCallback::process()` (SiteToSiteClient.h:308) we calculated the number of bytes received, but then we did not log it but always logged 0 instead.

I did not deal with warnings related to the min/max macros, as those will be fixed by #837 

---

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/civetweb/tests/ListenHTTPTests.cpp
```diff
@@ -169,7 +168,6 @@ class ListenHTTPTestsFixture {
   }
 
  protected:
-  char* tmp_dir_format;
```
adamdebreceni, line 172: 👍

extensions/libarchive/BinFiles.h
```diff
@@ -91,7 +92,7 @@ class Bin {
       if (flow->getAttribute(fileCount_, value)) {
         try {
           // for defrag case using the identification
-          int count = std::stoi(value);
+          size_t count = std::stoul(value);
```
adamdebreceni, line 95: could use 
```
uint32_t count;
utils::internal::ValueParser(value).parse(count).parseEnd()
```

adamdebreceni: no `size_t` overload unfortunately

libminifi/include/io/Serializable.h
```diff
@@ -44,8 +44,9 @@ Integral byteSwap(Integral i) {
 #ifdef htonll
   return htonll(i);
 #else
-  #define htonll_r(x) ((((uint64_t)htonl(x)) << 32) + htonl((x) >> 32))
-  return htonll_r(i);
+  uint32_t lower_32_bits = static_cast<uint32_t>(i);
```
adamdebreceni, line 47: are these two implementations equivalent? on big-endian systems `htonll` is noop while the other one swaps the bytes anyway

adamdebreceni: similarly, the other `byteSwap` overloads all seem to be noop on big-endian systems

fgerlits: The new code is equivalent to the old code which used the `#define`, but you are right that it's not always equivalent to `htonll`.  We check `is_little_endian` every time before we call `byteSwap`; it might be better to move that check inside `byteSwap`, but I didn't think that was in scope for this change.

extensions/libarchive/BinFiles.h
fgerlits: I wanted to keep changes minimal, just making sure warnings, especially warnings about implicit narrowing, are gone.

libminifi/include/io/Serializable.h
adamdebreceni: agree that it is not in scope
👍

libminifi/include/io/CRCStream.h
```diff
@@ -224,7 +222,7 @@ template<typename T>
 int CRCStream<T>::readData(uint8_t *buf, int buflen) {
   int ret = child_stream_->read(buf, buflen);
   if (ret > 0) {
-    crc_ = crc32(crc_, buf, ret);
+    crc_ = crc32(gsl::narrow<uLong>(crc_), buf, ret);
```
szaszm, line 225: I suggest changing the type of `crc_` to `uint32_t` and `static_assert` that `uLong` has the same width and signedness as `uint32_t`. This should make `gsl::narrow` calls redundant.

nanofi/src/core/cstream.c
```diff
@@ -153,9 +148,9 @@ int readUTFLen(uint32_t * utflen, cstream * stream) {
   return ret;
 }
 
-int readUTF(char * buf, uint64_t buflen, cstream * stream) {
+int readUTF(char * buf, uint32_t buflen, cstream * stream) {
```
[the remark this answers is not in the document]
fgerlits: Because everywhere we call this function, we call it with an argument of type `uint32_t`, and the return value (output argument) of `readUTFLen()` is also of type `uint32_t`.

libminifi/include/io/CRCStream.h
fgerlits: `uLong` in zlib is a typedef to `unsigned long`, so on 64-bit Linux it is 64 bits long.  I have changed the type of `crc_` to `uLong`, so now we only have one `gsl::narrow` (called once per instance) instead of three (called many times).

libminifi/src/sitetosite/SiteToSiteClient.cpp
```diff
@@ -499,8 +499,9 @@ int16_t SiteToSiteClient::send(std::string transactionID, DataPacket *packet, co
       return -1;
     }
 
-    ret = transaction->getStream().writeData(reinterpret_cast<uint8_t *>(const_cast<char*>(packet->payload_.c_str())), len);
-    if (ret != (int64_t)len) {
+    ret = transaction->getStream().writeData(reinterpret_cast<uint8_t *>(const_cast<char*>(packet->payload_.c_str())),
+                                             gsl::narrow<int>(len));
+    if (ret != static_cast<int64_t>(len)) {
```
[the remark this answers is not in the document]
fgerlits: fixed

extensions/http-curl/processors/InvokeHTTP.cpp
```diff
@@ -337,7 +337,7 @@ void InvokeHTTP::onTrigger(const std::shared_ptr<core::ProcessContext> &context,
     const std::vector<char> &response_body = client.getResponseBody();
     const std::vector<std::string> &response_headers = client.getHeaders();
 
-    int64_t http_code = client.getResponseCode();
+    int http_code = gsl::narrow<int>(client.getResponseCode());
```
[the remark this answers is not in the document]
fgerlits: I have removed this `gsl::narrow()` call.

libminifi/include/io/CRCStream.h
szaszm: Oh, I assumed it must be 32 bit, because it's crc32. Great, thanks.

PR #854, opened 2020-07-28
MINIFICPP-1310 Fix the CMakeSettings.json file used by Visual Studio

fgerlits:
* Change build type to 32-bit RelWithDebInfo
* Copy all the variable settings from win_build_vs.bat
* Set up the search path for mc.exe correctly (mixing individual items and lists doesn't work)
* Update Windows.md

---

### For all changes:
     in the commit message?

### For code changes:

### Note:


szaszm:
Why change to 32 bit? Why RelWithDebInfo?

fgerlits:
> Why change to 32 bit? Why RelWithDebInfo?

I wanted to be as close as possible to `win_build_vs.bat`, which builds a 32-bit Release version by default, but also include debug symbols.

CMakeSettings.json
```diff
@@ -2,97 +2,97 @@
...
         },
         {
-          "name": "EXCLUDE_BOOST",
-          "value": "TRUE"
+          "name": "SKIP_TESTS",
+          "value": "OFF"
+        },
+        {
+          "name": "STRICT_GSL_CHECKS",
+          "value": "AUDIT"
```
szaszm, line 92: I suggest `ON` by default instead of `AUDIT`. Audit checks shouldn't be enabled anywhere other than the CI by default, IMO.

https://github.com/gsl-lite/gsl-lite/#contract-checking-configuration-macros

CMakeSettings.json
```diff
@@ -2,97 +2,97 @@
...
+      "inheritEnvironments": [ "msvc_x64_x86" ],
       "buildRoot": "${env.USERPROFILE}\\CMakeBuilds\\${workspaceHash}\\build\\${name}",
       "installRoot": "${env.USERPROFILE}\\CMakeBuilds\\${workspaceHash}\\install\\${name}",
       "cmakeCommandArgs": "",
-      "buildCommandArgs": "-m -v:minimal",
+      "buildCommandArgs": "",
       "variables": [
         {
-          "name": "OPENSSL_OFF",
+          "name": "INSTALLER_MERGE_MODULES",
```
szaszm, line 15: I suggest removing entries that are already cmake defaults so that cmake default changes implicitly apply here as well. I think only `CMAKE_BUILD_TYPE_INIT`, `WIN32`, `USE_SHARED_LIBS`, `DISABLE_CONTROLLER`, `FORCE_WINDOWS`, `DISABLE_SCRIPTING` and `EXCLUDE_BOOST` should stay.

CMakeSettings.json
fgerlits: done

CMakeSettings.json
fgerlits: I think there is value in `CMakeSettings.json` being in sync with `win_build_vs.bat`; if we change a default, that will not automatically get applied there, either.  So I'd rather leave it like this.

CMakeSettings.json
```diff
@@ -2,97 +2,97 @@
   // See https://go.microsoft.com//fwlink//?linkid=834763 for more information about this file.
   "configurations": [
     {
-      "name": "x64-Release",
-      "generator": "Visual Studio 15 2017 Win64",
+      "name": "x86-RelWithDebInfo",
+      "generator": "Visual Studio 15 2017",
       "configurationType": "RelWithDebInfo",
-      "inheritEnvironments": [ "msvc_x64_x64" ],
+      "inheritEnvironments": [ "msvc_x64_x86" ],
```
szaszm, line 8: I think this environment is invalid. I would simply go for `msvc_x86`.

https://docs.microsoft.com/en-us/cpp/build/cmakesettings-reference?view=vs-2019#environments

CMakeSettings.json
```diff
@@ -2,97 +2,97 @@
...
       "buildRoot": "${env.USERPROFILE}\\CMakeBuilds\\${workspaceHash}\\build\\${name}",
       "installRoot": "${env.USERPROFILE}\\CMakeBuilds\\${workspaceHash}\\install\\${name}",
       "cmakeCommandArgs": "",
-      "buildCommandArgs": "-m -v:minimal",
+      "buildCommandArgs": "",
```
szaszm, line 12: I think these empty `*CommandArgs` can be omitted.

CMakeSettings.json
fgerlits: Thanks, I have changed it to `msvc_x86`.  I got the architecture name from https://docs.microsoft.com/en-us/cpp/build/building-on-the-command-line?view=vs-2017, but you are right, the environment names are backwards.  This did work, but I don't know how or why. :)  I think VS may have simply ignored it, and it did an x86 build because of the generator setting.

CMakeSettings.json
fgerlits: I think these don't hurt, and they will make it easier to add command-line arguments later if we need them.

CMakeSettings.json
```diff
@@ -2,97 +2,97 @@
...
+          "name": "FORCE_WINDOWS",
+          "value": "ON"
         },
         {
-          "name": "ENABLE_LIBRDKAFKA",
+          "name": "USE_SYSTEM_UUID",
           "value": "OFF"
         },
         {
-          "name": "DISABLE_CURL",
```
hunyadi-dev, line 71: What happened to this flag?

szaszm: It is not present in `win_build_vs.bat` and @fgerlits changed this file to be roughly equivalent to `win_build_vs.bat /D`

PR #855, opened 2020-07-29
MINIFICPP-1288 - Remove FlowController::reload() as it is both flawed and unused

hunyadi-dev:
Move FlowController::reload() where it belongs. 🚮


jdye64:
Seems like the only issue with CI here was the job taking too long (and only for OSX build). I see nothing wrong here and I'm going to merge this.

PR #858, opened 2020-08-03
MINIFICPP-1300 - Decouple ControllerServiceProvider ownerships from use-only components

hunyadi-dev:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/coap/protocols/CoapC2Protocol.cpp
+++ b/extensions/coap/protocols/CoapC2Protocol.cpp
@@ -39,7 +39,7 @@ CoapProtocol::CoapProtocol(const std::string &name, const utils::Identifier &uui
 
 CoapProtocol::~CoapProtocol() = default;
 
-void CoapProtocol::initialize(const std::shared_ptr<core::controller::ControllerServiceProvider> &controller, const std::shared_ptr<Configure> &configure) {
+void CoapProtocol::initialize(core::controller::ControllerServiceProvider* controller, const std::shared_ptr<Configure> &configure) {
   RESTSender::initialize(controller, configure);
   if (configure->get("nifi.c2.coap.connector.service", controller_service_name_)) {
     auto service = controller->getControllerService(controller_service_name_);
--- a/libminifi/include/FlowController.h
+++ b/libminifi/include/FlowController.h
@@ -235,7 +235,7 @@ class FlowController : public core::controller::ControllerServiceProvider, publi
    * @param id service identifier
    * @return shared pointer to the controller service node or nullptr if it does not exist.
    */
-  std::shared_ptr<core::controller::ControllerServiceNode> getControllerServiceNode(const std::string &id) override;
+  std::shared_ptr<core::controller::ControllerServiceNode> getControllerServiceNode(const std::string &id) const override;
 
   void verifyCanStopReferencingComponents(std::shared_ptr<core::controller::ControllerServiceNode> &serviceNode) override;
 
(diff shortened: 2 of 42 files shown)


extensions/coap/protocols/CoapC2Protocol.cpp
```diff
@@ -39,7 +39,7 @@ CoapProtocol::CoapProtocol(const std::string &name, const utils::Identifier &uui
 
 CoapProtocol::~CoapProtocol() = default;
 
-void CoapProtocol::initialize(const std::shared_ptr<core::controller::ControllerServiceProvider> &controller, const std::shared_ptr<Configure> &configure) {
+void CoapProtocol::initialize(core::controller::ControllerServiceProvider* controller, const std::shared_ptr<Configure> &configure) {
```
adamdebreceni, line 42: if we take a pointer, it's our responsibility to null-check it, we assume here that it is non-null, we should either handle nullptr or take a reference, (in theory we could use a gls::non_null<...>, but references are more succinctly command validity)

adamdebreceni: I take it back, since we intend to store the pointer (in `initialize`) I agree with the pointer

libminifi/include/FlowController.h
```diff
@@ -339,6 +339,12 @@ class FlowController : public core::controller::ControllerServiceProvider, publi
    */
   virtual void initializePaths(const std::string &adjustedFilename);
 
+  template <typename T, typename = typename std::enable_if<std::is_base_of<SchedulingAgent, T>::value>::type>
+  void conditionalReloadScheduler(std::shared_ptr<T>& scheduler, const bool condition) {
```
adamdebreceni, line 343: would make this template private 
(as a matter of fact, I would make this whole protected block private, and use the discussed TestAccessor method, but that can wait, as I believe that would be out of scope for this PR)

hunyadi-dev: Changed this to private.

libminifi/include/core/ProcessContextBuilder.h
```diff
@@ -40,6 +40,7 @@
 #include "core/Repository.h"
 #include "core/FlowFile.h"
 #include "VariableRegistry.h"
+#include "utils/gsl.h"
```
szaszm, line 43: I think this include is not necessary, probably a leftover from a previous development version.

nanofi/include/cxx/Instance.h
```diff
@@ -134,7 +134,7 @@ class Instance {
 
   void transfer(const std::shared_ptr<FlowFileRecord> &ff, const std::shared_ptr<minifi::io::DataStream> &stream = nullptr) {
     std::shared_ptr<core::controller::ControllerServiceProvider> controller_service_provider = nullptr;
-    auto processContext = std::make_shared<core::ProcessContext>(proc_node_, controller_service_provider, no_op_repo_, no_op_repo_, configure_, content_repo_);
+    auto processContext = std::make_shared<core::ProcessContext>(proc_node_, controller_service_provider.get(), no_op_repo_, no_op_repo_, configure_, content_repo_);
```
szaszm, line 137: Why no `nullptr` substitution in this file?

hunyadi-dev: I do not know. Will replace this with a `nullptr` literal.

hunyadi-dev: Updated.

libminifi/include/core/ProcessContextBuilder.h
hunyadi-dev: Good catch, this is now removed.

szaszm (approved):

arpadboda (approved):

PR #861, opened 2020-08-04
MINIFICPP-1312 - Fix flaky FlowControllerTest

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/test/flow-tests/FlowControllerTests.cpp
+++ b/libminifi/test/flow-tests/FlowControllerTests.cpp
@@ -144,9 +144,8 @@ TEST_CASE("Flow shutdown waits for a while", "[TestFlow2]") {
   testController.startFlow();
 
   // wait for the source processor to enqueue its flowFiles
-  busy_wait(std::chrono::milliseconds{50}, [&] {return root->getTotalFlowFileCount() != 0;});
+  busy_wait(std::chrono::milliseconds{50}, [&] {return root->getTotalFlowFileCount() == 3;});
 
-  REQUIRE(root->getTotalFlowFileCount() != 0);
   REQUIRE(sourceProc->trigger_count.load() == 1);
 
   execSinkPromise.set_value();
@@ -181,9 +180,8 @@ TEST_CASE("Flow stopped after grace period", "[TestFlow3]") {
   testController.startFlow();
 
   // wait for the source processor to enqueue its flowFiles
-  busy_wait(std::chrono::milliseconds{50}, [&] {return root->getTotalFlowFileCount() != 0;});
+  busy_wait(std::chrono::milliseconds{50}, [&] {return root->getTotalFlowFileCount() == 3;});
 
-  REQUIRE(root->getTotalFlowFileCount() != 0);
   REQUIRE(sourceProc->trigger_count.load() == 1);
 
   execSinkPromise.set_value();
@@ -220,9 +218,8 @@ TEST_CASE("Extend the waiting period during shutdown", "[TestFlow4]") {
   testController.startFlow();
 
   // wait for the source processor to enqueue its flowFiles
-  busy_wait(std::chrono::milliseconds{50}, [&] {return root->getTotalFlowFileCount() != 0;});
+  busy_wait(std::chrono::milliseconds{50}, [&] {return root->getTotalFlowFileCount() == 3;});
 
-  REQUIRE(root->getTotalFlowFileCount() != 0);
   REQUIRE(sourceProc->trigger_count.load() == 1);
 
   std::thread shutdownThread([&]{
@@ -230,13 +227,15 @@ TEST_CASE("Extend the waiting period during shutdown", "[TestFlow4]") {
     controller->stop(true);
   });
 
-  int extendCount = 0;
+  unsigned int extendCount = 0;
   while (extendCount++ < 5 && controller->isRunning()) {
-    testController.logger_->log_info("Controller still running, extend the waiting period, ff count: %d", static_cast<int>(root->getTotalFlowFileCount()));
+    testController.getLogger()->log_info("Controller still running, extend the waiting period %u, ff count: %u", extendCount, static_cast<unsigned int>(root->getTotalFlowFileCount()));
     std::this_thread::sleep_for(std::chrono::milliseconds{500});
     timeout_ms += 500;
     testController.configuration_->set(minifi::Configure::nifi_flowcontroller_drain_timeout, std::to_string(timeout_ms) + " ms");
   }
+
+  REQUIRE(!controller->isRunning());
 
   shutdownThread.join();
 
--- a/libminifi/test/flow-tests/FlowControllerTests.cpp
+++ b/libminifi/test/flow-tests/FlowControllerTests.cpp
@@ -138,9 +138,7 @@ TEST_CASE("Flow shutdown waits for a while", "[TestFlow2]") {
   std::promise<void> execSinkPromise;
   std::future<void> execSinkFuture = execSinkPromise.get_future();
   sinkProc->onTriggerCb_ = [&] {
-    testController.logger_->log_info("sinkProc onTrigger started");
     execSinkFuture.wait();
-    testController.logger_->log_info("sinkProc onTrigger calls impl");
   };
 
   testController.startFlow();
@@ -148,11 +146,9 @@ TEST_CASE("Flow shutdown waits for a while", "[TestFlow2]") {
   // wait for the source processor to enqueue its flowFiles
   busy_wait(std::chrono::milliseconds{50}, [&] {return root->getTotalFlowFileCount() != 0;});
 
-  testController.logger_->log_info("Verifying conditions");
   REQUIRE(root->getTotalFlowFileCount() != 0);
   REQUIRE(sourceProc->trigger_count.load() == 1);
 
-  testController.logger_->log_info("Continue sinkProc and stop controller");
   execSinkPromise.set_value();
   controller->stop(true);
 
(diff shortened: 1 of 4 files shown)


libminifi/test/flow-tests/FlowControllerTests.cpp
```diff
@@ -133,12 +144,9 @@ TEST_CASE("Flow shutdown waits for a while", "[TestFlow2]") {
   testController.startFlow();
 
   // wait for the source processor to enqueue its flowFiles
-  int tryCount = 0;
-  while (tryCount++ < 10 && root->getTotalFlowFileCount() != 3) {
-    std::this_thread::sleep_for(std::chrono::milliseconds{20});
-  }
+  busy_wait(std::chrono::milliseconds{50}, [&] {return root->getTotalFlowFileCount() != 0;});
```
fgerlits, line 147: You could use `verifyEventHappenedInPollTime` in IntegrationTestUtils.h, it is almost the same as `busy_wait`.

```suggestion
  REQUIRE(verifyEventHappenedInPollTime(std::chrono::milliseconds{50}, [&] {return root->getTotalFlowFileCount() != 0;}));
```

fgerlits, line 147: Is there no way to avoid changing `== 3` to `!= 0`?  The test condition is much weaker now.

libminifi/test/flow-tests/FlowControllerTests.cpp
```diff
@@ -133,12 +144,9 @@ TEST_CASE("Flow shutdown waits for a while", "[TestFlow2]") {
...
 
-  REQUIRE(root->getTotalFlowFileCount() == 3);
+  REQUIRE(root->getTotalFlowFileCount() != 0);
```
fgerlits, line 149: why do we need this REQUIRE? `busy_wait` (or `verifyEventHappenedInPollTime` if you change to that) already checked this condition in the previous line

libminifi/test/flow-tests/FlowControllerTests.cpp
```diff
@@ -215,20 +220,19 @@ TEST_CASE("Extend the waiting period during shutdown", "[TestFlow4]") {
...
 
   std::thread shutdownThread([&]{
     execSinkPromise.set_value();
     controller->stop(true);
   });
 
-  while (controller->isRunning()) {
+  int extendCount = 0;
+  while (extendCount++ < 5 && controller->isRunning()) {
+    testController.logger_->log_info("Controller still running, extend the waiting period, ff count: %d", static_cast<int>(root->getTotalFlowFileCount()));
```
fgerlits, line 235: it would be useful to include the value of `extendCount ` in this log

libminifi/test/flow-tests/FlowControllerTests.cpp
hunyadi-dev, line 147: Why not simply:
```c++
assert(verifyEventHappenedInPollTime(std::chrono::milliseconds{50}, [&] {return root->getTotalFlowFileCount() != 0;})
```

libminifi/test/flow-tests/FlowControllerTests.cpp
hunyadi-dev, line 149: Would we not have already thrown if this was not true?

libminifi/test/flow-tests/FlowControllerTests.cpp
```diff
@@ -215,20 +220,19 @@ TEST_CASE("Extend the waiting period during shutdown", "[TestFlow4]") {
...
   REQUIRE(sourceProc->trigger_count.load() == 1);
 
   std::thread shutdownThread([&]{
     execSinkPromise.set_value();
     controller->stop(true);
   });
 
-  while (controller->isRunning()) {
+  int extendCount = 0;
+  while (extendCount++ < 5 && controller->isRunning()) {
```
hunyadi-dev, line 234: Maybe we could have more attempts with smaller timeout increases.

libminifi/test/flow-tests/FlowControllerTests.cpp
hunyadi-dev: The same applies to the other similar changes.

libminifi/test/flow-tests/FlowControllerTests.cpp
hunyadi-dev: The same applies to the other similar changes.

libminifi/test/flow-tests/FlowControllerTests.cpp
hunyadi-dev, line 235: Does `%u` not work here?

libminifi/test/flow-tests/FlowControllerTests.cpp
adamdebreceni: only seemingly weaker, since the `3` flowFiles are generated in batch, and their insertion is one atomic operation

adamdebreceni: the `sleep_for` inside the `verifyEventHappenedInPollTime` is problematic, as I would like to wait for the sourceProcessor to get executed once, and as it turns out a on the github agents a `sleep_for(20ms)` might happily wait more than `100ms` 

(note that I can't signal the main thread from the `onTrigger` as the shutdown logic checks the actual contents of the connections, so if I signal from the `onTrigger` and initiate a flowController shutdown the generated flowFiles might not yet have been enqueued and do not trigger the graceful shutdown)

hunyadi-dev:
I see I have been reviewing parallel to @fgerlits. Please mark the duplicates resolved :)

libminifi/test/flow-tests/FlowControllerTests.cpp
adamdebreceni: done

libminifi/test/flow-tests/FlowControllerTests.cpp
adamdebreceni: removed the superfluous checks

libminifi/test/flow-tests/FlowControllerTests.cpp
adamdebreceni: added

libminifi/test/flow-tests/FlowControllerTests.cpp
adamdebreceni: removed

libminifi/test/flow-tests/FlowControllerTests.cpp
adamdebreceni: wrote back the `== 3` as it should make no difference

adamdebreceni: see [this](https://github.com/apache/nifi-minifi-cpp/pull/861#discussion_r465030783), we could extend the `verifyEventHappenedInPollTime` with a `poll_interval_ms` where `-1` would indicate a busy checking, what do you think?

fgerlits: I would prefer a function without a timeout parameter (possibly with both the timeout-y and timeout-less version calling a common implementation).

hunyadi-dev: I cannot see where to link points, but -1 from me for the additional argument as deducing if we need to sleep should be a compile-time operation.

adamdebreceni: @hunyadi-dev I don't see why it should be compile-time decidable if we need sleep but alright will not add an extra parameter, @fgerlits could you expand on your proposal?

fgerlits: @adebreceni It wasn't really a proposal, but I would prefer two separate functions `verifyEventHappenedInPollTime()` and `verifyEventHappenedInPollTimeWithBusyWait()` (just an example, feel free to find a shorter name) rather than a single function where an argument value of -1 means "use the busy version".

[TBH, I am not a fan of the `verifyEventHappenedInPollTime` name, as something like `REQUIRE(happensBeforeTimeout(...))` would read better, but renaming it is probably not in scope for this PR.]

libminifi/test/flow-tests/FlowControllerTests.cpp
szaszm: shorter sleeps and checking for a timeout instead of run count should achieve the same goal without multiplying the uncertainty of the agents.

adamdebreceni: note that here we are not simply waiting for the controller to shutdown, but we are verifying if we can progressively delay the shutdown by adjusting the timeout parameter, shorter sleeps seem to be relatively more imprecise than longer ones on the agents, I understand now that this loop might not be the best so I will unroll it to properly convey its purpose

libminifi/test/flow-tests/FlowControllerTests.cpp
adamdebreceni: changed it to `verifyWithBusyWait`

libminifi/test/flow-tests/FlowControllerTests.cpp
```diff
@@ -79,6 +79,17 @@ Flow Controller:
 Remote Processing Groups:
 )";
 
+template<typename Fn>
+bool verifyWithBusyWait(std::chrono::milliseconds timeout, Fn&& fn) {
+  auto start = std::chrono::steady_clock::now();
+  while (std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - start) < timeout) {
```
fgerlits, line 85: minor, but I don't think you need the `duration_cast`: durations in different units can be compared

szaszm:
Merged in 53ba922, but forgot to link the PR in the commit.

PR #862, opened 2020-08-04
MINIFICPP-1318 migrate PersistenceTests to /var/tmp

szaszm:
Related: [MINIFICPP-1188](https://issues.apache.org/jira/browse/MINIFICPP-1188)
______
Thank you for submitting a contribution to Apache NiFi - MiNiFi C++.

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [x] Is there a JIRA ticket associated with this PR? Is it referenced
     in the commit message?

- [x] Does your PR title start with MINIFICPP-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.

- [x] Has your PR been rebased against the latest commit within the target branch (typically master)?

- [x] Is your initial contribution a single, squashed commit?

### For code changes:
- [x] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?
- [x] If applicable, have you updated the LICENSE file?
- [x] If applicable, have you updated the NOTICE file?

### For documentation related changes:
- [x] Have you ensured that format looks appropriate for the output in which it is rendered?

### Note:
Please ensure that once the PR is submitted, you check travis-ci for build issues and submit an update to your PR as soon as possible.

PR #859, opened 2020-08-03
MINIFICPP-1313 Do a multifile lookup in onSchedule

fgerlits:
https://issues.apache.org/jira/browse/MINIFICPP-1313

Do a first multifile lookup in `onSchedule()` to make sure TailFile can see the files which were already present before minifi started.  This fixes an intermittent CI failure caused by TailFileTests running on the virtual machine less than 10 minutes after bootup.

---

Thank you for submitting a contribution to Apache NiFi - MiNiFi C++.

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [x] Is there a JIRA ticket associated with this PR? Is it referenced
     in the commit message?

- [x] Does your PR title start with MINIFICPP-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.

- [x] Has your PR been rebased against the latest commit within the target branch (typically master)?

- [x] Is your initial contribution a single, squashed commit?

### For code changes:
- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?
- [ ] If applicable, have you updated the LICENSE file?
- [ ] If applicable, have you updated the NOTICE file?

### For documentation related changes:
- [ ] Have you ensured that format looks appropriate for the output in which it is rendered?

### Note:
Please ensure that once the PR is submitted, you check travis-ci for build issues and submit an update to your PR as soon as possible.

PR #863, opened 2020-08-05
WIP: MINIFICPP-1318 move test rocksdb state to /var/tmp

szaszm:
Missed TailFileTests state storage in #862 
______

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/test/flow-tests/FlowControllerTests.cpp
+++ b/libminifi/test/flow-tests/FlowControllerTests.cpp
@@ -79,6 +79,17 @@ Connections:
 Remote Processing Groups:
 )";
 
+template<typename Fn>
+bool verifyWithBusyWait(std::chrono::milliseconds timeout, Fn&& fn) {
+  auto start = std::chrono::steady_clock::now();
+  while (std::chrono::steady_clock::now() - start < timeout) {
+    if (fn()) {
+      return true;
+    }
+  }
+  return false;
+}
+
 TEST_CASE("Flow shutdown drains connections", "[TestFlow1]") {
   TestControllerWithFlow testController(yamlConfig);
   auto controller = testController.controller_;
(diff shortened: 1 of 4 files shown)


szaszm:
> There are a lot more:
> 
> ```
> ~/src/minifi$ git grep '"/tmp' | wc -l
> 147
> ```
> 
> the rest do not need to be changed?

They are fine as long as we don't try to create a rocksdb repository there. The problem is that rocksdb is configured to use direct IO, i.e. bypass the page cache, but tmpfs is just a mounted page cache without backing storage, so it makes no sense to use direct IO there.

arpadboda:
@szaszm : not sure if the CI failures we face on U16 jobs are related or not, but I would prefer to investigate before proceeding with this.

fgerlits:
> @szaszm : not sure if the CI failures we face on U16 jobs are related or not, but I would prefer to investigate before proceeding with this.

I am 99% sure that both of those failures in TailFileTests are due to the bug fixed in #859.  That PR has been merged now, so rebasing on top of it should get rid of the test failures.

libminifi/test/flow-tests/FlowControllerTests.cpp
```diff
@@ -193,7 +198,7 @@ TEST_CASE("Extend the waiting period during shutdown", "[TestFlow4]") {
   auto controller = testController.controller_;
   auto root = testController.root_;
 
-  int timeout_ms = 1000;
+  unsigned int timeout_ms = 1000;
```
adam-markovics, line 201: As this is a duration its type could also be std::chrono::duration, with no _ms suffix needed then.

szaszm:
The CI issue is fixed in #879, thanks to @fgerlits

PR #865, opened 2020-08-07
MINIFICPP-1325 - Separate process group processor lookups for IDs and names, allow constructing ID-s from strings

hunyadi-dev:
This is a cleanup PR before I start a refactor on configuration parsing, IDs and names lookups.

### DIFF
--- a/libminifi/include/core/ProcessGroup.h
+++ b/libminifi/include/core/ProcessGroup.h
@@ -199,10 +199,26 @@ class ProcessGroup {
   void removeProcessGroup(ProcessGroup *child);
   // ! Add connections
   void addConnection(std::shared_ptr<Connection> connection);
+  // Generic find
+  template <typename Fun>
+  std::shared_ptr<Processor> findProcessor(Fun&& condition) const {
+    std::lock_guard<std::recursive_mutex> lock(mutex_);
+    const auto found = std::find_if(processors_.cbegin(), processors_.cend(), condition);
+    if (found != processors_.cend()) {
+      return *found;
+    }
+    for (const auto& processGroup : child_process_groups_) {
+      const std::shared_ptr<Processor> processor = processGroup->findProcessor(condition);
+      if (processor) {
+        return processor;
+      }
+    }
+    return nullptr;
+  }
   // findProcessor based on UUID
-  std::shared_ptr<Processor> findProcessor(utils::Identifier &uuid);
+  std::shared_ptr<Processor> findProcessorById(const utils::Identifier& uuid) const;
   // findProcessor based on name
-  std::shared_ptr<Processor> findProcessor(const std::string &processorName);
+  std::shared_ptr<Processor> findProcessorByName(const std::string &processorName) const;
 
   void getAllProcessors(std::vector<std::shared_ptr<Processor>> &processor_vec);
   /**
@@ -274,7 +290,7 @@ class ProcessGroup {
 
  private:
   // Mutex for protection
-  std::recursive_mutex mutex_;
+  mutable std::recursive_mutex mutex_;
   // Logger
   std::shared_ptr<logging::Logger> logger_;
   // Prevent default copy constructor and assignment operation
--- a/libminifi/include/core/ProcessGroup.h
+++ b/libminifi/include/core/ProcessGroup.h
@@ -201,7 +201,7 @@ class ProcessGroup {
   void addConnection(std::shared_ptr<Connection> connection);
   // Generic find
   template <typename Fun>
-  std::shared_ptr<Processor> findProcessor(Fun&& condition) const {
+  std::shared_ptr<Processor> findProcessor(Fun condition) const {
     std::lock_guard<std::recursive_mutex> lock(mutex_);
     const auto found = std::find_if(processors_.cbegin(), processors_.cend(), condition);
     if (found != processors_.cend()) {
--- a/libminifi/include/utils/Id.h
+++ b/libminifi/include/utils/Id.h
@@ -107,9 +107,12 @@ class IdentifierBase {
 typedef uint8_t UUID_FIELD[16];
 
 class Identifier : public IdentifierBase<UUID_FIELD, std::string> {
+  static const char* UUID_FORMAT_STRING;
+
  public:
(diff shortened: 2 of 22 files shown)


libminifi/src/core/ProcessGroup.cpp
```diff
@@ -263,26 +263,21 @@ void ProcessGroup::stopProcessing(const std::shared_ptr<TimerDrivenSchedulingAge
   }
 }
 
-std::shared_ptr<Processor> ProcessGroup::findProcessor(utils::Identifier &uuid) {
-  std::lock_guard<std::recursive_mutex> lock(mutex_);
-  std::shared_ptr<Processor> ret = NULL;
-  for (auto processor : processors_) {
-    logger_->log_debug("find processor %s", processor->getName());
+std::shared_ptr<Processor> ProcessGroup::findProcessorById(const utils::Identifier& uuid) const {
+  const auto id_matches = [&] (const std::shared_ptr<Processor>& processor) {
```
lordgamez, line 267: It might be safer to only capture uuid by reference if this changes in the future

hunyadi-dev: I disagree, capturing by `&` is safe long as the lambda never outlives its scope.

libminifi/src/core/ProcessGroup.cpp
```diff
@@ -263,26 +263,21 @@ void ProcessGroup::stopProcessing(const std::shared_ptr<TimerDrivenSchedulingAge
...
-  }
-  for (auto processGroup : child_process_groups_) {
-    logger_->log_debug("find processor child %s", processGroup->getName());
-    std::shared_ptr<Processor> processor = processGroup->findProcessor(uuid);
-    if (processor)
-      return processor;
-  }
-  return ret;
+std::shared_ptr<Processor> ProcessGroup::findProcessorByName(const std::string &processorName) const {
+  const auto name_matches = [&] (const std::shared_ptr<Processor>& processor) {
```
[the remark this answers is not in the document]
hunyadi-dev: Capturing by `&` is safe long as the lambda never outlives its scope. Also, `this` is captured here as well.

libminifi/include/core/ProcessGroup.h
```diff
@@ -199,10 +199,26 @@ class ProcessGroup {
   void removeProcessGroup(ProcessGroup *child);
   // ! Add connections
   void addConnection(std::shared_ptr<Connection> connection);
+  // Generic find
+  template <typename Fun>
+  std::shared_ptr<Processor> findProcessor(Fun&& condition) const {
```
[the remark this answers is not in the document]
hunyadi-dev: No, you are right, I realized that I could not forward here, so I removed the syntax for it, but left the &&.

hunyadi-dev:
> For some reason I couldn't reply to the other review comments, so writing it here:
> +1 for explicitly listing the captured variables, just for clarity, not safety.

I would say listing the captured variables is just clutter-code in these cases. The lambdas mentioned captures everything in scope: there are no locals, `this` is captured and the single function args are also inputs for the lambda.

libminifi/src/utils/Id.cpp
```diff
@@ -48,6 +48,8 @@ namespace nifi {
 namespace minifi {
 namespace utils {
 
+const char* Identifier::UUID_FORMAT_STRING = "%02hhx%02hhx%02hhx%02hhx-%02hhx%02hhx-%02hhx%02hhx-%02hhx%02hhx-%02hhx%02hhx%02hhx%02hhx%02hhx%02hhx";
```
adamdebreceni, line 51: would `constexpr` allow us to make it an in-class initializer?

hunyadi-dev: I think it only works in c++17 or maybe even c++14.

adamdebreceni: you are right that from `c++17` `inline` comes to the rescue, but even in `c++11` we may (must) add an initializer to the declaration of a literal constexpr variable (and it is enough if it is not odr-used) (see [here](https://en.cppreference.com/w/cpp/language/static) in section `Constant static members`)

```
#include <stdio.h>

struct A{
  static constexpr const char* FORMAT = "%d\n";
};

int main(){
  // uncommenting the following will generate a linker error
  // auto& ref = A::FORMAT;
  printf(A::FORMAT, 5);
}
```
it is not pivotal to move it to the header, I just wanted to share that it is possible

szaszm: Please make sure I can't reassign the string to something else. `constexpr` or top-level `const` achieves that.

hunyadi-dev: Made `UUID_FORMAT_STRING` `constexpr`.

libminifi/include/utils/Id.h
```diff
@@ -107,9 +107,12 @@ class IdentifierBase {
 typedef uint8_t UUID_FIELD[16];
 
 class Identifier : public IdentifierBase<UUID_FIELD, std::string> {
+  static constexpr const char* UUID_FORMAT_STRING = "%02hhx%02hhx%02hhx%02hhx-%02hhx%02hhx-%02hhx%02hhx-%02hhx%02hhx-%02hhx%02hhx%02hhx%02hhx%02hhx%02hhx";
```
szaszm, line 110: Pre-C++17, this is only a declaration that requires a definition. After C++17, this is redefined to be a definition and the namespace-scope one is a deprecated but accepted declaration.
See the bottom is this page: https://en.cppreference.com/w/cpp/language/static

Put this in the .cpp file: `constexpr const char* Identifier::UUID_FORMAT_STRING;`.

adamdebreceni: I believe it is not necessary since UUID_FORMAT_STRING is not odr-used

> A variable whose name appears as a potentially-evaluated expression is odr-used unless it is an
object that satisfies the requirements for appearing in a constant expression (5.19) and the lvalue-to-rvalue
conversion (4.1) is immediately applied.

szaszm (approved):

arpadboda (approved):

PR #866, opened 2020-08-10
MINIFICPP-1321 Support attribute strategies in MergeContent processor

lordgamez:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/utils/IntegrationTestUtils.h
+++ b/libminifi/include/utils/IntegrationTestUtils.h
@@ -32,13 +32,13 @@ namespace utils {
 
 template <class Rep, class Period, typename Fun>
 bool verifyEventHappenedInPollTime(const std::chrono::duration<Rep, Period>& wait_duration, Fun&& check) {
-  std::chrono::system_clock::time_point wait_end = std::chrono::system_clock::now() + wait_duration;
+  std::chrono::steady_clock::time_point wait_end = std::chrono::steady_clock::now() + wait_duration;
   do {
     if (std::forward<Fun>(check)()) {
       return true;
     }
     std::this_thread::sleep_for(std::chrono::milliseconds(100));
-  } while (std::chrono::system_clock::now() < wait_end);
+  } while (std::chrono::steady_clock::now() < wait_end);
   return false;
 }
 
--- a/libminifi/test/flow-tests/FlowControllerTests.cpp
+++ b/libminifi/test/flow-tests/FlowControllerTests.cpp
@@ -79,6 +79,17 @@ Connections:
 Remote Processing Groups:
 )";
 
+template<typename Fn>
+bool verifyWithBusyWait(std::chrono::milliseconds timeout, Fn&& fn) {
+  auto start = std::chrono::steady_clock::now();
+  while (std::chrono::steady_clock::now() - start < timeout) {
+    if (fn()) {
+      return true;
+    }
+  }
+  return false;
+}
+
 TEST_CASE("Flow shutdown drains connections", "[TestFlow1]") {
   TestControllerWithFlow testController(yamlConfig);
   auto controller = testController.controller_;
@@ -133,12 +144,9 @@ TEST_CASE("Flow shutdown waits for a while", "[TestFlow2]") {
   testController.startFlow();
 
   // wait for the source processor to enqueue its flowFiles
-  int tryCount = 0;
-  while (tryCount++ < 10 && root->getTotalFlowFileCount() != 3) {
-    std::this_thread::sleep_for(std::chrono::milliseconds{20});
-  }
+  auto flowFilesEnqueued = [&] {return root->getTotalFlowFileCount() == 3;};
+  REQUIRE(verifyWithBusyWait(std::chrono::milliseconds{50}, flowFilesEnqueued));
 
-  REQUIRE(root->getTotalFlowFileCount() == 3);
   REQUIRE(sourceProc->trigger_count.load() == 1);
 
   execSinkPromise.set_value();
@@ -173,12 +181,9 @@ TEST_CASE("Flow stopped after grace period", "[TestFlow3]") {
   testController.startFlow();
 
   // wait for the source processor to enqueue its flowFiles
-  int tryCount = 0;
-  while (tryCount++ < 10 && root->getTotalFlowFileCount() != 3) {
-    std::this_thread::sleep_for(std::chrono::milliseconds{20});
-  }
+  auto flowFilesEnqueued = [&] {return root->getTotalFlowFileCount() == 3;};
+  REQUIRE(verifyWithBusyWait(std::chrono::milliseconds{50}, flowFilesEnqueued));
 
-  REQUIRE(root->getTotalFlowFileCount() == 3);
   REQUIRE(sourceProc->trigger_count.load() == 1);
 
   execSinkPromise.set_value();
@@ -193,7 +198,7 @@ TEST_CASE("Extend the waiting period during shutdown", "[TestFlow4]") {
   auto controller = testController.controller_;
   auto root = testController.root_;
 
-  int timeout_ms = 1000;
+  unsigned int timeout_ms = 1000;
 
   testController.configuration_->set(minifi::Configure::nifi_flowcontroller_drain_timeout, std::to_string(timeout_ms) + " ms");
 
@@ -215,12 +220,9 @@ TEST_CASE("Extend the waiting period during shutdown", "[TestFlow4]") {
   testController.startFlow();
 
   // wait for the source processor to enqueue its flowFiles
-  int tryCount = 0;
-  while (tryCount++ < 10 && root->getTotalFlowFileCount() != 3) {
-    std::this_thread::sleep_for(std::chrono::milliseconds{20});
-  }
+  auto flowFilesEnqueued = [&] {return root->getTotalFlowFileCount() == 3;};
+  REQUIRE(verifyWithBusyWait(std::chrono::milliseconds{50}, flowFilesEnqueued));
 
-  REQUIRE(root->getTotalFlowFileCount() == 3);
   REQUIRE(sourceProc->trigger_count.load() == 1);
 
   std::thread shutdownThread([&]{
@@ -228,11 +230,20 @@ TEST_CASE("Extend the waiting period during shutdown", "[TestFlow4]") {
     controller->stop(true);
   });
 
-  while (controller->isRunning()) {
-    std::this_thread::sleep_for(std::chrono::milliseconds{500});
+  auto shutdownInitiated = std::chrono::steady_clock::now();
+  auto shutdownDuration = [&] {return std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - shutdownInitiated);};
+
+  std::this_thread::sleep_for(std::chrono::milliseconds{500});
+  while (shutdownDuration() < std::chrono::milliseconds(2500) && controller->isRunning()) {
     timeout_ms += 500;
+    testController.getLogger()->log_info("Controller still running after %u ms, extending the waiting period to %u ms, ff count: %u",
+        static_cast<unsigned int>(shutdownDuration().count()), timeout_ms, static_cast<unsigned int>(root->getTotalFlowFileCount()));
     testController.configuration_->set(minifi::Configure::nifi_flowcontroller_drain_timeout, std::to_string(timeout_ms) + " ms");
+    std::this_thread::sleep_for(std::chrono::milliseconds{500});
   }
+
+  REQUIRE(!controller->isRunning());
+  REQUIRE(shutdownDuration().count() > 1500);
 
   shutdownThread.join();
 
--- a/libminifi/test/flow-tests/TestControllerWithFlow.h
+++ b/libminifi/test/flow-tests/TestControllerWithFlow.h
@@ -25,6 +25,12 @@ class TestControllerWithFlow: public TestController{
     LogTestController::getInstance().setTrace<processors::TestProcessor>();
     LogTestController::getInstance().setTrace<processors::TestFlowFileGenerator>();
     LogTestController::getInstance().setTrace<minifi::Connection>();
+    LogTestController::getInstance().setTrace<core::Connectable>();
+    LogTestController::getInstance().setTrace<minifi::SchedulingAgent>();
+    LogTestController::getInstance().setTrace<minifi::ThreadedSchedulingAgent>();
+    LogTestController::getInstance().setTrace<core::Processor>();
+    LogTestController::getInstance().setTrace<minifi::TimerDrivenSchedulingAgent>();
+    LogTestController::getInstance().setTrace<minifi::EventDrivenSchedulingAgent>();
 
     char format[] = "/tmp/flowTest.XXXXXX";
     std::string dir = createTempDirectory(format);


extensions/libarchive/MergeContent.cpp
```diff
@@ -62,6 +62,11 @@ core::Property MergeContent::KeepPath(
   core::PropertyBuilder::createProperty("Keep Path")
   ->withDescription("If using the Zip or Tar Merge Format, specifies whether or not the FlowFiles' paths should be included in their entry")
   ->withDefaultValue(false)->build());
+core::Property MergeContent::AttributeStrategy(
+  core::PropertyBuilder::createProperty("Attribute Strategy")
+  ->withDescription("Determines which FlowFile attributes should be added to the bundle.")
```
hunyadi-dev, line 67: Can you please extend the description to the one used in the docs here?
http://nifi.apache.org/docs/nifi-docs/components/org.apache.nifi/nifi-standard-nar/1.11.4/org.apache.nifi.processors.standard.MergeContent/

extensions/libarchive/MergeContent.cpp
```diff
@@ -140,6 +146,10 @@ void MergeContent::onSchedule(core::ProcessContext *context, core::ProcessSessio
   if (context->getProperty(KeepPath.getName(), value) && !value.empty()) {
     org::apache::nifi::minifi::utils::StringUtils::StringToBool(value, keepPath_);
   }
+  value = "";
+  if (context->getProperty(AttributeStrategy.getName(), value) && !value.empty()) {
+      this->attributeStrategy_ = value;
```
hunyadi-dev, line 151: What is the point of having an explicit this?

extensions/libarchive/MergeContent.cpp
```diff
@@ -140,6 +146,10 @@ void MergeContent::onSchedule(core::ProcessContext *context, core::ProcessSessio
   if (context->getProperty(KeepPath.getName(), value) && !value.empty()) {
     org::apache::nifi::minifi::utils::StringUtils::StringToBool(value, keepPath_);
   }
+  value = "";
```
hunyadi-dev, line 149: What is the value of resetting `value` between uses?

extensions/libarchive/MergeContent.cpp
```diff
@@ -271,6 +281,16 @@ bool MergeContent::processBin(core::ProcessContext *context, core::ProcessSessio
       return false;
     }
     session->putAttribute(mergeFlow, BinFiles::FRAGMENT_COUNT_ATTRIBUTE, std::to_string(bin->getSize()));
+
+    std::unique_ptr<AttributeMerger> attributeMerger;
+    if (attributeStrategy_ == ATTRIBUTE_STRATEGY_KEEP_COMMON)
+      attributeMerger = std::unique_ptr < AttributeMerger > (new KeepOnlyCommonAttributesMerger(session, bin->getFlowFile(), mergeFlow));
```
hunyadi-dev, line 287: We have `utils::make_unique` available in `GeneralUtils.h`.

extensions/libarchive/MergeContent.cpp
```diff
@@ -271,6 +281,16 @@ bool MergeContent::processBin(core::ProcessContext *context, core::ProcessSessio
...
+    else if (attributeStrategy_ == ATTRIBUTE_STRATEGY_KEEP_ALL_UNIQUE)
+      attributeMerger = std::unique_ptr < AttributeMerger > (new KeepAllUniqueAttributesMerger(session, bin->getFlowFile(), mergeFlow));
```
hunyadi-dev, line 289: Minor, but padding `<`-s from both sides makes it seem like they are relational operators.
See Templates and Casts in the google styleguide we follow:
https://google.github.io/styleguide/cppguide.html#Horizontal_Whitespace

extensions/libarchive/MergeContent.cpp
```diff
@@ -271,6 +281,16 @@ bool MergeContent::processBin(core::ProcessContext *context, core::ProcessSessio
...
+    else
+      return false;
```
hunyadi-dev, line 291: Maybe we should [log something](https://xkcd.com/2200/) here.

extensions/libarchive/MergeContent.cpp
```diff
@@ -342,6 +362,56 @@ std::shared_ptr<core::FlowFile> ZipMerge::merge(core::ProcessContext *context, c
...
+      processFlowFile(flow, commonAttributes);
+    }
+  }
+  return commonAttributes;
+}
+
+void KeepOnlyCommonAttributesMerger::processFlowFile(const std::shared_ptr<core::FlowFile> &flow, std::map<std::string, std::string>& commonAttributes) {
+  auto flowAttributes = flow->getAttributes();
+  for (auto it = commonAttributes.cbegin(); it != commonAttributes.cend();) {
+    if (flowAttributes.find(it->first) != flowAttributes.end()) {
```
adamdebreceni, line 390: with `FlowFile::getAttribute` we could both check for existence and extract the value in one motion (here for `flowAttributes` and in the other `processFlowFile` for `commonAttributes`)

extensions/libarchive/MergeContent.cpp
```diff
@@ -342,6 +362,56 @@ std::shared_ptr<core::FlowFile> ZipMerge::merge(core::ProcessContext *context, c
   return flowFile;
 }
 
+void AttributeMerger::mergeAttributes() {
+  std::map<std::string, std::string> commonAttributes = getCommonAttributes();
```
hunyadi-dev, line 366: Could be moved directly to the for loop below.

adamdebreceni, line 366: we could do without the local variable, also I would rename `getCommonAttributes` to `getMergedAttributes` (or something else) as "common" has a specific meaning in this context, i.e. attributes that have the same value in all flowFiles, (similarly for the parameter `commonAttributes` in the `processFlowFile` methods)

extensions/libarchive/MergeContent.cpp
adamdebreceni, line 287: I would keep it on the stack and immediately use it like `KeepOnlyCommonAttributesMerger(session, bin->getFlowFile(), mergeFlow).mergeAttributes()` it really is a single-use object (and we have to be careful to keep it that way as `KeepAllUniqueAttributesMerger::mergeAttributes` is not idempotent)

extensions/libarchive/MergeContent.cpp
adamdebreceni: consistency, all others above this follow the same pattern, same for resetting the `value`, I'm not sure if removing the `this` improves the code or not

extensions/libarchive/MergeContent.cpp
adamdebreceni: I see now that `mergeBin` follows the same pattern, although there `->merge` could throw

extensions/libarchive/MergeContent.cpp
hunyadi-dev: We denote member variables with a trailing `_`. There is no added readibility in further specifying its explicitness.

lordgamez: I think having the `this` keyword is just redundant here if we have the `_` suffix for member variables, and as resetting the `value` has no reason in this case I can just remove them in all occurrences and keep the consistency.

extensions/libarchive/MergeContent.cpp
lordgamez: Good point, I'll also remove them at line 268-272 in this case.

extensions/libarchive/MergeContent.cpp
```diff
@@ -342,6 +362,56 @@ std::shared_ptr<core::FlowFile> ZipMerge::merge(core::ProcessContext *context, c
...
+
+  for (const auto& pair : commonAttributes) {
+    session_->putAttribute(mergeFlow_, pair.first, pair.second);
+  }
+}
+
+std::map<std::string, std::string> AttributeMerger::getCommonAttributes() {
+  std::map<std::string, std::string> commonAttributes;
```
hunyadi-dev, line 374: This could potentially simplify things:
https://en.cppreference.com/w/cpp/algorithm/set_intersection

extensions/libarchive/MergeContent.cpp
```diff
@@ -342,6 +362,56 @@ std::shared_ptr<core::FlowFile> ZipMerge::merge(core::ProcessContext *context, c
...
+    }
+  }
+  return commonAttributes;
+}
+
+void KeepOnlyCommonAttributesMerger::processFlowFile(const std::shared_ptr<core::FlowFile> &flow, std::map<std::string, std::string>& commonAttributes) {
+  auto flowAttributes = flow->getAttributes();
+  for (auto it = commonAttributes.cbegin(); it != commonAttributes.cend();) {
+    if (flowAttributes.find(it->first) != flowAttributes.end()) {
+      if (flowAttributes[it->first] != commonAttributes[it->first])
```
hunyadi-dev, line 391: Minor, but do not use `operator[]` on a map for strict element access. Use `std::map::at()` instead as you never want any insertion here.

hunyadi-dev, line 391: `commonAttributes[it->first] == it->second`?

extensions/libarchive/MergeContent.cpp
```diff
@@ -342,6 +362,56 @@ std::shared_ptr<core::FlowFile> ZipMerge::merge(core::ProcessContext *context, c
...
+
+void KeepAllUniqueAttributesMerger::processFlowFile(const std::shared_ptr<core::FlowFile> &flow, std::map<std::string, std::string>& commonAttributes) {
+  auto flowAttributes = flow->getAttributes();
+  for (const auto& attr : flowAttributes) {
+    if (commonAttributes.find(attr.first) != commonAttributes.end()) {
+      if (commonAttributes[attr.first] != attr.second) {
+        commonAttributes.erase(attr.first);
+        removed_attributes_.push_back(attr.first);
+      }
+    } else if (std::find(removed_attributes_.cbegin(), removed_attributes_.cend(), attr.first) == removed_attributes_.cend()) {
```
hunyadi-dev, line 409: I am not sure I understand why we do this.

extensions/libarchive/MergeContent.cpp
lordgamez: I think throwing would be a better way in this scenario as this should be unreachable because of the allowed values, so it would suggest a bigger problem. Same for the merge strategy, although there is an actual unreachable code there as `mergeFormat_` is checked both on line 258 and 260 so we could never reach the `return false` statement on line 267.

extensions/libarchive/MergeContent.h
```diff
@@ -259,6 +261,42 @@ class ZipMerge: public ArchiveMerge, public MergeBin {
...
+    : session_(session)
+    , flows_(flows)
+    , mergeFlow_(mergeFlow) {}
+  void mergeAttributes();
+  virtual ~AttributeMerger() = default;
+protected:
+  std::map<std::string, std::string> getCommonAttributes();
+  virtual void processFlowFile(const std::shared_ptr<core::FlowFile> &flow, std::map<std::string, std::string>& commonAttributes) = 0;
+
+  std::deque<std::shared_ptr<core::FlowFile>> &flows_;
```
hunyadi-dev, line 276: How about using a simple `core::FlowFile*` pointer here?

hunyadi-dev: Maybe some fancy `gsl_not_null`

extensions/libarchive/MergeContent.cpp
lordgamez: Iterating through 3 flowfiles in order A, B and C, if all 3 have the same attribute but B has a different value for it, then we remove the attribute from the common attributes map. But reaching C we would readd the attribute to the common attributes map because it is not present, as it was removed previously due to the conflicting value. So in this case I check if it was previously removed or not and only add the new attribute if it was not previously seen.

extensions/libarchive/MergeContent.h
```diff
@@ -259,6 +261,42 @@ class ZipMerge: public ArchiveMerge, public MergeBin {
...
+  core::ProcessSession *session_;
+  std::shared_ptr<core::FlowFile> &mergeFlow_;
```
hunyadi-dev, line 278: These could go directly to `mergeAttributes()` as args.

extensions/libarchive/MergeContent.h
```diff
@@ -43,6 +43,8 @@ namespace processors {
 #define MERGE_FORMAT_AVRO_VALUE "Avro"
 #define DELIMITER_STRATEGY_FILENAME "Filename"
 #define DELIMITER_STRATEGY_TEXT "Text"
+#define ATTRIBUTE_STRATEGY_KEEP_COMMON "Keep Only Common Attributes"
+#define ATTRIBUTE_STRATEGY_KEEP_ALL_UNIQUE "Keep All Unique Attributes"
```
hunyadi-dev, line 47: I am not too fond of having these as defines. Maybe they belong under `MergeContent` or `BinFiles` as `static constexpr const char*`-s.

extensions/libarchive/MergeContent.h
lordgamez: Why would be the benefit using a raw pointer over shared_ptr& here?

hunyadi-dev: I take it back, I see that this is actually a reference to the `BinFiles::queue_` that probably owns the flowfiles (it is irrelevant for this PR if it should be shared or not).

As you only take `const &`-s to the elements, you should probably make the member `const &` as well.

extensions/libarchive/MergeContent.cpp
lordgamez: This would require creating a new map for every intersection's output as it seems I cannot update the same map for multiple intersections, that would be costly.

hunyadi-dev: Not necessarily, if you move into the algo:
https://en.cppreference.com/w/cpp/iterator/make_move_iterator

extensions/libarchive/MergeContent.cpp
```diff
@@ -252,36 +255,43 @@ bool MergeContent::processBin(core::ProcessContext *context, core::ProcessSessio
...
+  }
+  session->putAttribute(mergeFlow, BinFiles::FRAGMENT_COUNT_ATTRIBUTE, std::to_string(bin->getSize()));
+
+  if (attributeStrategy_ == ATTRIBUTE_STRATEGY_KEEP_COMMON)
+    KeepOnlymergedAttributesMerger(bin->getFlowFile()).mergeAttributes(session, mergeFlow);
+  else if (attributeStrategy_ == ATTRIBUTE_STRATEGY_KEEP_ALL_UNIQUE)
+    KeepAllUniqueAttributesMerger(bin->getFlowFile()).mergeAttributes(session, mergeFlow);
+  else {
+    logger_->log_error("Attribute strategy not supported %s", attributeStrategy_);
+    throw minifi::Exception(ExceptionType::PROCESSOR_EXCEPTION, "Invalid attribute strategy: " + attributeStrategy_);
```
hunyadi-dev, line 284: I agree with this behaviour, but just in case, can you check, what the NiFi guys do in similar cases?

hunyadi-dev:
Thanks for the modifications, I think [this one](https://github.com/apache/nifi-minifi-cpp/pull/866#discussion_r468394688) is still relevant: 

Current:
```c++
if (mergedAttributes.find(attr.first) != mergedAttributes.end()) {
  if (mergedAttributeIt->second != attr.second) {
    mergedAttributes.erase(attr.first);
    removed_attributes_.push_back(attr.first);
  }
} else if (std::find(removed_attributes_.cbegin(), removed_attributes_.cend(), attr.first) == removed_attributes_.cend()) {
  mergedAttributes[attr.first] = attr.second;
}
```

Suggestion to avoid double map lookup:
```c++
auto mergedAttributeFound = mergedAttributes.find(attr.first);
if (mergedAttributeFound != mergedAttributes.end()) {
  if (mergedAttributeFound->second != attr.second) {
    mergedAttributes.erase(attr.first);
    removed_attributes_.push_back(attr.first);
  }
} else if (std::find(removed_attributes_.cbegin(), removed_attributes_.cend(), attr.first) == removed_attributes_.cend()) {
  mergedAttributeIt->second = attr.second;
}
```

You can further simplify this by checking the removed elements first:
```c++
if(std::find(removed_attributes_.cbegin(), removed_attributes_.cend(), attr.first) != removed_attributes_.cend()) {
  continue;
}
std::map<std::string, std::string>::iterator insertion_res;
bool insertion_happened;
std::tie(insertion_res, insertion_happened) = mergedAttributes.insert(attr);
if(!insertion_happened && insertion_res->second != attr.second) {
  mergedAttributes.erase(insertion_res);
  removed_attributes_.push_back(attr.first);
}
```

Another thing worth potentially improving is looping on `flowAttributes` by move, as we can consume it during the loop.

extensions/libarchive/MergeContent.cpp
```diff
@@ -342,6 +352,54 @@ std::shared_ptr<core::FlowFile> ZipMerge::merge(core::ProcessContext *context, c
...
+      mergedAttributes = flow->getAttributes();
+      isFirst = false;
+    } else {
+      processFlowFile(flow, mergedAttributes);
+    }
+  }
+  return mergedAttributes;
+}
+
+void KeepOnlymergedAttributesMerger::processFlowFile(const std::shared_ptr<core::FlowFile> &flow, std::map<std::string, std::string>& mergedAttributes) {
```
hunyadi-dev, line 375: Minor, but the capitalization of `merged` looks off.

extensions/libarchive/MergeContent.cpp
lordgamez: Are you suggesting something like this?

`auto attr = flow->getAttributes();`
`std::map<std::string, std::string> tmpMerged;`
`std::set_intersection(std::make_move_iterator(mergedAttributes.begin()), std::make_move_iterator(mergedAttributes.end()), std::make_move_iterator(attr.begin()), std::make_move_iterator(attr.end()), std::inserter(tmpMerged, tmpMerged.begin()));`
`mergedAttributes = std::move(tmpMerged);`

Wouldn't it invalidate the original attributes of the original flowfile or it doesn't matter in this case?

lordgamez: Oh now I see we can consume the flow attributes as it's just a copy never mind then.

lordgamez:
> Thanks for the modifications, I think [this one](https://github.com/apache/nifi-minifi-cpp/pull/866#discussion_r468394688) is still relevant:
> 
> Current:
> 
> ```c++
> if (mergedAttributes.find(attr.first) != mergedAttributes.end()) {
>   if (mergedAttributeIt->second != attr.second) {
>     mergedAttributes.erase(attr.first);
>     removed_attributes_.push_back(attr.first);
>   }
> } else if (std::find(removed_attributes_.cbegin(), removed_attributes_.cend(), attr.first) == removed_attributes_.cend()) {
>   mergedAttributes[attr.first] = attr.second;
> }
> ```
> 
> Suggestion to avoid double map lookup:
> 
> ```c++
> auto mergedAttributeFound = mergedAttributes.find(attr.first);
> if (mergedAttributeFound != mergedAttributes.end()) {
>   if (mergedAttributeFound->second != attr.second) {
>     mergedAttributes.erase(attr.first);
>     removed_attributes_.push_back(attr.first);
>   }
> } else if (std::find(removed_attributes_.cbegin(), removed_attributes_.cend(), attr.first) == removed_attributes_.cend()) {
>   mergedAttributeIt->second = attr.second;
> }
> ```
> 
> You can further simplify this by checking the removed elements first:
> 
> ```c++
> if(std::find(removed_attributes_.cbegin(), removed_attributes_.cend(), attr.first) != removed_attributes_.cend()) {
>   continue;
> }
> std::map<std::string, std::string>::iterator insertion_res;
> bool insertion_happened;
> std::tie(insertion_res, insertion_happened) = mergedAttributes.insert(attr);
> if(!insertion_happened && insertion_res->second != attr.second) {
>   mergedAttributes.erase(insertion_res);
>   removed_attributes_.push_back(attr.first);
> }
> ```
> 
> Another thing worth potentially improving is looping on `flowAttributes` by move, as we can consume it during the loop.

Sorry I somehow missed this comment, thanks for the tip I'll update the commit with this.

extensions/libarchive/MergeContent.cpp
hunyadi-dev: Probably the not the most efficient, but very readable:

Assuming 
 1. A := The current common attributes map
 1. B := The current flow-file attribute map
 1. X := Keys(A) ∩ Keys(B)

Then the two kinds of mergers you want is:
 1. **Keep All Unique Attributes:** (Elements of A at X) ∩ (Elements of B at X)
 1. **Keep Only Common Attributes:** A ∩ B

libminifi/test/archive-tests/MergeFileTests.cpp
```diff
@@ -829,3 +829,152 @@ TEST_CASE("MergeFileOnAttribute", "[mergefiletest5]") {
...
+  auto output = testController.output;
+
+  {
+    std::ofstream expectfileFirst(EXPECT_MERGE_CONTENT_FIRST, std::ios::binary);
+
+    // Create and write to the test file
+    for (int i = 0; i < 3; i++) {
+      std::ofstream tmpfile;
+      std::string flowFileName = std::string(FLOW_FILE) + "." + std::to_string(i) + ".txt";
+      tmpfile.open(flowFileName.c_str(), std::ios::binary);
```
szaszm, line 921:
```suggestion
      std::string flowFileName = std::string(FLOW_FILE) + "." + std::to_string(i) + ".txt";
      std::ofstream tmpfile(flowFileName.c_str(), std::ios::binary);
```

extensions/libarchive/MergeContent.cpp
lordgamez: It seems to me that NiFi just crashes on a NullPointerException because when they retrieve the AttributeStrategy object for an unknown strategy, it returns a null and then the getMergedAttributes() is called on that :)

extensions/libarchive/MergeContent.cpp
lordgamez: The only problem in the unique attribute case is after the result of the first intersection we remove the attributes with conflicting values, then in the next processing of the next flowfile attributes the removed attribute can reappear in that flowfile and we handle it as a unique attribute adding it to the final merged attribute map. (More simply: missing the removed attributes handling)

lordgamez: I used the set_intersection for the common attributes, but kept the original solution for the unique attributes

extensions/libarchive/MergeContent.cpp
lordgamez: Good catch, it was a find and replace error :)

extensions/libarchive/MergeContent.h
```diff
@@ -259,9 +259,44 @@ class ZipMerge: public ArchiveMerge, public MergeBin {
...
+  AttributeMerger(std::deque<std::shared_ptr<org::apache::nifi::minifi::core::FlowFile>> &flows)
+    : flows_(flows) {}
+  void mergeAttributes(core::ProcessSession *session, std::shared_ptr<core::FlowFile> &mergeFlow);
+  virtual ~AttributeMerger() = default;
+protected:
+  std::map<std::string, std::string> getMergedAttributes();
+  virtual void processFlowFile(const std::shared_ptr<core::FlowFile> &flow, std::map<std::string, std::string>& mergedAttributes) = 0;
+
+  const std::deque<std::shared_ptr<core::FlowFile>> &flows_;
+};
```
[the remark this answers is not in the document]
lordgamez: I decided to go with the classes as the std::accumulate only supports move from c++20 and it would be unoptimal to create a temporal map of every partial result in case of a large number of flow files.

lordgamez:
Moved to https://github.com/apache/nifi-minifi-cpp/pull/870 due to pull bot incompatibility with PR from main branch.

PR #869, opened 2020-08-12
MINIFICPP-1328 - Improve performance of FileStream

arpadboda:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/src/io/FileStream.cpp
+++ b/libminifi/src/io/FileStream.cpp
@@ -34,19 +34,17 @@ FileStream::FileStream(const std::string &path, bool append)
       path_(path),
       offset_(0) {
   file_stream_ = std::unique_ptr<std::fstream>(new std::fstream());
-  if (append)
+  if (append) {
     file_stream_->open(path.c_str(), std::fstream::in | std::fstream::out | std::fstream::app | std::fstream::binary);
-  else
-    file_stream_->open(path.c_str(), std::fstream::out | std::fstream::binary);
-  file_stream_->seekg(0, file_stream_->end);
-  file_stream_->seekp(0, file_stream_->end);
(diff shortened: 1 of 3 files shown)


libminifi/src/io/FileStream.cpp
```diff
@@ -101,13 +96,12 @@ int FileStream::writeData(std::vector<uint8_t> &buf, int buflen) {
...
 int FileStream::writeData(uint8_t *value, int size) {
   if (!IsNullOrEmpty(value)) {
-    std::lock_guard<std::recursive_mutex> lock(file_lock_);
+    std::lock_guard<std::mutex> lock(file_lock_);
     if (file_stream_->write(reinterpret_cast<const char*>(value), size)) {
       offset_ += size;
       if (offset_ > length_) {
         length_ = offset_;
       }
-      file_stream_->seekg(offset_);
```
arpadboda, line 110: Hope I don't miss anything here due to lack of coffee, but it seems to be that there is no real point in seeking to where we just finished writing.

szaszm: This seeks the input position indicator after write has automatically seeked the output indicator. Yes, they are separate, and no, I don't find this intuitive either.
https://godbolt.org/z/oe1M4j

fgerlits: I guess the `tellg` and `tellp` pointers could in theory be different, but they have never been, and if the standard library changes this, that will break almost all programs in existence, so I don't think they will. :)

szaszm: Yup, I assumed that `stringstream` would be similar to `fstream`, but apparently no. No problem in this case, thanks @fgerlits.

PR #870, opened 2020-08-12
MINIFICPP-1321 Support attribute strategies in MergeContent processor

lordgamez:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/libarchive/MergeContent.cpp
+++ b/extensions/libarchive/MergeContent.cpp
@@ -117,7 +117,7 @@ void MergeContent::onSchedule(core::ProcessContext *context, core::ProcessSessio
   std::string value;
   BinFiles::onSchedule(context, sessionFactory);
   if (context->getProperty(MergeStrategy.getName(), value) && !value.empty()) {
-    mergeStratgey_ = value;
+    mergeStrategy_ = value;
   }
   if (context->getProperty(MergeFormat.getName(), value) && !value.empty()) {
     mergeFormat_ = value;
@@ -126,7 +126,7 @@ void MergeContent::onSchedule(core::ProcessContext *context, core::ProcessSessio
     correlationAttributeName_ = value;
   }
   if (context->getProperty(DelimiterStrategy.getName(), value) && !value.empty()) {
-    delimiterStratgey_ = value;
+    delimiterStrategy_ = value;
   }
   if (context->getProperty(Header.getName(), value) && !value.empty()) {
     header_ = value;
@@ -143,12 +143,12 @@ void MergeContent::onSchedule(core::ProcessContext *context, core::ProcessSessio
   if (context->getProperty(AttributeStrategy.getName(), value) && !value.empty()) {
     attributeStrategy_ = value;
   }
-  if (mergeStratgey_ == MERGE_STRATEGY_DEFRAGMENT) {
+  if (mergeStrategy_ == MERGE_STRATEGY_DEFRAGMENT) {
     binManager_.setFileCount(FRAGMENT_COUNT_ATTRIBUTE);
   }
-  logger_->log_debug("Merge Content: Strategy [%s] Format [%s] Correlation Attribute [%s] Delimiter [%s]", mergeStratgey_, mergeFormat_, correlationAttributeName_, delimiterStratgey_);
+  logger_->log_debug("Merge Content: Strategy [%s] Format [%s] Correlation Attribute [%s] Delimiter [%s]", mergeStrategy_, mergeFormat_, correlationAttributeName_, delimiterStrategy_);
   logger_->log_debug("Merge Content: Footer [%s] Header [%s] Demarcator [%s] KeepPath [%d]", footer_, header_, demarcator_, keepPath_);
-  if (delimiterStratgey_ == DELIMITER_STRATEGY_FILENAME) {
+  if (delimiterStrategy_ == DELIMITER_STRATEGY_FILENAME) {
     if (!header_.empty()) {
       this->headerContent_ = readContent(header_);
     }
@@ -159,7 +159,7 @@ void MergeContent::onSchedule(core::ProcessContext *context, core::ProcessSessio
         this->demarcatorContent_ = readContent(demarcator_);
     }
   }
-  if (delimiterStratgey_ == DELIMITER_STRATEGY_TEXT) {
+  if (delimiterStrategy_ == DELIMITER_STRATEGY_TEXT) {
     this->headerContent_ = header_;
     this->footerContent_ = footer_;
     this->demarcatorContent_ = demarcator_;
@@ -173,7 +173,7 @@ std::string MergeContent::getGroupId(core::ProcessContext *context, std::shared_
     if (flow->getAttribute(correlationAttributeName_, value))
       groupId = value;
   }
-  if (groupId.empty() && mergeStratgey_ == MERGE_STRATEGY_DEFRAGMENT) {
+  if (groupId.empty() && mergeStrategy_ == MERGE_STRATEGY_DEFRAGMENT) {
     if (flow->getAttribute(FRAGMENT_ID_ATTRIBUTE, value))
       groupId = value;
   }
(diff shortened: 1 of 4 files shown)


extensions/libarchive/MergeContent.h
```diff
@@ -259,20 +259,56 @@ class ZipMerge: public ArchiveMerge, public MergeBin {
...
+
+private:
+  std::vector<std::string> removed_attributes_;
+};
+
 // MergeContent Class
 class MergeContent : public processors::BinFiles {
- public:
+public:
+  static constexpr const char *ATTRIBUTE_STRATEGY_KEEP_COMMON = "Keep Only Common Attributes";
```
adamdebreceni, line 297: I'm all for using static members to store the attribute name, but then we should make all of them static members not just the new ones, if moving the others seems problematic, we should reassess leaving them be for now

extensions/libarchive/MergeContent.cpp
```diff
@@ -156,7 +159,7 @@ void MergeContent::onSchedule(core::ProcessContext *context, core::ProcessSessio
         this->demarcatorContent_ = readContent(demarcator_);
     }
   }
-  if (delimiterStratgey_ == DELIMITER_STRATEGY_TEXT) {
+  if (delimiterStrategy_ == DELIMITER_STRATEGY_TEXT) {
     this->headerContent_ = header_;
     this->footerContent_ = footer_;
```
adamdebreceni, line 164: as we picked the "this-elimination" route, we should remove `this->` everywhere in this file

extensions/libarchive/MergeContent.h
szaszm: Instead of members, I would move them to the implementation file into a private namespace if possible.

lordgamez: It's possible, but it would also require moving the ArchiveMerge::WriteCallback's and the MergeContent's constructor's implementation to the cpp file. In that case it might be better to move all the implementation details to the cpp file.

extensions/libarchive/MergeContent.cpp
```diff
@@ -342,6 +352,50 @@ std::shared_ptr<core::FlowFile> ZipMerge::merge(core::ProcessContext *context, c
...
 }
 
+void AttributeMerger::mergeAttributes(core::ProcessSession *session, std::shared_ptr<core::FlowFile> &merge_flow) {
+  for (const auto& pair : getMergedAttributes()) {
+    session->putAttribute(merge_flow, pair.first, pair.second);
+  }
+}
+
+std::map<std::string, std::string> AttributeMerger::getMergedAttributes() {
+  std::map<std::string, std::string> merged_attributes;
```
hunyadi-dev, line 362: Maybe `std::accumulate` could make this simpler:
```c++
if(flows_.empty()) {
  return {}
}
auto merge_attributes = [&] (
    const std::map<std::string, std::string>& merged_attributes,
    const std::shared_ptr<core::FlowFile>& flow)
    { processFlowFile(flow, merged_attributes); }
return std::accumulate(std::next(flows_.cbegin()), flows.cend(), flows_.begin()->getAttributes(), merge_attributes);

File: extensions/libarchive/MergeContent.cpp [hunk 4]
```diff
@@ -342,6 +352,50 @@ std::shared_ptr<core::FlowFile> ZipMerge::merge(core::ProcessContext *context, c
...
+      processFlowFile(flow, merged_attributes);
+    }
+  }
+  return merged_attributes;
+}
+
+void KeepOnlyCommonAttributesMerger::processFlowFile(const std::shared_ptr<core::FlowFile> &flow_file, std::map<std::string, std::string>& merged_attributes) {
+  auto flow_attributes = flow_file->getAttributes();
+  std::map<std::string, std::string> tmp_merged;
+  std::set_intersection(std::make_move_iterator(merged_attributes.begin()), std::make_move_iterator(merged_attributes.end()),
```
hunyadi-dev, line 378: 👍

extensions/libarchive/MergeContent.cpp
lordgamez: I tried this solution before as @szaszm also suggested it, and it was working and was a quite elegant solution, but I dropped it as the [cppreference](https://en.cppreference.com/w/cpp/algorithm/accumulate) suggested that std::move is only used in its implementation from C++20, so copying every temporal map result when there is a large number of flow files may be slow.

extensions/libarchive/MergeContent.h
lordgamez: I left them as ::processor namespace level constants as both tests and different classes use these constants currently.

extensions/libarchive/MergeContent.cpp
hunyadi-dev: You mean this move?
Screenshot 2020-08-13 at 18 23 20
The lambda captures by reference, so it should not be a problem.

The assembly generated is also the same, see here:
https://quick-bench.com/q/J7uVr8A_Ib62WQrJm9QXm7zK5Wo

lordgamez: There was an error in the code you sent, both versions were called with manualLooping() :) But there was no difference in the result after fixing it.

I was thinking about that move, but actually I was thinking referring to the other case, the Keep All Unique Attributes scenario. In that case we currently merge the attributes in place, but using the std::accumulate we cannot do that, and as there is no move support, I cannot consume the parameter, so I have to make a copy of that map. You can check the difference here:
https://quick-bench.com/q/RauiICzYXNNBplQmLFj2bZ7z14E

Btw thanks for this tool, I didn't know about this and it's pretty handy :)

szaszm: I've got this down to near-equal by introducing a pointer-indirection to the map and processing it in place anyway, but this probably defeats the purpose. (Let me join the quick-bench party! :smile: )
https://quick-bench.com/q/7kxa6NSSuo9Dki-J0Mllh3vABeQ

lordgamez: A bit circumvented solution, but I am okay with it. At this point I have no preference one way or another so I can add it in the next commit if there are no objections.

extensions/libarchive/MergeContent.h
```diff
@@ -32,17 +32,19 @@ namespace nifi {
...
+static constexpr const char *MERGE_FORMAT_ZIP_VALUE = "ZIP";
+static constexpr const char *MERGE_FORMAT_FLOWFILE_STREAM_V3_VALUE = "FlowFile Stream, v3";
+static constexpr const char *MERGE_FORMAT_FLOWFILE_STREAM_V2_VALUE = "FlowFile Stream, v2";
+static constexpr const char *MERGE_FORMAT_FLOWFILE_TAR_V1_VALUE = "FlowFile Tar, v1";
+static constexpr const char *MERGE_FORMAT_CONCAT_VALUE = "Binary Concatenation";
+static constexpr const char *MERGE_FORMAT_AVRO_VALUE = "Avro";
+static constexpr const char *DELIMITER_STRATEGY_FILENAME = "Filename";
+static constexpr const char *DELIMITER_STRATEGY_TEXT = "Text";
+static constexpr const char *ATTRIBUTE_STRATEGY_KEEP_COMMON = "Keep Only Common Attributes";
+static constexpr const char *ATTRIBUTE_STRATEGY_KEEP_ALL_UNIQUE = "Keep All Unique Attributes";
```
szaszm, line 47: I recommend either moving these to a subnamespace or moving them back inside the class so that they are not in the scope of all processors.

Also, in namespace scope constant, the `static` keyword is redundant because they have static storage duration and internal linkage by default.

extensions/libarchive/MergeContent.cpp
szaszm: Please don't. This was just a demonstration, but it defeats the purpose of accumulate by circumventing the assignment part. If you prefer STL algorithms, a for_each should be able to do the same.

extensions/libarchive/MergeContent.h
```diff
@@ -32,17 +32,21 @@ namespace nifi {
...
+constexpr const char *MERGE_FORMAT_FLOWFILE_STREAM_V3_VALUE = "FlowFile Stream, v3";
+constexpr const char *MERGE_FORMAT_FLOWFILE_STREAM_V2_VALUE = "FlowFile Stream, v2";
+constexpr const char *MERGE_FORMAT_FLOWFILE_TAR_V1_VALUE = "FlowFile Tar, v1";
+constexpr const char *MERGE_FORMAT_CONCAT_VALUE = "Binary Concatenation";
+constexpr const char *MERGE_FORMAT_AVRO_VALUE = "Avro";
+constexpr const char *DELIMITER_STRATEGY_FILENAME = "Filename";
+constexpr const char *DELIMITER_STRATEGY_TEXT = "Text";
+constexpr const char *ATTRIBUTE_STRATEGY_KEEP_COMMON = "Keep Only Common Attributes";
+constexpr const char *ATTRIBUTE_STRATEGY_KEEP_ALL_UNIQUE = "Keep All Unique Attributes";
+}
```
adamdebreceni, line 49: we could add a comment at the closing bracket about what namespace is being closed, maybe some empty lines before and after the body of the namespace

adamdebreceni (approved):

extensions/libarchive/MergeContent.h
szaszm: https://google.github.io/styleguide/cppguide.html#Namespaces
https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace

szaszm (approved):

arpadboda (approved):

PR #874, opened 2020-08-17
MINIFICPP-1325 - Refactor and test YAML connection parsing

hunyadi-dev:
This is a refactor and test only PR - done as prework for name and UUID usage rework. It is probably simplest to **review this PR on a per commit basis**.

This PR fixes faulty scenarios for parsing configuration values related to connections. 
1. We agreed to fix this one:
https://issues.apache.org/jira/browse/MINIFICPP-1331
1. The others are documented here.


arpadboda:
@hunyadi-dev : quite a lot of CI failures, could you take a look at them before people start reviewing?

extensions/standard-processors/tests/unit/YamlConnectionParserTest.cpp
```diff
@@ -0,0 +1,200 @@
...
+      REQUIRE(1 == relationship_count);
+    }
+  }
+  SECTION("Queue size limits are read") {
+    YAML::Node connection_node = YAML::Load(std::string {
+        "max work queue size: 231\n"
+        "max work queue data size: 12 MB\n" });
+    YamlConnectionParser yaml_connection_parser(connection_node, "test_node", parent_ptr, logger);
+    REQUIRE(231 == yaml_connection_parser.getWorkQueueSizeFromYaml());
+    REQUIRE(12582912 == yaml_connection_parser.getWorkQueueDataSizeFromYaml());  // 12 * 1024 * 1024 B
```
adamdebreceni, line 70: we could move the comment and let the compiler do the multiplication

extensions/standard-processors/tests/unit/YamlConnectionParserTest.cpp
```diff
@@ -0,0 +1,200 @@
...
+    }
+    SECTION("With a configuration that has values of incorrect format") {
+      YAML::Node connection_node = YAML::Load(std::string {
+          "max work queue size: 2 KB\n"
+          "max work queue data size: 10 Incorrect\n"
+          "flowfile expiration: 12\n"
+          "drop empty: sup\n"});
+      YamlConnectionParser yaml_connection_parser(connection_node, "test_node", parent_ptr, logger);
+      CHECK_NOTHROW(yaml_connection_parser.getWorkQueueSizeFromYaml());
+      CHECK_NOTHROW(yaml_connection_parser.getWorkQueueDataSizeFromYaml());
```
adamdebreceni, line 183: why are we verifying incorrect behavior?

hunyadi-dev: We need to be backward compatible with configuration files.

adamdebreceni: I understand we would like to maintain backward compatibility, so would we only throw when the strict parsing is ready?

extensions/standard-processors/tests/unit/YamlConnectionParserTest.cpp
hunyadi-dev: Maybe I am wrong, but I have the gut feeling that this compiles ever-so-slightly quicker. I will change it if you convince me there is no compile-time penalty :D

adamdebreceni: leave it if you can tell me what `64 * 1024 * 1024` is faster than the compiler, it is about having a single source of truth, and in this case we could express in code, what is written in comment (me moving ~~in my chair~~ the cursor during compilation probably has a greater effect on the compilation time than this multiplication)

hunyadi-dev: I do not think you not-liking comments because they can lie is a good excuse for asking this change. Using comments for explaining a calculation is my choice of style (and I gave you the reason, I think it theoretically - albeit rather slowly - might increase compilation time).

Why do you think as the author of this PR, I should follow your favourite way of doing things, when the current proposal is just as expressive? Are you not being a bit arrogant?

libminifi/src/core/yaml/CheckRequiredField.cpp
```diff
@@ -0,0 +1,55 @@
...
+      const YAML::Node name_node = yamlNode->as<YAML::Node>()["name"];
+      // Build a helpful error message for the user so they can fix the
+      // invalid YAML config file, using the component name if present
+      errMsg =
+          name_node ?
+              "Unable to parse configuration file for component named '" + name_node.as<std::string>() + "' as required field '" + fieldName + "' is missing" :
+              "Unable to parse configuration file as required field '" + fieldName + "' is missing";
+      if (!yamlSection.empty()) {
+        errMsg += " [in '" + yamlSection + "' section of configuration file]";
+      }
```
adamdebreceni, line 43: it seems like we could access the line and column number of the `YAML::Node` with the `Mark` method, could this provide even more info to the user?

extensions/standard-processors/tests/unit/YamlConnectionParserTest.cpp
adamdebreceni: > But remember: while comments are very important, the best code is self-documenting.

from [this](https://google.github.io/styleguide/cppguide.html#Comments)

szaszm: Please remember the Apache motto "Community Over Code". (from https://www.apache.org/theapacheway/)
Good code is high priority but keeping a healthy community is higher. We should be open to constructive criticism, but also accept when we cannot reach a consensus and not sacrifice the community morale on the altar of tiny technical disagreements.

My decision making preference is: consensus > healthy debate > majority vote.

szaszm: This is how I solved the same problem some time ago: https://github.com/apache/nifi-minifi-cpp/pull/875/files#diff-42dd8bec1f7b514486d8de629c9a608eR64

Take what you like.

hunyadi-dev: But the code \*is\* self-documenting, the number it shows is obvious. I would definitely change this if I saw any chance that anyone who reads this couldn't immediately understand what the number means from the context.

Just for reference, I asked my girlfriend who has nothing to do with coding what it could mean that we read in 12 MB and get the number 12582912 and she could immediately tell that this is probably the number of bytes.

However, I decided to go further than this, and add an explanation, and my choice of style is a comment for the reason mentioned above. I call this 100% bikeshedding: https://www.wikiwand.com/en/Law_of_triviality

libminifi/src/core/yaml/CheckRequiredField.cpp
hunyadi-dev: Added as requested.

hunyadi-dev:
Closing this PR and opening a new one for the same content rebased.

PR #875, opened 2020-08-19
MINIFICPP-1332 Prevent errneous behavior by stopping FlowController in low disk space conditions

szaszm:
Stop the FlowController if available disk space is <100MB, restart when it goes above 150MB.

___

### For all changes:
     in the commit message?

### For code changes:

### Note:


szaszm:
In case you want to review, these are the problems I'm planning to fix:
- don't use exceptions for control flow
- nuke DiskSpaceWatchdog class to multiple parts because it's doing too much on its own

libminifi/include/utils/OptionalUtils.h
```diff
@@ -37,11 +39,47 @@ optional<typename gsl_lite::remove_cvref<T>::type> optional_from_ptr(T&& obj) {
...
+  if (o.has_value()) {
+    return make_optional(minifiutils::invoke(std::forward<F>(f), *o));
+  } else {
+    return nullopt;
+  }
+}
+
+// bind
+template<typename SourceType, typename F>
+auto operator>>=(const optional<SourceType>& o, F&& f) noexcept(noexcept(minifiutils::invoke(std::forward<F>(f), *o)))
```
adamdebreceni, line 72: I'd rather we use the conventional `map` and `flatMap` names, could we create a custom `optional` class (with `map` and `flatMap`) extending the `nonstd::optional`, and make that available as `utils::optional`?

szaszm: I'd prefer to avoid modifying the optional class. What do you think about this syntax?
```
utils::optional<T> opt{...};
auto map_result = opt | utils::map([](T) -> U {...});
auto flat_map_result = opt | utils::flatMap([](T) -> utils::optional<U>{...});
```

adamdebreceni: yes, that could work

libminifi/src/DiskSpaceWatchdog.cpp
```diff
@@ -0,0 +1,133 @@
...
+  if (restart_bytes < stop_bytes) { throw std::runtime_error{"disk space watchdog stop threshold must be <= restart threshold"}; }
+  constexpr auto mebibytes = 1024 * 1024;
+  return {
+      interval_ms.value_or(chr::seconds{1}),
+      stop_bytes.value_or(15 * mebibytes),
+      restart_bytes.value_or(20 * mebibytes)
+  };
+}
+
+struct callback {
```
lordgamez, line 71: Even for a small scope I think it would be better to use a more distinguished name and capitalized.

libminifi/src/DiskSpaceWatchdog.cpp
```diff
@@ -0,0 +1,133 @@
...
+  Config cfg;
+  gsl::not_null<FlowController*> flow_controller;
+  bool stopped;
```
lordgamez, line 74: Initialize bool

libminifi/test/unit/GeneralUtilsTest.cpp
```diff
@@ -0,0 +1,117 @@
...
+#include <functional>
+#include <string>
+#include <type_traits>
+
+#include "../TestBase.h"
+#include "utils/GeneralUtils.h"
+
+namespace utils = org::apache::nifi::minifi::utils;
+
+static_assert(std::is_same<decltype(utils::make_unique<char16_t>()), std::unique_ptr<char16_t>>::value, "utils::make_unique type is correct");
```
lordgamez, line 28: Shouldn't be the message "type is incorrect"? The message appears if the assertion fails. Same for the static_asserts below.

szaszm: How about "type must be correct"? That would result in something like "assertion failure: ... type must be correct". 

I avoid negative message because I perceive assertion to be a positive check and its message should be about the check, not the error IMHO.

lordgamez: I see, yes that should be okay.

libminifi/src/DiskSpaceWatchdog.cpp
szaszm: It would cause the callback type to no longer be an aggregate in C++11 (fixed in C++14).
https://godbolt.org/z/MGo5cf
https://en.cppreference.com/w/cpp/language/aggregate_initialization

in particular:
> no default member initializers (since C++11) (until C++14)

libminifi/src/DiskSpaceWatchdog.cpp
szaszm: I can't give it a name that any more distinguished than "DiskSpaceWatchdogCallback", but the "DiskSpaceWatchdog" part should be obvious from the context (DiskSpaceWatchdog implementation file). Renaming anyway, because it doesn't hurt. :smile: 

If I misunderstood you in some way, please correct me.

libminifi/src/DiskSpaceWatchdog.cpp
lordgamez: Thanks, I really appreciate the info, I did not know about this C++11 specific restriction.

libminifi/include/utils/file/PathUtils.h
```diff
@@ -47,7 +53,52 @@ std::string getFullPath(const std::string& path);
...
+  std::uintmax_t free;
+  std::uintmax_t available;
+
+  friend bool operator==(const space_info& a, const space_info& b) noexcept {
+    return a.capacity == b.capacity && a.free == b.free && a.available == b.available;
+  }
+};
+
+class filesystem_error : public std::system_error {
+  std::shared_ptr<const std::pair<std::string, std::string>> paths_involved_;
```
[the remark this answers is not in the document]
szaszm: fixed in 13116f0

libminifi/src/DiskSpaceWatchdog.cpp
szaszm: renamed in 13116f0

libminifi/test/unit/GeneralUtilsTest.cpp
szaszm: changed in 13116f0

libminifi/include/utils/OptionalUtils.h
szaszm: fixed in 13116f0 and fdcec88

libminifi/src/DiskSpaceWatchdog.cpp
szaszm: I didn't know it either until I ran into the compilation error. :smile:

libminifi/src/DiskSpaceWatchdog.cpp
lordgamez: Thanks, it's just for making it easier to search for its references.

lordgamez (approved):

libminifi/include/properties/Properties.h
```diff
@@ -82,6 +84,16 @@ class Properties {
    */
   int getInt(const std::string &key, int default_value) const;
 
+  utils::optional<std::string> get(const std::string& key) const {
+    std::string result;
+    const bool found = get(key, result);
```
arpadboda, line 89: Just fancy, for me this is the case when ternary operator simplifies the code:
```
return get(key, result) ? result : utils::nullopt;
```

Feel free to ignore in case you doesn't share this opinion.

libminifi/include/utils/IntervalSwitch.h
```diff
@@ -0,0 +1,80 @@
...
+
+template<typename T, typename Comp = std::less<T>>
+class IntervalSwitch {
+ public:
+  IntervalSwitch(T lower_threshold, T upper_threshold, const IntervalSwitchState initial_state = IntervalSwitchState::UPPER)
+      :lower_threshold_{std::move(lower_threshold)}, upper_threshold_{std::move(upper_threshold)}, state_{initial_state} {
+    gsl_Expects(!less_(upper_threshold_, lower_threshold_));
+  }
+
+  detail::SwitchReturn operator()(const T& value) {
```
arpadboda, line 52: This seems a bit overcomplicated to be with multiple local members and return.

Suggestion:
```
    const auto old_state = state_;
    if (less_(value, lower_threshold_)) {
      state_ = IntervalSwitchState::LOWER;
    } else if (!less_(value, upper_threshold_)) {
      state_ = IntervalSwitchState::UPPER;
    } 
   return {state_, state_ != old_state};
```

Az the "else if" branch was actually doing nothing, inverting the condition could help removing it completely.

libminifi/include/properties/Properties.h
szaszm: I totally agree but they need to be of the same type to be able to use the ternary operator and in this case I rely on conversions from different types to construct the return value.

libminifi/include/utils/IntervalSwitch.h
szaszm: My intention was to write code that describes the intent as much as possible, but you're right, I wrote code that's describing trivialities. I'll change it to be shorter.

libminifi/include/properties/Properties.h
szaszm:
```
/home/szaszm/nifi-minifi-cpp-2/libminifi/include/properties/Properties.h: In member function ‘nonstd::optional_lite::optional<std::__cxx11::basic_string<char> > org::apache::nifi::minifi::Properties::get(const string&) const’:
/home/szaszm/nifi-minifi-cpp-2/libminifi/include/properties/Properties.h:89:29: error: operands to ‘?:’ have different types ‘std::string’ {aka ‘std::__cxx11::basic_string<char>’} and ‘const nonstd::optional_lite::nullopt_t’
   89 |     return get(key, result) ? result : utils::nullopt;
      |            ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
```

libminifi/src/utils/file/PathUtils.cpp
```diff
@@ -84,13 +86,58 @@ std::string PathUtils::getFullPath(const std::string& path) {
...
 
-std::string PathUtils::globToRegex(std::string glob) {
+std::string globToRegex(std::string glob) {
   utils::StringUtils::replaceAll(glob, ".", "\\.");
   utils::StringUtils::replaceAll(glob, "*", ".*");
   utils::StringUtils::replaceAll(glob, "?", ".");
   return glob;
 }
 
+space_info space(const path p, std::error_code& ec) noexcept {
```
lordgamez, line 96: Please rename 'p' to something more searchable as it has a bit larger scope. If it clashes with the type name it is usually recommended to use the '_' suffix in these cases.

lordgamez (approved):
LGTM

libminifi/include/properties/Properties.h
arpadboda: 😞

arpadboda (approved):

libminifi/include/DiskSpaceWatchdog.h
```diff
@@ -0,0 +1,60 @@
...
+namespace disk_space_watchdog {
+struct Config {
+  std::chrono::milliseconds interval;
+  std::uintmax_t stop_threshold_bytes;
+  std::uintmax_t restart_threshold_bytes;
+};
+
+Config read_config(const Configure&);
+
+inline utils::IntervalSwitch<std::uintmax_t> disk_space_interval_switch(Config config) {
```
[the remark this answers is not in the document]
szaszm: It has the size of 3 doubles/pointers on my machine and is trivially copiable, so I thought it's better to copy it than pass by const reference.
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-in
The text says that 2-3 doubles/pointers is cheap to copy, yet the enforcement section suggests flagging anything over `2*sizeof(void*)`. I'm still leaning towards copy, so let me know if you insist on the change given the above.

libminifi/src/utils/file/PathUtils.cpp
szaszm: I'm not a fan of trailing underscores unless really necessary, so I went with `path` and let them clash. Let me know if you're fine with this approach.

libminifi/include/DiskSpaceWatchdog.h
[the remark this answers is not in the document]
lordgamez: You can leave it as it is, I do not insist on it if it's still cheap to copy. My only concern was if this config may be expanded in the future, but I suppose the watchdog configuration will not really change.

libminifi/src/utils/file/PathUtils.cpp
lordgamez: I'm fine with that as well.

libminifi/include/DiskSpaceWatchdog.h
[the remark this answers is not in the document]
adamdebreceni: I would be really surprised if it made any difference performance-wise, this function is most probably going to get inlined either way with the optimization level we are using

libminifi/include/utils/GeneralUtils.h
```diff
@@ -43,6 +43,7 @@ using std::make_unique;
 
 template<typename T, typename = typename std::enable_if<std::is_integral<T>::value>::type>
 T intdiv_ceil(T numerator, T denominator) {
+  gsl_Expects(denominator != 0);
   // note: division and remainder is 1 instruction on x86
   return numerator / denominator + (numerator % denominator > 0);
```
adamdebreceni, line 48: I know this is old code, but unless we restrict the domain of this function to non-negative arguments, it might give incorrect results, e.g. `intdiv_ceil(-5, -3) == 1`

PR #877, opened 2020-08-24
MINIFICPP-1339 - Reopen rocksdb database when space becomes available on the disk

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/rocksdb-repos/FlowFileRepository.cpp
+++ b/extensions/rocksdb-repos/FlowFileRepository.cpp
@@ -37,6 +37,10 @@ namespace core {
 namespace repository {
 
 void FlowFileRepository::flush() {
+  auto opendb = db_->open();
+  if (!opendb) {
+    return;
+  }
   rocksdb::WriteBatch batch;
   rocksdb::ReadOptions options;
 
@@ -54,7 +58,7 @@ void FlowFileRepository::flush() {
     }
   }
 
-  auto multistatus = db_->MultiGet(options, keys, &values);
+  auto multistatus = opendb->MultiGet(options, keys, &values);
 
   for(size_t i=0; i<keys.size() && i<values.size() && i<multistatus.size(); ++i) {
     if(!multistatus[i].ok()) {
@@ -71,7 +75,7 @@ void FlowFileRepository::flush() {
     batch.Delete(keys[i]);
   }
 
-  auto operation = [this, &batch]() { return db_->Write(rocksdb::WriteOptions(), &batch); };
+  auto operation = [this, &batch, &opendb]() { return opendb->Write(rocksdb::WriteOptions(), &batch); };
 
   if (!ExecuteWithRetry(operation)) {
     for (const auto& key: keystrings) {
@@ -91,14 +95,18 @@ void FlowFileRepository::flush() {
 }
 
 void FlowFileRepository::printStats() {
+  auto opendb = db_->open();
+  if (!opendb) {
+    return;
+  }
   std::string key_count;
-  db_->GetProperty("rocksdb.estimate-num-keys", &key_count);
+  opendb->GetProperty("rocksdb.estimate-num-keys", &key_count);
 
   std::string table_readers;
-  db_->GetProperty("rocksdb.estimate-table-readers-mem", &table_readers);
+  opendb->GetProperty("rocksdb.estimate-table-readers-mem", &table_readers);
 
   std::string all_memtables;
-  db_->GetProperty("rocksdb.cur-size-all-mem-tables", &all_memtables);
+  opendb->GetProperty("rocksdb.cur-size-all-mem-tables", &all_memtables);
 
   logger_->log_info("Repository stats: key count: %s, table readers size: %s, all memory tables size: %s",
       key_count, table_readers, all_memtables);
@@ -122,9 +130,9 @@ void FlowFileRepository::run() {
 }
 
 void FlowFileRepository::prune_stored_flowfiles() {
+  utils::optional<minifi::internal::OpenRocksDB> opendb;
   rocksdb::DB* used_database;
   std::unique_ptr<rocksdb::DB> stored_database;
-  bool corrupt_checkpoint = false;
   if (nullptr != checkpoint_) {
     rocksdb::Options options;
     options.create_if_missing = true;
@@ -134,14 +142,18 @@ void FlowFileRepository::prune_stored_flowfiles() {
     if (status.ok()) {
       stored_database.reset(used_database);
     } else {
-      used_database = db_;
+      opendb = db_->open();
+      if (!opendb) {
+        return;
+      }
+      used_database = opendb->get();
     }
   } else {
     logger_->log_trace("Could not open checkpoint as object doesn't exist. Likely not needed or file system error.");
     return;
   }
 
-  rocksdb::Iterator* it = used_database->NewIterator(rocksdb::ReadOptions());
+  auto it = std::unique_ptr<rocksdb::Iterator>(used_database->NewIterator(rocksdb::ReadOptions()));
   for (it->SeekToFirst(); it->Valid(); it->Next()) {
     std::shared_ptr<FlowFileRecord> eventRead = std::make_shared<FlowFileRecord>(shared_from_this(), content_repo_);
     std::string key = it->key().ToString();
@@ -155,7 +167,7 @@ void FlowFileRepository::prune_stored_flowfiles() {
         search = connectionMap.find(eventRead->getConnectionUuid());
         found = (search != connectionMap.end());
       }
-      if (!corrupt_checkpoint && found) {
+      if (found) {
         // we find the connection for the persistent flowfile, create the flowfile and enqueue that
         std::shared_ptr<core::FlowFile> flow_file_ref = std::static_pointer_cast<core::FlowFile>(eventRead);
         eventRead->setStoredToRepository(true);
@@ -171,8 +183,6 @@ void FlowFileRepository::prune_stored_flowfiles() {
       keys_to_delete.enqueue(key);
     }
   }
-
-  delete it;
 }
 
 bool FlowFileRepository::ExecuteWithRetry(std::function<rocksdb::Status()> operation) {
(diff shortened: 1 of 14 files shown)


extensions/rocksdb-repos/RocksDatabase.h
```diff
@@ -0,0 +1,98 @@
...
+
+  rocksdb::Status Write(const rocksdb::WriteOptions& options, rocksdb::WriteBatch* updates);
+
+  rocksdb::Status Delete(const rocksdb::WriteOptions& options, const rocksdb::Slice& key);
+
+  rocksdb::Status Merge(const rocksdb::WriteOptions& options, const rocksdb::Slice& key, const rocksdb::Slice& value);
+
+  bool GetProperty(const rocksdb::Slice& property, std::string* value);
+
+  rocksdb::Iterator* NewIterator(const rocksdb::ReadOptions& options);
```
fgerlits, line 58: this could return a `unique_ptr`, as we always wrap the result into one

extensions/rocksdb-repos/RocksDbStream.h
```diff
@@ -160,7 +160,7 @@ class RocksDbStream : public io::BaseStream {
 
   std::string value_;
 
-  rocksdb::DB *db_;
+  gsl::not_null<minifi::internal::RocksDatabase*> db_;
```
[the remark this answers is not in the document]
adamdebreceni: I was under the impression that taking a reference indicates that we are not planning on storing it, is this not a generally recognized convention?

extensions/rocksdb-repos/RocksDatabase.h
adamdebreceni: done

extensions/rocksdb-repos/FlowFileRepository.cpp
```diff
@@ -71,7 +75,7 @@ void FlowFileRepository::flush() {
     batch.Delete(keys[i]);
   }
 
-  auto operation = [this, &batch]() { return db_->Write(rocksdb::WriteOptions(), &batch); };
+  auto operation = [this, &batch, &opendb]() { return opendb->Write(rocksdb::WriteOptions(), &batch); };
```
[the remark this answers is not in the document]
adamdebreceni: done

extensions/rocksdb-repos/RocksDatabase.h
```diff
@@ -0,0 +1,98 @@
...
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace internal {
+
+class RocksDatabase;
+
+// Not thread safe
+class OpenRocksDB {
```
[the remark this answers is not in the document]
adamdebreceni: done

extensions/rocksdb-repos/RocksDbStream.h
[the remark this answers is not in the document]
fgerlits: I was not aware of this convention; to me, a reference member variable sounds like a clear expression of "this object depends on `db_` being non-null and should not outlive it".  I would be interested to hear what others think.

extensions/rocksdb-repos/FlowFileRepository.cpp
```diff
@@ -134,14 +142,18 @@ void FlowFileRepository::prune_stored_flowfiles() {
     if (status.ok()) {
```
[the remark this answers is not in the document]
adamdebreceni: made it possible to create a `RocksDatabase` in a `ReadOnly` mode, so `::open` acts accordingly, this cleared up the three different DB variables

extensions/rocksdb-repos/RocksDatabase.cpp
```diff
@@ -0,0 +1,138 @@
...
+  impl_.reset();
+}
+
+utils::optional<OpenRocksDB> RocksDatabase::open() {
+  std::lock_guard<std::mutex> db_guard{ mtx_ };
+  if (!impl_) {
+    // database is not opened yet
+    rocksdb::DB* db_instance = nullptr;
+    rocksdb::Status result;
+    if (mode_ == Mode::ReadWrite) {
```
fgerlits, line 117: I would use a `switch` so that we'll get a compiler warning if a new enum value is added and we forget to handle it.  But I realize that both of these events are unlikely, so I don't insist. :)

extensions/rocksdb-repos/FlowFileRepository.cpp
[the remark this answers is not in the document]
fgerlits: very nice, thank you!

extensions/rocksdb-repos/RocksDbStream.h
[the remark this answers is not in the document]
szaszm: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#note-392 Last sentence of "Exceptions":
> Note that using a reference member is almost always wrong.
Sadly there is no explanation and I couldn't find one elsewhere either.

I was not aware of the convention either, that taking a reference indicates not storing it. Do you have any resources explaining it? My disk space watchdog PR is affected because it takes ctor args by reference just to store them in a `gsl::not_null<T*>` member.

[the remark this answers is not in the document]
szaszm: also related: https://abseil.io/tips/177

extensions/rocksdb-repos/RocksDatabase.cpp
adamdebreceni: could we promote such a warning to an error at the file/function/block level?

extensions/rocksdb-repos/RocksDbStream.h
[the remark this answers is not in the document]
adamdebreceni: I couldn't find a convincing argument either, only a gut feeling that having `U f(T& t)` and `U g(T* t)`, after calling `g` I'm more reluctant to discard the `t` I provided than after calling `f`, I feel like if I write `f` I'm signaling that I only need the object for the duration of the function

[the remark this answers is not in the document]
adamdebreceni: but since nobody seems to share this feeling (here or the "internet") I will modify the constructor to take a reference 👍

extensions/rocksdb-repos/RocksDatabase.cpp
```diff
@@ -0,0 +1,139 @@
...
+  return result;
+}
+
+std::vector<rocksdb::Status> OpenRocksDB::MultiGet(const rocksdb::ReadOptions& options, const std::vector<rocksdb::Slice>& keys, std::vector<std::string>* values) {
+  std::vector<rocksdb::Status> results = impl_->MultiGet(options, keys, values);
+  for (const auto& result : results) {
+    if (result == rocksdb::Status::NoSpace()) {
+      db_->invalidate();
+    }
+  }
```
lordgamez, line 51: Could the db be invalidated on the first bad result only?

```suggestion
    if (std::any_of(results.begin(), results.end(), [](const rocksdb::Status& status){ return status == rocksdb::Status::NoSpace();})
      db_->invalidate();
```

adamdebreceni: nice catch (although `invalidate` is idempotent), I just added a `break`, because the proposed solution is too verbose for my liking, I dream of a day when we can do something like `results.some(_ == rocksdb::Status::NoSpace())` but that will probably come in `C++never`

extensions/rocksdb-repos/RocksDatabase.cpp
adamdebreceni: 🤔 I think we should enable that error, although not in this PR

extensions/rocksdb-repos/RocksDatabase.cpp
fgerlits: I think 
```c++
auto is_nospace = [](auto status){ return status == rocksdb::Status::NoSpace(); };
if (std::ranges::any_of(results, is_nospace) { ... }
```
is OK, and it is reality in C++20, so it will be available in minificpp in about ten years. :)

fgerlits (approved):

lordgamez (approved):

extensions/rocksdb-repos/FlowFileRepository.cpp
```diff
@@ -37,6 +37,10 @@ namespace core {
 namespace repository {
 
 void FlowFileRepository::flush() {
+  auto opendb = db_->open();
+  if (!opendb) {
+    return;
```
adam-markovics, line 42: Is this okay, shouldn't we log or throw?

extensions/rocksdb-repos/FlowFileRepository.cpp
```diff
@@ -194,22 +201,27 @@ bool FlowFileRepository::ExecuteWithRetry(std::function<rocksdb::Status()> opera
  * Returns True if there is data to interrogate.
  * @return true if our db has data stored.
  */
-bool FlowFileRepository::need_checkpoint(){
-  std::unique_ptr<rocksdb::Iterator> it = std::unique_ptr<rocksdb::Iterator>(db_->NewIterator(rocksdb::ReadOptions()));
+bool FlowFileRepository::need_checkpoint(minifi::internal::OpenRocksDB& opendb){
+  auto it = opendb.NewIterator(rocksdb::ReadOptions());
   for (it->SeekToFirst(); it->Valid(); it->Next()) {
```
adam-markovics, line 206: Why is this a loop if it returns the first time?

extensions/rocksdb-repos/FlowFileRepository.cpp
```diff
@@ -91,14 +95,18 @@ void FlowFileRepository::flush() {
 }
 
 void FlowFileRepository::printStats() {
+  auto opendb = db_->open();
+  if (!opendb) {
+    return;
```
adam-markovics, line 100: Is this okay, shouldn't we log or throw?

extensions/rocksdb-repos/FlowFileRepository.cpp
adamdebreceni: an unsuccessful `open` now logs an error

extensions/rocksdb-repos/FlowFileRepository.cpp
adamdebreceni: logging error on failed `open`

adam-markovics (approved):

arpadboda (approved):

PR #885, opened 2020-08-31
MINIFICPP-1344 - Correct rest endpoint for buckets on updateFromPayload

hunyadi-dev:
I tried to predict what kind of URL the function meant to parse and changed the implementation based on that. The only important change in logic is that now the function looks for "buckets" in the REST endpoint path instead of "bucket".

### DIFF
--- a/libminifi/src/core/FlowConfiguration.cpp
+++ b/libminifi/src/core/FlowConfiguration.cpp
@@ -67,33 +67,23 @@ std::shared_ptr<core::Processor> FlowConfiguration::createProvenanceReportTask()
   return processor;
 }
 
-std::unique_ptr<core::ProcessGroup> FlowConfiguration::updateFromPayload(const std::string &source, const std::string &yamlConfigPayload) {
+std::unique_ptr<core::ProcessGroup> FlowConfiguration::updateFromPayload(const std::string& url, const std::string& yamlConfigPayload) {
   auto old_services = controller_services_;
   auto old_provider = service_provider_;
   controller_services_ = std::make_shared<core::controller::ControllerServiceMap>();
   service_provider_ = std::make_shared<core::controller::StandardControllerServiceProvider>(controller_services_, nullptr, configuration_);
   auto payload = getRootFromPayload(yamlConfigPayload);
-  if (!source.empty() && payload != nullptr) {
-    std::string host, protocol, path, query, url = source;
-    int port = -1;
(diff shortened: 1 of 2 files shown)


libminifi/src/core/FlowConfiguration.cpp
```diff
@@ -67,33 +67,24 @@ std::shared_ptr<core::Processor> FlowConfiguration::createProvenanceReportTask()
...
-        }
-      }
-
-      if (str == "bucket") {
-        if (i + 1 < path_split.size()) {
-          bucket_id = path_split.at(i + 1);
-          i++;
-        }
+    auto path_split = utils::StringUtils::split(url, "/");
+    // This function might not do what the original implementater expected from it (https://issues.apache.org/jira/browse/MINIFICPP-1344)
```
fgerlits, line 79: I don't understand this comment; I think it can be deleted.  (Just line 79; lines 80-81 are useful.)

libminifi/src/core/FlowConfiguration.cpp
```diff
@@ -67,33 +67,24 @@ std::shared_ptr<core::Processor> FlowConfiguration::createProvenanceReportTask()
...
+    // Registry API docs: nifi.apache.org/docs/nifi-registry-docs/rest-api/index.html
+    // GET /buckets/{bucketId}/flows/{flowId}: Gets a flow
+    const auto bucket_token_found = std::find(path_split.cbegin(), path_split.cend(), "buckets");
+    if (bucket_token_found != path_split.cend() && std::next(bucket_token_found) != path_split.cend()) {
+      bucket_id = *std::next(bucket_token_found);
+      const auto flows_token_found = std::find(std::next(bucket_token_found, 2), path_split.cend(), "flows");
```
fgerlits, line 85: Do we need a `find` here?  This will accept things like `.../buckets/JACKnJILLs/colors/RED/animals/ANT/flows/DANUBE` -- do we want to allow this?

hunyadi-dev: The original implementation was even less strict, looking for the token anywhere. It is a good question how we want to treat the example you listed, to me it seems like it clearly has a `bucket` and a `flow` we can extract.

libminifi/src/core/FlowConfiguration.cpp
hunyadi-dev: Fixed.

fgerlits (approved):

szaszm (approved):

PR #886, opened 2020-08-31
MINIFICPP-1323 Encrypt sensitive properties

fgerlits:
https://issues.apache.org/jira/browse/MINIFICPP-1323

Encrypt sensitive properties in the `minifi.properties` file.  The change has two parts:
1. implement a new command-line tool called `encrypt-config` which can encrypt the sensitive properties
1. decrypt the sensitive properties during startup.

Please also review the wiki page https://cwiki.apache.org/confluence/display/MINIFI/Encrypt+sensitive+configuration+properties as part of this pull request.

---

### For all changes:
     in the commit message?

### For code changes:

### Note:


adamdebreceni:
cool stuff, as I understand it only aims to protect the user when they (accidentally or intentionally) expose the `minifi.properties` file

fgerlits:
> cool stuff, as I understand it only aims to protect the user when they (accidentally or intentionally) expose the `minifi.properties` file

Yes, that is true.  The idea is that
1. it is enough to protect `bootstrap.conf` and you can leave `minifi.properties` world-readable, so non-entitled users can see the value of most settings except for these sensitive ones; and
2. if you want to find out the value of these sensitive properties, you should have to do some work; you can't see them by accident.

hunyadi-dev:
From the documentation:

Screenshot 2020-09-01 at 10 36 56

Are we sure that having special characters like `=` in the properties file can not hurt us?

encrypt-config/ConfigFile.cpp
```diff
@@ -0,0 +1,195 @@
...
+}
+
+ConfigLine::ConfigLine(const std::string& key, const std::string& value)
+  : line_{key + "=" + value}, key_{key}, value_{value} {
+}
+
+void ConfigLine::updateValue(const std::string& value) {
+  auto pos = line_.find('=');
+  if (pos != std::string::npos) {
+    line_.replace(pos + 1, std::string::npos, value);
```
hunyadi-dev, line 57: Would this not be past-end if one would want to assign an empty-string as a value?

encrypt-config/ConfigFile.cpp
```diff
@@ -0,0 +1,195 @@
...
+ConfigFile::ConfigFile(const std::string& file_path) {
+  std::ifstream file{file_path};
+  std::string line;
+  while (std::getline(file, line)) {
+    config_lines_.push_back(ConfigLine{line});
+  }
+}
+
+ConfigFile::Lines::const_iterator ConfigFile::findKey(const std::string& key) const {
+  return std::find_if(config_lines_.begin(), config_lines_.end(), [&key](const ConfigLine& config_line) {
```
hunyadi-dev, line 73: Minor, but could be `cbegin()` and `cend()`.

encrypt-config/ConfigFile.cpp
```diff
@@ -0,0 +1,195 @@
...
+}
+
+ConfigFile::Lines::iterator ConfigFile::findKey(const std::string& key) {
+  return std::find_if(config_lines_.begin(), config_lines_.end(), [&key](const ConfigLine& config_line) {
+    return config_line.key_ == key;
+  });
+}
+
+utils::optional<std::string> ConfigFile::getValue(const std::string& key) const {
+  auto it = findKey(key);
```
hunyadi-dev, line 85: Could be `const`.

encrypt-config/ConfigFile.cpp
```diff
@@ -0,0 +1,195 @@
...
+  if (it != config_lines_.end()) {
+    return it->value_;
+  } else {
+    return utils::nullopt;
+  }
+}
+
+void ConfigFile::update(const std::string& key, const std::string& value) {
+  const auto it = findKey(key);
```
hunyadi-dev, line 94: This though does not seem like a const iterator.

encrypt-config/ConfigFile.cpp
fgerlits: It would, but that is legal.  If `pos + 1 == line_.length()`, then the `replace` becomes an `append`.  If `pos + 1 > line_.length()`, that would be a problem, but I don't think that can happen.  I have added unit tests for this function.

encrypt-config/ConfigFile.cpp
fgerlits: done

encrypt-config/ConfigFile.cpp
fgerlits: done

encrypt-config/ConfigFile.cpp
fgerlits: Yeah, it's a const variable of non-const iterator type -- I find that confusing, too. :)  I have removed the `const`.

encrypt-config/ConfigFile.cpp
```diff
@@ -0,0 +1,195 @@
...
+void ConfigFile::append(const std::string& key, const std::string& value) {
+  config_lines_.emplace_back(key, value);
+}
+
+int ConfigFile::erase(const std::string& key) {
+  auto has_this_key = [&key](const ConfigLine& line) { return line.key_ == key; };
+  auto new_end = std::remove_if(config_lines_.begin(), config_lines_.end(), has_this_key);
+  auto num_removed = std::distance(new_end, config_lines_.end());
+  config_lines_.erase(new_end, config_lines_.end());
+  return gsl::narrow<int>(num_removed);
```
hunyadi-dev, line 121: Why not simply returning `std::size_t`?

fgerlits: It isn't `size_t`, it is a signed type: `Lines::difference_type`, which is too long, or (probably) `std::ptrdiff_t`, which is misleading -- `int` just seemed simpler.

If you have erased more than INT_MAX lines from the config file, you should get an exception, as something is surely wrong.

encrypt-config/ConfigFile.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+  return num_properties_encrypted;
+}
+
+std::vector<std::string> ConfigFile::getSensitiveProperties() const {
+  std::vector<std::string> sensitive_properties(DEFAULT_SENSITIVE_PROPERTIES.begin(), DEFAULT_SENSITIVE_PROPERTIES.end());
+  utils::optional<std::string> additional_sensitive_props_list = getValue(ADDITIONAL_SENSITIVE_PROPS_PROPERTY_NAME);
+  if (additional_sensitive_props_list) {
+    std::vector<std::string> additional_sensitive_properties = utils::StringUtils::split(*additional_sensitive_props_list, ",");
+    sensitive_properties = mergeProperties(sensitive_properties, additional_sensitive_properties);
```
hunyadi-dev, line 166: additional_sensitive_properties could be moved.

encrypt-config/ConfigFile.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+  return sensitive_properties;
+}
+
+std::vector<std::string> ConfigFile::mergeProperties(std::vector<std::string> properties,
+                                                     const std::vector<std::string>& additional_properties) {
+  for (const auto& property_name : additional_properties) {
+    std::string property_name_trimmed = utils::StringUtils::trim(property_name);
+    if (!property_name_trimmed.empty()) {
+      properties.push_back(property_name_trimmed);
```
hunyadi-dev, line 181: Could be move instead of a copy.

encrypt-config/ConfigFile.cpp
hunyadi-dev, line 166: `mergeProperties` currently takes sensitive_properties by copy. You can move it in, or capture by reference and modify in-place.

hunyadi-dev (approved):
Looks good and very well tested.

encrypt-config/ConfigFile.cpp
fgerlits: I don't want to clutter the code with too many `std::move()`s and `&&`s which are not important semantically, only might make the the code run slightly faster.  Note that is in a command-line tool which already runs in less than 0.1 second, and will only be used a few times a year, at most.

encrypt-config/ConfigFile.cpp
```diff
@@ -0,0 +1,195 @@
...
+std::vector<std::string> ConfigFile::getSensitiveProperties() const {
+  std::vector<std::string> sensitive_properties(DEFAULT_SENSITIVE_PROPERTIES.begin(), DEFAULT_SENSITIVE_PROPERTIES.end());
+  utils::optional<std::string> additional_sensitive_props_list = getValue(ADDITIONAL_SENSITIVE_PROPS_PROPERTY_NAME);
+  if (additional_sensitive_props_list) {
+    std::vector<std::string> additional_sensitive_properties = utils::StringUtils::split(*additional_sensitive_props_list, ",");
+    sensitive_properties = mergeProperties(sensitive_properties, additional_sensitive_properties);
+  }
+
+  auto not_found = [this](const std::string& property_name) { return !getValue(property_name); };
+  auto new_end = std::remove_if(sensitive_properties.begin(), sensitive_properties.end(), not_found);
```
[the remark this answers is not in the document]
fgerlits: done

encrypt-config/ConfigFile.cpp
fgerlits: done

encrypt-config/ConfigFile.cpp
fgerlits: see above; I think this would mean more code for a tiny performance improvement which we don't need

encrypt-config/ConfigFile.cpp
```diff
@@ -0,0 +1,195 @@
...
+      ++num_properties_encrypted;
+    }
+  }
+
+  return num_properties_encrypted;
+}
+
+std::vector<std::string> ConfigFile::getSensitiveProperties() const {
+  std::vector<std::string> sensitive_properties(DEFAULT_SENSITIVE_PROPERTIES.begin(), DEFAULT_SENSITIVE_PROPERTIES.end());
+  utils::optional<std::string> additional_sensitive_props_list = getValue(ADDITIONAL_SENSITIVE_PROPS_PROPERTY_NAME);
```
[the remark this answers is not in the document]
fgerlits: done

encrypt-config/EncryptConfig.cpp
```diff
@@ -0,0 +1,146 @@
...
+  }
+}
+
+std::string EncryptConfig::base64DecodeAndValidateKey(const std::string& key) const {
+  std::string base64_decoded_key = utils::StringUtils::from_base64(key);
+  if (base64_decoded_key.size() == ENCRYPTION_KEY_SIZE) {
+    return base64_decoded_key;
+  } else {
+    std::stringstream error;
+    error << "The encryption key\n    " << ENCRYPTION_KEY_PROPERTY_NAME << '=' << key << '\n'
```
alopresto, line 105: I don't think the key value should be printed here.

libminifi/src/Configure.cpp
```diff
@@ -68,6 +74,37 @@ const char *Configure::nifi_state_management_provider_local = "nifi.state.manage
...
+    const std::string& property_key = property.first;
+    const std::string& property_value = property.second;
+
+    utils::optional<std::string> encryption_type = get(property_key + ".protected");
+    if (Decryptor::isEncrypted(encryption_type)) {
+      std::string decrypted_property_value;
+      try {
+        decrypted_property_value = decryptor.decrypt(utils::StringUtils::from_base64(property_value), property_key);
+      } catch (const std::exception& ex) {
+        logger_->log_error("Could not decrypt property %s=%s; error: %s", property_key, property_value, ex.what());
```
alopresto, line 95: If the encryption type isn't recognized, this could leak the "encrypted" value (which may be plaintext).

encrypt-config/EncryptConfig.cpp
fgerlits: I have removed it.

libminifi/src/Configure.cpp
fgerlits: I have removed the `property_value` from the error message.

encrypt-config/ConfigFile.h
```diff
@@ -0,0 +1,85 @@
...
+  std::vector<std::string> getSensitiveProperties() const;
+  static std::vector<std::string> mergeProperties(std::vector<std::string> properties,
+                                                  const std::vector<std::string>& additional_properties);
+
+  Lines config_lines_;
+};
+
+inline bool operator==(const ConfigFile& left, const ConfigFile& right) {
+  return left.config_lines_ == right.config_lines_;
+}
```
szaszm, line 79: Both equality comparisons should have an `operator!=` counterpart, preferably implemented in terms of `operator==`.

encrypt-config/ConfigFile.h
```diff
@@ -0,0 +1,85 @@
...
+class ConfigLine {
+public:
+  ConfigLine(std::string line);
+  ConfigLine(const std::string& key, const std::string& value);
+
+  void updateValue(const std::string& value);
+
+  std::string line_;
+  std::string key_;
+  std::string value_;
```
szaszm, line 40: These should be private because there is an invariant to be enforced. https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-struct

Alternatively, we should not keep the original line, only the key and value. In this case, the class essentially becomes `std::pair<std::string, std::string>`.

```suggestion
  std::string key() const { return key_; }
  std::string value() const { return value_; }
 private:
  std::string line_;
  std::string key_;
  std::string value_;
```

fgerlits: We need the original line, because I do not want to reformat the line unless we modify it.  For example, if there are no properties to be encrypted, `encrypt-config` should be a no-op.

szaszm: Could you add a comment about this design decision? It will help future readers.

encrypt-config/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+cmake_minimum_required(VERSION 3.7)
+project(encrypt-config)
```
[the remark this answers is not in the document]
fgerlits: I have removed these 3 lines

encrypt-config/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+set(VERSION, "0.1")
+
+file(GLOB ENCRYPT_CONFIG_FILES  "*.cpp")
+add_executable(encrypt-config "${ENCRYPT_CONFIG_FILES}")
+include_directories(../libminifi/include  ../thirdparty/cxxopts/include)
```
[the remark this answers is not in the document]
fgerlits: If I change it to `target_include_directories`, I get an error message which says

```
Determining if the pthread_create exist failed with the following output:
Change Dir: /home/fgerlits/src/minifi/build/CMakeFiles/CMakeTmp

Run Build Command:"/usr/bin/make" "cmTC_c8414/fast"
/usr/bin/make -f CMakeFiles/cmTC_c8414.dir/build.make CMakeFiles/cmTC_c8414.dir/build
make[1]: Entering directory '/home/fgerlits/src/minifi/build/CMakeFiles/CMakeTmp'
Building C object CMakeFiles/cmTC_c8414.dir/CheckSymbolExists.c.o
/usr/bin/cc    -o CMakeFiles/cmTC_c8414.dir/CheckSymbolExists.c.o   -c /home/fgerlits/src/minifi/build/CMakeFiles/CMakeTmp/CheckSymbolExists.c
Linking C executable cmTC_c8414
/usr/bin/cmake -E cmake_link_script CMakeFiles/cmTC_c8414.dir/link.txt --verbose=1
/usr/bin/cc      -rdynamic CMakeFiles/cmTC_c8414.dir/CheckSymbolExists.c.o  -o cmTC_c8414 
CMakeFiles/cmTC_c8414.dir/CheckSymbolExists.c.o: In function `main':
CheckSymbolExists.c:(.text+0x1b): undefined reference to `pthread_create'
collect2: error: ld returned 1 exit status
CMakeFiles/cmTC_c8414.dir/build.make:97: recipe for target 'cmTC_c8414' failed
make[1]: *** [cmTC_c8414] Error 1
make[1]: Leaving directory '/home/fgerlits/src/minifi/build/CMakeFiles/CMakeTmp'
Makefile:126: recipe for target 'cmTC_c8414/fast' failed
make: *** [cmTC_c8414/fast] Error 2

File /home/fgerlits/src/minifi/build/CMakeFiles/CMakeTmp/CheckSymbolExists.c:
/* */
#include <pthread.h>

int main(int argc, char** argv)
{
  (void)argv;
#ifndef pthread_create
  return ((int*)(&pthread_create))[argc];
#else
  (void)argc;
  return 0;
#endif
}
```

If I simply delete this line, then the includes are missing from the compile options.  I have tried various ways of telling CMake that `encrypt-config` depends on `minifi` and `cxxopts` and should inherit their include and link flags, but none of them worked.

I agree that the CMake build structure should be cleared up, but I suggest we do that separately from this pull request.

encrypt-config/tests/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+
+file(GLOB ENCRYPT_CONFIG_TESTS  "*.cpp")
+
+file(GLOB ENCRYPT_CONFIG_SOURCES  "${CMAKE_SOURCE_DIR}/encrypt-config/*.cpp")
```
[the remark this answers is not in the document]
fgerlits: Can you suggest an alternative?  I don't want to list each file, because then we would need to modify the CMakeLists.txt file every time we add a new .cpp file, which is easy to forget.

Note that globs like this are used all over minifi.  If we want to clear it up, I think we should do it in a separate pull request.

encrypt-config/tests/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+set(ENCRYPT_CONFIG_TEST_COUNT 0)
+foreach(testfile ${ENCRYPT_CONFIG_TESTS})
+  get_filename_component(testfilename "${testfile}" NAME_WE)
+  add_executable(${testfilename} "${testfile}" ${ENCRYPT_CONFIG_SOURCES})
+
+  target_include_directories(${testfilename} BEFORE PRIVATE "${CMAKE_SOURCE_DIR}/thirdparty/catch")
+  target_include_directories(${testfilename} BEFORE PRIVATE "${CMAKE_SOURCE_DIR}/encrypt-config")
+  target_include_directories(${testfilename} BEFORE PRIVATE "${CMAKE_SOURCE_DIR}/libminifi/include")
+  target_include_directories(${testfilename} BEFORE PRIVATE "${CMAKE_SOURCE_DIR}/libminifi/test")
+  target_include_directories(${testfilename} BEFORE PRIVATE "${CMAKE_SOURCE_DIR}/thirdparty/cxxopts/include")
```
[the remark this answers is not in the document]
fgerlits: Same as above: simply removing these lines doesn't work; I'm sure there is a better way, but I haven't been able to find it.  A clear-up could be done as part of a separate pull request.  Note that `cmake/BuildTests.cmake` looks like this, too.

encrypt-config/EncryptConfig.cpp
```diff
@@ -0,0 +1,146 @@
...
+#include "ConfigFile.h"
+#include "utils/file/FileUtils.h"
+#include "utils/OptionalUtils.h"
+
+namespace {
+const char* CONF_DIRECTORY_NAME = "conf";
+const char* BOOTSTRAP_FILE_NAME = "bootstrap.conf";
+const char* MINIFI_PROPERTIES_FILE_NAME = "minifi.properties";
+const char* ENCRYPTION_KEY_PROPERTY_NAME = "nifi.bootstrap.sensitive.key";
+const size_t ENCRYPTION_KEY_SIZE = 32;
```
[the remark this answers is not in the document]
fgerlits: fixed

encrypt-config/ConfigFile.h
```diff
@@ -0,0 +1,85 @@
...
+}
+
+class ConfigFile {
+public:
+  ConfigFile(const std::string& file_path);
+
+  utils::optional<std::string> getValue(const std::string& key) const;
+  void update(const std::string& key, const std::string& value);
+  void insertAfter(const std::string& after_key, const std::string& key, const std::string& value);
+  void append(const std::string& key, const std::string& value);
```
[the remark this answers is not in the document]
fgerlits: We use plain `update` in most places, once we use "`update` or `append`", and once we use "`update` or `insertAfter`".

I can replace the one "`update` or `append`" call with a `set()`; I could also replace all `update()` calls with `set()` -- but I'm not convinced either option would improve the code.

If you feel strongly about this, I will do it (tell me which of the above options you would prefer).

encrypt-config/ConfigFile.h
```diff
@@ -0,0 +1,85 @@
...
+
+  void updateValue(const std::string& value);
+
+  std::string line_;
+  std::string key_;
+  std::string value_;
+};
+
+inline bool operator==(const ConfigLine& left, const ConfigLine& right) {
+  return left.line_ == right.line_;
```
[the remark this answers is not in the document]
fgerlits: This is used in the tests only, to verify that the input file is not modified at all in a certain use case.  I don't want to allow `encrypt-config` to reformat the config file when it is not supposed to make any changes.

encrypt-config/ConfigFile.h
```diff
@@ -0,0 +1,85 @@
...
+  void update(const std::string& key, const std::string& value);
+  void insertAfter(const std::string& after_key, const std::string& key, const std::string& value);
+  void append(const std::string& key, const std::string& value);
+  int erase(const std::string& key);
+
+  void writeTo(const std::string& file_path) const;
+
+  size_t size() const { return config_lines_.size(); }
+
+  int encryptSensitiveProperties(const utils::crypto::Bytes& encryption_key);
```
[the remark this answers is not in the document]
fgerlits: Done.  I have also moved `decryptSensitiveProperties()` from `Configuration` to `Decryptor` as that violated the same principle even more.

encrypt-config/ConfigFile.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+  std::string key = utils::StringUtils::trim(line.substr(0, index_of_first_equals_sign));
+  if (key.empty()) { return; }
+
+  key_ = key;
+  value_ = utils::StringUtils::trim(line.substr(index_of_first_equals_sign + 1));
+}
+
+ConfigLine::ConfigLine(const std::string& key, const std::string& value)
+  : line_{key + "=" + value}, key_{key}, value_{value} {
```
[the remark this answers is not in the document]
fgerlits: done

encrypt-config/ConfigFile.cpp
```diff
@@ -0,0 +1,195 @@
...
+  if (pos != std::string::npos) {
+    line_.replace(pos + 1, std::string::npos, value);
+    value_ = value;
+  } else {
+    throw std::invalid_argument{"Cannot update value in config line: it does not contain an = sign!"};
+  }
+}
+
+ConfigFile::ConfigFile(const std::string& file_path) {
+  std::ifstream file{file_path};
```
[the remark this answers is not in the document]
fgerlits: done

encrypt-config/tests/CMakeLists.txt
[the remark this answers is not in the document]
szaszm: Listing each file is the alternative, but ok.

encrypt-config/ConfigFile.h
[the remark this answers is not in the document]
szaszm: I would only replace the one "update or append" with set. If you're not convinced, then I don't want to force this, it was just an idea.

encrypt-config/CMakeLists.txt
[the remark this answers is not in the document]
szaszm: As an acceptable workaround until the cleanup, I suggest changing to `target_include_directories` and adding `target_link_libraries(encrypt-config Threads::Threads)` against the pthread error.

libminifi/include/properties/Decryptor.h
```diff
@@ -0,0 +1,55 @@
...
+  std::string decrypt(const std::string& encrypted_text, const std::string& aad) const;
+  void decryptSensitiveProperties(Configure& configure) const;
+
+ private:
+  const utils::crypto::Bytes encryption_key_;
+};
+
+inline Decryptor::Decryptor(const utils::crypto::Bytes& encryption_key) : encryption_key_(encryption_key) {}
+
+inline bool Decryptor::isEncrypted(const utils::optional<std::string>& encryption_type) {
```
szaszm, line 44: I think this name is misleading, because the question it answer is not whether the argument is encrypted, but rather whether it's a valid encryption type string. I suggest calling it `isValidEncryptionType` or similar, move it near `utils::crypto::EncryptionType` and make it take a string rather than an optional string. (Move the optional validity check to the usage site)

later:
After looking at the test case, I realize that this is checking that it valid and it's not plaintext, so the encryption type is encrypted. I maintain that since it's checking a property of an encryption type string, it should be near the EncryptionType class. But not sure how to resolve the naming confusion, maybe you have a better idea?

libminifi/include/utils/EncryptionUtils.h
```diff
@@ -0,0 +1,105 @@
...
+
+std::string bytesToString(const Bytes& bytes);
+
+Bytes randomBytes(size_t num_bytes);
+
+struct EncryptionType {
+  static const EVP_AEAD* cipher();
+  static std::string name();
+  static size_t nonceLength();
+};
```
szaszm, line 44: Why are these all `static`? I know that currently only one encryption type is supported, but I think it should either be non-static so that it's more general, or be called `XChaCha20_Poly1305_EncryptionType` so that it doesn't give the impression of a general class.

main/MiNiFiMain.cpp
```diff
@@ -53,13 +53,41 @@
...
+bool containsEncryptedProperties(const minifi::Configure& minifi_properties) {
+  const auto is_encrypted_property_marker = [&minifi_properties](const std::string& property_name) {
+    return utils::StringUtils::endsWith(property_name, ".protected") &&
+        minifi::Decryptor::isEncrypted(minifi_properties.get(property_name));
+  };
+  const auto property_names = minifi_properties.getConfiguredKeys();
+  return std::any_of(property_names.begin(), property_names.end(), is_encrypted_property_marker);
+}
+
+void decryptSensitiveProperties(minifi::Configure& minifi_properties, const std::string& minifi_home, logging::Logger& logger) {
```
szaszm, line 75: Both with this and encrypt-config: I generally like functions that take a value and return a value better than functions that take a reference to some mutable memory and modify it in place.

szaszm (commented):
The majority of my comments are about API design questions/issues, hopefully not affecting your in-progress changes.

libminifi/include/properties/Decryptor.h
fgerlits: I have renamed it to `isValidEncryptionMarker`

libminifi/include/utils/EncryptionUtils.h
fgerlits: I wanted to group together the set of constants related to the encryption cipher.  This could be a namespace instead of an empty struct, but I don't think there is a huge difference.  I don't want to add extra complexity now to support a multi-cipher use case which will probably not happen, especially with libsodium, which hides the cipher from the user and may change it in future versions.

main/MiNiFiMain.cpp
fgerlits: Do you think it's worth copying the whole `Configure` object?  It's large.

encrypt-config/CMakeLists.txt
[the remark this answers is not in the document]
fgerlits: Hm... simply changing to `target_include_directories` worked now.  Not sure if I did something wrong earlier, or some changes since then have made it work.

main/MiNiFiMain.cpp
szaszm: In case of encrypt-config: definitely, since a few microseconds delay doesn't hurt the user experience. My estimate: 100 properties * 3 std::strings (~allocations) per property * ~500 ns per allocation ~= 150000 ns = 150 μs
In case of `decryptSensitiveProperties`: probably yes, because it runs only once during startup/initialization, but I'm not 100% sure about this one.
Also, maybe moving resources can help avoid copies. (pass-by-value and move into the value, transform private copy, return private copy with NRVO or implicit move).

If you disagree, that's also fine with me, I'm not insisting on this change.

fgerlits:
Replaced by pull request https://github.com/apache/nifi-minifi-cpp/pull/914

PR #890, opened 2020-09-02
MINIFICPP-1288 - FlowController start/load refactor

hunyadi-dev:
**It is easiest to review this PR on a per-commit basis.**

The FlowController methods: `start`, `stop`, `load`, `unload` (and maybe even `waitUnload`), were strongly coupled and were controlled through FlowController states that made it difficult to argue about what goes on in each of them.

`load` had two versions with only minor differences, one reload (with a few extra steps) and one non-reload. The version that did the reload was only called during `applyConfiguration`, so I moved the extra behaviour required there (and renamed the related functions accordingly to prove there were no other instances of calling reload).

`load` was also changing the state of `initialized_` read by only itself and `start`, but it is easy to argue that `load` is never called twice and `start` is never called without a preceeding `load`, proving this variable clutter.

### DIFF
--- a/libminifi/include/FlowController.h
+++ b/libminifi/include/FlowController.h
@@ -342,6 +342,7 @@ class FlowController : public core::controller::ControllerServiceProvider, publi
  private:
   void restartThreadPool();
   void initializeUninitializedSchedulers();
+  void reinitializeSchedulersWithNewThreadPool();
 
   template <typename T, typename = typename std::enable_if<std::is_base_of<SchedulingAgent, T>::value>::type>
   void conditionalReloadScheduler(std::shared_ptr<T>& scheduler, const bool condition) {
--- a/libminifi/include/FlowController.h
+++ b/libminifi/include/FlowController.h
@@ -100,7 +100,9 @@ class FlowController : public core::controller::ControllerServiceProvider, publi
   }
 
   // Load flow xml from disk, after that, create the root process group and its children, initialize the flows
-  virtual void load(const std::shared_ptr<core::ProcessGroup> &root = nullptr, bool reload = false);
+  // virtual void load(const std::shared_ptr<core::ProcessGroup> &root = nullptr, bool reload = false);
+  virtual void load_without_reload(const std::shared_ptr<core::ProcessGroup> &root = nullptr);
+  virtual void load_with_reload(const std::shared_ptr<core::ProcessGroup> &root = nullptr);
 
   // Whether the Flow Controller is start running
   bool isRunning() override {
@@ -338,6 +340,9 @@ class FlowController : public core::controller::ControllerServiceProvider, publi
   utils::optional<std::chrono::milliseconds> loadShutdownTimeoutFromConfiguration();
 
  private:
+  void restartThreadPool();
+  void initializeUninitializedSchedulers();
+
   template <typename T, typename = typename std::enable_if<std::is_base_of<SchedulingAgent, T>::value>::type>
   void conditionalReloadScheduler(std::shared_ptr<T>& scheduler, const bool condition) {
     if (condition) {
--- a/libminifi/include/FlowController.h
+++ b/libminifi/include/FlowController.h
@@ -101,7 +101,7 @@ class FlowController : public core::controller::ControllerServiceProvider, publi
 
   // Load flow xml from disk, after that, create the root process group and its children, initialize the flows
   // virtual void load(const std::shared_ptr<core::ProcessGroup> &root = nullptr, bool reload = false);
-  virtual void load_without_reload(const std::shared_ptr<core::ProcessGroup> &root = nullptr);
+  virtual void load(const std::shared_ptr<core::ProcessGroup> &root = nullptr);
 
   // Whether the Flow Controller is start running
   bool isRunning() override {
--- a/libminifi/include/FlowController.h
+++ b/libminifi/include/FlowController.h
@@ -102,7 +102,6 @@ class FlowController : public core::controller::ControllerServiceProvider, publi
   // Load flow xml from disk, after that, create the root process group and its children, initialize the flows
   // virtual void load(const std::shared_ptr<core::ProcessGroup> &root = nullptr, bool reload = false);
   virtual void load_without_reload(const std::shared_ptr<core::ProcessGroup> &root = nullptr);
-  virtual void load_with_reload(const std::shared_ptr<core::ProcessGroup> &root = nullptr);
 
   // Whether the Flow Controller is start running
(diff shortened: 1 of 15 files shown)


libminifi/include/FlowController.h
```diff
@@ -100,18 +100,15 @@ class FlowController : public core::controller::ControllerServiceProvider, publi
...
   bool isRunning() override {
     return running_.load() || updating_.load();
   }
 
-  // Whether the Flow Controller has already been initialized (loaded flow XML)
-  virtual bool isInitialized() {
-    return initialized_.load();
-  }
   // Start to run the Flow Controller which internally start the root process group and all its children
+  // int16_t start() override;
```
arpadboda, line 111: What's the purpose of this new comment here?

libminifi/include/FlowController.h
```diff
@@ -338,6 +335,10 @@ class FlowController : public core::controller::ControllerServiceProvider, publi
   utils::optional<std::chrono::milliseconds> loadShutdownTimeoutFromConfiguration();
 
  private:
+  void restartThreadPool();
+  void initializeUninitializedSchedulers();
+  void reinitializeSchedulersWithNewThreadPool();
```
arpadboda, line 340: I'm not sure if the name of this function actually reflect what it does. 
There is no new threadpool, as far as I see it's created only once and stopped/started according to usage later.

libminifi/src/FlowController.cpp
```diff
@@ -311,56 +302,62 @@ void FlowController::unload() {
...
+}
+
+void FlowController::initializeUninitializedSchedulers() {
+  conditionalReloadScheduler<TimerDrivenSchedulingAgent>(timer_scheduler_, !timer_scheduler_);
+  conditionalReloadScheduler<EventDrivenSchedulingAgent>(event_scheduler_, !event_scheduler_);
+  conditionalReloadScheduler<CronDrivenSchedulingAgent>(cron_scheduler_, !cron_scheduler_);
+}
+
+void FlowController::reinitializeSchedulersWithNewThreadPool() {
+  using ControllerServiceProvider = core::controller::ControllerServiceProvider;
```
arpadboda, line 324: Just fancy: in case we increase code readability this way, why don't we alias the templated gsl::not_null class instead of the parameter?

arpadboda (commented):
Most looks good, added some comments.

libminifi/include/FlowController.h
hunyadi-dev: It is a leftover from the refactoring, will be removed.

libminifi/include/FlowController.h
hunyadi-dev: It calls `ThreadPool::shutdown` which clears any running activities, joins all threads, and clears all worker threads as well (also resets config). The `ThreadPool` object is not new as in newly allocated, but new as the pool brand new threads.

I agree that it is not the best name, will rename it to `reinitializeSchedulersWithClearedThreadPool()`.

libminifi/src/FlowController.cpp
hunyadi-dev: The reason I introduced this was actually because the lines were too long otherwise, and my ocd wanted them to align :D.

Will add an alias for the `gsl::not_null`.

libminifi/src/FlowController.cpp
```diff
@@ -320,12 +320,12 @@ void FlowController::initializeUninitializedSchedulers() {
...
   restartThreadPool();
-  timer_scheduler_ = std::make_shared<TimerDrivenSchedulingAgent>(gsl::not_null<ControllerServiceProvider*>(this), provenance_repo_, flow_file_repo_, content_repo_, configuration_, thread_pool_);
-  event_scheduler_ = std::make_shared<EventDrivenSchedulingAgent>(gsl::not_null<ControllerServiceProvider*>(this), provenance_repo_, flow_file_repo_, content_repo_, configuration_, thread_pool_);
-  cron_scheduler_ = std::make_shared<CronDrivenSchedulingAgent>(gsl::not_null<ControllerServiceProvider*>(this), provenance_repo_, flow_file_repo_, content_repo_, configuration_, thread_pool_);
+  timer_scheduler_ = std::make_shared<TimerDrivenSchedulingAgent>(NonNullControllerServiceProviderPtr(this), provenance_repo_, flow_file_repo_, content_repo_, configuration_, thread_pool_);
+  event_scheduler_ = std::make_shared<EventDrivenSchedulingAgent>(NonNullControllerServiceProviderPtr(this), provenance_repo_, flow_file_repo_, content_repo_, configuration_, thread_pool_);
+  cron_scheduler_ = std::make_shared<CronDrivenSchedulingAgent>(NonNullControllerServiceProviderPtr(this), provenance_repo_, flow_file_repo_, content_repo_, configuration_, thread_pool_);
```
szaszm, line 328: Consider using `gsl::make_not_null` and relying on implicit conversion to base class pointer.

szaszm (commented):
Since the requirement of "one has to call load before start" is not documented anywhere, I would restore the `initialized_` variable to prevent invalid usage of `FlowController`.
I fell into this trap of only calling `stop`/`start` instead of `stop`,`unload`/`load`,`start` while working on #875

libminifi/src/FlowController.cpp
hunyadi-dev: This is an explicit conversion.

hunyadi-dev: What is the advantake of using `make_not_null`? It does nothing, does it?

hunyadi-dev:
> Since the requirement of "one has to call load before start" is not documented anywhere, I would restore the `initialized_` variable to prevent invalid usage of `FlowController`.
> I fell into this trap of only calling `stop`/`start` instead of `stop`,`unload`/`load`,`start` while working on #875

I see what you mean, but I disagree. I would definitely avoid adding states like this to an object. I found it already quite a task to decipher in what state can we reach each of the member functions. How about either of these options:
 1. I can make the FlowController a state machine with well defined transitions between states
 1. I can add checks to the `start` method that ensure that the members set in `load` are indeed set.

szaszm:
> I see what you mean, but I disagree. I would definitely avoid adding states like this to an object. I found it already quite a task to decipher in what state can we reach each of the member functions. How about either of these options:
>     1. I can make the FlowController a state machine with well defined transitions between states
>     2. I can add checks to the `start` method that ensure that the members set in `load` are indeed set.

I don't really understand what you mean by making `FlowController` a state machine or checking members, but in either case, if you can make it hard or impossible to use `FlowController` incorrectly, that would be a big plus in my opinion.

PR #891, opened 2020-09-02
MINIFICPP-1349 ConsumeWindowsEventLogs should honor batch commit size in a single session

lordgamez:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/windows-event-log/ConsumeWindowsEventLog.cpp
+++ b/extensions/windows-event-log/ConsumeWindowsEventLog.cpp
@@ -317,7 +317,7 @@ bool ConsumeWindowsEventLog::commitAndSaveBookmark(const std::wstring &bookmark_
   return true;
 }
 
-std::wstring ConsumeWindowsEventLog::populateSessionWithEventLogs(const std::shared_ptr<core::ProcessSession> &session,
+std::wstring ConsumeWindowsEventLog::processEventLogs(const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSession> &session,
     size_t& processed_event_count, const EVT_HANDLE& event_query_results) {
   std::wstring bookmark_xml;
   logger_->log_trace("Enumerating the events in the result set after the bookmarked event.");
@@ -345,16 +345,8 @@ std::wstring ConsumeWindowsEventLog::populateSessionWithEventLogs(const std::sha
       putEventRenderFlowFileToSession(event_render, *session);
     }
   }
-  return bookmark_xml;
-}
-
-void ConsumeWindowsEventLog::processEventLogs(const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSession> &session,
-    size_t& processed_event_count, const EVT_HANDLE& event_query_results) {
-  std::wstring bookmark_xml = populateSessionWithEventLogs(session, processed_event_count, event_query_results);
   logger_->log_trace("Finished enumerating events.");
-  if (processed_event_count > 0 && !commitAndSaveBookmark(bookmark_xml, session)) {
-    context->yield();
-  }
+  return bookmark_xml;
 }
 
 void ConsumeWindowsEventLog::onTrigger(const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSession> &session) {
@@ -403,7 +395,12 @@ void ConsumeWindowsEventLog::onTrigger(const std::shared_ptr<core::ProcessContex
   }
 
   refreshTimeZoneData();
-  processEventLogs(context,session, processed_event_count, event_query_results);
+  auto bookmark_xml = processEventLogs(context,session, processed_event_count, event_query_results);
+
+  if (processed_event_count > 0 && !commitAndSaveBookmark(bookmark_xml, session)) {
+    context->yield();
+    return;
+  }
 }
 
 wel::WindowsEventLogHandler ConsumeWindowsEventLog::getEventLogHandler(const std::string & name) {
--- a/extensions/windows-event-log/ConsumeWindowsEventLog.cpp
+++ b/extensions/windows-event-log/ConsumeWindowsEventLog.cpp
@@ -175,9 +175,9 @@ ConsumeWindowsEventLog::ConsumeWindowsEventLog(const std::string& name, utils::I
 }
 
 void ConsumeWindowsEventLog::notifyStop() {
-  std::lock_guard<std::mutex> lock(onTriggerMutex_);
+  std::lock_guard<std::mutex> lock(on_trigger_mutex_);
   logger_->log_trace("start notifyStop");
-  pBookmark_.reset();
+  bookmark_.reset();
   if (hMsobjsDll_) {
     if (FreeLibrary(hMsobjsDll_)) {
       hMsobjsDll_ = nullptr;
@@ -277,16 +277,16 @@ void ConsumeWindowsEventLog::onSchedule(const std::shared_ptr<core::ProcessConte
   bool processOldEvents{};
   context->getProperty(ProcessOldEvents.getName(), processOldEvents);
 
-  if (!pBookmark_) {
+  if (!bookmark_) {
     std::string bookmarkDir;
     context->getProperty(BookmarkRootDirectory.getName(), bookmarkDir);
     if (bookmarkDir.empty()) {
       logger_->log_error("State Directory is empty");
       throw Exception(PROCESS_SCHEDULE_EXCEPTION, "State Directory is empty");
     }
-    pBookmark_ = std::make_unique<Bookmark>(wstrChannel_, wstrQuery_, bookmarkDir, getUUIDStr(), processOldEvents, state_manager_, logger_);
-    if (!*pBookmark_) {
-      pBookmark_.reset();
+    bookmark_ = std::make_unique<Bookmark>(wstrChannel_, wstrQuery_, bookmarkDir, getUUIDStr(), processOldEvents, state_manager_, logger_);
+    if (!*bookmark_) {
+      bookmark_.reset();
       throw Exception(PROCESS_SCHEDULE_EXCEPTION, "Bookmark is empty");
     }
   }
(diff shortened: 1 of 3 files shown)


extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -298,120 +298,108 @@ void ConsumeWindowsEventLog::onSchedule(const std::shared_ptr<core::ProcessConte
...
+  if (session->outgoingConnectionsFull("success")) {
+    logger_->log_debug("Outgoing success connection is full");
+    return false;
+  }
+
+  return true;
+}
+
+std::wstring ConsumeWindowsEventLog::processEventLogs(const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSession> &session,
+    size_t& processed_event_count, const EVT_HANDLE& event_query_results) {
```
fgerlits, line 321: I would do this the other way round: return the number of events processed, and use an out parameter for the bookmark XML.  Not important, but I think that is a more common pattern.

Another option would be to return a struct `{ size_t processed_event_count; std::wstring bookmark_xml; }`.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -298,120 +298,108 @@ void ConsumeWindowsEventLog::onSchedule(const std::shared_ptr<core::ProcessConte
...
-        commitAndSaveBookmarkCount = eventCount;
-      }
-    }
-  }
-
-  logger_->log_trace("Finish enumerating events.");
-
-  if (eventCount > commitAndSaveBookmarkCount) {
-    commitAndSaveBookmark(bookmarkXml);
+  if (processed_event_count > 0 && !commitAndSaveBookmark(bookmark_xml, session)) {
```
fgerlits, line 400: I think we should yield if no events were processed, too.  I know we didn't before this change, but that seems like a bug.
```suggestion
  if (processed_event_count == 0 || !commitAndSaveBookmark(bookmark_xml, session)) {
```

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
fgerlits, line 321: could this be `const`?

extensions/windows-event-log/tests/ConsumeWindowsEventLogTests.cpp
```diff
@@ -367,27 +372,19 @@ void batchCommitSizeTestHelper(int batch_commit_size, int expected_num_commits)
...
-  batchCommitSizeTestHelper(4, 2);
-  batchCommitSizeTestHelper(3, 2);
-  batchCommitSizeTestHelper(2, 3);
-  batchCommitSizeTestHelper(1, 5);
-  batchCommitSizeTestHelper(0, 1);
+  batchCommitSizeTestHelper(1000);
+  batchCommitSizeTestHelper(5);
+  batchCommitSizeTestHelper(4);
+  batchCommitSizeTestHelper(1);
+  batchCommitSizeTestHelper(0);
```
fgerlits, line 389: I think the test is clearer if you explicitly specify the expected result, as before.  Maybe the number of events could be a parameter, too, eg. `batchCommitSizeTestHelper(int num_events_read, int batch_size, int num_events_processed)`.
```suggestion
  batchCommitSizeTestHelper(5, 1000, 5);
  batchCommitSizeTestHelper(5, 5, 5);
  batchCommitSizeTestHelper(5, 4, 4);
  batchCommitSizeTestHelper(5, 1, 1);
  batchCommitSizeTestHelper(5, 0, 5);
```

extensions/windows-event-log/tests/ConsumeWindowsEventLogTests.cpp
```diff
@@ -335,21 +335,26 @@ TEST_CASE("ConsumeWindowsEventLog prints events in XML correctly", "[onTrigger]"
...
+    REQUIRE(LogTestController::getInstance().contains(R"(<EventID Qualifiers="0">14985</EventID>)"));
+    REQUIRE(LogTestController::getInstance().contains(R"(<Level>4</Level>)"));
+    REQUIRE(LogTestController::getInstance().contains(R"(<Task>0</Task>)"));
+    REQUIRE(LogTestController::getInstance().contains(R"(<Keywords>0x80000000000000</Keywords><TimeCreated SystemTime=")"));
     // the timestamp (when the event was published) goes here
     REQUIRE(LogTestController::getInstance().contains(R"("/><EventRecordID>)"));
     // the ID of the event goes here (a number)
-    REQUIRE(LogTestController::getInstance().contains(R"(</EventRecordID><Channel>Application</Channel><Computer>)"));
+    REQUIRE(LogTestController::getInstance().contains(R"(</EventRecordID>)"));
+    REQUIRE(LogTestController::getInstance().contains(R"(<Channel>Application</Channel><Computer>)"));
```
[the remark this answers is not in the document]
lordgamez: The reason I weakened the condition is because the test was failing on my system, as the output of the event was something like [this](https://social.msdn.microsoft.com/Forums/windowsdesktop/en-US/5383445d-437e-4ad5-bf03-9584aef9527a/httpschemasmicrosoftcomwin200408eventseventquotgt?forum=windowsdesktopsearchhelp). There are a couple of additional tags in between, like the Correlation and Execution tag after the EventRecordID, which seem to be valid according to the schema. I am not sure what causes the difference.

extensions/windows-event-log/tests/ConsumeWindowsEventLogTests.cpp
```diff
@@ -367,27 +372,19 @@ void batchCommitSizeTestHelper(int batch_commit_size, int expected_num_commits)
...
-    reportEvent(APPLICATION_CHANNEL, "Event two");
-    reportEvent(APPLICATION_CHANNEL, "Event three");
-    reportEvent(APPLICATION_CHANNEL, "Event four");
-    reportEvent(APPLICATION_CHANNEL, "Event five");
-
-    test_controller.runSession(test_plan);
-
-    REQUIRE(LogTestController::getInstance().countOccurrences("processQueue commit") == expected_num_commits);
-  }
+  std::vector<std::string> events{"Event one", "Event two", "Event three", "Event four", "Event five"};
```
hunyadi-dev, line 375: Minor, but why not just a simple range based for here?

hunyadi-dev (approved):

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
lordgamez: As both are output parameters I decided to return a tuple with both values. Added in [24855ee](https://github.com/apache/nifi-minifi-cpp/pull/891/commits/24855eeb70fffe2f33b065bcd75dc77f8d5aa3f5)

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
lordgamez: Done in [82aac08](https://github.com/apache/nifi-minifi-cpp/pull/891/commits/82aac084e181abb4f213287ebd4e883255a73739)

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
lordgamez: Unfortunately no, because of `getEventLogHandler` in the `createEventRender` call. But I made 2 other methods const in [0a32d4e](https://github.com/apache/nifi-minifi-cpp/pull/891/commits/0a32d4e3eb179ea9fbf8d2914c70997430f91c58)

extensions/windows-event-log/tests/ConsumeWindowsEventLogTests.cpp
lordgamez: You are right, I made it more explicit in [fa06429](https://github.com/apache/nifi-minifi-cpp/pull/891/commits/fa064298d99a1e0efc2609548062efd5f815085a)

extensions/windows-event-log/tests/ConsumeWindowsEventLogTests.cpp
lordgamez: Maybe I just wanted to keep it in a single line, but yea I see it's better to have a simple for loop here, fixed in [fa06429](https://github.com/apache/nifi-minifi-cpp/pull/891/commits/fa064298d99a1e0efc2609548062efd5f815085a)

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -298,120 +298,111 @@ void ConsumeWindowsEventLog::onSchedule(const std::shared_ptr<core::ProcessConte
...
-  }
-
-  logger_->log_trace("Finish enumerating events.");
+  std::wstring bookmark_xml;
+  std::tie(processed_event_count, bookmark_xml) = processEventLogs(context, session, event_query_results);
 
-  if (eventCount > commitAndSaveBookmarkCount) {
-    commitAndSaveBookmark(bookmarkXml);
+  if (processed_event_count == 0 || !commitAndSaveBookmark(bookmark_xml, session)) {
+    context->yield();
```
adamdebreceni, line 404: are we sure, we want to yield a CWEL processors with no events to process? e.g. GetFile does not yield if the input directory is empty

lordgamez: @fgerlits Could you elaborate why did you suggest that we should yield in this case?

fgerlits: I think we generally want to yield if there is nothing to do, so we wait a bit longer between `onTrigger()` calls and we don't waste resources.  @adebreceni Do you know why GetFile does not yield when it has nothing to do?

adamdebreceni: I don't know why GetFile doesn't yield, but if the purpose of `yield` is resource optimization it is rightfully here

adamdebreceni (approved):

fgerlits (approved):

arpadboda (approved):

PR #893, opened 2020-09-03
MINIFICPP-1288 - FlowController C2 init refactor

hunyadi-dev:
Split up FlowController::initializeC2, remove atomicity from c2_initialized_, delete flow_update_ field.

### DIFF
--- a/libminifi/src/FlowController.cpp
+++ b/libminifi/src/FlowController.cpp
@@ -95,7 +95,6 @@ FlowController::FlowController(std::shared_ptr<core::Repository> provenance_repo
   }
   utils::IdGenerator::getIdGenerator()->generate(uuid_);
   setUUID(uuid_);
-  flow_update_ = false;
   // Setup the default values
   if (flow_configuration_ != nullptr) {
     configuration_filename_ = flow_configuration_->getConfigurationPath();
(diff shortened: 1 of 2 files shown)


libminifi/src/FlowController.cpp
```diff
@@ -216,9 +215,6 @@ bool FlowController::applyConfiguration(const std::string &source, const std::st
   bool started = false;
   try {
     load(this->root_, true);
-    flow_update_ = true;
-    started = start() == 0;
```
lordgamez, line 220: Was this supposed to be removed? The code in the conditional branch `if (started) {` below now seems to be dead code.

libminifi/src/FlowController.cpp
```diff
@@ -441,43 +429,40 @@ void FlowController::initializeC2() {
...
+  return agent_identifier;
+}
 
-  if (c2_initialized_ && !flow_update_) {
-    return;
+void FlowController::addQueueMetricsToDeviceInformation() {
+  std::shared_ptr<state::response::QueueMetrics> queueMetrics = std::make_shared<state::response::QueueMetrics>();
+  std::map<std::string, std::shared_ptr<Connection>> connections;
+  root_->getConnections(connections);
+  for (auto con : connections) {
```
lordgamez, line 451: This could be `const auto&`

libminifi/src/FlowController.cpp
hunyadi-dev: Honestly, I don't remember what was exactly extracted into this PR. I think this was opened to ease the review load on #890, but the boundaries are probably not correctly set. Closing this PR.

hunyadi-dev:
Probably part of #890, not worth rebasing separately.

PR #895, opened 2020-09-04
MINIFICPP-1352 - Comment out unused parameters (for enabling -Wall)

hunyadi-dev:
Script used:
```bash
cat build_logs.log | grep -v "thirdparty" | sort | uniq | egrep "\[-Wunused-parameter*\]" | tr ":" " " | cut -d" " -f1,2,9 |
tr "'" " " | tr -s " " | xargs -n3 python -c 'import os,sys; print("%s %s %s" % (os.path.abspath(sys.argv[1]), sys.argv[2], sys.argv[3]))' | 
sort | uniq | xargs -n 3 sh -c 'perl -pi -e "$. == $2 && s;( *([*&])? *)$3( *= *?[\w:]*(\(\))?)* *(?=[\),]);\2 /*$3*/\3;" $1' sh
```

Manual edits in:
> nanofi/include/sitetosite/CPeer.h (due to C compatibility enforcing naming arguments)
> nanofi/include/sitetosite/CRawSocketProtocol.h (due to C compatibility enforcing naming arguments)
> nanofi/src/sitetosite/CRawSocketProtocol.c (due to C compatibility enforcing naming arguments)
> libminifi/include/core/Repository.h (due to line becoming too long for linter checks)
> libminifi/include/utils/file/FileUtils.h (due to platform specific parameter usage)

### DIFF
--- a/extensions/civetweb/processors/ListenHTTP.cpp
+++ b/extensions/civetweb/processors/ListenHTTP.cpp
@@ -99,7 +99,7 @@ void ListenHTTP::initialize() {
   setSupportedRelationships(relationships);
 }
 
-void ListenHTTP::onSchedule(core::ProcessContext *context, core::ProcessSessionFactory *sessionFactory) {
+void ListenHTTP::onSchedule(core::ProcessContext *context, core::ProcessSessionFactory* /*sessionFactory*/) {
   std::string basePath;
 
   if (!context->getProperty(BasePath.getName(), basePath)) {
@@ -230,7 +230,7 @@ void ListenHTTP::onSchedule(core::ProcessContext *context, core::ProcessSessionF
 
 ListenHTTP::~ListenHTTP() = default;
 
-void ListenHTTP::onTrigger(core::ProcessContext *context, core::ProcessSession *session) {
+void ListenHTTP::onTrigger(core::ProcessContext* /*context*/, core::ProcessSession *session) {
   logger_->log_debug("OnTrigger ListenHTTP");
   processIncomingFlowFile(session);
   processRequestBuffer(session);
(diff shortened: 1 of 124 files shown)


extensions/civetweb/processors/ListenHTTP.cpp
```diff
@@ -211,7 +211,7 @@ void ListenHTTP::onSchedule(core::ProcessContext *context, core::ProcessSessionF
 
 ListenHTTP::~ListenHTTP() = default;
 
-void ListenHTTP::onTrigger(core::ProcessContext *context, core::ProcessSession *session) {
+void ListenHTTP::onTrigger(core::ProcessContext* /*context*/, core::ProcessSession *session) {
```
lordgamez, line 214: I would remove commented variables where the parameter is self-explanatory or has even less meaning than the typename, like ProcessContext - context, CivetServer - server, Configure - configure, ProcessSessionFactory - sessionFactory, ProcessGroup - pg, Parameters params, Repository - repo and so on. If you have an idea it could be filtered and replaced easily go ahead, if it would be too exhausting, just leave it for now.

lordgamez (approved):

libminifi/include/utils/file/FileUtils.h
```diff
@@ -115,14 +115,15 @@ class FileUtils {
    * @param force_posix returns the posix path separator ('/'), even when not on posix. Useful when dealing with remote posix paths.
    * @return the path separator character
    */
-  static char get_separator(bool force_posix = false) {
 #ifdef WIN32
+  static char get_separator(bool force_posix = false) {
```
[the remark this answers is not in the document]
hunyadi-dev: Fixed by having the closing brace appear in both versions inside the macro.

szaszm:
I'd be fine with this after a rebase

szaszm (approved):
I'm fine with merging this after @lordgamez 's comment is closed (either with or without change, but with an answer).

extensions/civetweb/processors/ListenHTTP.cpp
hunyadi-dev: I don't have a quick-fix for that. People probably need to edit those cases one-by-one when they encounter them.

PR #896, opened 2020-09-04
MINIFICPP-1353 Fix a heap-use-after-free error

fgerlits:
https://issues.apache.org/jira/browse/MINIFICPP-1353

Declare the variables in the right order in `InvokeHTTP::onTrigger()` to make sure that the destructors run in the right order.

In the long term, the client should probably own the callbacks (and callbacks should own their sub-callbacks), but that would be a bigger change, and is less important.

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/http-curl/processors/InvokeHTTP.cpp
+++ b/extensions/http-curl/processors/InvokeHTTP.cpp
@@ -283,6 +283,7 @@ void InvokeHTTP::onTrigger(const std::shared_ptr<core::ProcessContext> &context,
   // create a transaction id
   std::string tx_id = generateId();
 
+  // Note: callback must be declared before callbackObj so that they are destructed in the correct order
   std::unique_ptr<utils::ByteInputCallBack> callback = nullptr;
   std::unique_ptr<utils::HTTPUploadCallback> callbackObj = nullptr;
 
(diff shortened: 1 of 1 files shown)


szaszm (approved):

extensions/http-curl/processors/InvokeHTTP.cpp
```diff
@@ -283,6 +283,10 @@ void InvokeHTTP::onTrigger(const std::shared_ptr<core::ProcessContext> &context,
   // create a transaction id
   std::string tx_id = generateId();
 
+  std::unique_ptr<utils::ByteInputCallBack> callback = nullptr;
+  std::unique_ptr<utils::HTTPUploadCallback> callbackObj = nullptr;
+
+  // Client declared after the callbacks to make sure the callbacks are still available when the client is destructed
```
hunyadi-dev, line 289: Might be worth mentioning that the destruction of `callback` and `callbackObj` also matters.

hunyadi-dev (approved):

extensions/http-curl/processors/InvokeHTTP.cpp
fgerlits: Good point, I have added a comment.  This is a good example when a comment is necessary, but it is also a code smell. :)

lordgamez (approved):

PR #897, opened 2020-09-06
MINIFICPP-1357 Implement and test proxy handling in InvokeHTTP

lordgamez:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/docker/test/integration/test_http.py
+++ b/docker/test/integration/test_http.py
@@ -21,7 +21,6 @@ def test_invoke_listen():
     """
     Verify sending using InvokeHTTP to a receiver using ListenHTTP.
     """
-
     invoke_flow = (GetFile('/tmp/input')
                    >> LogAttribute()
                    >> InvokeHTTP('http://minifi-listen:8080/contentListener', method='POST'))
@@ -30,6 +29,29 @@ def test_invoke_listen():
 
     with DockerTestCluster(SingleFileOutputValidator('test')) as cluster:
         cluster.put_test_data('test')
+        cluster.deploy_flow(listen_flow, name='minifi-listen')
+        cluster.deploy_flow(invoke_flow, name='minifi-invoke')
+
+        assert cluster.check_output()
+
+def test_invoke_listen_with_proxy():
+    """
+    Verify sending through a proxy using InvokeHTTP to a receiver using ListenHTTP.
+    """
+    invoke_flow = (GetFile('/tmp/input')
+                   >> LogAttribute()
+                   >> InvokeHTTP('http://minifi-listen:8080/contentListener',
+                                 method='POST',
+                                 proxy_host='http-proxy',
+                                 proxy_port='3128',
+                                 proxy_username='admin',
+                                 proxy_password='test101'))
+
+    listen_flow = ListenHTTP(8080) >> LogAttribute() >> PutFile('/tmp/output')
+
+    with DockerTestCluster(SingleFileOutputValidator('test')) as cluster:
+        cluster.put_test_data('test')
+        cluster.deploy_flow(None, engine='http-proxy')
         cluster.deploy_flow(listen_flow, name='minifi-listen')
         cluster.deploy_flow(invoke_flow, name='minifi-invoke')
 
--- a/docker/test/integration/test_http.py
+++ b/docker/test/integration/test_http.py
@@ -56,3 +56,4 @@ def test_invoke_listen_with_proxy():
         cluster.deploy_flow(invoke_flow, name='minifi-invoke')
 
         assert cluster.check_output()
+        assert cluster.check_http_proxy_access()
--- a/extensions/http-curl/processors/InvokeHTTP.cpp
+++ b/extensions/http-curl/processors/InvokeHTTP.cpp
@@ -145,8 +145,8 @@ void InvokeHTTP::initialize() {
   properties.insert(ProxyHost);
   properties.insert(ProxyPort);
   properties.insert(ProxyUsername);
-  properties.insert(UseChunkedEncoding);
   properties.insert(ProxyPassword);
+  properties.insert(UseChunkedEncoding);
   properties.insert(ContentType);
   properties.insert(SendBody);
   properties.insert(DisablePeerVerification);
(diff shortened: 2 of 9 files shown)


extensions/http-curl/processors/InvokeHTTP.cpp
```diff
@@ -246,6 +246,20 @@ void InvokeHTTP::onSchedule(const std::shared_ptr<core::ProcessContext> &context
   if (context->getProperty(DisablePeerVerification.getName(), disablePeerVerification)) {
     utils::StringUtils::StringToBool(disablePeerVerification, disable_peer_verification_);
   }
+
+  std::string proxy_value;
+  if (context->getProperty(ProxyHost.getName(), proxy_value) && !proxy_value.empty()) {
+    proxy_.host = proxy_value;
+  }
+  if (context->getProperty(ProxyPort.getName(), proxy_value) && !proxy_value.empty()) {
+    proxy_.port = std::stoi(proxy_value);
```
adamdebreceni, line 255: `getProperty` is a templated method, that depending on the second argument's type carries out checks and conversions from `std::string`, so feel free to pass `proxy.port` as the second argument here, moreover I believe `getProperty` adheres to the contract "the out argument is not modified on failure, missing value", so we don't even need the "if"s here

libminifi/include/utils/HTTPClient.h
```diff
@@ -300,6 +319,9 @@ class BaseHTTPClient {
   virtual void setUploadCallback(HTTPUploadCallback *callbackObj) {
   }
 
+  virtual void setSeekFunction(HTTPUploadCallback *callbackObj) {
+  }
```
adamdebreceni, line 323: it seems like `HTTPClient` is the only class inheriting from it, and noone directly instantiates a BaseHTTPClient, could we make these noop virtual methods pure virtual?

extensions/http-curl/processors/InvokeHTTP.cpp
arpadboda: The other reason to remove if statements is updating the config: we should null the members, otherwise proxy properties can't be removed without restarting minifi.

hunyadi-dev:
Please check the CI errors, they seem related to the change.

libminifi/include/utils/ByteArrayCallback.h
```diff
@@ -57,6 +57,7 @@ class ByteInputCallBack : public InputStreamCallback {
   }
 
   virtual void seek(size_t pos) {
+    ptr = &vec[pos];
```
szaszm, line 60: What's the purpose of this change? I think this may break `getBuffer`.

lordgamez: Curl proxy authentication requires 2 steps (getting the available authentication methods then authenticate with the chosen one) and after the first step it needs a data rewind which requires implementing the seek function. You are right, it broke the getBuffer that's what partly caused the test failures, it is fixed in [fd3a7c6](https://github.com/apache/nifi-minifi-cpp/pull/897/commits/fd3a7c653202c7cdd460153175d3deb13b7651ce)

extensions/http-curl/processors/InvokeHTTP.h
```diff
@@ -155,8 +155,10 @@ class InvokeHTTP : public core::Processor {
   bool use_chunked_encoding_{false};
   // penalize on no retry
   bool penalize_no_retry_{false};
-  // disable peer verification ( makes susceptible for MITM attacks )
-  bool disable_peer_verification_{false};
+  bool disable_peer_verification_{false}; // ( makes susceptible for MITM attacks )
+  bool follow_redirects_{true};
```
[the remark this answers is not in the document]
lordgamez: Fixed in  [e0e31c5](https://github.com/apache/nifi-minifi-cpp/pull/897/commits/e0e31c5e96e0ad29c54ca09230a845588664b864)

docker/test/integration/test_http.py
```diff
@@ -34,3 +33,26 @@ def test_invoke_listen():
...
+
+    listen_flow = ListenHTTP(8080) >> LogAttribute() >> PutFile('/tmp/output')
+
+    with DockerTestCluster(SingleFileOutputValidator('test')) as cluster:
+        cluster.put_test_data('test')
+        cluster.deploy_flow(None, engine='http-proxy')
+        cluster.deploy_flow(listen_flow, name='minifi-listen')
+        cluster.deploy_flow(invoke_flow, name='minifi-invoke')
+
+        assert cluster.check_output()
```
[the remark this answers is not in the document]
lordgamez: We could check the acces logs of the http-proxy, that's what I was checking manually, I will add it to the test.

libminifi/include/utils/HTTPClient.h
lordgamez: Done in [fd3a7c6](https://github.com/apache/nifi-minifi-cpp/pull/897/commits/fd3a7c653202c7cdd460153175d3deb13b7651ce)

extensions/http-curl/processors/InvokeHTTP.cpp
lordgamez: Done in [2bc2b75](https://github.com/apache/nifi-minifi-cpp/pull/897/commits/429675f5d9a526c529080f4468d170e320f21dca)

libminifi/include/utils/ByteArrayCallback.h
```diff
@@ -65,7 +66,7 @@ class ByteInputCallBack : public InputStreamCallback {
   }
 
   virtual char *getBuffer(size_t pos) {
-    return ptr + pos;
+    return reinterpret_cast<char*>(&vec[0]) + pos;
```
szaszm, line 69: Please use `vec.data()` to avoid out-of-bounds indexing when `vec` is empty, or index with pos.
```suggestion
    return &vec[pos];
```
or
```suggestion
    return vec.data() + pos;
```

lordgamez: Done in [839ed47](https://github.com/apache/nifi-minifi-cpp/pull/897/commits/839ed47d2c3618a1fb1443f0929bfc8908051a87)

docker/test/integration/test_http.py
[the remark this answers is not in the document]
lordgamez: Added check in [5047001](https://github.com/apache/nifi-minifi-cpp/pull/897/commits/504700148228b571ee69a2d58322c235d5f71428)

szaszm (approved):

adamdebreceni (approved):

PR #900, opened 2020-09-08
MINIFICPP-1350 - Explicit serializer for MergeContent

adamdebreceni:
Best to review these changes commit-by-commit as there was a refactoring commit that touched quite a few files.

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/bustache/ApplyTemplate.h
+++ b/extensions/bustache/ApplyTemplate.h
@@ -61,7 +61,7 @@ class ApplyTemplate : public core::Processor {
   class WriteCallback : public OutputStreamCallback {
    public:
     WriteCallback(const std::string &templateFile, const std::shared_ptr<core::FlowFile> &flow_file);
-    int64_t process(std::shared_ptr<io::BaseStream> stream);
+    int64_t process(const std::shared_ptr<io::BaseStream>& stream);
 
    private:
     std::shared_ptr<logging::Logger> logger_;
--- a/extensions/libarchive/MergeContent.cpp
+++ b/extensions/libarchive/MergeContent.cpp
@@ -51,7 +51,11 @@ core::Property MergeContent::MergeStrategy(
 core::Property MergeContent::MergeFormat(
   core::PropertyBuilder::createProperty("Merge Format")
   ->withDescription("Merge Format")
-  ->withAllowableValues<std::string>({merge_content_options::MERGE_FORMAT_CONCAT_VALUE, merge_content_options::MERGE_FORMAT_TAR_VALUE, merge_content_options::MERGE_FORMAT_ZIP_VALUE})
+  ->withAllowableValues<std::string>({
+      merge_content_options::MERGE_FORMAT_CONCAT_VALUE,
+      merge_content_options::MERGE_FORMAT_TAR_VALUE,
+      merge_content_options::MERGE_FORMAT_ZIP_VALUE,
+      merge_content_options::MERGE_FORMAT_FLOWFILE_STREAM_V3_VALUE})
   ->withDefaultValue(merge_content_options::MERGE_FORMAT_CONCAT_VALUE)->build());
 core::Property MergeContent::CorrelationAttributeName("Correlation Attribute Name", "Correlation Attribute Name", "");
 core::Property MergeContent::DelimiterStrategy(
@@ -74,11 +78,6 @@ core::Property MergeContent::AttributeStrategy(
                     "only the attributes that exist on all FlowFiles in the bundle, with the same value, will be preserved.")
   ->withAllowableValues<std::string>({merge_content_options::ATTRIBUTE_STRATEGY_KEEP_COMMON, merge_content_options::ATTRIBUTE_STRATEGY_KEEP_ALL_UNIQUE})
   ->withDefaultValue(merge_content_options::ATTRIBUTE_STRATEGY_KEEP_COMMON)->build());
-core::Property MergeContent::FlowFileSerializer(
-  core::PropertyBuilder::createProperty("Flow File Serializer")
-  ->withDescription("Determines how to flow files should be serialized before merging")
-  ->withAllowableValues<std::string>({merge_content_options::SERIALIZER_PAYLOAD, merge_content_options::SERIALIZER_FLOW_FILE_V3})
-  ->withDefaultValue(merge_content_options::SERIALIZER_PAYLOAD)->build());
 core::Relationship MergeContent::Merge("merged", "The FlowFile containing the merged content");
 
 void MergeContent::initialize() {
@@ -99,7 +98,6 @@ void MergeContent::initialize() {
   properties.insert(Demarcator);
   properties.insert(KeepPath);
   properties.insert(AttributeStrategy);
-  properties.insert(FlowFileSerializer);
   setSupportedProperties(properties);
   // Set the supported relationships
   std::set<core::Relationship> relationships;
@@ -134,7 +132,6 @@ void MergeContent::onSchedule(core::ProcessContext *context, core::ProcessSessio
   context->getProperty(Demarcator.getName(), demarcator_);
   context->getProperty(KeepPath.getName(), keepPath_);
   context->getProperty(AttributeStrategy.getName(), attributeStrategy_);
-  context->getProperty(FlowFileSerializer.getName(), flowFileSerializer_);
 
   validatePropertyOptions();
 
@@ -143,6 +140,19 @@ void MergeContent::onSchedule(core::ProcessContext *context, core::ProcessSessio
   }
   logger_->log_debug("Merge Content: Strategy [%s] Format [%s] Correlation Attribute [%s] Delimiter [%s]", mergeStrategy_, mergeFormat_, correlationAttributeName_, delimiterStrategy_);
   logger_->log_debug("Merge Content: Footer [%s] Header [%s] Demarcator [%s] KeepPath [%d]", footer_, header_, demarcator_, keepPath_);
+
+  if (mergeFormat_ != merge_content_options::MERGE_FORMAT_CONCAT_VALUE) {
+    if (!header_.empty()) {
+      logger_->log_warn("Header property only works with the Binary Concatenation format, value [%s] is ignored", header_);
+    }
+    if (!footer_.empty()) {
+      logger_->log_warn("Footer property only works with the Binary Concatenation format, value [%s] is ignored", footer_);
+    }
+    if (!demarcator_.empty()) {
+      logger_->log_warn("Demarcator property only works with the Binary Concatenation format, value [%s] is ignored", demarcator_);
+    }
+  }
+
   if (delimiterStrategy_ == merge_content_options::DELIMITER_STRATEGY_FILENAME) {
     if (!header_.empty()) {
       headerContent_ = readContent(header_);
@@ -170,7 +180,8 @@ void MergeContent::validatePropertyOptions() {
 
   if (mergeFormat_ != merge_content_options::MERGE_FORMAT_CONCAT_VALUE &&
       mergeFormat_ != merge_content_options::MERGE_FORMAT_TAR_VALUE &&
-      mergeFormat_ != merge_content_options::MERGE_FORMAT_ZIP_VALUE) {
+      mergeFormat_ != merge_content_options::MERGE_FORMAT_ZIP_VALUE &&
+      mergeFormat_ != merge_content_options::MERGE_FORMAT_FLOWFILE_STREAM_V3_VALUE) {
     logger_->log_error("Merge format not supported %s", mergeFormat_);
     throw minifi::Exception(ExceptionType::PROCESSOR_EXCEPTION, "Invalid merge format: " + mergeFormat_);
   }
@@ -185,12 +196,6 @@ void MergeContent::validatePropertyOptions() {
       attributeStrategy_ != merge_content_options::ATTRIBUTE_STRATEGY_KEEP_ALL_UNIQUE) {
     logger_->log_error("Attribute strategy not supported %s", attributeStrategy_);
     throw minifi::Exception(ExceptionType::PROCESSOR_EXCEPTION, "Invalid attribute strategy: " + attributeStrategy_);
-  }
-
-  if (flowFileSerializer_ != merge_content_options::SERIALIZER_PAYLOAD &&
-      flowFileSerializer_ != merge_content_options::SERIALIZER_FLOW_FILE_V3) {
-    logger_->log_error("FlowFile serializer not supported %s", flowFileSerializer_);
-    throw minifi::Exception(ExceptionType::PROCESSOR_EXCEPTION, "Invalid flowFile serializer: " + flowFileSerializer_);
   }
 }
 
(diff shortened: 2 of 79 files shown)


libminifi/include/io/StreamPipe.h
```diff
@@ -0,0 +1,110 @@
...
+ public:
+  virtual ~InputStreamCallback() = default;
+  // virtual void process(std::ifstream *stream) = 0;
+
+  virtual int64_t process(std::shared_ptr<io::BaseStream> stream) = 0;
+};
+class OutputStreamCallback {
+ public:
+  virtual ~OutputStreamCallback() = default;
+  virtual int64_t process(std::shared_ptr<io::BaseStream> stream) = 0;
```
lordgamez, line 43: Could these parameters be `const std::shared_ptr<io::BaseStream>&` instead? We may not need to increase the refcount here.

lordgamez:
Please check the linter as well. It warns about some missing includes.

libminifi/include/io/StreamPipe.h
```diff
@@ -0,0 +1,110 @@
...
+namespace apache {
+namespace nifi {
+namespace minifi {
+
+// FlowFile IO Callback functions for input and output
+// throw exception for error
+class InputStreamCallback {
+ public:
+  virtual ~InputStreamCallback() = default;
+  // virtual void process(std::ifstream *stream) = 0;
```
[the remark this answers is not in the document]
adamdebreceni: removed

extensions/libarchive/MergeContent.cpp
```diff
@@ -147,6 +152,9 @@ void MergeContent::onSchedule(core::ProcessContext *context, core::ProcessSessio
   if (context->getProperty(AttributeStrategy.getName(), value) && !value.empty()) {
     attributeStrategy_ = value;
   }
+  if (context->getProperty(FlowFileSerializer.getName(), value) && !value.empty()) {
+    flowFileSerializer_ = value;
+  }
 
   validatePropertyOptions();
```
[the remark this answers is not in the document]
adamdebreceni: done

extensions/libarchive/MergeContent.cpp
```diff
@@ -147,6 +152,9 @@ void MergeContent::onSchedule(core::ProcessContext *context, core::ProcessSessio
   if (context->getProperty(AttributeStrategy.getName(), value) && !value.empty()) {
     attributeStrategy_ = value;
   }
+  if (context->getProperty(FlowFileSerializer.getName(), value) && !value.empty()) {
```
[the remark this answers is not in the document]
adamdebreceni: done

libminifi/include/io/StreamPipe.h
adamdebreceni: changed it, I had to touch quite a few files as well

lordgamez (approved):

szaszm:
There are a bunch of changes changing "pass shared_ptr by value" to "pass shared_ptr by const ref". Why?
If we intend to share ownership, it should be passed by value. If we just want to use the underlying object, passing a reference or raw pointer to that is probably better.

lordgamez:
> There are a bunch of changes changing "pass shared_ptr by value" to "pass shared_ptr by const ref". Why?
> If we intend to share ownership, it should be passed by value. If we just want to use the underlying object, passing a reference or raw pointer to that is probably better.
> 
> Related very incomplete guideline: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-sharedptrparam-const

In this case I suppose we only want to use the underlying object so we do not need to pass by value and we can use a raw pointer or a simple reference here. The only difference maybe is that if we decide later that we need to share the ownership with someone in the call chain we have the possibility. As I recall clang static analyzer usually suggests using shared_ptr const refs in these cases.

extensions/bustache/ApplyTemplate.h
```diff
@@ -61,7 +61,7 @@ class ApplyTemplate : public core::Processor {
   class WriteCallback : public OutputStreamCallback {
    public:
     WriteCallback(const std::string &templateFile, const std::shared_ptr<core::FlowFile> &flow_file);
-    int64_t process(std::shared_ptr<io::BaseStream> stream);
+    int64_t process(const std::shared_ptr<io::BaseStream>& stream);
```
adamdebreceni, line 64: I would say, that using `const shared_ptr<>&` is certainly an improvement, as @lordgamez said it does not force us to take ownership, and it is agreement with the proliferation of managed pointers in the rest of the system (so it lets us call into interfaces expecting a managed ptr), I don't have a strong opinion about it though, because I plan on removing the whole `*StreamCallback` stuff, so if you feel like it should be changed, decide which one should it be:

1. shared_ptr<BaseStream>
2. const shared_ptr<BaseStream>&
3. BaseStream&
4. BaseStream*
5. gsl::not_null<BaseStream&>

szaszm: Understood. I'd go for 3.

lordgamez: I'm okay with any options of 2-5.

szaszm: Note: I didn't really review this PR, just briefly looked at the diff and shared my thoughts about it. Me generally preferring 3. is because it's normally not nullable and doesn't enforce any lifetime semantics of the object, which is normally the ideal when we only care about the object. When storing as a data member, I prefer 5. or 4, because they are mutable and sometimes one needs nullable members to be able to implement move semantics.

With that said, I can accept any other way, in any case, especially given good reasoning (e.g. the proliferation of shared_ptr throughout the codebase).

adamdebreceni: as it turns out there are a number of implementations that store the stream, so that (third-party) callbacks may access it, e.g. `CompressContent::WriteCallback`, `JniByteInputStream`, `FocusArchiveEntry::ReadCallback`, `ArchiveMerge::WriteCallback`, `UnfocusArchiveEntry::WriteCallback`, `LuaInputStreamCallback`, `PyOutputStreamCallback`, moreover the hash functions in `HashContent` expect a `shared_ptr`

base on this coupled with the diff explosion that would result from rewriting all the `stream->write` to `stream.write`, I propose we leave this be for now, and return to it in a later PR

libminifi/include/FlowFileRecord.h
```diff
@@ -39,6 +39,7 @@
 #include "ResourceClaim.h"
 #include "Connection.h"
 #include "io/OutputStream.h"
+#include "io/StreamPipe.h"
```
szaszm, line 42: Is this a leftover?

libminifi/include/io/StreamPipe.h
```diff
@@ -0,0 +1,110 @@
...
+
+#include <memory>
+#include <utility>
+#include "BaseStream.h"
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+
```
szaszm, line 31: I'd expect these to be in the `io` namespace like everything else in the io directory.

extensions/libarchive/MergeContent.cpp
```diff
@@ -72,10 +74,12 @@ core::Property MergeContent::AttributeStrategy(
                     "only the attributes that exist on all FlowFiles in the bundle, with the same value, will be preserved.")
   ->withAllowableValues<std::string>({merge_content_options::ATTRIBUTE_STRATEGY_KEEP_COMMON, merge_content_options::ATTRIBUTE_STRATEGY_KEEP_ALL_UNIQUE})
   ->withDefaultValue(merge_content_options::ATTRIBUTE_STRATEGY_KEEP_COMMON)->build());
+core::Property MergeContent::FlowFileSerializer(
+  core::PropertyBuilder::createProperty("Flow File Serializer")
+  ->withDescription("Determines how to flow files should be serialized before merging")
+  ->withAllowableValues<std::string>({merge_content_options::SERIALIZER_PAYLOAD, merge_content_options::SERIALIZER_FLOW_FILE_V3})
+  ->withDefaultValue(merge_content_options::SERIALIZER_PAYLOAD)->build());
```
szaszm, line 81: NiFi uses the merge format property for this purpose, with the new serializer bound to the equivalent of `BinaryConcatenationMerge`. Why is it separate here?

libminifi/CMakeLists.txt
```diff
@@ -89,7 +89,7 @@ if (NOT OPENSSL_OFF)
 	set(TLS_SOURCES "src/io/tls/*.cpp")
 endif()
 
-file(GLOB SOURCES  "src/utils/file/*.cpp" "src/sitetosite/*.cpp"  "src/core/logging/*.cpp"  "src/core/state/*.cpp" "src/core/state/nodes/*.cpp" "src/c2/protocols/*.cpp" "src/c2/triggers/*.cpp" "src/c2/*.cpp" "src/io/*.cpp" ${SOCKET_SOURCES} ${TLS_SOURCES} "src/core/controller/*.cpp" "src/controllers/*.cpp" "src/controllers/keyvalue/*.cpp" "src/core/*.cpp"  "src/core/repository/*.cpp" "src/core/yaml/*.cpp" "src/core/reporting/*.cpp"  "src/provenance/*.cpp" "src/utils/*.cpp" "src/*.cpp")
+file(GLOB SOURCES  "src/utils/file/*.cpp" "src/sitetosite/*.cpp"  "src/core/logging/*.cpp"  "src/core/state/*.cpp" "src/core/state/nodes/*.cpp" "src/c2/protocols/*.cpp" "src/c2/triggers/*.cpp" "src/c2/*.cpp" "src/io/*.cpp" ${SOCKET_SOURCES} ${TLS_SOURCES} "src/core/controller/*.cpp" "src/controllers/*.cpp" "src/controllers/keyvalue/*.cpp" "src/core/*.cpp"  "src/core/repository/*.cpp" "src/core/yaml/*.cpp" "src/core/reporting/*.cpp" "src/serialization/*.cpp" "src/provenance/*.cpp" "src/utils/*.cpp" "src/*.cpp")
```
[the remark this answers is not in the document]
adamdebreceni: agree with the rationale, although it is rare that we add new files, in case we go down the "explicit" path I would do it a separate PR and for all source files, imo it would be strange if we used it for some files but not for others

extensions/libarchive/MergeContent.cpp
adamdebreceni: imo the idea of WHAT we want to merge, and HOW we want to merge are orthogonal, here `Flow File Serializer` answers the WHAT and `Merge Format` answers the HOW

szaszm: I agree on the implementation level, but on the interface, compatibility with NiFi offers a greater benefit than separation of concerns IMO. Now you can create a tar of FFv3-serialized flow files, but what would you do with it if NiFi can't handle it?

libminifi/include/FlowFileRecord.h
adamdebreceni: intentional as I didn't want to add the includes, as this change of moving `*StreamCallback` to a separate file is already tangential to the issue, I would like to do it in a separate PR, likewise with moving it to the `io` namespace
[the ticket](https://issues.apache.org/jira/browse/MINIFICPP-1386)

libminifi/include/io/StreamPipe.h
```diff
@@ -0,0 +1,110 @@
...
+    }
+    int remaining = readRet;
+    int transferred = 0;
+    while (remaining > 0) {
+      int writeRet = dst->write(buffer + transferred, remaining);
+      // TODO(adebreceni):
+      //   write might return 0, e.g. in case of a congested server
+      //   what should we return then?
+      //     - the number of bytes read or
+      //     - the number of bytes wrote
```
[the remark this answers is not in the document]
adamdebreceni: given where `internal::pipe` is used I would generalize the question to "what should `*StreamCallback::process` return?"

e.g. if `ZipProcessor::process` returns `10` what could that mean? should we retry `process` hoping that the (implicitly captured) `output` is able to accept the data, or does that mean that we transferred all the data? (if it returns readByteCount we are no closer to the answer as the bytes might have been read but not successfully transferred)

currently I think returning anything other than [`Failure`, `Success`, `Retry`?] is of little utility as most such callbacks cannot handle retry

extensions/libarchive/MergeContent.cpp
adamdebreceni: should be a straightforward change, this being an interface level decision I would also like to ask @arpadboda to confirm dropping the separate `Flow File Serializer` property and merging it into `Merge Format`

libminifi/include/io/StreamPipe.h
adamdebreceni: as expressed above, this is a tangential change and it would be better if the namespace change gets its own ticket [here](https://issues.apache.org/jira/browse/MINIFICPP-1386)

extensions/libarchive/MergeContent.cpp
arpadboda: The current implementation makes sense in a way, but I also think that compatibility with NiFi would bring more benefits, so please merge in case it's not a big change.

adamdebreceni: the serializer has been integrated into the Merge Format (using the NiFi-equivalent `FlowFile Stream, v3` name)

libminifi/include/io/StreamPipe.h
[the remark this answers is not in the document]
szaszm: I don't think we can give strong guarantees with streams, unless we make them transactional, which would take significant effort. In case of a ZipProcessor, I'd return the number of input bytes that successfully made it through the pipeline. If this is not possible, which I think is likely, then just return Failure.

What would be the meaning of `Retry`? Otherwise I'm also fine with the [`Failure`, `Success`] range.

extensions/libarchive/MergeContent.h
```diff
@@ -138,67 +107,87 @@ class BinaryConcatenationMerge : public MergeBin {
...
 };
 
 
 // Archive Class
 class ArchiveMerge {
-public:
-  // Nest Callback Class for read stream
-  class ReadCallback: public InputStreamCallback {
+ public:
+ class ArchiveWriter : public io::OutputStream {
```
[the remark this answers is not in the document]
szaszm: c5021a9

szaszm (approved):

szaszm:
I did some pre-merge manual testing of FFv3. For a while everything worked fine and just as I removed the GenerateFlowFile -> MergeContent connection, NiFi started spewing out these exceptions:
```
2020-10-22 19:16:28,318 ERROR [Timer-Driven Process Thread-6] o.a.n.processors.standard.UnpackContent UnpackContent[id=514003df-0175-1000-9df2-b9696fdbaa96] Unable to unpack StandardFlowFileRecord
[uuid=6803473c-2f9c-443f-900a-1f58f8721af2,claim=,offset=0,name=1603386311828679904,size=0] due to org.apache.nifi.processor.exception.ProcessException: IOException thrown from UnpackContent[id=51
4003df-0175-1000-9df2-b9696fdbaa96]: java.io.IOException: Not in FlowFile-v3 format; routing to failure: org.apache.nifi.processor.exception.ProcessException: IOException thrown from UnpackContent
[id=514003df-0175-1000-9df2-b9696fdbaa96]: java.io.IOException: Not in FlowFile-v3 format
org.apache.nifi.processor.exception.ProcessException: IOException thrown from UnpackContent[id=514003df-0175-1000-9df2-b9696fdbaa96]: java.io.IOException: Not in FlowFile-v3 format
at org.apache.nifi.controller.repository.StandardProcessSession.write(StandardProcessSession.java:2770)
at org.apache.nifi.processors.standard.UnpackContent$FlowFileStreamUnpacker$1.process(UnpackContent.java:441)
[...]
```
Apparently 0 sized flow files with the mime.type of `application/flowfile-v3` appeared in NiFi.
If this is expected, then fine, otherwise could you check it?

adamdebreceni:
nice finding @szaszm! as previously discussed, it seems to be unrelated to this PR, as the same behavior is also observable on the current main, the cause of this issue is that for some reason the `FlowFileRepository::prune_stored_flowfiles` gets executed multiple times, reenqueuing an already deleted flowFile (we load it from the checkpoint we create beforehand where the deletion did not happen), but by then the content resource is deleted, yielding and empty resource (which is an invalid FFv3 format)

the created jira ticket is [here](https://issues.apache.org/jira/browse/MINIFICPP-1399)

PR #911, opened 2020-09-18
MINIFICPP-1373 - Add proper FlowFile::getAttribute getter [QoL][mini-PR]

hunyadi-dev:
Simple change extracted as separate commit, so that further changes will be easier to review.

### DIFF
--- a/libminifi/include/core/FlowFile.h
+++ b/libminifi/include/core/FlowFile.h
@@ -25,6 +25,7 @@
 #include <utility>
 #include <vector>
 
+#include "utils/OptionalUtils.h"
 #include "utils/TimeUtil.h"
 #include "ResourceClaim.h"
 #include "Connectable.h"
@@ -134,7 +135,9 @@ class FlowFile : public CoreComponent, public ReferenceContainer {
    * @param value value to set
    * @return result of finding key
    */
-  bool getAttribute(std::string key, std::string& value) const;
+  bool getAttribute(const std::string& key, std::string& value) const;
+
+  utils::optional<std::string> getAttribute(const std::string& key) const;
 
   /**
    * Updates the value in the attribute map that corresponds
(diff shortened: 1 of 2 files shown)


libminifi/include/core/FlowFile.h
```diff
@@ -129,7 +130,9 @@ class FlowFile : public core::Connectable, public ReferenceContainer {
    * @param value value to set
    * @return result of finding key
    */
-  bool getAttribute(std::string key, std::string& value) const;
+  bool getAttribute(const std::string& key, std::string& value) const;
+
+  utils::optional<std::reference_wrapper<const std::string>> getAttribute(const std::string& key) const;
```
szaszm, line 135: Why reference wrapper? I would avoid returning a reference to some internal data by default because of possible lifetime issues. If there is a certain need for this, I would call the function something else, like `getAttributeRef`.

arpadboda: +1. 

Moreover I also think that copying a string is not going to be the bottleneck in this software, so for safety reasons I would simply make it a copy.

hunyadi-dev: One needs to call `.value().get()` or `->get()` on an optional of reference handlers. It indeed would be an issue if someone seriously misused `auto` for holding the return value, but otherwise this is even safer than returning a const ref.
https://godbolt.org/z/jbhf3n

Will change this to **return a copy**  :cry:

szaszm: Note: An optional of reference handlers should in general be replaced by a non-owning pointer. They have the same semantics, but tooling is better equipped to handle pointers.

`auto` generally means safe value semantics, and it becoming a "serious misuse" is another footgun that reference types currently have to live with. How is this safer otherwise, than returning a const ref or non-owning pointer to const?

(I'm continuing because I want to either convince you that this is better or understand your standpoint and why I'm wrong.)

hunyadi-dev: I am not arguing for safety, to plut it blunt I just do not like unnecessary copies.

szaszm (approved):

PR #914, opened 2020-09-29
MINIFICPP-1323 Encrypt sensitive properties using libsodium

fgerlits:
https://issues.apache.org/jira/browse/MINIFICPP-1323

Encrypt sensitive properties in the minifi.properties file. The change has four parts:

 * introduce a new dependency: libsodium
 * wrap libsodium functions in `EncryptionUtils`
 * implement a new command-line tool called `encrypt-config` which can encrypt the sensitive properties
 * decrypt the sensitive properties during startup.

Please also review the wiki page https://cwiki.apache.org/confluence/display/MINIFI/Encrypt+sensitive+configuration+properties as part of this pull request.

---

### For all changes:
     in the commit message?

### For code changes:

### Note:


encrypt-config/EncryptConfig.cpp
```diff
@@ -0,0 +1,150 @@
...
+    std::cout << "Generating a new encryption key...\n";
+    utils::crypto::Bytes encryption_key = utils::crypto::generateKey();
+    writeEncryptionKeyToBootstrapFile(encryption_key);
+    std::cout << "Wrote the new encryption key to " << bootstrapFilePath() << '\n';
+    return encryption_key;
+  }
+}
+
+std::string EncryptConfig::hexDecodeAndValidateKey(const std::string& key) const {
+  std::string binary_key = utils::StringUtils::from_hex(key);
```
adamdebreceni, line 104: I'm not sure what the expected behavior would be, but be aware that `from_hex` simply skips non-hex characters in the string, so `from_hex("aT_bZa9") == "\xab\xa9"`

fgerlits: I think that's fine.  It may even be useful, if we want to use an externally-generated key, to be able to format it using separators, eg. "aa411f289c91685e-f9d5a9e5a4fad939-3ff4c7a78ab97848-4323488caed7a9ab".

libminifi/include/properties/Properties.h
```diff
@@ -65,6 +67,13 @@ class Properties {
    */
   bool get(const std::string &key, std::string &value) const;
 
+  /**
+   * Returns the config value.
+   * @param key key to look up
+   * @returns the value if found, otherwise nullopt
+   */
+  utils::optional<std::string> get(const std::string &key) const;
```
adamdebreceni, line 75: 👍

main/MiNiFiMain.cpp
```diff
@@ -208,6 +238,10 @@ int main(int argc, char **argv) {
   configure->setHome(minifiHome);
   configure->loadConfigureFile(DEFAULT_NIFI_PROPERTIES_FILE);
 
+  if (containsEncryptedProperties(*configure)) {
+    decryptSensitiveProperties(*configure, minifiHome, *logger);
```
adamdebreceni, line 242: as I see, we update the `configure` object in `decryptSensitiveProperties `, one possible problem with this, is that on a flow update (from a C2 agent) the configurations file is persisted by default (to store the new flow url) this could inadvertently leak the decrypted values

fgerlits: Ouch.  Yes, that is a serious problem.

lordgamez (commented):
Just some minor comments, LGTM.

encrypt-config/ConfigFile.cpp
```diff
@@ -0,0 +1,166 @@
...
+}  // namespace
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace encrypt_config {
+
+ConfigLine::ConfigLine(std::string line) : line_(line) {
+  line = utils::StringUtils::trim(line);
```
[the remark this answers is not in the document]
fgerlits: I feel fairly strongly that we should only update the properties we are encrypting, and not touch the rest of the configuration file.  If I run the script and it says "encrypted 1 property", then the diff between the old and new file should be something like
```
-nifi.rest.api.password=password123
+nifi.rest.api.password=iARpRfcyawNeEkEYLcLfatLUABuf9Nq3||OKHK6goZgGPskfKBow7CARSwdhyExPr1xEzE
+nifi.rest.api.password.protected=xsalsa20poly1305
```
and nothing else.

encrypt-config/ConfigFileEncryptor.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+#include "utils/StringUtils.h"
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace encrypt_config {
+
+int encryptSensitivePropertiesInFile(ConfigFile& config_file, const utils::crypto::Bytes& encryption_key) {
```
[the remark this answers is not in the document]
fgerlits: I am not totally convinced: to me, `int` means "a smallish integer, don't care too much which type", which is what I meant here; but I have changed it to `uint32_t`.

encrypt-config/EncryptConfig.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+#include <stdexcept>
+
+#include "ConfigFile.h"
+#include "ConfigFileEncryptor.h"
+#include "cxxopts.hpp"
+#include "utils/file/FileUtils.h"
+#include "utils/OptionalUtils.h"
+
+namespace {
```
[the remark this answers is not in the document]
fgerlits: That reference says "do _not_ use blank lines in general, but you can use them in certain cases", and it's not clear this is one of those cases, but OK.

encrypt-config/ConfigFile.cpp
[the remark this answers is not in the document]
lordgamez: I understand, it definitely removes any clutter from the diff file.

encrypt-config/EncryptConfig.cpp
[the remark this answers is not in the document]
lordgamez: Thanks, I was referring to the last use case, where this could help with the readability a bit.

`Blank lines immediately inside a declaration of a namespace or block of namespaces may help readability by visually separating the load-bearing content from the (largely non-semantic) organizational wrapper.`

lordgamez (approved):

cmake/BundledLibSodium.cmake
```diff
@@ -0,0 +1,95 @@
...
+    set(PASSTHROUGH_VARIABLES ${PASSTHROUGH_VARIABLES} "-DEXPORTED_LIBSODIUM_INCLUDE_DIRS=${LIBSODIUM_INCLUDE_DIRS}" CACHE STRING "" FORCE)
+    set(PASSTHROUGH_VARIABLES ${PASSTHROUGH_VARIABLES} "-DEXPORTED_LIBSODIUM_LIBRARIES=${LIBSODIUM_LIBRARIES}" CACHE STRING "" FORCE)
+
+    # Create imported targets
+    file(MAKE_DIRECTORY ${LIBSODIUM_INCLUDE_DIRS})
+
+    add_library(libsodium STATIC IMPORTED)
+    set_target_properties(libsodium PROPERTIES IMPORTED_LOCATION "${LIBSODIUM_LIBRARIES}")
+    add_dependencies(libsodium libsodium-external)
+    set_property(TARGET libsodium APPEND PROPERTY INTERFACE_INCLUDE_DIRECTORIES "${LIBSODIUM_INCLUDE_DIRS}")
```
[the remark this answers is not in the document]
fgerlits: Both of these require CMake >= 3.11 (ticket: https://gitlab.kitware.com/cmake/cmake/-/issues/15689, PR: https://gitlab.kitware.com/cmake/cmake/-/merge_requests/1264).

Do you think it's worth bumping the minimum CMake version for this?  Ubuntu 18.04 comes with cmake 3.10 by default.

CMakeLists.txt
```diff
@@ -245,6 +245,12 @@ if (NOT OPENSSL_OFF)
 	set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DOPENSSL_SUPPORT")
 endif()
 
+# libsodium
+include(BundledLibSodium)
+use_bundled_libsodium("${CMAKE_CURRENT_SOURCE_DIR}" "${CMAKE_CURRENT_BINARY_DIR}")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DSODIUM_STATIC=1")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DSODIUM_STATIC=1")
```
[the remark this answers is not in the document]
fgerlits: requires CMake >= 3.11, see below

encrypt-config/EncryptConfig.cpp
fgerlits: I have added a comment.

encrypt-config/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+file(GLOB ENCRYPT_CONFIG_FILES  "*.cpp")
+add_executable(encrypt-config "${ENCRYPT_CONFIG_FILES}")
+target_include_directories(encrypt-config PRIVATE ../libminifi/include  ../thirdparty/cxxopts/include)
+target_wholearchive_library(encrypt-config minifi)
+target_link_libraries(encrypt-config cxxopts)
```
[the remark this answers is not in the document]
fgerlits: fixed

encrypt-config/ConfigFile.cpp
```diff
@@ -0,0 +1,166 @@
...
+#include "ConfigFile.h"
+
+#include <algorithm>
+#include <fstream>
+#include <utility>
+
+#include "utils/StringUtils.h"
+
+namespace {
+constexpr std::array<const char*, 2> DEFAULT_SENSITIVE_PROPERTIES{"nifi.security.client.pass.phrase",
```
alopresto, line 27: Is there a canonical enumeration of the default properties somewhere?

szaszm: You can find all of the properties in `libminifi/include/properties/Configure.h`.

encrypt-config/tests/ConfigFileEncryptorTests.cpp
```diff
@@ -0,0 +1,128 @@
...
+namespace minifi {
+namespace encrypt_config {
+
+// NOTE(fgerlits): these ==/!= operators are in the test file on purpose, and should not be part of production code,
+// as they take a varying amount of time depending on which character in the line differs, so they would open up
+// our code to timing attacks.  If you need == in production code, make sure to compare all pairs of chars/lines.
+bool operator==(const ConfigLine& left, const ConfigLine& right) { return left.getLine() == right.getLine(); }
+bool operator!=(const ConfigLine& left, const ConfigLine& right) { return !(left == right); }
+bool operator==(const ConfigFile& left, const ConfigFile& right) { return left.config_lines_ == right.config_lines_; }
+bool operator!=(const ConfigFile& left, const ConfigFile& right) { return !(left == right); }
```
szaszm, line 57: Is this a real risk? `ConfigLine` and `ConfigFile` contain the properties file with the sensitive properties in ciphertext, so it's not sensitive IMO.

I'm not asking this to be moved back, just want to know how real the risk is.

fgerlits: ConfigFile contains the plaintext initially, and it gets encrypted later.  I don't think there is a risk now, but we want to do the right thing in case this gets exposed to users later.  I think this is more about principles: if a string can contain sensitive information, remember not to compare it using the default `==`.

encrypt-config/EncryptConfig.cpp
```diff
@@ -0,0 +1,153 @@
...
+        << "is invalid; delete it to generate a new key.";
+    throw std::runtime_error{error.str()};
+  }
+}
+
+void EncryptConfig::writeEncryptionKeyToBootstrapFile(const utils::crypto::Bytes& encryption_key) const {
+  std::string key_encoded = utils::StringUtils::to_hex(utils::crypto::bytesToString(encryption_key));
+  encrypt_config::ConfigFile bootstrap_file{std::ifstream{bootstrapFilePath()}};
+
+  if (bootstrap_file.getValue(ENCRYPTION_KEY_PROPERTY_NAME)) {
```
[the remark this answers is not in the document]
fgerlits: I have added `hasValue`, although I'm not sure if the better readability compensates for the increased amount of code.

encrypt-config/tests/ConfigFileTests.cpp
```diff
@@ -0,0 +1,224 @@
...
+  // whitespace is preserved in the key, but not in the value
+  REQUIRE(can_update_value("nifi.some.key= some_value", "some_value", "nifi.some.key=some_value"));
+  REQUIRE(can_update_value("nifi.some.key = some_value  ", "some_value", "nifi.some.key =some_value"));
+  REQUIRE(can_update_value("  nifi.some.key=some_value", "some_value", "  nifi.some.key=some_value"));
+  REQUIRE(can_update_value("  nifi.some.key =\tsome_value\r", "some_value", "  nifi.some.key =some_value"));
+}
+
+TEST_CASE("ConfigFile creates an empty object from a nonexistent file", "[encrypt-config][constructor]") {
+  ConfigFile test_file{std::ifstream{"resources/nonexistent-minifi.properties"}};
+  REQUIRE(test_file.size() == 0);
```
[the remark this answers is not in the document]
fgerlits: In the case of the `bootstrap.conf` file, I wanted to handle the "file does not exist" case and the "file exists but does not contain the `nifi.bootstrap.sensitive.key` property" case in the same way.

If the `minifi.properties` file either does not exist or is empty, the user will get an error message (EncryptConfig.cpp:135).

encrypt-config/tests/ConfigFileTests.cpp
```diff
@@ -0,0 +1,224 @@
...
+  }
+}
+
+TEST_CASE("ConfigFile can add a new setting after an existing setting", "[encrypt-config][insertAfter]") {
+  ConfigFile test_file{std::ifstream{"resources/minifi.properties"}};
+
+  SECTION("valid key") {
+    test_file.insertAfter("nifi.rest.api.password", "nifi.rest.api.password.protected", "XChaCha20-Poly1305");
+    REQUIRE(test_file.size() == 102);
+    REQUIRE(test_file.getValue("nifi.rest.api.password.protected") == utils::optional<std::string>{"XChaCha20-Poly1305"});
```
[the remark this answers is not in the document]
fgerlits: fixed

szaszm (approved):

main/MiNiFiMain.cpp
fgerlits: fixed in https://github.com/apache/nifi-minifi-cpp/pull/914/commits/10259562e26b6be5c8dbd6ba200118dc43f0d1e5

libminifi/src/Decryptor.cpp
```diff
@@ -0,0 +1,54 @@
...
+namespace nifi {
+namespace minifi {
+
+utils::optional<minifi::Decryptor> Decryptor::create(const std::string& minifi_home) {
+  minifi::Properties bootstrap_conf;
+  bootstrap_conf.setHome(minifi_home);
+  bootstrap_conf.loadConfigureFile(DEFAULT_NIFI_BOOTSTRAP_FILE);
+  return bootstrap_conf.getString(CONFIG_ENCRYPTION_KEY_PROPERTY_NAME)
+      | utils::map([](const std::string& encryption_key_hex) { return utils::StringUtils::from_hex(encryption_key_hex); })
+      | utils::map([](const std::string& encryption_key) { return utils::crypto::stringToBytes(encryption_key); })
```
szaszm, line 47: Sadly overloaded functions (like `from_hex`) and constructor calls (like `Decryptor`) are not [`Callable`](https://en.cppreference.com/w/cpp/named_req/Callable), but function pointers are, so this line can be simplified:
```suggestion
      | utils::map(&utils::crypto::stringToBytes)
```

fgerlits: done: https://github.com/apache/nifi-minifi-cpp/pull/914/commits/54f405621368410d9ac27947d455eb345dc641fe

libminifi/include/properties/Properties.h
```diff
@@ -65,7 +65,7 @@ class Properties {
    * @param value value in which to place the map's stored property value
    * @returns true if found, false otherwise.
    */
-  bool get(const std::string &key, std::string &value) const;
+  bool getString(const std::string &key, std::string &value) const;
```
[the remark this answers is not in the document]
fgerlits: This way it is easier to see which function is getting called.  We don't need `get()` to be virtual, and hiding a non-virtual function isn't nice.  Also, Properties has a `getInt()` method, so `getString()` makes sense.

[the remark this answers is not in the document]
szaszm: > We don't need get() to be virtual, and hiding a non-virtual function isn't nice.

I don't think hiding is a big issue. What I don't like is that a change in a derived class forces the parent class to change in an undesirable way.

> Properties has a getInt() method, so getString() makes sense.

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Ro-equivalent

Also, getInt isn't too important, just a convenience overload from the times before optional and map. `Properties` is a string key-string value store, so I as a user shouldn't need to say `getString` when retrieving a value, because I already said that by using the `Properties` class.

encrypt-config/ConfigFile.cpp
```diff
@@ -0,0 +1,171 @@
...
+int ConfigFile::erase(const std::string& key) {
+  auto has_this_key = [&key](const ConfigLine& line) { return line.getKey() == key; };
+  auto new_end = std::remove_if(config_lines_.begin(), config_lines_.end(), has_this_key);
+  auto num_removed = std::distance(new_end, config_lines_.end());
+  config_lines_.erase(new_end, config_lines_.end());
+  return gsl::narrow<int>(num_removed);
+}
+
+void ConfigFile::writeTo(const std::string& file_path) const {
+  std::ofstream file{file_path};
```
szaszm, line 131: `std::ofstream` doesn't throw on error, just sets error bits by default. This hides errors and causes me as a user to not realize that despite the success output, nothing has changes because I was not running as admin.

Luckily its error reporting behavior can be configured.

```suggestion
  std::ofstream file{file_path};
  file.exceptions(std::ios::failbit | std::ios::badbit);
```

encrypt-config/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+# specific language governing permissions and limitations
+# under the License.
+
+file(GLOB ENCRYPT_CONFIG_FILES  "*.cpp")
+add_executable(encrypt-config "${ENCRYPT_CONFIG_FILES}")
+target_include_directories(encrypt-config PRIVATE ../libminifi/include  ../thirdparty/cxxopts/include)
+target_wholearchive_library(encrypt-config minifi)
+target_link_libraries(encrypt-config cxxopts libsodium)
+set_target_properties(encrypt-config PROPERTIES OUTPUT_NAME encrypt-config)
+install(TARGETS encrypt-config RUNTIME DESTINATION bin COMPONENT bin)
```
szaszm, line 25: Multi configuration generators (visual studio) don't save the last build type in the cache, so cpack apparently has no way of knowing that it has to get the binary from the Debug/RelWithDebInfo directory and defaults to Release (the first one AFAIK). Not sure why the issue only happens with `install(TARGETS)`, but the fix is to append pass `-C <configuration>` to cpack.

Could you add this to `win_build_vs.bat`?

fgerlits: yes, I will, thank you!

encrypt-config/ConfigFile.cpp
fgerlits: done: https://github.com/apache/nifi-minifi-cpp/pull/914/commits/ec7277ca5be16565d25738d23f898d7e46608c5c

libminifi/include/properties/Properties.h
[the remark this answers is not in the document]
fgerlits: I think that by naming both functions in the parent and child classes `get()`, we would be setting a trap for our future selves, and for anyone who will work on this code.  I agree `getString()` is not a great name, but I would prefer to rename it to anything other than `get()`.  How about `value()`, `getProperty()`, `getPropertyValue()` or something similar?

arpadboda (approved):

encrypt-config/ConfigFileEncryptor.cpp
```diff
@@ -0,0 +1,64 @@
...
+  int num_properties_encrypted = 0;
+
+  for (const auto& property_key : config_file.getSensitiveProperties()) {
+    utils::optional<std::string> property_value = config_file.getValue(property_key);
+    if (!property_value) { continue; }
+
+    std::string encryption_type_key = property_key + ".protected";
+    utils::optional<std::string> encryption_type = config_file.getValue(encryption_type_key);
+
+    if (!encryption_type || encryption_type->empty() || *encryption_type == "plaintext") {
```
adamdebreceni, line 41: what if the user accidentally tries to encrypt a file that already contains encrypted values with a different encryption key, should they get a warning?

fgerlits: You mean if the user removed/altered the encryption key but forgot to reset the property values to their unencrypted state, and then re-ran the script?  The script will tell them that no new properties were encrypted; I think that is enough of a warning.

adamdebreceni (approved):

libminifi/include/properties/Properties.h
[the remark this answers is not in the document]
szaszm: After discussion, we agreed to leave it as is, because all of the alternatives were suboptimal.

szaszm (approved):

PR #915, opened 2020-09-30
MINIFICPP-1376 Create PutS3Object processor

lordgamez:
Implementation of PutS3Object processor. The purpose of the processor is to upload an S3 object created from a flow file to AWS storage.

Jira issue: https://issues.apache.org/jira/browse/MINIFICPP-1376

---------------------------------------------------------------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/aws/AWSLoader.h
```diff
@@ -16,43 +16,48 @@
...
   }
 
   /**
    * Gets the class name for the object
    * @return class name for the processor.
    */
-  virtual std::vector<std::string> getClassNames() override{
+  std::vector<std::string> getClassNames() override {
     std::vector<std::string> class_names;
     class_names.push_back("AWSCredentialsService");
```
adamdebreceni, line 52: don't we need to add `PutS3Object` to the `class_names` vector as well?

lordgamez: I suppose we do, I will have to check how it worked without it even in EFM. Fixed in [d2fc8cd](https://github.com/apache/nifi-minifi-cpp/pull/915/commits/d2fc8cdf064177ec7188ff4e80228724c1711fb2)

extensions/aws/processors/PutS3Object.h
```diff
@@ -0,0 +1,188 @@
...
+
+    ReadCallback(uint64_t flow_size, const minifi::aws::s3::PutObjectRequestParameters& options, aws::s3::S3WrapperBase* s3_wrapper)
+      : flow_size_(flow_size)
+      , options_(std::move(options))
+      , s3_wrapper_(s3_wrapper) {
+    }
+
+    ~ReadCallback() = default;
+
+    int64_t process(std::shared_ptr<io::BaseStream> stream) {
```
adamdebreceni, line 131: we should add `override` to `~ReadCallback` and `process`

extensions/aws/processors/PutS3Object.cpp
```diff
@@ -0,0 +1,487 @@
...
+  std::string value;
+  if (context->getProperty(EndpointOverrideURL, value, flow_file) && !value.empty()) {
+    s3_wrapper_->setEndpointOverrideUrl(value);
+    logger_->log_debug("PutS3Object: Endpoint Override URL [%d]", value);
+  }
+
+  return setAccessControl(context, flow_file);
+}
+
+void PutS3Object::onTrigger(const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSession> &session) {
```
adamdebreceni, line 449: some processors are now receiving a `Batch Size` property to consume more flowFiles (but at most `Batch Size`) in a single trigger, this processor looks like a good candidate for such a behavior, it could also be implemented later in a follow-up PR, what do you think?

extensions/aws/processors/PutS3Object.cpp
```diff
@@ -0,0 +1,487 @@
...
+  relationships.insert(Success);
+  setSupportedRelationships(relationships);
+}
+
+minifi::utils::optional<Aws::Auth::AWSCredentials> PutS3Object::getAWSCredentialsFromControllerService(const std::shared_ptr<core::ProcessContext> &context) const {
+  std::string service_name;
+  if (context->getProperty(AWSCredentialsProviderService.getName(), service_name) && !service_name.empty()) {
+    std::shared_ptr<core::controller::ControllerService> service = context->getControllerService(service_name);
+    if (nullptr != service) {
+      auto aws_credentials_service = std::static_pointer_cast<minifi::aws::controllers::AWSCredentialsService>(service);
```
adamdebreceni, line 226: is `static_pointer_cast` a safe choice here? we are fetching (instantiating maybe?) a ControllerService based on a configuration parameter, could that contain a value that results in a ControllerService other than AWSCredentialsService?

extensions/aws/processors/PutS3Object.cpp
```diff
@@ -0,0 +1,487 @@
...
+  for (const auto& user : minifi::utils::StringUtils::split(comma_separated_list, ",")) {
+    if (is_first) {
+      is_first = false;
+    } else {
+      result_list += ", ";
+    }
+
+    auto trimmed_user = minifi::utils::StringUtils::trim(user);
+    static const std::regex email_pattern("(\\w+)(\\.|_)?(\\w*)@(\\w+)(\\.(\\w+))+");
+    if (std::regex_match(trimmed_user, email_pattern)) {
```
adamdebreceni, line 372: doesn't `\w` also match `_`?
also note that this won't match all valid email addresses (if that is the purpose), the regex for that is quite... a mouthful

extensions/aws/processors/PutS3Object.cpp
lordgamez: I think it's a good idea as an upload could take a some time depending on the latency/region/object size so the scheduling and load balancing could be optimized with this. I'm still discussing some requirements with @arpadboda I'll check with him about this as well.

extensions/aws/processors/PutS3Object.cpp
lordgamez: That can be a possible user error in case the user gives the name of another service here. It can cause an error which could be avoided in this case, I'll use a dynamic cast here instead.

extensions/aws/processors/PutS3Object.cpp
lordgamez: Yes, you are right and it is unnecessary to complicate it in this case.

lordgamez: Fixed in [2e71a70](https://github.com/apache/nifi-minifi-cpp/pull/915/commits/2e71a707a31f5564360c40c7bfdd38c3d97302ef)

extensions/aws/processors/PutS3Object.cpp
lordgamez: Fixed in [2e71a70](https://github.com/apache/nifi-minifi-cpp/pull/915/commits/2e71a707a31f5564360c40c7bfdd38c3d97302ef)

extensions/aws/processors/PutS3Object.h
lordgamez: As there is no need for override in the destructor I removed it and added the override to the process in [2e71a70](https://github.com/apache/nifi-minifi-cpp/pull/915/commits/2e71a707a31f5564360c40c7bfdd38c3d97302ef)

extensions/aws/processors/PutS3Object.cpp
adamdebreceni: the regex rejects a number of otherwise valid email addresses, e.g. the "local-part" (part before the '@' character) may contain "!~#" and others, it could even be quoted or contain comments, the domain part is similarly complex
I don't know what the requirement is here, do we aim to catch all valid email addresses? (if not, could checking for the '@' character be enough?)

extensions/aws/processors/PutS3Object.cpp
```diff
@@ -0,0 +1,489 @@
...
+  relationships.insert(Success);
+  setSupportedRelationships(relationships);
+}
+
+minifi::utils::optional<Aws::Auth::AWSCredentials> PutS3Object::getAWSCredentialsFromControllerService(const std::shared_ptr<core::ProcessContext> &context) const {
+  std::string service_name;
+  if (context->getProperty(AWSCredentialsProviderService.getName(), service_name) && !service_name.empty()) {
+    std::shared_ptr<core::controller::ControllerService> service = context->getControllerService(service_name);
+    if (nullptr != service) {
+      auto aws_credentials_service = std::dynamic_pointer_cast<minifi::aws::controllers::AWSCredentialsService>(service);
```
adamdebreceni, line 226: should we omit the `nullptr != service` check? `dynamic_pointer_cast` should be safe to use on an empty `shared_ptr`

extensions/aws/processors/PutS3Object.cpp
lordgamez: The goal here would be to distinguish email addresses and AWS ids . I am not sure if quotes and comments should be accepted in this case, but for our goal I suppose checking the '@' character should be enough.

adamdebreceni: is the format of AWS id specified? or is it "not email address"?

extensions/aws/processors/PutS3Object.cpp
lordgamez: But in that case the `aws_credentials_service` would be nullptr as well after the `dynamic_pointer_cast` and we would call on the nullptr in the next line.

adamdebreceni: yes, but we are doing a check there too, ultimately it is not crucial to make this change

extensions/aws/processors/PutS3Object.cpp
lordgamez: In this case as we only accept email addresses and AWS ids so currently I only checked if it is an email address or not in which case we assume it is an AWS id. The canonical user id is not really specified it is only defined [here](https://docs.aws.amazon.com/general/latest/gr/acct-identifiers.html) as "An alpha-numeric identifier, such as 79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be, that is an obfuscated form of the AWS account ID.". It seems to me it is a 64 long hex string literal, but I am not sure if the length is fixed as there are some questions about it [here for example](https://forums.aws.amazon.com/thread.jspa?threadID=286019).

extensions/aws/processors/PutS3Object.cpp
lordgamez: I discussed with @arpadboda and we do not really have any benefit of the batch processing in case of an `EVENT_DRIVEN` processor (which should be a main use case) as we still have to do every processing step on a single flow file basis. Unfortunately [S3 does not support batch upload of objects](https://stackoverflow.com/questions/15050146/is-it-possible-to-perform-a-batch-upload-to-amazon-s3) so cannot benefit from a single request batch upload either.

extensions/aws/processors/PutS3Object.cpp
lordgamez: Checking only for '@' for email addresses in [967b355](https://github.com/apache/nifi-minifi-cpp/pull/915/commits/967b355ca776f8a2dba705e390a91bc6aa5380a3)

extensions/aws/processors/PutS3Object.h
```diff
@@ -0,0 +1,195 @@
...
+  constexpr const char *CA_CENTRAL_1 = "ca-central-1";
+  constexpr const char *AF_SOUTH_1 = "af-south-1";
+  constexpr const char *AP_EAST_1 = "ap-east-1";
+  constexpr const char *AP_NORTHEAST_3 = "ap-northeast-3";
+  constexpr const char *CN_NORTHWEST_1 = "cn-northwest-1";
+  constexpr const char *EU_SOUTH_1 = "eu-south-1";
+  constexpr const char *EU_WEST_3 = "eu-west-3";
+  constexpr const char *EU_NORTH_1 = "eu-north-1";
+  constexpr const char *ME_SOUTH_1 = "me-south-1";
+  constexpr const char *US_GOV_EAST_1 = "us-gov-east-1";
```
adamdebreceni, line 71: 1. I think we do not indent inside namespaces (although I do not think that it hinders readability here)
2. also do you think that putting the regions in lexicographical order could help us add-remove (or diff) the list in the future?

lordgamez: Fixed in [dff0c8a](https://github.com/apache/nifi-minifi-cpp/pull/915/commits/dff0c8a327c5c6e13a4f6008314986e3bd57b441)

adamdebreceni (approved):

PROCESSORS.md
```diff
@@ -1125,3 +1126,40 @@ In the list below, the names of required properties appear in bold. Any other pr
...
+### Description
+
+Puts FlowFiles to an Amazon S3 Bucket. The upload uses either the PutS3Object method. The PutS3Object method send the file in a single synchronous call, but it has a 5GB size limit. Larger files sent using the multipart upload methods are currently not supported. The AWS libraries select an endpoint URL based on the AWS region, but this can be overridden with the 'Endpoint Override URL' property for use with other S3-compatible endpoints. The S3 API specifies that the maximum file size for a PutS3Object upload is 5GB.
+### Properties
+
+In the list below, the names of required properties appear in bold. Any other properties (not in bold) are considered optional. The table also indicates any default values, and whether a property supports the NiFi Expression Language.
+
+| Name | Default Value | Allowable Values | Description |
+| - | - | - | - |
+|**Object Key**|${filename}||The key of the S3 object<br/>**Supports Expression Language: true**|
```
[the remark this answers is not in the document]
lordgamez: Sounds good to me, I am not familiar how common are agents without EL support, but it's reasonable to support those as well. In that case the access control properties should lose the default values too, as if those are invalid the put operation fails on AWS. Those properties are not mandatory so we can have empty defaults if that sounds okay?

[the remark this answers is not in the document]
szaszm: I'm not familiar enough with S3 to know this, but if empty is a reasonable default that's usable in simple setups, then I support it.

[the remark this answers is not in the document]
lordgamez: These values control the permissions of the uploaded object. We can give email addresses or AWS IDs that should have read/write/full control permissions on the uploaded object, to which by default only the owner (the uploader's account) have access if no other permissions are given. I think that should be okay in our case. The current default values like ${s3.permissions.full.users} are from the Nifi version of PutS3Object which can probably come from other AWS processors, but we don't have those yet, so we can remove those.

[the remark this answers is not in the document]
szaszm: Interesting point. NiFi compatibility would be nice, but in NiFi, expression language is not an optional feature. That made me neutral, I'm fine with either option.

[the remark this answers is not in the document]
lordgamez: I searched for the `s3.permission` parameters in NiFi if any of the processors have them as outputs, but I could not find any, only as the default input values of PutS3Object. I am not sure, where these defaults come from, but I suppose in this case it would be more advantageous to have support for agents without EL support if we cannot find any other references to them. I removed the EL dependent default values in [bc6b752](https://github.com/apache/nifi-minifi-cpp/pull/915/commits/bc6b752e03ddd29dc150cbb598414cb332ca5592)

cmake/BundledAwsSdkCpp.cmake
```diff
@@ -23,8 +23,8 @@ function(use_bundled_libaws SOURCE_DIR BINARY_DIR)
         set(SUFFIX "a")
     endif()
     set(BYPRODUCTS
-            "lib/libaws-cpp-sdk-core.${SUFFIX}"
-            "lib/libaws-cpp-sdk-s3.${SUFFIX}")
+            "${CMAKE_INSTALL_LIBDIR}/libaws-cpp-sdk-core.${SUFFIX}"
```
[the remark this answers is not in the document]
lordgamez: Added it in [8a05f1a](https://github.com/apache/nifi-minifi-cpp/pull/915/commits/8a05f1a3821aabe6a72607d9dae28a418dd9f57d) and I also found some problems with the AWS build on Windows. Fixed them in [42eb471](https://github.com/apache/nifi-minifi-cpp/pull/915/commits/42eb471011efb7c93e13a553d6a04f129ccd1439)

extensions/aws/processors/PutS3Object.cpp
```diff
@@ -0,0 +1,495 @@
...
+
+  if (!context->getProperty(StorageClass.getName(), put_s3_request_params_.storage_class)
+      || put_s3_request_params_.storage_class.empty()
+      || STORAGE_CLASSES.find(put_s3_request_params_.storage_class) == STORAGE_CLASSES.end()) {
+    throw Exception(PROCESS_SCHEDULE_EXCEPTION, "Storage Class property missing or invalid");
+  }
+  logger_->log_debug("PutS3Object: Storage Class [%s]", put_s3_request_params_.storage_class);
+
+  std::string value;
+  if (!context->getProperty(Region.getName(), value) || value.empty() || REGIONS.find(value) == REGIONS.end()) {
```
arpadboda, line 323: Just fancy: in case you use set, you can rely on count instead of comparing the result of find to end.

extensions/aws/processors/PutS3Object.cpp
```diff
@@ -0,0 +1,495 @@
...
+}
+
+std::string PutS3Object::parseAccessControlList(const std::string &comma_separated_list) const {
+  std::string result_list;
+  bool is_first = true;
+  for (const auto& user : minifi::utils::StringUtils::split(comma_separated_list, ",")) {
+    if (is_first) {
+      is_first = false;
+    } else {
+      result_list += ", ";
```
arpadboda, line 354: This is already implemented in join. 

I bit more simple way of doing this:
```
auto users : minifi::utils::StringUtils::split(comma_separated_list, ",")
for(auto& user: users) {
      auto trimmed_user = minifi::utils::StringUtils::trim(user);
    if (trimmed_user.find('@') != std::string::npos) {
      user = "emailAddress=\"" + trimmed_user + "\"";
    } else {
      user += "id=" + trimmed_user;
    }
}
return minifi::utils::StringUtils::join(", ", users);
```

arpadboda (commented):
Mosty looks good, added some minor comments

extensions/aws/processors/PutS3Object.h
```diff
@@ -0,0 +1,197 @@
...
+    int64_t process(std::shared_ptr<io::BaseStream> stream) override {
+      if (flow_size_ > MAX_SIZE) {
+        return -1;
+      }
+      std::vector<uint8_t> buffer;
+      auto data_stream = std::make_shared<std::stringstream>();
+      buffer.reserve(BUFFER_SIZE);
+      read_size_ = 0;
+      while (read_size_ < flow_size_) {
+        auto next_read_size = flow_size_ - read_size_ < BUFFER_SIZE ? flow_size_ - read_size_ : BUFFER_SIZE;
```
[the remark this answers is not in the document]
lordgamez: Fixed in [e994558](https://github.com/apache/nifi-minifi-cpp/pull/915/commits/e994558ba8904488b0e3489d51236fb7d4e31ce1)

extensions/aws/utils/AWSSdkLogger.cpp
```diff
@@ -0,0 +1,105 @@
...
+    case Aws::Utils::Logging::LogLevel::Fatal:
+      logger_->log_error("[%s] %s", tag, format_str);
+      break;
+    default:
+      break;
+  }
+}
+
+void AWSSdkLogger::LogStream(Aws::Utils::Logging::LogLevel log_level, const char* tag, const Aws::OStringStream &message_stream) {
+  switch (log_level) {
```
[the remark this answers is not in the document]
lordgamez: Fixed in [e994558](https://github.com/apache/nifi-minifi-cpp/pull/915/commits/e994558ba8904488b0e3489d51236fb7d4e31ce1)

extensions/aws/processors/PutS3Object.cpp
lordgamez: Fixed in [e994558](https://github.com/apache/nifi-minifi-cpp/pull/915/commits/e994558ba8904488b0e3489d51236fb7d4e31ce1)

extensions/aws/processors/PutS3Object.cpp
lordgamez: Fixed in [e994558](https://github.com/apache/nifi-minifi-cpp/pull/915/commits/e994558ba8904488b0e3489d51236fb7d4e31ce1)

arpadboda (approved):

PR #920, opened 2020-10-07
MINIFICPP-1296 - All tests should use volatile state storage

adam-markovics:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/standard-processors/controllers/UnorderedMapKeyValueStoreService.h
+++ b/extensions/standard-processors/controllers/UnorderedMapKeyValueStoreService.h
@@ -29,6 +29,7 @@
 #include "core/logging/Logger.h"
 #include "core/logging/LoggerConfiguration.h"
 #include "core/Resource.h"
+#include "controllers/keyvalue/PersistableKeyValueStoreService.h"
 
 namespace org {
 namespace apache {
@@ -36,12 +37,13 @@ namespace nifi {
 namespace minifi {
 namespace controllers {
 
-class UnorderedMapKeyValueStoreService : virtual public KeyValueStoreService {
+/// Key-value store serice purely in RAM without disk usage
+class UnorderedMapKeyValueStoreService : virtual public PersistableKeyValueStoreService {
  public:
   explicit UnorderedMapKeyValueStoreService(const std::string& name, utils::Identifier uuid = utils::Identifier());
   explicit UnorderedMapKeyValueStoreService(const std::string& name, const std::shared_ptr<Configure>& configuration);
 
-  virtual ~UnorderedMapKeyValueStoreService();
+  ~UnorderedMapKeyValueStoreService() override;
 
   bool set(const std::string& key, const std::string& value) override;
 
@@ -54,6 +56,10 @@ class UnorderedMapKeyValueStoreService : virtual public KeyValueStoreService {
   bool clear() override;
 
   bool update(const std::string& key, const std::function<bool(bool /*exists*/, std::string& /*value*/)>& update_func) override;
+
+  bool persist() override {
+    return true;
+  }
 
  protected:
   std::unordered_map<std::string, std::string> map_;
--- a/extensions/standard-processors/controllers/UnorderedMapPersistableKeyValueStoreService.h
+++ b/extensions/standard-processors/controllers/UnorderedMapPersistableKeyValueStoreService.h
@@ -48,7 +48,9 @@ class UnorderedMapPersistableKeyValueStoreService : public AbstractAutoPersistin
   static core::Property File;
 
   void onEnable() override;
-  void initialize() override;
+  void initialize() override {
+    initializeNonVirtual();
+  }
   void notifyStop() override;
 
   bool set(const std::string& key, const std::string& value) override;
@@ -59,7 +61,9 @@ class UnorderedMapPersistableKeyValueStoreService : public AbstractAutoPersistin
 
   bool update(const std::string& key, const std::function<bool(bool /*exists*/, std::string& /*value*/)>& update_func) override;
 
-  bool persist() override;
+  bool persist() override {
+    persistNonVirtual();
+  }
 
  protected:
   using AbstractAutoPersistingKeyValueStoreService::getImpl;
@@ -75,10 +79,12 @@ class UnorderedMapPersistableKeyValueStoreService : public AbstractAutoPersistin
 
   bool load();
 
-  std::string escape(const std::string& str);
   bool parseLine(const std::string& line, std::string& key, std::string& value);
 
  private:
+  void initializeNonVirtual();
+  bool persistNonVirtual();
+
   std::shared_ptr<logging::Logger> logger_;
 };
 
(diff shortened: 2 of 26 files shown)


extensions/standard-processors/controllers/UnorderedMapKeyValueStoreService.h
```diff
@@ -29,19 +29,21 @@
...
 
 namespace org {
 namespace apache {
 namespace nifi {
 namespace minifi {
 namespace controllers {
 
-class UnorderedMapKeyValueStoreService : virtual public KeyValueStoreService {
+/// Key-value store serice purely in RAM without disk usage
+class UnorderedMapKeyValueStoreService : virtual public PersistableKeyValueStoreService {
```
szaszm, line 41: If it's RAM only, then why is it `PersistableKeyValueStoreService`?

extensions/standard-processors/controllers/UnorderedMapPersistableKeyValueStoreService.h
```diff
@@ -59,20 +61,30 @@ class UnorderedMapPersistableKeyValueStoreService : public AbstractAutoPersistin
...
   std::string file_;
 
   bool load();
 
-  std::string escape(const std::string& str);
   bool parseLine(const std::string& line, std::string& key, std::string& value);
 
  private:
+  void initializeNonVirtual();
+  bool persistNonVirtual();
```
szaszm, line 86: Why are these necessary? If subclasses need to call the base class implementation, they can always qualify the call with the base class name to avoid dynamic dispatching,

szaszm (commented):
just a cursory first pass

libminifi/include/utils/ThreadPool.h
```diff
@@ -227,11 +227,10 @@ class ThreadPool {
    * Returns true if a task is running.
    */
   bool isTaskRunning(const std::string &identifier) const {
-    try {
-      return task_status_.at(identifier) == true;
-    } catch (const std::out_of_range &) {
+    const auto iter = task_status_.find(identifier);
```
adamdebreceni, line 230: I know it's old code but I wonder why aren't we locking the `worker_queue_mutex_` or the `manager_mutex_` mutex here, even though we lock at least one of them every other place we access the `task_status_` map

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -497,11 +497,8 @@ bool TailFile::recoverState(const std::shared_ptr<core::ProcessContext>& context
...
   if (state_manager_->get(state_map)) {
-    for (size_t i = 0U;; i++) {
-      std::string name;
-      try {
-        name = state_map.at("file." + std::to_string(i) + ".name");
-      } catch (...) {
+    for (size_t i = 0U;; ++i) {
+      if (state_map.find("file." + std::to_string(i) + ".name") == state_map.end()) {
         break;
       }
```
adamdebreceni, line 503: I am genuinely curious what was the reasoning behind this code, could we be missing something by rewriting it to `std::map<...>::find`?

extensions/standard-processors/tests/unit/TailFileTests.cpp
```diff
@@ -1129,8 +1129,11 @@ TEST_CASE("TailFile finds and finishes the renamed file and continues with the n
   utils::Identifier tail_file_uuid = utils::IdGenerator::getIdGenerator()->generate();
   const core::Relationship success_relationship{"success", "everything is fine"};
 
+  // use persistent state storage that defaults to rocksDB, not volatile
+  const auto configuration = std::make_shared<minifi::Configure>();
```
adamdebreceni, line 1133: is this necessary? `TestController::createPlan` default constructs a `Configure` object if `nullptr` is provided, and by default a `RocksDbPersistableKeyValueStoreService` is constructed

extensions/standard-processors/controllers/UnorderedMapKeyValueStoreService.h
adam-markovics: Because AbstractCoreComponentStateManagerProvider (grandparent of this class) declares a pure virtual persistImpl() method that is implemented in base class PersistableKeyValueStoreService by calling persist(). Basically, persist() is a hook, that can be called with any implementation, doing a null operation when it is volatile.

extensions/standard-processors/controllers/UnorderedMapPersistableKeyValueStoreService.h
adam-markovics: Because initialize() and persist() were called in constructor/destructor and they are virtual methods.

libminifi/include/utils/ThreadPool.h
adam-markovics: Good question, I added locking `worker_queue_mutex_` as it is locked by most other locations of accessing `task_status_`, and its documentation comment says `mutex to  protect task status and delayed queue`.

extensions/standard-processors/processors/TailFile.cpp
adam-markovics: I don't think we would miss anything. `std::unordered_map<...>::at()` only throws if the requested element is not found, the new code is also checking that. My goal was to eliminate exceptions as they are confusing during debugging and should not be used for control flow. And especially eliminating `catch (...)` from wherever possible.

extensions/standard-processors/tests/unit/TailFileTests.cpp
adam-markovics: No, UnorderedMapKeyValueStoreService will be the default from now on for nullptr:
https://github.com/apache/nifi-minifi-cpp/pull/920/files/74dc246506795eb3832547da796c227d03ecb321#diff-530e3557b16fbcc2c012a6ac10efccf62258f51004a7c61d3191334cbd3e3461
In this test case, persistency is more of a requirement, so volatile is not sufficient.

libminifi/test/TestBase.h
```diff
@@ -282,18 +274,10 @@ class TestPlan {
...
-  }
-
   std::shared_ptr<logging::Logger> getLogger() const {
     return logger_;
   }
 
-  std::string getStateDir() {
-    return state_dir_;
-  }
-
```
szaszm, line 296:
```
/home/szaszm/nifi-minifi-cpp-2/extensions/sftp/tests/ListSFTPTests.cpp: In member function ‘void ListSFTPTestsFixture::createPlan(org::apache::nifi::minifi::utils::Identifier*)’:
/home/szaszm/nifi-minifi-cpp-2/extensions/sftp/tests/ListSFTPTests.cpp:92:64: error: ‘using element_type = class TestPlan’ {aka ‘class TestPlan’} has no member named ‘getStateDir’
92 |     const std::string state_dir = plan == nullptr ? "" : plan->getStateDir();
|                                                                ^~~~~~~~~~~
make[2]: *** [extensions/sftp/tests/CMakeFiles/ListSFTPTests.dir/build.make:82: extensions/sftp/tests/CMakeFiles/ListSFTPTests.dir/ListSFTPTests.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:8318: extensions/sftp/tests/CMakeFiles/ListSFTPTests.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

szaszm (changes requested):

szaszm (approved):

adamdebreceni (approved):

adamdebreceni:
the `ListSFTPTests` failed on `ubuntu-16.04-all` on all runs I saw, (here in the PR, and on your own fork run), could we re-run it again (preferably on both places) to verify if the failures are only transient?

szaszm:
ListSFTPTests is still failing after another rerun, so this needs some attention.

szaszm:
35999e0 (on main) broke the build

PR #924, opened 2020-10-13
MINIFICPP-1389 - Upgrade librdkafka to version 1.5.0

hunyadi-dev:
The current version of librdkafka we use has cherry-picked patches applied to it and also does not support transactions required for the implementation of ConsumeKafka processor.


fgerlits:
I think we need to update the LICENSE file with the new versions of the licenses.

hunyadi-dev:
> I think we need to update the LICENSE file with the new versions of the licenses.

Updated the librdkafka license files. However I am not sure if we should also include the licenses from the dependencies of librdkafka itself. Opened a JIRA ticket for the Apache legal team here:
https://issues.apache.org/jira/browse/LEGAL-546

If the current format is not acceptable, I will open a separate correction PR.

szaszm:
I'm pretty sure the answer is yes: we need to include all transitive dependencies' licenses and notices. Otherwise it would be trivial to circumvent these requirements by creating a wrapper project that depends on the software and exposes all of its components.

fgerlits (commented):
We already include these sub-licenses in our LICENSE file (lines 761 - 944), we just need to update them to the newer version (and add/remove some which were added/removed in librdkafka).

hunyadi-dev:
@fgerlits @szaszm updated the transitive licenses.

szaszm (approved):

arpadboda (approved):
LGTM, thanks!

I especially like the deleted patches. :)

LICENSE
```diff
@@ -808,12 +957,27 @@ This product bundles sys/queue.h within librdkafka, which under a 3-clause UC Re
  * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
  * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
  * SUCH DAMAGE.
+ *
+ *  @(#)queue.h 8.5 (Berkeley) 8/20/94
+ * $FreeBSD$
  
 This product bundles regexp.c and regexp.h within librdkafka offered under a public domain license as below.
 
-These libraries are in the public domain (or the equivalent where that is not possible). You can do anything you want with them. You have no legal obligation to do anything else, although I appreciate attribution.
```
fgerlits, line 814: I think we should keep this line (also in msi/LICENSE.txt)

fgerlits (approved):

LICENSE
hunyadi-dev: Updated!

PR #926, opened 2020-10-19
MINIFICPP-1387 - Speed up Identifier to string conversion

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/utils/Id.h
+++ b/libminifi/include/utils/Id.h
@@ -73,6 +73,14 @@ class Identifier {
 
   bool isNil() const;
 
+  // Numerous places query the string representation
+  // just to then forward the temporary to build logs,
+  // streams, or others. Dynamically allocating in these
+  // instances is wasteful as we immediately discard
+  // the result. The difference on the test machine is 8x,
+  // building the representation itself takes 10ns, while
+  // subsequently turning it into a std::string would take
+  // 70ns more.
   UUIDString to_string() const;
 
   static utils::optional<Identifier> parse(const std::string& str);
--- a/libminifi/include/utils/Id.h
+++ b/libminifi/include/utils/Id.h
@@ -51,8 +51,6 @@ namespace nifi {
 namespace minifi {
 namespace utils {
 
-using UUIDString = SmallString<36>;
-
 class Identifier {
   friend struct IdentifierTestAccessor;
   static constexpr const char* UUID_FORMAT_STRING = "%02hhx%02hhx%02hhx%02hhx-%02hhx%02hhx-%02hhx%02hhx-%02hhx%02hhx-%02hhx%02hhx%02hhx%02hhx%02hhx%02hhx";
@@ -81,7 +79,7 @@ class Identifier {
   // building the representation itself takes 10ns, while
   // subsequently turning it into a std::string would take
   // 70ns more.
-  UUIDString to_string() const;
+  SmallString<36> to_string() const;
 
   static utils::optional<Identifier> parse(const std::string& str);
 
--- a/libminifi/include/utils/Id.h
+++ b/libminifi/include/utils/Id.h
@@ -51,6 +51,8 @@ namespace nifi {
 namespace minifi {
 namespace utils {
 
+using UUIDString = SmallString<36>;
+
 class Identifier {
   friend struct IdentifierTestAccessor;
   static constexpr const char* UUID_FORMAT_STRING = "%02hhx%02hhx%02hhx%02hhx-%02hhx%02hhx-%02hhx%02hhx-%02hhx%02hhx-%02hhx%02hhx%02hhx%02hhx%02hhx%02hhx";
@@ -71,7 +73,7 @@ class Identifier {
 
   bool isNil() const;
 
-  SmallString<36> to_string() const;
+  UUIDString to_string() const;
 
   static utils::optional<Identifier> parse(const std::string& str);
 
--- a/libminifi/include/utils/Id.h
+++ b/libminifi/include/utils/Id.h
@@ -23,7 +23,7 @@
 #include <memory>
 #include <string>
 #include <thread>
-#include "FixedLengthString.h"
+#include "SmallString.h"
 
 #ifndef WIN32
 class uuid;
@@ -71,7 +71,7 @@ class Identifier {
 
   bool isNil() const;
 
-  FixedLengthString<36> to_string() const;
+  SmallString<36> to_string() const;
 
   static utils::optional<Identifier> parse(const std::string& str);
 
--- a/libminifi/include/utils/Id.h
+++ b/libminifi/include/utils/Id.h
@@ -54,7 +54,7 @@ namespace utils {
 class Identifier {
   friend struct IdentifierTestAccessor;
   static constexpr const char* UUID_FORMAT_STRING = "%02hhx%02hhx%02hhx%02hhx-%02hhx%02hhx-%02hhx%02hhx-%02hhx%02hhx-%02hhx%02hhx%02hhx%02hhx%02hhx%02hhx";
-  static constexpr const char hex_lut[] = "0123456789abcdef";
+  static constexpr const char* hex_lut = "0123456789abcdef";
 
  public:
   using Data = std::array<uint8_t, 16>;
--- a/libminifi/include/utils/Id.h
+++ b/libminifi/include/utils/Id.h
@@ -23,6 +23,7 @@
 #include <memory>
 #include <string>
 #include <thread>
+#include "FixedLengthString.h"
 
 #ifndef WIN32
 class uuid;
@@ -53,6 +54,7 @@ namespace utils {
 class Identifier {
   friend struct IdentifierTestAccessor;
   static constexpr const char* UUID_FORMAT_STRING = "%02hhx%02hhx%02hhx%02hhx-%02hhx%02hhx-%02hhx%02hhx-%02hhx%02hhx-%02hhx%02hhx%02hhx%02hhx%02hhx%02hhx";
+  static constexpr const char hex_lut[] = "0123456789abcdef";
 
  public:
   using Data = std::array<uint8_t, 16>;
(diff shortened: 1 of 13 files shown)


fgerlits (approved):
LGTM

Do you have some data on how much faster the new code is?  You could attach that to the Jira.

adamdebreceni:
Given a GenerateFlowFile -> LogAttribute flow with 10'000 batchSize (for both) and 10 byte payload, the length of the onTrigger is the following:
Screenshot 2020-10-06 at 15 45 46

libminifi/include/utils/Id.h
```diff
@@ -65,10 +67,11 @@ class Identifier {
 
   bool operator!=(const Identifier& other) const;
   bool operator==(const Identifier& other) const;
+  bool operator<(const Identifier& other) const;
 
   bool isNil() const;
 
-  std::string to_string() const;
+  SmallString<36> to_string() const;
```
szaszm: This should get a comment explaining the reason for using `SmallString<36>` instead of `std::string` to avoid a future refactoring changing it back to `std::string`.

libminifi/include/utils/SmallString.h
```diff
@@ -0,0 +1,86 @@
...
+namespace minifi {
+namespace utils {
+
+template<size_t N>
+class SmallString : public std::array<char, N + 1> {
+ public:
+  operator std::string() const {  // NOLINT
+    return {this->data()};
+  }
+
```
arpadboda, line 36: I think having a c_str() member returning data would be nice.
Helps to be used as a string and zero cost as the internal representation is a c_str anyway.

libminifi/include/utils/SmallString.h
```diff
@@ -0,0 +1,86 @@
...
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace utils {
+
+template<size_t N>
+class SmallString : public std::array<char, N + 1> {
+ public:
+  operator std::string() const {  // NOLINT
```
szaszm, line 33: I would make this `explicit`, or maybe even create a separate conversion function, e.g. a `to_string` free function mimicking `std::to_string` with ADL. I'm not 100% sure that the example is a good idea, but it seems to be useful.

The important part is to avoid implicit conversion to `std::string`.

libminifi/include/utils/Id.h
arpadboda, line 74: I wonder if we could typedef this somewhere to UUIDString or something talkative, rather then just the length referring to the content.

libminifi/src/core/ProcessSession.cpp
```diff
@@ -66,10 +66,12 @@ ProcessSession::~ProcessSession() {
 }
 
 void ProcessSession::add(const std::shared_ptr<core::FlowFile> &record) {
-  if (_updatedFlowFiles.find(record->getUUIDStr()) != _updatedFlowFiles.end()) {
+  utils::Identifier uuid;
+  record->getUUID(uuid);
+  if (_updatedFlowFiles.find(uuid) != _updatedFlowFiles.end()) {
```
szaszm, line 71: I would consider adding a version of `getUUID` that returns the result as a return value. It looks like `Identifier` can hold `Nil`, so that is something to keep in mind and something that makes returning even an `optional` unnecessary. I don't understand why the current version returns a bool of false in the return value instead of just returning a `Nil` UUID.

libminifi/include/utils/Id.h
adamdebreceni: added UUIDString alias

libminifi/include/utils/SmallString.h
adamdebreceni: added c_str method, and using it instead of data wherever applicable

libminifi/include/utils/Id.h
szaszm: I disagree with this idea because the intent of the function is to return a string, not to return an UUIDString, the same way `std::to_string` doesn't return IntStrings and LongStrings, but just `std::string`.

In general I think these typedefs can easily obscure the code if not used with care. There doesn't seem to be an industry-wide discussion about this, so this is purely my opinion, not a guideline.

libminifi/include/utils/SmallString.h
adamdebreceni: in this case I would keep it implicit since using `SmallString` is purely an optimization and it should be kept as transparent as possible

libminifi/include/utils/Id.h
adamdebreceni: I generally like if the types convey their usage, so we can have functions expecting `UUIDString` and we will make no mistake passing it a `"banana"` (the intent of the `Identifier::to_string` is not to return a `std::string` but to stringify its value, and that representation is quite constrained compared to `std::string`)

libminifi/include/utils/SmallString.h
adamdebreceni: on a practical note, there are quite a number of places passing `::getUUIDStr()` where a `std::string` is expected, IMO we should rewrite all these functions to expect an `Identifier`

libminifi/include/utils/Id.h
adamdebreceni: on a second thought, this whole "pass identifiers as strings" interface needs a rethinking, since the only place we should call `Identifier::to_string` should be when we create logs or other outputs for human consumption

libminifi/src/core/ProcessSession.cpp
adamdebreceni: the jury is still out on if we should consider `Nil` an `invalid` uuid, in the spec it is only mentioned as "special", but I am all for restricting our set of valid uuid's to not include `Nil` and then we can abolish all the `bool getUUID(Identifier&)` stuff

(the `getUUID` implementation itself already considers `Nil` to be invalid, so we just have to commit to this direction)

libminifi/include/utils/Id.h
szaszm: You convinced me that returning a UUIDString is not a bad thing in itself.

> I generally like if the types convey their usage, so we can have functions expecting `UUIDString`

But they accept any `SmallString<36>` including other typedefs, the type safety is not great here and we can make the mistake of passing `"banana789012345678901234567890123456"` (36 chars). Returning or passing `UUIDString` this way is a lie because we return/accept much more than just that.

Since this is a minor detail and strong typedefs don't exist, I'm fine with either way with a preference towards `SmallString<36>` for less lying versus `UUIDString` for more talkative naming.

libminifi/src/core/ProcessSession.cpp
szaszm: if `Nil` is valid:
`getUUID` should return the `Nil` UUID, because it's a valid value, nothing special about them mandates special treatment. `getUUID` can just `return uuid_`.

if `Nil` is invalid:
`getUUID` can use `Nil` to represent an invalid value, instead of using `false`, simplifying the interface. Currently the only invalid value is `Nil`, so `getUUID` can just `return uuid_`.

libminifi/include/utils/SmallString.h
szaszm: ok

libminifi/src/core/ProcessSession.cpp
adamdebreceni: you are right, whatever we consider `Nil` to be (invalid or valid), the `getUUID` implementation is the same, created a ticket for refactoring our `Identifier` treatment e.g. don't pass `.getUUIDStr()` when a `.getUUID()` would be perfectly fine, and to replace all `.getUUID(uuid)` with the new implementation, I imagine there are quite a number of places that will be touched by that PR

the created ticket is [here](https://issues.apache.org/jira/browse/MINIFICPP-1395)

libminifi/include/utils/Id.h
adamdebreceni: added a comment

arpadboda: As far as I see smallsring<N> cannot be constructed (even explicitly) from const char * with the proper length. Which is exactly what I would like to achieve. 
You can convert smallstring (and any typedefs based to that) to string, but you cannot anyhow create those by accident, only well designed APIs should be provide an instance of these. 

```
std::array<char, 10> temp_str = "banana123";

12:36: error: conversion from 'const char [10]' to non-scalar type 'std::array<char, 10ul>' requested
```

And I don't see any user-defined ctor in our code to make it possible. 

So the only thing we can do wrong is to typedef the same length of smallstring differently in different usecases. Which is a valid concerns as those could be mixed accidentally, but I don't think that living without typedefs helps that: if you see two APIs providing or requesting a given length of smallstr, you have no idea if they are compatible or not.

szaszm: My point is that if we follow this pattern and say we define `MySpecialMD5HashString` in the format of `"[[0123456789abcdef0123456789abcdef]]"`, with `using MySpecialMD5HashString = SmallString<36>;`, then `MySpecialMD5HashString` and `UUIDString` will be the same type and we get no errors or warnings when mixing them up.

Not a very practical example, but that's why I have a preference for avoiding such type aliases.

arpadboda: Okay, I get you. 

Wonder if we can find a solution in between. For eg. UUIDString is not a typedef, but publicly inherits SmallString, constructors are private and ID (where we create those) is a friend class?

szaszm: On the flipside, that would add some boilerplate.

If we exclude the original `SmallString<36>`, I think both the current type alias and this strong typedef-like but compatible new type are acceptable.

adamdebreceni: previously I wrote:

> I generally like if the types convey their usage, so we can have functions expecting `UUIDString`

this would make sense if we wanted functions expecting a `UUIDString`, I argue that we don't and wherever we pass/store a stringified `Identifier` we instead should pass/store an `Identifier`, [this](https://issues.apache.org/jira/browse/MINIFICPP-1395) is the issue aiming to make this change, where I think `UUIDString` would be removed

in light of this I would remove the `UUIDString` typedef

szaszm: That's indeed a better solution :+1:

adamdebreceni: @arpadboda are you too okay with this solution?

arpadboda: Works for me

adamdebreceni: done

szaszm (approved):

PR #932, opened 2020-10-28
MINIFICPP-1395 - Use Identifier instead of its stringified form wherever possible

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/controllers/keyvalue/AbstractCoreComponentStateManagerProvider.h
+++ b/libminifi/include/controllers/keyvalue/AbstractCoreComponentStateManagerProvider.h
@@ -45,14 +45,14 @@ class AbstractCoreComponentStateManagerProvider : public std::enable_shared_from
     std::shared_ptr<AbstractCoreComponentStateManagerProvider> provider_;
     utils::Identifier id_;
     bool state_valid_;
-    std::unordered_map<std::string, std::string> state_;
+    core::CoreComponentState state_;
 
    public:
     AbstractCoreComponentStateManager(std::shared_ptr<AbstractCoreComponentStateManagerProvider> provider, const utils::Identifier& id);
 
-    bool set(const std::unordered_map<std::string, std::string>& kvs) override;
+    bool set(const core::CoreComponentState& kvs) override;
 
-    bool get(std::unordered_map<std::string, std::string>& kvs) override;
+    bool get(core::CoreComponentState& kvs) override;
 
     bool clear() override;
 
@@ -60,14 +60,14 @@ class AbstractCoreComponentStateManagerProvider : public std::enable_shared_from
   };
 
  protected:
-  virtual bool setImpl(const utils::Identifier& key, const std::string& value) = 0;
-  virtual bool getImpl(const utils::Identifier& key, std::string& value) = 0;
+  virtual bool setImpl(const utils::Identifier& key, const std::string& serialized_state) = 0;
+  virtual bool getImpl(const utils::Identifier& key, std::string& serialized_state) = 0;
   virtual bool getImpl(std::map<utils::Identifier, std::string>& kvs) = 0;
   virtual bool removeImpl(const utils::Identifier& key) = 0;
   virtual bool persistImpl() = 0;
 
-  virtual std::string serialize(const std::unordered_map<std::string, std::string>& kvs);
-  bool deserialize(const std::string& serialized, std::unordered_map<std::string, std::string>& kvs);
+  virtual std::string serialize(const core::CoreComponentState& kvs);
+  bool deserialize(const std::string& serialized, core::CoreComponentState& kvs);
 };
 
 }  // namespace controllers
--- a/libminifi/include/controllers/keyvalue/AbstractCoreComponentStateManagerProvider.h
+++ b/libminifi/include/controllers/keyvalue/AbstractCoreComponentStateManagerProvider.h
@@ -20,6 +20,7 @@
 #include <unordered_map>
 #include <string>
 #include <memory>
+#include <map>
 
 #include "core/Core.h"
 #include "core/CoreComponentState.h"
@@ -35,19 +36,19 @@ class AbstractCoreComponentStateManagerProvider : public std::enable_shared_from
  public:
   virtual ~AbstractCoreComponentStateManagerProvider();
 
-  std::shared_ptr<core::CoreComponentStateManager> getCoreComponentStateManager(const std::string& uuid) override;
+  std::shared_ptr<core::CoreComponentStateManager> getCoreComponentStateManager(const utils::Identifier& uuid) override;
 
-  std::unordered_map<std::string, std::unordered_map<std::string, std::string>> getAllCoreComponentStates() override;
+  std::map<utils::Identifier, std::unordered_map<std::string, std::string>> getAllCoreComponentStates() override;
 
   class AbstractCoreComponentStateManager : public core::CoreComponentStateManager{
    private:
     std::shared_ptr<AbstractCoreComponentStateManagerProvider> provider_;
-    std::string id_;
+    utils::Identifier id_;
     bool state_valid_;
     std::unordered_map<std::string, std::string> state_;
 
    public:
-    AbstractCoreComponentStateManager(std::shared_ptr<AbstractCoreComponentStateManagerProvider> provider, const std::string& id);
+    AbstractCoreComponentStateManager(std::shared_ptr<AbstractCoreComponentStateManagerProvider> provider, const utils::Identifier& id);
 
     bool set(const std::unordered_map<std::string, std::string>& kvs) override;
 
@@ -59,10 +60,10 @@ class AbstractCoreComponentStateManagerProvider : public std::enable_shared_from
   };
 
  protected:
-  virtual bool setImpl(const std::string& key, const std::string& value) = 0;
-  virtual bool getImpl(const std::string& key, std::string& value) = 0;
-  virtual bool getImpl(std::unordered_map<std::string, std::string>& kvs) = 0;
-  virtual bool removeImpl(const std::string& key) = 0;
+  virtual bool setImpl(const utils::Identifier& key, const std::string& value) = 0;
+  virtual bool getImpl(const utils::Identifier& key, std::string& value) = 0;
+  virtual bool getImpl(std::map<utils::Identifier, std::string>& kvs) = 0;
+  virtual bool removeImpl(const utils::Identifier& key) = 0;
   virtual bool persistImpl() = 0;
 
   virtual std::string serialize(const std::unordered_map<std::string, std::string>& kvs);
--- a/libminifi/include/core/CoreComponentState.h
+++ b/libminifi/include/core/CoreComponentState.h
@@ -22,7 +22,6 @@
 
 #include <cstdint>
 #include <memory>
-#include <unordered_map>
 #include <map>
 #include <string>
 
--- a/libminifi/include/core/CoreComponentState.h
+++ b/libminifi/include/core/CoreComponentState.h
@@ -23,6 +23,7 @@
 #include <cstdint>
 #include <memory>
 #include <unordered_map>
+#include <map>
 #include <string>
 
 namespace org {
@@ -48,13 +49,13 @@ class CoreComponentStateManagerProvider {
  public:
   virtual ~CoreComponentStateManagerProvider() = default;
 
-  virtual std::shared_ptr<CoreComponentStateManager> getCoreComponentStateManager(const std::string& uuid) = 0;
+  virtual std::shared_ptr<CoreComponentStateManager> getCoreComponentStateManager(const utils::Identifier& uuid) = 0;
 
   virtual std::shared_ptr<CoreComponentStateManager> getCoreComponentStateManager(const CoreComponent& component) {
-    return getCoreComponentStateManager(component.getUUIDStr());
+    return getCoreComponentStateManager(component.getUUID());
   }
 
-  virtual std::unordered_map<std::string, std::unordered_map<std::string, std::string>> getAllCoreComponentStates() = 0;
+  virtual std::map<utils::Identifier, std::unordered_map<std::string, std::string>> getAllCoreComponentStates() = 0;
 };
 
 }  // namespace core
--- a/libminifi/include/core/CoreComponentState.h
+++ b/libminifi/include/core/CoreComponentState.h
@@ -22,6 +22,7 @@
 
 #include <cstdint>
 #include <memory>
+#include <unordered_map>
 #include <map>
 #include <string>
 
--- a/libminifi/include/core/CoreComponentState.h
+++ b/libminifi/include/core/CoreComponentState.h
@@ -32,13 +32,15 @@ namespace nifi {
 namespace minifi {
 namespace core {
 
+using CoreComponentState = std::unordered_map<std::string, std::string>;
+
 class CoreComponentStateManager {
  public:
   virtual ~CoreComponentStateManager() = default;
 
-  virtual bool set(const std::unordered_map<std::string, std::string>& kvs) = 0;
+  virtual bool set(const CoreComponentState& kvs) = 0;
 
-  virtual bool get(std::unordered_map<std::string, std::string>& kvs) = 0;
+  virtual bool get(CoreComponentState& kvs) = 0;
 
   virtual bool clear() = 0;
 
@@ -55,7 +57,7 @@ class CoreComponentStateManagerProvider {
     return getCoreComponentStateManager(component.getUUID());
   }
 
-  virtual std::map<utils::Identifier, std::unordered_map<std::string, std::string>> getAllCoreComponentStates() = 0;
+  virtual std::map<utils::Identifier, CoreComponentState> getAllCoreComponentStates() = 0;
 };
 
 }  // namespace core
--- a/libminifi/include/core/logging/Logger.h
+++ b/libminifi/include/core/logging/Logger.h
@@ -59,9 +59,16 @@ inline char const* conditional_conversion(const utils::SmallString<N>& arr) {
   return arr.c_str();
 }
 
-template<typename T>
+template<typename T, typename = typename std::enable_if<
+    std::is_arithmetic<T>::value ||
+    std::is_enum<T>::value ||
+    std::is_pointer<T>::value>::type>
 inline T conditional_conversion(T const& t) {
   return t;
+}
+
+inline char const* conditional_conversion(const char* str) {
+  return str;
 }
 
 template<typename ... Args>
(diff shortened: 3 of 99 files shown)


libminifi/include/core/CoreComponentState.h
```diff
@@ -48,13 +49,13 @@ class CoreComponentStateManagerProvider {
...
-  virtual std::shared_ptr<CoreComponentStateManager> getCoreComponentStateManager(const std::string& uuid) = 0;
+  virtual std::shared_ptr<CoreComponentStateManager> getCoreComponentStateManager(const utils::Identifier& uuid) = 0;
 
   virtual std::shared_ptr<CoreComponentStateManager> getCoreComponentStateManager(const CoreComponent& component) {
-    return getCoreComponentStateManager(component.getUUIDStr());
+    return getCoreComponentStateManager(component.getUUID());
   }
 
-  virtual std::unordered_map<std::string, std::unordered_map<std::string, std::string>> getAllCoreComponentStates() = 0;
+  virtual std::map<utils::Identifier, std::unordered_map<std::string, std::string>> getAllCoreComponentStates() = 0;
```
arpadboda, line 58: What's the reason behind this change?
If you do this, unordered map include can be removed imho.

libminifi/include/sitetosite/SiteToSite.h
```diff
@@ -246,24 +246,24 @@ class Transaction {
...
-
-    // Generate the global UUID for the transaction
-    uuid_str_ = uuid_.to_string();
   }
   // Destructor
   virtual ~Transaction() = default;
   // getUUIDStr
-  std::string getUUIDStr() {
-    return uuid_str_;
+  utils::SmallString<36> getUUIDStr() {
```
arpadboda, line 253: const

libminifi/src/CronDrivenSchedulingAgent.cpp
```diff
@@ -56,8 +56,8 @@ utils::TaskRescheduleInfo CronDrivenSchedulingAgent::run(const std::shared_ptr<c
       } else {
         Bosma::Cron schedule(processor->getCronPeriod());
         result = schedule.cron_to_next(from);
-        last_exec_[uuidStr] = result;
-        schedules_.insert(std::make_pair(uuidStr, schedule));
+        last_exec_[uuid] = result;
+        schedules_.insert(std::make_pair(uuid, schedule));
```
arpadboda, line 60: It's old code, but it's a map! :)

schedules_[uuid] = schedule;

arpadboda (commented):
Mostly looks good, some minor comments.

libminifi/src/controllers/keyvalue/PersistableKeyValueStoreService.cpp
```diff
@@ -29,20 +29,29 @@ PersistableKeyValueStoreService::PersistableKeyValueStoreService(const std::stri
...
-bool PersistableKeyValueStoreService::getImpl(std::unordered_map<std::string, std::string>& kvs) {
-  return get(kvs);
+bool PersistableKeyValueStoreService::getImpl(std::map<utils::Identifier, std::string>& kvs) {
+  std::unordered_map<std::string, std::string> states;
+  if (!get(states)) {
+    return false;
+  }
+  kvs.clear();
+  for (const auto& state : states) {
+    const auto uuid = utils::Identifier::parse(state.first);
```
fgerlits, line 47: it is not immediately clear to me from this code what kind of thing `uuid` is; I would either change `auto` to `utils::optional<Identifier>` or change the name of the variable to something like `optional_uuid`

libminifi/src/controllers/keyvalue/PersistableKeyValueStoreService.cpp
```diff
@@ -29,20 +29,29 @@ PersistableKeyValueStoreService::PersistableKeyValueStoreService(const std::stri
...
+    kvs[uuid.value()] = state.second;
```
fgerlits, line 48: what happens if this throws?  where do we catch it?

libminifi/src/core/reporting/SiteToSiteProvenanceReportingTask.cpp
```diff
@@ -68,16 +68,29 @@ void setJsonStr(const std::string& key, const std::string& value, rapidjson::Val
...
+rapidjson::Value getStringValue(const std::string& value, rapidjson::Document::AllocatorType& alloc) {
   rapidjson::Value Val;
   Val.SetString(value.c_str(), value.length(), alloc);
   return Val;
 }
 
-void appendJsonStr(const std::string& value, rapidjson::Value& parent, rapidjson::Document::AllocatorType& alloc) { // NOLINT
+template<size_t N>
+rapidjson::Value getStringValue(const utils::SmallString<N>& value, rapidjson::Document::AllocatorType& alloc) {
+  rapidjson::Value Val;
```
fgerlits, line 79: very minor, but I would rename this variable (and the other `valueVal`s) to `target_value` or `json_value` or something similar to match our naming conventions

libminifi/include/core/CoreComponentState.h
adamdebreceni: since it manages CoreComponents (it's in the name) and each CoreComponent has an id, it makes sense to use the id directly instead of `std::string`, if you are asking about the `unordered_map` => `map` change, I tried to avoid writing a hash implementation for the `Identifier` (which we then would have to manually specify each time we write the unordered_map type)

libminifi/src/provenance/Provenance.cpp
```diff
@@ -346,25 +352,33 @@ bool ProvenanceEventRecord::DeSerialize(const uint8_t *buffer, const size_t buff
...
+      ret = outStream.read(childUUIDStr);
       if (ret <= 0) {
         return false;
       }
-      this->addChildUuid(childUUID);
+      utils::optional<utils::Identifier> childUUID = utils::Identifier::parse(childUUIDStr);
+      if (!childUUID) {
+        return false;
+      }
+      this->addChildUuid(childUUID.value());
```
fgerlits, line 381: we do this a lot; it could be pulled out to a function, or even maybe to a new `outStream.read(utils::Identifier&)` overload

libminifi/src/CronDrivenSchedulingAgent.cpp
adamdebreceni: I think they have different semantics, your proposal overwrites if it already exists, the current version does not change the map if there was a same key in it, unclear if that was intentional (probably not) but I wanted to change semantics as little as possible

libminifi/include/core/CoreComponentState.h
arpadboda: Okay, makes sense, thanks!

libminifi/src/CronDrivenSchedulingAgent.cpp
arpadboda: 👍

arpadboda (approved):

libminifi/test/TestBase.cpp
```diff
@@ -104,9 +104,9 @@ std::shared_ptr<core::Processor> TestPlan::addProcessor(const std::shared_ptr<co
     connection->setDestination(processor);
 
     utils::Identifier uuid_copy, uuid_copy_next;
-    last->getUUID(uuid_copy);
+    uuid_copy = last->getUUID();
     connection->setSourceUUID(uuid_copy);
-    processor->getUUID(uuid_copy_next);
+    uuid_copy_next = processor->getUUID();
     connection->setDestinationUUID(uuid_copy_next);
```
fgerlits, line 110: minor, but this could be simplified to
```c++
connection->setSourceUUID(last->getUUID());
connection->setDestinationUUID(processor->getUUID());
```
(also further down in this file)

nanofi/src/cxx/Plan.cpp
```diff
@@ -304,11 +304,11 @@ std::shared_ptr<minifi::Connection> ExecutionPlan::connectProcessors(std::shared
   connection->setSource(src_proc);
 
   utils::Identifier uuid_copy, uuid_copy_next;
-  src_proc->getUUID(uuid_copy);
+  uuid_copy = src_proc->getUUID();
   connection->setSourceUUID(uuid_copy);
```
fgerlits, line 308: here, too, this could be `connection->setSourceUUID(src_proc->getUUID());` etc

libminifi/include/utils/SmallString.h
```diff
@@ -34,6 +34,10 @@ class SmallString : public std::array<char, N + 1> {
     return {c_str()};
   }
 
+  size_t length() const noexcept {
```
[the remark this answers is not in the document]
adamdebreceni: I intended to mimic the signature of `std::string::length` on purpose, to make it as similar as possible, and that is only `constexpr` from `c++20`, should I change it to `constexpr`?

libminifi/include/utils/ThreadPool.h
```diff
@@ -45,6 +45,8 @@ namespace nifi {
 namespace minifi {
 namespace utils {
 
+using TaskId = std::string;
```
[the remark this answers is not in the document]
adamdebreceni: `GetTCP` creates a worker with `"workers"` name

https://github.com/apache/nifi-minifi-cpp/blob/d94e14f5fadb59a7b07da968b609d4f0831f4c27/extensions/standard-processors/processors/GetTCP.cpp#L290

libminifi/include/utils/SmallString.h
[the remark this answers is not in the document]
fgerlits: I would, as it does not affect its usage, but I don't insist

libminifi/include/utils/ThreadPool.h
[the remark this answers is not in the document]
fgerlits: I see, OK

szaszm (commented):
Awesome changes, some questions and suggestions below.

libminifi/include/core/logging/Logger.h
```diff
@@ -59,11 +59,18 @@ inline char const* conditional_conversion(const utils::SmallString<N>& arr) {
...
+    std::is_arithmetic<T>::value ||
+    std::is_enum<T>::value ||
+    std::is_pointer<T>::value>::type>
 inline T conditional_conversion(T const& t) {
   return t;
 }
 
+inline char const* conditional_conversion(const char* str) {
+  return str;
+}
```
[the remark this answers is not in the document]
adamdebreceni: for some reason I cannot reply to your comment about `std::hash` specialization:
from [google guidelines](https://google.github.io/styleguide/cppguide.html#std_hash)

> Do not define specializations of std::hash.

libminifi/src/ThreadedSchedulingAgent.cpp
```diff
@@ -110,16 +110,15 @@ void ThreadedSchedulingAgent::schedule(std::shared_ptr<core::Processor> processo
...
 
 void ThreadedSchedulingAgent::stop() {
   SchedulingAgent::stop();
   std::lock_guard<std::mutex> lock(mutex_);
-  for (const auto& p : processors_running_) {
-    logger_->log_error("SchedulingAgent is stopped before processor was unscheduled: %s", p);
-    thread_pool_.stopTasks(p);
+  for (const auto& processor_id : processors_running_) {
+    logger_->log_error("SchedulingAgent is stopped before processor was unscheduled: %s", processor_id.to_string());
+    thread_pool_.stopTasks(processor_id.to_string());
```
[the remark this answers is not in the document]
adamdebreceni: yes that seems to be the case, (so processors with greater than one maxConcurrentTasks share the same `task_status_` flag)

libminifi/src/controllers/keyvalue/PersistableKeyValueStoreService.cpp
```diff
@@ -29,20 +29,31 @@ PersistableKeyValueStoreService::PersistableKeyValueStoreService(const std::stri
...
+  std::unordered_map<std::string, std::string> states;
+  if (!get(states)) {
+    return false;
+  }
+  kvs.clear();
+  for (const auto& state : states) {
+    utils::optional<utils::Identifier> optional_uuid = utils::Identifier::parse(state.first);
+    if (optional_uuid) {
+      kvs[optional_uuid.value()] = state.second;
+    }
```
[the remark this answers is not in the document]
adamdebreceni: since a `PerisistableKeyValueStoreService (PKVSS)` ISA `KeyValueStoreService (KVSS)` (publicly inherits) which can store arbitrary `std::string` keys, it is not impossible to cast a `PKVSS` to a `KVSS`, add some "invalid" key, and "mess up" the `PKVSS`, I would refactor this to not inherit from `KVSS` but contain an instance of it, for now I will add a log message

libminifi/include/core/logging/Logger.h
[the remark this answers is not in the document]
adamdebreceni: for this comment, it is needed so we can allow [this](https://godbolt.org/z/9drc8f) to compile, and now that I wrote this down, I realize I could simply remove the `const&` qualifier on the template

```
#include <iostream>

struct A{
  static constexpr const char* STR = "abcd";
};

template<typename T, typename = typename std::enable_if<
    std::is_arithmetic<T>::value ||
    std::is_enum<T>::value ||
    std::is_pointer<T>::value>::type>
inline T conditional_conversion(T const& t) {
  return t;
}

// inline char const* conditional_conversion(const char* str) {
//   return str;
// }

int main(){
  std::cout << conditional_conversion(A::STR) << std::endl;
}
```

libminifi/include/controllers/keyvalue/AbstractCoreComponentStateManagerProvider.h
```diff
@@ -59,10 +60,10 @@ class AbstractCoreComponentStateManagerProvider : public std::enable_shared_from
...
 
  protected:
-  virtual bool setImpl(const std::string& key, const std::string& value) = 0;
-  virtual bool getImpl(const std::string& key, std::string& value) = 0;
-  virtual bool getImpl(std::unordered_map<std::string, std::string>& kvs) = 0;
-  virtual bool removeImpl(const std::string& key) = 0;
+  virtual bool setImpl(const utils::Identifier& key, const std::string& value) = 0;
+  virtual bool getImpl(const utils::Identifier& key, std::string& value) = 0;
+  virtual bool getImpl(std::map<utils::Identifier, std::string>& kvs) = 0;
+  virtual bool removeImpl(const utils::Identifier& key) = 0;
```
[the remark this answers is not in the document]
adamdebreceni: renamed the parameter, also added type alias for the `CoreComponentState` to `unordered_map<string, string>`

libminifi/src/ThreadedSchedulingAgent.cpp
```diff
@@ -110,16 +110,15 @@ void ThreadedSchedulingAgent::schedule(std::shared_ptr<core::Processor> processo
...
+  processors_running_.insert(processor->getUUID());
 }
 
 void ThreadedSchedulingAgent::stop() {
   SchedulingAgent::stop();
   std::lock_guard<std::mutex> lock(mutex_);
   for (const auto& p : processors_running_) {
-    logger_->log_error("SchedulingAgent is stopped before processor was unscheduled: %s", p);
-    thread_pool_.stopTasks(p);
+    logger_->log_error("SchedulingAgent is stopped before processor was unscheduled: %s", p.to_string());
```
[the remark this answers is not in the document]
adamdebreceni: done

libminifi/src/controllers/keyvalue/PersistableKeyValueStoreService.cpp
adamdebreceni: done

libminifi/src/controllers/keyvalue/PersistableKeyValueStoreService.cpp
adamdebreceni: rewrote it, to ignore such cases, but log errors, (should be refactored to contain an instance of `KeyValueStoreService` instead of extending it)

libminifi/src/provenance/Provenance.cpp
```diff
@@ -128,7 +128,7 @@ bool ProvenanceEventRecord::Serialize(org::apache::nifi::minifi::io::BufferStrea
     return false;
   }
 
-  ret = outStream.write(this->flow_uuid_);
+  ret = outStream.write(this->flow_uuid_.to_string());
```
fgerlits, line 131: these `.to_string()`s are no longer needed here, right?

nanofi/src/cxx/Plan.cpp
adamdebreceni: done

libminifi/test/TestBase.cpp
adamdebreceni: done

libminifi/src/provenance/Provenance.cpp
adamdebreceni: added, also added an `OutputStream:read` overload for symmetry

fgerlits (approved):

libminifi/include/core/logging/Logger.h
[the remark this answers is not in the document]
szaszm: The reason you couldn't reply to my comment about hash is because it was a reply to this discussion: https://github.com/apache/nifi-minifi-cpp/pull/932#discussion_r513307831

To this thread: acknowledged, I'm fine with either way.

libminifi/src/ThreadedSchedulingAgent.cpp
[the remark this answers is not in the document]
szaszm: @arpadboda could you confirm or clarify, as the author/major refactorer of this code?

libminifi/include/core/CoreComponentState.h
szaszm: marking as resolved again as per https://github.com/apache/nifi-minifi-cpp/pull/932#discussion_r514057657

libminifi/src/controllers/keyvalue/PersistableKeyValueStoreService.cpp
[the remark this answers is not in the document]
szaszm: Sounds like a violation of the Liskov substitution principle. Thanks for the mitigation.

szaszm (approved):

libminifi/include/utils/SmallString.h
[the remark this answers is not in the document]
adamdebreceni: made it constexpr

libminifi/src/ThreadedSchedulingAgent.cpp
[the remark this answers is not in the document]
arpadboda: I think it's intentional. 
The IDs are only used to identify the tasks belonging to a given component, the threadpool otherwise doesn't care. 
I wouldn't call it a beautiful design, but works as expected, so I kept it when the code was refactored.

PR #935, opened 2020-11-10
MINIFICPP-1404 Add option to disable unity build of AWS library

lordgamez:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,6 +72,9 @@ if (${FORCE_COLORED_OUTPUT})
 	   add_compile_options (-fcolor-diagnostics)
 	endif ()
 endif ()
+option(AWS_ENABLE_UNITY_BUILD "If enabled, AWS SDK libraries will be built as a single, generated .cpp file. \
+	This can significantly reduce static library size as well as speed up a single compilation time, but it is regenerated \
+	and recompiled in every iterative build instance. Turn off to avoid recompilation." ON)
 
 cmake_dependent_option(ASAN_BUILD "Uses AddressSanitizer to instrument the code" OFF "NOT WIN32" OFF)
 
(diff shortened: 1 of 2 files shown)


szaszm (approved):
Tested, works, but I'm not sure this change adds a lot of value to the project. I would expect a third party lib to only change when it's updated to a newer version, which is a rare occurrence. Since it's behind an option, it doesn't hurt too much either. Treat my approval as a verification that this change is successfully achieving its goals but I'm neutral on whether we need this feature.

lordgamez:
> Tested, works, but I'm not sure this change adds a lot of value to the project. I would expect a third party lib to only change when it's updated to a newer version, which is a rare occurrence. Since it's behind an option, it doesn't hurt too much either. Treat my approval as a verification that this change is successfully achieving its goals but I'm neutral on whether we need this feature.

It was my personal experience that lead to this change, as working on the AWS features every newly introduced test change or minor code change triggered the regeneration and recompilation of  the AWS SDK if unity build was enabled. This became tiresome after a while as changes took disproportionately long time to compile because of this. I did not want to disable the unity build by default because it makes the binary 20-30MB smaller in my experience, so I chose this option, but I opened the PR separately from other changes to be open for discussions whether it's okay have this option or just locally disable it in the CMake files when developing the AWS extension. This may be more explicit and could be found more easily with some explanation if anyone else experiences the same problem.

szaszm:
I wonder why changes in our aws code triggers recompilation of the third party aws lib. It's as is libaws had a dependency on the minifi aws extension rather than the other way around.

lordgamez:
We currently have 2 dependencies from the sdk, the aws-cpp-sdk-s3 and the aws-cpp-sdk-core libraries. When we trigger a build after a dependent change the unity build generates new ub_S3.cpp and a ub_core.cpp files from the source files of those libraries. CMake only checks the timestamp of these source files so even with if the content is unchanged, they are rebuilt.

arpadboda (approved):

PR #938, opened 2020-11-16
MINIFICPP-1408 thrown objects should be derived from std::exception

szaszm:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/sftp/client/SFTPClient.h
+++ b/extensions/sftp/client/SFTPClient.h
@@ -27,6 +27,7 @@
 #include <vector>
 
 #include "Exception.h"
+#include "utils/Enum.h"
 #include "utils/HTTPClient.h"
 #include "core/logging/Logger.h"
 #include "core/logging/LoggerConfiguration.h"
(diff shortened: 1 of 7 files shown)


lordgamez (approved):

extensions/sftp/client/SFTPClient.h
```diff
@@ -49,6 +50,12 @@ enum class SFTPError : uint8_t {
   SFTP_ERROR_UNEXPECTED
 };
 
+struct SFTPException : Exception {
+  explicit SFTPException(const SFTPError err)
+      :Exception{ExceptionType::FILE_OPERATION_EXCEPTION, fmt::format("SFTP Error {0}", static_cast<int>(err))}
```
adamdebreceni, line 55: I am not saying that we should change it in this PR (or in any PR for that matter), but turning `SFTPError` into a `SMART_ENUM` could benefit us, (e.g. here we could then write `err.toString()`)

adamdebreceni: this is just something to think about, not a change request

szaszm: I thought about it but then went for the easier solution. I might change it after I figure out the reason for the test failures.

adamdebreceni (approved):

extensions/sftp/client/SFTPClient.h
szaszm: I have some negative feedback regarding the conversion to SMART_ENUM: the enum values are implicitly convertible to integer which makes it less safe than `enum class`. This resulted in compiling but invalid code when the wrong overload of `LastSFTPError::operator=` was called and I had to introduce differently named functions.

I'm leaning towards keeping SMART_ENUM in this case but I'll be more careful about introducing it in new places, weighing the convenient string conversion against the safety tradeoff in each case.

adamdebreceni: yes, the implicit convertibility is indeed troublesome (and the extra scope specifier we save when accessing an instance might not be worth it), although in this specific case I would have gone with changing the parameter type from `const SFTPError&` to `SFTPError::Type`

arpadboda (approved):

PR #939, opened 2020-11-19
MINIFICPP-1409 Add default credentials provider chain support to AWS authentication

lordgamez:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/CONTROLLERS.md
+++ b/CONTROLLERS.md
@@ -23,8 +23,8 @@
 
 ### Description
 
-Manages the Amazon Web Services (AWS) ```Access Key``` and ```Secret Access Key``` for a AWS account. This allows for multiple 
-AWS credential services to be defined. This also allows for multiple AWS related processors to reference this single 
+Manages the Amazon Web Services (AWS) credentials for an AWS account. This allows for multiple
+AWS credential services to be defined. This also allows for multiple AWS related processors to reference this single
 controller service so that AWS credentials can be managed and controlled in a central location.
 
 ### Properties
@@ -35,5 +35,7 @@ default values, and whether a property supports the NiFi Expression Language.
 
 | Name | Default Value | Allowable Values | Expression Language Supported? | Description |
 | - | - | - | - | - |
-| **Access Key** | | | Yes | Specifies the AWS Access Key |
-| **Secret Key** | | | Yes | Specifies the AWS Secret Key |
+|**Use Default Credentials**|false||No|If true, uses the Default Credential chain, including EC2 instance profiles or roles, environment variables, default user credentials, etc.|
+|Access Key|||Yes|Specifies the AWS Access Key|
+|Secret Key|||Yes|Specifies the AWS Secret Key|
+|Credentials File|||No|Path to a file containing AWS access key and secret key in properties file format. Properties used: accessKey and secretKey|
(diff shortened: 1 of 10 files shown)


hunyadi-dev (approved):
Looks good to me!

extensions/aws/AWSCredentialsProvider.cpp
```diff
@@ -0,0 +1,94 @@
...
+  secret_key_ = secret_key;
+}
+
+void AWSCredentialsProvider::setCredentialsFile(const std::string &credentials_file) {
+  credentials_file_ = credentials_file;
+}
+
+minifi::utils::optional<Aws::Auth::AWSCredentials> AWSCredentialsProvider::getAWSCredentials() {
+  if (use_default_credentials_) {
+    logger_->log_debug("Trying to use default AWS credentials provider chain.");
```
fgerlits, line 61: How often is this function called?  If not too frequently, maybe these logs could be on info level.

extensions/aws/processors/PutS3Object.cpp
```diff
@@ -220,57 +227,29 @@ minifi::utils::optional<Aws::Auth::AWSCredentials> PutS3Object::getAWSCredential
...
-    return prop_cred.value();
-  }
-
-  auto file_cred = getAWSCredentialsFromFile(context);
-  if (file_cred) {
-    logger_->log_info("AWS Credentials successfully set from file");
-    return file_cred.value();
-  }
-
+    const std::shared_ptr<core::FlowFile> &flow_file) {
```
fgerlits, line 232: why is this no longer const?

lordgamez: We are now using the new AwsCredentialsProvider class to retrieve the credentials if no controller service is used. This provider is a member which is changed in this function so it cannot be const anymore.

extensions/aws/AWSCredentialsProvider.cpp
lordgamez: Currently it is called for every onTrigger instance in the PutS3Object as the credentials can come from flowfile attributes as expression language variables. It may be too frequent in some use cases if we change it to info level.

fgerlits (approved):

CONTROLLERS.md
```diff
@@ -35,5 +35,7 @@ default values, and whether a property supports the NiFi Expression Language.
 
 | Name | Default Value | Allowable Values | Expression Language Supported? | Description |
 | - | - | - | - | - |
-| **Access Key** | | | Yes | Specifies the AWS Access Key |
-| **Secret Key** | | | Yes | Specifies the AWS Secret Key |
+|**Use Default Credentials**|false||No|If true, uses the Default Credential chain, including EC2 instance profiles or roles, environment variables, default user credentials, etc.|
+|Access Key|||Yes|Specifies the AWS Access Key|
```
arpadboda, line 39: Access key being required seems to be a bit strange to me when I can use default credentials.

I have a feeling that we need custom validation here: either both keys should be provided or default cred should be used.

arpadboda (commented):
Mostly looks good, added a comment on property validation.

CONTROLLERS.md
lordgamez: I actually changed it to not be required anymore (removed the * in the docs). The validation starts with the Use Default Credentials, if it is set it will go through the AWS chain and try to get the credentials from there. If it's not found then it checks if the Access Key and Secret Key are set and uses them as credentials if both are set. If those are not set either it checks if the Credentials File is set and tries to read the access key and secret key properties from that file.

arpadboda: Okay, looks good, thanks!

arpadboda (approved):

PR #941, opened 2020-11-23
MINIFICPP-1411 - Convert \n newline characters to <br/> tags in the docs generation

hunyadi-dev:
When agent documentation generation is run, processor property descriptions that contain `\n` characters are not properly generated, as manual linebreaks in github markdown tables are not supported. These are to be replaced with `<br/>` html tags.


arpadboda (approved):
Good spot, thanks!

PR #950, opened 2020-12-04
MINIFICPP-1405 - Update civetweb version and correct build flags used by it

hunyadi-dev:
Also, disable building duktape and lua support for civetweb.

Seems like civetweb had an issue using the available compiler flags. I opened a PR for them [here](https://github.com/civetweb/civetweb/pull/948), until they release a version with it, we can use the fix as a patch.


thirdparty/civetweb/civetweb.patch
```diff
@@ -1,61 +1,45 @@
...
+From cf90b37da9432cf957f4f2e3ce837d6c73647802 Mon Sep 17 00:00:00 2001
+From: Adam Hunyadi <hunyadi.dev@gmail.com>
+Date: Fri, 4 Dec 2020 11:26:12 +0100
+Subject: [PATCH] Fix checks for available compiler flags
+
+---
+ cmake/AddCCompilerFlag.cmake   | 5 ++---
+ cmake/AddCXXCompilerFlag.cmake | 5 ++---
+ 2 files changed, 4 insertions(+), 6 deletions(-)
+
```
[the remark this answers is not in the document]
hunyadi-dev: This was auto-generated by git. Will remove the redundant parts but I did not think this would be manually read.

cmake/BundledCivetWeb.cmake
```diff
@@ -30,8 +30,8 @@ function(use_bundled_civetweb SOURCE_DIR BINARY_DIR)
     endif()
 
     set(BYPRODUCTS
-            "lib/${PREFIX}civetweb.${SUFFIX}"
-            "lib/${PREFIX}civetweb-cpp.${SUFFIX}"
+            "${CMAKE_INSTALL_LIBDIR}/${PREFIX}civetweb.${SUFFIX}"
+            "${CMAKE_INSTALL_LIBDIR}/${PREFIX}civetweb-cpp.${SUFFIX}"
```
fgerlits, line 34: I think you need to do the same on line 80, as well

hunyadi-dev: Good catch!!

fgerlits: yeah -- that makes me wonder if adding `-DCMAKE_INSTALL_LIBDIR=lib` (as is done in BundledLibRdKafka) would be a better solution

but I think either way is fine

fgerlits (approved):

cmake/BundledCivetWeb.cmake
```diff
@@ -30,8 +30,8 @@ function(use_bundled_civetweb SOURCE_DIR BINARY_DIR)
     endif()
 
     set(BYPRODUCTS
-            "lib/${PREFIX}civetweb.${SUFFIX}"
-            "lib/${PREFIX}civetweb-cpp.${SUFFIX}"
+            "${CMAKE_INSTALL_LIBDIR}/${PREFIX}civetweb.${SUFFIX}"
```
lordgamez, line 33: [GNUInstallDirs](https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html) should be included to use CMAKE_INSTALL_* variables on unix systems and CMAKE_INSTALL_LIBDIR should be set explicitly on Windows.

hunyadi-dev: Updated, now GNUInstallDirs are included when not on windows and libdir is set accordingly.

lordgamez (approved):

arpadboda (approved):

hunyadi-dev:
Changing this to the safer v1.12 in #952, closing this PR.

PR #952, opened 2020-12-07
MINIFICPP-1405 - Update civetweb version and correct build flags used by it [v1.12 version]

hunyadi-dev:
Also, disable building duktape and lua support for civetweb.

Seems like civetweb had an issue using the available compiler flags. I opened a PR for them [here](https://github.com/civetweb/civetweb/pull/948), until they release a version with it, we can use the fix as a patch.

Seems like the v1.13 of civetweb fails to properly pick up openssl (could be related that we use libress that defines a bogus openssl version). This branch is a fix with the 1.12 version of civetweb.


arpadboda (approved):

hunyadi-dev:
Seems like this is the safe version of #950 closing that one.

fgerlits (approved):

lordgamez (approved):

PR #957, opened 2020-12-11
MINIFICPP-1407 - Fetch flow configuration file from C2 if not found locally

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:


encrypt-config/tests/ConfigFileEncryptorTests.cpp
```diff
@@ -28,7 +28,7 @@ size_t base64_length(size_t unencoded_length) {
   return (unencoded_length + 2) / 3 * 4;
 }
 
-bool check_encryption(const ConfigFile& test_file, const std::string& property_name, size_t original_value_length) {
+bool check_encryption(const  ConfigFile& test_file, const std::string& property_name, size_t original_value_length) {
```
adamdebreceni, line 31: accidental space

encrypt-config/tests/ConfigFileEncryptorTests.cpp
```diff
@@ -67,7 +69,7 @@ TEST_CASE("ConfigFileEncryptor can encrypt the sensitive properties", "[encrypt-
       "6q9u8LEDy1/CdmSBm8oSqPS/Ds5UOD2nRouP8yUoK10="));
 
   SECTION("default properties") {
-    ConfigFile test_file{std::ifstream{"resources/minifi.properties"}};
+     ConfigFile test_file{std::ifstream{"resources/minifi.properties"}};
```
adamdebreceni, line 72: incorrect indentation (and more in this file)

libminifi/include/FlowController.h
```diff
@@ -304,6 +192,9 @@ class FlowController : public core::controller::ControllerServiceProvider,  publ
 
   std::vector<BackTrace> getTraces() override;
 
+ private:
+  std::unique_ptr<core::ProcessGroup> instantiateNetwork();
```
adamdebreceni, line 196: not the best name, to-be-renamed

libminifi/include/properties/Properties.h
```diff
@@ -45,19 +51,19 @@ class Properties {
 
   // Clear the load config
   void clear() {
-    std::lock_guard<std::mutex> lock(mutex_);
+    std::lock_guard<std::recursive_mutex> lock(mutex_);
```
adamdebreceni, line 54: forgot why we needed the `recursive_mutex`, add explanation pls

extensions/http-curl/tests/HTTPHandlers.h
```diff
@@ -447,35 +447,43 @@ class HeartbeatHandler : public ServerAwareHandler {
...
 
     return true;
   }
 
-  bool handleGet(CivetServer *server, struct mg_connection *conn) override {
-    std::ifstream myfile(test_file_location_.c_str(), std::ios::in | std::ios::binary);
-    if (myfile.good()) {
-      std::string str((std::istreambuf_iterator<char>(myfile)), (std::istreambuf_iterator<char>()));
+ private:
+  std::string test_file_location_;
```
arpadboda, line 472: nitpicking: can be const as only the ctor can set it.

extensions/http-curl/tests/HTTPHandlers.h
```diff
@@ -499,8 +507,8 @@ class C2UpdateHandler : public ServerAwareHandler {
   }
 
   std::atomic<size_t> calls_{0};
```
arpadboda, line 509: Does this have to be public?

libminifi/include/core/controller/ForwardingControllerServiceProvider.h
```diff
@@ -0,0 +1,136 @@
...
+#include "ControllerServiceNode.h"
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace core {
+namespace controller {
+
+class ForwardingControllerServiceProvider : public ControllerServiceProvider {
```
arpadboda, line 35: Why do we need pimpl here?

At first look I'm not sure I perfectly understand the reason of introducing this class.

libminifi/include/properties/Properties.h
```diff
@@ -51,13 +57,13 @@ class Properties {
...
-    properties_[key] = value;
+    properties_[key] = PropertyValue{value, true};
     dirty_ = true;
   }
   // Check whether the config value existed
-  bool has(std::string key) const {
+  bool has(const std::string& key) const {
     std::lock_guard<std::mutex> lock(mutex_);
-    return properties_.count(key) > 0;
+    return properties_.find(key) != properties_.end();
```
arpadboda, line 66: This seems like a question of preference. As either way is correct, I think every author I free do decide which one to use, but I wouldn't change it. It just adds extra diffs to the code review without value.

libminifi/include/core/controller/ForwardingControllerServiceProvider.h
adamdebreceni: its purpose is to make the `FlowController` class skinnier as I was finding myself scrolling a lot, and I thought it is a general-enough behavior to factor out

arpadboda: I agree with the general purpose (make flowcontroller tinier), but not sure if pimpl is the best solution to achieve this. Usually introduces a lot of boilerplate, so I'm against it unless we can prove that we really need it. 

If we don't have any better idea, let's do it, but I hope that we will. @szaszm @fgerlits what's your opinion?

adamdebreceni: note that it was already present in `FlowController`, it has just been moved out, I am all for eliminating the need for it altogether

extensions/http-curl/tests/HTTPHandlers.h
adamdebreceni: made it protected and added getter

libminifi/include/properties/Properties.h
adamdebreceni: reverted

libminifi/include/core/controller/ForwardingControllerServiceProvider.h
szaszm: Please note that the pointer was present in the old `FlowController.h:343` and the members at old `FlowController.cpp:363-522`.

I wouldn't support introducing a pimpl here, but I support the extraction of this existing pimpl to its own class, because it means one less responsibility for the bloated `FlowController` and because it's a meaningful class on its own.

I also support eliminating it altogether, but it might require more changes and I don't think it's a good idea to further inflate this PR.

arpadboda: Okay, I like this approach.

Let's leave this PR as it is and create a follow-up to find a better solution later.

arpadboda (approved):

PR #958, opened 2020-12-12
MINIFICPP-1432 Remove the timing-sensitivity of NetworkPrioritizerServiceTests, version 1

fgerlits:
NOTE: alternative PR to #960, only one of the two should be merged!

https://issues.apache.org/jira/browse/MINIFICPP-1432

* Add a mockable Clock field to `NetworkPrioritizerService`
* Fix and clean up the tests
* Fix some bugs in `NetworkPrioritizerService`:
   - `max_payload_` was not set in `onEnable()`
   - only payloads strictly smaller than `max_payload_` were allowed
   - use `steady_clock` instead of `system_clock`

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/controllers/NetworkPrioritizerService.h
+++ b/libminifi/include/controllers/NetworkPrioritizerService.h
@@ -18,11 +18,13 @@
 #ifndef LIBMINIFI_INCLUDE_CONTROLLERS_NETWORKPRIORITIZERSERVICE_H_
 #define LIBMINIFI_INCLUDE_CONTROLLERS_NETWORKPRIORITIZERSERVICE_H_
 
-#include <string>
-#include <vector>
 #include <iostream>
-#include <memory>
 #include <limits>
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
+
 #include "core/Resource.h"
 #include "utils/StringUtils.h"
 #include "io/validation.h"
@@ -43,7 +45,9 @@ namespace controllers {
  */
 class NetworkPrioritizerService : public core::controller::ControllerService, public minifi::io::NetworkPrioritizer, public std::enable_shared_from_this<NetworkPrioritizerService> {
  public:
-  explicit NetworkPrioritizerService(const std::string &name, const utils::Identifier& uuid = {})
+  explicit NetworkPrioritizerService(const std::string& name,
+                                     const utils::Identifier& uuid = {},
+                                     std::shared_ptr<utils::timeutils::Clock> clock = std::make_shared<utils::timeutils::SteadyClock>())
       : ControllerService(name, uuid),
         enabled_(false),
         max_throughput_(std::numeric_limits<uint64_t>::max()),
@@ -53,6 +57,7 @@ class NetworkPrioritizerService : public core::controller::ControllerService, pu
         timestamp_(0),
         bytes_per_token_(0),
         verify_interfaces_(true),
+        clock_(std::move(clock)),
         logger_(logging::LoggerFactory<NetworkPrioritizerService>::getLogger()) {
   }
 
@@ -121,6 +126,7 @@ class NetworkPrioritizerService : public core::controller::ControllerService, pu
   bool verify_interfaces_;
 
  private:
+  std::shared_ptr<utils::timeutils::Clock> clock_;
   std::shared_ptr<logging::Logger> logger_;
 };
 
--- a/libminifi/include/utils/TimeUtil.h
+++ b/libminifi/include/utils/TimeUtil.h
@@ -54,6 +54,29 @@ inline uint64_t getTimeNano() {
 }
 
 /**
+ * Mockable clock classes
+ */
+class Clock {
+ public:
+  virtual ~Clock() = default;
+  virtual std::chrono::milliseconds timeSinceEpoch() const = 0;
+};
+
+class SystemClock : public Clock {
+ public:
+  std::chrono::milliseconds timeSinceEpoch() const override {
+    return std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch());
+  }
+};
+
+class SteadyClock : public Clock {
+ public:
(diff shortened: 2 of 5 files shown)


arpadboda (approved):

libminifi/test/unit/NetworkPrioritizerServiceTests.cpp
```diff
@@ -41,129 +47,105 @@ TEST_CASE("TestPrioritizerOneInterface", "[test1]") {
...
+    REQUIRE("eth0" == parent_controller->getInterface(5).getInterface());
+    // triggered the max throughput on eth0, switching to eth1
+    REQUIRE("eth1" == parent_controller->getInterface(5).getInterface());
+    REQUIRE("eth1" == parent_controller->getInterface(5).getInterface());
   }
-}
 
+  SECTION("Can keep sending on eth0 if we wait between packets") {
+    for (int i = 0; i < 100; i++) {
+      REQUIRE("eth0" == parent_controller->getInterface(10).getInterface());
```
adamdebreceni, line 115: this is a little unclear to me, so `getInterface(size)` returns a network interface that we should use to send the next `size` number of bytes? what controls how long we should wait (5ms in this case) until the same interface is available?

fgerlits: > `getInterface(size)` returns a network interface that we should use to send the next size number of bytes?

Yes.

> what controls how long we should wait (5ms in this case) until the same interface is available?

It is a (version of the) [token bucket algorithm](https://en.wikipedia.org/wiki/Token_bucket): conceptually, the bucket starts out with `b` tokens and `t` tokens are added to the bucket each millisecond, and each send operation removes `s` times `size` tokens from the bucket, if there are enough tokens available, otherwise it fails (and does not remove any tokens).

[In practice, the bucket is filled in `getInterface()` according to the number of elapsed milliseconds since the last call to `getInterface()`, and the way the `b`, `t` and `s` parameters are set is a bit strange (`t` is always 2, `b` and `s` depend on `MaxThroughput`, but differently depending on whenter `MaxThroughput` is less or more than 1 kB).]

Since I don't think anybody actually uses this class at the moment, I did not clean it up; I only fixed the unit test so that it is no longer dependent on timing, and fixed a few obvious bugs in the service code.

adamdebreceni: thank you for clarifying!

adamdebreceni (approved):

libminifi/include/controllers/NetworkPrioritizerService.h
```diff
@@ -121,6 +126,7 @@ class NetworkPrioritizerService : public core::controller::ControllerService, pu
   bool verify_interfaces_;
 
  private:
+  std::shared_ptr<utils::timeutils::Clock> clock_;
```
szaszm, line 129: Why shared_ptr, what are the other owners?

adamdebreceni: in the tests the clock is being advanced externally

libminifi/include/utils/TimeUtil.h
```diff
@@ -53,6 +53,29 @@ inline uint64_t getTimeNano() {
   return std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
 }
 
+/**
+ * Mockable clock classes
+ */
+class Clock {
+ public:
+  virtual ~Clock() = default;
+  virtual std::chrono::milliseconds timeSinceEpoch() const = 0;
```
szaszm, line 62: This could be type erased with `std::function<std::chrono::milliseconds()>` instead of a class hierarchy for simplicity.

adamdebreceni: also: I wouldn't introduce a `SystemClock` class until it is actually needed

adamdebreceni: regarding the `std::function<...>`, I would keep the `Clock` class, as IMO it better describes its usage

libminifi/include/controllers/NetworkPrioritizerService.h
fgerlits: The service and the unit test.  If only the service owned it, the unit test would need to have a naked pointer (or reference) to the clock, and rely on the order of the operations to make sure it remains valid -- I have considered this, but did not like it.  Since the `clock_` field is only created once, and never copied, there isn't much of an overhead.

libminifi/include/utils/TimeUtil.h
fgerlits: Yes, that would be another way of doing it, but I'm not sure it would be simpler.  Do you think it would be so much simpler that it's worth rewriting this?

szaszm: Currently I see the `Clock` class hierarchy as just related functions in disguise. This change would allow us to get rid of the classes and replace them with free functions (e.g. `get_time`, `get_steady_clock_time` and a lambda with a captured local for the manual case), which is closer to the truth IMO.

I think functions are simpler than classes (one operation vs multiple operations + data), so for my code, I would opt for that approach, but I'm totally fine with you not sharing my viewpoint and keeping the code as is.

For the readability argument, I'm arguing that the usage we describe is retrieving the time, and that can be described by naming the functions and the type-erased member something along the lines of `get_time_`.

libminifi/include/controllers/NetworkPrioritizerService.h
adamdebreceni: info(not ment to argue either for or against) : if we want to reference a member of a `shared_ptr` pointee, and not accidentally mess up, we can share the lifetime of the pointee while pointing to the member, so the "container" is kept alive even if we are only using its member

hunyadi-dev (approved):
I prefer this version to the one with a member function hidden as variable. Approved both of them though.

PR #961, opened 2020-12-16
MINIFICPP-1440 PublishKafka processor's "security protocol" should ha…

dam4rus:
…ve allowable values

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/librdkafka/PublishKafka.cpp
+++ b/extensions/librdkafka/PublishKafka.cpp
@@ -126,7 +126,11 @@ const core::Property PublishKafka::CompressCodec(
 const core::Property PublishKafka::MaxFlowSegSize(
     core::PropertyBuilder::createProperty("Max Flow Segment Size")->withDescription("Maximum flow content payload segment size for the kafka record. 0 B means unlimited.")
         ->isRequired(false)->withDefaultValue<core::DataSizeValue>("0 B")->build());
-const core::Property PublishKafka::SecurityProtocol("Security Protocol", "Protocol used to communicate with brokers", "");
+const core::Property PublishKafka::SecurityProtocol(
+        core::PropertyBuilder::createProperty("Security Protocol")
+        ->withDescription("Protocol used to communicate with brokers")
+        ->withAllowableValues<std::string>({SECURITY_PROTOCOL_PLAINTEXT, SECURITY_PROTOCOL_SSL, SECURITY_PROTOCOL_SASL_PLAINTEXT, SECURITY_PROTOCOL_SASL_SSL})
+        ->build());
 const core::Property PublishKafka::SecurityCA("Security CA", "File or directory path to CA certificate(s) for verifying the broker's key", "");
 const core::Property PublishKafka::SecurityCert("Security Cert", "Path to client's public key (PEM) used for authentication", "");
 const core::Property PublishKafka::SecurityPrivateKey("Security Private Key", "Path to client's private key (PEM) used for authentication", "");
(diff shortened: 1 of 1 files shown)


extensions/librdkafka/PublishKafka.cpp
```diff
@@ -126,7 +126,11 @@ const core::Property PublishKafka::CompressCodec(
 const core::Property PublishKafka::MaxFlowSegSize(
     core::PropertyBuilder::createProperty("Max Flow Segment Size")->withDescription("Maximum flow content payload segment size for the kafka record. 0 B means unlimited.")
         ->isRequired(false)->withDefaultValue<core::DataSizeValue>("0 B")->build());
-const core::Property PublishKafka::SecurityProtocol("Security Protocol", "Protocol used to communicate with brokers", "");
+const core::Property PublishKafka::SecurityProtocol(
+        core::PropertyBuilder::createProperty("Security Protocol")
+        ->withDescription("Protocol used to communicate with brokers")
+        ->withAllowableValues<std::string>({SECURITY_PROTOCOL_PLAINTEXT, SECURITY_PROTOCOL_SSL, SECURITY_PROTOCOL_SASL_PLAINTEXT, SECURITY_PROTOCOL_SASL_SSL})
```
szaszm, line 132: When looking at `PublishKafka.cpp:686`, it seems like only `SECURITY_PROTOCOL_SSL` is actually supported, and the default empty value is interpreted as plaintext.

I suggest restricting the set of "allowable values" to plaintext and ssl, and handling plaintext near `PublishKafka.cpp:687`. I think doing nothing is a valid handling of plaintext.

szaszm (commented):
Thanks for your first contribution. We generally merge PRs after 2-3 approvals and when all comments are resolved. In case there are any questions, feel free to reach out.

extensions/librdkafka/PublishKafka.cpp
hunyadi-dev: I would also add a default value for `PLAINTEXT` as we are trying to make the behaviour consistent with NiFi:
https://nifi.apache.org/docs/nifi-docs/components/nifi-docs/components/org.apache.nifi/nifi-kafka-2-0-nar/1.9.0/org.apache.nifi.processors.kafka.pubsub.PublishKafka_2_0/

dam4rus: > When looking at `PublishKafka.cpp:686`, it seems like only `SECURITY_PROTOCOL_SSL` is actually supported, and the default empty value is interpreted as plaintext.

Should I leave the unused defines in? If it doesn't breaks backward compatibility removing them would make the code a bit clearer.

> I suggest restricting the set of "allowable values" to plaintext and ssl, and handling plaintext near PublishKafka.cpp:687. I think doing nothing is a valid handling of plaintext.

What's the suggested coding style in this case? Having an empty `else if (value == SECURITY_PROTOCOL_PLAINTEXT)` branch or just using `else if (value != SECURITY_PROTOCOL_PLAINTEXT)` to handle unexpected values?

> I would also add a default value for PLAINTEXT as we are trying to make the behaviour consistent with NiFi:
https://nifi.apache.org/docs/nifi-docs/components/nifi-docs/components/org.apache.nifi/nifi-kafka-2-0-nar/1.9.0/org.apache.nifi.processors.kafka.pubsub.PublishKafka_2_0/

Nifi also describes it as a required value. Should I add that too?

hunyadi-dev: 1. Feel free to delete unused parts.
2. Up to you. If you feel fancy, you can extend the property API to be able to return the allowable values and check against presence there.
3. Yes.  The more important thing is that we are compatible with NiFi configs, but compatibility in the other direction is probably nice to have.

szaszm (approved):

fgerlits (approved):

hunyadi-dev (approved):

PR #964, opened 2020-12-16
MINIFICPP-1443 Fix 'possible loss of data' warnings

fgerlits:
... and a few others which were related or nearby in the code.

https://issues.apache.org/jira/browse/MINIFICPP-1443

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/rocksdb-repos/DatabaseContentRepository.cpp
+++ b/extensions/rocksdb-repos/DatabaseContentRepository.cpp
@@ -17,11 +17,14 @@
  */
 
 #include "DatabaseContentRepository.h"
+
 #include <memory>
 #include <string>
+
 #include "RocksDbStream.h"
 #include "rocksdb/merge_operator.h"
 #include "utils/GeneralUtils.h"
+#include "utils/gsl.h"
 #include "Exception.h"
 
 namespace org {
@@ -84,7 +87,7 @@ void DatabaseContentRepository::Session::commit() {
     if (outStream == nullptr) {
       throw Exception(REPOSITORY_EXCEPTION, "Couldn't open the underlying resource for write: " + resource.first->getContentFullPath());
     }
-    const auto size = resource.second->size();
+    const int size = gsl::narrow<int>(resource.second->size());
     if (outStream->write(const_cast<uint8_t*>(resource.second->getBuffer()), size) != size) {
       throw Exception(REPOSITORY_EXCEPTION, "Failed to write new resource: " + resource.first->getContentFullPath());
     }
@@ -94,7 +97,7 @@ void DatabaseContentRepository::Session::commit() {
     if (outStream == nullptr) {
       throw Exception(REPOSITORY_EXCEPTION, "Couldn't open the underlying resource for append: " + resource.first->getContentFullPath());
     }
-    const auto size = resource.second->size();
+    const int size = gsl::narrow<int>(resource.second->size());
     if (outStream->write(const_cast<uint8_t*>(resource.second->getBuffer()), size) != size) {
       throw Exception(REPOSITORY_EXCEPTION, "Failed to append to resource: " + resource.first->getContentFullPath());
     }
--- a/extensions/rocksdb-repos/FlowFileRepository.cpp
+++ b/extensions/rocksdb-repos/FlowFileRepository.cpp
@@ -17,10 +17,6 @@
  */
 #include "FlowFileRepository.h"
 
-#include <rocksdb/options.h>
-#include <rocksdb/write_batch.h>
-#include <rocksdb/slice.h>
-
 #include <chrono>
 #include <memory>
 #include <string>
@@ -28,6 +24,9 @@
 #include <vector>
 #include <list>
 
+#include "rocksdb/options.h"
+#include "rocksdb/write_batch.h"
+#include "rocksdb/slice.h"
 #include "FlowFileRecord.h"
 #include "utils/gsl.h"
 
--- a/extensions/rocksdb-repos/FlowFileRepository.cpp
+++ b/extensions/rocksdb-repos/FlowFileRepository.cpp
@@ -15,12 +15,11 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#include "FlowFileRecord.h"
 #include "FlowFileRepository.h"
 
-#include "rocksdb/options.h"
-#include "rocksdb/write_batch.h"
-#include "rocksdb/slice.h"
+#include <rocksdb/options.h>
+#include <rocksdb/write_batch.h>
+#include <rocksdb/slice.h>
 
 #include <chrono>
 #include <memory>
@@ -28,6 +27,9 @@
 #include <utility>
 #include <vector>
 #include <list>
+
+#include "FlowFileRecord.h"
+#include "utils/gsl.h"
 
 namespace org {
 namespace apache {
@@ -68,7 +70,7 @@ void FlowFileRepository::flush() {
     }
 
     utils::Identifier containerId;
-    auto eventRead = FlowFileRecord::DeSerialize(reinterpret_cast<const uint8_t *>(values[i].data()), values[i].size(), content_repo_, containerId);
+    auto eventRead = FlowFileRecord::DeSerialize(reinterpret_cast<const uint8_t *>(values[i].data()), gsl::narrow<int>(values[i].size()), content_repo_, containerId);
     if (eventRead) {
       purgeList.push_back(eventRead);
     }
@@ -152,7 +154,7 @@ void FlowFileRepository::prune_stored_flowfiles() {
   auto it = opendb->NewIterator(rocksdb::ReadOptions());
   for (it->SeekToFirst(); it->Valid(); it->Next()) {
     utils::Identifier containerId;
-    auto eventRead = FlowFileRecord::DeSerialize(reinterpret_cast<const uint8_t *>(it->value().data()), it->value().size(), content_repo_, containerId);
+    auto eventRead = FlowFileRecord::DeSerialize(reinterpret_cast<const uint8_t *>(it->value().data()), gsl::narrow<int>(it->value().size()), content_repo_, containerId);
     std::string key = it->key().ToString();
     if (eventRead) {
       // on behalf of the just resurrected persisted instance
--- a/libminifi/include/provenance/Provenance.h
+++ b/libminifi/include/provenance/Provenance.h
@@ -18,12 +18,13 @@
 #ifndef LIBMINIFI_INCLUDE_PROVENANCE_PROVENANCE_H_
 #define LIBMINIFI_INCLUDE_PROVENANCE_PROVENANCE_H_
 
-#include <memory>
+#include <algorithm>
 #include <atomic>
 #include <cstdint>
 #include <cstring>
 #include <iostream>
 #include <map>
+#include <memory>
 #include <set>
 #include <string>
 #include <thread>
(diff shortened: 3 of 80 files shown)


libminifi/src/provenance/Provenance.cpp
```diff
@@ -178,8 +181,8 @@ bool ProvenanceEventRecord::Serialize(org::apache::nifi::minifi::io::BufferStrea
 
   if (this->_eventType == ProvenanceEventRecord::FORK || this->_eventType == ProvenanceEventRecord::CLONE || this->_eventType == ProvenanceEventRecord::JOIN) {
     // write UUIDs
-    uint32_t number = this->_parentUuids.size();
-    ret = outStream.write(number);
+    uint32_t parent_uuids_size = gsl::narrow<uint32_t>(this->_parentUuids.size());
```
adamdebreceni, line 184: an alternative name could be `parent_uuids_count`, although not sure if it is "better"

libminifi/src/provenance/Provenance.cpp
```diff
@@ -189,8 +192,8 @@ bool ProvenanceEventRecord::Serialize(org::apache::nifi::minifi::io::BufferStrea
         return false;
       }
     }
-    number = this->_childrenUuids.size();
-    ret = outStream.write(number);
+    uint32_t chldren_uuids_size = gsl::narrow<uint32_t>(this->_childrenUuids.size());
```
adamdebreceni, line 195: typo in variable name, missing `i` from `children`

libminifi/src/utils/StringUtils.cpp
```diff
@@ -95,21 +95,23 @@ bool StringUtils::StringToFloat(std::string input, float &output, FailurePolicy
...
     }
-    if (beg_seq < 0)
-      break;
     end_seq = source_string.find("}", beg_seq + 2);
-    if (end_seq < 0)
+    if (end_seq == std::string::npos) {
       break;
-    if (end_seq - (beg_seq + 2) < 0) {
+    }
+    if (end_seq < beg_seq + 2) {
```
adamdebreceni, line 114: isn't this condition always false?

libminifi/src/utils/StringUtils.cpp
```diff
@@ -95,21 +95,23 @@ bool StringUtils::StringToFloat(std::string input, float &output, FailurePolicy
 }
 
 std::string StringUtils::replaceEnvironmentVariables(std::string& original_string) {
-  int32_t beg_seq = 0;
-  int32_t end_seq = 0;
+  size_t beg_seq = 0;
```
adamdebreceni, line 98: should we use `std::string::size_type`?

libminifi/test/unit/ZlibStreamTests.cpp
```diff
@@ -97,7 +106,8 @@ TEST_CASE("gzip compression and decompression pipeline", "[basic]") {
     for (size_t i = 0U; i < 1024U; i++) {
       std::generate(buf.begin(), buf.end(), [&](){return dist(gen);});
       original += std::string(reinterpret_cast<const char*>(buf.data()), buf.size());
-      REQUIRE(buf.size() == compressStream.write(buf.data(), buf.size()));
+      int buffer_size = gsl::narrow<int>(buf.size());
+      REQUIRE(buffer_size == compressStream.write(buf.data(), gsl::narrow<int>(buffer_size)));
```
adamdebreceni, line 110: the `gsl::narrow` at line `110` is not necessary

main/AgentDocs.cpp
```diff
@@ -47,7 +47,7 @@ std::string AgentDocs::extractClassName(const std::string &processor) const {
   auto lastOfIdx = processor.find_last_of(".");
   if (lastOfIdx != std::string::npos) {
     lastOfIdx++;  // if a value is found, increment to move beyond the .
-    int nameLength = processor.length() - lastOfIdx;
+    size_t nameLength = processor.length() - lastOfIdx;
     std::string processorName = processor.substr(lastOfIdx, nameLength);
```
adamdebreceni, line 51: similarly to the ones you replaced the `nameLength` is likewise unnecessary

nanofi/tests/CTestsBase.h
```diff
@@ -138,7 +138,8 @@ class TestControllerWithTemporaryWorkingDirectory {
   }
 
   ~TestControllerWithTemporaryWorkingDirectory() {
-    chdir(old_cwd_);
+    int success = chdir(old_cwd_);
+    (void) success;
```
adamdebreceni, line 142: is this added to silence an "unused return value" warning?

nanofi/include/sitetosite/CPeer.h
```diff
@@ -79,15 +79,15 @@ static void setHostName(struct SiteToSiteCPeer * peer, const char * hostname) {
...
   }
   return;
 }
 
 static void setPort(struct SiteToSiteCPeer * peer, uint16_t port) {
   peer->_port = port;
   if(peer->_url != NULL) {
-    int i;
+    size_t i;
     for(i = strlen(peer->_url) -1; i >= 0; --i) {  // look for the last ':' in the string
```
adamdebreceni, line 91: won't this result in an infinite loop?

libminifi/src/provenance/Provenance.cpp
fgerlits: yes, I think it's slightly better -- changed it to `..._count`

libminifi/src/utils/StringUtils.cpp
fgerlits: Yes.  I think the intention was to check for `<=`, as a `"... ${} ..."` input causes an infinite loop without this check.  I have changed it to `<=` and added some more unit tests.  (Which showed another bug, which I also fixed.)

libminifi/src/utils/StringUtils.cpp
fgerlits: I think we can assume that `string::size_type` will always be `size_t`, but you are right, `string::size_type` is more correct, so I changed it.

nanofi/tests/CTestsBase.h
fgerlits: yes

nanofi/include/sitetosite/CPeer.h
fgerlits: Good catch, thanks!  I have changed the type of `i` back to `int`.

nanofi/include/sitetosite/CPeer.h
```diff
@@ -88,7 +88,7 @@ static void setPort(struct SiteToSiteCPeer * peer, uint16_t port) {
   peer->_port = port;
   if(peer->_url != NULL) {
     int i;
-    for(i = strlen(peer->_url) -1; i >= 0; --i) {  // look for the last ':' in the string
+    for(i = (int)(strlen(peer->_url)) - 1; i >= 0; --i) {  // look for the last ':' in the string
```
adamdebreceni, line 91: the google coding guideline suggests using `static_cast` here (some other places touched by this PR also use C-style casts in the old code)

fgerlits: This is nanofi code, which is supposed to be in C.  This is not entirely true, as there is some C++ code in it already, but I did not want to add more.

libminifi/src/utils/StringUtils.cpp
```diff
@@ -94,37 +94,39 @@ bool StringUtils::StringToFloat(std::string input, float &output, FailurePolicy
...
       beg_seq += 2;
       continue;
     }
-    const std::string env_field = source_string.substr(beg_seq + 2, end_seq - (beg_seq + 2));
-    const std::string env_field_wrapped = source_string.substr(beg_seq, end_seq + 1);
-    if (env_field.empty()) {
+    auto env_var_length = end_seq - (beg_seq + 2);
+    const std::string env_var = source_string.substr(beg_seq + 2, env_var_length);
+    const std::string env_var_wrapped = source_string.substr(beg_seq, env_var_length + 3);
+    if (env_var.empty()) {
```
adamdebreceni, line 120: it seems like this condition is not possible anymore given line `113`

fgerlits: Good point.  I have removed it.

adamdebreceni (approved):

extensions/rocksdb-repos/FlowFileRepository.cpp
```diff
@@ -15,12 +15,11 @@
...
  * limitations under the License.
  */
-#include "FlowFileRecord.h"
 #include "FlowFileRepository.h"
 
-#include "rocksdb/options.h"
-#include "rocksdb/write_batch.h"
-#include "rocksdb/slice.h"
+#include <rocksdb/options.h>
+#include <rocksdb/write_batch.h>
```
arpadboda, line 21: Nitpicking: these are 3rd party headers.
As we fix the order here, I don't think these should be included before stl headers.

libminifi/include/provenance/Provenance.h
```diff
@@ -360,7 +361,7 @@ class ProvenanceEventRecord : public core::SerializableComponent {
   bool DeSerialize(const std::shared_ptr<core::SerializableComponent> &repo);
 
   uint64_t getEventTime(const uint8_t *buffer, const size_t bufferSize) {
-    int size = bufferSize > 72 ? 72 : bufferSize;
+    int size = gsl::narrow<int>(bufferSize > 72 ? 72 : bufferSize);
```
arpadboda, line 364: I think this is simply an std::min written in complex form :)

libminifi/include/utils/StringUtils.h
```diff
@@ -159,9 +160,9 @@ class StringUtils {
   }
 
   inline static std::string hex_ascii(const std::string& in) {
-    int len = in.length();
+    size_t len = in.length();
```
arpadboda, line 163: I think we can simply remove the variable, it's only referenced in the header of the loop. 

Completely unrelated, possible optimisation here: guess the lenght of the output and reserve place for the result string with that size to avoid reallocations caused by push_back.

arpadboda (commented):
Mostly looks good, added some minor comments.

extensions/rocksdb-repos/FlowFileRepository.cpp
fgerlits: fixed

libminifi/include/provenance/Provenance.h
fgerlits: true -- I've replaced it with `min`

libminifi/include/utils/StringUtils.h
fgerlits: I have inlined the variable, and added a `reserve` (though I don't think this will increase performance by a measurable amount).

arpadboda (approved):

extensions/rocksdb-repos/DatabaseContentRepository.cpp
```diff
@@ -94,7 +97,7 @@ void DatabaseContentRepository::Session::commit() {
     if (outStream == nullptr) {
       throw Exception(REPOSITORY_EXCEPTION, "Couldn't open the underlying resource for append: " + resource.first->getContentFullPath());
     }
-    const auto size = resource.second->size();
+    const int size = gsl::narrow<int>(resource.second->size());
```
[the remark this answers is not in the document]
fgerlits: @arpadboda was quicker than you :)

yes, this would have been an improvement, but I don't think it's essential, so let's leave it as it is, for now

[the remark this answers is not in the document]
szaszm: I know, it's fine :)

PR #968, opened 2021-01-04
MINIFICPP-1445 - Refactor docker integration test frame

hunyadi-dev:
No functional changes were introduced in this PR, just separating integration test components  into python modules based on their functionality (and wrapping config serializers into classes).

The new directory structure is:
```bash
➜  tree docker/test/integration/minifi

minifi
├── core
│   ├── Cluster.py
│   ├── Connectable.py
│   ├── ControllerService.py
│   ├── __init__.py
│   ├── InputPort.py
│   ├── Processor.py
│   ├── RemoteProcessGroup.py
│   ├── SingleNodeDockerCluster.py
│   └── SSLContextService.py
├── flow_serialization
│   ├── __init__.py
│   ├── Minifi_flow_yaml_serializer.py
│   └── Nifi_flow_xml_serializer.py
├── __init__.py
├── processors
│   ├── GenerateFlowFile.py
│   ├── GetFile.py
│   ├── __init__.py
│   ├── InvokeHTTP.py
│   ├── ListenHTTP.py
│   ├── LogAttribute.py
│   ├── PublishKafka.py
│   ├── PublishKafkaSSL.py
│   ├── PutFile.py
│   └── PutS3Object.py
└── test
    └── __init__.py

4 directories, 24 files
```

Also, added __pycache__ directories to the .gitignore.


lordgamez (approved):

arpadboda:
Seems to conflict with a recently merged PR, could you address that?

docker/test/integration/test_s2s.py
```diff
@@ -14,8 +14,6 @@
 # limitations under the License.
 
 from minifi import *
-from minifi.test import *
```
arpadboda, line 17: Seeing import * deleted always warms my heart :)

arpadboda (approved):
LGTM!

Let's wait for the CI to complete.

docker/test/integration/test_s2s.py
hunyadi-dev: We still import everything, except now we don't have two separate modules that export them rather a single one. However, **now at last it is possible** to import only what is needed.

arpadboda: The main reason behind my feelings on import * is that it's impossible to find the source of a given symbol when * is used across the code. So we world has just become a better place. At least our code is a bit easier to read. :)

arpadboda (approved):

PR #970, opened 2021-01-05
MINIFICPP-1396 Create FetchS3Object processor

lordgamez:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/PROCESSORS.md
+++ b/PROCESSORS.md
@@ -415,7 +415,7 @@ In the list below, the names of required properties appear in bold. Any other pr
 |Proxy Username|||Username to set when authenticating against proxy<br/>**Supports Expression Language: true**|
 |Proxy Password|||Password to set when authenticating against proxy<br/>**Supports Expression Language: true**|
 |Version|||The Version of the Object to download<br/>**Supports Expression Language: true**|
-|Requester Pays|false||If true, indicates that the requester consents to pay any charges associated with retrieving objects from the S3 bucket. This sets the 'x-amz-request-payer' header to 'requester'.|
+|**Requester Pays**|false||If true, indicates that the requester consents to pay any charges associated with retrieving objects from the S3 bucket. This sets the 'x-amz-request-payer' header to 'requester'.|
 ### Relationships
 
 | Name | Description |
--- a/PROCESSORS.md
+++ b/PROCESSORS.md
@@ -408,7 +408,7 @@ In the list below, the names of required properties appear in bold. Any other pr
 |Credentials File|||Path to a file containing AWS access key and secret key in properties file format. Properties used: accessKey and secretKey|
 |AWS Credentials Provider service|||The name of the AWS Credentials Provider controller service that is used to obtain AWS credentials.|
 |**Region**|us-west-2|af-south-1<br/>ap-east-1<br/>ap-northeast-1<br/>ap-northeast-2<br/>ap-northeast-3<br/>ap-south-1<br/>ap-southeast-1<br/>ap-southeast-2<br/>ca-central-1<br/>cn-north-1<br/>cn-northwest-1<br/>eu-central-1<br/>eu-north-1<br/>eu-south-1<br/>eu-west-1<br/>eu-west-2<br/>eu-west-3<br/>me-south-1<br/>sa-east-1<br/>us-east-1<br/>us-east-2<br/>us-gov-east-1<br/>us-gov-west-1<br/>us-west-1<br/>us-west-2|AWS Region|
-|**Communications Timeout**|30 sec|||
+|**Communications Timeout**|30 sec||Sets the timeout of the communication between the AWS server and the client|
 |Endpoint Override URL|||Endpoint URL to use instead of the AWS default including scheme, host, port, and path. The AWS libraries select an endpoint URL based on the AWS region, but this property overrides the selected endpoint URL, allowing use with other S3-compatible endpoints.<br/>**Supports Expression Language: true**|
 |Proxy Host|||Proxy host name or IP<br/>**Supports Expression Language: true**|
 |Proxy Port|||The port number of the proxy host<br/>**Supports Expression Language: true**|
--- a/PROCESSORS.md
+++ b/PROCESSORS.md
@@ -17,6 +17,7 @@
 - [ExecuteScript](#executescript)
 - [ExtractText](#extracttext)
 - [FetchOPCProcessor](#fetchopcprocessor)
+- [FetchS3Object](#fetchs3object)
 - [FetchSFTP](#fetchsftp)
 - [FocusArchiveEntry](#focusarchiveentry)
 - [GenerateFlowFile](#generateflowfile)
@@ -387,6 +388,40 @@ In the list below, the names of required properties appear in bold. Any other pr
 | - | - |
 |failure|Retrieved OPC-UA nodes where value cannot be extracted (only if enabled)|
 |success|Successfully retrieved OPC-UA nodes|
+
+
+## FetchS3Object
+
+### Description
+
+Retrieves the contents of an S3 Object and writes it to the content of a FlowFile
+### Properties
+
+In the list below, the names of required properties appear in bold. Any other properties (not in bold) are considered optional. The table also indicates any default values, and whether a property supports the NiFi Expression Language.
+
+| Name | Default Value | Allowable Values | Description |
+| - | - | - | - |
+|Object Key|||The key of the S3 object. If none is given the filename attribute will be used by default.<br/>**Supports Expression Language: true**|
+|**Bucket**|||The S3 bucket<br/>**Supports Expression Language: true**|
+|Access Key|||AWS account access key<br/>**Supports Expression Language: true**|
+|Secret Key|||AWS account secret key<br/>**Supports Expression Language: true**|
+|Credentials File|||Path to a file containing AWS access key and secret key in properties file format. Properties used: accessKey and secretKey|
+|AWS Credentials Provider service|||The name of the AWS Credentials Provider controller service that is used to obtain AWS credentials.|
+|**Region**|us-west-2|af-south-1<br/>ap-east-1<br/>ap-northeast-1<br/>ap-northeast-2<br/>ap-northeast-3<br/>ap-south-1<br/>ap-southeast-1<br/>ap-southeast-2<br/>ca-central-1<br/>cn-north-1<br/>cn-northwest-1<br/>eu-central-1<br/>eu-north-1<br/>eu-south-1<br/>eu-west-1<br/>eu-west-2<br/>eu-west-3<br/>me-south-1<br/>sa-east-1<br/>us-east-1<br/>us-east-2<br/>us-gov-east-1<br/>us-gov-west-1<br/>us-west-1<br/>us-west-2|AWS Region|
+|**Communications Timeout**|30 sec|||
+|Endpoint Override URL|||Endpoint URL to use instead of the AWS default including scheme, host, port, and path. The AWS libraries select an endpoint URL based on the AWS region, but this property overrides the selected endpoint URL, allowing use with other S3-compatible endpoints.<br/>**Supports Expression Language: true**|
+|Proxy Host|||Proxy host name or IP<br/>**Supports Expression Language: true**|
+|Proxy Port|||The port number of the proxy host<br/>**Supports Expression Language: true**|
+|Proxy Username|||Username to set when authenticating against proxy<br/>**Supports Expression Language: true**|
+|Proxy Password|||Password to set when authenticating against proxy<br/>**Supports Expression Language: true**|
+|Version|||The Version of the Object to download<br/>**Supports Expression Language: true**|
+|Requester Pays|false||If true, indicates that the requester consents to pay any charges associated with retrieving objects from the S3 bucket. This sets the 'x-amz-request-payer' header to 'requester'.|
+### Relationships
+
+| Name | Description |
+| - | - |
+|failure|FlowFiles are routed to failure relationship|
+|success|FlowFiles are routed to success relationship|
 
 
 ## FetchSFTP
--- a/PROCESSORS.md
+++ b/PROCESSORS.md
@@ -219,7 +219,7 @@ Deletes FlowFiles on an Amazon S3 Bucket. If attempting to delete a file that do
 
 In the list below, the names of required properties appear in bold. Any other properties (not in bold) are considered optional. The table also indicates any default values, and whether a property supports the NiFi Expression Language.
 
-| Name | Default Value | Allowable Values | Description |
+| Name | <div style="width:7em">Default Value</div> | <div style="width:8em">Allowable Values</div> | Description |
 | - | - | - | - |
 |Object Key|||The key of the S3 object. If none is given the filename attribute will be used by default.<br/>**Supports Expression Language: true**|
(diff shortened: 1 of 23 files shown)


extensions/aws/processors/FetchS3Object.cpp
```diff
@@ -0,0 +1,122 @@
...
+    return;
+  }
+
+  if (!getExpressionLanguageSupportedProperties(context, flow_file)) {
+    session->transfer(flow_file, Failure);
+    return;
+  }
+
+  WriteCallback callback(flow_file->getSize(), get_object_params_, s3_wrapper_.get());
+  session->write(flow_file, &callback);
```
arpadboda, line 91: I wonder if it may throw anyhow. Or just fails?
I don't say that the code is wrong, just would like to double-check to make sure we don't execute rollbacks in case there is an issue with one specific flowfile.

extensions/aws/processors/FetchS3Object.h
```diff
@@ -0,0 +1,110 @@
...
+class FetchS3Object : public S3Processor {
+ public:
+  static constexpr char const* ProcessorName = "FetchS3Object";
+
+  // Supported Properties
+  static const core::Property Version;
+  static const core::Property RequesterPays;
+
+  // Supported Relationships
+  static const core::Relationship Failure;
```
arpadboda, line 51: Is there any aws processor these two relationships are not applicable to?
In case the answer is yes, they are fine here, otherwise I would move them to the base.

extensions/aws/s3/S3WrapperBase.cpp
```diff
@@ -32,6 +33,17 @@ namespace minifi {
...
 namespace s3 {
 
+void GetObjectResult::setFilePaths(const std::string& key) {
+  const auto last_slash = key.find_last_of('/');
+  absolute_path = key;
+  if (last_slash != std::string::npos) {
+    path = key.substr(0, last_slash);
+    filename = key.substr(last_slash + 1);
+  } else {
+    filename = key;
```
arpadboda, line 43: Don't we have a splitter in pathutils or stringutils already?

arpadboda (commented):
Mostly looks good, some minor comments.

extensions/aws/processors/FetchS3Object.cpp
lordgamez: After checking the code there seems to be no way for this to throw. The callback can only fail and return 0 if failure happens. The write may throw an exception only if the return value of the callback is less than 0.

extensions/aws/processors/FetchS3Object.h
lordgamez: The next S3 processor depending on this is the ListS3 which only has a success relationship.

extensions/aws/s3/S3WrapperBase.cpp
lordgamez: You are right there is a getFileNameAndPath in pathutils, but that uses OS specific separators, but in AWS the object key always uses '/' character as a separator.

arpadboda: I think some of our such util functions already have a "force posix" bool argument to address this, I think we can apply that for this specific function and use it here.

extensions/aws/processors/FetchS3Object.cpp
arpadboda: Okay, thanks.

extensions/aws/s3/S3WrapperBase.cpp
lordgamez: Found it in the FileUtils, the split_path function does what we need here, thanks! Added it in [fd2e4fa](https://github.com/apache/nifi-minifi-cpp/pull/970/commits/fd2e4facc77e51e6e6fe05280c9ab6d563ee5114)

extensions/aws/s3/S3WrapperBase.cpp
```diff
@@ -68,13 +75,13 @@ void S3WrapperBase::setCannedAcl(Aws::S3::Model::PutObjectRequest& request, cons
   request.SetACL(CANNED_ACL_MAP.at(canned_acl));
 }
 
-std::string S3WrapperBase::getExpiryDate(const std::string& expiration) {
+std::pair<std::string, std::string> S3WrapperBase::getExpirationPair(const std::string& expiration) {
   static const std::regex expr = std::regex("expiry-date=\"(.*)\", rule-id=\"(.*)\"");
```
adamdebreceni, line 79: shouldn't we use the `Regex` class defined in `RegexUtils.h`? (that contains some "magic" for `<GCC 4.9`)

extensions/aws/s3/S3WrapperBase.cpp
```diff
@@ -68,13 +75,13 @@ void S3WrapperBase::setCannedAcl(Aws::S3::Model::PutObjectRequest& request, cons
   request.SetACL(CANNED_ACL_MAP.at(canned_acl));
 }
 
-std::string S3WrapperBase::getExpiryDate(const std::string& expiration) {
+std::pair<std::string, std::string> S3WrapperBase::getExpirationPair(const std::string& expiration) {
```
adamdebreceni, line 78: even though it is only used twice, we could create a `struct` (given how both members of the pair are of the same type)

extensions/aws/s3/S3WrapperBase.cpp
arpadboda: Good shout

lordgamez: Fixed in [8ddbc42](https://github.com/apache/nifi-minifi-cpp/pull/970/commits/8ddbc42c1e49ffcbd7764be5b95535bbf9e65813)

extensions/aws/s3/S3WrapperBase.cpp
lordgamez: Fixed in [ad6ee95](https://github.com/apache/nifi-minifi-cpp/pull/970/commits/ad6ee955fc776b4816c32b56a802e5a5e30fe41c)

extensions/aws/processors/FetchS3Object.cpp
```diff
@@ -0,0 +1,122 @@
...
+}
+
+bool FetchS3Object::getExpressionLanguageSupportedProperties(
+    const std::shared_ptr<core::ProcessContext> &context,
+    const std::shared_ptr<core::FlowFile> &flow_file) {
+  if (!S3Processor::getExpressionLanguageSupportedProperties(context, flow_file)) {
+    return false;
+  }
+  get_object_params_.bucket = std::move(bucket_);
+  get_object_params_.object_key = std::move(object_key_);
```
adamdebreceni, line 70: it seems like `object_key_` and `bucket_` are a processor members, and we read/write them from possible multiple concurrent `onTrigger`s

adamdebreceni: the same goes for, I think, `aws_credentials_provider_`and `s3_wrapper_` where we call `S3WrapperBase::setProxy` and `S3WrapperBase::setCredentials` from the `onTrigger`s

lordgamez: Could you please elaborate on this a bit more? I am not that familiar with the scheduling of the onTrigger calls. Are you saying that it is possible that the same processor's onTrigger call can be scheduled on separate threads in the threadpool and depending on the timing they might overlap causing the concurrency?

arpadboda: Yes, this can happen depending on configurations.
In most of the processors we just prevent this by having a mutex for the whole ontrigger call, in this case we might just protect some common resources.

adamdebreceni: yes, as far as I know, `onSchedule` is guaranteed to be called once for a processor, but multiple `onTrigger`s can run concurrently on multiple threads, the number of such `onTrigger`s is controlled by the `max concurrent tasks` property of the processor

adamdebreceni: of course we could always just ignore this property and override `getMaxConcurrentTasks` to return `1`, but I don't (yet) think there is an inherent reason to do so, for this processor (family)

PROCESSORS.md
```diff
@@ -389,6 +390,40 @@ In the list below, the names of required properties appear in bold. Any other pr
...
+
+| Name | Default Value | Allowable Values | Description |
+| - | - | - | - |
+|Object Key|||The key of the S3 object. If none is given the filename attribute will be used by default.<br/>**Supports Expression Language: true**|
+|**Bucket**|||The S3 bucket<br/>**Supports Expression Language: true**|
+|Access Key|||AWS account access key<br/>**Supports Expression Language: true**|
+|Secret Key|||AWS account secret key<br/>**Supports Expression Language: true**|
+|Credentials File|||Path to a file containing AWS access key and secret key in properties file format. Properties used: accessKey and secretKey|
+|AWS Credentials Provider service|||The name of the AWS Credentials Provider controller service that is used to obtain AWS credentials.|
+|**Region**|us-west-2|af-south-1<br/>ap-east-1<br/>ap-northeast-1<br/>ap-northeast-2<br/>ap-northeast-3<br/>ap-south-1<br/>ap-southeast-1<br/>ap-southeast-2<br/>ca-central-1<br/>cn-north-1<br/>cn-northwest-1<br/>eu-central-1<br/>eu-north-1<br/>eu-south-1<br/>eu-west-1<br/>eu-west-2<br/>eu-west-3<br/>me-south-1<br/>sa-east-1<br/>us-east-1<br/>us-east-2<br/>us-gov-east-1<br/>us-gov-west-1<br/>us-west-1<br/>us-west-2|AWS Region|
```
hunyadi-dev, line 410: Very minor, but this list of allowable values does not look well at all when rendered:
image

It might be worth doing some manual tweaking on it, eg changing font size and adding some padding to the top and bottom. Same stands for DeleteS3Object.

libminifi/test/aws-tests/FetchS3ObjectTests.cpp
```diff
@@ -0,0 +1,174 @@
...
+class FetchS3ObjectTestsFixture : public S3TestsFixture<minifi::aws::processors::FetchS3Object> {
+ public:
+  FetchS3ObjectTestsFixture() {
+    auto putfile = plan->addProcessor(
+      "PutFile",
+      "PutFile",
+      core::Relationship("success", "d"),
+      true);
+    char output_dir_mask[] = "/tmp/gt.XXXXXX";
+    output_dir = test_controller.createTempDirectory(output_dir_mask);
```
hunyadi-dev, line 33: You can get rid of the mask by using:
```c++
output_dir = utils::createTempDir(&testController);
```

libminifi/test/aws-tests/FetchS3ObjectTests.cpp
```diff
@@ -0,0 +1,174 @@
...
+  setRequiredProperties();
+  setProxy();
+  test_controller.runSession(plan, true);
+  checkProxySettings();
+}
+
+TEST_CASE_METHOD(FetchS3ObjectTestsFixture, "Test default properties", "[awsS3Config]") {
+  setRequiredProperties();
+  test_controller.runSession(plan, true);
+  REQUIRE(LogTestController::getInstance().contains("key:s3.bucket value:" + S3_BUCKET));
```
hunyadi-dev, line 106: Minor but `contains` is slow. `utils::verifyLogLinePresenceInPollTime` should be a better choice here.

libminifi/test/aws-tests/FetchS3ObjectTests.cpp
```diff
@@ -0,0 +1,174 @@
...
+    setUseDefaultCredentialsChain(aws_credentials_service);
+    setCredentialsService();
+  }
+
+  test_controller.runSession(plan, true);
+  REQUIRE(mock_s3_wrapper_ptr->getCredentials().GetAWSAccessKeyId() == "key");
+  REQUIRE(mock_s3_wrapper_ptr->getCredentials().GetAWSSecretKey() == "secret");
+}
+
+TEST_CASE_METHOD(FetchS3ObjectTestsFixture, "Test required property not set", "[awsS3Config]") {
```
hunyadi-dev, line 75: :+1:  for proper negative tests

libminifi/test/aws-tests/FetchS3ObjectTests.cpp
```diff
@@ -0,0 +1,174 @@
...
+TEST_CASE_METHOD(FetchS3ObjectTestsFixture, "Test subdirectories on AWS", "[awsS3Config]") {
+  setRequiredProperties();
+  plan->setProperty(s3_processor, "Object Key", "dir1/dir2/logs.txt");
+  test_controller.runSession(plan, true);
+  REQUIRE(LogTestController::getInstance().contains("key:filename value:logs.txt"));
+  REQUIRE(LogTestController::getInstance().contains("key:path value:dir1/dir2"));
+  REQUIRE(LogTestController::getInstance().contains("key:absolute.path value:dir1/dir2/logs.txt"));
+  std::ifstream s3_file(output_dir + "/" + INPUT_FILENAME);
+  std::string output((std::istreambuf_iterator<char>(s3_file)), std::istreambuf_iterator<char>());
+  REQUIRE(output.empty());
```
hunyadi-dev, line 152: Probably a good candidate to extract to FileUtils.h

hunyadi-dev (approved):
LGTM, only minor suggestions added.

extensions/aws/processors/FetchS3Object.cpp
lordgamez: Thanks for the info and great catch! I removed the members that were modified in the onTrigger calls and added a mutex for the S3Wrapper calls to guarantee that the S3 configuration is not changed between AWS calls. Change is done in [e59580a](https://github.com/apache/nifi-minifi-cpp/pull/970/commits/e59580a5cbd1ef78a05da67664103fcfaeb4a062)

PROCESSORS.md
lordgamez: I adjusted the cell width with a fix but scalable div in [9417f98](https://github.com/apache/nifi-minifi-cpp/pull/970/commits/9417f98d7314b35bf2d346fbd83aaa92ff58b58d). The region values should not wrap anymore.

libminifi/test/aws-tests/FetchS3ObjectTests.cpp
lordgamez: Added in [7b57ab6](https://github.com/apache/nifi-minifi-cpp/pull/970/commits/7b57ab6c5a077ae14d490eb325a3f049bf629d68)

libminifi/test/aws-tests/FetchS3ObjectTests.cpp
lordgamez: Replaced it in [840a645](https://github.com/apache/nifi-minifi-cpp/pull/970/commits/840a645918fb63a6bde7a8c702e0a58125f0f085).

libminifi/test/aws-tests/FetchS3ObjectTests.cpp
lordgamez: Extraced in [6097258](https://github.com/apache/nifi-minifi-cpp/pull/970/commits/60972586e82e9e7786adf01ce84c305e88dc7224)

adamdebreceni (approved):

PR #973, opened 2021-01-07
MINIFICPP-1446 - SQL extension doesn't compile on Debian

arpadboda:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/cmake/BundledOSSPUUID.cmake
+++ b/cmake/BundledOSSPUUID.cmake
@@ -21,11 +21,11 @@ function(use_bundled_osspuuid SOURCE_DIR BINARY_DIR)
     message("Using bundled ossp-uuid")
 
     # Define patch step
-    # Avoid failure and prompt in case already applied
-    set(PC "${Patch_EXECUTABLE}" -p1 -N -i "${SOURCE_DIR}/thirdparty/ossp-uuid/ossp-uuid-mac-fix.patch" &&
-           "${Patch_EXECUTABLE}" -p1 -N -i "${SOURCE_DIR}/thirdparty/ossp-uuid/ossp-uuid-no-prog.patch" || 
-           "${Patch_EXECUTABLE}" -p1 -R -i "${SOURCE_DIR}/thirdparty/ossp-uuid/ossp-uuid-mac-fix.patch" &&
-           "${Patch_EXECUTABLE}" -p1 -R -i "${SOURCE_DIR}/thirdparty/ossp-uuid/ossp-uuid-no-prog.patch")
+    # if already applied, reverse application should succeed
+    set(PC bash -c "set -x && (\"${Patch_EXECUTABLE}\" -p1 -N -i \"${SOURCE_DIR}/thirdparty/ossp-uuid/ossp-uuid-mac-fix.patch\" &&\
+            \"${Patch_EXECUTABLE}\" -p1 -N -i \"${SOURCE_DIR}/thirdparty/ossp-uuid/ossp-uuid-no-prog.patch\") ||\
+            (\"${Patch_EXECUTABLE}\" -p1 -R --dry-run -i \"${SOURCE_DIR}/thirdparty/ossp-uuid/ossp-uuid-mac-fix.patch\" &&\
+            \"${Patch_EXECUTABLE}\" -p1 -R --dry-run -i \"${SOURCE_DIR}/thirdparty/ossp-uuid/ossp-uuid-no-prog.patch\")")
 
     # Define byproducts
     set(BYPRODUCTS "lib/libuuid.a"
--- a/cmake/BundledOSSPUUID.cmake
+++ b/cmake/BundledOSSPUUID.cmake
@@ -21,8 +21,11 @@ function(use_bundled_osspuuid SOURCE_DIR BINARY_DIR)
     message("Using bundled ossp-uuid")
 
     # Define patch step
-    set(PC "${Patch_EXECUTABLE}" -p1 -i "${SOURCE_DIR}/thirdparty/ossp-uuid/ossp-uuid-mac-fix.patch" &&
-           "${Patch_EXECUTABLE}" -p1 -i "${SOURCE_DIR}/thirdparty/ossp-uuid/ossp-uuid-no-prog.patch")
+    # Avoid failure and prompt in case already applied
+    set(PC "${Patch_EXECUTABLE}" -p1 -N -i "${SOURCE_DIR}/thirdparty/ossp-uuid/ossp-uuid-mac-fix.patch" &&
+           "${Patch_EXECUTABLE}" -p1 -N -i "${SOURCE_DIR}/thirdparty/ossp-uuid/ossp-uuid-no-prog.patch" || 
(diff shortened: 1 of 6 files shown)


szaszm:
Not debian, but I get this error when trying to build this PR with SQL enabled
```make[2]: *** No rule to make target 'extensions/sql/thirdparty/iodbc-install//lib64/libiodbc.a', needed by 'main/minifi'.  Stop.```
on main I get a similar error, but for `libsoci_odbc.a` instead.

**Comment by arpadboda** [comment]
> Not debian, but I get this error when trying to build this PR with SQL enabled
> `make[2]: *** No rule to make target 'extensions/sql/thirdparty/iodbc-install//lib64/libiodbc.a', needed by 'main/minifi'. Stop.`
> on main I get a similar error, but for `libsoci_odbc.a` instead.

This PR was meant to address the latter, but cmake caching is so f... up that you never know when you have the good setup. Will tackle this a bit more.

**Comment by szaszm** [comment]
A bit more info after trying to build this PR:
```
szaszm@szaszm-pc:~/nifi-minifi-cpp-2/cmake-build-relwithdebinfo-gcc-large$ find -name libsoci_odbc.a
./extensions/sql/soci-external-prefix/src/soci-external-build/lib/libsoci_odbc.a
./extensions/sql/thirdparty/soci-install/lib64/libsoci_odbc.a
szaszm@szaszm-pc:~/nifi-minifi-cpp-2/cmake-build-relwithdebinfo-gcc-large$ find -name libiodbc.a
./extensions/sql/thirdparty/iodbc-install/lib/libiodbc.a
./extensions/sql/thirdparty/iodbc-src/iodbc/.libs/libiodbc.a
```

**Review by szaszm — APPROVED** [review]

File: extensions/sql/CMakeLists.txt [hunk 1]
```diff
@@ -89,11 +84,12 @@ find_package(Patch)
 set(PC "${Patch_EXECUTABLE}" -p1 -i "${CMAKE_CURRENT_SOURCE_DIR}/patch/soci.patch")
 
 # Define byproducts
-if(NOT APPLE AND CMAKE_SIZEOF_VOID_P EQUAL 8)
-	include(GNUInstallDirs)
-	set(LIBDIR "${CMAKE_INSTALL_LIBDIR}")
-else()
+# This should be based on GNUInstallDirs, but it's done wrong in Soci:
+# https://github.com/SOCI/soci/blob/release/4.0/CMakeLists.txt#L140
+if(APPLE OR CMAKE_SIZEOF_VOID_P EQUAL 4)
```
lordgamez, line 89: Shouldn't this be always "lib" for WIN32 as well?

arpadboda: This is now consistent with what the lib does internally (see the link in the comment above), so I assume this to be correct for Win as well.

lordgamez (approved):

cmake/BundledOSSPUUID.cmake
```diff
@@ -21,8 +21,8 @@ function(use_bundled_osspuuid SOURCE_DIR BINARY_DIR)
     message("Using bundled ossp-uuid")
 
     # Define patch step
-    set(PC "${Patch_EXECUTABLE}" -p1 -i "${SOURCE_DIR}/thirdparty/ossp-uuid/ossp-uuid-mac-fix.patch" &&
-           "${Patch_EXECUTABLE}" -p1 -i "${SOURCE_DIR}/thirdparty/ossp-uuid/ossp-uuid-no-prog.patch")
+    set(PC "${Patch_EXECUTABLE}" -p1 -N -r- -i "${SOURCE_DIR}/thirdparty/ossp-uuid/ossp-uuid-mac-fix.patch" &&
+           "${Patch_EXECUTABLE}" -p1 -N -r- -i "${SOURCE_DIR}/thirdparty/ossp-uuid/ossp-uuid-no-prog.patch" || :)
```
fgerlits, line 25: That `|| :` is not very nice -- is it there because of the "patch already applied" error we sometimes get?  I haven't seen that error too many times, and I'd rather do a clean rebuild (until we can fix the problem properly) than possibly ignore real errors thrown by `patch`.

szaszm: That was me, I admit it's not a nice solution. The intention was to stop patch from throwing errors when it has already been applied, but the only way was to stop it from throwing errors altogether.

I got an idea that maybe instead of `|| :` at the end, we could try reverse applying with `--dry-run` to not actually change anything. This would verify that either we could apply the patch or it could be reverse applied, i.e. it has already been applied. Would this be a good enough workaround for you or do you want to revert the whole change?

szaszm: Note for @arpadboda: A clean build works on gentoo without the changes in this file.

fgerlits: I'm OK with a change that ignores "already applied" errors only and not all errors.  I would also be OK with reverting this part of the change, so we just live with the "already applied" errors until we figure out how to fix them properly.

szaszm (approved):

fgerlits (approved):

cmake/BundledOSSPUUID.cmake
```diff
@@ -21,8 +21,11 @@ function(use_bundled_osspuuid SOURCE_DIR BINARY_DIR)
     message("Using bundled ossp-uuid")
 
     # Define patch step
-    set(PC "${Patch_EXECUTABLE}" -p1 -i "${SOURCE_DIR}/thirdparty/ossp-uuid/ossp-uuid-mac-fix.patch" &&
-           "${Patch_EXECUTABLE}" -p1 -i "${SOURCE_DIR}/thirdparty/ossp-uuid/ossp-uuid-no-prog.patch")
+    set(PC "${Patch_EXECUTABLE}" -p1 -N -i "${SOURCE_DIR}/thirdparty/ossp-uuid/ossp-uuid-mac-fix.patch" &&
+            "${Patch_EXECUTABLE}" -p1 -N -i "${SOURCE_DIR}/thirdparty/ossp-uuid/ossp-uuid-no-prog.patch" ||
+            # if already applied, reverse application should succeed
+            "${Patch_EXECUTABLE}" -p1 -R --dry-run -i "${SOURCE_DIR}/thirdparty/ossp-uuid/ossp-uuid-mac-fix.patch" &&
+            "${Patch_EXECUTABLE}" -p1 -R --dry-run -i "${SOURCE_DIR}/thirdparty/ossp-uuid/ossp-uuid-no-prog.patch")
```
fgerlits, line 28: It would be nicer to have a `( ... )` around the `--dry-run` commands.  Otherwise, if the first two patch commands succeed, we skip the third command, but execute the fourth.  This is not a big problem, as the fourth command should succeed, but it isn't nice.

fgerlits (changes requested):
testing, please ignore

cmake/BundledOSSPUUID.cmake
szaszm: https://github.com/szaszm/nifi-minifi-cpp/commit/fee00580740c578a64956d3e448c7729f9bb44b0

fgerlits (approved):

PR #974, opened 2021-01-11
MINIFICPP-1449 Add pause and resume command to C2

lordgamez:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/core/state/UpdateController.h
+++ b/libminifi/include/core/state/UpdateController.h
@@ -148,16 +148,16 @@ class UpdateRunner : public utils::AfterExecute<Update> {
   std::chrono::milliseconds delay_;
 };
 
-class PausableStateController {
+class Pausable {
  public:
-  virtual ~PausableStateController() = default;
+  virtual ~Pausable() = default;
 
   virtual int16_t pause() = 0;
 
   virtual int16_t resume() = 0;
 };
 
-class StateController : public PausableStateController {
+class StateController : public Pausable {
  public:
   virtual ~StateController() = default;
 
--- a/libminifi/include/core/state/UpdateController.h
+++ b/libminifi/include/core/state/UpdateController.h
@@ -148,7 +148,16 @@ class UpdateRunner : public utils::AfterExecute<Update> {
   std::chrono::milliseconds delay_;
 };
 
-class StateController {
+class PausableStateController {
+ public:
+  virtual ~PausableStateController() = default;
+
+  virtual int16_t pause() = 0;
+
+  virtual int16_t resume() = 0;
+};
+
+class StateController : public PausableStateController {
  public:
   virtual ~StateController() = default;
 
@@ -165,8 +174,6 @@ class StateController {
   virtual int16_t stop() = 0;
 
   virtual bool isRunning() = 0;
-
-  virtual int16_t pause() = 0;
 };
 
 /**
--- a/libminifi/include/utils/ThreadPool.h
+++ b/libminifi/include/utils/ThreadPool.h
@@ -228,18 +228,14 @@ class ThreadPool {
   void stopTasks(const TaskId &identifier);
 
   /**
-   * attempts to resume tasks with the provided identifier.
-   * @param identifier for worker tasks. Note that these tasks won't
-   * immediately start.
+   * resumes work queue processing.
    */
-  void resumeTasks(const TaskId &identifier);
+  void resume();
 
   /**
-   * attempts to pause tasks with the provided identifier.
-   * @param identifier for worker tasks. Note that these tasks won't
-   * immediately pause.
+   * pauses work queue processing
    */
-  void pauseTasks(const TaskId &identifier);
+  void pause();
 
   /**
    * Returns true if a task is running.
@@ -249,7 +245,7 @@ class ThreadPool {
     const auto iter = task_status_.find(identifier);
     if (iter == task_status_.end())
       return false;
-    return iter->second == TaskState::RUNNING;
+    return iter->second;
   }
 
   bool isRunning() const {
@@ -309,12 +305,6 @@ class ThreadPool {
   }
 
  protected:
-  enum class TaskState {
-    STOPPED,
-    PAUSED,
-    RUNNING
-  };
-
   std::thread createThread(std::function<void()> &&functor) {
     return std::thread([ functor ]() mutable {
       functor();
@@ -365,13 +355,11 @@ class ThreadPool {
 // notification for new delayed tasks that's before the current ones
   std::condition_variable delayed_task_available_;
 // map to identify if a task should be
-  std::map<TaskId, TaskState> task_status_;
+  std::map<TaskId, bool> task_status_;
 // manager mutex
   std::recursive_mutex manager_mutex_;
   // thread pool name
   std::string name_;
-  // map of paused tasks moved from worker queue that should not run at the time
-  std::unordered_map<TaskId, std::vector<Worker<T>>> paused_tasks_;
 
   /**
    * Call for the manager to start worker threads
--- a/libminifi/include/utils/ThreadPool.h
+++ b/libminifi/include/utils/ThreadPool.h
@@ -27,6 +27,7 @@
 #include <atomic>
 #include <mutex>
 #include <map>
+#include <unordered_map>
 #include <vector>
 #include <queue>
 #include <future>
@@ -370,7 +371,7 @@ class ThreadPool {
   // thread pool name
   std::string name_;
   // map of paused tasks moved from worker queue that should not run at the time
-  std::map<TaskId, Worker<T>> paused_tasks_;
+  std::unordered_map<TaskId, std::vector<Worker<T>>> paused_tasks_;
 
   /**
    * Call for the manager to start worker threads
--- a/libminifi/include/utils/ThreadPool.h
+++ b/libminifi/include/utils/ThreadPool.h
@@ -227,6 +227,20 @@ class ThreadPool {
   void stopTasks(const TaskId &identifier);
 
   /**
+   * attempts to resume tasks with the provided identifier.
+   * @param identifier for worker tasks. Note that these tasks won't
+   * immediately start.
+   */
+  void resumeTasks(const TaskId &identifier);
+
+  /**
+   * attempts to pause tasks with the provided identifier.
+   * @param identifier for worker tasks. Note that these tasks won't
+   * immediately pause.
+   */
+  void pauseTasks(const TaskId &identifier);
+
+  /**
    * Returns true if a task is running.
    */
   bool isTaskRunning(const TaskId &identifier) {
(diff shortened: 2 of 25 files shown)


libminifi/include/utils/ThreadPool.h
```diff
@@ -344,11 +364,13 @@ class ThreadPool {
...
   std::condition_variable delayed_task_available_;
 // map to identify if a task should be
-  std::map<TaskId, bool> task_status_;
+  std::map<TaskId, TaskState> task_status_;
 // manager mutex
   std::recursive_mutex manager_mutex_;
   // thread pool name
   std::string name_;
+  // map of paused tasks moved from worker queue that should not run at the time
+  std::map<TaskId, Worker<T>> paused_tasks_;
```
adamdebreceni, line 373: there could be multiple tasks with the same `TaskId` when the maxConcurrentTasks is greater than 1, this way we could discard tasks on a pause + resume

lordgamez: Thanks for the info, I was not aware of that. I modified the map to be able to hold multiple tasks for a taskid in [cd5817c](https://github.com/apache/nifi-minifi-cpp/pull/974/commits/cd5817cd7254305b382734d954ebc5b659453519)

lordgamez:
The C2PauseResumeTest is flaky because of the possible timings of the GET request and the HeartBeat messages, so it still needs to be stabilized.

arpadboda (commented):
Wonder if tasks need to have the property to indicate paused state.

So currently I think we have a dedicated threadpool for c2 (heartbeating). Because of this the flowcontroller's threadpool can be completely paused.

In this case the scheduling agent and the tasks don't need to be impacted, only the threadpool workers top picking up tasks while the flowcontroller is paused. 

What do you think?

lordgamez:
> Wonder if tasks need to have the property to indicate paused state.
> 
> So currently I think we have a dedicated threadpool for c2 (heartbeating). Because of this the flowcontroller's threadpool can be completely paused.
> 
> In this case the scheduling agent and the tasks don't need to be impacted, only the threadpool workers top picking up tasks while the flowcontroller is paused.
> 
> What do you think?

That's actually what I wanted to go with first, but a problem occurred when I implemented it. When the ThreadPool dequeues the worker queue and sees that the dequeued task is in stopped state it does not re-enqueue the task, but simply throws it away. The task remains in the task states map, but changing its state has no effect afterwards. This is why I wanted to distinguish a stopped task from a paused task by moving the paused tasks to separate map when they are processed and later re-enqueueing them when they are resumed thus keeping the original behavior of the stopped tasks.

adamdebreceni:
> > Wonder if tasks need to have the property to indicate paused state.
> > So currently I think we have a dedicated threadpool for c2 (heartbeating). Because of this the flowcontroller's threadpool can be completely paused.
> > In this case the scheduling agent and the tasks don't need to be impacted, only the threadpool workers top picking up tasks while the flowcontroller is paused.
> > What do you think?
> 
> That's actually what I wanted to go with first, but a problem occurred when I implemented it. When the ThreadPool dequeues the worker queue and sees that the dequeued task is in stopped state it does not re-enqueue the task, but simply throws it away. The task remains in the task states map, but changing its state has no effect afterwards. This is why I wanted to distinguish a stopped task from a paused task by moving the paused tasks to separate map when they are processed and later re-enqueueing them when they are resumed thus keeping the original behavior of the stopped tasks.

I feel like there are two contending "pause/resume" operation semantics, one regarding the FlowController (@arpadboda ), and one regarding the processors (@lordgamez)

1. which one is the current PR about? (it seem to me, that the first one)
2. if the first one, do we plan on extending it to support the second one?
3. if the first one, and we don't plan on supporting processor-level pause/resume why doesn't a ThreadPool simply stop dequeuing tasks on pause

lordgamez:
> > > Wonder if tasks need to have the property to indicate paused state.
> > > So currently I think we have a dedicated threadpool for c2 (heartbeating). Because of this the flowcontroller's threadpool can be completely paused.
> > > In this case the scheduling agent and the tasks don't need to be impacted, only the threadpool workers top picking up tasks while the flowcontroller is paused.
> > > What do you think?
> > 
> > 
> > That's actually what I wanted to go with first, but a problem occurred when I implemented it. When the ThreadPool dequeues the worker queue and sees that the dequeued task is in stopped state it does not re-enqueue the task, but simply throws it away. The task remains in the task states map, but changing its state has no effect afterwards. This is why I wanted to distinguish a stopped task from a paused task by moving the paused tasks to separate map when they are processed and later re-enqueueing them when they are resumed thus keeping the original behavior of the stopped tasks.
> 
> I feel like there are two contending "pause/resume" operation semantics, one regarding the FlowController (@arpadboda ), and one regarding the processors (@lordgamez)
> 
> 1. which one is the current PR about? (it seem to me, that the first one)
> 2. if the first one, do we plan on extending it to support the second one?
> 3. if the first one, and we don't plan on supporting processor-level pause/resume why doesn't a ThreadPool simply stop dequeuing tasks on pause

I suppose in our case as we only want to pause the workflow, it would suffice to stop the dequeuing the the tasks. I am not sure if we have any plans in the future to pause-resume on processor level, or if we should have that option open, maybe @arpadboda can answer that.

arpadboda:
@adamdebreceni : currently it's about the first (stopping the whole flow). Being able to stop one processor might be handy in the future. 

@lordgamez : for pausing the whole threadpool, I think the easy solution is sending back the worker threads to sleep when they wake up. 
Somewhere in ```ThreadPool<T>::run_tasks```

I just realised that stopping the worker_queue_ actually results in this.

lordgamez:
@adamdebreceni @arpadboda I changed the semantics to stop the thread pool when we pause and restart again when we resume in [89ea022](https://github.com/apache/nifi-minifi-cpp/pull/974/commits/89ea02226dd48fcd288ce82354fb9e6ba8125b74). I misunderstood a previous comment as previously I stopped the ThreadedScheduler not the ThreadPool. This time it works by only stopping the ThreadPool.

libminifi/src/FlowController.cpp
```diff
@@ -360,6 +360,33 @@ int16_t FlowController::start() {
...
 
+int16_t FlowController::pause() {
+  std::lock_guard<std::recursive_mutex> flow_lock(mutex_);
+  if (!running_) {
+    logger_->log_warn("Can not pause flow controller that is not running");
+    return 0;
+  }
+
+  logger_->log_info("Pausing Flow Controller");
+  timer_scheduler_->pause();
```
arpadboda, line 371: As far as I remember the threadpool is currently the member of flowcontroller. 
Only a reference to that is passed to all schedulers, so schedulers share a common threadpool. 

Which means that we notify all scheduling agents to indirectly pause our own threadpool, but at least it's done 3 times. :)

I don't think that the scheduling agents even need to be aware to being paused and resumed. I doens't hurt, but I think pausing the threadpool should be the responsibility of the flowcontroller, which makes this code less complex, too.

arpadboda (commented):
Mostly looks good, nice tests, one part where I disagree with the implementation.

libminifi/src/FlowController.cpp
lordgamez: Thanks, I totally missed that they shared the same threadpool! I fixed it in [e69f7cf](https://github.com/apache/nifi-minifi-cpp/pull/974/commits/e69f7cfdcff8ec6fa3aa3c4e34199b90e9761270)

arpadboda (approved):
LGTM, thanks!

arpadboda:
I think we should investigate the CI failure here:

```
C2PauseResumeTest: /home/runner/work/nifi-minifi-cpp/nifi-minifi-cpp/extensions/http-curl/tests/HTTPHandlers.h:432: void HeartbeatHandler::verify(mg_connection*): Assertion `ok' failed.
```

lordgamez:
> I think we should investigate the CI failure here:
> 
> ```
> C2PauseResumeTest: /home/runner/work/nifi-minifi-cpp/nifi-minifi-cpp/extensions/http-curl/tests/HTTPHandlers.h:432: void HeartbeatHandler::verify(mg_connection*): Assertion `ok' failed.
> ```

I could not reproduce the issue locally after a few hundred runs with Release and Debug builds. I checked the failed assertion and it is a rapidjson parse failure of either an acknowledge or a heartbeat message sent by the Minifi process. It may have been a corrupted message due to some environment related issue, I'm not sure of the root cause, but if it resurfaces we could open an new issue to the transient failures epic.

libminifi/include/core/state/UpdateController.h
```diff
@@ -148,7 +148,16 @@ class UpdateRunner : public utils::AfterExecute<Update> {
...
 
-class StateController {
+class PausableStateController {
+ public:
+  virtual ~PausableStateController() = default;
+
+  virtual int16_t pause() = 0;
+
+  virtual int16_t resume() = 0;
+};
```
szaszm, line 158: There is nothing in this interface that is related to `StateController`. This interface should probably be named `Pausable` instead. This also implies some C2Agent and related changes. The C2 part probably doesn't need to care what kind of object it pauses, but just pause and resume something that is `Pausable` and was passed as the ctor argument to be paused/resumed.

I was thinking about defining `Pausable` as a runtime concept with reference semantics, but that is only if you like the idea of not having to depend on/derive from a class to "be" `Pausable` enough to take on significantly increased implementation complexity. The implementation would be a wrapper class with a pointer to a private abstract class (concept) similar to the current one and a constructor template that instantiates a private class template (model) implementing the private concept class.

szaszm:
Please don't forget to update the [C2 Design](https://cwiki.apache.org/confluence/display/MINIFI/C2+Design) wiki page.

libminifi/include/core/state/UpdateController.h
lordgamez: I updated the name of the interface in [11cca31](https://github.com/apache/nifi-minifi-cpp/pull/974/commits/11cca310b1dd516fbdaf0e1264fbe6f0c552803e). I would stay with this implementation now instead of a Pausable concept, it seems more simple and explicit for me.

szaszm (approved):
LGTM

lordgamez:
> > I think we should investigate the CI failure here:
> > ```
> > C2PauseResumeTest: /home/runner/work/nifi-minifi-cpp/nifi-minifi-cpp/extensions/http-curl/tests/HTTPHandlers.h:432: void HeartbeatHandler::verify(mg_connection*): Assertion `ok' failed.
> > ```
> 
> I could not reproduce the issue locally after a few hundred runs with Release and Debug builds. I checked the failed assertion and it is a rapidjson parse failure of either an acknowledge or a heartbeat message sent by the Minifi process. It may have been a corrupted message due to some environment related issue, I'm not sure of the root cause, but if it resurfaces we could open an new issue to the transient failures epic.

The issue seems to be reappearing, needs further investigation.

lordgamez:
C2PauseResumeTests should be okay now. After reproducing the issue and applying the change of the last commit, 1000 exections were run without failures on both ubuntu 20.04 and 16.04 distros. The last failure on windows-vs2017 is a know issue [MINIFICPP-1438](https://issues.apache.org/jira/browse/MINIFICPP-1438)

PR #975, opened 2021-01-12
MINIFICPP-1400 Create ListS3 processor

lordgamez:
Jira ticket: https://issues.apache.org/jira/browse/MINIFICPP-1400

---------------------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/aws/processors/ListS3.cpp
```diff
@@ -0,0 +1,296 @@
...
+uint64_t ListS3::getLatestListedKeyTimestamp(const std::unordered_map<std::string, std::string> &state) {
+  std::string stored_listed_key_timestamp_str;
+  auto it = state.find(LATEST_LISTED_KEY_TIMESTAMP);
+  if (it != state.end()) {
+    stored_listed_key_timestamp_str = it->second;
+  }
+
+  int64_t stored_listed_key_timestamp = 0;
+  if (!stored_listed_key_timestamp_str.empty()) {
+    core::Property::StringToInt(stored_listed_key_timestamp_str, stored_listed_key_timestamp);
```
adamdebreceni, line 192: it seems like `StringToInt` properly handles the empty string so this check is not necessary

extensions/aws/processors/ListS3.cpp
```diff
@@ -0,0 +1,296 @@
...
+  }
+
+  return stored_listed_key_timestamp;
+}
+
+std::tuple<int64_t, std::vector<std::string>> ListS3::getCurrentState(const std::shared_ptr<core::ProcessContext> &context) {
```
adamdebreceni, line 198: do you think it would be worth creating a `State` struct to replace the `tuple`?

extensions/aws/processors/ListS3.cpp
```diff
@@ -0,0 +1,296 @@
...
+    for (const auto& tag : get_object_tags_result.value()) {
+      session->putAttribute(flow_file, "s3.tag." + tag.first, tag.second);
+    }
+  } else {
+    logger_->log_warn("Failed to get object tags for object %s in bucket %s", object_attributes.filename, bucket);
+  }
+}
+
+void ListS3::writeUserMetadata(
+    const std::string &bucket,
```
adamdebreceni, line 150: `bucket` parameter is unused

extensions/aws/processors/ListS3.cpp
adamdebreceni: we could even move `was_object_listed_already` functionality into that struct (and the update logic in `onTrigger` as something like `registerObject`)

extensions/aws/s3/S3Wrapper.cpp
```diff
@@ -30,46 +37,253 @@ namespace minifi {
...
+  auto request = createListRequest<Aws::S3::Model::ListObjectVersionsRequest>(params);
+  std::vector<ListedObjectAttributes> attribute_list;
+  nonstd::optional_lite::optional<Aws::S3::Model::ListObjectVersionsResult> aws_result;
+  do {
+    aws_result = request_sender_->sendListVersionsRequest(request);
+    if (!aws_result) {
+      return minifi::utils::nullopt;
+    }
+    const auto& versions = aws_result->GetVersions();
+    logger_->log_debug("AWS S3 List operation returned %d versions. This result is truncated: %s", versions.size(), aws_result->GetIsTruncated() ? "true" : "false");
```
adamdebreceni, line 223: I'm not sure if printing (probably) `size_t` with the specifier `%d` is safe

extensions/aws/s3/S3Wrapper.h
```diff
@@ -31,11 +47,190 @@ namespace minifi {
...
-  minifi::utils::optional<Aws::S3::Model::GetObjectResult> sendGetObjectRequest(const Aws::S3::Model::GetObjectRequest& request) override;
+  void setCannedAcl(Aws::S3::Model::PutObjectRequest& request, const std::string& canned_acl) const;
+  int64_t writeFetchedBody(Aws::IOStream& source, const int64_t data_size, const std::shared_ptr<io::BaseStream>& output);
+  Expiration getExpiration(const std::string& expiration);
+  static std::string getEncryptionString(Aws::S3::Model::ServerSideEncryption encryption);
+
+  minifi::utils::optional<std::vector<ListedObjectAttributes>> listVersions(const ListRequestParameters& params);
+  minifi::utils::optional<std::vector<ListedObjectAttributes>> listObjects(const ListRequestParameters& params);
+  void addListResults(const Aws::Vector<Aws::S3::Model::ObjectVersion>& content, const uint64_t min_object_age, std::vector<ListedObjectAttributes>& listed_objects);
+  void addListResults(const Aws::Vector<Aws::S3::Model::Object>& content, const uint64_t min_object_age, std::vector<ListedObjectAttributes>& listed_objects);
```
adamdebreceni, line 190: the top level `const` on the parameters only has an effect in the definition

extensions/aws/s3/S3Wrapper.cpp
lordgamez: Fixed in [d1af764](https://github.com/apache/nifi-minifi-cpp/pull/975/commits/d1af764a93439b099a8bee23f6d397772eefc0ff)

extensions/aws/processors/ListS3.cpp
lordgamez: Removed in [d1af764](https://github.com/apache/nifi-minifi-cpp/pull/975/commits/d1af764a93439b099a8bee23f6d397772eefc0ff)

extensions/aws/processors/ListS3.cpp
lordgamez: Good idea! Added in [d1af764](https://github.com/apache/nifi-minifi-cpp/pull/975/commits/d1af764a93439b099a8bee23f6d397772eefc0ff)

extensions/aws/processors/ListS3.cpp
lordgamez: Fixed in [d1af764](https://github.com/apache/nifi-minifi-cpp/pull/975/commits/d1af764a93439b099a8bee23f6d397772eefc0ff)

extensions/aws/s3/S3Wrapper.h
lordgamez: Could you elaborate on this please? I'm not sure I understand the reason behind this comment.

adamdebreceni: the `const` in the parameter `const uint64_t min_object_age` is an "implementation detail" https://abseil.io/tips/109

lordgamez: Thanks for the link, I never thought about it from this point of view, but it makes sense to follow the mentioned guideline.

lordgamez: Updated in [4507f26](https://github.com/apache/nifi-minifi-cpp/pull/975/commits/4507f26caf3e1bec015f411fc1e924a08fa2c9a3)

extensions/aws/processors/ListS3.cpp
```diff
@@ -0,0 +1,294 @@
...
+  auto aws_results = s3_wrapper_.listBucket(list_request_params_);
+  if (!aws_results) {
+    logger_->log_error("Failed to list S3 bucket %s", list_request_params_.bucket);
+    context->yield();
+    return;
+  }
+
+  if (aws_results->size() == 0) {
+    logger_->log_debug("No S3 object were found in bucket %s", list_request_params_.bucket);
+    return;
```
adamdebreceni, line 252: CWEL yields when there are no events, GetFile does not, should we yield here?

extensions/aws/s3/S3Wrapper.cpp
```diff
@@ -30,46 +37,253 @@ namespace minifi {
...
+  buffer.reserve(BUFFER_SIZE);
 
-  if (outcome.IsSuccess()) {
-      logger_->log_info("Added S3 object '%s' to bucket '%s'", request.GetKey(), request.GetBucket());
-      return outcome.GetResultWithOwnership();
-  } else {
-    logger_->log_error("PutS3Object failed with the following: '%s'", outcome.GetError().GetMessage());
+  int64_t write_size = 0;
+  while (write_size < data_size) {
+    auto next_write_size = data_size - write_size < BUFFER_SIZE ? data_size - write_size : BUFFER_SIZE;
```
adamdebreceni, line 152: this looks like `std::min`

extensions/aws/s3/S3Wrapper.cpp
```diff
@@ -30,46 +37,253 @@ namespace minifi {
...
+    if (!source.read(reinterpret_cast<char*>(buffer.data()), next_write_size)) {
```
adamdebreceni, line 153: could the `Aws::IOStream` stream contain more than the body? if it only contains the body, we could use the `readsome` method and read until eof, so we wouldn't need the `data_size` parameter

extensions/aws/s3/S3Wrapper.cpp
```diff
@@ -30,46 +37,253 @@ namespace minifi {
...
+  auto request = createListRequest<Aws::S3::Model::ListObjectsV2Request>(params);
+  std::vector<ListedObjectAttributes> attribute_list;
+  nonstd::optional_lite::optional<Aws::S3::Model::ListObjectsV2Result> aws_result;
+  do {
+    aws_result = request_sender_->sendListObjectsRequest(request);
+    if (!aws_result) {
+      return minifi::utils::nullopt;
+    }
+    const auto& objects = aws_result->GetContents();
+    logger_->log_debug("AWS S3 List operation returned %d objects. This result is truncated: %s", objects.size(), aws_result->GetIsTruncated() ? "true" : "false");
```
adamdebreceni, line 244: the `%d`  might not be the proper format specifier

extensions/aws/s3/S3Wrapper.h
```diff
@@ -31,11 +47,190 @@ namespace minifi {
...
+};
+
+struct ListedObjectAttributes {
+  std::string filename;
+  std::string etag;
+  bool is_latest = false;
+  int64_t last_modified = 0;
+  int length = 0;
+  std::string store_class;
+  std::string version;
```
adamdebreceni, line 156: it seems like `version` is an optional field and we treat the empty string as the `null`, now there are many examples of this throughout our codebase so it is not unheard of, but do you think using `optional` to make this explicit could be beneficial?

extensions/aws/processors/ListS3.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+void ListS3::onSchedule(const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSessionFactory> &sessionFactory) {
+  S3Processor::onSchedule(context, sessionFactory);
+
+  state_manager_ = context->getStateManager();
+  if (state_manager_ == nullptr) {
+    throw Exception(PROCESSOR_EXCEPTION, "Failed to get StateManager");
+  }
+
+  auto common_properties = getCommonELSupportedProperties(context, nullptr);
```
adamdebreceni, line 98: passing `nullptr` will call a possible expression script with `nullptr`, will that properly throw an exception? e.g. it seems like `Bucket` property supports expression language, what happens if the user accidentally supplies a script value for that property that tries to access the null flowFile?

extensions/aws/processors/ListS3.cpp
lordgamez: I'm not sure, even though we could yield if we did not get any data just to wait some more time for additional objects, we cannot be sure, that the next schedule time will not provide new data.

extensions/aws/s3/S3Wrapper.h
lordgamez: I don't think in this case we would have a benefit for this field to be optional as an empty version is not a valid value, so we cannot really distinguish an empty version from an unset version, they should mean the same. I think here it is easier just to check for an empty string if we want to check if it has been set.

extensions/aws/processors/ListS3.cpp
lordgamez: As far as I see calling the getProperty with nullptr a minifi::expression::Parameters is created with a weak_ptr from the passed flow_file (where the default value is also nullptr) before evaluating the expression. Every time the the flow_file is used it tries to lock the weak_ptr and only uses it when the lock succeeds, so it should be safe.

extensions/aws/s3/S3Wrapper.cpp
lordgamez: Fixed in [27ff518](https://github.com/apache/nifi-minifi-cpp/pull/975/commits/27ff518407185f7876d96dfac2566bfae9eda62b)

extensions/aws/s3/S3Wrapper.cpp
lordgamez: Fixed in [27ff518](https://github.com/apache/nifi-minifi-cpp/pull/975/commits/27ff518407185f7876d96dfac2566bfae9eda62b)

extensions/aws/s3/S3Wrapper.cpp
lordgamez: Somehow I get segmentation fault when I use `readsome` and `eof` check combination here. I'm not sure of the root cause, I'll try to investigate it next week.

extensions/aws/processors/ListS3.cpp
lordgamez: I checked the Nifi implementation for reference and it seems it yields in case there is no new S3 object to list. I'll update the implementation accordingly.

lordgamez: Updated in [8216738](https://github.com/apache/nifi-minifi-cpp/pull/975/commits/8216738921e0d5368e532b09aaf618fc61bcbaaf)

extensions/aws/s3/S3Wrapper.cpp
lordgamez: Even though I managed to make the `readsome` method work in the unit tests, it seems to fail in the docker tests. While reading from the stream, no data is returned and the callback's `process` method fails. The definition of `readsome` says that the function "Extracts up to count **immediately available** characters from the input stream." and notes that "The behavior of this function is **highly implementation-specific**." which seems to be unreliable. Some have noted in other words that the function only returns characters that are immediately available in the underlying buffer. Because of this I would prefer staying with the use of the `read` function in this case.

adamdebreceni: I see now that `readsome` would not be a wise choice here, I just wanted to avoid passing in the body length, if we read until `eof` anyway, but I see no indication that the two (reading until body length and reading until `eof`) are equivalent

extensions/aws/processors/ListS3.cpp
```diff
@@ -0,0 +1,295 @@
...
+    }
+  } else {
+    logger_->log_warn("Failed to get object metadata for object %s in bucket %s", params.object_key, params.bucket);
+  }
+}
+
+std::vector<std::string> ListS3::getLatestListedKeys(const std::unordered_map<std::string, std::string> &state) {
+  std::vector<std::string> latest_listed_keys;
+  for (const auto& kvp : state) {
+    if (kvp.first.rfind(LATEST_LISTED_KEY_PREFIX, 0) == 0) {
```
adamdebreceni, line 175: `LATEST_LISTED_KEY_PREFIX` being a prefix of `LATEST_LISTED_KEY_TIMESTAMP` results in a subtle bug where we append the timestamp to the `latest_listed_keys` continuously growing, we should change either the prefix or the timestamp key to avoid this

extensions/aws/processors/ListS3.cpp
```diff
@@ -0,0 +1,295 @@
...
+  writeObjectTags(list_request_params_.bucket, object_attributes, session, flow_file);
+  writeUserMetadata(object_attributes, session, flow_file);
+
+  session->transfer(flow_file, Success);
+}
+
+void ListS3::onTrigger(const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSession> &session) {
+  logger_->log_debug("ListS3 onTrigger");
+
+  auto aws_results = s3_wrapper_.listBucket(list_request_params_);
```
adamdebreceni, line 243: would it be possible to use the previous `onTrigger`s timestamp to update the minimum object age and avoid refetching the whole object list each time?

extensions/aws/processors/ListS3.cpp
lordgamez: Thank, great catch! Fixed in [72b5ba7](https://github.com/apache/nifi-minifi-cpp/pull/975/commits/72b5ba7a66b05f7f03580900b5e1fac2d086b6e5)

extensions/aws/processors/ListS3.cpp
lordgamez: Unfortunately the AWS SDK client does not provide the minimum object age as a parameter so we still have to do the filtering by hand either in the `onTrigger` or in the `S3Wrapper`, so all of the objects will be retrieved at all times. Even if we would pass the new timestamp to the `S3Wrapper` the objects still need to be filtered in the `onTrigger` as well to see if there were additional objects fetched with that last timestamp since the previous listing. I would not mix these two filters, I think it's a bit more easier to follow the logic if we have a separate filter with the user given minimum object age in the `S3Wrapper`, and have another filter in the `onTrigger` with the parameters retrieved from our last stored state.

adamdebreceni: thank you, makes sense

extensions/aws/processors/ListS3.cpp
```diff
@@ -0,0 +1,295 @@
...
+
+  aws::s3::HeadObjectRequestParameters params;
+  params.bucket = list_request_params_.bucket;
+  params.object_key = object_attributes.filename;
+  params.version = object_attributes.version;
+  params.requester_pays = requester_pays_;
+  auto head_object_tags_result = s3_wrapper_.headObject(params);
+  if (head_object_tags_result) {
+    for (const auto& metadata : head_object_tags_result->user_metadata_map) {
+      session->putAttribute(flow_file, "s3.user.metadata." + metadata.first, metadata.second);
```
adamdebreceni, line 165: the head result also contains other metadata (e.g. mime type) but only the user defined metadata are forwarded onto the flowFile, is this intentional?

lordgamez: Yes, ListS3 written attributes only contain the user metadata in [NiFi](https://nifi.apache.org/docs/nifi-docs/components/org.apache.nifi/nifi-aws-nar/1.12.1/org.apache.nifi.processors.aws.s3.ListS3/index.html) as well. It is usually intended to be used together with FetchS3Object processor which is responsible for getting the mime type.

adamdebreceni: I see, thank you

adamdebreceni (approved):

extensions/aws/processors/DeleteS3Object.cpp
```diff
@@ -30,6 +30,11 @@ namespace minifi {
 namespace aws {
 namespace processors {
 
+const core::Property DeleteS3Object::ObjectKey(
+  core::PropertyBuilder::createProperty("Object Key")
+    ->withDescription("The key of the S3 object. If none is given the filename attribute will be used by default.")
+    ->supportsExpressionLanguage(true)
+    ->build());
```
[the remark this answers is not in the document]
lordgamez: Are you referring to the `${filename}` attribute value? In the PutS3Object processor we removed that default value on purpose and used this implementation to support users who do not have expression language support. This is why I used the same method here.

extensions/aws/s3/S3ClientRequestSender.h
```diff
@@ -0,0 +1,47 @@
...
+class S3ClientRequestSender : public S3RequestSender {
+ public:
+  minifi::utils::optional<Aws::S3::Model::PutObjectResult> sendPutObjectRequest(const Aws::S3::Model::PutObjectRequest& request) override;
+  bool sendDeleteObjectRequest(const Aws::S3::Model::DeleteObjectRequest& request) override;
+  minifi::utils::optional<Aws::S3::Model::GetObjectResult> sendGetObjectRequest(const Aws::S3::Model::GetObjectRequest& request) override;
+  minifi::utils::optional<Aws::S3::Model::ListObjectsV2Result> sendListObjectsRequest(const Aws::S3::Model::ListObjectsV2Request& request) override;
+  minifi::utils::optional<Aws::S3::Model::ListObjectVersionsResult> sendListVersionsRequest(const Aws::S3::Model::ListObjectVersionsRequest& request) override;
+  minifi::utils::optional<Aws::S3::Model::GetObjectTaggingResult> sendGetObjectTaggingRequest(const Aws::S3::Model::GetObjectTaggingRequest& request) override;
+  minifi::utils::optional<Aws::S3::Model::HeadObjectResult> sendHeadObjectRequest(const Aws::S3::Model::HeadObjectRequest& request) override;
+};
```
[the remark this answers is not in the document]
lordgamez: The ListS3 processor is supposed to be the last S3 processor for the time being, I'm not sure if we have any additional S3 functionality in the future. Do you have any particular use case in mind for the async API? In that case I suppose we should have the same async API in the S3Wrapper as well, to be able to get the processor-desired result in an async way. In my opinion we should only invest in implementing and testing an async API when we have a use case for it, I'm not sure if it would be worth right now.

extensions/aws/processors/DeleteS3Object.cpp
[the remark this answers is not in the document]
szaszm: I didn't realize it was an attribute, sorry.

extensions/aws/s3/S3ClientRequestSender.h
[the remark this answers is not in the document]
szaszm: Async operations generally scale better, that's why I was raising this. The functionality and use cases are the same, but it would probably result in 1-2 orders of magnitude better performance. Now we have to wait for a request to finish before we can proceed. We could pipeline multiple requests with an async API.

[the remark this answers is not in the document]
lordgamez: I think the async API could help us in a way that currently we always wait for the request to finish because of the `s3_wrapper_mutex_` which helps to avoid overwriting client configuration of the S3 client in concurrent `onTrigger` calls. Because of the mutex only 1 request can run at a time, and you are right that making the client async, waiting for the result could be done afterwards. Another solution I was thinking about is making the S3Wrapper stateless and passing all the configuration parameters to every single call, making it thread safe to call `onTrigger` function parallel which would create separate clients with separate configurations. What do you think?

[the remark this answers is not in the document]
szaszm: Regarding S3 client, if it's cheap to create, then I would go for solution 2, otherwise solution 1. But you're more familiar with the ecosystem, so I trust your judgement.

In any case, an easy future-proofing (pun intended) could be to wrap the current solution in `async(deferred, ...)` and convert to actual async later. After giving this a second thought, async conversion would probably require other changes as well, so this may not be a good idea after all.

It's your call whether you see any value in any of the ideas raised. I'm fine with no changes as well.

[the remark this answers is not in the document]
lordgamez: I think it's a good and idea to improve upon this and it should be investigated further or rather experimented with the available solutions. As it impacts all the S3 API and the callbacks in each S3 processor and test and therefore all S3 processors themselves, I would move this to a followup PR so I created a separate Jira ticket for this: https://issues.apache.org/jira/browse/MINIFICPP-1476

lordgamez:
After rebase moved ListS3 tests to the Python Behave framework. This implies some additional changes that should be re-checked.

extensions/aws/s3/S3Wrapper.cpp
```diff
@@ -30,46 +37,253 @@ namespace minifi {
...
+  return Expiration{results[1], results[2]};
+}
+
+std::string S3Wrapper::getEncryptionString(Aws::S3::Model::ServerSideEncryption encryption) {
+  if (encryption == Aws::S3::Model::ServerSideEncryption::NOT_SET) {
+    return "";
+  }
+
+  auto it = std::find_if(SERVER_SIDE_ENCRYPTION_MAP.begin(), SERVER_SIDE_ENCRYPTION_MAP.end(),
+    [&](const std::pair<std::string, const Aws::S3::Model::ServerSideEncryption&> pair) {
```
szaszm, line 94: Did you mean to take a const ref to a pair of two values (i.e. swap `&` and `>`)?

extensions/aws/s3/S3Wrapper.cpp
```diff
@@ -30,46 +37,253 @@ namespace minifi {
...
-  auto outcome = s3_client.PutObject(request);
+void HeadObjectResult::setFilePaths(const std::string& key) {
+  absolute_path = key;
+  std::tie(path, filename) = minifi::utils::file::FileUtils::split_path(key, true /*force_posix*/);
+}
+
+S3Wrapper::S3Wrapper() : request_sender_(minifi::utils::make_unique<S3ClientRequestSender>()) {
+}
+
+S3Wrapper::S3Wrapper(std::unique_ptr<S3RequestSender> request_sender) : request_sender_(std::move(request_sender)) {
```
szaszm, line 48: Since `unique_ptr` is move-only and we're just storing it, we can just take an rvalue reference for the same effect with one less move operation, assuming no elision.

extensions/aws/s3/S3Wrapper.cpp
```diff
@@ -30,46 +37,253 @@ namespace minifi {
...
+  Aws::S3::Model::DeleteObjectRequest request;
+  request.SetBucket(bucket);
+  request.SetKey(object_key);
+  if (!version.empty()) {
+    request.SetVersionId(version);
+  }
+  return request_sender_->sendDeleteObjectRequest(request);
+}
+
+int64_t S3Wrapper::writeFetchedBody(Aws::IOStream& source, const int64_t data_size, const std::shared_ptr<io::BaseStream>& output) {
```
szaszm, line 145: This could be static. I would take the stream by reference.

extensions/aws/s3/S3Wrapper.cpp
```diff
@@ -30,46 +37,253 @@ namespace minifi {
...
+  static const int64_t BUFFER_SIZE = 4096;
+  std::vector<uint8_t> buffer;
+  buffer.reserve(BUFFER_SIZE);
```
szaszm, line 148: This should be `resize`, not `reserve`, otherwise writing to it is undefined behavior (even though it usually works).

extensions/aws/s3/S3Wrapper.cpp
```diff
@@ -30,46 +37,253 @@ namespace minifi {
...
+      logger_->log_debug("Object version '%s' of key '%s' skipped due to minimum object age filter", version.GetVersionId(), version.GetKey());
+      continue;
+    }
+
+    ListedObjectAttributes attributes;
+    attributes.etag = minifi::utils::StringUtils::removeFramingCharacters(version.GetETag(), '"');
+    attributes.filename = version.GetKey();
+    attributes.is_latest = version.GetIsLatest();
+    attributes.last_modified = version.GetLastModified().Millis();
+    attributes.length = version.GetSize();
```
szaszm, line 188: This is a narrowing conversion. `length` should be wide enough to properly store the size.

extensions/aws/s3/S3Wrapper.cpp
```diff
@@ -30,46 +37,253 @@ namespace minifi {
...
+  auto request = createListRequest<Aws::S3::Model::ListObjectVersionsRequest>(params);
+  std::vector<ListedObjectAttributes> attribute_list;
+  nonstd::optional_lite::optional<Aws::S3::Model::ListObjectVersionsResult> aws_result;
+  do {
+    aws_result = request_sender_->sendListVersionsRequest(request);
+    if (!aws_result) {
+      return minifi::utils::nullopt;
+    }
+    const auto& versions = aws_result->GetVersions();
+    logger_->log_debug("AWS S3 List operation returned %zu versions. This result is truncated: %s", versions.size(), aws_result->GetIsTruncated() ? "true" : "false");
```
szaszm, line 223:
```suggestion
    logger_->log_debug("AWS S3 List operation returned %zu versions. This result is%s truncated.", versions.size(), aws_result->GetIsTruncated() ? "" : " not");
```

extensions/aws/processors/ListS3.h
```diff
@@ -0,0 +1,111 @@
...
+  // Supported Relationships
+  static const core::Relationship Success;
+
+  explicit ListS3(std::string name, minifi::utils::Identifier uuid = minifi::utils::Identifier())
+    : S3Processor(name, uuid, logging::LoggerFactory<ListS3>::getLogger()) {
+  }
+
+  explicit ListS3(std::string name, minifi::utils::Identifier uuid, std::unique_ptr<aws::s3::S3RequestSender> s3_request_sender)
+    : S3Processor(std::move(name), uuid, logging::LoggerFactory<ListS3>::getLogger(), std::move(s3_request_sender)) {
+  }
```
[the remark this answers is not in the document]
lordgamez: Updated in all S3 processors in 65dde7c661bfac32918522999752e6310a5becf7

docker/test/integration/steps/steps.py
```diff
@@ -227,10 +238,20 @@ def step_impl(context, cluster_name):
...
+def step_impl(context, content, file_name, path, seconds):
+    time.sleep(seconds)
+    context.test.add_test_data(path, content, file_name)
+
 @then("a flowfile with the content \"{content}\" is placed in the monitored directory in less than {duration}")
 def step_impl(context, content, duration):
     context.test.check_for_file_with_content_generated(content, timeparse(duration))
 
+@then("{number_of_files:d} flowfiles are placed in the monitored directory in less than {duration}")
+@then("{number_of_files:d} flowfile is placed in the monitored directory in less than {duration}")
```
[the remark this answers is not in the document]
lordgamez: Updated in 623540cadc8eae8f3c1de4439af140f23d8ba5bf

extensions/aws/processors/ListS3.cpp
```diff
@@ -0,0 +1,295 @@
...
+    }
+  } else {
+    logger_->log_warn("Failed to get object metadata for object %s in bucket %s", params.object_key, params.bucket);
+  }
+}
+
+std::vector<std::string> ListS3::getLatestListedKeys(const std::unordered_map<std::string, std::string> &state) {
```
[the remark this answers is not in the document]
lordgamez: Updated in 0c24748c92b8fdd6c4ec1070e86a30916e4db885

extensions/aws/processors/ListS3.h
```diff
@@ -0,0 +1,111 @@
...
+    const std::shared_ptr<core::ProcessSession> &session,
+    const std::shared_ptr<core::FlowFile> &flow_file);
+  void writeUserMetadata(
+    const aws::s3::ListedObjectAttributes &object_attributes,
+    const std::shared_ptr<core::ProcessSession> &session,
+    const std::shared_ptr<core::FlowFile> &flow_file);
+  std::vector<std::string> getLatestListedKeys(const std::unordered_map<std::string, std::string> &state);
+  uint64_t getLatestListedKeyTimestamp(const std::unordered_map<std::string, std::string> &state);
+  ListingState getCurrentState(const std::shared_ptr<core::ProcessContext> &context);
+  void storeState(const std::shared_ptr<core::ProcessContext> &context, const ListingState &latest_listing_state);
```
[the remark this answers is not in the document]
lordgamez: Updated in 0c24748c92b8fdd6c4ec1070e86a30916e4db885

extensions/aws/processors/ListS3.h
```diff
@@ -0,0 +1,111 @@
...
+  void createNewFlowFile(
+    const std::shared_ptr<core::ProcessSession> &session,
+    const aws::s3::ListedObjectAttributes &object_attributes);
```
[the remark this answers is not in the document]
lordgamez: Updated in 0c24748c92b8fdd6c4ec1070e86a30916e4db885

extensions/aws/s3/S3Wrapper.cpp
```diff
@@ -30,46 +37,253 @@ namespace minifi {
...
+}
+
+Expiration S3Wrapper::getExpiration(const std::string& expiration) {
+  minifi::utils::Regex expr("expiry-date=\"(.*)\", rule-id=\"(.*)\"");
+  const auto match = expr.match(expiration);
+  const auto& results = expr.getResult();
+  if (!match || results.size() < 3)
+    return Expiration{};
+  return Expiration{results[1], results[2]};
+}
```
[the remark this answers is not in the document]
lordgamez: Updated in 0c24748c92b8fdd6c4ec1070e86a30916e4db885

extensions/aws/s3/S3Wrapper.h
```diff
@@ -31,11 +47,190 @@ namespace minifi {
 namespace aws {
 namespace s3 {
 
-class S3Wrapper : public S3WrapperBase {
+static const std::unordered_map<std::string, Aws::S3::Model::StorageClass> STORAGE_CLASS_MAP {
```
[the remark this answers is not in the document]
lordgamez: These maps are used to initialize some static class members in PutS3Objects. If `static` is removed, because of the order of the initialization those members will not be initialized.

extensions/aws/s3/S3Wrapper.h
```diff
@@ -31,11 +47,190 @@ namespace minifi {
...
+  Expiration getExpiration(const std::string& expiration);
+  static std::string getEncryptionString(Aws::S3::Model::ServerSideEncryption encryption);
+
+  minifi::utils::optional<std::vector<ListedObjectAttributes>> listVersions(const ListRequestParameters& params);
+  minifi::utils::optional<std::vector<ListedObjectAttributes>> listObjects(const ListRequestParameters& params);
+  void addListResults(const Aws::Vector<Aws::S3::Model::ObjectVersion>& content, uint64_t min_object_age, std::vector<ListedObjectAttributes>& listed_objects);
+  void addListResults(const Aws::Vector<Aws::S3::Model::Object>& content, uint64_t min_object_age, std::vector<ListedObjectAttributes>& listed_objects);
+
+  template<typename ListRequest>
+  ListRequest createListRequest(const ListRequestParameters& params) {
```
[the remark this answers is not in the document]
lordgamez: Updated in 6a44b4beb6b511546226d01655cad0d5c8688fc7

extensions/aws/utils/AWSInitializer.h
```diff
@@ -20,6 +20,7 @@
 #pragma once
 
 #include "aws/core/Aws.h"
+#include "aws/core/utils/xml/XmlSerializer.h"
```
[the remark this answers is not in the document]
lordgamez: Moved to `S3RequestSender` header in c2539fcb0e56fb00750f2c3b4dd657f34835f8ec. The XMLDocument class is needed to instantiate the S3 Request classes and it is not implicitly included.

extensions/aws/s3/S3Wrapper.cpp
lordgamez: Nice catch, thanks, it was a typo, fixed in ee41601f1811fbc6bc575bdae25d450709d29f50

extensions/aws/s3/S3Wrapper.cpp
lordgamez: I am not sure how would you go about it, could you please elaborate? How would you initialize the `unique_ptr` with a `S3RequestSender&&`?

extensions/aws/s3/S3Wrapper.cpp
lordgamez: Updated in 322c3f0d787b1fbd68518f572337951d466bdf3a

extensions/aws/s3/S3Wrapper.cpp
lordgamez: Updated in 322c3f0d787b1fbd68518f572337951d466bdf3a I think the change is also present in the PutAzureBlobStorage PR

extensions/aws/s3/S3Wrapper.cpp
lordgamez: Good catch, updated in 322c3f0d787b1fbd68518f572337951d466bdf3a

extensions/aws/s3/S3Wrapper.cpp
lordgamez: Updated in 322c3f0d787b1fbd68518f572337951d466bdf3a

extensions/aws/s3/S3Wrapper.cpp
szaszm:
```suggestion
S3Wrapper::S3Wrapper(std::unique_ptr<S3RequestSender>&& request_sender) : request_sender_(std::move(request_sender)) {
```

extensions/aws/processors/ListS3.cpp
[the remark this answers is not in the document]
szaszm: Please update `getLatestListedKeyTimestamp` too

extensions/aws/s3/S3Wrapper.cpp
```diff
@@ -30,46 +37,291 @@ namespace minifi {
...
-  auto outcome = s3_client.GetObject(request);
+minifi::utils::optional<GetObjectResult> S3Wrapper::getObject(const GetObjectRequestParameters& get_object_params, const std::shared_ptr<io::BaseStream>& out_body) {
+  auto request = createFetchObjectRequest<Aws::S3::Model::GetObjectRequest>(get_object_params);
+  auto aws_result = request_sender_->sendGetObjectRequest(request);
+  if (!aws_result) {
+    return minifi::utils::nullopt;
+  }
+  auto result = fillFetchObjectResult<Aws::S3::Model::GetObjectResult, GetObjectResult>(get_object_params, aws_result.value());
+  result.write_size = writeFetchedBody(aws_result->GetBody(), aws_result->GetContentLength(), *out_body);
+  return result;
```
szaszm, line 173: I'd convert this to take a reference, too.
```suggestion
minifi::utils::optional<GetObjectResult> S3Wrapper::getObject(const GetObjectRequestParameters& get_object_params, io::BaseStream& out_body) {
  auto request = createFetchObjectRequest<Aws::S3::Model::GetObjectRequest>(get_object_params);
  auto aws_result = request_sender_->sendGetObjectRequest(request);
  if (!aws_result) {
    return minifi::utils::nullopt;
  }
  auto result = fillFetchObjectResult<Aws::S3::Model::GetObjectResult, GetObjectResult>(get_object_params, *aws_result);
  result.write_size = writeFetchedBody(aws_result->GetBody(), aws_result->GetContentLength(), out_body);
  return result;
```

extensions/aws/s3/S3Wrapper.cpp
szaszm: Could you update `listObjects` in the same manner?

extensions/aws/s3/S3Wrapper.cpp
lordgamez: Updated in 4001582e1efc9d5c697854e8b4477eb393d6e2f3

extensions/aws/s3/S3Wrapper.cpp
lordgamez: Thanks, sorry I misinterpreted your comment at first, updated in 4001582e1efc9d5c697854e8b4477eb393d6e2f3

extensions/aws/processors/ListS3.cpp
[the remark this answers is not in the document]
lordgamez: Updated in 4001582e1efc9d5c697854e8b4477eb393d6e2f3

szaszm (approved):

PR #979, opened 2021-01-26
MINIFICPP-1456 Introduce PutAzureBlobStorage processor

lordgamez:
Introducing PutAzureBlobStorage processor using [Azure SDK for CPP](https://github.com/Azure/azure-sdk-for-cpp) with similar functionality of the [NiFi equivalent](https://nifi.apache.org/docs/nifi-docs/components/org.apache.nifi/nifi-azure-nar/1.12.1/org.apache.nifi.processors.azure.storage.PutAzureBlobStorage/index.html). Jira ticket: https://issues.apache.org/jira/browse/MINIFICPP-1456

-------------------------------------------------------------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


cmake/BundledAzureSdkCpp.cmake
```diff
@@ -0,0 +1,126 @@
...
+    set_target_properties(AZURE::azure-core PROPERTIES IMPORTED_LOCATION "${AZURE_CORE_LIB}")
+    add_dependencies(AZURE::azure-core azure-sdk-cpp-external-build)
+    set_property(TARGET AZURE::azure-core APPEND PROPERTY INTERFACE_INCLUDE_DIRECTORIES ${LIBAZURE_INCLUDE_DIRS})
+    set_property(TARGET AZURE::azure-core APPEND PROPERTY INTERFACE_LINK_LIBRARIES LibXml2::LibXml2 CURL::libcurl OpenSSL::Crypto OpenSSL::SSL Threads::Threads nlohmann_json::nlohmann_json)
+    if (APPLE)
+        set_property(TARGET AZURE::azure-core APPEND PROPERTY INTERFACE_LINK_LIBRARIES "-framework CoreFoundation")
+    endif()
+    if (WIN32)
+        set_property(TARGET AZURE::azure-core APPEND PROPERTY INTERFACE_LINK_LIBRARIES winhttp.lib)
+    endif()
```
szaszm, line 88: Instead of setting properties directly, we should prefer using target commands to set them indirectly. This way is shorter, looks better and is more commonly found in examples and docs.
```suggestion
    target_include_directories(AZURE::azure-core INTERFACE ${LIBAZURE_INCLUDE_DIRS})
    target_link_libraries(AZURE::azure-core INTERFACE LibXml2::LibXml2 CURL::libcurl OpenSSL::Crypto OpenSSL::SSL Threads::Threads nlohmann_json::nlohmann_json)
    if (APPLE)
        target_link_libraries(AZURE::azure-core INTERFACE "-framework CoreFoundation")
    endif()
    if (WIN32)
        target_link_libraries(AZURE::azure-core INTERFACE winhttp.lib)
    endif()
```

extensions/azure/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+include(${CMAKE_SOURCE_DIR}/extensions/ExtensionHeader.txt)
+
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+include_directories(controllerservices processors storage ${CMAKE_SOURCE_DIR}/extensions/azure)
```
szaszm, line 24: Avoid global commands like `include_directories` and `link_libraries`. Use target equivalents like `target_include_directories` and `target_link_libraries`.
For example, you can set the required standard with `target_compile_features(minifi-azure PUBLIC cxx_std_14)`

NOTICE
```diff
@@ -55,6 +58,7 @@ This software includes third party software subject to the following copyrights:
 - libsodium - Copyright (c) 2013 - 2018 Frank Denis under the ISC software license
 - IANA timezone database - public domain
 - date (HowardHinnant/date) - notices below
+- JSON for Modern C++ (nlohmann/json) - Copyright (c) 2013-2021 Niels Lohmann
```
szaszm, line 61: Is this a hard dependency of the azure sdk? Because we already have `rapidjson` for json handling, so if it's possible to avoid bringing in another lib for the same purpose, I would try.

LICENSE
```diff
@@ -2958,3 +2958,53 @@ OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
...
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
```
szaszm, line 3010: `msi/LICENSE.txt` should be a copy of this file. I've just submitted a pull request #983 to synchronize it, because it is out of date. Please consider updating it in this PR as well, so that the merge conflict will remind us to check both while merging.

NOTICE
lordgamez: In the latest version it has become a new dependency for azure sdk and does not compile without it. We may investigate if it's viable to patch and replace it with rapidjson in the sdk, although it may be cumbersome to maintain it in the future.

szaszm: Ok, don't bother

cmake/BundledAzureSdkCpp.cmake
lordgamez: Updated in [3dba6a8](https://github.com/apache/nifi-minifi-cpp/pull/979/commits/3dba6a8440d962cd546a7975e4fa9b21386ab6d0)

extensions/azure/CMakeLists.txt
lordgamez: Updated in [3dba6a8](https://github.com/apache/nifi-minifi-cpp/pull/979/commits/3dba6a8440d962cd546a7975e4fa9b21386ab6d0)

LICENSE
lordgamez: Updated the LICENSE.txt in [3dba6a8](https://github.com/apache/nifi-minifi-cpp/pull/979/commits/3dba6a8440d962cd546a7975e4fa9b21386ab6d0)

win_build_vs.bat
```diff
@@ -47,6 +47,7 @@ for %%x in (%*) do (
     if [%%~x] EQU [/M]           set installer_merge_modules=ON
     if [%%~x] EQU [/C]           set build_coap=ON
     if [%%~x] EQU [/A]           set build_AWS=ON
+    if [%%~x] EQU [/Z]           set build_azure=ON
```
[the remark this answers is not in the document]
lordgamez: Updated

extensions/azure/controllerservices/AzureStorageCredentialsService.h
```diff
@@ -0,0 +1,85 @@
...
+
+  explicit AzureStorageCredentialsService(const std::string &name, const std::shared_ptr<Configure> &configuration)
+      : ControllerService(name),
+        logger_(logging::LoggerFactory<AzureStorageCredentialsService>::getLogger()) {
+  }
+
+  void initialize() override;
+
+  void yield() override {
+  };
```
adamdebreceni, line 57: `;` accidental empty declaration?

extensions/azure/processors/PutAzureBlobStorage.cpp
```diff
@@ -0,0 +1,225 @@
...
+  azure::storage::AzureStorageCredentials credentials;
+  context->getProperty(StorageAccountName, credentials.storage_account_name, flow_file);
+  context->getProperty(StorageAccountKey, credentials.storage_account_key, flow_file);
+  context->getProperty(SASToken, credentials.sas_token, flow_file);
+  context->getProperty(CommonStorageAccountEndpointSuffix, credentials.endpoint_suffix, flow_file);
+  context->getProperty(ConnectionString, credentials.connection_string, flow_file);
+  return credentials.getConnectionString();
+}
+
+void PutAzureBlobStorage::createAzureStorageClient(const std::string &connection_string, const std::string &container_name) {
```
adamdebreceni, line 146: we could indicate with a comment or with a dummy lock parameter, that the `azure_storage_mutex_` should be acquired before calling this method

extensions/azure/processors/PutAzureBlobStorage.h
```diff
@@ -0,0 +1,138 @@
...
+    ReadCallback(uint64_t flow_size, azure::storage::BlobStorage& blob_storage_wrapper, const std::string &blob_name)
+      : flow_size_(flow_size)
+      , blob_storage_wrapper_(blob_storage_wrapper)
+      , blob_name_(blob_name) {
+    }
+
+    int64_t process(const std::shared_ptr<io::BaseStream>& stream) override {
+      std::vector<uint8_t> buffer;
+      buffer.reserve(flow_size_);
+      int read_ret = stream->read(buffer.data(), flow_size_);
```
adamdebreceni, line 85: note that `std::vector<T>::data` is only guaranteed to return a pointer such that `[data(), data() + size())` is a valid range, and for empty vectors it might return whatever, we should use `resize` [1](https://stackoverflow.com/questions/38972443/use-stdvectordata-after-reserve)

extensions/azure/storage/AzureBlobStorage.cpp
```diff
@@ -0,0 +1,83 @@
...
+    UploadBlobResult result;
+    result.length = buffer_size;
+    result.primary_uri = container_client_->GetUrl();
+    if (response->ETag.HasValue()) {
+      result.etag = response->ETag.ToString();
+    }
+    result.timestamp = response->LastModified.GetString(Azure::Core::DateTime::DateFormat::Rfc1123);
+    return result;
+  } catch (const std::runtime_error&) {
+    return utils::nullopt;
```
adamdebreceni, line 74: should we log here?

extensions/azure/processors/PutAzureBlobStorage.cpp
```diff
@@ -0,0 +1,225 @@
...
+  context->getProperty(CreateContainer.getName(), create_container_);
+}
+
+std::string PutAzureBlobStorage::getConnectionStringFromControllerService(const std::shared_ptr<core::ProcessContext> &context) const {
+  std::string service_name;
+  if (!context->getProperty(AzureStorageCredentialsService.getName(), service_name) || service_name.empty()) {
+    return "";
+  }
+
+  std::shared_ptr<core::controller::ControllerService> service = context->getControllerService(service_name);
```
adamdebreceni, line 121: it seems, here the user provided the `service_name` but found no such service, we should consider this an error, and log it accordingly, and a few lines below this, the service we acquired turns out not to be an instance of `AzureStorageCredentialsService` the fact of which we could log as well

extensions/azure/processors/PutAzureBlobStorage.h
adamdebreceni: this is also an issue in `S3WrapperBase.cpp:147`, `PutS3Object.h:94` and `PublishMQTT.h:82`

extensions/azure/controllerservices/AzureStorageCredentialsService.h
lordgamez: Updated in [8853da](https://github.com/apache/nifi-minifi-cpp/pull/979/commits/8853da7346c55a08a326609e97ad09675443e7fb)

extensions/azure/processors/PutAzureBlobStorage.cpp
lordgamez: Added comment in [8853da](https://github.com/apache/nifi-minifi-cpp/pull/979/commits/8853da7346c55a08a326609e97ad09675443e7fb)

extensions/azure/processors/PutAzureBlobStorage.h
lordgamez: Fixed in [8853da](https://github.com/apache/nifi-minifi-cpp/pull/979/commits/8853da7346c55a08a326609e97ad09675443e7fb). Used read overload with vector parameter where resize is used already, where it was possible.

extensions/azure/storage/AzureBlobStorage.cpp
lordgamez: Added logs in [8853da](https://github.com/apache/nifi-minifi-cpp/pull/979/commits/8853da7346c55a08a326609e97ad09675443e7fb)

extensions/azure/processors/PutAzureBlobStorage.cpp
lordgamez: Added error logs in [8853da](https://github.com/apache/nifi-minifi-cpp/pull/979/commits/8853da7346c55a08a326609e97ad09675443e7fb)

lordgamez:
The docker tests were moved to the new Python Behave framework after rebase. Some missing test assertions were added to the test framework and additional startup checks were implemented.

docker/test/integration/MiNiFi_integration_test_driver.py
```diff
@@ -110,6 +110,12 @@ def start(self):
                 startup_success = cluster.wait_for_app_logs("Starting Flow Controller...", 120)
             elif cluster.get_engine() == "kafka-broker":
                 startup_success = cluster.wait_for_app_logs("Startup complete.", 120)
+            elif cluster.get_engine() == "http-proxy":
```
hunyadi-dev, line 113: Thanks for finding what container logs to wait for, great addition.

libminifi/test/azure-tests/PutAzureBlobStorageTests.cpp
```diff
@@ -0,0 +1,288 @@
...
+      new minifi::azure::processors::PutAzureBlobStorage("PutAzureBlobStorage", utils::Identifier(), std::move(mock_blob_storage)));
+    char input_dir_mask[] = "/tmp/gt.XXXXXX";
+    auto input_dir = test_controller.createTempDirectory(input_dir_mask);
+    std::ofstream input_file_stream(input_dir + utils::file::FileUtils::get_separator() + "input_data.log");
+    input_file_stream << TEST_DATA;
+    input_file_stream.close();
+    get_file = plan->addProcessor("GetFile", "GetFile");
+    plan->setProperty(get_file, processors::GetFile::Directory.getName(), input_dir);
+    plan->setProperty(get_file, processors::GetFile::KeepSourceFile.getName(), "false");
+    update_attribute = plan->addProcessor(
```
hunyadi-dev, line 109: Minor, but can these not be combined into a single-line expression?
```c++
update_attribute = plan->addProcessor("UpdateAttribute", "UpdateAttribute", { {"success", "d"} },  true);
```

libminifi/test/azure-tests/PutAzureBlobStorageTests.cpp
```diff
@@ -0,0 +1,288 @@
...
+    LogTestController::getInstance().setTrace<minifi::core::ProcessSession>();
+    LogTestController::getInstance().setTrace<processors::GetFile>();
+    LogTestController::getInstance().setDebug<processors::UpdateAttribute>();
+    LogTestController::getInstance().setDebug<processors::LogAttribute>();
+    LogTestController::getInstance().setTrace<minifi::azure::processors::PutAzureBlobStorage>();
+
+    // Build MiNiFi processing graph
+    plan = test_controller.createPlan();
+    mock_blob_storage_ptr = new MockBlobStorage();
+    std::unique_ptr<minifi::azure::storage::BlobStorage> mock_blob_storage(mock_blob_storage_ptr);
```
hunyadi-dev, line 98: Minor, but maybe `utils::make_unique`?

libminifi/test/azure-tests/PutAzureBlobStorageTests.cpp
```diff
@@ -0,0 +1,288 @@
...
+    plan->setProperty(update_attribute, "test.sas_token", SAS_TOKEN, true);
+    plan->setProperty(put_azure_blob_storage, "SAS Token", "${test.sas_token}");
+    test_controller.runSession(plan, true);
+    REQUIRE(mock_blob_storage_ptr->getConnectionString() == "AccountName=" + STORAGE_ACCOUNT_NAME + ";SharedAccessSignature=" + SAS_TOKEN);
+  }
+
+  SECTION("Account name and SAS token with question mark set in properties") {
+    plan->setProperty(update_attribute, "test.account_name", STORAGE_ACCOUNT_NAME, true);
+    plan->setProperty(put_azure_blob_storage, "Storage Account Name", "${test.account_name}");
+    plan->setProperty(update_attribute, "test.sas_token", "?" + SAS_TOKEN, true);
```
hunyadi-dev, line 213: What does this test?

libminifi/test/azure-tests/PutAzureBlobStorageTests.cpp
```diff
@@ -0,0 +1,288 @@
+/**
```
hunyadi-dev, line 1: Great tests!

extensions/azure/storage/AzureBlobStorage.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+void AzureBlobStorage::resetClientIfNeeded(const std::string &connection_string, const std::string &container_name) {
+  if (connection_string == connection_string_ && container_name_ == container_name) {
+    logger_->log_debug("Client credentials have not changed, no need to reset client");
+    return;
+  }
+  connection_string_ = connection_string;
+  container_name_ = container_name;
+  logger_->log_debug("Client has been reset with new credentials");
+  container_client_ = minifi::utils::make_unique<Azure::Storage::Blobs::BlobContainerClient>(Azure::Storage::Blobs::BlobContainerClient::CreateFromConnectionString(connection_string, container_name));
```
hunyadi-dev, line 45: Minor, but calling `::reset` would be a little bit less boilerplate.

extensions/mqtt/processors/PublishMQTT.h
```diff
@@ -79,7 +79,7 @@ class PublishMQTT : public processors::AbstractMQTTProcessor {
       if (flow_size_ < max_seg_size_)
         max_seg_size_ = flow_size_;
       std::vector<unsigned char> buffer;
-      buffer.reserve(max_seg_size_);
+      buffer.resize(max_seg_size_);
```
hunyadi-dev, line 82: Does this change belong here?
If so, moving this to the constructor is more optimal:
```c++
std::vector<unsigned char> buffer(max_seg_size_);
```

extensions/aws/s3/S3WrapperBase.cpp
```diff
@@ -144,7 +144,7 @@ bool S3WrapperBase::deleteObject(const std::string& bucket, const std::string& o
 int64_t S3WrapperBase::writeFetchedBody(Aws::IOStream& source, const int64_t data_size, const std::shared_ptr<io::BaseStream>& output) {
   static const uint64_t BUFFER_SIZE = 4096;
   std::vector<uint8_t> buffer;
-  buffer.reserve(BUFFER_SIZE);
+  buffer.resize(BUFFER_SIZE);
```
hunyadi-dev, line 147: Both the static bufffer size and the preallocation adds performance overhead.
```c++
std::vector<unsigned char> buffer(4096);
```

extensions/azure/AzureLoader.h
```diff
@@ -0,0 +1,69 @@
...
+    } else {
+      return nullptr;
+    }
+  }
+
+  static bool added;
+};
+
+extern "C" {
+DLL_EXPORT void *createAzureFactory(void);
```
hunyadi-dev, line 68: Both here and at the declaration: do we need to use `(void)` as opposed to just `()`?

extensions/azure/processors/PutAzureBlobStorage.cpp
```diff
@@ -0,0 +1,229 @@
...
+    SASToken,
+    CommonStorageAccountEndpointSuffix,
+    ConnectionString,
+    AzureStorageCredentialsService,
+    Blob,
+    CreateContainer
+  });
+  // Set the supported relationships
+  setSupportedRelationships({
+    Failure,
```
hunyadi-dev, line 106: This order is a bit pessimistic :smile:

extensions/azure/processors/PutAzureBlobStorage.h
```diff
@@ -0,0 +1,137 @@
...
+    uint64_t flow_size_;
+    azure::storage::BlobStorage &blob_storage_wrapper_;
+    std::string blob_name_;
+    utils::optional<azure::storage::UploadBlobResult> result_ = utils::nullopt;
+  };
+
+ private:
+  friend class ::PutAzureBlobStorageTestsFixture;
+
+  explicit PutAzureBlobStorage(std::string name, minifi::utils::Identifier uuid, std::unique_ptr<storage::BlobStorage> blob_storage_wrapper)
```
hunyadi-dev, line 110: I would by default to using const ref args, but at least, why not take `uuid` by const ref?

extensions/azure/processors/PutAzureBlobStorage.h
```diff
@@ -0,0 +1,137 @@
...
+  static const core::Relationship Failure;
+  static const core::Relationship Success;
+
+  explicit PutAzureBlobStorage(std::string name, minifi::utils::Identifier uuid = minifi::utils::Identifier())
+    : PutAzureBlobStorage(name, uuid, nullptr) {
+  }
+
+  ~PutAzureBlobStorage() override = default;
+
+  bool supportsDynamicProperties() override { return true; }
```
hunyadi-dev, line 69: What are the dynamic properties used for?

extensions/azure/processors/PutAzureBlobStorage.cpp
```diff
@@ -0,0 +1,229 @@
...
+void PutAzureBlobStorage::onTrigger(const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSession> &session) {
+  logger_->log_debug("PutAzureBlobStorage onTrigger");
+  std::shared_ptr<core::FlowFile> flow_file = session->get();
+  if (!flow_file) {
+    return;
+  }
+
+  auto connection_string = getConnectionString(context, flow_file);
+  if (connection_string.empty()) {
+    logger_->log_error("Connection string is empty!");
```
hunyadi-dev, line 178: Minor, but shouldn't these missing properties be checked `onSchedule`?

extensions/azure/processors/PutAzureBlobStorage.cpp
```diff
@@ -0,0 +1,229 @@
...
+    }
+    PutAzureBlobStorage::ReadCallback callback(flow_file->getSize(), *blob_storage_wrapper_, blob_name);
+    session->read(flow_file, &callback);
+    upload_result = callback.getResult();
+  }
+
+  if (!upload_result) {
+    logger_->log_error("Failed to upload blob '%s' to Azure storage container '%s'", blob_name, container_name);
+    session->transfer(flow_file, Failure);
+  } else {
```
hunyadi-dev, line 212: Minor, but adding the scope could be omitted if the unhappy path returned.

extensions/aws/processors/S3Processor.cpp
```diff
@@ -143,11 +143,13 @@ minifi::utils::optional<Aws::Auth::AWSCredentials> S3Processor::getAWSCredential
 
   std::shared_ptr<core::controller::ControllerService> service = context->getControllerService(service_name);
   if (!service) {
+    logger_->log_error("AWS credentials service with name: '%s' could not be found", service_name.c_str());
```
hunyadi-dev, line 146: I think you don't need `c_str()` here.

docker/test/integration/minifi/processors/PutAzureBlobStorage.py
```diff
@@ -0,0 +1,13 @@
...
+class PutAzureBlobStorage(Processor):
+    def __init__(self):
+        super(PutAzureBlobStorage, self).__init__('PutAzureBlobStorage',
+                                                  properties={
+                                                      'Container Name': 'test_container',
+                                                      'Connection String': 'DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://azure-storage-server:10000/devstoreaccount1;QueueEndpoint=http://azure-storage-server:10001/devstoreaccount1;',
+                                                      'Blob': 'test_blob',
+                                                      'Create Container': 'true',
+                                                  },
+                                                  auto_terminate=['success'])
```
hunyadi-dev, line 13: Does this processor not have a failure relationship as well?

docker/test/integration/minifi/core/DockerTestCluster.py
```diff
@@ -110,6 +111,12 @@ def check_s3_server_object_metadata(self, content_type="application/octet-stream
         server_metadata = json.loads(metadata_json)
         return server_metadata["contentType"] == content_type and metadata == server_metadata["userMetadata"]
 
+    @retry_check()
+    def check_azure_storage_server_data(self, test_data):
+        data_file = subprocess.check_output(["docker", "exec", "azure-storage-server", "find", "/data/__blobstorage__", "-type", "f"]).decode(sys.stdout.encoding).strip()
```
hunyadi-dev, line 116: `sys.stdout.encoding` is not GitHub Actions safe. See #1018 .
image

libminifi/test/azure-tests/PutAzureBlobStorageTests.cpp
lordgamez: SAS usually generated in HTTP GET parameter format starting with `?` mark followed by `parameter=value` pairs. NiFi only accepts the SAS token with the starting `?` mark but the connection string we build does not require this so I wanted to accept it in both form. I test if both formats result in the same connection string.

libminifi/test/azure-tests/PutAzureBlobStorageTests.cpp
lordgamez: Thanks!

libminifi/test/azure-tests/PutAzureBlobStorageTests.cpp
lordgamez: Unfortunately make_unique only works with constructor parameters, but does not work with an existing object.

extensions/azure/storage/AzureBlobStorage.cpp
lordgamez: `CreateFromConnectionString` does not return a pointer but an object so we cannot use `reset` here.

extensions/mqtt/processors/PublishMQTT.h
lordgamez: @adamdebreceni suggested using resize instead of reserve, and I changed it in all occurrences. I'll move it to constructor.

extensions/azure/AzureLoader.h
lordgamez: It is supposed to be the explicit way of saying the function "takes no parameters" as opposed to "taking an unspecified number of arguments of unspecified type" in C according to https://stackoverflow.com/questions/51032/is-there-a-difference-between-foovoid-and-foo-in-c-or-c

libminifi/test/azure-tests/PutAzureBlobStorageTests.cpp
lordgamez: Updated in [48ec7a](https://github.com/apache/nifi-minifi-cpp/pull/979/commits/48ec7a2ebb6d109c147f2a3b83aa2ae8996acfa2)

extensions/mqtt/processors/PublishMQTT.h
lordgamez: Updated in [48ec7a](https://github.com/apache/nifi-minifi-cpp/pull/979/commits/48ec7a2ebb6d109c147f2a3b83aa2ae8996acfa2)

extensions/aws/s3/S3WrapperBase.cpp
lordgamez: Updated in [48ec7a](https://github.com/apache/nifi-minifi-cpp/pull/979/commits/48ec7a2ebb6d109c147f2a3b83aa2ae8996acfa2)

extensions/azure/processors/PutAzureBlobStorage.cpp
lordgamez: Updated in [48ec7a](https://github.com/apache/nifi-minifi-cpp/pull/979/commits/48ec7a2ebb6d109c147f2a3b83aa2ae8996acfa2)

extensions/azure/processors/PutAzureBlobStorage.h
lordgamez: Updated in [48ec7a](https://github.com/apache/nifi-minifi-cpp/pull/979/commits/48ec7a2ebb6d109c147f2a3b83aa2ae8996acfa2)

extensions/azure/processors/PutAzureBlobStorage.h
lordgamez: My mistake, they're not used in this processor. Updated in [48ec7a](https://github.com/apache/nifi-minifi-cpp/pull/979/commits/48ec7a2ebb6d109c147f2a3b83aa2ae8996acfa2)

extensions/azure/processors/PutAzureBlobStorage.cpp
lordgamez: Added checks in [48ec7a](https://github.com/apache/nifi-minifi-cpp/pull/979/commits/48ec7a2ebb6d109c147f2a3b83aa2ae8996acfa2)

extensions/azure/processors/PutAzureBlobStorage.cpp
lordgamez: Updated in [48ec7a](https://github.com/apache/nifi-minifi-cpp/pull/979/commits/48ec7a2ebb6d109c147f2a3b83aa2ae8996acfa2)

extensions/aws/processors/S3Processor.cpp
lordgamez: Updated in [48ec7a](https://github.com/apache/nifi-minifi-cpp/pull/979/commits/48ec7a2ebb6d109c147f2a3b83aa2ae8996acfa2)

docker/test/integration/minifi/processors/PutAzureBlobStorage.py
lordgamez: Updated in [48ec7a](https://github.com/apache/nifi-minifi-cpp/pull/979/commits/48ec7a2ebb6d109c147f2a3b83aa2ae8996acfa2)

docker/test/integration/minifi/core/DockerTestCluster.py
lordgamez: Updated in [48ec7a](https://github.com/apache/nifi-minifi-cpp/pull/979/commits/48ec7a2ebb6d109c147f2a3b83aa2ae8996acfa2). It may cause a merge conflict, but I will solve it if it occurs.

libminifi/test/azure-tests/PutAzureBlobStorageTests.cpp
hunyadi-dev: I meant something like:
`const auto mock_blob_storage = utils::make_unique<MockBlobStorage>();`

docker/test/integration/minifi/core/DockerTestCluster.py
hunyadi-dev: Hope I can merge first :D

extensions/aws/s3/S3WrapperBase.cpp
```diff
@@ -142,13 +142,11 @@ bool S3WrapperBase::deleteObject(const std::string& bucket, const std::string& o
...
 int64_t S3WrapperBase::writeFetchedBody(Aws::IOStream& source, const int64_t data_size, const std::shared_ptr<io::BaseStream>& output) {
-  static const uint64_t BUFFER_SIZE = 4096;
-  std::vector<uint8_t> buffer;
-  buffer.reserve(BUFFER_SIZE);
+  std::vector<uint8_t> buffer(4096);
 
   int64_t write_size = 0;
   while (write_size < data_size) {
-    auto next_write_size = data_size - write_size < BUFFER_SIZE ? data_size - write_size : BUFFER_SIZE;
+    auto next_write_size = data_size - write_size < 4096 ? data_size - write_size : 4096;
```
adamdebreceni, line 149: I know it's old code, but this looks like `std::min(data_size - write_size, buffer.size())`

lordgamez: That change is actually part of #1020

libminifi/test/azure-tests/PutAzureBlobStorageTests.cpp
lordgamez: Sorry, I misunderstood the comment, updated in [8da842](https://github.com/apache/nifi-minifi-cpp/pull/979/commits/8da842073ee927f193f29db5da6603f691000345)

extensions/azure/processors/PutAzureBlobStorage.h
```diff
@@ -0,0 +1,136 @@
...
+ private:
+  friend class ::PutAzureBlobStorageTestsFixture;
+
+  explicit PutAzureBlobStorage(const std::string& name, const minifi::utils::Identifier& uuid, std::unique_ptr<storage::BlobStorage> blob_storage_wrapper)
+    : core::Processor(name, uuid)
+    , blob_storage_wrapper_(std::move(blob_storage_wrapper)) {
+  }
+
+  std::string getConnectionStringFromControllerService(const std::shared_ptr<core::ProcessContext> &context) const;
+  std::string getAzureConnectionStringFromProperties(
```
szaszm, line 115: `getAzureConnectionStringFromProperties` could be made static and possibly hidden in the implementation file.

libminifi/test/azure-tests/PutAzureBlobStorageTests.cpp
```diff
@@ -0,0 +1,276 @@
...
+    LogTestController::getInstance().setDebug<processors::UpdateAttribute>();
+    LogTestController::getInstance().setDebug<processors::LogAttribute>();
+    LogTestController::getInstance().setTrace<minifi::azure::processors::PutAzureBlobStorage>();
+
+    // Build MiNiFi processing graph
+    plan = test_controller.createPlan();
+    auto mock_blob_storage = utils::make_unique<MockBlobStorage>();
+    mock_blob_storage_ptr = mock_blob_storage.get();
+    put_azure_blob_storage = std::shared_ptr<minifi::azure::processors::PutAzureBlobStorage>(
+      new minifi::azure::processors::PutAzureBlobStorage("PutAzureBlobStorage", utils::Identifier(), std::move(mock_blob_storage)));
```
szaszm, line 100: Use `std::make_shared`

lordgamez:
https://github.com/apache/nifi-minifi-cpp/pull/979#discussion_r589398242

I think it may happen that the next flow file contains proper authentication info and in that case we should not yield.

libminifi/test/azure-tests/PutAzureBlobStorageTests.cpp
lordgamez: Unfortunately `std::make_shared` does not work with the used private constructor, because only the `PutAzureBlobStorageTestsFixture` is the friend class of `PutAzureBlobStorage`, but it is private in the context of the `make_shared` call.

extensions/azure/processors/PutAzureBlobStorage.h
lordgamez: Updated in 1f07ca2f5b54c5ba7f4975634e9e320c1c0c9449

extensions/azure/storage/BlobStorage.h
```diff
@@ -0,0 +1,63 @@
...
+  std::string etag;
+  std::size_t length;
+  std::string timestamp;
+};
+
+class BlobStorage {
+ public:
+  BlobStorage(const std::string &connection_string, const std::string &container_name)
+    : connection_string_(connection_string)
+    , container_name_(container_name) {
```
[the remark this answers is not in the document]
lordgamez: Updated in 1f07ca2f5b54c5ba7f4975634e9e320c1c0c9449

libminifi/test/azure-tests/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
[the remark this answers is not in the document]
lordgamez: Updated in 1f07ca2f5b54c5ba7f4975634e9e320c1c0c9449

szaszm (approved):

PR #984, opened 2021-01-28
MINIFICPP-1462 Fix incorrect usage of move constructors and assignment operators

dam4rus:
Apparently this PR raises some questions about how to properly fix this.

Some of the classes are actually never instantiated on the stack. So the move constructors are actually never called. All of the classes with deleted copy and move constructors in the PR are such classes. I think it's safer to delete the special functions for such classes instead of them doing something wrong if somehow they are called.

Other classes, like Peer, had only a move constructor which would lead to strange behaviors. A user-defined move constructor deletes the copy constructor, however they had defaulted copy and move assignment operators. So they could have been copied with assignment but not through constructor. I don't believe this was intentional?

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/mqtt/controllerservice/MQTTControllerService.h
+++ b/extensions/mqtt/controllerservice/MQTTControllerService.h
@@ -49,17 +49,11 @@ class Message {
       : topic_(topic),
         data_((uint8_t*) data, ((uint8_t*)data + dataLen)) {
   }
-  explicit Message(const Message &&other)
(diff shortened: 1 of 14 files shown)


szaszm (approved):
Issues around copy/move were commonplace in the codebase for a long time, but most commonly touched and trivially changed parts are now fixed. Thanks for cleaning up more of these.

arpadboda (approved):
LGTM, thanks!

PR #992, opened 2021-02-05
MINIFICPP-1467 Remove mutex from Connection

dam4rus:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/utils/MinifiConcurrentQueue.h
+++ b/libminifi/include/utils/MinifiConcurrentQueue.h
@@ -148,13 +148,38 @@ class ConcurrentQueue {
 template <typename T>
 class LockedConcurrentQueue {
  public:
+  using reference = typename std::deque<T>::reference;
+  using const_reference = typename std::deque<T>::const_reference;
+
   explicit LockedConcurrentQueue(ConcurrentQueue<T>& concurrentQueue)
       : queue_(std::ref(concurrentQueue.queue_))
       , lock_(concurrentQueue.mtx_) {
   }
 
-  std::deque<T>* operator->() const { return &queue_.get(); }
-  std::deque<T>& operator*() const { return queue_; }
+  void push(const T& value) {
+    queue_.get().push_back(value);
+  }
+
+  void push(T&& value) {
+    queue_.get().push_back(std::move(value));
+  }
+
+  bool empty() const {
+    return queue_.get().empty();
+  }
+
+  reference front() {
+    return queue_.get().front();
+  }
+
+  const_reference front() const {
+    return queue_.get().front();
+  }
+
+  void pop() {
+    queue_.get().pop_front();
+  }
+
  private:
   std::reference_wrapper<std::deque<T>> queue_;
   std::unique_lock<std::mutex> lock_;
--- a/libminifi/include/utils/MinifiConcurrentQueue.h
+++ b/libminifi/include/utils/MinifiConcurrentQueue.h
@@ -33,10 +33,14 @@ namespace nifi {
 namespace minifi {
 namespace utils {
 
+template <typename T>
+class LockedConcurrentQueue;
+
 // Provides a queue API and guarantees no race conditions in case of multiple producers and consumers.
 // Guarantees elements to be dequeued in order of insertion
 template <typename T>
 class ConcurrentQueue {
+  friend class LockedConcurrentQueue<T>;
  public:
   ConcurrentQueue() = default;
 
@@ -87,6 +91,10 @@ class ConcurrentQueue {
     queue_.emplace_back(std::forward<Args>(args)...);
   }
 
+  LockedConcurrentQueue<T> lock() {
+    return LockedConcurrentQueue<T>(*this);
+  }
+
  private:
   ConcurrentQueue(ConcurrentQueue&& other, std::lock_guard<std::mutex>&)
     : queue_(std::move(other.queue_)) {}
@@ -135,6 +143,23 @@ class ConcurrentQueue {
   std::deque<T> queue_;
 };
 
+// Enables batched operations on a ConcurrentQueue by holding a reference to the queue and locking it's mutex
+// until it leaves the scope. Can help performance by avoiding multiple locking in loops, etc.
+template <typename T>
+class LockedConcurrentQueue {
+ public:
+  explicit LockedConcurrentQueue(ConcurrentQueue<T>& concurrentQueue)
+    : queue_(std::ref(concurrentQueue.queue_))
+    , lock(concurrentQueue.mtx_)
+    {
+    }
+
+  std::deque<T>* operator->() const { return &queue_.get(); }
+  std::deque<T>& operator*() const { return queue_; }
+ private:
+  std::unique_lock<std::mutex> lock;
+  std::reference_wrapper<std::deque<T>> queue_;
+};
 
 // A ConcurrentQueue extended with a condition variable to be able to block and wait for incoming data
 // Stopping interrupts all consumers without a chance to consume remaining elements in the queue although elements can still be enqueued
--- a/libminifi/include/utils/MinifiConcurrentQueue.h
+++ b/libminifi/include/utils/MinifiConcurrentQueue.h
@@ -150,8 +150,7 @@ class LockedConcurrentQueue {
  public:
   explicit LockedConcurrentQueue(ConcurrentQueue<T>& concurrentQueue)
       : queue_(std::ref(concurrentQueue.queue_))
-      , lock_(concurrentQueue.mtx_)
-  {
+      , lock_(concurrentQueue.mtx_) {
   }
 
   std::deque<T>* operator->() const { return &queue_.get(); }
--- a/libminifi/include/utils/MinifiConcurrentQueue.h
+++ b/libminifi/include/utils/MinifiConcurrentQueue.h
@@ -149,16 +149,16 @@ template <typename T>
 class LockedConcurrentQueue {
  public:
   explicit LockedConcurrentQueue(ConcurrentQueue<T>& concurrentQueue)
-    : queue_(std::ref(concurrentQueue.queue_))
-    , lock(concurrentQueue.mtx_)
-    {
-    }
+      : queue_(std::ref(concurrentQueue.queue_))
+      , lock_(concurrentQueue.mtx_)
+  {
+  }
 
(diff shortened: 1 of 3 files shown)


fgerlits (commented):
This solution exposes the internal implementation of `ConcurrentQueue` and gives direct access to the `deque` contained in it.  It would be good if we could avoid this.

One alternative solution would be to add a `push_all(items)` and a `pop_all()` method to `ConcurrentQueue`.

dam4rus:
I agree that this implementation doesn't fit the current design but to avoid needless lockings and potentially tank the performance I choose this over `push_all`, `pop_all`

For example:
At https://github.com/apache/nifi-minifi-cpp/pull/992/files#diff-edfe02f672dad6c32d860c97ea73e5f4b95889387942c8b372554ba43fda14cdL159 `flows` are filtered so I would need to create a new container. It's not a big issue since `shared_ptr` is fairly cheap to copy but nonetheless felt unnecessary. Also, what would be the implementation of `push_all`? Should it accept `std::vector`, `std::queue` or iterator pair?

But the biggest issue is here https://github.com/apache/nifi-minifi-cpp/pull/992/files#diff-edfe02f672dad6c32d860c97ea73e5f4b95889387942c8b372554ba43fda14cdL181 , since there's both pop and push in the same loop if the flow file is penalised, which breaks the loop and put's the flow file back into the queue. Maybe `pop`-ing should be moved to the end of the loop to avoid an unnecessary `pop` and `push` if the flow file is penalised? So it's not really a `pop_all`. So without exposing the underlying queue somehow, this would require some refactoring and/or hurting the performance by locking the mutex with every call to `ConcurrentQueue`'s methods.

fgerlits:
Yes, using (a modified or extended version of) ConcurrentQueue [rather than exposing the internal mutex and deque and using those] would require some refactoring of the Connection class.  That is not necessarily a bad thing, as `Connection::poll()` contains some duplicated code.

I also can't see a way to avoid some performance degradation, as you mentioned.  But I don't think that is a big problem, as the difference in performance would be small.

To be honest, at this point I would vote for keeping the old code as it was, but let's see what other people say.

dam4rus:
TBH, I don't really see how `Connection::poll()` could be refactored, since I'm not super familiar with the code base, to avoid performance degradation with `ConcurrentQueue` since it has a more complex structure. Using a lambda would be weird since the loop has `break` in it. Maybe something like `ConcurrentQueue::consume_while(function(item)->bool)` could be implemented, where a `false` return value would indicate the `break` and that the item isn't poped out of the queue. My biggest issue with this approach is that I generally don't think that having a side-effect in a consumer lambda is good practise.

I went with this design because it required the least modification and the performance is the same as before. I kinda agree that maybe leaving it as it was is fine.

szaszm:
About exposing internals: I don't think it's happening here. I view `LockedConcurrentQueue` as an implementation detail of `ConcurrentQueue`, so they are one unit, even though they are syntactically implemented as two classes. `friend` is not evil IMO.

I like the design, but on the flipside, it introduces more complexity than it removes. Overall, I support this change. If we happen to reuse this change at least once more, then the introduced complexity would pay off.

Let's wait for more opinions!

adamdebreceni:
@szaszm I think @fgerlits was referring to the `std::deque<T>* LockedConcurrentQueue::operator->() const { return &queue_.get(); }` providing direct access to the underlying queue, if `LockedConcurrentQueue` would implement the necessary methods (`push_back`, `pop_front`, `front`, `empty` by the looks of it) this could be a solid solution IMO

szaszm:
Good point @adamdebreceni, that would indeed be a very welcome change. Client code being dependent on the internal std::deque could prevent us from moving to a better implementation in the future.

libminifi/include/utils/MinifiConcurrentQueue.h
```diff
@@ -135,6 +143,22 @@ class ConcurrentQueue {
   std::deque<T> queue_;
 };
 
+// Enables batched operations on a ConcurrentQueue by holding a reference to the queue and locking it's mutex
```
hunyadi-dev, line 146: Very minor, but you probably meant "its" instead of it's.

libminifi/include/utils/MinifiConcurrentQueue.h
```diff
@@ -135,6 +143,22 @@ class ConcurrentQueue {
...
+// until it leaves the scope. Can help performance by avoiding multiple locking in loops, etc.
+template <typename T>
+class LockedConcurrentQueue {
+ public:
+  explicit LockedConcurrentQueue(ConcurrentQueue<T>& concurrentQueue)
+      : queue_(std::ref(concurrentQueue.queue_))
+      , lock_(concurrentQueue.mtx_) {
+  }
+
+  std::deque<T>* operator->() const { return &queue_.get(); }
```
hunyadi-dev, line 156: This exposes the underlying queue object. This is a big no-no.

libminifi/include/utils/MinifiConcurrentQueue.h
```diff
@@ -87,6 +91,10 @@ class ConcurrentQueue {
     queue_.emplace_back(std::forward<Args>(args)...);
   }
 
+  LockedConcurrentQueue<T> lock() {
```
hunyadi-dev, line 94: Why would you do this? If you really want to extend the functionality of the class, instead of adding friend accessors, make your queue inherit from the `ConcurrentQueue`.

libminifi/include/utils/MinifiConcurrentQueue.h
dam4rus: Good point.

libminifi/src/Connection.cpp
```diff
@@ -224,11 +216,11 @@ std::shared_ptr<core::FlowFile> Connection::poll(std::set<std::shared_ptr<core::
 }
 
 void Connection::drain(bool delete_permanently) {
-  std::lock_guard<std::mutex> lock(mutex_);
+  auto lockedQueue = queue_.lock();
```
hunyadi-dev, line 219: A better pattern would be just to lock the queue until you move the data out of it.

libminifi/include/utils/MinifiConcurrentQueue.h
dam4rus: It has already been discussed and if everyone agrees with this approach I'm totally fine with providing a public API instead of overloaded operators.

libminifi/src/Connection.cpp
```diff
@@ -157,15 +149,15 @@ void Connection::put(const std::shared_ptr<core::FlowFile>& flow) {
 
 void Connection::multiPut(std::vector<std::shared_ptr<core::FlowFile>>& flows) {
   {
-    std::lock_guard<std::mutex> lock(mutex_);
+    auto lockedQueue = queue_.lock();
```
hunyadi-dev, line 152: Instead of this, I would create an interface on the queue that performs batch insertion. This piece of API could be better tested as well.

libminifi/include/utils/MinifiConcurrentQueue.h
dam4rus: Because `LockedConcurrentQueue` **is not a** `ConcurrentQueue`. It's just a utility class that holds a lock to a `ConcurrentQueue`'s mutex and provides access to the underlying queue until it goes out of scope. So instead of locking the mutex with every operation you can do multiple operations with a single locking. I'm open to a better name for the class.

libminifi/src/Connection.cpp
dam4rus: You mean only lock it before calling empty and unlock after pop_front? That's not an equivalent refactor. I don't know how costly the operations are after pop-ing the element but locking in a loop can definitely degrade the performance.

hunyadi-dev (changes requested):
The old implementation seems more robust and exposing the implementation of ConcurrentQueue should be definitely avoided. Consider refactoring the Connection class and extending the interface of ConcurrentQueue. If it is really not the right tool to handle the queue of connections then [we would be better of not using it](https://www.wikiwand.com/en/Law_of_the_instrument). I am not convinced that that locking the connection mutexes is throttling performance.

dam4rus:
I'm pretty sure that locking multiple times in a loop can [degrade performance](https://quick-bench.com/q/6Tb3pu8sUC80EtpJtxht7uxHlLE) significantly and [that's not what you want](https://www.techiedelight.com/why-vector-class-java-obsolete/#:~:text=This%20is%20because%20Vector%20synchronizes,synchronized%20and%20not%20individual%20operations.) most of the time. While `ConcurrentQueue::pop_all` and `ConcurrentQueue::push_all` can solve some cases but for more complex cases like `Connection::poll` it's kinda hard to write a generic solution. I can probably make it work somehow but I'm not sure that it would be easier to reason about that this solution.

I've created this PR mainly as a PoC and I agree that a public API would be better than exposing the underlying queue. I don't mind if you guys would rather extend the ConcurrentQueue class or leave it as it is.

libminifi/src/Connection.cpp
```diff
@@ -179,56 +171,44 @@ void Connection::multiPut(std::vector<std::shared_ptr<core::FlowFile>>& flows) {
...
-        break;
-      }
-      std::shared_ptr<Connectable> connectable = std::static_pointer_cast<Connectable>(shared_from_this());
-      item->setConnection(connectable);
-      logger_->log_debug("Dequeue flow file UUID %s from connection %s", item->getUUIDStr(), name_);
-      return item;
+    // Flow record not expired
+    if (item->isPenalized()) {
+      // Flow record was penalized
+      break;
```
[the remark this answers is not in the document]
dam4rus: Damn. You are right. Changing behavior wasn't intentional. I'll revert this back while still removing the code duplication if that's okay.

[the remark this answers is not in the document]
dam4rus: @fgerlits Your PR also changes the underlying queue in Connection, which makes rebasing almost impossible. I wonder if with your changes does this PR has any merit? At this point I think abandoning this ticket would be the best option.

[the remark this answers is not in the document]
fgerlits: Yes, you're right, the two changes were at cross purposes -- this was not intentional, sorry.

fgerlits:
Closing this, as we ended up going with a different approach, see PR #1027.

PR #994, opened 2021-02-05
MINIFICPP-1434: Fix flaky CSite2SiteTests on macOS

martinzink:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/nanofi/tests/CSite2SiteTests.cpp
+++ b/nanofi/tests/CSite2SiteTests.cpp
@@ -17,6 +17,7 @@
  */
 
 #include <stdlib.h>
+#include <signal.h>
 
 #include <algorithm>
 #include <chrono>
@@ -68,6 +69,9 @@ void wait_until(std::atomic<bool>& b) {
 }
 
 TEST_CASE("TestSetPortId", "[S2S1]") {
+#ifndef WIN32
+  signal(SIGPIPE, SIG_IGN);
+#endif
   SiteToSiteCPeer peer;
   initPeer(&peer, "fake_host", 65433);
   CRawSiteToSiteClient * protocol = (CRawSiteToSiteClient*)malloc(sizeof(CRawSiteToSiteClient));
@@ -166,8 +170,12 @@ void different_version_bootstrap(minifi::io::BaseStream* stream, TransferState& 
   sunny_path_bootstrap(stream, transfer_state, s2s_data);
 }
 
+
 TEST_CASE("TestSiteToBootStrap", "[S2S3]") {
 
+#ifndef WIN32
+  signal(SIGPIPE, SIG_IGN);
+#endif
   std::array<std::function<void(minifi::io::BaseStream*, TransferState&, S2SReceivedData&)>, 2> bootstrap_functions =
       {sunny_path_bootstrap, different_version_bootstrap};
 
--- a/nanofi/tests/CSite2SiteTests.cpp
+++ b/nanofi/tests/CSite2SiteTests.cpp
@@ -17,7 +17,9 @@
  */
 
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


martinzink:
The issue is that on macOS the default sigpipe handler crashes the application.
In production code we are already ignoring this by setting the sigpipe handle to nothing. see https://github.com/apache/nifi-minifi-cpp/blob/main/main/MiNiFiMain.cpp#L181
Given this I think it would be appropriate to ignore it in these tests as well.

I've tested this on my fork with custom CI, on both macos-xcode12.0 and macos-xcode11.2.1 with 200 runs
`make test ARGS="--timeout 300 -j4 --output-on-failure -R CSite2SiteTests --repeat-until-fail 200`
Without the fix both configuration failed (both around the 100th run), with the fix they passed 3 times (thats 1200 runs without failure)

nanofi/tests/CSite2SiteTests.cpp
```diff
@@ -166,8 +170,12 @@ void different_version_bootstrap(minifi::io::BaseStream* stream, TransferState&
   sunny_path_bootstrap(stream, transfer_state, s2s_data);
 }
 
+
```
szaszm, line 173: I don't think we need this empty line

nanofi/tests/CSite2SiteTests.cpp
```diff
@@ -17,6 +17,7 @@
  */
 
 #include <stdlib.h>
+#include <signal.h>
```
szaszm, line 20: This should be wrapped in `#ifndef WIN32` as well

nanofi/tests/CSite2SiteTests.cpp
martinzink: fixed

nanofi/tests/CSite2SiteTests.cpp
martinzink: fixed

szaszm (approved):

lordgamez (approved):

nanofi/tests/CSite2SiteTests.cpp
```diff
@@ -167,7 +173,9 @@ void different_version_bootstrap(minifi::io::BaseStream* stream, TransferState&
 }
 
 TEST_CASE("TestSiteToBootStrap", "[S2S3]") {
-
+#ifndef WIN32
+  signal(SIGPIPE, SIG_IGN);
```
arpadboda, line 177: Could you elaborate how this fixes the issue?

martinzink: Sure, a SIGPIPE is sent to a process if it tried to write to a socket that had been shutdown for writing or isn't connected (anymore), this can easily happen in multithreaded enviroment. 
The issue is that on macOS the default sigpipe handler crashes the application.
In production code we are already ignoring this signal by setting the sigpipe handle to nothing. see https://github.com/apache/nifi-minifi-cpp/blob/main/main/MiNiFiMain.cpp#L181

Given that in production code we are ignoring the signal, I think it would be best to ignore it here as well.

I've tested this on my fork with custom CI, on both macos-xcode12.0 and macos-xcode11.2.1 with 200 runs
`make test ARGS="--timeout 300 -j4 --output-on-failure -R CSite2SiteTests --repeat-until-fail 200`
Without the fix both configuration failed (both around the 100th run), with the fix they passed 3 times (that's 1200 runs without failure)

arpadboda: Thanks for the details, makes sense!
At least whoever finds this change lager, will see the reasoning here.

arpadboda (approved):

PR #999, opened 2021-02-09
MINIFICPP-1485 Improve 'exclusive property' error message (mini-PR)

fgerlits:
https://issues.apache.org/jira/browse/MINIFICPP-1485

Old error message:
```
[error] Unable to parse configuration file for component named 'InvokeHTTP' because property
'SSL Context Service' is exclusive of property 'Remote URL' values matching '^http:.*$'
[in 'Processors' section of configuration file]
```
New error message:
```
[error] Unable to parse configuration file for component named 'InvokeHTTP' because property
'SSL Context Service' must not be set when the value of property 'Remote URL' matches '^http:.*$'
[in 'Processors' section of configuration file]
```

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/standard-processors/tests/unit/YamlConfigurationTests.cpp
+++ b/extensions/standard-processors/tests/unit/YamlConfigurationTests.cpp
@@ -709,7 +709,7 @@ TEST_CASE("Test Exclusive Property 2", "[YamlConfigurationExclusiveProperty2]") 
   } catch (const std::exception &e) {
     config_failed = true;
     REQUIRE("Unable to parse configuration file for component named 'component A' because "
-        "property 'Prop B' is exclusive of property 'Prop A' values matching '^val.*$' "
+        "property 'Prop B' must not be set when the value of property 'Prop A' matches '^val.*$' "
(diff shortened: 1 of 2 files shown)


arpadboda (approved):

hunyadi-dev (approved):
Looks good to me. However, I think readibility could be further improved with an error message like: 
> The property "SSL Context Service" can only be used for "Remote URL" values that do not start with "http:".

fgerlits:
> Looks good to me. However, I think readibility could be further improved with an error message like:
> 
> > The property "SSL Context Service" can only be used for "Remote URL" values that do not start with "http:".

Yes, that would be better.  But our current `->withExclusiveProperty()` allows any regular expression, and parsing the regex to check whether it is of the form `^something.*$` would be overkill, I think.

I would have done this the other way round, eg. `->withOtherPropertyMatchingRegex()` so we can say "property X must match regex ... if property Y is set" -- but changing this would break our yaml API, so I don't think we can do it.

lordgamez (approved):

szaszm (approved):

PR #1000, opened 2021-02-10
MINIFICPP-1451: ThreadPoolAdjust test transiently failed on Windows CI

martinzink:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/http-curl/tests/ThreadPoolAdjust.cpp
+++ b/extensions/http-curl/tests/ThreadPoolAdjust.cpp
@@ -31,11 +31,11 @@
 #include "processors/LogAttribute.h"
 #include "utils/IntegrationTestUtils.h"
 
-constexpr uint64_t INCREASED_WAITTIME_MSECS = DEFAULT_WAITTIME_MSECS * 1.5;
+constexpr uint64_t WAITTIME_MSECS = 5000;
 
 class HttpTestHarness : public IntegrationBase {
  public:
-  HttpTestHarness() : IntegrationBase(INCREASED_WAITTIME_MSECS) {
+  HttpTestHarness() : IntegrationBase(WAITTIME_MSECS) {
     char format[] = "/tmp/ssth.XXXXXX";
     dir = testController.createTempDirectory(format);
   }
--- a/extensions/http-curl/tests/ThreadPoolAdjust.cpp
+++ b/extensions/http-curl/tests/ThreadPoolAdjust.cpp
@@ -31,9 +31,11 @@
 #include "processors/LogAttribute.h"
 #include "utils/IntegrationTestUtils.h"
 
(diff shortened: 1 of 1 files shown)


extensions/http-curl/tests/ThreadPoolAdjust.cpp
```diff
@@ -31,9 +31,11 @@
 #include "processors/LogAttribute.h"
 #include "utils/IntegrationTestUtils.h"
 
+constexpr uint64_t INCREASED_WAITTIME_MSECS = DEFAULT_WAITTIME_MSECS * 1.5;
+
 class HttpTestHarness : public IntegrationBase {
  public:
-  HttpTestHarness() {
+  HttpTestHarness() : IntegrationBase(INCREASED_WAITTIME_MSECS) {
```
martinzink, line 38: Looking at the available logs on github, this issue is fairly regular (happened 7 times in the last 90 days, only on windows)

The strange thing is, usually by the time the assertion message appears the required log lines are already present.
So I suspected that the given wait time for these log messages are too low to be consistent for the windows CI. (I wasn't able to reproduce the issue on my windows machine)
I've tested the fix with a custom CI that's tries to run this test 200 times on windows2017 and windows2019.
Without the fix it wasn't able to run the 200 test run without failure 4 times out of 4. (Failures happened after the 1st, 11th, 12th, 103rd test run)
  https://github.com/martinzink/nifi-minifi-cpp/actions/runs/551019180
  https://github.com/martinzink/nifi-minifi-cpp/actions/runs/551315036
With the increased wait time it passed 6 times out of 6. (1200 successful test run)
  https://github.com/martinzink/nifi-minifi-cpp/actions/runs/551722774
  https://github.com/martinzink/nifi-minifi-cpp/actions/runs/552033701
  https://github.com/martinzink/nifi-minifi-cpp/actions/runs/552535235

fgerlits, line 38: I would hard-code the wait time value here instead of defining it as `DEFAULT_WAITTIME_MSECS * 1.5`, because I don't think we want this wait time to change if we change the default later.

Also, since `wait_time_` is only used as a parameter to `verifyLogLinePresenceInPollTime()`, ie. it will stop waiting earlier if the log line is found earlier, we could make it longer, eg. 5 seconds or even 10 seconds.

martinzink: Changed it to be hard-coded 5000 ms, instead of the previously calculated 4500 ms.

fgerlits (approved):

lordgamez (approved):
LGTM

arpadboda (approved):

PR #1005, opened 2021-02-16
MINIFICPP-1496 "Enable all extensions" in bootstrap.sh should not enable ASAN_BUILD

fgerlits:
https://issues.apache.org/jira/browse/MINIFICPP-1496

ASAN_BUILD should be handled like the the other Build Options (like USE_SHARED_LIBS and SKIP_TESTS) and not like extensions.  In particular, if you select option 2, "Enable all extensions", it should not be enabled.

(Ideally, the same should apply to ENABLE_JNI, but that would be a bigger change [see the Jira; thanks to @adebreceni for pointing this out], so I won't address that in this PR.)

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -290,6 +290,9 @@ add_dependency COAP_ENABLED "automake"
 add_dependency COAP_ENABLED "autoconf"
 add_dependency COAP_ENABLED "libtool"
 
+add_disabled_option JNI_ENABLED ${FALSE} "ENABLE_JNI"
+add_dependency JNI_ENABLED "jnibuild"
+
 add_disabled_option OPENCV_ENABLED ${FALSE} "ENABLE_OPENCV"
 
 add_disabled_option OPENCV_ENABLED ${FALSE} "ENABLE_OPENCV"
@@ -314,9 +317,6 @@ add_dependency TENSORFLOW_ENABLED "tensorflow"
 
 add_disabled_option OPC_ENABLED ${FALSE} "ENABLE_OPC"
 add_dependency OPC_ENABLED "mbedtls"
-
-JNI_ENABLED=${FALSE}
-add_dependency JNI_ENABLED "jnibuild"
 
 USE_SHARED_LIBS=${TRUE}
 TESTS_DISABLED=${FALSE}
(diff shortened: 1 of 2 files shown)


adamdebreceni (approved):

bootstrap.sh
```diff
@@ -318,14 +315,16 @@ add_dependency TENSORFLOW_ENABLED "tensorflow"
 add_disabled_option OPC_ENABLED ${FALSE} "ENABLE_OPC"
 add_dependency OPC_ENABLED "mbedtls"
 
+JNI_ENABLED=${FALSE}
+add_dependency JNI_ENABLED "jnibuild"
```
adamdebreceni, line 319: it seems to me, that `JNI_ENABLED` no longer being added to `OPTIONS` affects the platform specific dependency installer scripts, specifically they iterate over the `OPTIONS` and then installer their dependencies, could this cause that java will no longer be installed even if we enable jni?

adamdebreceni, line 319: it seems to me, that JNI_ENABLED no longer being added to OPTIONS affects the platform specific dependency installer scripts, specifically they iterate over the OPTIONS and then install their dependencies, could this cause that java will no longer be installed even if we enable jni?

adamdebreceni (changes requested):

bootstrap.sh
fgerlits: Yes, you're right.  I have reverted the JNI part of the change.

adamdebreceni (approved):

lordgamez (approved):

arpadboda (approved):

bootstrap.sh
```diff
@@ -320,12 +320,11 @@ add_dependency OPC_ENABLED "mbedtls"
 
 USE_SHARED_LIBS=${TRUE}
 TESTS_DISABLED=${FALSE}
+ASAN_ENABLED=${FALSE}
 
 ## name, default, values
 add_multi_option BUILD_PROFILE "RelWithDebInfo" "RelWithDebInfo" "Debug" "MinSizeRel" "Release"
 
-add_disabled_option ASAN_ENABLED ${FALSE} "ASAN_BUILD"
```
arpadboda, line 327: Are you sure this could be removed? 
For eg., settings are persisted without this?

adamdebreceni: nice, yes indeed, it has to be explicitly added to `save_state`

fgerlits: Indeed, it was missing from `bt_state`.  Fixed in https://github.com/apache/nifi-minifi-cpp/pull/1005/commits/28970606a01063e4491bbb35ca4de5899e1cd865.

This bootstrap script is a bit of a mess, and should be refactored, but I think that should be done in a separate PR.

arpadboda: Agreed.

PR #1012, opened 2021-02-19
MINIFICPP-1499 Add clcache to Windows builds

lordgamez:
After checking the available solutions for caching the build objects on Windows with Visual Studio, clcache was chosen. From the available options unfortunately sccache did not work properly with msbuild, and clcache seemed to be a better option than cclash.

Runtimes measured in Github Actions:
Without caching:
vs2017 build: 1h 28m 37s
windows-vs2019: 56m 44s

With caching:
windows-vs2017: 51m 24s
windows-vs2019: 27m 33s

-----------------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,9 +65,20 @@ jobs:
     name: "windows-vs2017"
     runs-on: windows-2016
     timeout-minutes: 120
+    env:
+      CLCACHE_DIR: ${{ GITHUB.WORKSPACE }}\clcache
+      CLCACHE_CL: C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.16.27023\bin\Hostx86\x86\cl_original.exe
     steps:
       - id: checkout
         uses: actions/checkout@v2
+      - id: cache
+        uses: actions/cache@v2
+        with:
+          path: ${{ env.CLCACHE_DIR }}
+          key: windows-vs2017-clcache-${{github.ref}}-${{github.sha}}
+          restore-keys: |
+            windows-vs2017-clcache-${{github.ref}}-
+            windows-vs2017-clcache-refs/heads/main-
       - name: Setup PATH
         uses: microsoft/setup-msbuild@v1.0.2
       - id: install-sqliteodbc-driver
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -64,27 +64,64 @@ jobs:
...
+          path: ${{ env.CLCACHE_DIR }}
+          key: windows-vs2017-clcache-${{github.ref}}-${{github.sha}}
+          restore-keys: |
+            windows-vs2017-clcache-${{github.ref}}-
+            windows-vs2017-clcache-refs/heads/main-
       - name: Setup PATH
         uses: microsoft/setup-msbuild@v1.0.2
+      - name: Setup clcache
+        run: |
+          (New-Object System.Net.WebClient).DownloadFile('https://github.com/frerich/clcache/releases/download/v4.2.0/clcache-4.2.0.zip', "$pwd\clcache.zip")
```
szaszm, line 86: clcache is not maintained, so I'm not a fan of using it.

.github/workflows/ci.yml
```diff
@@ -64,27 +64,64 @@ jobs:
...
+          Expand-Archive -Force -Path clcache.zip -DestinationPath "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.28.29333\bin\Hostx64\x64";
+          move "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.28.29333\bin\Hostx64\x64\cl.exe" "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.28.29333\bin\Hostx64\x64\cl_original.exe"
+          move "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.28.29333\bin\Hostx64\x64\cl.exe.config" "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.28.29333\bin\Hostx64\x64\cl_original.exe.config"
+          move "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.28.29333\bin\Hostx64\x64\clcache.exe" "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.28.29333\bin\Hostx64\x64\cl.exe"
```
szaszm, line 124: Toolchain version numbers like those in the path change with every upgrade. I expect this to break every few months, when GitHub decides to upgrade its toolchain.

szaszm (commented):
Due to the issues below, I think the maintenance costs make this change unfeasible.

.github/workflows/ci.yml
lordgamez: I will give another go with sccache and see if I can make it work. Unfortunately I only found 3 options of caching for MSVC (2 of which are not maintained), but still from what I've found clcache seem to be the preferred one. There are some forks with active development, the author of clcache advises people to merge new required features to those repositories. Although I think for our use case the last release version may suffice if sccache cannot be used.

.github/workflows/ci.yml
lordgamez: You are right, I will find a solution for this.

.github/workflows/ci.yml
szaszm: Did you try setting `CMAKE_CXX_COMPILER=sccache cl`?

lordgamez: I remember I tried `CMAKE_CXX_COMPILER_LAUNCHER` and it didn't work, maybe I tried this as well, but I'm not sure, I will give it a go.

.github/workflows/ci.yml
szaszm: I suggest looking at `%VSToolsInstallDir%` which is defined to be `C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023\` on my vs2017 installation and `C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.28.29333\` on vs2019

.github/workflows/ci.yml
lordgamez: Unfortunately I could not make sccache work with cmake and msbuild. I tried setting cmake parameters like the `CMAKE_CXX_COMPILER` and `CMAKE_CXX_COMPILER_LAUNCHER` but sccache was not used. I also tried multiple methods mentioned in https://github.com/mozilla/sccache/issues/107, like setting the msbuild configuration parameters and replacing the cl.exe, but unfortunately non of them seem to work. Replacing the cl.exe with sccache and adjusting the PATH was the only method where I could see the cache calls in the sccache statistics, but all of them were marked as `Non-cacheable calls`.

.github/workflows/ci.yml
lordgamez: Removed the hardcoded versions in [b2f4a71](https://github.com/apache/nifi-minifi-cpp/pull/1012/commits/b2f4a712d7f84c3dec0e35d5146e162bf35def6d)

adamdebreceni:
even though clcache is unmaintained, I am not against using it in the CI as the speed-up is substantial, at worst it could introduce inconsistencies between a fresh build a build using clcache (and in worst case false negatives), but as I understand clcache users voiced no such events/concerns

fgerlits (commented):
I have tested this in my fork.  It works very well on a 2nd CI run in the same branch: https://github.com/fgerlits/nifi-minifi-cpp/actions/runs/729202774 but it does not speed up a compile of the same code in a new branch: https://github.com/fgerlits/nifi-minifi-cpp/actions/runs/729538166.  Is there a way to make the cache work across branches?

lordgamez:
> I have tested this in my fork. It works very well on a 2nd CI run in the same branch: https://github.com/fgerlits/nifi-minifi-cpp/actions/runs/729202774 but it does not speed up a compile of the same code in a new branch: https://github.com/fgerlits/nifi-minifi-cpp/actions/runs/729538166. Is there a way to make the cache work across branches?

Currently all cached builds try use 2 restore-keys for retrieving cache: the one for the current branch and the one for the main branch. It works the same way for the Windows builds. If a new branch is built and no cache is available for that yet, it will use the cache from the main branch. As the change is not yet submitted there is no available cache from the main branch that's why it is slow on the first compilation, but it should be available after the PR is merged.

fgerlits (approved):
OK, that makes sense, thanks!

szaszm:
I wanted to clarify my stance on this issue. I maintain that I'm opposed to introducing CI caching at the expense of a dependency on an unmaintained project, but if there is a majority support for it, I'm not vetoing it.
Based on [this](https://www.apache.org/foundation/voting.html#expressing-votes-1-0-1-and-fractions), I would be a -0.

arpadboda (approved):

PR #1013, opened 2021-02-22
MINIFICPP-1487 Do not trigger the processor if the incoming queue has penalized flow files only

fgerlits:
https://issues.apache.org/jira/browse/MINIFICPP-1487

When checking whether a processor has work to do, so it should be triggered, handle penalized flow files (with unexpired penalty) correctly.  For example, if all the incoming flow files are penalized, the processor should not be triggered.

Before this change, when all incoming flow files were penalized, the processor would get  triggered repeatedly in a tight loop, causing 100% CPU usage (of one core).

A weakness of this solution is that we can't modify the penalty expiration of a flow file after it was added to a queue (Connection).  We currently don't do that, but I couldn't find a simple way to make sure it can never happen in the future.

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/standard-processors/tests/unit/ProcessorTests.cpp
+++ b/extensions/standard-processors/tests/unit/ProcessorTests.cpp
@@ -614,3 +614,66 @@ TEST_CASE("TestRPGWithoutHostInvalidPort", "[TestRPG5]") {
 TEST_CASE("TestRPGValid", "[TestRPG6]") {
   testRPGBypass("", "8080", "8080", false);
 }
+
+TEST_CASE("A Processor detects correctly if it has incoming flow files it can process", "[isWorkAvailable]") {
+  LogTestController::getInstance().setDebug<core::Processor>();
+
+  const auto repo = std::make_shared<TestRepository>();
+  const auto content_repo = std::make_shared<core::repository::VolatileContentRepository>();
+  content_repo->initialize(std::make_shared<minifi::Configure>());
+
+  const std::shared_ptr<core::Processor> processor = std::make_shared<processors::LogAttribute>("test_processor");
+  const auto incoming_connection = std::make_shared<minifi::Connection>(repo, content_repo, "incoming_connection");
+  incoming_connection->addRelationship(core::Relationship{"success", ""});
+  incoming_connection->setDestinationUUID(processor->getUUID());
+  processor->addConnection(incoming_connection);
+  processor->initialize();
+
+  const auto processor_node = std::make_shared<core::ProcessorNode>(processor);
+  const auto context = std::make_shared<core::ProcessContext>(processor_node, nullptr, repo, repo, content_repo);
+  const auto session_factory = std::make_shared<core::ProcessSessionFactory>(context);
+  const auto session = session_factory->createSession();
+
+  SECTION("Initially, the queue is empty, so there is no work available") {
+    REQUIRE_FALSE(processor->isWorkAvailable());
+  }
+
+  SECTION("When a non-penalized flow file is queued, there is work available") {
+    const auto flow_file = session->create();
+    incoming_connection->put(flow_file);
+
+    REQUIRE(processor->isWorkAvailable());
+  }
+
+  SECTION("When a penalized flow file is queued, there is no work available (until the penalty expires)") {
+    const auto flow_file = session->create();
+    session->penalize(flow_file);
+    incoming_connection->put(flow_file);
+
+    REQUIRE_FALSE(processor->isWorkAvailable());
+  }
+
+  SECTION("If there is both a penalized and a non-penalized flow file queued, there is work available") {
+    const auto normal_flow_file = session->create();
+    incoming_connection->put(normal_flow_file);
+
+    const auto penalized_flow_file = session->create();
+    session->penalize(penalized_flow_file);
+    incoming_connection->put(penalized_flow_file);
+
+    REQUIRE(processor->isWorkAvailable());
+  }
+
+  SECTION("When a penalized flow file is queued, there is work available after the penalty expires") {
+    processor->setPenalizationPeriodMsec(10);
+
+    const auto flow_file = session->create();
+    session->penalize(flow_file);
+    incoming_connection->put(flow_file);
+
+    REQUIRE_FALSE(processor->isWorkAvailable());
+    const auto penalty_has_expired = [flow_file] { return !flow_file->isPenalized(); };
+    REQUIRE(utils::verifyEventHappenedInPollTime(std::chrono::seconds{1}, penalty_has_expired, std::chrono::milliseconds{10}));
+    REQUIRE(processor->isWorkAvailable());
+  }
+}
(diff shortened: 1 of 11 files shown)


szaszm (approved):
My only objection is the code duplication in `callOnTrigger`, but as it stems from the duplicate onTrigger overloads, I'm not insisting on a change.

I appreciate this optimization as well as the refactoring of `Processor`. LGTM.

libminifi/src/core/Processor.cpp
```diff
@@ -277,6 +286,28 @@ void Processor::onTrigger(const std::shared_ptr<ProcessContext> &context, const
   }
 }
 
+void Processor::callOnTrigger(const std::shared_ptr<ProcessContext>& context, const std::shared_ptr<ProcessSession>& session) {
+  uint64_t num_incoming_flow_files_queued_before = numIncomingFlowFilesQueued();
+  onTrigger(context, session);  // virtual method overridden by subclasses
+  uint64_t num_incoming_flow_files_queued_after = numIncomingFlowFilesQueued();
+
+  if (num_incoming_flow_files_queued_before != 0 && num_incoming_flow_files_queued_before == num_incoming_flow_files_queued_after) {
```
arpadboda, line 294: As we don't hold lock on the incoming connection during the whole ontrigger execution, flowfiles can be added there in the meanwhile. This condition can be satisfied during normal operation.

libminifi/src/core/Processor.cpp
```diff
@@ -277,6 +286,28 @@ void Processor::onTrigger(const std::shared_ptr<ProcessContext> &context, const
   }
 }
 
+void Processor::callOnTrigger(const std::shared_ptr<ProcessContext>& context, const std::shared_ptr<ProcessSession>& session) {
+  uint64_t num_incoming_flow_files_queued_before = numIncomingFlowFilesQueued();
+  onTrigger(context, session);  // virtual method overridden by subclasses
```
hunyadi-dev, line 291: The lock is unlocked inbetween the calls. Could there be no issue with subsequent calls to this function on multiple threads? Eg. can this case not happen?
1. Count flowfiles before
1. Call onTrigger
1. Add files to incoming connections on a different thread
1. Count flowfiles after

libminifi/src/core/Processor.cpp
```diff
@@ -223,7 +223,18 @@ bool Processor::flowFilesQueued() {
   return false;
 }
 
-bool Processor::flowFilesOutGoingFull() {
+uint64_t Processor::numIncomingFlowFilesQueued() const {
+  std::lock_guard<std::mutex> lock{mutex_};
+
+  uint64_t sum = 0;
+  for (const auto& connectable : _incomingConnections) {
+    auto connection = std::static_pointer_cast<Connection>(connectable);
```
hunyadi-dev, line 231: Does this assume that on a connectable, every incoming connection is actually a `Connection`?  If so, why not change the type of the `_incomingConnections` to a `std::set<std::shared_ptr<Connection>>`?

libminifi/src/core/Processor.cpp
```diff
@@ -277,6 +286,28 @@ void Processor::onTrigger(const std::shared_ptr<ProcessContext> &context, const
...
+  onTrigger(context, session);  // virtual method overridden by subclasses
+  uint64_t num_incoming_flow_files_queued_after = numIncomingFlowFilesQueued();
+
+  if (num_incoming_flow_files_queued_before != 0 && num_incoming_flow_files_queued_before == num_incoming_flow_files_queued_after) {
+    logger_->log_info("Processor %s processed no incoming flow files (out of %" PRIu64 "), so it will yield", name_, num_incoming_flow_files_queued_before);
+    yield();
+  }
+}
+
+void Processor::callOnTrigger(ProcessContext* context, ProcessSession* session) {
```
hunyadi-dev, line 300: Is this different from the other overload? Can the other one not just call to this one?

extensions/standard-processors/tests/unit/ProcessorTests.cpp
```diff
@@ -614,3 +614,109 @@ TEST_CASE("TestRPGWithoutHostInvalidPort", "[TestRPG5]") {
...
+
+  return process_incoming_queue->isYield();
+}
+
+TEST_CASE("If there are no incoming flow files, then there is no automatic yield", "[AutomaticYield]") {
+  SECTION("0 flow files in the queue, we don't try to process any") {
+    auto is_yield = testAutomaticYieldWhenNoIncomingFlowFilesAreProcessed(0, 0);
+    REQUIRE_FALSE(is_yield);
+  }
+  SECTION("0 flow files in the queue, we try to process one") {
```
hunyadi-dev, line 692: This description is a bit misleading as we don't call `Processession::remove` if there is nothing to process. It might be also worth updating this function not to take `nullptr` as its arg.

libminifi/src/core/Processor.cpp
arpadboda: https://github.com/apache/nifi-minifi-cpp/pull/1013#discussion_r580306578

That's what I say, too. :)

libminifi/src/core/Processor.cpp
fgerlits: Yes, good point -- I'll need to rethink this.

libminifi/src/core/Processor.cpp
fgerlits: It does seem like we always store Connections in it, so this is a good idea, although not related to this change.

libminifi/src/core/Processor.cpp
fgerlits: They call different `onTrigger()` overloads.  But I will need a different approach in any case, so this is moot.

extensions/standard-processors/tests/unit/ProcessorTests.cpp
fgerlits: We call `session->get()` once, which is what I meant by "try to process one".

Which function do you mean?  Probably best to discuss offline.

libminifi/src/core/Processor.cpp
fgerlits: I have rewritten the code to use a different approach: now `isWorkAvailable()` returns false when all the incoming flow files are penalized.  This will cause the processor to yield for a `nifi.bored.yield.duration` amount of time, so it will no longer spin in a tight loop.

arpadboda (approved):
Nicely simplified logic in ProcessSession.

libminifi/src/Connection.cpp
```diff
@@ -227,8 +212,7 @@ void Connection::drain(bool delete_permanently) {
   std::lock_guard<std::mutex> lock(mutex_);
```
hunyadi-dev, line 212: Might be not related to this PR, but are we happy just locking here? Can we not resume a blocked insertion after the queue is supposedly drained?

hunyadi-dev (approved):
Good PR, the change looks robust and well tested.

libminifi/src/Connection.cpp
fgerlits: I don't think we can (or want to) stop new flow files from being added to the connection queue after a `drain()`, and I would say a `put()` which resumes after it has acquired the lock happens after the `drain()`, even if it started before.

But yes, I agree it's not related to this change. :)

PR #1015, opened 2021-02-24
MINIFICPP-1509 - Use _stat64 on windows to handle large files

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/standard-processors/processors/GetFile.cpp
+++ b/extensions/standard-processors/processors/GetFile.cpp
@@ -239,14 +239,15 @@ bool GetFile::acceptFile(std::string fullName, std::string name, const GetFileRe
     return false;
   }
 #endif
+  uint64_t file_size = gsl::narrow<uint64_t>(statbuf.st_size);
+  uint64_t modifiedTime = gsl::narrow<uint64_t>(statbuf.st_mtime) * 1000;
 
-  if (request.minSize > 0 && statbuf.st_size < request.minSize)
+  if (request.minSize > 0 && file_size < request.minSize)
     return false;
 
-  if (request.maxSize > 0 && statbuf.st_size > request.maxSize)
+  if (request.maxSize > 0 && file_size > request.maxSize)
     return false;
 
-  uint64_t modifiedTime = ((uint64_t) (statbuf.st_mtime) * 1000);
   uint64_t fileAge = utils::timeutils::getTimeMillis() - modifiedTime;
   if (request.minAge > 0 && fileAge < request.minAge)
     return false;
@@ -267,7 +268,7 @@ bool GetFile::acceptFile(std::string fullName, std::string name, const GetFileRe
     return false;
   }
 
-  metrics_->input_bytes_ += statbuf.st_size;
+  metrics_->input_bytes_ += file_size;
   metrics_->accepted_files_++;
   return true;
 }
--- a/extensions/standard-processors/processors/GetFile.cpp
+++ b/extensions/standard-processors/processors/GetFile.cpp
@@ -228,42 +228,48 @@ void GetFile::pollListing(std::queue<std::string> &list, const GetFileRequest &r
 bool GetFile::acceptFile(std::string fullName, std::string name, const GetFileRequest &request) {
   logger_->log_trace("Checking file: %s", fullName);
 
+#ifdef WIN32
+  struct _stat64 statbuf;
+  if (_stat64(fullName.c_str(), &statbuf) != 0) {
+    return false;
+  }
+#else
   struct stat statbuf;
+  if (stat(fullName.c_str(), &statbuf) != 0) {
+    return false;
+  }
+#endif
 
-  if (stat(fullName.c_str(), &statbuf) == 0) {
-    if (request.minSize > 0 && statbuf.st_size < (int32_t) request.minSize)
-      return false;
+  if (request.minSize > 0 && statbuf.st_size < request.minSize)
+    return false;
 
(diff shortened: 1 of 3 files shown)


adamdebreceni:
Note: eventually we shouldn't include `<sys/stat.h>` but provide a platform-independent abstraction for it, `FileUtils.h` could be that abstraction but since some functions do not unambiguously signal a non-existent file (e.g. `file_size` returns 0 if the stat fails) currently it is not entirely viable to replace the explicit `stat` in `GetFile` or `PutFile` and possible other places

extensions/standard-processors/processors/GetFile.cpp
```diff
@@ -228,42 +228,48 @@ void GetFile::pollListing(std::queue<std::string> &list, const GetFileRequest &r
...
+  struct _stat64 statbuf;
+  if (_stat64(fullName.c_str(), &statbuf) != 0) {
+    return false;
+  }
+#else
   struct stat statbuf;
+  if (stat(fullName.c_str(), &statbuf) != 0) {
+    return false;
+  }
+#endif
```
adamdebreceni, line 241: we could use `utils::file::exists` and then call `utils::file::file_size` and `utils::file::last_write_time` separately, my concern with this, is that we took one (supposedly) atomic `stat`-call and made three separate calls, on the other side, that would limit the number of places `stat` is explicitly called from, moreover if we trace where this `acceptFile` is called from (`file::list_dir`) we could even ignore the possibility of a non-existing file

1. call `stat` explicitly (current)
2. use 3 calls (exists, file_size, last_write_time)
3. use 2 calls (file_size, last_write_time) because we don't care about non-existing files because a) this method is called with existing files b) we will handle them anyway downstream when we actually read it
4. something else (make file_size, last_write_time return optionals, etc )

libminifi/include/utils/file/FileUtils.h
```diff
@@ -248,18 +248,16 @@ inline uint64_t last_write_time(const std::string &path) {
   if (ec.value() == 0) {
     return result;
   }
-#else
-#ifdef WIN32
-  struct _stat result;
-  if (_stat(path.c_str(), &result) == 0) {
+#elif WIN32
```
fgerlits, line 251: I think this should be `#elif defined(WIN32)`

libminifi/include/utils/file/FileUtils.h
```diff
@@ -334,20 +332,28 @@ inline bool get_uid_gid(const std::string &path, uint64_t &uid, uint64_t &gid) {
...
+  }
+  return S_ISDIR(dir_stat.st_mode) != 0;
+#endif
 }
 
 inline bool exists(const std::string& path) {
 #ifdef USE_BOOST
   return boost::filesystem::exists(path);
 #else
 #ifdef WIN32
```
lordgamez, line 354: You could merge these as well, like you did in `last_write_time`

adamdebreceni: done

extensions/standard-processors/processors/GetFile.cpp
```diff
@@ -228,42 +228,48 @@ void GetFile::pollListing(std::queue<std::string> &list, const GetFileRequest &r
...
 
-  if (stat(fullName.c_str(), &statbuf) == 0) {
-    if (request.minSize > 0 && statbuf.st_size < (int32_t) request.minSize)
-      return false;
-
-    if (request.maxSize > 0 && statbuf.st_size > (int32_t) request.maxSize)
-      return false;
+  if (request.minSize > 0 && statbuf.st_size < request.minSize)
```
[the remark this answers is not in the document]
adamdebreceni: added gsl::narrow casts

libminifi/include/utils/file/FileUtils.h
adamdebreceni: done

extensions/standard-processors/processors/GetFile.cpp
```diff
@@ -228,42 +228,49 @@ void GetFile::pollListing(std::queue<std::string> &list, const GetFileRequest &r
...
+    return false;
+  }
+#else
   struct stat statbuf;
+  if (stat(fullName.c_str(), &statbuf) != 0) {
+    return false;
+  }
+#endif
+  uint64_t file_size = gsl::narrow<uint64_t>(statbuf.st_size);
+  uint64_t modifiedTime = gsl::narrow<uint64_t>(statbuf.st_mtime) * 1000;
```
hunyadi-dev, line 243: Is this safe (both for the narrowing before scaling and the type usage)? Shouldn't we use `std::difftime`?

hunyadi-dev (approved):

extensions/standard-processors/processors/GetFile.cpp
adamdebreceni: you are right about our inconsistent time representations (we use `uint64_t`, `int64_t` and `time_t` in `TimeUtil.h` to represent time) we should unify it, based on the POSIX and Windows documentations, as I understand we may (cautiously) assume that `st_mtime` yields the Unix time, so these narrowings and the subsequent scaling should be fine

fgerlits (approved):

lordgamez (approved):
Thanks for the fixes, LGTM

arpadboda (approved):

extensions/standard-processors/processors/GetFile.cpp
szaszm: I'd stay with 1

szaszm (approved):

PR #1017, opened 2021-02-26
MINIFICPP-1515 - Add integration tests testing different flowfile sizes in a simple flow

hunyadi-dev:
Issues with big flowfiles were reported on [stack overflow](https://stackoverflow.com/questions/66330866/minifi-getfile-processor-fails-to-get-large-files/66334615?noredirect=1#comment117275399_66334615).

Seems like this is related to twi issues, one a narrowing exception happens when trying to determine the length of the file to be written into the content repository, and another is that we use `_stat` on windows even though we should be using `_stat64` to query files larger than 2GB.

In order to avoid these issues to reappear, we should add some integration test coverage.

This PR will be in draft status until the mentioned issues are fixed. Also, currently integration test logs are bloated for scenario outline based testing, this is to be fixed as [part of this task](https://issues.apache.org/jira/browse/MINIFICPP-1515).

### DIFF
--- a/docker/test/integration/minifi/validators/SingleFileContentHashValidator.py
+++ b/docker/test/integration/minifi/validators/SingleFileContentHashValidator.py
@@ -27,17 +27,16 @@ class SingleFileContentHashValidator(FileOutputValidator):
 
         listing = listdir(full_dir)
         if listing:
-            for l in listing:
-                logging.info("name:: %s", l)
-            out_file_name = listing[0]
-            logging.info("dir %s -- name %s", full_dir, out_file_name)
-            full_path = join(full_dir, out_file_name)
-            if not os.path.isfile(full_path):
-                return self.valid
-            
-            actual_md5_hash = md5(full_path)
-            logging.info("expected hash: %s -- actual: %s", self.expected_md5_hash, actual_md5_hash)
-            if self.expected_md5_hash == actual_md5_hash:
-                self.valid = True
-
+            all_matches = True
+            for out_file_name in listing:
+                logging.info("dir %s -- name %s", full_dir, out_file_name)
+                full_path = join(full_dir, out_file_name)
+                if not os.path.isfile(full_path):
+                    all_matches = False
+                    break
+                actual_md5_hash = md5(full_path)
+                logging.info("expected hash: %s -- actual: %s", self.expected_md5_hash, actual_md5_hash)
+                if self.expected_md5_hash != actual_md5_hash:
+                    all_matches = False
+            self.valid = all_matches
         return self.valid
--- a/docker/test/integration/minifi/validators/SingleFileContentHashValidator.py
+++ b/docker/test/integration/minifi/validators/SingleFileContentHashValidator.py
@@ -1,0 +1,43 @@
+import logging
+import os
+
+from os import listdir
+from os.path import join
+
+from .FileOutputValidator import FileOutputValidator
+from ..core.HashUtils import md5
+
+class SingleFileContentHashValidator(FileOutputValidator):
+    """
+    Validates the content of a single file in the given directory.
+    """
+
+    def __init__(self, expected_md5_hash, subdir=''):
+        self.valid = False
+        self.expected_md5_hash = expected_md5_hash
+        self.subdir = subdir
+
+    def validate(self):
+        self.valid = False
+        full_dir = os.path.join(self.output_dir, self.subdir)
+        logging.info("Output folder: %s", full_dir)
+
+        if not os.path.isdir(full_dir):
+            return self.valid
+
+        listing = listdir(full_dir)
+        if listing:
+            for l in listing:
+                logging.info("name:: %s", l)
+            out_file_name = listing[0]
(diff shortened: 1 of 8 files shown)


docker/test/integration/minifi/validators/SingleFileContentHashValidator.py
```diff
@@ -0,0 +1,43 @@
...
+
+        listing = listdir(full_dir)
+        if listing:
+            for l in listing:
+                logging.info("name:: %s", l)
+            out_file_name = listing[0]
+            logging.info("dir %s -- name %s", full_dir, out_file_name)
+            full_path = join(full_dir, out_file_name)
+            if not os.path.isfile(full_path):
+                return self.valid
```
fgerlits, line 36: Why not `continue` instead of `return`?  If 'l' is a subdirectory in `output_dir`, we want to keep looking for matching files.

docker/test/integration/minifi/validators/SingleFileContentHashValidator.py
```diff
@@ -0,0 +1,43 @@
...
+        self.valid = False
+        full_dir = os.path.join(self.output_dir, self.subdir)
+        logging.info("Output folder: %s", full_dir)
+
+        if not os.path.isdir(full_dir):
+            return self.valid
+
+        listing = listdir(full_dir)
+        if listing:
+            for l in listing:
```
fgerlits, line 30: I'm not a fan of one-char variable names, especially `l` (and `O` etc) -- this could be `file_name`, for example

fgerlits (commented):
Looks good.  Please rebase it on top of the current main and resolve the conflicts.

docker/test/integration/minifi/validators/SingleFileContentHashValidator.py
hunyadi-dev: Unfortunately, the flow name is randomly generated by `GetFile`. I would also like to have a failure if there are multiple files, but only one matches the criteria so I went with a full loop.

docker/test/integration/minifi/validators/SingleFileContentHashValidator.py
hunyadi-dev: Updated

hunyadi-dev:
The check still shows that **we do not handle the file size 2.1 GiB properly**. Merge once it is fixed.

fgerlits:
> The check still shows that we do not handle the file size 2.1 GiB properly. Merge only once it is fixed.

Right -- #1028 needs to be merged first, and probably its successor (which will handle `OutputStream::write`), too.

szaszm:
> Right -- #1028 needs to be merged first, and probably its successor (which will handle `OutputStream::write`), too.

I'm close to being done with the write refactor. I'll submit a PR next week against the read refactor branch (i.e. #1028) to keep the size managable, and both can be merged together after they're reviewed and approved.

fgerlits:
This still fails after merging #1028 and #1083.  The main problem is that `SingleFileContentHashValidator` reports a hash mismatch on the 2.1 GiB test file:
```
      INFO:root:Output file created: /tmp/.nifi-test-output.fe59da51-9bb0-4c9f-9966-876ec2ea543b/.4c3b8487-5767-4b29-830e-7e83d4e3e5ec.053203f8-ce84-11eb-82db-0242c0a81002
      INFO:root:Output folder: /tmp/.nifi-test-output.fe59da51-9bb0-4c9f-9966-876ec2ea543b/
      INFO:root:dir /tmp/.nifi-test-output.fe59da51-9bb0-4c9f-9966-876ec2ea543b/ -- name .4c3b8487-5767-4b29-830e-7e83d4e3e5ec.053203f8-ce84-11eb-82db-0242c0a81002
      INFO:root:expected hash: 3531295d2dc5e038e4f4e684d020167b -- actual: 2073b6f0bc607551c97ed79b4e0de87e
```
I suspect this is problem with the test and not with GetFile/PutFile, but it needs to be investigated.

Another, minor issue is that the 20 seconds timeout needs to be increased, as running the test with the 2.1 GiB test file takes around 40 seconds on my computer.

Also, `OutputEventHandler` prints "Output file modified" more than 2 million times, which is annoying.

fgerlits:
@hunyadi-dev do you want to fix this?  I can do it if you don't feel like it.

szaszm:
After rebasing to main, the tests seem to be running successfully, thanks to #1131. I think we can merge this after a successful CI run, I've rebased it to the latest main.

lordgamez:
> This test passes for me locally after rebasing on top of main, so I think it's ready for merging.
> 
> In order to not increase the running time of the CI job by too much, I suggest making the timeout configurable and lower, eg. 1 second for the first 3 tests and 10 seconds for the last two.

Rebased and fixed the runtime issues (the file observer wanted to log the content of the file on every modification which prevented to get the actual notifications in time when large files flooded the logger), I think this could be merged if all tests pass.

lordgamez (approved):

szaszm:
closing due to inactivity

PR #1018, opened 2021-02-26
MINIFICPP-1513 - Increase timeouts for behave tests involving http proxies, clean up logging

hunyadi-dev:
Currently python based the integration tests instantiate separate loggers. This makes it difficult to capture logging when multiple test are run with the integration tests (eg. single Scenario Outline with multiple examples).

Also, tests involving http proxies seem to occasionally time out/fail on Github Actions. The timeout issue is mitigated by increasing the execution time limit. The failure is related to improperly recognized `stdout` encoding (see example below, this is not linked to using behave). In case of encoding is not recognized `utf8` will used as fallback value.

image

### DIFF
--- a/docker/test/integration/MiNiFi_integration_test_driver.py
+++ b/docker/test/integration/MiNiFi_integration_test_driver.py
@@ -22,7 +22,6 @@ from minifi.validators.SingleFileOutputValidator import SingleFileOutputValidato
 
 class MiNiFi_integration_test():
     def __init__(self, context):
-        logging.info("MiNiFi_integration_test init")
         self.test_id = str(uuid.uuid4())
         self.clusters = {}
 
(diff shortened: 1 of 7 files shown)


docker/test/integration/MiNiFi_integration_test_driver.py
```diff
@@ -22,7 +22,6 @@
 
 class MiNiFi_integration_test():
     def __init__(self, context):
-        logging.info("MiNiFi_integration_test init")
```
hunyadi-dev, line 25: Logging before  `before_all` is triggered is removed, because that seems to use the previous `logging` instantiation.

lordgamez (approved):

arpadboda (approved):

PR #1019, opened 2021-03-01
MINIFICPP-1517 - Fix script engine integration test log cluttering

hunyadi-dev:
Currently integration tests used to produce errors like these logged due to `ScriptEngine` issues:
```
[warning] Cannot load SentimentAnalysis because of ModuleNotFoundError: No module named 'vaderSentiment'
```

For the time of building the integration test docker image, we should disable scripting. This is to be reenabled when the `ScriptEngine` issue is fixed or when there is an explicit need for it in the integration tests.


arpadboda (approved):
LGTM! Let's wait CI to complete before merging!

docker/Dockerfile
```diff
@@ -115,6 +115,7 @@ ARG DISABLE_LIBARCHIVE
 ARG DISABLE_LZMA
 ARG DISABLE_BZIP2
 ARG DISABLE_SCRIPTING
+ARG DISABLE_PYTHON_SCRIPTING
```
lordgamez, line 118: DockerBuild.sh and DockerConfig.cmake should be updated as well to pass this parameter

hunyadi-dev: Updated.

lordgamez (approved):

szaszm (approved):

PR #754, opened 2020-04-14
MINIFICPP-1151 fix most 1st party compiler warnings

szaszm:
Remaining:
 - MSVC: multiple warning LNK4221: This object file does not define any previously undefined public symbols, so it will not be used by any link operation that consumes this library
 - MSVC/expression language: multiple warning C4566: character represented by universal-character-name '\u????' cannot be represented in the current code page (1252)
 - clang/expression language yacc: warning: fix-its can be applied.  Rerun with option '--update'. [-Wother]

Squashed commit of the following:

commit 98a04a4c83bef712b04fce151d98414a142d6ee4
Author: Marton Szasz <szaszm01@gmail.com>
Date:   Tue Apr 14 15:00:54 2020 +0200

    MINIFICPP-1151 windows (except expression language)

commit d58e7342f2c5ebbfa6e4c664666b6f0f81355489
Author: Marton Szasz <szaszm01@gmail.com>
Date:   Tue Apr 7 10:18:26 2020 +0200

    MINIFICPP-1151 linux, clang 10

    note: rocksdb compilation correctly fails with clang 10. We will need to
    upgrade or patch it.

commit 995d6f7031947e4fc6e4370a2b04eba998d84ada
Author: Marton Szasz <szaszm01@gmail.com>
Date:   Mon Feb 10 21:35:29 2020 +0100

    verified on clang6/linux

    There are some warnings due to overriding a ROCKSDB_ASSERT macro to
    throw and it's expanded inside `noexcept` functions. No idea how to fix
    that.

commit b20761da23fde4b108c5bda736c1f14cc822e652
Author: Marton Szasz <szaszm01@gmail.com>
Date:   Mon Feb 10 18:05:55 2020 +0100

    fix compilation errors

commit acbebcbbcef05a294e8a7cecb2da8e0202ffb53a
Author: Marton Szasz <szaszm01@gmail.com>
Date:   Mon Feb 10 15:58:05 2020 +0100

    MINIFICPP-1151 round #1: GCC 7 on ubuntu, default config

Thank you for submitting a contribution to Apache NiFi - MiNiFi C++.

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [x] Is there a JIRA ticket associated with this PR? Is it referenced
     in the commit message?

- [x] Does your PR title start with MINIFICPP-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.

[...]

PR #1020, opened 2021-03-01
MINIFICPP-1352 Enable -Wall and -Wextra and resolve related warnings

lordgamez:
This change:
- Adds Wall and Wextra to Mac and Linux builds
- Adds FAIL_ON_WARNINGS CMake flags to enable compilation errors on warnings
- Resolves current warnings with gcc and clang compilers
- Adds 20.04 clang build with all extensions to the CI workflow
- Enables FAIL_ON_WARNINGS flag in Mac and multiple Linux builds

Note: It may be easier to review on a per-commit basis as most commits focus on a single warning type

### For all changes:
     in the commit message?

### For code changes:

### Note:


CMakeLists.txt
```diff
@@ -315,11 +313,28 @@ target_compile_definitions(gsl-lite INTERFACE ${GslDefinitions})
...
+# Setup passthrough args used by third parties
+set(PASSTHROUGH_CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+set(PASSTHROUGH_CMAKE_C_FLAGS ${CMAKE_C_FLAGS})
+
+# Setup warning flags
+if(MSVC)
+  if(FAIL_ON_WARNINGS)
+		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /WX")
+  endif()
+else()
```
hunyadi-dev, line 330: Minor, but omitting the `FAIL_ON_WARNINGS_FLAG` variable would probably be more readable:
```cmake
else()
  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra")
  if(FAIL_ON_WARNINGS)
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror")
  endif()	
endif()
```

extensions/coap/nanofi/coap_functions.c
```diff
@@ -115,6 +121,7 @@ int create_endpoint_context(coap_context_t **ctx, const char *node, const char *
 }
 
 struct coap_pdu_t *create_request(struct coap_context_t *ctx, struct coap_session_t *session, coap_optlist_t **optlist, unsigned char code, coap_str_const_t *ptr) {
+  UNUSED(ctx);
```
hunyadi-dev, line 124: If you don't use it, I think its better not to name it instead.

extensions/coap/tests/CoapIntegrationBase.h
```diff
@@ -45,6 +45,7 @@ class CoapIntegrationBase : public IntegrationBase {
   }
 
   void run(const utils::optional<std::string>& test_file_location = {}, const utils::optional<std::string>& bootstrap_file = {}) override {
+    (void)bootstrap_file;  // against unused variable warnings
```
hunyadi-dev, line 48: Why don't you leavev the argument unnamed instead?

extensions/librdkafka/PublishKafka.cpp
```diff
@@ -407,7 +407,7 @@ class ReadCallback : public InputStreamCallback {
       return 0;
     }
 
-    for (size_t segment_num = 0; read_size_ < flow_size_; ++segment_num) {
+    for (size_t segment_num = 0; gsl::narrow<uint32_t>(read_size_) < flow_size_; ++segment_num) {
```
hunyadi-dev, line 410: Can we not actually change the `read_size_` type?

extensions/librdkafka/rdkafka_utils.h
```diff
@@ -47,7 +47,7 @@ struct rd_kafka_conf_deleter {
 
 struct rd_kafka_producer_deleter {
   void operator()(rd_kafka_t* ptr) const noexcept {
-    rd_kafka_resp_err_t flush_ret = rd_kafka_flush(ptr, 10000 /* ms */);  // Matching the wait time of KafkaConnection.cpp
+    // rd_kafka_resp_err_t flush_ret = rd_kafka_flush(ptr, 10000 /* ms */);  // Matching the wait time of KafkaConnection.cpp
```
hunyadi-dev, line 50: Why did you comment this out?
This is part of the proper termination sequence:
https://github.com/edenhill/librdkafka/blob/master/INTRODUCTION.md#termination

extensions/coap/nanofi/coap_functions.c
lordgamez: Unfortunately this part is C code, and C99 standard does not allow parameters without names, otherwise you get a `parameter name omitted` error:

[6.9.1.5] If the declarator includes a parameter type list, the declaration of each parameter shall include an identifier, except for the special case of a parameter list consisting of a single parameter of type void, in which case there shall not be an identifier. No declaration list shall follow.

extensions/librdkafka/PublishKafka.cpp
lordgamez: I was wondering the same, but I couldn't decide if the value can go below zero, but I suppose it shouldn't. I will change the type.

extensions/librdkafka/rdkafka_utils.h
lordgamez: My mistake, the problem was the unused parameter and I commented it out without realizing that we need the flush call. Thanks for the info.

extensions/pcap/CapturePacket.h
```diff
@@ -167,7 +167,7 @@ class CapturePacket : public core::Processor {
 
 REGISTER_RESOURCE(CapturePacket, "CapturePacket captures and writes one or more packets into a PCAP file that will be used as the content of a flow file."
     " Configuration options exist to adjust the batching of PCAP files. PCAP batching will place a single PCAP into a flow file. "
-    "A regular expression selects network interfaces. Bluetooth network interfaces can be selected through a separate option.")
+    "A regular expression selects network interfaces. Bluetooth network interfaces can be selected through a separate option.");
```
hunyadi-dev, line 170: How did this work beforehand? Do we even compile this?

extensions/script/python/ExecutePythonProcessor.h
```diff
@@ -34,6 +34,8 @@
 #include "PythonScriptEngine.h"
 #include "core/Property.h"
 
+#pragma GCC visibility push(hidden)
```
hunyadi-dev, line 37: Does this solve the `declared with greater visibility than the type of its field` error? If so, how does this work?
I found these refs: [[1]](https://github.com/vgc/vgc/issues/11), [[2]](https://stackoverflow.com/questions/49252686/exported-class-with-non-exported-data-member?newreg=b2adafbe461c42faa7888ff23b6b05c3) on the issue; how does this relate?

extensions/sensors/GetEnvironmentalSensors.h
```diff
@@ -75,7 +75,7 @@ class GetEnvironmentalSensors : public SensorBase {
   static std::shared_ptr<utils::IdGenerator> id_generator_;
 };
 
-REGISTER_RESOURCE(GetEnvironmentalSensors, "Provides sensor information from known sensors to include humidity and pressure data")
+REGISTER_RESOURCE(GetEnvironmentalSensors, "Provides sensor information from known sensors to include humidity and pressure data");
```
hunyadi-dev, line 78: Again, why did this work, if the semicolon is required?

extensions/pcap/CapturePacket.h
lordgamez: I changed REGISTER_RESOURCE, removing the semicolon from the macro, so that the semicolon is required now. Previously we sometimes put the semicolon there, sometimes we did not, causing additional semicolon warnings.

extensions/sensors/GetEnvironmentalSensors.h
lordgamez: See the answer above. SMART_ENUM has the semicolon as part of the macro, and it is used consistently.

extensions/script/python/ExecutePythonProcessor.h
lordgamez: Yes that's the error and those refs are related. There was a similar fix by @szaszm previously in #754, he also mentioned that the fix is due to pybind11 requiring `-fvisibility=hidden`. This solution is now applied in a wider scope by including the classes using PyProcessSession.

extensions/standard-processors/processors/GenerateFlowFile.cpp
```diff
@@ -87,7 +87,7 @@ void generateData(std::vector<char>& data, bool textData = false) {
     std::uniform_int_distribution<> distr(0, index_of_last_char);
     auto rand = std::bind(distr, eng);
     std::generate_n(data.begin(), data.size(), rand);
-    std::for_each(data.begin(), data.end(), [](char & c) { c = TEXT_CHARS[c];});
+    std::for_each(data.begin(), data.end(), [](char & c) { c = TEXT_CHARS[static_cast<uint8_t>(c)];});
```
hunyadi-dev, line 90: Minor, but this could be moved to `rand`.

CMakeLists.txt
lordgamez: Updated in [75e1aa](https://github.com/apache/nifi-minifi-cpp/pull/1020/commits/75e1aa33766b2b43b135ebfd2aac312e796d463a)

extensions/coap/tests/CoapIntegrationBase.h
lordgamez: I think I did not realize you can leave the argument unnamed when it has a default value. Updated in [75e1aa](https://github.com/apache/nifi-minifi-cpp/pull/1020/commits/75e1aa33766b2b43b135ebfd2aac312e796d463a)

extensions/librdkafka/PublishKafka.cpp
lordgamez: Updated in [75e1aa](https://github.com/apache/nifi-minifi-cpp/pull/1020/commits/75e1aa33766b2b43b135ebfd2aac312e796d463a)

extensions/librdkafka/rdkafka_utils.h
lordgamez: Updated in [75e1aa](https://github.com/apache/nifi-minifi-cpp/pull/1020/commits/75e1aa33766b2b43b135ebfd2aac312e796d463a)

extensions/standard-processors/processors/GenerateFlowFile.cpp
lordgamez: Sorry I'm not sure what do you mean here. Could you give a code example/suggestion?

extensions/standard-processors/tests/integration/TLSServerSocketSupportedProtocolsTest.cpp
```diff
@@ -151,10 +151,12 @@ class SimpleSSLTestClient  {
   gsl::not_null<std::shared_ptr<logging::Logger>> logger_{gsl::make_not_null(logging::LoggerFactory<SimpleSSLTestClient>::getLogger())};
 
   static SocketDescriptor openConnection(const char *host_name, const char *port, logging::Logger& logger) {
-    struct addrinfo hints = {0}, *addrs;
+    struct addrinfo hints;
+    hints.ai_flags = 0;
```
hunyadi-dev: Is this the same behaviour? What was the issue with the `{0}` initialization?
https://man7.org/linux/man-pages/man3/getaddrinfo.3.html

lordgamez: You are right it wasn't the same behavior and it failed on VS2019 build. I fixed it in [a89e77](https://github.com/apache/nifi-minifi-cpp/pull/1020/commits/a89e77484a6ea3651a2952a6a9236ea0e7027e17). The warning I got from this is that it did not initialize all members with the `{0}` initialization.

libminifi/src/utils/StringUtils.cpp
```diff
@@ -305,8 +305,10 @@ bool StringUtils::from_base64(uint8_t* data, size_t* data_length, const char* ba
     case 1:
       return false;
     case 2:
-      digits[2] = 0x00;
     case 3: {
+      if (digit_counter == 2) {
```
hunyadi-dev, line 309: You probably did not mean to do this.

nanofi/examples/monitor_directory.c
```diff
@@ -75,11 +77,11 @@ int main(int argc, char **argv) {
 
   port.port_id = portStr;
 
-  C2_Server server;
-  server.url = argv[4];
-  server.ack_url = argv[5];
-  server.identifier = "monitor_directory";
-  server.type = REST;
+  // C2_Server server;
```
hunyadi-dev, line 80: This is probably leftover commented-out code.

nanofi/include/sitetosite/CPeer.h
```diff
@@ -74,8 +74,8 @@ static void setHostName(struct SiteToSiteCPeer * peer, const char * hostname) {
   peer->_url = (char*)malloc(host_len + 14);  // +1 for trailing zero, 1 for ':', at most 5 for port, 7 for "nifi://" suffix
   memset(peer->_url, 0, host_len + 14);  // make sure to have zero padding no matter the length of the port
   strncpy(peer->_host, hostname, host_len);
-  strncpy(peer->_url, "nifi://", 7);
-  strncpy(peer->_url + 7, hostname, host_len);
+  memcpy(peer->_url, "nifi://", 7);
```
hunyadi-dev, line 77: Do you not just mask out the reported mistake of not copying `\0`?
```bash
nifi://
0123456 --> 7 characters copied, missing \0.
```

nanofi/include/sitetosite/CRawSocketProtocol.h
```diff
@@ -76,23 +76,15 @@ int16_t sendPacket(struct CRawSiteToSiteClient * client, const char * transactio
 
 CTransaction* createTransaction(struct CRawSiteToSiteClient * client, TransferDirection direction);
 
-static const char * getResourceName() {
+static inline const char * getResourceName() {
```
hunyadi-dev, line 79: Why all the explicit inlining?

hunyadi-dev (commented):
Went through all the long list of changes, looks mostly good, but added some questions/recommendations. I suggest splitting up the PR into smaller parts next time.

nanofi/include/sitetosite/CRawSocketProtocol.h
lordgamez: According to @arpadboda the static function definitions in the headers were introduced to bypass link issues, not an actual solution to the problem. They cause `-Wunused-function` as they should be referenced in every single translation unit where they are included. The original link issues should be resolved in the future, but as this codebase is abandoned at the moment, the easiest solution was to mark the static functions inline, which bypasses the warning.

nanofi/include/sitetosite/CPeer.h
lordgamez: Updated in 828f564566b9e7fa50bfe0f37bcbf0996b22de67

nanofi/examples/monitor_directory.c
lordgamez: Removed commented code in ee40d7658735b8ebe61f88956dfa6581da3cb863

libminifi/src/utils/StringUtils.cpp
lordgamez: After further checking the issue clang does not include the implicit-fallthrough warning in Wall and Wextra flags, so it is only GCC specific issue. To be backward compatible with older GCC versions I used `// fall through` marker comment.  Updated in 9b337b70b01441aea445a5d31a75c88c7ab6f680.

extensions/standard-processors/processors/GenerateFlowFile.cpp
lordgamez: Updated in 2d4d00f390e68a1128e5e9964b988c76fe3c4505

extensions/openwsman/processors/SourceInitiatedSubscriptionListener.cpp
```diff
@@ -249,11 +249,11 @@ bool SourceInitiatedSubscriptionListener::loadState() {
 
 std::string SourceInitiatedSubscriptionListener::Handler::millisecondsToXsdDuration(int64_t milliseconds) {
   char buf[1024];
-  snprintf(buf, sizeof(buf), "PT%lld.%03lldS", milliseconds / 1000, milliseconds % 1000);
+  snprintf(buf, sizeof(buf), "PT%ld.%03ldS", milliseconds / 1000, milliseconds % 1000);
```
szaszm, line 252: Both the old and the new version are incorrect. The format specifier of `int64_t` is the `PRId64` macro from `<cinttypes>`.
```suggestion
  snprintf(buf, sizeof(buf), "PT%" PRId64 ".%03" PRId64 "S", milliseconds / 1000, milliseconds % 1000);
```

extensions/rocksdb-repos/RocksDatabase.h
```diff
@@ -84,6 +84,7 @@ class RocksDatabase {
   };
 
   RocksDatabase(const rocksdb::Options& options, const std::string& name, Mode mode = Mode::ReadWrite);
+  virtual ~RocksDatabase() = default;
```
szaszm, line 87: Since there are no derived classes of `RocksDatabase`, I would rather make all members non-virtual.

extensions/script/lua/LuaBaseStream.cpp
```diff
@@ -54,15 +54,15 @@ std::string LuaBaseStream::read(size_t len) {
   // http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2012/n3337.pdf
   auto read = stream_->read(reinterpret_cast<uint8_t *>(&buffer[0]), static_cast<int>(len));
 
-  if (read != len) {
-    buffer.resize(static_cast<size_t >(read));
+  if (gsl::narrow<size_t>(read) != len) {
+    buffer.resize(gsl::narrow<size_t>(read));
```
szaszm, line 58: Unsafe narrowing: `read` can be `-1` in case of an error.

extensions/script/lua/LuaBaseStream.cpp
```diff
@@ -54,15 +54,15 @@ std::string LuaBaseStream::read(size_t len) {
...
   }
 
-  return std::move(buffer);
+  return buffer;
 }
 
 size_t LuaBaseStream::write(std::string buf) {
-  return static_cast<size_t>(stream_->write(reinterpret_cast<uint8_t *>(const_cast<char *>(buf.data())),
+  return gsl::narrow<size_t>(stream_->write(reinterpret_cast<uint8_t *>(const_cast<char *>(buf.data())),
```
szaszm, line 65: same here

extensions/script/python/PythonScriptEngine.cpp
```diff
@@ -33,7 +33,7 @@ Interpreter *getInterpreter() {
 }
 
 PythonScriptEngine::PythonScriptEngine() {
-  auto intepreter = getInterpreter();
+  getInterpreter();
```
szaszm, line 36: There doesn't seem to be any magic global initialization happening here, so I suggest removing the calls altogether.

extensions/standard-processors/processors/ListenSyslog.cpp
```diff
@@ -187,7 +187,7 @@ void ListenSyslog::runThread() {
         struct sockaddr_in cli_addr;
         clilen = sizeof(cli_addr);
         int recvlen = recvfrom(_serverSocket, _buffer, sizeof(_buffer), 0, (struct sockaddr *) &cli_addr, &clilen);
-        if (recvlen > 0 && (uint64_t) (recvlen + getEventQueueByteSize()) <= _recvBufSize) {
+        if (recvlen > 0 && (uint64_t) (recvlen + getEventQueueByteSize()) <= gsl::narrow<uint64_t>(_recvBufSize)) {
```
szaszm, line 190: Unsafe narrowing. `_recvBufSize` is parsed from a property, meaning it has to be validated. `gsl::narrow` should be used when you "promise" the value will fit in the destination type, or otherwise you're willing to crash the program.

extensions/standard-processors/processors/ListenSyslog.cpp
```diff
@@ -204,7 +204,7 @@ void ListenSyslog::runThread() {
           logger_->log_debug("ListenSysLog client socket %d close", clientSocket);
           it = _clientSockets.erase(it);
         } else {
-          if ((uint64_t) (recvlen + getEventQueueByteSize()) <= _recvBufSize) {
+          if ((uint64_t) (recvlen + getEventQueueByteSize()) <= gsl::narrow<uint64_t>(_recvBufSize)) {
```
szaszm, line 207: same as above

extensions/standard-processors/processors/ListenSyslog.h
```diff
@@ -191,7 +191,7 @@ class ListenSyslog : public core::Processor {
   void pollEvent(std::queue<SysLogEvent> &list, int maxSize) {
     std::lock_guard<std::mutex> lock(mutex_);
 
-    while (!_eventQueue.empty() && (maxSize == 0 || list.size() < maxSize)) {
+    while (!_eventQueue.empty() && (maxSize == 0 || list.size() < gsl::narrow<uint32_t>(maxSize))) {
```
szaszm, line 194: same as above

libminifi/include/utils/Monitors.h
```diff
@@ -41,7 +41,7 @@ class AfterExecute {
 
   AfterExecute() = default;
 
-  explicit AfterExecute(AfterExecute &&other) {
+  explicit AfterExecute(AfterExecute&& /*other*/) {
```
szaszm, line 44: Can we remove `explicit` and explicitly `default` this?

libminifi/test/unit/GeneralUtilsTest.cpp
```diff
@@ -130,7 +130,7 @@ TEST_CASE("GeneralUtils::invoke FunctionObject", "[invoke function object]") {
   REQUIRE(false == utils::invoke(&free_function, false));
 
   const auto n = 3;
-  const auto int_timesn = [n](const int i) { return n * i; };
+  const auto int_timesn = [=](const int i) { return n * i; };
```
szaszm, line 133: I prefer the old version here.

nanofi/src/api/nanofi.cpp
```diff
@@ -457,7 +457,7 @@ int get_content(const flow_file_record* ff, uint8_t* target, int size) {
     return stream->read(target, size);
   } else {
     file_buffer fb = file_to_buffer(ff->contentLocation);
-    size_t copy_size = size < fb.file_len ? size : gsl::narrow<size_t>(fb.file_len);
+    size_t copy_size = static_cast<size_t>(size) < fb.file_len ? size : gsl::narrow<size_t>(fb.file_len);
```
szaszm, line 460: I would check for negative `size` before casting to avoid surprises.

szaszm (changes requested):

bstrp_functions.sh
```diff
@@ -375,6 +376,7 @@ show_supported_features() {
   echo "4. Use Shared Dependency Links .$(print_feature_status USE_SHARED_LIBS)"
   echo "5. Build Profile ...............$(print_multi_option_status BUILD_PROFILE)"
   echo "6. Create ASAN build ...........$(print_feature_status ASAN_ENABLED)"
+  echo "7. Treat warnings as errors.....$(print_feature_status FAIL_ON_WARNINGS)"
```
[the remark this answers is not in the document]
lordgamez: Sure, updated in 267cda83a77a5a6d05f9df714d0c923814351826

cmake/Date.cmake
```diff
@@ -69,6 +69,9 @@ if (NOT date_src_POPULATED)
     add_library(date::tz ALIAS date-tz)
     target_include_directories(date-tz PUBLIC ${DATE_INCLUDE_DIR})
     target_compile_features(date-tz PUBLIC cxx_std_11)
+    if (NOT WIN32)
+        target_compile_options(date-tz PRIVATE "-Wno-error")
```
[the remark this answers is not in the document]
lordgamez: I tried that at first, but it seemed to be way more problematic as a lot of CMakeFiles needed to be updated for that. The other reason I decided to do it this way is because code added in the future will be automatically have the warning compile options, without the need to explicitly define them again. In the third party libraries they are not part of the PASSTHROUGH cmake options, only the Date.cmake file had to be updated with this option as they do not use those options.

controller/Controller.h
```diff
@@ -36,7 +36,7 @@ bool sendSingleCommand(std::unique_ptr<minifi::io::Socket> socket, uint8_t op, c
   minifi::io::BufferStream stream;
   stream.write(&op, 1);
   stream.write(value);
-  return socket->write(const_cast<uint8_t*>(stream.getBuffer()), gsl::narrow<int>(stream.size())) == stream.size();
+  return gsl::narrow<size_t>(socket->write(const_cast<uint8_t*>(stream.getBuffer()), gsl::narrow<int>(stream.size()))) == stream.size();
```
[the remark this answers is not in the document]
lordgamez: Updated in 267cda83a77a5a6d05f9df714d0c923814351826

extensions/coap/controllerservice/CoapResponse.h
```diff
@@ -73,7 +73,7 @@ class CoapResponse {
    * Returns a const pointer to the constant data.
    * @return data pointer.
    */
-  const uint8_t * const getData() const {
+  uint8_t* getData() const {
```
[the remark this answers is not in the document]
lordgamez: Updated in 267cda83a77a5a6d05f9df714d0c923814351826

extensions/openwsman/processors/SourceInitiatedSubscriptionListener.cpp
lordgamez: Updated in 267cda83a77a5a6d05f9df714d0c923814351826

extensions/rocksdb-repos/RocksDatabase.h
lordgamez: Updated in 267cda83a77a5a6d05f9df714d0c923814351826

extensions/script/lua/LuaBaseStream.cpp
lordgamez: Updated in 267cda83a77a5a6d05f9df714d0c923814351826

extensions/script/lua/LuaBaseStream.cpp
lordgamez: Updated in 267cda83a77a5a6d05f9df714d0c923814351826

extensions/script/python/PythonScriptEngine.cpp
lordgamez: I am a bit unsure about this. Checking back in #500 this seems to be always part of the initialization, but I do not know how those members actually work in the context of the python interpeter. Is there a way to test this?

extensions/standard-processors/processors/ListenSyslog.cpp
lordgamez: Updated in 267cda83a77a5a6d05f9df714d0c923814351826

extensions/standard-processors/processors/ListenSyslog.cpp
lordgamez: Updated in 267cda83a77a5a6d05f9df714d0c923814351826

libminifi/include/utils/Monitors.h
lordgamez: Yes, updated in 267cda83a77a5a6d05f9df714d0c923814351826

libminifi/test/unit/GeneralUtilsTest.cpp
lordgamez: Removed the capture altogether as it is not needed for this test. Updated in 267cda83a77a5a6d05f9df714d0c923814351826

extensions/script/python/PythonScriptEngine.cpp
szaszm: You're right, the `py::scoped_interpreter` member of `Interpreter` is actually initializing global state. Nevermind.

nanofi/src/api/nanofi.cpp
lordgamez: Added check before cast.

extensions/openwsman/processors/SourceInitiatedSubscriptionListener.cpp
```diff
@@ -563,7 +563,7 @@ bool SourceInitiatedSubscriptionListener::Handler::handleSubscriptionManager(str
     ws_xml_add_child(subscribe_node, XML_NS_EVENTING, WSEVENT_EXPIRES, millisecondsToXsdDuration(processor_.subscription_expiration_interval_).c_str());
 
     // Body/Filter
-    WsXmlNodeH filter_node = ws_xml_add_child(subscribe_node, XML_NS_WS_MAN, WSM_FILTER, processor_.xpath_xml_query_.c_str());
+    // WsXmlNodeH filter_node = ws_xml_add_child(subscribe_node, XML_NS_WS_MAN, WSM_FILTER, processor_.xpath_xml_query_.c_str());
```
szaszm, line 566: I believe this might be an accidental behavior change. `ws_xml_add_child` has side effects, it adds a child to the parameter node `subscribe_node`.

lordgamez: I believe you're right. Readded in 72ecc8d8d60244e8087510f05fa7399324b215cb

szaszm (approved):

hunyadi-dev (approved):

szaszm:
`-DENABLE_SENSORS=ON -DFAIL_ON_WARNINGS=ON` seems to be broken because `thirdparty/RTIMULib` is added with `add_subdirectory`, inherits the `Werror` and has warnings. Not sure if this should be fixed in context of this issue or a new one.

PR #1022, opened 2021-03-05
Refactor ThreadPool::execute to return a future since it always retur…

dam4rus:
…ned true

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/standard-processors/processors/GetTCP.cpp
+++ b/extensions/standard-processors/processors/GetTCP.cpp
@@ -285,17 +285,14 @@ void GetTCP::onTrigger(const std::shared_ptr<core::ProcessContext> &context, con
       } else {
         logger_->log_error("Could not create socket for %s", endpoint);
       }
-      std::unique_ptr<utils::AfterExecute<int>> after_execute = std::unique_ptr<utils::AfterExecute<int>>(new SocketAfterExecute(running_, endpoint, &live_clients_, &mutex_));
+      auto after_execute = std::unique_ptr<utils::AfterExecute<int>>(new SocketAfterExecute(running_, endpoint, &live_clients_, &mutex_));
       utils::Worker<int> functor(f_ex, "workers", std::move(after_execute));
-      auto* future = new std::future<int>(client_thread_pool_.execute(std::move(functor)));
-      live_clients_[endpoint] = future;
+      live_clients_[endpoint] = client_thread_pool_.execute(std::move(functor));
     } else {
-      if (!endPointFuture->second->valid()) {
-        delete endPointFuture->second;
-        std::unique_ptr<utils::AfterExecute<int>> after_execute = std::unique_ptr<utils::AfterExecute<int>>(new SocketAfterExecute(running_, endpoint, &live_clients_, &mutex_));
+      if (!endPointFuture->second.valid()) {
+        auto after_execute = std::unique_ptr<utils::AfterExecute<int>>(new SocketAfterExecute(running_, endpoint, &live_clients_, &mutex_));
         utils::Worker<int> functor(f_ex, "workers", std::move(after_execute));
-        auto* future = new std::future<int>(client_thread_pool_.execute(std::move(functor)));
-        live_clients_[endpoint] = future;
+        endPointFuture->second = client_thread_pool_.execute(std::move(functor));
       } else {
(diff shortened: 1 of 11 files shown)


libminifi/test/unit/ThreadPoolTests.cpp
```diff
@@ -70,8 +70,7 @@ TEST_CASE("ThreadPoolTest1", "[TPT1]") {
   std::function<bool()> f_ex = function;
   utils::Worker<bool> functor(f_ex, "id");
   pool.start();
-  std::future<bool> fut;
-  REQUIRE(true == pool.execute(std::move(functor), fut));
+  auto fut = pool.execute(std::move(functor));
   fut.wait();
```
adamdebreceni, line 74: should we assert the validity of the returned future in this test?

dam4rus: I'm not sure about whether it makes sense to test for validity or not to be honest. Judging by the documentation https://en.cppreference.com/w/cpp/thread/future/valid `std::future::valid` only returns `false` if `std::future::get` has been called. I can revert the assertion but I assume it would be always `true`.

extensions/standard-processors/processors/GetTCP.cpp
```diff
@@ -285,21 +285,17 @@ void GetTCP::onTrigger(const std::shared_ptr<core::ProcessContext> &context, con
       } else {
         logger_->log_error("Could not create socket for %s", endpoint);
       }
-      auto* future = new std::future<int>();
       std::unique_ptr<utils::AfterExecute<int>> after_execute = std::unique_ptr<utils::AfterExecute<int>>(new SocketAfterExecute(running_, endpoint, &live_clients_, &mutex_));
       utils::Worker<int> functor(f_ex, "workers", std::move(after_execute));
-      if (client_thread_pool_.execute(std::move(functor), *future)) {
-        live_clients_[endpoint] = future;
-      }
+      auto* future = new std::future<int>(client_thread_pool_.execute(std::move(functor)));
```
hunyadi-dev, line 290: Who cleans up these futures allocated with `new`?

libminifi/test/unit/ThreadPoolTests.cpp
adamdebreceni: a default constructed future and a future that has been moved from are also `valid() == false`, my only problem is that operating on an "invalid" future is undefined behavior (although encouraged to throw), but I believe we can safely assume that `ThreadPool<T>::execute` returns a valid future, in which case I would remove the validity checks from `SchedulingAgent.cpp`

dam4rus: Sorry for the confusing comment. I meant that in this case `valid` would be only `false` after calling `get`/`share`. If I get this right, `std::promise::get_future` can't return an invalid `std::future` since it throws an exception when called on an invalid `std::promise`. E.g. it has no state or `get_future` has already been called once.
I'm not super familiar with `std::promise` to be honest but I think it's safe to remove validity checks in these cases. Wouldn't an invalid `std::future` would have cause problems by now?

extensions/standard-processors/processors/GetTCP.cpp
szaszm: It's cleaned up by subsequent `onTrigger` calls, see the branch below. The list is also modified by `SocketAfterExecute`, which looks like it leaks on error, but I didn't verify.

dam4rus: Looking at the code it looks like `SocketAfterExecute::isFinished` is indeed leaking the future. I think simply storing an `std::future` in the `GetTCP::live_clients_` should clear this up.
The whole `SocketAfterExecute` class is kinda suspicious. It stores a copy of `GetTCP::running_` while storing a pointer to `GetTCP::live_clients_` and `GetTCP::mutex_`. Isn't it supposed to take a reference to `GetTCP::running_` as well?
What are the lifetime relation of these objects? Can `SocketAfterExecute` outlive `GetTCP`? If so storing an `std::weak_ptr` would be safer than raw pointers. Otherwise references should would work as well but harder to reason about. Maybe store `GetTCP::running_`, `GetTCP::live_clients_` and `GetTCP::mutex_` in a struct and create a `std::shared_ptr` to it?

szaszm:
closing due to inactivity

PR #1027, opened 2021-03-11
MINIFICPP-1487 Use a single queue for flow files

fgerlits:
https://issues.apache.org/jira/browse/MINIFICPP-1487

This PR has some overlap with #1025 (the first two commits), so it may be easier to review the commits separately, skipping those.  We may merge either of these PRs, both, or neither.

Instead of a separate `deque` for non-penalized flow files and a `priority_queue` for penalized flow files, this PR uses a single `priority_queue` for both, setting a "due date" to be the insertion time for non-penalized flow files and the penalty expiration time for penalized flow files.

This makes the code in FlowFileQueue simpler, easier to reason about, and the unit tests more robust.  On the other hand, `push` and `pop` are now O(log n) for non-penalized flow files, too, instead of O(1).  I think this is an acceptable cost, as O(log n) is not significantly slower than O(1), but I'm interested in what others think.

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/standard-processors/tests/unit/ProcessorTests.cpp
+++ b/extensions/standard-processors/tests/unit/ProcessorTests.cpp
@@ -702,7 +702,7 @@ TEST_CASE_METHOD(ProcessorWithIncomingConnectionTest, "A Processor detects corre
 TEST_CASE_METHOD(ProcessorWithIncomingConnectionTest, "A failed and re-penalized flow file does not block the incoming queue of the Processor", "[penalize]") {
   processor_->setPenalizationPeriod(std::chrono::milliseconds{100});
   const auto penalized_flow_file = session_->create();
-  session_->penalize(penalized_flow_file);  // first penalty duration is 100 ms
+  session_->penalize(penalized_flow_file);
   incoming_connection_->put(penalized_flow_file);
   const auto penalty_has_expired = [penalized_flow_file] { return !penalized_flow_file->isPenalized(); };
   REQUIRE(utils::verifyEventHappenedInPollTime(std::chrono::seconds{1}, penalty_has_expired, std::chrono::milliseconds{10}));
@@ -719,7 +719,7 @@ TEST_CASE_METHOD(ProcessorWithIncomingConnectionTest, "A failed and re-penalized
   const auto next_flow_file_1 = incoming_connection_->poll(expired_flow_files);
   REQUIRE(next_flow_file_1 == penalized_flow_file);
 
-  session_->penalize(penalized_flow_file);  // second penalty duration is 200 ms
+  session_->penalize(penalized_flow_file);
   incoming_connection_->put(penalized_flow_file);
   std::this_thread::sleep_for(std::chrono::milliseconds{110});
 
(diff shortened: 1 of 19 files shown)


adamdebreceni:
image

!mind the baseline labeled "empty", that is a constant cost of the benchmarking system!

I have measured if there is observable penalty going to O(log n), these tests operate (push/pop) on 10 `std::shared_ptr` instances, popping 10 elements from a 1000000 long priority_queue takes less than a microsecond on my machine, of course compared to the deque it is a slowdown on the order of 10, but based on this, I would not worry about it being a bottleneck of some sort

adamdebreceni:
image


!mind the baseline labeled "empty", that is a constant cost of the benchmarking system!

I have measured if there is observable penalty going to O(log n), these tests operate (push/pop) on 10 `std::shared_ptr<int>` instances (with custom comparator), popping 10 elements from a 1000000 long priority_queue takes less than a `6.5 us` on my machine, of course compared to the deque it is a slowdown by a factor of ~100, but based on this, I would not worry about it being a bottleneck of some sort

szaszm (approved):

libminifi/src/utils/FlowFileQueue.cpp
```diff
@@ -30,70 +30,41 @@ bool FlowFileQueue::FlowFilePenaltyExpirationComparator::operator()(const value_
...
-  }
-
-  if (!priority_queue_.empty()) {
-    value_type next_flow_file = priority_queue_.top();
-    priority_queue_.pop();
-    return next_flow_file;
-  }
-
-  throw std::logic_error{"forcePop() called on an empty FlowFileQueue"};
+  value_type next_flow_file = queue_.top();
```
adamdebreceni, line 37: should we move from `queue_.top()`?

extensions/standard-processors/tests/unit/ProcessorTests.cpp
```diff
@@ -615,65 +615,131 @@ TEST_CASE("TestRPGValid", "[TestRPG6]") {
...
+  incoming_connection_->put(flow_file_2);
+  const auto flow_file_3 = session_->create();
+  incoming_connection_->put(flow_file_3);
+
+  REQUIRE(incoming_connection_->isWorkAvailable());
+  std::set<std::shared_ptr<core::FlowFile>> expired_flow_files;
+  const auto next_flow_file_1 = incoming_connection_->poll(expired_flow_files);
+  REQUIRE(next_flow_file_1 == penalized_flow_file);
+
+  session_->penalize(penalized_flow_file);  // second penalty duration is 200 ms
```
adamdebreceni, line 722: does this comment apply?

libminifi/src/utils/FlowFileQueue.cpp
fgerlits: I don't think we can: `top()` returns a const reference (and is a const member function).

extensions/standard-processors/tests/unit/ProcessorTests.cpp
fgerlits: Nope.  Removed in https://github.com/apache/nifi-minifi-cpp/pull/1027/commits/a42c324123487d8a05be45c236ef1b42fd24457e.

libminifi/src/utils/FlowFileQueue.cpp
adamdebreceni: that is a bummer, there really is no non-const overload :(

adamdebreceni (approved):

lordgamez (approved):

PR #1028, opened 2021-03-12
MINIFICPP-1507 convert InputStream::read to size_t

szaszm:
Converting `write` is done in #1083 to keep the PR size managable. Errors are now reported as `static_cast<size_t>(-1)`. 

_______

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/http-curl/tests/HTTPHandlers.h
+++ b/extensions/http-curl/tests/HTTPHandlers.h
@@ -214,7 +214,7 @@ class FlowFileResponder : public ServerAwareHandler {
       minifi::io::CivetStream civet_stream(conn);
       minifi::io::CRCStream < minifi::io::CivetStream > stream(gsl::make_not_null(&civet_stream));
       uint32_t num_attributes;
-      int read;
+      size_t read;
       uint64_t total_size = 0;
       read = stream.read(num_attributes);
       if (!isServerRunning())return false;
@@ -241,9 +241,10 @@ class FlowFileResponder : public ServerAwareHandler {
       flow->data.resize(gsl::narrow<size_t>(length));
       flow->total_size = total_size;
 
-      read = stream.read(flow->data.data(), gsl::narrow<int>(length));
-      if (!isServerRunning())return false;
-      assert(read == gsl::narrow<int>(length));
+      read = stream.read(flow->data.data(), length);
+      if (!isServerRunning()) return false;
+      (void)read;
+      assert(read == length);
 
       if (!invalid_checksum) {
         site2site_rest_resp = std::to_string(stream.getCRC());
--- a/extensions/http-curl/tests/HTTPHandlers.h
+++ b/extensions/http-curl/tests/HTTPHandlers.h
@@ -213,38 +213,51 @@ class FlowFileResponder : public ServerAwareHandler {
     if (!wrong_uri) {
       minifi::io::CivetStream civet_stream(conn);
       minifi::io::CRCStream < minifi::io::CivetStream > stream(gsl::make_not_null(&civet_stream));
-      uint32_t num_attributes;
-      size_t read;
+      uint32_t num_attributes = 0;
       uint64_t total_size = 0;
-      read = stream.read(num_attributes);
-      if (!isServerRunning())return false;
-      assert(read > 0); total_size += read;
+      {
+        const auto read = stream.read(num_attributes);
+        if (!isServerRunning()) return false;
+        assert(read > 0);
+        total_size += read;
+      }
 
-      auto flow = std::make_shared<FlowObj>();
+      const auto flow = std::make_shared<FlowObj>();
 
       for (uint32_t i = 0; i < num_attributes; i++) {
         std::string name, value;
-        read = stream.read(name, true);
-        if (!isServerRunning())return false;
-        assert(read > 0); total_size += read;
-        read = stream.read(value, true);
-        if (!isServerRunning())return false;
-        assert(read > 0); total_size += read;
+        {
+          const auto read = stream.read(name, true);
+          if (!isServerRunning()) return false;
+          assert(read > 0);
+          total_size += read;
+        }
+        {
+          const auto read = stream.read(value, true);
+          if (!isServerRunning()) return false;
+          assert(read > 0);
+          total_size += read;
+        }
         flow->attributes[name] = value;
       }
       uint64_t length;
-      read = stream.read(length);
-      if (!isServerRunning())return false;
-      assert(read > 0); total_size += read;
+      {
+        const auto read = stream.read(length);
+        if (!isServerRunning()) return false;
+        assert(read > 0);
+        total_size += read;
+      }
 
       total_size += length;
       flow->data.resize(gsl::narrow<size_t>(length));
       flow->total_size = total_size;
 
-      read = stream.read(flow->data.data(), length);
-      if (!isServerRunning()) return false;
-      (void)read;
-      assert(read == length);
+      {
+        const auto read = stream.read(flow->data.data(), length);
+        if (!isServerRunning()) return false;
+        (void)read;
+        assert(read == length);
+      }
 
       if (!invalid_checksum) {
         site2site_rest_resp = std::to_string(stream.getCRC());
--- a/extensions/libarchive/CompressContent.h
+++ b/extensions/libarchive/CompressContent.h
@@ -187,11 +187,14 @@ public:
       return ret;
     }
 
-    static la_ssize_t archive_read(struct archive*, void *context, const void **buff) {
+    static la_ssize_t archive_read(struct archive* archive, void *context, const void **buff) {
       auto *callback = (WriteCallback *) context;
       callback->session_->read(callback->flow_, &callback->readDecompressCb_);
       *buff = callback->readDecompressCb_.buffer;
-      if (io::isError(callback->readDecompressCb_.stream_read_result)) return -1;
+      if (io::isError(callback->readDecompressCb_.stream_read_result)) {
+        archive_set_error(archive, EIO, "Error reading flowfile");
+        return -1;
+      }
       return gsl::narrow<la_ssize_t>(callback->readDecompressCb_.stream_read_result);
     }
 
--- a/extensions/libarchive/CompressContent.h
+++ b/extensions/libarchive/CompressContent.h
@@ -21,6 +21,7 @@
 #define __COMPRESS_CONTENT_H__
 
 #include <cinttypes>
+#include <utility>
 
 #include "archive_entry.h"
 #include "archive.h"
@@ -43,7 +44,7 @@ namespace minifi {
 namespace processors {
 
 // CompressContent Class
-class CompressContent: public core::Processor {
+class CompressContent : public core::Processor {
 public:
   // Constructor
   /*!
@@ -56,7 +57,7 @@ public:
     , encapsulateInTar_(false) {
   }
   // Destructor
-  virtual ~CompressContent() = default;
+  ~CompressContent() override = default;
   // Processor Name
   static constexpr char const* ProcessorName = "CompressContent";
   // Supported Properties
@@ -94,8 +95,8 @@ public:
     ReadCallbackCompress(std::shared_ptr<core::FlowFile> &flow, struct archive *arch, struct archive_entry *entry) :
         flow_(flow), arch_(arch), entry_(entry), status_(0), logger_(logging::LoggerFactory<CompressContent>::getLogger()) {
     }
-    ~ReadCallbackCompress() = default;
-    int64_t process(const std::shared_ptr<io::BaseStream>& stream) {
+    ~ReadCallbackCompress() override = default;
+    int64_t process(const std::shared_ptr<io::BaseStream>& stream) override {
       uint8_t buffer[4096U];
       int64_t ret = 0;
       uint64_t read_size = 0;
@@ -107,24 +108,24 @@ public:
         return -1;
       }
       while (read_size < flow_->getSize()) {
-        ret = stream->read(buffer, sizeof(buffer));
-        if (ret < 0) {
+        const auto readret = stream->read(buffer, sizeof(buffer));
+        if (io::isError(readret)) {
           status_ = -1;
           return -1;
         }
-        if (ret > 0) {
-          ret = archive_write_data(arch_, buffer, gsl::narrow<size_t>(ret));
+        if (readret > 0) {
+          ret = archive_write_data(arch_, buffer, readret);
           if (ret < 0) {
             logger_->log_error("Compress Content archive error %s", archive_error_string(arch_));
             status_ = -1;
             return -1;
           }
-          read_size += ret;
+          read_size += gsl::narrow<uint64_t>(ret);
         } else {
           break;
         }
       }
-      return read_size;
+      return gsl::narrow<int64_t>(read_size);
     }
     std::shared_ptr<core::FlowFile> flow_;
     struct archive *arch_;
@@ -133,28 +134,24 @@ public:
     std::shared_ptr<logging::Logger> logger_;
   };
   // Nest Callback Class for read stream from flow for decompress
-  class ReadCallbackDecompress: public InputStreamCallback {
-  public:
-    ReadCallbackDecompress(const std::shared_ptr<core::FlowFile> &flow) :
-        read_size_(0), offset_(0), flow_(flow) {
-      origin_offset_ = flow_->getOffset();
+  struct ReadCallbackDecompress : InputStreamCallback {
+    explicit ReadCallbackDecompress(std::shared_ptr<core::FlowFile> flow) :
+        flow_file(std::move(flow)) {
     }
-    ~ReadCallbackDecompress() = default;
-    int64_t process(const std::shared_ptr<io::BaseStream>& stream) {
-      read_size_ = 0;
-      stream->seek(offset_);
-      int readRet = stream->read(buffer_, sizeof(buffer_));
-      read_size_ = readRet;
-      if (readRet > 0) {
-        offset_ += read_size_;
+    ~ReadCallbackDecompress() override = default;
+    int64_t process(const std::shared_ptr<io::BaseStream>& stream) override {
+      stream->seek(offset);
+      const auto readRet = stream->read(buffer, sizeof(buffer));
+      stream_read_result = readRet;
+      if (!io::isError(readRet)) {
+        offset += readRet;
       }
-      return readRet;
+      return gsl::narrow<int64_t>(readRet);
     }
-    int64_t read_size_;
-    uint8_t buffer_[8192];
-    uint64_t offset_;
-    uint64_t origin_offset_;
-    std::shared_ptr<core::FlowFile> flow_;
+    size_t stream_read_result = 0;  // read size or error code, to be checked with io::isError
+    uint8_t buffer[8192] = {0};
+    size_t offset = 0;
+    std::shared_ptr<core::FlowFile> flow_file;
   };
   // Nest Callback Class for write stream
   class WriteCallback: public OutputStreamCallback {
@@ -190,16 +187,12 @@ public:
       return ret;
     }
 
-    static la_ssize_t archive_read(struct archive *arch, void *context, const void **buff) {
-      WriteCallback *callback = (WriteCallback *) context;
+    static la_ssize_t archive_read(struct archive*, void *context, const void **buff) {
+      auto *callback = (WriteCallback *) context;
       callback->session_->read(callback->flow_, &callback->readDecompressCb_);
-      if (callback->readDecompressCb_.read_size_ >= 0) {
-        *buff = callback->readDecompressCb_.buffer_;
-        return gsl::narrow<la_ssize_t>(callback->readDecompressCb_.read_size_);
-      } else {
-        archive_set_error(arch, EIO, "Error reading flowfile");
-        return -1;
-      }
+      *buff = callback->readDecompressCb_.buffer;
+      if (io::isError(callback->readDecompressCb_.stream_read_result)) return -1;
+      return gsl::narrow<la_ssize_t>(callback->readDecompressCb_.stream_read_result);
     }
 
     static la_int64_t archive_skip(struct archive* /*a*/, void* /*client_data*/, la_int64_t /*request*/) {
@@ -383,13 +376,14 @@ public:
           std::vector<uint8_t> buffer(16 * 1024U);
           int64_t read_size = 0;
           while (read_size < gsl::narrow<int64_t>(writer_.flow_->getSize())) {
-            int ret = inputStream->read(buffer.data(), gsl::narrow<int>(buffer.size()));
-            if (ret < 0) {
+            const auto ret = inputStream->read(buffer.data(), buffer.size());
+            if (io::isError(ret)) {
               return -1;
             } else if (ret == 0) {
               break;
             } else {
-              if (outputStream_->write(buffer.data(), ret) != ret) {
+              const auto writeret = outputStream_->write(buffer.data(), ret);
+              if (io::isError(writeret) || gsl::narrow<size_t>(writeret) != ret) {
                 return -1;
               }
               read_size += ret;
@@ -414,7 +408,7 @@ public:
 
       success_ = filterStream->isFinished();
 
-      return flow_->getSize();
+      return gsl::narrow<int64_t>(flow_->getSize());
     }
   };
 
@@ -444,7 +438,7 @@ private:
   }
 
   std::shared_ptr<logging::Logger> logger_;
-  int compressLevel_;
+  int compressLevel_{};
   CompressionMode compressMode_;
   ExtendedCompressionFormat compressFormat_;
   bool updateFileName_;
--- a/extensions/mqtt/processors/PublishMQTT.h
+++ b/extensions/mqtt/processors/PublishMQTT.h
@@ -20,6 +20,8 @@
 #ifndef __PUBLISH_MQTT_H__
 #define __PUBLISH_MQTT_H__
 
+#include <limits>
+
 #include "FlowFileRecord.h"
 #include "core/Processor.h"
 #include "core/ProcessSession.h"
@@ -29,6 +31,7 @@
 #include "core/logging/LoggerConfiguration.h"
 #include "MQTTClient.h"
 #include "AbstractMQTTProcessor.h"
+#include "utils/gsl.h"
 
 namespace org {
 namespace apache {
@@ -50,7 +53,7 @@ class PublishMQTT : public processors::AbstractMQTTProcessor {
     max_seg_size_ = ULLONG_MAX;
   }
   // Destructor
-  virtual ~PublishMQTT() = default;
+  ~PublishMQTT() override = default;
   // Processor Name
   static constexpr char const* ProcessorName = "PublishMQTT";
   // Supported Properties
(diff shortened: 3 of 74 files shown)


libminifi/test/archive-tests/MergeFileTests.cpp
```diff
@@ -111,8 +111,8 @@ std::vector<FixedBuffer> read_archives(const FixedBuffer& input) {
   class ArchiveEntryReader {
    public:
     explicit ArchiveEntryReader(archive* arch) : arch(arch) {}
-    int read(uint8_t* out, std::size_t len) {
-      return gsl::narrow<int>(archive_read_data(arch, out, len));
+    size_t read(uint8_t* out, std::size_t len) {
+      return gsl::narrow_cast<size_t>(archive_read_data(arch, out, len));
```
fgerlits, line 115: I know this is just a test file helper, but it would be safer to explicitly test for -1 and then use `gsl::narrow`, instead of using `gsl::narrow_cast` and risk that the return value will be something like `static_cast<size_t>(-3)`, which is not handled by `FixedBuffer::write()`.

fgerlits (commented):
Thanks for cleaning this up!

extensions/sftp/client/SFTPClient.cpp
```diff
@@ -577,20 +577,20 @@ bool SFTPClient::putFile(const std::string& path, io::BaseStream& input, bool ov
     }
     logger_->log_trace("Read %d bytes", read_ret);
     total_read += read_ret;
-    ssize_t remaining = read_ret;
+    auto remaining = gsl::narrow<size_t>(read_ret);
```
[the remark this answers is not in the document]
szaszm: Indeed, it was an oversight. Fixing in the next commit.

extensions/standard-processors/processors/ExtractText.cpp
```diff
@@ -126,22 +125,22 @@ int64_t ExtractText::ReadCallback::process(const std::shared_ptr<io::BaseStream>
   if (sizeLimitStr.empty())
     size_limit = DEFAULT_SIZE_LIMIT;
   else if (sizeLimitStr != "0")
-    size_limit = std::stoi(sizeLimitStr);
+    size_limit = gsl::narrow_cast<size_t>(std::stoi(sizeLimitStr));
```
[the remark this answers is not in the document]
szaszm: It comes from the user and negative numbers shouldn't crash the program, but they are not "valid" either.

libminifi/include/io/InputStream.h
```diff
@@ -39,43 +39,43 @@ class InputStream : public virtual Stream {
    * reads a byte array from the stream
    * @param value reference in which will set the result
    * @param len length to read
-   * @return resulting read size
+   * @return resulting read size or static_cast<size_t>(-1) on error or static_cast<size_t>(-2) for ClientSocket EAGAIN
```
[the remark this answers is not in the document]
szaszm: I felt the same, but didn't have a good idea how to resolve it. Refactoring Socket classes to use some other form of handling EAGAIN would be preferred, but that seems too out of scope here, and this change set feels dangerous enough already.

I like the idea of `Stream::isError`, I'll change the checks.

libminifi/src/c2/ControllerSocketProtocol.cpp
```diff
@@ -248,11 +248,11 @@ void ControllerSocketProtocol::initialize(core::controller::ControllerServicePro
 void ControllerSocketProtocol::parse_content(const std::vector<C2ContentResponse> &content) {
   for (const auto &payload_content : content) {
     if (payload_content.name == "Components") {
-      for (auto content : payload_content.operation_arguments) {
+      for (const auto& content_ : payload_content.operation_arguments) {
```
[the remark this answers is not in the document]
szaszm: A minimal change to avoid shadowing the parameter. I'm open to alternatives.

libminifi/src/io/tls/SecureDescriptorStream.cpp
```diff
@@ -71,34 +71,29 @@ int SecureDescriptorStream::write(const uint8_t *value, int size) {
...
-          sslStatus = SSL_get_error(ssl_, status);
-        } while (status < 0 && sslStatus == SSL_ERROR_WANT_READ);
+  if (IsNullOrEmpty(buf)) return static_cast<size_t>(-1);
+  size_t total_read = 0;
+  uint8_t* writepos = buf;
+  while (buflen > total_read) {
+    int status;
+    int sslStatus;
+    do {
+      status = SSL_read(ssl_, writepos, gsl::narrow<int>(std::min(buflen - total_read, gsl::narrow<size_t>(std::numeric_limits<int>::max()))));
```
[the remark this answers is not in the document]
szaszm: The code is not prepared to handle exceptions and I don't want to rely on `gsl::narrow` throwing instead of calling `terminate`.

If there is any code that relies on `gsl::narrow` throwing vs terminating, that's an error.

libminifi/include/io/InputStream.h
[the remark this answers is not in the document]
szaszm: What do you think about `io::isError`?

extensions/standard-processors/processors/ExtractText.cpp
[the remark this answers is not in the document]
fgerlits: I think throwing an exception in `onSchedule()` would be better than treating a limit of "-2" as a limit of 18.4 exabytes, but this is how the code was working before this PR, and I agree that changing it would be out of scope.

libminifi/include/io/InputStream.h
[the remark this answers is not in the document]
fgerlits: yup, sounds good

libminifi/src/c2/ControllerSocketProtocol.cpp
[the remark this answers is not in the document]
fgerlits: Ah OK, I didn't see the name of the argument.  We could rename `content_` to `operation_argument`, but it's not vital.

libminifi/src/io/tls/SecureDescriptorStream.cpp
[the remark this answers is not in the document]
fgerlits: OK, that makes sense.  This line is quite long and hard to read, though; pulling out the 3rd argument (with or without the `gsl::narrow`) and giving it a name could make it more readable.

extensions/standard-processors/processors/ExtractText.cpp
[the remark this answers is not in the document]
szaszm: Actually it would make sense to cast them to `size_t` and treat `-1` as "essentially no limit" instead of throwing. If you think this is a good idea, I'll change it to `static_cast<size_t>(...)`.

libminifi/src/FlowFileRecord.cpp
```diff
@@ -181,12 +179,12 @@ std::shared_ptr<FlowFileRecord> FlowFileRecord::DeSerialize(io::InputStream& inS
   }
 
   ret = inStream.read(file->uuid_);
-  if (ret <= 0) {
+  if (ret == static_cast<size_t>(-1)) {
```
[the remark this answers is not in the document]
szaszm: Definitely an oversight, will fix. If you find other suspicious places, feel free to point out. I might have missed some in the 100+ "Find usages" results.

extensions/script/python/PyBaseStream.cpp
```diff
@@ -49,13 +49,11 @@ py::bytes PyBaseStream::read(size_t len) {
 
   std::vector<uint8_t> buffer(len);
 
-  auto read = stream_->read(buffer.data(), static_cast<int>(len));
-  auto result = py::bytes(reinterpret_cast<char *>(buffer.data()), static_cast<size_t>(read));
-
-  return result;
+  const auto read = stream_->read(buffer.data(), len);
+  return py::bytes(reinterpret_cast<char *>(buffer.data()), static_cast<size_t>(read));
```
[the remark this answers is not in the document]
szaszm: resolved in 34b60f0

extensions/sftp/client/SFTPClient.cpp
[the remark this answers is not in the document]
szaszm: resolved in 34b60f0

libminifi/include/sitetosite/SiteToSiteClient.h
```diff
@@ -65,19 +67,14 @@ class SiteToSiteClient : public core::Connectable {
...
         ssl_context_service_(nullptr),
         logger_(logging::LoggerFactory<SiteToSiteClient>::getLogger()) {
-    _supportedVersion[0] = 5;
-    _supportedVersion[1] = 4;
-    _supportedVersion[2] = 3;
-    _supportedVersion[3] = 2;
-    _supportedVersion[4] = 1;
     _currentVersion = _supportedVersion[0];
     _currentVersionIndex = 0;
     _supportedCodecVersion[0] = 1;
```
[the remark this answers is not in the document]
szaszm: done in 34b60f0

libminifi/src/FlowFileRecord.cpp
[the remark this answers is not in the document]
szaszm: resolved in 34b60f0

libminifi/src/sitetosite/SiteToSiteClient.cpp
```diff
@@ -31,47 +31,37 @@ namespace sitetosite {
...
-
-  if (resCode == NULL) {
-    // Not a valid respond code
+  if (!resCode) {
     return -1;
   }
   if (resCode->hasDescription) {
     ret = peer_->read(message);
-    if (ret <= 0)
+    if (ret != static_cast<size_t>(-1))
```
fgerlits, line 55: missing `== 0 ||`

thirdparty/google-styleguide/run_linter.sh
```diff
@@ -40,5 +40,7 @@ done
 
 HEADERS=`find $INCLUDE_DIRS -name '*.h' | sort | uniq | tr '\n' ' '`
 SOURCES=`find $SOURCE_DIRS -name  '*.cpp' | sort | uniq | tr '\n' ' '`
-REPOSITORY="$(realpath --physical "$(dirname "$0")/../..")"
+# this realpath alternative should work on mac
+alias prealpath="python -c 'import os, sys; print(os.path.realpath(sys.argv[1]))'"
```
fgerlits, line 44: where is this alias used?

libminifi/src/io/InputStream.cpp
```diff
@@ -30,42 +30,43 @@ namespace nifi {
...
 }
 
-int InputStream::read(std::string &str, bool widen) {
+size_t InputStream::read(std::string &str, bool widen) {
   uint32_t len = 0;
-  int ret = 0;
+  size_t ret = 0;
   if (!widen) {
     uint16_t shortLength = 0;
     ret = read(shortLength);
```
fgerlits, line 72: I can't comment at line 78, but it needs to be updated from `<= 0` to `== 0 || == static_cast(-1)`, too

thirdparty/google-styleguide/run_linter.sh
szaszm: In the end I incorporated it to the below command, because the alias didn't work in subshells. I forgot to remove this one.

libminifi/include/io/InputStream.h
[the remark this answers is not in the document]
szaszm: done in 0c23d4e

libminifi/src/sitetosite/SiteToSiteClient.cpp
```diff
@@ -31,47 +31,43 @@ namespace sitetosite {
...
-  if (resCode == NULL) {
-    // Not a valid respond code
+  if (!resCode) {
     return -1;
   }
   if (resCode->hasDescription) {
-    ret = peer_->read(message);
-    if (ret <= 0)
+    const auto ret = peer_->read(message);
+    if (ret == 0 || !io::isError(ret))
```
fgerlits, line 61:
```suggestion
    if (ret == 0 || io::isError(ret))
```

libminifi/test/rocksdb-tests/DBContentRepositoryTests.cpp
```diff
@@ -140,7 +140,7 @@ TEST_CASE("Test Empty Claim", "[TestDBCR3]") {
   std::string readstr;
 
   // -1 tell us we have an invalid stream
```
fgerlits, line 142: minor, but you changed "-1" to "error" in the other place -- should be changed here, too

extensions/standard-processors/processors/ExtractText.cpp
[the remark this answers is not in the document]
szaszm: changed to `static_cast` in 0c23d4e

libminifi/src/c2/ControllerSocketProtocol.cpp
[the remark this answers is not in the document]
szaszm: renamed to `operation_argument` in 0c23d4e

libminifi/src/io/tls/SecureDescriptorStream.cpp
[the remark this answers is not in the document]
szaszm: named the third argument in 0c23d4e

libminifi/test/archive-tests/MergeFileTests.cpp
szaszm: Introduced a check for negative return values in 0c23d4e

libminifi/test/rocksdb-tests/DBContentRepositoryTests.cpp
szaszm: changed in 46f3b0e

libminifi/src/io/InputStream.cpp
szaszm: fixed in 0c23d4e

thirdparty/google-styleguide/run_linter.sh
szaszm: removed the leftover line in 0c23d4e

libminifi/src/sitetosite/SiteToSiteClient.cpp
szaszm: fixed partially in 0c23d4e and finally in 46f3b0e after https://github.com/apache/nifi-minifi-cpp/pull/1028#discussion_r596108625

libminifi/test/rocksdb-tests/ContentSessionTests.cpp
```diff
@@ -56,11 +56,10 @@ const std::shared_ptr<minifi::io::BaseStream>& operator>>(const std::shared_ptr<
   str = "";
   uint8_t buffer[4096]{};
   while (true) {
-    size_t ret = stream->read(buffer, sizeof(buffer));
+    const auto ret = stream->read(buffer, sizeof(buffer));
     REQUIRE(ret >= 0);
```
fgerlits, line 60:
```suggestion
    REQUIRE(!minifi::io::isError(ret));
```

libminifi/test/rocksdb-tests/DBContentRepositoryTests.cpp
```diff
@@ -179,7 +179,7 @@ TEST_CASE("Delete NonExistent Claim", "[TestDBCR4]") {
 
   std::string readstr;
 
-  // -1 tell us we have an invalid stream
+  // error tells us we have an invalid stream
```
fgerlits, line 182: it would be better to remove this comment, as we don't check for error here (also at line 230)

libminifi/test/rocksdb-tests/ContentSessionTests.cpp
szaszm: We need both checks, and the new one is one line below. Maybe I misunderstood your point?

szaszm: I think I get it: `ret >= 0` is always true, I'll remove it.

fgerlits: somehow I didn't see the other check below it -- but yes, this can be removed

szaszm: done in 141917b

libminifi/test/rocksdb-tests/DBContentRepositoryTests.cpp
szaszm: done in 141917b

extensions/script/lua/LuaBaseStream.cpp
```diff
@@ -52,16 +52,11 @@ std::string LuaBaseStream::read(size_t len) {
...
-
-  if (gsl::narrow<size_t>(read) != len) {
-    buffer.resize(gsl::narrow<size_t>(read));
+  const auto read = stream_->read(reinterpret_cast<uint8_t *>(&buffer[0]), len);
+  if (!io::isError(read) && read != len) {
+    buffer.resize(read);
   }
-
-  return buffer;
+  return io::isError(read) ? std::string{} : buffer;
```
fgerlits, line 59: We used to throw an exception (in an obfuscated way) in the error case, now we return an empty string.  This is probably good, but I don't know if we want to make a change like this in a refactoring PR.  If we do want to make this change, we should probably change line 42, too.

szaszm: My understanding is that the old version returned `nullptr` as `std::string`, invoking the `std::string::string(const char*)` constructor which led to undefined behavior in case of an error. I thought replacing this with a defined empty string would be better, but I might be missing something.

szaszm: This snippet was changed during the rebase, btw. Previously it was a simple comparison + resize without the `narrow` and the `if (read < 0)` condition

fgerlits: yup, I said "probably good", you said "would be better", there is no disagreement there :)

libstdc++ throws an exception, but you're right that other implementations may behave differently

my point is that the same thing happens at line 42

szaszm: I didn't know that libstdc++ throws there. Despite being an unrelated change, I really don't want to rely on undefined behavior there, so I insist on a change.

In the old main version (that this PR was originally based on), there was no string construction from nullptr here (but the one on line 42 was already present). I believe that change on main to be an error, so by moving to returning an empty string, we basically revert to the old, correct version, the only change being that we no longer return an empty string of size `len`.

If you insist on throwing, I can change this to throw something explicitly, but I prefer the old behavior of returning an empty string. I want to change the occurrence on line 42 to be consistent with this one, too.

fgerlits: Yes, I think returning an empty string at line 42 would be fine.

fgerlits (approved):

extensions/libarchive/CompressContent.h
```diff
@@ -190,16 +189,11 @@ class CompressContent: public core::Processor {
...
       callback->session_->read(callback->flow_, &callback->readDecompressCb_);
-      if (callback->readDecompressCb_.read_size_ >= 0) {
-        *buff = callback->readDecompressCb_.buffer_;
-        return gsl::narrow<la_ssize_t>(callback->readDecompressCb_.read_size_);
-      } else {
-        archive_set_error(arch, EIO, "Error reading flowfile");
-        return -1;
-      }
+      *buff = callback->readDecompressCb_.buffer_;
+      return gsl::narrow<la_ssize_t>(callback->readDecompressCb_.read_size_);
```
adamdebreceni, line 196: [this](http://manpages.org/archive_read_open2/3) I-don't-know-how-trustworthy site says:

> On error, the read callback should invoke Fn archive_set_error to register an error code and message and return -1.

extensions/libarchive/CompressContent.h
```diff
@@ -383,13 +377,14 @@ class CompressContent: public core::Processor {
...
-            if (ret < 0) {
+            const auto ret = inputStream->read(buffer.data(), buffer.size());
+            if (io::isError(ret)) {
               return -1;
             } else if (ret == 0) {
               break;
             } else {
-              if (outputStream_->write(buffer.data(), ret) != ret) {
+              const auto writeret = outputStream_->write(buffer.data(), ret);
+              if (!io::isError(writeret) && gsl::narrow<size_t>(writeret) != ret) {
```
adamdebreceni, line 387: previously this returned -1 on error, shouldn't it be?
```suggestion
              if (io::isError(writeret) || gsl::narrow<size_t>(writeret) != ret) {
```

extensions/mqtt/processors/PublishMQTT.h
```diff
@@ -74,19 +77,21 @@ class PublishMQTT : public processors::AbstractMQTTProcessor {
...
         max_seg_size_ = flow_size_;
+      gsl_Expects(max_seg_size_ < gsl::narrow<uint64_t>(std::numeric_limits<int>::max()));
       std::vector<unsigned char> buffer;
       buffer.reserve(max_seg_size_);
       read_size_ = 0;
       status_ = 0;
       while (read_size_ < flow_size_) {
-        int readRet = stream->read(&buffer[0], max_seg_size_);
+        // MQTTClient_message::payloadlen is int, so we can't handle 2GB+
+        int readRet = static_cast<int>(stream->read(&buffer[0], max_seg_size_));
```
adamdebreceni, line 91: we could check the return value of `stream->read(...)` with `io::isError` before casting to `int`

extensions/libarchive/CompressContent.h
szaszm: Good catch. The site lists the official man page contents in a web format, so I would say the contents are valid. I've mistakenly removed the error branch because after converting `read_size_` to `size_t`, it was a dead branch and I didn't realize `read_size_` comes from a stream read return value. Will fix and rename `read_size_` to `stream_read_result`.

szaszm: fixed in 7955d00

extensions/mqtt/processors/PublishMQTT.h
szaszm: done in 7ef4b05

extensions/tensorflow/TFExtractTopLabels.cpp
```diff
@@ -134,9 +136,9 @@ int64_t TFExtractTopLabels::LabelsReadCallback::process(const std::shared_ptr<io
   buf.resize(buf_size);
 
   while (total_read < stream->size()) {
-    auto read = stream->read(reinterpret_cast<uint8_t *>(&buf[0]), static_cast<int>(buf_size));
-
-    for (auto i = 0; i < read; i++) {
+    const auto read = stream->read(reinterpret_cast<uint8_t *>(&buf[0]), buf_size);
+    if (io::isError(read)) break;
```
adamdebreceni, line 140: not sure how other `Callback::process` methods work when they encounter an error, do they generally return what they managed to process or do they signal an error?

libminifi/src/io/InputStream.cpp
```diff
@@ -30,42 +30,43 @@ namespace nifi {
...
-  if (buffer.size() < gsl::narrow<size_t>(len)) {
+size_t InputStream::read(std::vector<uint8_t>& buffer, size_t len) {
+  if (buffer.size() < len) {
     buffer.resize(len);
   }
-  int ret = read(buffer.data(), len);
-  buffer.resize((std::max)(ret, 0));
+  const auto ret = read(buffer.data(), len);
+  if (io::isError(ret)) return ret;
+  buffer.resize((std::max)(ret, size_t{0}));
```
adamdebreceni, line 39: is the `std::max` still necessary?

libminifi/src/io/InputStream.cpp
```diff
@@ -84,9 +85,9 @@ int InputStream::read(std::string &str, bool widen) {
   }
 
   std::vector<uint8_t> buffer(len);
-  uint32_t bytes_read = gsl::narrow<uint32_t>(read(buffer.data(), len));
+  const auto bytes_read = gsl::narrow<uint32_t>(read(buffer.data(), len));
   if (bytes_read != len) {
-    return -1;
+    return static_cast<size_t>(-1);
```
adamdebreceni, line 90: as classes generally only override `read(uint8_t*, size_t)` won't using this swallow meaningful error codes coming from a base implementation? e.g. as I understand `Socket::read` might return `size_t{-2}`

libminifi/test/BufferReader.h
```diff
@@ -44,7 +44,7 @@ class BufferReader : public org::apache::nifi::minifi::InputStreamCallback {
   }
 
   int64_t process(const std::shared_ptr<org::apache::nifi::minifi::io::BaseStream>& stream) {
-    return write(*stream.get(), stream->size());
+    return static_cast<int64_t>(write(*stream.get(), stream->size()));
```
adamdebreceni, line 47: why not `gsl::narrow`?

libminifi/test/archive-tests/CompressContentTests.cpp
```diff
@@ -41,29 +41,23 @@
...
-  }
-  ~ReadCallback() {
-    if (buffer_)
-      delete[] buffer_;
-    if (archive_buffer_)
-      delete[] archive_buffer_;
-  }
-  int64_t process(const std::shared_ptr<minifi::io::BaseStream>& stream) {
+  explicit ReadCallback(size_t size)
+      :buffer_size_{size}
```
adamdebreceni, line 49: minor: a comment warning us that other member initializers depend on `buffer_size_` might help future readers

extensions/libarchive/CompressContent.h
adamdebreceni: `archive_set_error` should also be restored according to the link

lordgamez (approved):
LGTM, thanks for all the boy scouting.

extensions/tensorflow/TFExtractTopLabels.cpp
szaszm: read/write usually signal an error, but it's not consistent. I think Callback::process just forwards these results.

libminifi/include/io/Stream.h
```diff
@@ -24,14 +24,23 @@ namespace nifi {
 namespace minifi {
 namespace io {
 
+inline bool isError(const size_t read_return) noexcept {
+  return read_return == static_cast<size_t>(-1)   // general error
+      || read_return == static_cast<size_t>(-2);  // Socket EAGAIN, to be refactored to eliminate this error condition
```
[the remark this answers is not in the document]
szaszm: `static_cast<size_t>(-2)` is a special case, I don't want to touch that. I'm open to extracting `static_cast<size_t>(-1)` to a symbolic constant, will probably do it later.

libminifi/src/io/InputStream.cpp
szaszm: It would, but there is only one occurrence of returning -2 and I think it doesn't hit this path. That one occurrence should be cleaned up in the future anyway, so it's not an important issue IMO.
I will change it to forward the original return value, but if it mattered, we would already have an issue.

libminifi/test/BufferReader.h
szaszm: Because I wanted `static_cast<size_t>(-1)` to become `-1`. I will change it to explicitly check for error.

szaszm:
Manually cancelled github actions CI runs triggered by the push event after every job was successful from the pull_request event.

libminifi/include/io/Stream.h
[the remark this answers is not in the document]
szaszm: Introduced `constexpr auto minifi::io::STREAM_ERROR = static_cast<size_t>(-1);` in 47358cb

extensions/http-curl/tests/HTTPHandlers.h
```diff
@@ -214,7 +214,7 @@ class FlowFileResponder : public ServerAwareHandler {
       minifi::io::CivetStream civet_stream(conn);
       minifi::io::CRCStream < minifi::io::CivetStream > stream(gsl::make_not_null(&civet_stream));
       uint32_t num_attributes;
-      int read;
+      size_t read;
```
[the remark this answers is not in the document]
szaszm: refactored the function in 47358cb

libminifi/test/archive-tests/CompressContentTests.cpp
szaszm: added comment in 47358cb

libminifi/test/BufferReader.h
szaszm: changed in 47358cb

libminifi/src/io/InputStream.cpp
szaszm: changed to forward errors in 47358cb

libminifi/src/io/InputStream.cpp
szaszm: no, removed in 47358cb

extensions/libarchive/CompressContent.h
szaszm: restored `archive_set_error` in e8a0a33

adamdebreceni (approved):

libminifi/include/sitetosite/SiteToSiteClient.h
```diff
@@ -43,15 +45,14 @@ namespace sitetosite {
...
-      : payload_(payload),
-        logger_reference_(logger) {
-    _size = 0;
-    transaction_ = transaction;
-    _attributes = attributes;
+  DataPacket(std::shared_ptr<logging::Logger> logger, std::shared_ptr<Transaction> transaction, std::map<std::string, std::string> attributes, const std::string &payload)
+      : _attributes{std::move(attributes)},
+        transaction_{std::move(transaction)},
+        payload_{payload},
+        logger_reference_{std::move(logger)} {
```
martinzink, line 52: why is this better than the const& and copy?

martinzink: looked more into it, this might be better actually so consider my comments obsolete

martinzink (approved):

PR #1037, opened 2021-03-23
MINIFICPP-1520 Fix PublishKafka properties to support expression language

fgerlits:
https://issues.apache.org/jira/browse/MINIFICPP-1520

Contains the following fixes:
   * Known Brokers, Client Name, Delivery Guarantee now support expression language
   * Request Timeout and Message Timeout no longer claim to support expression language (as this is incompatible with TimePeriodValue validation)
   * deprecate Message Key Field, which didn't work, and replace it with Kafka Key (which is what this property is called in NiFi); Kafka Key supports expression language
   * dynamic properties support expression language now

Note: the unit test is not great, as it requires a running Kafka instance on the host, and it looks like it's included in CTest but isn't really.  However, this is better than nothing, it's the same as the other two existing Kafka unit tests, and I have opened a Jira to fix these problems: https://issues.apache.org/jira/browse/MINIFICPP-1530

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/PROCESSORS.md
+++ b/PROCESSORS.md
@@ -966,13 +966,13 @@ In the list below, the names of required properties appear in bold. Any other pr
 |Compress Codec|none||compression codec to use for compressing message sets|
 |Debug contexts|||A comma-separated list of debug contexts to enable.Including: generic, broker, topic, metadata, feature, queue, msg, protocol, cgrp, security, fetch, interceptor, plugin, consumer, admin, eos, all|
 |Delivery Guarantee|1||Specifies the requirement for guaranteeing that a message is sent to Kafka. Valid values are 0 (do not wait for acks), -1 or all (block until message is committed by all in sync replicas) or any concrete number of nodes.<br/>**Supports Expression Language: true**|
+|Kafka Key|||The key to use for the message. If not specified, the UUID of the flow file is used as the message key.<br/>**Supports Expression Language: true**|
 |Kerberos Keytab Path|||The path to the location on the local filesystem where the kerberos keytab is located. Read permission on the file is required.|
 |Kerberos Principal|||Keberos Principal|
 |Kerberos Service Name|||Kerberos Service Name|
 |**Known Brokers**|||A comma-separated list of known Kafka Brokers in the format <host>:<port><br/>**Supports Expression Language: true**|
 |Max Flow Segment Size|0 B||Maximum flow content payload segment size for the kafka record. 0 B means unlimited.|
 |Max Request Size|||Maximum Kafka protocol request message size|
-|Kafka Key|||The key to use for the message. If not specified, the UUID of the flow file is used as the message key.<br/>**Supports Expression Language: true**|
 |Message Key Field|||DEPRECATED, does not work -- use Kafka Key instead|
 |Message Timeout|30 sec||The total time sending a message could take|
 |Queue Buffering Max Time|||Delay to wait for messages in the producer queue to accumulate before constructing message batches|


extensions/librdkafka/PublishKafka.cpp
```diff
@@ -925,12 +931,10 @@ void PublishKafka::onTrigger(const std::shared_ptr<core::ProcessContext> &contex
     }
 
     std::string kafkaKey;
-    kafkaKey = "";
-    if (context->getDynamicProperty(MessageKeyField, kafkaKey, flowFile) && !kafkaKey.empty()) {
-      logger_->log_debug("PublishKafka: Message Key Field [%s]", kafkaKey);
-    } else {
+    if (!context->getProperty(KafkaKey, kafkaKey, flowFile) || kafkaKey.empty()) {
```
adamdebreceni, line 934: if `MessageKeyField` worked before and now it no longer works we should remove it altogether and not mark it deprecated, if we would like to enable a "soft transition" we should mark it deprecated but further support the feature and produce loud warnings that it will not be so for long

fgerlits: It's complicated.  `MessageKeyField` used to "work" in the sense that if you set it (e.g. to `kafka.message.key` or `${kafka.message.key}`) then every message sent to Kafka would have the message key set to this literal value ("kafka.message.key" or "${kafka.message.key}") without any substitution, i.e. all messages would have the same key.

The description of the `MessageKeyField` property was "The name of a field in the Input Records that should be used as the Key for the Kafka message. Supports Expression Language: true (will be evaluated using flow file attributes)", every part of which is false.

I hope nobody was using this, but I think that silently changing the message keys to the flow file UUID (which is what will happen if you were using `MessageKeyField` previously and don't change anything) is better than keeping the old behavior where every message had the same key.  Message keys are supposed to be unique.

What do you think of keeping the behavior as it is, but adding an error log if `MessageKeyField` is set?

adamdebreceni: I don't think this ever worked since we try to get the `MessageKeyField` as if it were a dynamic property (`context->getDynamicProperty`)

adamdebreceni: does an exception in an onSchedule really prevents a MiNiFi startup? that would be really bad, as there is no way to fix such a malformed flow through a c2 server

adamdebreceni: your approach of keeping it as is, and logging errors if `MessageKeyField` is set, could be our best bet, given the constraints

fgerlits: You're right about both MiNiFi startup and the old (non-)behavior or `MessageKeyField`.

I have added an error log in `onSchedule()` when `MessageKeyField` is set.

extensions/librdkafka/tests/CMakeLists.txt
```diff
@@ -29,10 +29,12 @@ FOREACH(testfile ${KAFKA_TESTS})
     createTests("${testfilename}")
     MATH(EXPR KAFKA_TEST_COUNT "${KAFKA_TEST_COUNT}+1")
     # The line below handles integration test
+    target_include_directories(${testfilename} BEFORE PRIVATE "../..")
```
szaszm, line 32: The comment above is outdated, I suggest removing it.

extensions/librdkafka/PublishKafka.cpp
```diff
@@ -541,6 +543,11 @@ void PublishKafka::onSchedule(const std::shared_ptr<core::ProcessContext> &conte
   conn_ = utils::make_unique<KafkaConnection>(key_);
   configureNewConnection(context);
 
+  std::string message_key_field;
+  if (context->getProperty(MessageKeyField.getName(), message_key_field)) {
+    logger_->log_error("The %s property is set. This property is DEPRECATED and has no effect; please use Kafka Key instead.", MessageKeyField.getName());
```
szaszm, line 548: I would prefer this to be a warning, not an error

adamdebreceni: I would go with error, as it uses a property that never worked and is only preserved to allow the processor to operate instead of fizzling out in onSchedule, personally I would even go for removing it, but I do not know how prevalent its usage is

szaszm: Having the property set is likely a mistake, but it doesn't cause data loss or failures down the line. Seeing this message doesn't mean a change in behavior from the happy path, only that the flow is incorrect, but still runnable.

fgerlits: I'm fine with logging it on either warning or error level, so we'll need a 4th vote to decide. :)

My reasoning for an error log was that if the user has set the MessageKeyField property, then their flow is not working as intended.  E.g. they may have wanted to set the message key to the value of a particular flow file attribute, but it is set to the flow file UUID instead -- this can cause difficult-to-debug problems down the line.

extensions/librdkafka/tests/CMakeLists.txt
fgerlits: I have removed the comment.

extensions/librdkafka/PublishKafka.cpp
arpadboda: I tend to agree with the error approach. Especially as this is only going to be logged once, so won't spam the log with error messages.

szaszm: ok

szaszm (approved):

arpadboda (approved):

PROCESSORS.md
```diff
@@ -971,13 +971,13 @@ In the list below, the names of required properties appear in bold. Any other pr
 |**Known Brokers**|||A comma-separated list of known Kafka Brokers in the format <host>:<port><br/>**Supports Expression Language: true**|
 |Max Flow Segment Size|0 B||Maximum flow content payload segment size for the kafka record. 0 B means unlimited.|
 |Max Request Size|||Maximum Kafka protocol request message size|
-|Message Key Field|||The name of a field in the Input Records that should be used as the Key for the Kafka message.
-Supports Expression Language: true (will be evaluated using flow file attributes)|
-|Message Timeout|30 sec||The total time sending a message could take<br/>**Supports Expression Language: true**|
+|Kafka Key|||The key to use for the message. If not specified, the UUID of the flow file is used as the message key.<br/>**Supports Expression Language: true**|
```
adamdebreceni, line 974: nitpick: other properties seem to be in an alphabetical order

adamdebreceni (approved):

PROCESSORS.md
fgerlits: fixed

lordgamez (approved):

PR #1041, opened 2021-04-06
MINIFICPP-1358 Add SendBody property functionality to InvokeHTTP processor

lordgamez:
Previously send-message-body property was available and documented in InvokeHTTP processor, but it had no functionality behind it. It is added and the documentation is updated with all available properties.

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/PROCESSORS.md
+++ b/PROCESSORS.md
@@ -693,7 +693,8 @@ In the list below, the names of required properties appear in bold. Any other pr
 |Put Response Body in Attribute|||If set, the response body received back will be put into an attribute of the original FlowFile instead of a separate FlowFile. The attribute key to put to is determined by evaluating value of this property.|
 |Read Timeout|15 secs||Max wait time for response from remote service.|
 |Remote URL|||Remote URL which will be connected to, including scheme, host, port, path.<br/>**Supports Expression Language: true**|
-|send-message-body|true||If true, sends the HTTP message body on POST/PUT/PATCH requests (default).  If false, suppresses the message body and content-type header for these requests.|
+|send-message-body|true||DEPRECATED. Only kept for backwards compatibility, no functionality is included.|
+|Send Message Body|true||If true, sends the HTTP message body on POST/PUT/PATCH requests (default).  If false, suppresses the message body and content-type header for these requests.|
 |SSL Context Service|||The SSL Context Service used to provide client certificate information for TLS/SSL (https) connections.|
 |Use Chunked Encoding|false||When POST'ing, PUT'ing or PATCH'ing content set this property to true in order to not pass the 'Content-length' header and instead send 'Transfer-Encoding' with a value of 'chunked'. This will enable the data transfer mechanism which was introduced in HTTP 1.1 to pass data of unknown lengths in chunks.|
 ### Relationships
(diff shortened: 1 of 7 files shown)


arpadboda (approved):

extensions/http-curl/processors/InvokeHTTP.cpp
```diff
@@ -95,9 +95,12 @@ core::Property InvokeHTTP::ContentType("Content-type", "The Content-Type to spec
                                        "POST or PATCH. In the case of an empty value after evaluating an expression language expression, "
                                        "Content-Type defaults to",
                                        "application/octet-stream");
-core::Property InvokeHTTP::SendBody("send-message-body", "If true, sends the HTTP message body on POST/PUT/PATCH requests (default).  "
-                                    "If false, suppresses the message body and content-type header for these requests.",
-                                    "true");
+core::Property InvokeHTTP::SendBody(
+    core::PropertyBuilder::createProperty("send-message-body", "Send Body")
```
szaszm, line 99: NiFi has a similar property called "Send Message Body". I would rename this to match.

PROCESSORS.md
```diff
@@ -684,15 +684,17 @@ In the list below, the names of required properties appear in bold. Any other pr
...
 |Include Date Header|true||Include an RFC-2616 Date header in the request.|
+|invokehttp-proxy-password|||Password to set when authenticating against proxy|
+|invokehttp-proxy-username|||Username to set when authenticating against proxy|
+|Penalize on "No Retry"|false||Enabling this property will penalize FlowFiles that are routed to the "No Retry" relationship.|
 |Proxy Host|||The fully qualified hostname or IP address of the proxy server|
 |Proxy Port|||The port of the proxy server|
+|Put Response Body in Attribute|||If set, the response body received back will be put into an attribute of the original FlowFile instead of a separate FlowFile. The attribute key to put to is determined by evaluating value of this property.|
 |Read Timeout|15 secs||Max wait time for response from remote service.|
 |Remote URL|||Remote URL which will be connected to, including scheme, host, port, path.<br/>**Supports Expression Language: true**|
+|send-message-body|true||If true, sends the HTTP message body on POST/PUT/PATCH requests (default).  If false, suppresses the message body and content-type header for these requests.|
```
szaszm, line 695: I would prefer to have the version with the capitals and spaces here and in the config, since this is facing the user and most of our properties are like that.

lordgamez: The version with capitals and spaces is just the "display name" of the property and cannot be used in a configuration, so I would not use it in the documentation. We could rename it, but even though it did not have any functionality behind it before, the property was available and could break some previous flows using it in the configuration.

extensions/http-curl/processors/InvokeHTTP.cpp
lordgamez: Updated in e27db873853dc331d7c9c20462eb9038f537a6a3

PROCESSORS.md
fgerlits: I think it's worth renaming all of these properties while we are still in a 0.x version and we can do it.  "Send Message Body", "Proxy Username", "Proxy Password" would match NiFi and it would be more consistent internally in MiNiFi, too.  Currently, we have "Proxy Host", "Proxy Port", but "invokehttp-proxy-username", "invokehttp-proxy-password", which is confusing.

lordgamez: I agree it would be more consistent and more intuitive to use those names. If property renaming is not a problem under version 1.0 then I'll change these properties.

lordgamez: Update in 54c835ee681bea18cc7172b309181d56e5ab5439

fgerlits (approved):

PROCESSORS.md
lordgamez: Updated in 321d274fd111d0fc64eff092df50fcbb68e6acd7 by deprecating the old property in favor of the new `Send Message Body` property.

szaszm (approved):

fgerlits (approved):

PR #1042, opened 2021-04-06
MINIFICPP-1457 - Implement InputRequirements

adam-markovics:

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/include/core/Annotation.h
```diff
@@ -0,0 +1,77 @@
...
+#include <string>
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace core {
+
+template <typename T>
+struct AnnotationBase {
```
adamdebreceni, line 30: are more kinds of annotations in the works? if so, we could drop the default definition, and specialize on `Annotation<EInputRequirement>` thus we wouldn't need the `InputRequirementAnnotationBase` anymore

extensions/standard-processors/processors/TailFile.h
```diff
@@ -71,7 +71,7 @@ enum class Mode {
   SINGLE, MULTIPLE, UNDEFINED
 };
 
-class TailFile : public core::Processor {
+class TailFile : public core::Processor, public core::InputRequirementAnnotation<core::EInputRequirement::INPUT_FORBIDDEN> {
```
adamdebreceni, line 74: we could introduce an `annotation` namespace in `core` and provide type aliases for these, like `core::annotation::InputForbidden`, this might make it more compact

libminifi/include/core/Annotation.h
adam-markovics: It's possible that we will have more annotations in the future. We can get rid of AnnotationBase, but not InputRequirementAnnotationBase, as a non-template base class in necessary for the dynamic cast in Processor::getInputRequirement().

adamdebreceni: is `dynamic_cast<const Annotation<EInputRequirement>*>(_)` malformed?

extensions/standard-processors/processors/TailFile.h
adam-markovics: As we could have more types of annotations (other than input), I slightly modified it to:
core::annotation::input::Forbidden

libminifi/include/core/Annotation.h
adam-markovics: Okay, I understand now, it's done.

libminifi/src/core/Processor.cpp
```diff
@@ -386,6 +379,39 @@ std::shared_ptr<Connectable> Processor::pickIncomingConnection() {
...
+      }
+      break;
+    }
+  }
+}
+
+EInputRequirement Processor::getInputRequirement() const {
+  // default input requirement
+  EInputRequirement myInputRequirement = EInputRequirement::INPUT_ALLOWED;
+  // get own input annotation if InputRequirementAnnotationBase is a base class
```
adamdebreceni, line 406: this comment got out of sync with the removal of `InputRequirementAnnotationBase`

adam-markovics: Done.

libminifi/src/ThreadedSchedulingAgent.cpp
```diff
@@ -88,6 +88,7 @@ void ThreadedSchedulingAgent::schedule(std::shared_ptr<core::Processor> processo
 
   auto sessionFactory = std::make_shared<core::ProcessSessionFactory>(processContext);
 
+  processor->validateAnnotations();
```
adamdebreceni, line 91: given our direction towards all kinds of flow-verification, this could be moved into either a `FlowController::verifyFlow(ProcessGroup&)` or `ProcessGroup::verify` (the former might be better as then we could access the services and repositories all at once) which could then be expanded as needed (calling it from `FlowController::load`)

libminifi/src/core/Processor.cpp
```diff
@@ -386,6 +379,39 @@ std::shared_ptr<Connectable> Processor::pickIncomingConnection() {
   return getNextIncomingConnectionImpl(rel_guard);
 }
 
+using annotation::input::EInputRequirement;
+
+void Processor::validateAnnotations() const {
```
adamdebreceni, line 384: some unit tests could 'secure' this method

.github/workflows/ci.yml
```diff
@@ -218,7 +218,7 @@ jobs:
           sudo ln -s /usr/lib/x86_64-linux-gnu/odbc/libsqlite3odbc.so /usr/lib/x86_64-linux-gnu/libsqlite3odbc.so
           echo "PATH=/usr/lib/ccache:$PATH" >> $GITHUB_ENV
       - id: build
-        run: sudo mount tmpfs -t tmpfs /tmp && ./bootstrap.sh -e -t && cd build  && cmake -DUSE_SHARED_LIBS= -DENABLE_OPENWSMAN=ON -DENABLE_OPENCV=ON -DENABLE_MQTT=ON -DENABLE_GPS=ON -DENABLE_USB_CAMERA=ON -DENABLE_LIBRDKAFKA=ON -DENABLE_OPC=ON -DENABLE_SFTP=ON -DENABLE_MQTT=ON -DENABLE_COAP=ON -DENABLE_PYTHON=ON -DENABLE_SQL=ON -DENABLE_AWS=ON -DENABLE_AZURE=ON -DSTRICT_GSL_CHECKS=AUDIT -DFAIL_ON_WARNINGS=ON .. &&  cmake --build . --parallel 4  && make test ARGS="--timeout 300 -j8 --output-on-failure"
```
arpadboda, line 221: Why remove MQTT?
Even if we don't test it, at least the build doesn't get broken.

extensions/standard-processors/processors/ExecuteProcess.h
```diff
@@ -51,7 +51,7 @@ namespace processors {
 #ifndef WIN32
 
 // ExecuteProcess Class
-class ExecuteProcess : public core::Processor {
+class ExecuteProcess : public core::Processor, public core::annotation::input::Forbidden {
```
arpadboda, line 54: I think this should allow - the process to be executed can be speficied in flowfile attributes and whatnot

.github/workflows/ci.yml
adam-markovics: Because it was duplicated.

extensions/standard-processors/processors/ExecuteProcess.h
adamdebreceni: it can be specified but it seems to me, that it won't be used, this might be a bug

adam-markovics: Input is also forbidden in NiFi, I will ask them what they think about this.

.github/workflows/ci.yml
arpadboda: Okay, makes sense :)

extensions/standard-processors/processors/ExecuteProcess.h
adam-markovics: ExecuteStreamCommand is the one in NiFi that accepts commands from an incoming flowfile. We don't have that (yet?).

arpadboda: Yep, makes sense to differentiate the two, but that's definitely out of scope of this change. 
Leave it as "allowed" for now, later we can align the processors to NiFi

lordgamez (approved):
LGTM

extensions/standard-processors/processors/ExecuteProcess.h
adam-markovics: Done.

libminifi/src/core/Processor.cpp
adam-markovics: Done.

libminifi/src/ThreadedSchedulingAgent.cpp
adam-markovics: Done.

arpadboda (approved):

libminifi/include/core/Annotation.h
```diff
@@ -0,0 +1,79 @@
...
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace core {
+namespace annotation {
+
+template <typename T>
+struct Annotation {
+  const T value_;
```
szaszm, line 32: This implicitly disables move, because it's a non-static const data member, inherited by the . It should be `static constexpr` IMO. It also pollutes the scope of derived classes with an overly generic name `value_`, because it's all used through inheritance.

I don't like using inheritance for this purpose. I would prefer if classes could just declare their requirements by creating a static const member or a type alias of a well known name and type, like `struct MyProcessor : Processor { using input_requirements = annotation::input::required; };`.

szaszm (commented):
While the PR looks functional, and the functionality is very appreciated, I have issues with the design. See my comment below.

libminifi/include/core/Processor.h
```diff
@@ -251,6 +252,10 @@ class Processor : public Connectable, public ConfigurableComponent, public std::
 
   std::shared_ptr<Connectable> pickIncomingConnection() override;
 
+  virtual void validateAnnotations() const;
```
adamdebreceni, line 255: are there any plans for processor-specific annotation validation making this method`virtual`?

libminifi/include/core/Annotation.h
adamdebreceni: compile-time would be a nice touch, but I don't see how could we verify at compile-time that a processor requiring inputs actually gets some input connections, also wouldn't this require the code to a) either cast all processors to their most derived type b) reimplement the input requirement validation for all processor types?

libminifi/include/core/Processor.h
adam-markovics: It's possible for specific processors to override this method. But because it's not done yet, I will remove this virtual specifier from here, it could be reintroduced later if needed.

libminifi/include/core/Annotation.h
adamdebreceni: agree with the member pollution concern, we could get rid of the `value_` altogether by trying successive `dynamic_cast`s to `Required`, `Allowed` and `Forbidden`

szaszm: Good point about the compile-time requirements, I didn't think about not having the static type being the most derived type. As a runtime-friendly alternative, we could avoid using inheritance by introducing a new virtual function, something along the lines of `getInputRequirements() const`, or passing the enum value to the `Processor` base constructor for storage.

libminifi/src/FlowController.cpp
```diff
@@ -302,6 +302,10 @@ void FlowController::load(const std::shared_ptr<core::ProcessGroup> &root, bool
       this->root_ = std::shared_ptr<core::ProcessGroup>(loadInitialFlow());
     }
 
+    if (root) {
+      root_->verify();
```
adamdebreceni, line 306: shouldn't we check `root_`?

adam-markovics: Yes, we should. Big thanks for noticing the typo!

adamdebreceni (approved):

libminifi/src/core/Processor.cpp
```diff
@@ -386,6 +379,38 @@ std::shared_ptr<Connectable> Processor::pickIncomingConnection() {
   return getNextIncomingConnectionImpl(rel_guard);
 }
 
+void Processor::validateAnnotations() const {
+  switch (getInputRequirement()) {
+    case annotation::Input::INPUT_REQUIRED: {
+      if (!hasIncomingConnections()) {
+        throw Exception(PROCESS_SCHEDULE_EXCEPTION, "INPUT_REQUIRED was specified for the processor, but no inputs were found");
```
szaszm, line 386: I would rather refer to incoming connections in the error message.
```suggestion
        throw Exception(PROCESS_SCHEDULE_EXCEPTION, "INPUT_REQUIRED was specified for the processor, but no incoming connections were found");
```

libminifi/src/core/Processor.cpp
```diff
@@ -386,6 +379,38 @@ std::shared_ptr<Connectable> Processor::pickIncomingConnection() {
...
+      }
+      return;
+    }
+    case annotation::Input::INPUT_ALLOWED:
+      return;
+    case annotation::Input::INPUT_FORBIDDEN: {
+      if (hasIncomingConnections()) {
+        throw Exception(PROCESS_SCHEDULE_EXCEPTION, "INPUT_FORBIDDEN was specified for the processor, but at least one input was found");
```
szaszm, line 394:
```suggestion
        throw Exception(PROCESS_SCHEDULE_EXCEPTION, "INPUT_FORBIDDEN was specified for the processor, but there are incoming connections");
```

szaszm (approved):

adamdebreceni (approved):

PR #1043, opened 2021-04-06
MINIFICPP-1359 Implement FollowRedirects property in InvokeHTTP

lordgamez:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/http-curl/processors/InvokeHTTP.cpp
+++ b/extensions/http-curl/processors/InvokeHTTP.cpp
@@ -75,7 +75,11 @@ core::Property InvokeHTTP::ReadTimeout(
 core::Property InvokeHTTP::DateHeader(
     core::PropertyBuilder::createProperty("Include Date Header")->withDescription("Include an RFC-2616 Date header in the request.")->isRequired(false)->withDefaultValue<bool>(true)->build());
 
-core::Property InvokeHTTP::FollowRedirects("Follow Redirects", "Follow HTTP redirects issued by remote server.", "True");
+core::Property InvokeHTTP::FollowRedirects(
+  core::PropertyBuilder::createProperty("Follow Redirects")
+  ->withDescription("Follow HTTP redirects issued by remote server.")
+  ->withDefaultValue<bool>(true)
+  ->build());
 core::Property InvokeHTTP::AttributesToSend("Attributes to Send", "Regular expression that defines which attributes to send as HTTP"
                                             " headers in the request. If not defined, no attributes are sent as headers.",
                                             "");
@@ -154,6 +158,7 @@ void InvokeHTTP::initialize() {
   properties.insert(SendBody);
   properties.insert(DisablePeerVerification);
   properties.insert(AlwaysOutputResponse);
+  properties.insert(FollowRedirects);
 
   setSupportedProperties(properties);
   // Set the supported relationships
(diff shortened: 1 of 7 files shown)


extensions/http-curl/tests/VerifyInvokeHTTPTest.cpp
```diff
@@ -120,6 +126,31 @@ class VerifyInvokeHTTPOKResponse : public VerifyInvokeHTTP {
   }
 };
 
+class VerifyInvokeHTTPOK200Response : public VerifyInvokeHTTP {
+ public:
+  void runAssertions() override {
+    using org::apache::nifi::minifi::utils::verifyLogLinePresenceInPollTime;
+    assert(verifyLogLinePresenceInPollTime(std::chrono::seconds(6),
+        "key:invokehttp.status.code value:200",
+        "response code 200"));
```
adamdebreceni, line 135: which handler responds with `200`? we redirect to `/`, does the civetweb have some default handler on `/`?

lordgamez: It seems so. At first I wanted to make a separate handler for it, but after seeing it automatically returns `HTTP 200 OK` after redirecting to `/` and no other handler returns code `200` I decided to use that.

arpadboda (approved):

extensions/http-curl/processors/InvokeHTTP.cpp
```diff
@@ -75,7 +75,11 @@ core::Property InvokeHTTP::ReadTimeout(
 core::Property InvokeHTTP::DateHeader(
     core::PropertyBuilder::createProperty("Include Date Header")->withDescription("Include an RFC-2616 Date header in the request.")->isRequired(false)->withDefaultValue<bool>(true)->build());
 
-core::Property InvokeHTTP::FollowRedirects("Follow Redirects", "Follow HTTP redirects issued by remote server.", "True");
+core::Property InvokeHTTP::FollowRedirects(
+  core::PropertyBuilder::createProperty("Follow Redirects")
+  ->withDescription("Follow HTTP redirects issued by remote server.")
+  ->withDefaultValue<bool>(true)
```
arpadboda, line 81: According to Curl, this is false by default:
https://curl.se/libcurl/c/CURLOPT_FOLLOWLOCATION.html

So I think this is a breaking change, I would prefer to have false here by default.

lordgamez: That's true, but I was checking the usage of this property in NiFi's version of [InvokeHTTP](https://nifi.apache.org/docs/nifi-docs/components/org.apache.nifi/nifi-standard-nar/1.13.2/org.apache.nifi.processors.standard.InvokeHTTP/index.html) and there the default value is True. Should we stil prefer the Curl's version for the default value?

arpadboda: Hmm, good question. 
My reasoning was not based on Curl's default, but what we did before (which was based on Curl's default).
As minifi is still 0.xxx version, I tend to agree with you and introduce the breaking change to align with NiFi in this aspect.

adamdebreceni (approved):

arpadboda (approved):

szaszm (approved):

PR #1044, opened 2021-04-06
MINIFICPP-1526 Add ConsumeJournald to consume systemd journal messages

szaszm:
- [x] libsystemd might be problematic license-wise. I might have to work around it with `dlopen`.
- [x] Still figuring out what would be the best way to test this. 
- [x] Tests are failing, work in progress, but submitting the PR anyway to see the reaction and gather opinions.

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/systemd/WorkerThread.h
```diff
@@ -0,0 +1,66 @@
...
+
+  ~WorkerThread();
+
+  template<typename... Args>
+  void enqueue(Args&& ... args) { work_.enqueue(std::forward<Args>(args)...); }
+
+ private:
+  void run() noexcept;
+
+  utils::ConditionConcurrentQueue<std::packaged_task<void()>> work_;
```
fgerlits, line 44: minor, but I would call this `queue_` or `work_queue_`

fgerlits (commented):
just some minor comments

CMakeLists.txt
```diff
@@ -564,6 +560,14 @@ if (ENABLE_ALL OR ENABLE_AZURE)
 	createExtension(AZURE-EXTENSIONS "AZURE EXTENSIONS" "This enables Azure support" "extensions/azure" "${TEST_DIR}/azure-tests")
 endif()
 
+## Add the systemd extension
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+	option(DISABLE_SYSTEMD "Disables the systemd extension." OFF)
```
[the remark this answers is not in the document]
szaszm: The reason it's `DISABLE_SYSTEMD` is that it's enabled by default on supported platforms and I think the pattern is that the option is the opposite of the default. There's already an exception to this, but it might be an accident or just that it doesn't correspont to an extension (AFAIK): `ENABLE_OPS`.
Given this, do you still want to change it?

extensions/systemd/ConsumeJournald.cpp
```diff
@@ -0,0 +1,254 @@
...
+  const auto journal_type = (context->getProperty(JournalType) | utils::flatMap(parse_journal_type)).value_or(JournalTypeEnum::System);
+  const auto process_old_messages = context->getProperty<bool>(ProcessOldMessages).value_or(false);
+  timestamp_format_ = [&context] {
+    auto tf_prop = context->getProperty(TimestampFormat).value_or(TimestampFormat.getDefaultValue());
+    if (tf_prop == "ISO" || tf_prop == "ISO 8601" || tf_prop == "ISO8601") return std::string{"%FT%T%Ez"};
+    return tf_prop;
+  }();
+
+  state_manager_ = context->getStateManager();
+  journal_ = worker_->enqueue([this, journal_type]{ return libwrapper_->openJournal(journal_type); }).get();
```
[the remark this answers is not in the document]
szaszm: The journal functions are "thread-agnostic", meaning for a given journal handle, they must be called from the same thread. In practice, they probably rely on thread-local storage.
See the NOTES of [`man sd_journal_open`](https://www.man7.org/linux/man-pages/man3/sd_journal_open_namespace.3.html).

extensions/systemd/WorkerThread.h
szaszm: I think there is no difference in readability, so I'm sticking with the current name, but if you disagree, or I should change it for some other reason, I'm open to arguments.

extensions/systemd/ConsumeJournald.cpp
```diff
@@ -0,0 +1,254 @@
...
+    }
+  }).get();
+  running_ = true;
+}
+
+void ConsumeJournald::onTrigger(core::ProcessContext* const context, core::ProcessSession* const session) {
+  gsl_Expects(context && session);
+  if (!running_.load(std::memory_order_acquire)) return;
+  auto cursor_and_messages = getCursorAndMessageBatch().get();
+  state_manager_->set({{"cursor", std::move(cursor_and_messages.first)}});
```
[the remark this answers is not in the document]
szaszm: This is not so minor. I honestly don't remember all the gotchas of the state manager, so thanks for pointing this out.

extensions/systemd/WorkerThread.h
fgerlits: To me, this is a queue containing pieces of work, not (as the name suggests) a piece of work itself.  But `work_` is also OK if you like it better.

fgerlits (approved):

lordgamez (approved):

extensions/systemd/WorkerThread.h
szaszm: I thought of work as uncountable in this context, but the queue indeed contains units of work. I normally prefer not to repeat information in the name that's already in the type, but since you understood it differently than I, it's probably a good idea to rename it.
Renaming to `task_queue_`, because I realized that the units of work are tasks. Let me know if you don't like this name.
5147b60bb

extensions/systemd/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+include(${CMAKE_SOURCE_DIR}/extensions/ExtensionHeader.txt)
+
+add_library(minifi-systemd STATIC ConsumeJournald.cpp WorkerThread.cpp libwrapper/LibWrapper.h Common.h libwrapper/LibWrapper.cpp libwrapper/DlopenWrapper.cpp libwrapper/DlopenWrapper.h)
```
adamdebreceni, line 22: do we need to add the headers?

extensions/systemd/ConsumeJournald.cpp
```diff
@@ -0,0 +1,262 @@
...
+  gsl_Expects(context && sessionFactory && !running_ && worker_);
+  using JournalTypeEnum = systemd::JournalType;
+
+  const auto parse_payload_format = [](const std::string& property_value) -> utils::optional<systemd::PayloadFormat> {
+    if (utils::StringUtils::equalsIgnoreCase(property_value, PAYLOAD_FORMAT_RAW)) return systemd::PayloadFormat::Raw;
+    if (utils::StringUtils::equalsIgnoreCase(property_value, PAYLOAD_FORMAT_SYSLOG)) return systemd::PayloadFormat::Syslog;
+    return utils::nullopt;
+  };
+  const auto parse_journal_type = [](const std::string& property_value) -> utils::optional<JournalTypeEnum> {
+    if (utils::StringUtils::equalsIgnoreCase(property_value, JOURNAL_TYPE_USER)) return JournalTypeEnum::User;
```
adamdebreceni, line 100: it seems that `ConsumeKafka` is the only other processor supporting case-insensitive property values, we should keep property values case sensitive until there is apt reason to decide otherwise

libminifi/include/utils/gsl.h
```diff
@@ -26,6 +28,26 @@ namespace minifi {
 
 namespace gsl = ::gsl_lite;
 
+namespace utils {
+namespace detail {
+template<typename T>
+using remove_cvref_t = typename std::remove_cv<typename std::remove_reference<T>::type>::type;
```
adamdebreceni, line 34: `gsl_lite` already has a `remove_cvref`, I do favor this version, we could change the `gsl_lite` one in `optional_from_ptr` to this one

libminifi/src/core/ProcessSession.cpp
```diff
@@ -241,6 +241,18 @@ void ProcessSession::write(const std::shared_ptr<core::FlowFile> &flow, OutputSt
   }
 }
 
+void ProcessSession::writeBuffer(const std::shared_ptr<core::FlowFile>& flow_file, gsl::span<const char> buffer) {
+  struct BufferOutputStreamCallback : OutputStreamCallback {
```
adamdebreceni, line 245: I'd rather we move out this `SpanOutputStreamCallback`, create a forwarding `ProcessSession::write` that takes a reference so we don't have to create local callback variables, and call that

libminifi/include/utils/GeneralUtils.h
```diff
@@ -174,6 +174,21 @@ auto invoke(F&& f, Args&&... args) MINIFICPP_UTIL_DEDUCED(detail::invoke_impl(st
...
+#endif /* < C++17 */
+
+namespace detail {
+struct dereference_t {
+  template<typename T>
+  T &operator()(T *ptr) const noexcept { return *ptr; }
+};
+}  // namespace detail
+
+CPP17_INLINE constexpr detail::dereference_t dereference{};
```
adamdebreceni, line 190: why do we need the `inline`, don't const variables have internal linkage?

lordgamez:
When I try testing the processor manually on my local machine it fails with the following error:

`[2021-05-13 14:00:02.434] [org::apache::nifi::minifi::core::ProcessGroup] [error] Failed to start processor d9606075-a35a-4509-b705-13e793e1b82b (ConsumeJournald): sd_journal_get_cursor: Cannot assign requested address`

I could only find a similar [issue](https://github.com/timberio/vector/issues/1081) but I did not find a solution for it, so it still needs to be investigated.

extensions/systemd/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+set_property(TARGET minifi-systemd PROPERTY POSITION_INDEPENDENT_CODE ON)
+
+target_link_libraries(minifi-systemd ${LIBMINIFI} Threads::Threads date::date)
+
+set(SYSTEMD-EXTENSION minifi-systemd PARENT_SCOPE)
+register_extension(minifi-systemd)
```
martinzink, line 28: I think, we should add this extension to the linter check as well
```suggestion
register_extension(minifi-systemd)
register_extension_linter(minifi-systemd-extension-linter)
```

this also brings out a couple linter violations which should be fixed as well

CMakeLists.txt
```diff
@@ -564,6 +560,14 @@ if (ENABLE_ALL OR ENABLE_AZURE)
 	createExtension(AZURE-EXTENSIONS "AZURE EXTENSIONS" "This enables Azure support" "extensions/azure" "${TEST_DIR}/azure-tests")
 endif()
 
+## Add the systemd extension
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
```
martinzink, line 564: How does this behave on distros without systemd? e.g Artix

extensions/systemd/ConsumeJournald.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+  state_manager_ = context->getStateManager();
+  // All journal-related API calls are thread-agnostic, meaning they need to be called from the same thread. In our environment,
+  // where a processor can easily be scheduled on different threads, we ensure this by executing all library calls on a dedicated
+  // worker thread. This is why all such operations are dispatched to a thread and immediately waited for in the initiating thread.
+  journal_ = worker_->enqueue([this, journal_type]{ return libwrapper_->openJournal(journal_type); }).get();
+  const auto seek_default = [this, process_old_messages](libwrapper::Journal& journal) {
+    if (process_old_messages) journal.seekHead();
+    else journal.seekTail();
+    state_manager_->set({{"cursor", getCursor()}});
```
martinzink, line 124: When I tried to run the processor locally (both on Ubuntu 18.04 and manjaro) the onSchedule failed because ​this getCursor calls returned an error.
sd_journal_get_cursor: Cannot assign requested address

From the man I think we are missing a sd_journal_next call (I hacked it into here, and now it seems working)

From the man
```
Note that sd_journal_get_cursor() will not work before
      ​sd_journal_next(3) (or related call) has been called at least
      ​once, in order to position the read pointer at a valid entry.

``

File: extensions/systemd/WorkerThread.h [hunk 13]
```diff
@@ -0,0 +1,66 @@
...
+
+  utils::ConditionConcurrentQueue<std::packaged_task<void()>> task_queue_;
+  std::thread thread_;
+};
+}  // namespace detail
+
+/**
+ * A worker that executes arbitrary functions with no parameters asynchronously on an internal thread, returning a future to the result.
+ */
+class Worker final {
```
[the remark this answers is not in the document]
szaszm: My impression is that using `utils::ThreadPool` would complicate the implementation so much that it would be more complex than reimplementing the needed bits. Now the whole implementation fits on a screen.

libminifi/include/utils/GeneralUtils.h
szaszm: I didn't think about it, just followed a common pattern. Apparently inline variables have external linkage, so translation units share a single instance.
https://wiki.tum.de/download/attachments/93291100/Riedel%20report%20-%20Constexpr%20and%20Inline%20Variables%20in%20C%2B%2B17.pdf?version=1&modificationDate=1532345169027&api=v2
https://www.learncpp.com/cpp-tutorial/global-constants-and-inline-variables/#urvanov-syntax-highlighter-60a2d99848d04196510799

CMakeLists.txt
szaszm: If the processor is used, it would fail to `dlopen("libsystemd.so.0")` during `onSchedule`

szaszm: I don't have such a distro available to me right now, so if you do and don't mind verifying that it compiles, that would be appreciated. The extension should compile without issues on such distros.

libminifi/src/core/ProcessSession.cpp
szaszm: I don't fully understand what you mean here.
What would this forwarding `write` take?
What's the problem with local callback variables? Lambdas are ruled out because of the `OutputStreamCallback` interface, but a preference for local scope remains.

szaszm: What do you think about a `write` that takes a lambda and a `writeBuffer` that passes a lambda to this new `write`?

extensions/systemd/ConsumeJournald.cpp
szaszm: Removed this call in f936b60 and updated the test to not expect state before a flow file is produced. Now the initial state is only stored after the first flow file is produced. This should not be a problem, but is a slight behavior change.

libminifi/include/utils/gsl.h
szaszm: changed to `detail::remove_cvref_t` from our own gsl.h in f936b60

extensions/systemd/ConsumeJournald.cpp
szaszm: changed to case-sensitive in f936b60

extensions/systemd/ConsumeJournald.cpp
```diff
@@ -0,0 +1,262 @@
...
+    if (utils::StringUtils::equalsIgnoreCase(property_value, PAYLOAD_FORMAT_SYSLOG)) return systemd::PayloadFormat::Syslog;
+    return utils::nullopt;
+  };
+  const auto parse_journal_type = [](const std::string& property_value) -> utils::optional<JournalTypeEnum> {
+    if (utils::StringUtils::equalsIgnoreCase(property_value, JOURNAL_TYPE_USER)) return JournalTypeEnum::User;
+    if (utils::StringUtils::equalsIgnoreCase(property_value, JOURNAL_TYPE_SYSTEM)) return JournalTypeEnum::System;
+    if (utils::StringUtils::equalsIgnoreCase(property_value, JOURNAL_TYPE_BOTH)) return JournalTypeEnum::Both;
+    return utils::nullopt;
+  };
+  batch_size_ = context->getProperty<size_t>(BatchSize).value_or(10);
```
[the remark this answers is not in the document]
szaszm: changed to throw in f936b60, that should also produce a log message IIRC, will test later

libminifi/include/core/FlowFile.h
```diff
@@ -161,6 +161,15 @@ class FlowFile : public CoreComponent, public ReferenceContainer {
...
   }
+  bool setAttribute(std::string&& key, const std::string& value) {
+    return attributes_.insert_or_assign(std::move(key), value).second;
+  }
+  bool setAttribute(const std::string& key, std::string&& value) {
+    return attributes_.insert_or_assign(key, std::move(value)).second;
+  }
+  bool setAttribute(std::string&& key, std::string&& value) {
+    return attributes_.insert_or_assign(std::move(key), std::move(value)).second;
+  }
```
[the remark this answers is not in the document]
szaszm: done in f936b60

extensions/systemd/CMakeLists.txt
szaszm: No, but it doesn't hurt either. I think I've manually removed some before, but CLion automatically adds them when creating a class. Removed all headers in f936b60.

extensions/systemd/ConsumeJournald.cpp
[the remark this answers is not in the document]
szaszm: verified:
```
2021-05-18 17:39:21.870] [org::apache::nifi::minifi::core::ProcessGroup] [error] Failed to start processor 66fbb65d-8b11-43a9-aaad-746a07d2fe35 (ConsumeJournald): Processor Operation: invalid payload format
```

CMakeLists.txt
martinzink: checked it, compiles without an issue

martinzink (approved):

CMakeLists.txt
szaszm: Awesome, thank you

CMakeLists.txt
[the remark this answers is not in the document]
szaszm: changed in f1cd129

libminifi/include/utils/GeneralUtils.h
adamdebreceni: the link cites performance reasons for using `inline constexpr` instead of `constexpr`, specifically the memory reserved in each translation unit that includes this header, and the per-TU initialization time, coupled with the fact that now we have different linkage depending on the C++ version, I would say it is not worth it, and would categorize this as premature optimization, I would be curious what others think, or possibly some other non-performance reasons

libminifi/src/core/ProcessSession.cpp
adamdebreceni: I recall a refactor concerning taking addresses of non-lvalues, due to this we have some (not horribly many) instances of creating local callbacks for the sole purpose of taking their addresses, creating a `ProcessSession(const shrd_ptr<FF>&, OutputStreamCallback&&)` would handle these "forwarding" use-cases (where the temporary callbacks only forward the stream data in some implementation dependent manner),
I just would like to avoid the proliferation of "input/output-dependent" methods on `ProcessSession` (like `writeFile`, `writeRawBuffer`, `writeString`, etc.)

szaszm: > I just would like to avoid the proliferation of "input/output-dependent" methods on ProcessSession (like writeFileStream, writeRawBuffer, writeString, etc.)

I don't want this either, but a buffer is a very basic concept and it's expected by almost all write functions in existence. A `gsl::span` is just a pointer and a size. I believe `writeBuffer` is the most basic and generic interface possible to write to a flow file, and therefore it should be in `ProcessSession`.

> I recall a refactor concerning taking addresses of non-lvalues, due to this we have some (not horribly many) instances of creating local callbacks for the sole purpose of taking their addresses, creating a ProcessSession(const shrd_ptr<FF>&, OutputStreamCallback&&) would handle these "forwarding" use-cases (where the temporary callbacks only forward the stream data in some implementation dependent manner),

What's your opinion about this? https://pastebin.com/q7d67x1j

adamdebreceni: compared to the current implementation, I would go with the current one, agree that a buffer is quite fundamental, one other idea we could entertain is a `write(const shrd_ptr<FF>&, const Container&)` template, but we could generalize it later when there is actual need for it, for now `writeBuffer` seems to do the job

extensions/systemd/WorkerThread.h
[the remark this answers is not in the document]
szaszm: Here's a list of the data members of ThreadPool:
```
// determines if threads are detached
  bool daemon_threads_;
  std::atomic<int> thread_reduction_count_;
// max worker threads
  int max_worker_threads_;
// current worker tasks.
  std::atomic<int> current_workers_;
  std::atomic<int> task_count_;
// thread queue
  std::vector<std::shared_ptr<WorkerThread>> thread_queue_;
// manager thread
  std::thread manager_thread_;
// the thread responsible for putting delayed tasks to the worker queue when they had to be put
  std::thread delayed_scheduler_thread_;
// conditional that's used to adjust the threads
  std::atomic<bool> adjust_threads_;
// atomic running boolean
  std::atomic<bool> running_;
// controller service provider
  std::shared_ptr<core::controller::ControllerServiceProvider> controller_service_provider_;
// integrated power manager
  std::shared_ptr<controllers::ThreadManagementService> thread_manager_;
  // thread queue for the recently deceased threads.
  ConcurrentQueue<std::shared_ptr<WorkerThread>> deceased_thread_queue_;
// worker queue of worker objects
  ConditionConcurrentQueue<Worker<T>> worker_queue_;
  std::priority_queue<Worker<T>, std::vector<Worker<T>>, DelayedTaskComparator<T>> delayed_worker_queue_;
// mutex to  protect task status and delayed queue
  std::mutex worker_queue_mutex_;
// notification for new delayed tasks that's before the current ones
  std::condition_variable delayed_task_available_;
// map to identify if a task should be
  std::map<TaskId, bool> task_status_;
// manager mutex
  std::recursive_mutex manager_mutex_;
  // thread pool name
  std::string name_;
```

and this is the same for `systemd::detail::WorkerThread`:
```
  utils::ConditionConcurrentQueue<std::packaged_task<void()>> task_queue_;
  std::thread thread_;
```

I think this shows that these things are not the same. `systemd::Worker` is just a thread that executes functions in order. `utils::ThreadPool` is a scheduler with priority support, a thread pool, it can be stopped, resumed, and a lot more.

libminifi/include/utils/GeneralUtils.h
szaszm: removed `inline` in 0c45ec9

extensions/systemd/WorkerThread.h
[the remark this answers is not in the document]
adamdebreceni: I ment, what the added complexity would be on the usage side, but lets go with the new implementation

adamdebreceni (approved):

PR #1047, opened 2021-04-08
MINIFICPP-1503 Build Tensorflow extension in the CI

lordgamez:
- Fix warnings in Tensorflow extension
- Pull Tensorflow dependencies from docker image in CI
- Build Tensorflow extension in ubuntu 20.04 job
----------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


.github/workflows/ci.yml
```diff
@@ -215,8 +215,14 @@ jobs:
           sudo ln -s /usr/lib/x86_64-linux-gnu/odbc/libsqlite3odbc.so /usr/lib/x86_64-linux-gnu/libsqlite3odbc.so
           echo "PATH=/usr/lib/ccache:$PATH" >> $GITHUB_ENV
           echo -e "127.0.0.1\t$HOSTNAME" | sudo tee -a /etc/hosts > /dev/null
+          docker pull lordgamez/tensorflow-cc-binaries:ubuntu-20.04
```
fgerlits, line 218: I don't know anything about tensorflow, but is there really no more standard way to fetch these files than using your private docker image?  For example, using `pip` or some official tensorflow docker image?

lordgamez: I'm not familiar with tensorflow either, but as I checked the APIs and the distribution it seemed that the only way to use the C++ API is to build your own libraries from the tensorflow sources. There is an official docker image and pip packages, but those only provide python packages. Additionally we currently use the [tensorflow_cc](https://github.com/FloopCZ/tensorflow_cc) wrapper library instead of the official tensorflow libs in our extension, which also includes only the source files in its releases. For me it seemed the the most viable option to package these libraries in a docker image, this way we have a documentation of the builds process as well in the form of the Dockerfile. But I'm open for any additional ideas, if we have some better options.

szaszm: I'm not a fan of pulling binaries from a personal namespace either. Can we pull original sources and build it ourselves, possibly caching the result?

lordgamez: The problem is the build time takes too long build it in the CI. On my machine building it on 12 threads took almost 2 hours just to build the Tensorflow libraries. Even if we cached the results Caching the results would probably help, but any time it would get invalidated it would hold the CI flow for hours.

szaszm: Then maybe it's better to keep it disabled until we figure out a solution.

lordgamez: If it's okay then I'll open a separate PR with only the build fixes and keep this PR open to be changed after we come up a solution.

szaszm: Sounds good to me. :+1:

lordgamez: Moved build fixes to #1096

PR #1055, opened 2021-04-20
MINIFICPP-1454 Reduce duplication of CMake parameters in docker build

lordgamez:

### For all changes:
     in the commit message?

### For code changes:

### Note:


cmake/DockerConfig.cmake
```diff
@@ -22,7 +22,6 @@ add_custom_target(
         -u 1000
         -g 1000
         -v ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH}
-        -c IMAGE_TYPE=release
```
adamdebreceni, line 25: should we use `-i release` here as other targets use it like that? or should we remove `-i, -u, -g` wherever they align with the defaults in `DockerBuild.sh`?

lordgamez: I think you are right, it should be the same parameter list everywhere for the sake of consistency, I added the `-i` switch here as well in c802edc

adamdebreceni (approved):

docker/DockerBuild.sh
```diff
@@ -112,7 +80,13 @@ while [[ $# -gt 0 ]]; do
...
       IFS='=' read -ra ARR <<< "$2"
       if [[ ${#ARR[@]} -gt 1 ]]; then
-        declare "${ARR[0]}"="${ARR[1]}"
+        if [ "${ARR[0]}" == "BUILD_NUMBER" ]; then
+          BUILD_NUMBER="${ARR[1]}"
+        elif [ "${ARR[0]}" == "DOCKER_BASE_IMAGE" ]; then
+          BUILD_ARGS="${BUILD_ARGS} --build-arg BASE_ALPINE_IMAGE=${ARR[1]}"
+        else
+          BUILD_ARGS="${BUILD_ARGS} --build-arg ${ARR[0]}=${ARR[1]}"
+        fi
```
fgerlits, line 89: We could add a new parameter, eg. `-n`, for `BUILD_NUMBER` and change DockerConfig.cmake to run `.../DockerBuild.sh ... -n ${BUILD_NUMBER} ... -c BASE_ALPINE_IMAGE=${DOCKER_BASE_IMAGE}`, then we wouldn't need this if-else here.

lordgamez: That would require some additional changes in our other CI environments, where the BUILD_NUMBER is used, I'm not sure it would worth breaking those.

fgerlits (approved):

martinzink (approved):

PR #1064, opened 2021-04-27
MINIFICPP-987 Improve error handling in GetFile

fgerlits:
https://issues.apache.org/jira/browse/MINIFICPP-987

* Get rid of the access() calls and catch file read exceptions inside the loop in onTrigger().
* Get rid of the ProcessSession::import() call and replace it with write().
* Remove the unused FileOutputCallback class.
* Log file removal failures.  In this case, we will keep reading and sending the file, but that is better than not sending some files.
* Add a few more unit tests.

---

### For all changes:
     in the commit message?

### For code changes:

### Note:


lordgamez (commented):
Looks good, only some minor comments.

extensions/standard-processors/processors/GetFile.cpp
```diff
@@ -159,47 +148,54 @@ void GetFile::onSchedule(core::ProcessContext *context, core::ProcessSessionFact
...
-  logger_->log_debug("Is listing empty after polling directory %i", isDirEmptyAfterPoll);
+  const bool is_dir_empty_after_poll = isListingEmpty();
+  logger_->log_debug("Listing is %s after polling directory", is_dir_empty_after_poll ? "empty" : "not empty");
+  if (is_dir_empty_after_poll) {
+    yield();
+    return;
+  }
+
+  std::queue<std::string> list_of_file_names = pollListing(request_.batchSize);
+  while (!list_of_file_names.empty()) {
```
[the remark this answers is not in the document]
fgerlits: done in https://github.com/apache/nifi-minifi-cpp/pull/1064/commits/3a154cc5756502d6b20e953b6eb31edd157d7e78

extensions/standard-processors/tests/unit/GetFileTests.cpp
```diff
@@ -91,12 +113,30 @@ TEST_CASE("GetFile: MaxSize", "[getFileFifo]") {  // NOLINT
...
+}
+
+TEST_CASE("GetFile keeps the source file if KeepSourceFile is true") {
+  GetFileTestController test_controller;
+  test_controller.setProperty(processors::GetFile::KeepSourceFile, "true");
+
+  test_controller.runSession();
+
+  REQUIRE(utils::file::FileUtils::exists(test_controller.input_file_name_));
+}
```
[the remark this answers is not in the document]
fgerlits: I have created a new Jira for this: https://issues.apache.org/jira/browse/MINIFICPP-1555

libminifi/src/utils/FileReaderCallback.cpp
```diff
@@ -0,0 +1,72 @@
...
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace utils {
+
+FileReaderCallback::FileReaderCallback(const std::string& file_name)
+    : logger_(core::logging::LoggerFactory<FileReaderCallback>::getLogger()) {
+  logger_->log_debug("Opening %s", file_name);
+  input_stream_.open(file_name.c_str(), std::fstream::in | std::fstream::binary);
+  if (!input_stream_.is_open() || !input_stream_.good()) {
```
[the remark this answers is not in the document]
fgerlits: I copied this mindlessly from `ProcessSession`, but I think you're right, `is_open()` and `good()` are the same immediately after `open()`, so we were checking it twice.  I have removed the `good()` check, because I think `is_open()` reads better in combination with `open()`.

extensions/standard-processors/processors/GetFile.cpp
```diff
@@ -158,76 +147,89 @@ void GetFile::onSchedule(core::ProcessContext *context, core::ProcessSessionFact
...
       }
-    } catch (std::exception &exception) {
-      logger_->log_debug("GetFile Caught Exception %s", exception.what());
-      throw;
-    } catch (...) {
-      throw;
     }
+  } catch (const std::exception& exception) {
+    logger_->log_error("GetFile caught exception while processing file '%s': %s", file_name, exception.what());
+    flow_file->setDeleted(true);
```
arpadboda, line 200: Do we need this? 
I would expect session->rollback() later to clean up all the "mess" we made in an interrupted onTrigger

extensions/standard-processors/processors/GetFile.cpp
```diff
@@ -158,76 +147,89 @@ void GetFile::onSchedule(core::ProcessContext *context, core::ProcessSessionFact
...
-  // Perform directory list
-
+void GetFile::onTrigger(core::ProcessContext* /*context*/, core::ProcessSession* session) {
   metrics_->iterations_++;
 
-  const bool isDirEmptyBeforePoll = isListingEmpty();
-  logger_->log_debug("Is listing empty before polling directory %i", isDirEmptyBeforePoll);
-  if (isDirEmptyBeforePoll) {
+  const bool is_dir_empty_before_poll = isListingEmpty();
+  logger_->log_debug("Listing is %s before polling directory", is_dir_empty_before_poll ? "empty" : "not empty");
```
arpadboda, line 154: Nitpicking: "empty" can be part of the format string, so this line can be 5 chars shorter :))

arpadboda (commented):
Mostly looks good to me, added one comment/question

extensions/standard-processors/processors/GetFile.cpp
fgerlits: We don't want to roll back if reading one file failed but another file succeeded; in this case, we want to ignore the failed file but process the successful one.

My understanding is that every flow file we create with `session->create()` has to either be transferred to an output relationship or marked as deleted, otherwise `routeFlowFile(flow_file)`, and therefore `session->commit()`, will throw.

extensions/standard-processors/processors/GetFile.cpp
fgerlits: I'll change it if you think saving 5 chars in the line (and 5 bytes in the code) is worth it, but I think it is more readable this way.

extensions/standard-processors/processors/GetFile.cpp
arpadboda: Okay, I need more coffee, didn't realise that this is a different scenario. :)

arpadboda (approved):

extensions/standard-processors/processors/GetFile.cpp
arpadboda: No, deffinitely not, leave it as it is. :)

extensions/standard-processors/processors/GetFile.cpp
```diff
@@ -158,76 +147,89 @@ void GetFile::onSchedule(core::ProcessContext *context, core::ProcessSessionFact
...
+      if (remove_status != 0) {
+        logger_->log_error("GetFile could not delete file '%s', error %d: %s", file_name, errno, strerror(errno));
       }
-    } catch (std::exception &exception) {
-      logger_->log_debug("GetFile Caught Exception %s", exception.what());
-      throw;
-    } catch (...) {
-      throw;
     }
+  } catch (const std::exception& exception) {
```
adamdebreceni, line 198: we could limit the kinds of exceptions we consider "recoverable" (i.e. discard that specific flowfile, but continue processor execution) we could go for `std::ios_base::failure`

libminifi/src/utils/FileReaderCallback.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+    const int len = gsl::narrow<int>(num_bytes_read);
+
+    output_stream->write(reinterpret_cast<uint8_t*>(buffer.data()), len);
+    num_bytes_written += len;
+  }
+  input_stream_.close();
+
+  logger_->log_debug("Finished reading %" PRIu64 " bytes from the file", num_bytes_written);
+  return num_bytes_written;
```
adamdebreceni, line 65: do we set the stream somewhere to throw on `failure`?

martinzink (approved):

extensions/standard-processors/processors/GetFile.cpp
fgerlits: Good point.  Fixed in https://github.com/apache/nifi-minifi-cpp/pull/1064/commits/0ea95c9fd071ab701a730c29abed60806b171b34; now non-IO exceptions are no longer caught, they will cause a rollback.

libminifi/src/utils/FileReaderCallback.cpp
fgerlits: There are problems with `exceptions()`, mainly that `read()` will set the failbit when it reads the last chunk of the file.  I ended up following the advice of https://gehrcke.de/2011/06/reading-files-in-c-using-ifstream-dealing-correctly-with-badbit-failbit-eofbit-and-perror/ in https://github.com/apache/nifi-minifi-cpp/pull/1064/commits/0ea95c9fd071ab701a730c29abed60806b171b34 (with some modifications because `read()` and `getline()` work slightly differently).

adamdebreceni: okay, now I see that an error during open sets the `failbit` (so we would ask for an exception on that bit) but `failbit` is also set during `eof`, but then we wouldn't want an exception

adamdebreceni (approved):

szaszm:
Should've been closed by 34603d0, but didn't close due to a github glitch.

PR #1071, opened 2021-05-13
MINIFICPP-1548 - Use the same rocksdb instance for multiple repositories

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/rocksdb-repos/DatabaseContentRepository.cpp
```diff
@@ -41,14 +42,17 @@ bool DatabaseContentRepository::initialize(const std::shared_ptr<minifi::Configu
...
   }
-  rocksdb::Options options;
-  options.create_if_missing = true;
-  options.use_direct_io_for_flush_and_compaction = true;
-  options.use_direct_reads = true;
-  options.merge_operator = std::make_shared<StringAppender>();
-  options.error_if_exists = false;
-  options.max_successive_merges = 0;
-  db_ = utils::make_unique<minifi::internal::RocksDatabase>(options, directory_);
+  auto db_opts = [] (internal::Writable<rocksdb::DBOptions>& db_opts) {
```
arpadboda, line 45: Nitpicking: as these are going to be functions we call later, I would prefer them to be named accordingly. "set_db_opts" for eg.

extensions/rocksdb-repos/controllers/RocksDbPersistableKeyValueStoreService.cpp
```diff
@@ -58,18 +58,18 @@ void RocksDbPersistableKeyValueStoreService::onEnable() {
...
-  options.write_buffer_size = 8 << 20;
-  options.max_write_buffer_number = 20;
-  options.min_write_buffer_number_to_merge = 1;
-  if (!always_persist_) {
-    options.manual_wal_flush = true;
-  }
-  db_ = utils::make_unique<minifi::internal::RocksDatabase>(options, directory_);
+  auto cf_opts = [] (minifi::internal::Writable<rocksdb::ColumnFamilyOptions>& cf_opts) {
+    cf_opts.set(&rocksdb::ColumnFamilyOptions::write_buffer_size, 8ULL << 20U);
+    cf_opts.set<int>(&rocksdb::ColumnFamilyOptions::max_write_buffer_number, 20);
```
arpadboda, line 69: Is this the cf used to store the state of processors?
In case it is, I would reduce the numbers here, that definitely doesn't need such huge buffers.

extensions/rocksdb-repos/database/OpenRocksDb.cpp
```diff
@@ -0,0 +1,116 @@
...
+    db_->invalidate();
+  }
+  return result;
+}
+
+std::vector<rocksdb::Status> OpenRocksDb::MultiGet(const rocksdb::ReadOptions& options, const std::vector<rocksdb::Slice>& keys, std::vector<std::string>* values) {
+  std::vector<rocksdb::Status> results = impl_->MultiGet(
+      options, std::vector<rocksdb::ColumnFamilyHandle*>(keys.size(), column_->handle.get()), keys, values);
+  for (const auto& result : results) {
+    if (result == rocksdb::Status::NoSpace()) {
```
arpadboda, line 52: How can we get no space when we try to get?
On the other side, is that the only error to trigger breaking?

extensions/rocksdb-repos/database/OpenRocksDb.cpp
```diff
@@ -0,0 +1,116 @@
...
+      db_->invalidate();
+      break;
+    }
+  }
+  return results;
+}
+
+rocksdb::Status OpenRocksDb::Write(const rocksdb::WriteOptions& options, internal::WriteBatch* updates) {
+  rocksdb::Status result = impl_->Write(options, &updates->impl_);
+  if (result == rocksdb::Status::NoSpace()) {
```
arpadboda, line 62: Idea: maybe this should be moved to a HandleResult(result) func or something like that, so we don't need to copy-paste and easier to be extended with different results.

extensions/rocksdb-repos/database/RocksDatabase.cpp
```diff
@@ -0,0 +1,91 @@
...
+    if (pos == std::string::npos) {
+      pos = path.find_last_of('\\');
+    }
+    if (pos == std::string::npos) {
+      logger_->log_error("Couldn't detect the column name in '%s'", uri);
+      return nullptr;
+    }
+    db_path = path.substr(0, pos);
+    db_column = path.substr(pos + 1);
+    logger_->log_trace("Using column '%s' in rocksdb database '%s'", db_column, db_path);
```
arpadboda, line 52: I think these might be more verbose than trace as they are only printed once and can be useful.

arpadboda (commented):
Looks good overall, added some minor comments.

extensions/rocksdb-repos/controllers/RocksDbPersistableKeyValueStoreService.cpp
adamdebreceni: removed the `max_write_buffer_number` which makes it default to 2

extensions/rocksdb-repos/database/OpenRocksDb.cpp
adamdebreceni: factored out error handling

extensions/rocksdb-repos/database/RocksDatabase.cpp
adamdebreceni: changed to info

martinzink:
the changes look good :+1: , but why did you revert to using separate dbs by default in https://github.com/apache/nifi-minifi-cpp/pull/1071/commits/d787a926dafcbd70074bbd3c9b94ab1a1a3ca3c8 ?
I also feel like if this shared db isnt the only option then we should document this `minifidb://` scheme and how to set these (nifi.flowfile.repository.directory.default, nifi.database.content.repository.directory.default, nifi.state.manangement.provider.local.path) values to use separate/same dbs (minifi.properties or README ?)

adamdebreceni:
> the changes look good 👍 , but why did you revert to using separate dbs by default in [d787a92](https://github.com/apache/nifi-minifi-cpp/commit/d787a926dafcbd70074bbd3c9b94ab1a1a3ca3c8) ?
> I also feel like if this shared db isnt the only option then we should document this `minifidb://` scheme and how to set these (nifi.flowfile.repository.directory.default, nifi.database.content.repository.directory.default, nifi.state.manangement.provider.local.path) values to use separate/same dbs (minifi.properties or README ?)

added documentation to `CONFIGURE.md`, adding the possibility and making it default are (IMO) two different decisions (the latter might not even be justified now), I feel like if we decide to change the default behavior we should do it in a separate PR

martinzink (approved):

extensions/rocksdb-repos/database/OpenRocksDb.cpp
arpadboda: Okay, makes sense, thanks!

arpadboda (approved):

CONFIGURE.md
```diff
@@ -166,6 +166,25 @@ folder. You may specify your own path in place of these defaults.
      nifi.flowfile.repository.directory.default=${MINIFI_HOME}/flowfile_repository
 	 nifi.database.content.repository.directory.default=${MINIFI_HOME}/content_repository
 
+It is also possible to use a single database to store multiple repositories with the `minifidb://` scheme.
+This could help with migration and centralize agent state persistence.
+
+     in minifi.properties
+     nifi.flowfile.repository.directory.default=minifidb://${MINIFI_HOME}/agent_state/flowfile
+	 nifi.database.content.repository.directory.default=minifidb://${MINIFI_HOME}/agent_state/content
+	 nifi.state.manangement.provider.local.path=minifidb://${MINIFI_HOME}/agent_state/processor_states
```
szaszm, line 175: What does the `minifidb://` scheme exactly mean? Does it interpret the last path segment as a column family and the segments before as a path to the database?

libminifi/include/utils/GeneralUtils.h
```diff
@@ -62,6 +62,15 @@ struct identity {
...
 #endif /* < C++20 */
 
+#if __cplusplus > 201703L
+using std::type_identity;
+#else
+template<typename T>
+struct type_identity {
+  using type = T;
+};
+#endif /* < C++20 */
```
szaszm, line 72: There is a [feature test macro](https://en.cppreference.com/w/cpp/feature_test) for this: `__cpp_lib_type_identity`
```suggestion
#if __cpp_lib_type_identity >= 201806L
using std::type_identity;
#else
template<typename T>
struct type_identity {
  using type = T;
};
#endif /* has std::type_identity */
```
It's not used elsewhere because I believe feature test macros were standardized in C++20, so I see no point in checking for features that were introduced earlier. And I don't know the macro for std::identity above. (Maybe ranges?)

extensions/rocksdb-repos/database/RocksDbInstance.h
```diff
@@ -0,0 +1,71 @@
...
+#include "utils/OptionalUtils.h"
+#include "RocksDbUtils.h"
+#include "rocksdb/db.h"
+#include "logging/Logger.h"
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace internal {
```
szaszm, line 31: Did you consider putting the extension, or the new classes, under a different namespace? e.g. `org::apache::nifi::minifi::rocksdbrepos`

adamdebreceni: it is certainly doable, should we go with "rocksdbrepos"?

szaszm: I'm fine with either that, `extensions::rocksdbrepos`, `extensions::rocksdb` or anything similar.

adamdebreceni: as I started refactoring, I am now leaning more towards leaving it as is, if you don't feel strongly about it

szaszm: ok, that's fine

szaszm (approved):

arpadboda:
Merged in 1ce101f33806de45da69b0d67df7858226349ffb

PR #1073, opened 2021-05-13
MINIFICPP-1541 Move ODBC dependent SQL testing to the docker integration tests

lordgamez:
- Refactor SQL processors to remove explicit `soci` library dependancy
- Create MockDB for unit tests
- Add option to use real SQLite database in unit tests
- Add integration tests to test SQL processors with PostgreSQL database

Jira ticket: https://issues.apache.org/jira/browse/MINIFICPP-1541

-------------------------------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


win_build_vs.bat
```diff
@@ -65,7 +65,7 @@ for %%x in (%*) do (
...
+cmake -G %generator% -A %build_platform% -DINSTALLER_MERGE_MODULES=%installer_merge_modules% -DTEST_CUSTOM_WEL_PROVIDER=%test_custom_wel_provider% -DENABLE_SQL=%build_SQL% -DUSE_REAL_ODBC_TEST_DRIVER=ON -DCMAKE_BUILD_TYPE_INIT=%cmake_build_type% -DCMAKE_BUILD_TYPE=%cmake_build_type% -DWIN32=WIN32 -DENABLE_LIBRDKAFKA=%build_kafka% -DENABLE_JNI=%build_jni% -DOPENSSL_OFF=OFF -DENABLE_COAP=%build_coap% -DENABLE_AWS=%build_AWS% -DENABLE_PDH=%build_PDH% -DENABLE_AZURE=%build_azure% -DENABLE_SFTP=%build_SFTP%  -DUSE_SHARED_LIBS=OFF -DDISABLE_CONTROLLER=ON  -DBUILD_ROCKSDB=ON -DFORCE_WINDOWS=ON -DUSE_SYSTEM_UUID=OFF -DDISABLE_LIBARCHIVE=OFF -DDISABLE_SCRIPTING=ON -DEXCLUDE_BOOST=ON -DENABLE_WEL=TRUE -DFAIL_ON_WARNINGS=OFF -DSKIP_TESTS=%skiptests% %strict_gsl_checks% %redist% -DENABLE_LINTER=%build_linter% .. && msbuild /m nifi-minifi-cpp.sln /property:Configuration=%cmake_build_type% /property:Platform=%build_platform% && copy main\%cmake_build_type%\minifi.exe main\
```
adamdebreceni, line 68: as I understand this will always make the tests try to use the real odbc driver on windows, shouldn't we introduce some flag which is only default-enabled in the ci?

libminifi/test/sql-tests/mocks/MockConnectors.cpp
```diff
@@ -0,0 +1,386 @@
...
+
+std::unique_ptr<Rowset> MockStatement::execute(const std::vector<std::string>& args) {
+  MockDB db(file_path_);
+  return db.execute(query_, args);
+}
+
+MockODBCConnection::MockODBCConnection(std::string connectionString)
+    : connection_string_(std::move(connectionString)) {
+  std::smatch match;
+  std::regex expr("Database=(.*)(;|$)");
```
adamdebreceni, line 365: not clear whether we need it or not, but there is a `Regex` class in `RegexUtils.h`, aiming to address GCC<4.9 compilers, should we use that here as well?

libminifi/test/sql-tests/mocks/MockConnectors.cpp
```diff
@@ -0,0 +1,386 @@
...
+  std::regex_search(replaced_query, match, expr);
+  auto cols = minifi::utils::StringUtils::splitAndTrim(match[1], ",");
+  if (cols[0] == "*") {
+    cols = {};
+  }
+  std::string table_name = match[2];
+  std::string condition_str = match[4];
+  std::function<bool(const MockRow&)> condition;
+  if (!condition_str.empty()) {
+    if (condition_str == "int_col > 103") {
```
adamdebreceni, line 249: I feel like we should go for a different solution than hard-coded conditions, I agree that writing an expression parser would complicate things a lot, could we reuse some expression parsing utility from our codebase? (e.g. from Expression.cpp)

libminifi/test/sql-tests/mocks/MockConnectors.cpp
```diff
@@ -0,0 +1,386 @@
...
+std::string MockRow::getString(std::size_t index) const {
+  return column_values_.at(index);
+}
+
+double MockRow::getDouble(std::size_t index) const {
+  return std::atof(column_values_.at(index).c_str());
+}
+
+int MockRow::getInteger(std::size_t index) const {
+  return std::atoi(column_values_.at(index).c_str());
```
adamdebreceni, line 60: should we use the `std::string` accepting version? (`std::stoi` and others)

libminifi/test/sql-tests/mocks/MockConnectors.cpp
szaszm: Two problems with reusing Expression.cpp:
1. It's not available when compiling with expression language support disabled
2. It's parsing nifi expression language, but this looks like SQL to me

Not sure what would be the best way to solve this.

adamdebreceni: maybe we could cover our test use-cases with some well-crafted regex-es (and maybe calling further regex-es on submatches if need be)

libminifi/test/sql-tests/mocks/MockConnectors.h
```diff
@@ -0,0 +1,175 @@
...
+ public:
+  MockRow(std::vector<std::string>& column_names, std::vector<DataType>& column_types, const std::vector<std::string>& column_values)
+    : column_names_(column_names), column_types_(column_types), column_values_(column_values) {
+  }
+
+  MockRow(MockRow&& other) = default;
+  MockRow(const MockRow& other) = default;
+  MockRow& operator=(MockRow&& other) {
+    column_names_ = other.column_names_;
+    column_types_ = other.column_types_;
```
adamdebreceni, line 47: I don't know the exact intention here, but I wager it won't be doing that

adamdebreceni: using `std::reference_wrapper` would most likely offer the semantics we are looking for here

adamdebreceni: also won't the lack of `noexcept` make it unavailable for `std::vector`?

libminifi/test/sql-tests/mocks/MockConnectors.cpp
lordgamez: My thinking was that as we are already on the way of dropping support for GCC 4.8 and we already have a ticket [MINIFICPP-1508](https://issues.apache.org/jira/browse/MINIFICPP-1508) for removing the `Regex` utility class it wouldn't be ideal to add more functionality to it.

libminifi/test/sql-tests/mocks/MockConnectors.cpp
lordgamez: I'll investigate some simple parsing solutions for handling our current cases. I think we need to find a good balance between maintainability and expandability. I used a hard coded condition as a start because we currently do not have too many conditions and it can be expanded easily, but if we add way more then it will become bloated. On the other side if we write a complex parser the maintainability would be way too much effort. I'll try to write a simple parser that would be sufficient for our cases without too much maintenance costs.

win_build_vs.bat
lordgamez: You are right, that's a better solution, fixed in 856a4349dda249088cc9fa104646cd4843f28780

libminifi/test/sql-tests/mocks/MockConnectors.cpp
lordgamez: Updated in e905b1377f3689f285913e3886deab32a06cea4d

libminifi/test/sql-tests/mocks/MockConnectors.h
lordgamez: You right this was wrong. The original goal was just to make the class copyable, and `std::reference_wrapper` could have been a good solution here. Finally I decided it's simpler just to use a raw pointer in this case. Updated in ee9452ebd591ff22139606ba2cd682df912d14f4

szaszm: I'd say just use pointers. Whether you write `*column_names_` or `column_names_.get()` doesn't really matter. Raw pointers are not evil as observer pointers.

libminifi/test/sql-tests/mocks/MockConnectors.cpp
```diff
@@ -0,0 +1,386 @@
...
+  std::string condition_str = match[4];
+  std::function<bool(const MockRow&)> condition;
+  if (!condition_str.empty()) {
+    if (condition_str == "int_col > 103") {
+      condition = [&](const MockRow& row){ return std::atoi(row.getValue("int_col").c_str()) > 103; };
+    } else if (condition_str == "int_col > 102") {
+      condition = [&](const MockRow& row){ return std::atoi(row.getValue("int_col").c_str()) > 102; };
+    } else if (condition_str == "int_col = 11") {
+      condition = [&](const MockRow& row){ return std::atoi(row.getValue("int_col").c_str()) == 11; };
+    } else if (condition_str == "int_col = 11 and text_col = banana") {
```
szaszm, line 255: Shouldn't "banana" be in apostrophes here? Unless it's another column.

extensions/sql/data/DatabaseConnectors.h
```diff
@@ -20,72 +20,68 @@
...
+  virtual std::size_t size() const = 0;
+  virtual std::string getColumnName(std::size_t index) const = 0;
+  virtual bool isNull(std::size_t index) const = 0;
+  virtual DataType getDataType(std::size_t index) const = 0;
+  virtual std::string getString(std::size_t index) const = 0;
+  virtual double getDouble(std::size_t index) const = 0;
+  virtual int getInteger(std::size_t index) const = 0;
+  virtual long long getLongLong(std::size_t index) const = 0;
+  virtual unsigned long long getUnsignedLongLong(std::size_t index) const = 0;
+  virtual std::tm getDate(std::size_t index) const = 0;
```
szaszm, line 53: This could be a `std::chrono::time_point` or if you only need the date part, then `date::year_month_day`. These types are easier to work with.

libminifi/test/sql-tests/mocks/MockConnectors.cpp
lordgamez: When replacing arguments I did not add apostrophes so it worked like this. The code was removed in favor of 96c3a9dea8e0d9c0492ccf5850184755651418e9

libminifi/test/sql-tests/mocks/MockConnectors.cpp
lordgamez: Added a parser for simple where conditions in 96c3a9dea8e0d9c0492ccf5850184755651418e9

extensions/sql/data/DatabaseConnectors.h
lordgamez: The specialization for `row::get` for the date type is `std::tm` in the implementing [SOCI library](http://soci.sourceforge.net/doc/master/types/) that's why I sticked with that type.

libminifi/include/utils/StringUtils.h
```diff
@@ -137,6 +137,7 @@ class StringUtils {
 
   static std::vector<std::string> split(const std::string &str, const std::string &delimiter);
   static std::vector<std::string> splitAndTrim(const std::string &str, const std::string &delimiter);
+  static std::vector<std::string> splitAndTrimOnString(std::string str, const std::string& delimiter);
```
szaszm, line 140: How is this different from splitAndTrim?

lordgamez: `splitAndTrim` only uses the first character of the given string as a delimiter, so it actually is using a character as a delimiter not a string. I'm not sure if that should be updated to have a character as an input delimiter and keep 2 overloads or the functions should be merged.

szaszm: 3 reasonable ways of using the delimiter string if it's longer than a character is
1. Match any of the characters in the string. Strings like this would be reasonable: `" \t;:,"`. C# does this.
2. Match the whole delimiter string as a delimiter. This is the case with ...OnString now. JavaScript and Python does this.
3. Match the whole delimiter string as a regex. Java does this, but I wouln't go this route.

Taking a string and only using its first character is a bit strange to me, I would change that behavior. The fact that it's trimmed after splitting should be relatively separate, but convenience functions are OK.

I can't think of a good and simple solution that does it all and that I wouldn't want to change eventually, so I'm not making any suggestions, but I'm happy to provide my opinion if you want to change it. Please consider adding comments that clarify the behavior.

adamdebreceni: wow this is "great", I wondered why the following test passed knowing this:
```
TEST_CASE("TestStringUtils::split4", "[test split classname]") {
  std::vector<std::string> expected = { "org", "apache", "nifi", "minifi", "utils", "StringUtils" };
  REQUIRE(expected == StringUtils::split(org::apache::nifi::minifi::core::getClassName<org::apache::nifi::minifi::utils::StringUtils>(), "::"));
}
```
turns out two wrongs make a right here :D 
the current implementation implicitly discards zero length segments, so `split("a,,b", ",") == ["a", "b"]` instead of `["a", "", "b"]`, I would definitely change this, I would keep a single function with `std::string` delimiter but with the "correct" semantics, if we would like to interpret each character as a delimiter later, we could add a `splitAny` or something along those lines

lordgamez: Thanks for both of your inputs. I changed `split` and `splitAndTrim` to use full strings and changed the semantics to keep the zero length segments. I also added `*RemovingEmpty` versions for utility. Changes are incorporated in 3ee4f1357bf03ec5bed69a3ca63801d128e739f8

lordgamez: Because of the new semantics of `split` there are still some issues that need to be resolved, putting it to draft until it's fixed.

lordgamez: Fixed the remaining issues.

szaszm (commented):
just a few minor ideas

szaszm (approved):
I'm +1 even without the above suggestions applied

extensions/sql/data/SociConnectors.cpp
```diff
@@ -0,0 +1,181 @@
...
+  return current_->get<long long>(index);
+}
+
+unsigned long long SociRow::getUnsignedLongLong(std::size_t index) const {
+  return current_->get<unsigned long long>(index);
+}
+
+std::tm SociRow::getDate(std::size_t index) const {
+  return current_->get<std::tm>(index);
+}
```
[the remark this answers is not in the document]
lordgamez: I considered it, but as these functions needed to be generalized in a common base class (to be later implemented in the mock and the soci implementation) I went with the pure virtual functions for each getter in the base `Row` class.

libminifi/test/sql-tests/mocks/MockODBCService.h
```diff
@@ -0,0 +1,66 @@
...
+
+  explicit MockODBCService(const std::string &name, const std::shared_ptr<Configure> &configuration)
+      : DatabaseService(name),
+        logger_(logging::LoggerFactory<MockODBCService>::getLogger()) {
+    setConfiguration(configuration);
+    initialize();
+  }
+
+  std::unique_ptr<sql::Connection> getConnection() const {
+    return std::unique_ptr<sql::Connection>(new MockODBCConnection(connection_string_));
```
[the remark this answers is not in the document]
lordgamez: Updated in 167cdb692db9fb1a8f429a8845447f98ee155103

.github/workflows/ci.yml
```diff
@@ -36,7 +36,7 @@ jobs:
           export LDFLAGS="-L/usr/local/opt/flex/lib"
           export CPPFLAGS="-I/usr/local/opt/flex/include"
           # CPPFLAGS are not recognized by cmake, so we have to force them to CFLAGS and CXXFLAGS to have flex 2.6 working
-          ./bootstrap.sh -e -t && cd build  && cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_FLAGS="${CPPFLAGS} ${CFLAGS}" -DCMAKE_CXX_FLAGS="${CPPFLAGS} ${CXXFLAGS}" -DENABLE_LUA_SCRIPTING=1 -DENABLE_SQL=ON -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_RULE_MESSAGES=OFF -DSTRICT_GSL_CHECKS=AUDIT -DFAIL_ON_WARNINGS=OFF .. && cmake --build . --parallel 4
+          ./bootstrap.sh -e -t && cd build  && cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_FLAGS="${CPPFLAGS} ${CFLAGS}" -DCMAKE_CXX_FLAGS="${CPPFLAGS} ${CXXFLAGS}" -DENABLE_LUA_SCRIPTING=1 -DENABLE_SQL=ON -DUSE_REAL_ODBC_TEST_DRIVER=ON -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_RULE_MESSAGES=OFF -DSTRICT_GSL_CHECKS=AUDIT -DFAIL_ON_WARNINGS=ON .. && cmake --build . --parallel 4
```
lordgamez, line 39: Had some conflicts in the ci.yml while rebasing and I changed the `FAIL_ON_WARNINGS` flag to `ON` in the mac build and added the `ENABLE_AZURE=ON` option in the clang build. These options were included in the previous base and changed on the main branch, but I think both should be included in the CI jobs. If it's not a problem I would like to include these in this PR.

win_build_vs.bat
adamdebreceni: you seem to have added the `/RO` flag for the ci jobs, but the flag we are looking for in the bat file is `/ODBC`

lordgamez: The flag we use in the bat file is actually `/RO`
```
if [%%~x] EQU [/RO] set real_odbc=ON
```

```
| /RO | Use real ODBC driver in tests instead of mock SQL driver |
```

adamdebreceni: oh you are right, I was looking at the linked commit only

libminifi/test/sql-tests/mocks/MockConnectors.cpp
adamdebreceni: a comment might come in handy, that it parses an `and`-separated list of conditions in the format of `<column_name> [<>=] <value>`

lordgamez: Added comment in 182d8a534045a230857d7c5dc5116466f5177bcf

adamdebreceni (approved):

PR #1053, opened 2021-04-19
MINIFICPP-1373 - Add integration tests for ConsumeKafka and test  cleanup issues

hunyadi-dev:
Docker based integration tests for kafka, and fix for issues related to docker container cleanup between test scenario runs.

PR #1076, opened 2021-05-20
MINIFICPP-1373 - Add integration tests for ConsumeKafka and test cleanup issues

lordgamez:
Continuation of PR #1053 to stabilize tests. Additional changes include:

- Remove sleep for kafka tests which required removing the manual offset reset in ConsumeKafka. The manual offset reset could poll some messages from the kafka broker before the first `onTrigger` call and missing the first messages due to this. It also interfered with `ConsumeKafka`'s `Offset Reset` property.
- Fix some validators to shorten the test runtimes
- Add debug messages to minifi test logs
- Remove unnecessary kafka consumer container
- Change GetFile processor in python tests to not keep source files and only be triggered once, therefore assertions can be more precise
- Fixed flake8 issues
- Removed unused `subdir` parameters

------------------------------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/docker/DockerVerify.sh
+++ b/docker/DockerVerify.sh
@@ -82,6 +82,15 @@ exec
   behave "${BEHAVE_OPTS[@]}" "features/kafka.feature" -n "A MiNiFi instance transfers data to a kafka broker" &&
   behave "${BEHAVE_OPTS[@]}" "features/kafka.feature" -n "PublishKafka sends flowfiles to failure when the broker is not available" &&
   behave "${BEHAVE_OPTS[@]}" "features/kafka.feature" -n "PublishKafka sends can use SSL connect" &&
+  behave "${BEHAVE_OPTS[@]}" "features/kafka.feature" -n "MiNiFi consumes data from a kafka topic" &&
+  behave "${BEHAVE_OPTS[@]}" "features/kafka.feature" -n "ConsumeKafka parses and uses kafka topics and topic name formats" &&
+  behave "${BEHAVE_OPTS[@]}" "features/kafka.feature" -n "ConsumeKafka key attribute is encoded according to the \"Key Attribute Encoding\" property" &&
+  behave "${BEHAVE_OPTS[@]}" "features/kafka.feature" -n "Headers on consumed kafka messages are extracted into attributes if requested on ConsumeKafka" &&
+  behave "${BEHAVE_OPTS[@]}" "features/kafka.feature" -n "ConsumeKafka transactional behaviour is supported" &&
+  behave "${BEHAVE_OPTS[@]}" "features/kafka.feature" -n "Key attribute is encoded according to the \"Key Attribute Encoding\" property" &&
+  behave "${BEHAVE_OPTS[@]}" "features/kafka.feature" -n "Messages are separated into multiple flowfiles if the message demarcator is present in the message" &&
+  behave "${BEHAVE_OPTS[@]}" "features/kafka.feature" -n "The ConsumeKafka \"Maximum Poll Records\" property sets a limit on the messages processed in a single batch" &&
+  behave "${BEHAVE_OPTS[@]}" "features/kafka.feature" -n "Unsupported encoding attributes for ConsumeKafka throw scheduling errors" &&
   behave "${BEHAVE_OPTS[@]}" "features/s3.feature" -n "A MiNiFi instance transfers encoded data to s3" &&
   behave "${BEHAVE_OPTS[@]}" "features/s3.feature" -n "A MiNiFi instance transfers encoded data through a http proxy to s3" &&
   behave "${BEHAVE_OPTS[@]}" "features/s3.feature" -n "A MiNiFi instance can remove s3 bucket objects" &&
--- a/docker/DockerVerify.sh
+++ b/docker/DockerVerify.sh
@@ -69,35 +69,9 @@ export PYTHONPATH
 # Add --no-logcapture to see logs inline
 BEHAVE_OPTS=(-f pretty --logging-level INFO --logging-clear-handlers)
 
+# Specify feature or scenario to run a specific test e.g.:
+# behave "${BEHAVE_OPTS[@]}" "features/file_system_operations.feature"
+# behave "${BEHAVE_OPTS[@]}" "features/file_system_operations.feature" -n "Get and put operations run in a simple flow"
 cd "${docker_dir}/test/integration"
 exec
-  behave "${BEHAVE_OPTS[@]}" "features/file_system_operations.feature" -n "Get and put operations run in a simple flow" &&
-  behave "${BEHAVE_OPTS[@]}" "features/file_system_operations.feature" -n "PutFile does not overwrite a file that already exists" &&
-  behave "${BEHAVE_OPTS[@]}" "features/s2s.feature" -n "A MiNiFi instance produces and transfers data to a NiFi instance via s2s" &&
-  behave "${BEHAVE_OPTS[@]}" "features/s2s.feature" -n "Zero length files are transfered between via s2s if the \"drop empty\" connection property is false" &&
-  behave "${BEHAVE_OPTS[@]}" "features/s2s.feature" -n "Zero length files are not transfered between via s2s if the \"drop empty\" connection property is true" &&
-  behave "${BEHAVE_OPTS[@]}" "features/http.feature" -n "A MiNiFi instance transfers data to another MiNiFi instance with message body" &&
-  behave "${BEHAVE_OPTS[@]}" "features/http.feature" -n "A MiNiFi instance sends data through a HTTP proxy and another one listens" &&
-  behave "${BEHAVE_OPTS[@]}" "features/http.feature" -n "A MiNiFi instance and transfers hashed data to another MiNiFi instance" &&
-  behave "${BEHAVE_OPTS[@]}" "features/http.feature" -n "A MiNiFi instance transfers data to another MiNiFi instance without message body" &&
-  behave "${BEHAVE_OPTS[@]}" "features/kafka.feature" -n "A MiNiFi instance transfers data to a kafka broker" &&
-  behave "${BEHAVE_OPTS[@]}" "features/kafka.feature" -n "PublishKafka sends flowfiles to failure when the broker is not available" &&
-  behave "${BEHAVE_OPTS[@]}" "features/kafka.feature" -n "PublishKafka sends can use SSL connect" &&
-  behave "${BEHAVE_OPTS[@]}" "features/kafka.feature" -n "MiNiFi consumes data from a kafka topic" &&
-  behave "${BEHAVE_OPTS[@]}" "features/kafka.feature" -n "ConsumeKafka parses and uses kafka topics and topic name formats" &&
-  behave "${BEHAVE_OPTS[@]}" "features/kafka.feature" -n "ConsumeKafka key attribute is encoded according to the \"Key Attribute Encoding\" property" &&
-  behave "${BEHAVE_OPTS[@]}" "features/kafka.feature" -n "Headers on consumed kafka messages are extracted into attributes if requested on ConsumeKafka" &&
-  behave "${BEHAVE_OPTS[@]}" "features/kafka.feature" -n "ConsumeKafka transactional behaviour is supported" &&
-  behave "${BEHAVE_OPTS[@]}" "features/kafka.feature" -n "Key attribute is encoded according to the \"Key Attribute Encoding\" property" &&
-  behave "${BEHAVE_OPTS[@]}" "features/kafka.feature" -n "Messages are separated into multiple flowfiles if the message demarcator is present in the message" &&
-  behave "${BEHAVE_OPTS[@]}" "features/kafka.feature" -n "The ConsumeKafka \"Maximum Poll Records\" property sets a limit on the messages processed in a single batch" &&
-  behave "${BEHAVE_OPTS[@]}" "features/kafka.feature" -n "Unsupported encoding attributes for ConsumeKafka throw scheduling errors" &&
-  behave "${BEHAVE_OPTS[@]}" "features/s3.feature" -n "A MiNiFi instance transfers encoded data to s3" &&
-  behave "${BEHAVE_OPTS[@]}" "features/s3.feature" -n "A MiNiFi instance transfers encoded data through a http proxy to s3" &&
-  behave "${BEHAVE_OPTS[@]}" "features/s3.feature" -n "A MiNiFi instance can remove s3 bucket objects" &&
-  behave "${BEHAVE_OPTS[@]}" "features/s3.feature" -n "Deletion of a s3 object through a proxy-server succeeds" &&
-  behave "${BEHAVE_OPTS[@]}" "features/s3.feature" -n "A MiNiFi instance can download s3 bucket objects directly" &&
-  behave "${BEHAVE_OPTS[@]}" "features/s3.feature" -n "A MiNiFi instance can download s3 bucket objects via a http-proxy" &&
-  behave "${BEHAVE_OPTS[@]}" "features/s3.feature" -n "A MiNiFi instance can list an S3 bucket directly" &&
-  behave "${BEHAVE_OPTS[@]}" "features/s3.feature" -n "A MiNiFi instance can list an S3 bucket objects via a http-proxy" &&
-  behave "${BEHAVE_OPTS[@]}" "features/azure_storage.feature" -n "A MiNiFi instance can upload data to Azure blob storage"
+  behave "${BEHAVE_OPTS[@]}"
--- a/docker/DockerVerify.sh
+++ b/docker/DockerVerify.sh
@@ -66,6 +66,7 @@ export PATH
 PYTHONPATH="${PYTHONPATH}:${docker_dir}/test/integration"
 export PYTHONPATH
 
+# Add --no-logcapture to see logs inline
 BEHAVE_OPTS=(-f pretty --logging-level INFO --logging-clear-handlers)
 
 cd "${docker_dir}/test/integration"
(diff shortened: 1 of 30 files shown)


docker/test/integration/environment.py
```diff
@@ -11,20 +12,31 @@ def raise_exception(exception):
...
+    del test
 
 
 def before_scenario(context, scenario):
-    use_fixture(test_driver_fixture, context)
+    if "skip" in scenario.effective_tags:
+        scenario.skip("Marked with @skip")
+        return
+
+    logging.info("Integration test setup at {time:%H:%M:%S:%f}".format(time=datetime.datetime.now()))
```
fgerlits, line 25: I think the colon between `%S` and `%f` is a typo, it should be a dot (also at line 34)

docker/test/integration/steps/steps.py
```diff
@@ -121,15 +133,71 @@ def step_impl(context):
...
 def step_impl(context, property_name, processor_name, property_value):
     processor = context.test.get_node_by_name(processor_name)
-    processor.set_property(property_name, property_value)
+    if property_value == "(not set)":
+        processor.unset_property(property_name)
+    else:
+        processor.set_property(property_name, property_value)
+
+
+@given("the \"{property_name}\" of the {processor_name} processor is set to match {key_attribute_encoding} encoded kafka message key \"{message_key}\"")
```
fgerlits, line 167: minor, and partly old code, but I would add "property" after the property name in all three of these steps, as I think `And the "Keep Source File" property of the GetFile processor is set to "false"` would read better than `And the "Keep Source File" of the GetFile processor is set to "false"`.

docker/test/integration/features/http.feature
```diff
@@ -18,10 +19,28 @@ Feature: Sending data using InvokeHTTP to a receiver using ListenHTTP
...
+    And a InvokeHTTP processor with the "Remote URL" property set to "http://secondary:8080/contentListener"
+    And the "HTTP Method" of the InvokeHTTP processor is set to "POST"
+    And the "success" relationship of the GetFile processor is connected to the InvokeHTTP
+
+    And a ListenHTTP processor with the "Listening Port" property set to "8080" in a "secondary" flow
+    And a PutFile processor with the "Directory" property set to "/tmp/output" in the "secondary" flow
+    And the "success" relationship of the ListenHTTP processor is connected to the PutFile
+
+    When both instances start up
+    Then two flowfiles with the contents "first message" and "second message" are placed in the monitored directory in less than 120 seconds
```
fgerlits, line 39: can this scenario be added (when fixed) in a later PR?  it is not related to ConsumeKafka

docker/test/integration/features/https.feature
```diff
@@ -1,3 +1,5 @@
+# Failing feature, needs to be fixed
```
fgerlits, line 1: a `TODO` or `FIXME` in the comment would make it easier to find later

docker/test/integration/minifi/core/DockerTestDirectoryBindings.py
```diff
@@ -77,6 +77,7 @@ def put_file_contents(file_abs_path, contents):
         logging.info('Writing %d bytes of content to file: %s', len(contents), file_abs_path)
         with open(file_abs_path, 'ab') as test_input_file:
             test_input_file.write(contents)
+        os.chmod(file_abs_path, 0o0777)
```
fgerlits, line 80: OoO I did not know this alternate syntax, it's interesting.  But I think `0777` is more standard and more readable.

docker/test/integration/minifi/core/FileSystemObserver.py
```diff
@@ -31,16 +31,21 @@ def restart_observer_if_needed(self):
...
+            return
         wait_start_time = time.perf_counter()
-        for i in range(0, max_files):
+        for _ in range(0, max_files):
             # Note: The timing on Event.wait() is inaccurate
-            self.done_event.wait(timeout_seconds)
-            self.done_event.clear()
-            current_time = time.perf_counter()
-            if timeout_seconds < (current_time - wait_start_time) or output_validator.validate():
+            self.done_event.wait(timeout_seconds - time.perf_counter() + wait_start_time)
```
fgerlits, line 42: (old code, but) 
It looks like this loop always runs only once [or maybe twice because of the mentioned inaccuracy], as the wait increment and the total allowed wait duration are both set to `timeout_seconds`.  Also, why is the loop written as if it's run `max_files` times?

I would 

- make the wait increment much smaller, eg. 10 or 100 ms
- change the loop to an infinite loop that we'll break out of when the timeout is over
- move the `wait` after the check so the extra check at line 37 can be removed

docker/test/integration/minifi/validators/NoFileOutPutValidator.py
```diff
@@ -12,15 +11,13 @@ class NoFileOutPutValidator(FileOutputValidator):
...
+        full_dir = os.path.join(self.output_dir)
         logging.info("Output folder: %s", full_dir)
-        listing = listdir(full_dir)
 
-        self.valid = not bool(listing)
+        if not os.path.isdir(full_dir):
+            return self.valid
 
+        self.valid = (0 == self.get_num_files(full_dir))
         return self.valid
```
fgerlits, line 23: this could be written in a shorter way:
```suggestion
        return os.path.isdir(full_dir) and 0 == self.get_num_files(full_dir)
```
and then `self.valid` can be removed

docker/test/integration/minifi/validators/NumFileRangeValidator.py
```diff
@@ -0,0 +1,24 @@
...
+        self.valid = False
+        full_dir = os.path.join(self.output_dir)
+        logging.info("Output folder: %s", full_dir)
+
+        if not os.path.isdir(full_dir):
+            return self.valid
+
+        num_files = self.get_num_files(full_dir)
+        self.valid = self.min_files < num_files and num_files < self.max_files
+        return self.valid
```
fgerlits, line 24: I would also get rid of `self.valid` here:
```suggestion
        if not os.path.isdir(full_dir):
            return False

        num_files = self.get_num_files(full_dir)
        return self.min_files < num_files and num_files < self.max_files
```

(and similarly in `SingleFileOutputValidator` and `SingleOrMoreFileOutputValidator`)

extensions/librdkafka/ConsumeKafka.cpp
```diff
@@ -78,7 +78,8 @@ core::Property ConsumeKafka::TopicNames(core::PropertyBuilder::createProperty("T
   ->build());
 
 core::Property ConsumeKafka::TopicNameFormat(core::PropertyBuilder::createProperty("Topic Name Format")
-  ->withDescription("Specifies whether the Topic(s) provided are a comma separated list of names or a single regular expression.")
+  ->withDescription("Specifies whether the Topic(s) provided are a comma separated list of names or a single regular expression. "
+      "Using regular expressions does not automatically discover kafka topics created after the processor started.")
```
fgerlits, line 82: `Kafka` is written with a capital `K` elsewhere, so should probably be here, too

extensions/librdkafka/ConsumeKafka.cpp
```diff
@@ -359,25 +362,6 @@ void ConsumeKafka::configure_new_connection(const core::ProcessContext& context)
...
-
-    if (!message_wrapper || RD_KAFKA_RESP_ERR_NO_ERROR != message_wrapper->err) {
-      break;
-    }
-    utils::print_kafka_message(*message_wrapper, *logger_);
-    // Commit offsets on broker for the provided list of partitions
-    logger_->log_info("Committing offset: %" PRId64 ".", message_wrapper->offset);
-    rd_kafka_commit_message(consumer_.get(), message_wrapper.get(), /* async = */ 0);
-  }
-  logger_->log_info("Done resetting offset manually.");
```
fgerlits, line 380: Do you know why this manual offset reset was added in the first place?

docker/test/integration/features/http.feature
lordgamez: Added Jira ticket https://issues.apache.org/jira/browse/MINIFICPP-1588

docker/test/integration/minifi/core/DockerTestDirectoryBindings.py
lordgamez: In python that syntax cannot be used.
```
    a = 0777
           ^
SyntaxError: leading zeros in decimal integer literals are not permitted; use an 0o prefix for octal integers
```

extensions/librdkafka/ConsumeKafka.cpp
lordgamez: I could not find out the reason for that. It seems to always set the offset to the latest message in the kafka topic, which contradicts the `OffsetReset` property. When I tested it manually with the `earliest` offset set it still retrieved the messages published to the Kafka broker after the Minifi agent was restarted. I tried to ask Adam Hunyadi about this functionality, but I did not get a reply about this in e-mail. To me it seems to be unneeded and faulty after testing the functionality.

docker/test/integration/minifi/core/DockerTestDirectoryBindings.py
fgerlits: Yeah, it looks like the 0777 syntax was valid in Python 2 but it got removed in Python 3.  Maybe someone pointed out to them that having both `0123` and `0o123` conflicts with "There's Only One Way To Do It" -- and then they picked the wrong one to keep. :D

docker/test/integration/minifi/core/FileSystemObserver.py
lordgamez: I'm not sure about this. Could you elaborate what do you mean by increments here? It seems to me that we are only waiting until a certain time point, the end of the timeout calculated from the start time of the function call. We are then waiting for the `done_event` to be set by a newly created file, for the number of `max_files` or until the designated time point is reached.

The first check is also needed IMO as it could happen that the observer is restarted after the required files are already created so we need to check them beforehand, because no new files will be created, so the event will not be notified by the observer.

extensions/librdkafka/ConsumeKafka.cpp
lordgamez: Update in cdb014f306f614bfad5e6120f089ab821c904677

docker/test/integration/minifi/validators/NumFileRangeValidator.py
lordgamez: Update in cdb014f306f614bfad5e6120f089ab821c904677

docker/test/integration/minifi/validators/NoFileOutPutValidator.py
lordgamez: Update in cdb014f306f614bfad5e6120f089ab821c904677

docker/test/integration/minifi/processors/GetFile.py
```diff
@@ -7,7 +7,7 @@ def __init__(self, input_dir="/tmp/input", schedule={'scheduling period': '2 sec
             'GetFile',
             properties={
                 'Input Directory': input_dir,
-                'Keep Source File': 'true'
+                'Keep Source File': 'false'
```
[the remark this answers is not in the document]
lordgamez: Update in cdb014f306f614bfad5e6120f089ab821c904677

docker/test/integration/features/https.feature
lordgamez: Update in cdb014f306f614bfad5e6120f089ab821c904677

docker/test/integration/steps/steps.py
lordgamez: Update in cdb014f306f614bfad5e6120f089ab821c904677

docker/test/integration/environment.py
lordgamez: Update in cdb014f306f614bfad5e6120f089ab821c904677

docker/test/integration/features/kafka.feature
```diff
@@ -57,3 +57,180 @@ Feature: Sending data to using Kafka streaming platform using PublishKafka
...
+
+    When all instances start up
+    And 1000 kafka messages are sent to the topic "ConsumeKafkaTest"
+
+    Then minimum <min expected messages>, maximum <max expected messages> flowfiles are produced and placed in the monitored directory in less than <polling time>
+
+  Examples: Message batching
+    | max poll records | scheduling period | polling time | min expected messages | max expected messages |
+    | 3                | 10 sec            | 60 seconds   | 12                    | 24                    |
+    | 6                | 10 sec            | 60 seconds   | 24                    | 48                    |
```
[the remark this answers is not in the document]
lordgamez: Update in cdb014f306f614bfad5e6120f089ab821c904677 to expect half of those messages in half of the polling interval

docker/test/integration/steps/steps.py
```diff
@@ -270,9 +367,119 @@ def step_impl(context, content, file_name, path, seconds):
...
+    context.test.check_for_multiple_files_generated(2, timeparse(duration), [content_1, content_2])
+
+
+@then("flowfiles with these contents are placed in the monitored directory in less than {duration}: \"{contents}\"")
+def step_impl(context, duration, contents):
+    contents_arr = contents.split(",")
+    context.test.check_for_multiple_files_generated(len(contents_arr), timeparse(duration), contents_arr)
+
+
+@then("minimum {lower_bound}, maximum {upper_bound} flowfiles are produced and placed in the monitored directory in less than {duration}")
```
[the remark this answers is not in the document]
lordgamez: Update in cdb014f306f614bfad5e6120f089ab821c904677

docker/DockerVerify.sh
```diff
@@ -72,28 +74,12 @@ export PATH
 PYTHONPATH="${PYTHONPATH}:${docker_dir}/test/integration"
 export PYTHONPATH
 
+# Add --no-logcapture to see logs inline
```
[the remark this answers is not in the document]
lordgamez: Update in cdb014f306f614bfad5e6120f089ab821c904677

docker/DockerVerify.sh
```diff
@@ -72,28 +74,13 @@ export PATH
...
-  behave "${BEHAVE_OPTS[@]}" "features/s3.feature" -n "A MiNiFi instance transfers encoded data through a http proxy to s3" &&
-  behave "${BEHAVE_OPTS[@]}" "features/s3.feature" -n "A MiNiFi instance can remove s3 bucket objects" &&
-  behave "${BEHAVE_OPTS[@]}" "features/s3.feature" -n "Deletion of a s3 object through a proxy-server succeeds" &&
-  behave "${BEHAVE_OPTS[@]}" "features/s3.feature" -n "A MiNiFi instance can download s3 bucket objects directly" &&
-  behave "${BEHAVE_OPTS[@]}" "features/s3.feature" -n "A MiNiFi instance can download s3 bucket objects via a http-proxy" &&
-  behave "${BEHAVE_OPTS[@]}" "features/s3.feature" -n "A MiNiFi instance can list an S3 bucket directly" &&
-  behave "${BEHAVE_OPTS[@]}" "features/s3.feature" -n "A MiNiFi instance can list an S3 bucket objects via a http-proxy" &&
-  behave "${BEHAVE_OPTS[@]}" "features/azure_storage.feature" -n "A MiNiFi instance can upload data to Azure blob storage"
+  # behave "${BEHAVE_OPTS[@]}"
+  behave "${BEHAVE_OPTS[@]}" "features/kafka.feature" -n "The ConsumeKafka \"Maximum Poll Records\" property sets a limit on the messages processed in a single batch"
```
fgerlits, line 86: did you check in this change by accident?

docker/test/integration/minifi/core/FileSystemObserver.py
fgerlits: Sorry, I misread the code, I thought the `wait()` was a sleep, and it was always going to wait for `timeout_seconds`.  Now I see it is waiting to be notified of a file system event.  This is fine.

fgerlits: Sorry, I misread the code, I thought the `wait()` was a sleep, and it was always going to wait for `timeout_seconds`. Now I see it is waiting to be notified of a file system event. This is fine.

docker/DockerVerify.sh
lordgamez: Yes, thank you for noticing! Left it there yesterday while testing and forgot about it :) Fixed it in 7a0dcab164ddd96d8edeaacbd2d61b945de4fdf9

fgerlits (approved):

docker/test/integration/steps/steps.py
```diff
@@ -180,6 +248,13 @@ def step_impl(context, file_name, content, path):
     context.test.add_test_data(path, content, file_name)
 
 
+@given("two files with content \"{content_1}\" and \"{content_2}\" are placed in \"{path}\"")
```
aminadinari19, line 251: I am not sure about the purpose of this step. It doesn't seem to be used anywhere and if it could be slightly less specific where it checks an arbitrary number of files?

lordgamez: You are right, when I removed the code for the failing scenario, it became unused. Removed it in 4248edb19945b6df8e3bd2b69add3d4ad252d7bc and added to https://issues.apache.org/jira/browse/MINIFICPP-1588 as a possible implementation, which can be changed later.

docker/test/integration/minifi/validators/NoContentCheckFileNumberValidator.py
```diff
@@ -0,0 +1,25 @@
...
+import os
+
+from .FileOutputValidator import FileOutputValidator
+
+
+class NoContentCheckFileNumberValidator(FileOutputValidator):
+    """
+    Validates the content of a single file in the given directory.
+    """
+
```
szaszm, line 11: This comment disagrees with the code

docker/test/integration/minifi/validators/SingleOrMoreFileOutputValidator.py
```diff
@@ -0,0 +1,22 @@
+import logging
+import os
+
+from .FileOutputValidator import FileOutputValidator
+
+
+class SingleOrMoreFileOutputValidator(FileOutputValidator):
+    """
+    Validates the content of a single file in the given directory.
+    """
```
szaszm, line 10: SingleOrMulti or OneOrMore (my preference) would be better naming. Also, the comment doesn't mention the possibility of multiple matches.

docker/test/integration/minifi/validators/NoContentCheckFileNumberValidator.py
lordgamez: Fixed in f1c9709ef6a9a2ea4ff8cf3e9e257123e8cd12dc

szaszm (approved):

PR #1078, opened 2021-05-21
MINIFICPP-1562 Use CLCACHE_NODIRECT option to avoid clcache failures

lordgamez:
The clcache object invalidation caused failures in the Github Actions Windows-VS2019 job previously which reappeared in run https://github.com/apache/nifi-minifi-cpp/runs/2628922164?check_suite_focus=true with the restore key: windows-vs2019-clcache-refs/heads/main-63f6e716f74f0936fc38b755eea23f83b481044c

Amending the PR by adding the CLCACHE_NODIRECT option to the job made the next run succeed with the same restore key: https://github.com/apache/nifi-minifi-cpp/runs/2630706634?check_suite_focus=true

Description of the option:

> **CLCACHE_NODIRECT**
> Disable direct mode. If this variable is set, clcache will always run preprocessor on source file and will hash preprocessor output to get cache key. Use this if you experience problems with direct mode or if you need built-in macroses like _TIME_ to work correctly.

--------------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,6 +106,7 @@ jobs:
           move "$cl_exe_dir\cl.exe.config" "$cl_exe_dir\cl_original.exe.config"
           move "$cl_exe_dir\clcache.exe" "$cl_exe_dir\cl.exe"
           echo "CLCACHE_CL=$cl_exe_dir\cl_original.exe" >> $env:GITHUB_ENV
+          echo "CLCACHE_NODIRECT=1" >> $env:GITHUB_ENV
       - name: build
         run: |
           PATH %PATH%;C:\Program Files (x86)\Windows Kits\10\bin\10.0.17763.0\x86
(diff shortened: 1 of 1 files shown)


szaszm (approved):
> adding the CLCACHE_NODIRECT option to the job made the next run succeed with the same restore key

This could happen because the hashes of the original and the preprocessed sources are different, so this was effectively a cache clear. FYI, but I don't have any other idea.

lordgamez:
> > adding the CLCACHE_NODIRECT option to the job made the next run succeed with the same restore key
> 
> This could happen because the hashes of the original and the preprocessed sources are different, so this was effectively a cache clear. FYI, but I don't have any other idea.

That is probably the case here judging from the build runtime. We should see if the issue reappears in the future with the `no direct` option. Maybe with the different cache key handling the issue will not persist anymore. There was a similar [missing object issue](https://github.com/frerich/clcache/issues/210) in the past with the `no direct` option which was fixed and could not be reproduced later, so I'm hoping this could help us as well.

fgerlits (approved):

martinzink (approved):

PR #1081, opened 2021-05-21
MINIFICPP-1565: Minor improvements to PerformanceDataMonitor

martinzink:

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/include/utils/MathUtils.h
```diff
@@ -0,0 +1,43 @@
...
+
+class MathUtils {
+ public:
+  static double round_to(double original, int8_t precision) {
+    if (precision < 0) {
+      return original;
+    } else {
+      double power_ten = pow(10, precision);
+      return std::round(original * power_ten) / power_ten;
+    }
```
lordgamez, line 35: You can lose the else branch here, with the early return before.

extensions/pdh/PerformanceDataMonitor.cpp
```diff
@@ -45,9 +45,15 @@ core::Property PerformanceDataMonitor::CustomPDHCounters(
...
-    withDefaultValue(JSON_FORMAT_STR)->build());
+    withAllowableValue(PRETTY_JSON_FORMAT_STR)->
+    withAllowableValue(COMPACT_JSON_FORMAT_STR)->
+    withAllowableValue(PRETTY_OPEN_TELEMETRY_FORMAT_STR)->
+    withAllowableValue(COMPACT_OPEN_TELEMETRY_FORMAT_STR)->
+    withDefaultValue(PRETTY_JSON_FORMAT_STR)->build());
+
+core::Property PerformanceDataMonitor::DoublePrecisionProperty(
+  core::PropertyBuilder::createProperty("Double Precision")->
+  withDescription("Rounds the double values to this precision (blank for maximum precision)")->build());
```
szaszm, line 56: What does precision mean in this context? Some power of 10 decimal place? Or power of 2? Or number of significant decimal digits? Or ...
This could use some more precise description.

libminifi/include/utils/MathUtils.h
```diff
@@ -0,0 +1,43 @@
...
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace utils {
+
+class MathUtils {
+ public:
+  static double round_to(double original, int8_t precision) {
```
lordgamez, line 29: I may go with utils::optional<uint8_t> instead, but I don't insist on it.

libminifi/include/utils/MathUtils.h
```diff
@@ -0,0 +1,43 @@
...
+    if (precision < 0) {
+      return original;
```
szaszm, line 31: Rounding to fractional decimal places like 10^-2 => 123.45 makes sense as well. Either implement the full expected behavior or move the implementation to somewhere local near the usage, so it's clear that this logic is related to the performance monitor functionality.

extensions/pdh/PerformanceDataMonitor.cpp
```diff
@@ -264,16 +281,29 @@ void PerformanceDataMonitor::setupMembersFromProperties(const std::shared_ptr<co
...
+    } else if (output_format_string == PRETTY_JSON_FORMAT_STR || output_format_string == COMPACT_JSON_FORMAT_STR) {
       output_format_ = OutputFormat::JSON;
+      pretty_output_ = output_format_string == PRETTY_JSON_FORMAT_STR;
+      logger_->log_trace("OutputFormat is configured to be %s JSON", pretty_output_ ? "pretty" : "compact");
     } else {
-      logger_->log_error("Invalid OutputFormat, defaulting to JSON");
       output_format_ = OutputFormat::JSON;
+      pretty_output_ = true;
+      logger_->log_error("Invalid OutputFormat, defaulting to %s JSON", pretty_output_ ? "pretty" : "compact");
+    }
```
lordgamez, line 296: I may extract this to a member function

extensions/pdh/PerformanceDataMonitor.cpp
```diff
@@ -264,16 +281,29 @@ void PerformanceDataMonitor::setupMembersFromProperties(const std::shared_ptr<co
...
+  }
+
+  std::string double_precision_string;
+  if (context->getProperty(DoublePrecisionProperty.getName(), double_precision_string)) {
```
lordgamez, line 300: The getProperty should do the conversion implicitly, so reading this to `double_precision_` variable should be enough.

extensions/pdh/PerformanceDataMonitor.h
```diff
@@ -75,7 +78,9 @@ class PerformanceDataMonitor : public core::Processor {
   void addCustomPDHCountersFromProperty(const std::string& custom_pdh_counters);
 
   OutputFormat output_format_;
+  bool pretty_output_;
```
lordgamez, line 81: This seems to be uninitialized.

extensions/pdh/PerformanceDataMonitor.cpp
```diff
@@ -45,9 +45,15 @@ core::Property PerformanceDataMonitor::CustomPDHCounters(
 core::Property PerformanceDataMonitor::OutputFormatProperty(
     core::PropertyBuilder::createProperty("Output Format")->
     withDescription("Format of the created flowfiles")->
-    withAllowableValue<std::string>(JSON_FORMAT_STR)->
-    withAllowableValue(OPEN_TELEMETRY_FORMAT_STR)->
-    withDefaultValue(JSON_FORMAT_STR)->build());
+    withAllowableValue(PRETTY_JSON_FORMAT_STR)->
+    withAllowableValue(COMPACT_JSON_FORMAT_STR)->
+    withAllowableValue(PRETTY_OPEN_TELEMETRY_FORMAT_STR)->
+    withAllowableValue(COMPACT_OPEN_TELEMETRY_FORMAT_STR)->
```
adamdebreceni, line 51: it seems `withAllowableValues` is more prevalent than `withAllowableValue`

extensions/pdh/PerformanceDataMonitor.cpp
```diff
@@ -102,25 +108,36 @@ void PerformanceDataMonitor::onTrigger(core::ProcessContext* context, core::Proc
...
+    session->transfer(flowFile, Success);
   }
-  utils::JsonOutputCallback callback(std::move(root));
-  session->write(flowFile, &callback);
-  session->transfer(flowFile, Success);
 }
 
 void PerformanceDataMonitor::initialize() {
-  setSupportedProperties({ CustomPDHCounters, PredefinedGroups, OutputFormatProperty });
+  setSupportedProperties({ CustomPDHCounters, PredefinedGroups, OutputFormatProperty, DoublePrecisionProperty });
```
adamdebreceni, line 125: I don't think we usually suffix properties with `Property`

extensions/pdh/PerformanceDataMonitor.cpp
```diff
@@ -264,16 +281,29 @@ void PerformanceDataMonitor::setupMembersFromProperties(const std::shared_ptr<co
...
+      logger_->log_trace("OutputFormat is configured to be %s OpenTelemetry", pretty_output_ ? "pretty" : "compact");
+    } else if (output_format_string == PRETTY_JSON_FORMAT_STR || output_format_string == COMPACT_JSON_FORMAT_STR) {
       output_format_ = OutputFormat::JSON;
+      pretty_output_ = output_format_string == PRETTY_JSON_FORMAT_STR;
+      logger_->log_trace("OutputFormat is configured to be %s JSON", pretty_output_ ? "pretty" : "compact");
     } else {
-      logger_->log_error("Invalid OutputFormat, defaulting to JSON");
       output_format_ = OutputFormat::JSON;
+      pretty_output_ = true;
+      logger_->log_error("Invalid OutputFormat, defaulting to %s JSON", pretty_output_ ? "pretty" : "compact");
```
adamdebreceni, line 295: we should throw here instead of defaulting, the user gave an invalid output format, that is an error IMO

libminifi/include/utils/MathUtils.h
```diff
@@ -0,0 +1,43 @@
...
+    } else {
+      double power_ten = pow(10, precision);
+      return std::round(original * power_ten) / power_ten;
```
adamdebreceni, line 34: as far as I know, precision has a different meaning, it is the number of "significant digits", so pinting `0.00123` with a precision of `2` should result in `0.0012` instead of `0` as this function does

adamdebreceni: a thought: since the `rapidjson::Document::Accept` is templated, we could wrap a writer (`rapidjons::Writer<...>`, `rapidjson::PrettyWriter<...>`) in a custom class, inheriting from it and shadowing the `Double` method to implement our custom double formatting, although this might not be trivial

martinzink: You are right, changed the function to round_to_decimal_places so its more clear what the function does.

extensions/pdh/PerformanceDataMonitor.cpp
martinzink: good idea changed it in [26802f7](https://github.com/martinzink/nifi-minifi-cpp/commit/26802f77ef539757fcedbffea0f596b2ef0cc828)

extensions/pdh/PerformanceDataMonitor.cpp
martinzink: changed to DecimalPlaces in [26802f7](https://github.com/martinzink/nifi-minifi-cpp/commit/26802f77ef539757fcedbffea0f596b2ef0cc828)

extensions/pdh/PerformanceDataMonitor.cpp
martinzink: good idea, changed it in [26802f7](https://github.com/martinzink/nifi-minifi-cpp/commit/26802f77ef539757fcedbffea0f596b2ef0cc828)

extensions/pdh/PerformanceDataMonitor.h
martinzink: good catch, fixed in [26802f7](https://github.com/martinzink/nifi-minifi-cpp/commit/26802f77ef539757fcedbffea0f596b2ef0cc828)

extensions/pdh/PerformanceDataMonitor.cpp
martinzink: Good idea, unfortunetly int8_t is not yet supported by Value.h and ValueParser.h (which the getProperty uses) but left TODO and will replace this in a separate PR if thats okay with you
(I didnt want to increase the scope of this PR with the int8_t Value implementation)

extensions/pdh/PerformanceDataMonitor.cpp
martinzink: extracted the different PropertyHandling functions to separete member functions in [26802f7](https://github.com/martinzink/nifi-minifi-cpp/commit/26802f77ef539757fcedbffea0f596b2ef0cc828)

libminifi/include/utils/MathUtils.h
martinzink: removed the IF now it works with negative decimal places as well (also renamed to round_to_decimal_places) in [26802f7](https://github.com/martinzink/nifi-minifi-cpp/commit/26802f77ef539757fcedbffea0f596b2ef0cc828)

libminifi/include/utils/MathUtils.h
martinzink: good idea thanks, removed the IF now it works as you suggested (also renamed to round_to_decimal_places) in [26802f7](https://github.com/martinzink/nifi-minifi-cpp/commit/26802f77ef539757fcedbffea0f596b2ef0cc828)

extensions/pdh/PerformanceDataMonitor.cpp
martinzink: Good point, changed it to "The number of decimal places to round the values to (blank for no rounding)" in [26802f7](https://github.com/martinzink/nifi-minifi-cpp/commit/26802f77ef539757fcedbffea0f596b2ef0cc828)

extensions/pdh/PerformanceDataMonitor.cpp
lordgamez: Sure it's okay in a separate PR, thanks!

martinzink:
> I would pass `utils::optional<int8_t>` by value. `sizeof(utils::optional<int8_t>)` should be 2 and copying is not expensive.
> 
> As a rule of thumb, if `sizeof(T) <= 32` and copying doesn't incur allocation or other expensive operations, I would pass by value.
> 
> https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#cp31-pass-small-amounts-of-data-between-threads-by-value-rather-than-by-reference-or-pointer

Changed it [aee9ab ](https://github.com/martinzink/nifi-minifi-cpp/commit/aee9ab6b573cb5c5d699dfc5f7aee6d77015f64d)

extensions/pdh/tests/PerformanceDataCounterTests.cpp
```diff
@@ -122,8 +122,8 @@ TEST_CASE("PDHCounterDataCollectionTest", "[pdhcounterdatacollectiontest]") {
   REQUIRE(int_counter.collectData());
 
   rapidjson::Document document(rapidjson::kObjectType);
-  double_counter.addToJson(document, document.GetAllocator());
-  int_counter.addToJson(document, document.GetAllocator());
+  double_counter.addToJson(document, document.GetAllocator(), utils::optional<int8_t>());
+  int_counter.addToJson(document, document.GetAllocator(), utils::optional<int8_t>());
```
szaszm, line 126: You can use `utils::nullopt` for empty optionals.

szaszm (approved):

extensions/pdh/tests/PerformanceDataCounterTests.cpp
martinzink: yeah that makes the code more readable, changed it

extensions/pdh/PerformanceDataMonitor.cpp
adamdebreceni: the same thing might apply to `OutputFormatProperty`

extensions/pdh/PerformanceDataMonitor.cpp
```diff
@@ -249,32 +265,56 @@ void PerformanceDataMonitor::addCustomPDHCountersFromProperty(const std::string&
...
     }
   }
+  logger_->log_trace("OutputFormat is configured to be %s %s", pretty_output_ ? "pretty" : "compact", output_format_ == OutputFormat::JSON ? "JSON" : "OpenTelemtry");
+}
+
+void PerformanceDataMonitor::setupDecimalPlacesFromProperties(const std::shared_ptr<core::ProcessContext>& context) {
+  std::string decimal_places_str;
+  if (context->getProperty(DecimalPlaces.getName(), decimal_places_str)) {  // TODO(mzink): Implement int8_t in Value.h to convert int8_t implicitly
+    try {
+      decimal_places_ = std::stoi(decimal_places_str);
```
adamdebreceni, line 304: watch out as `std::stoi` parses "dubious" strings as well (like `"-1 banana"`), using the `ValueParser` utility could elide this (unintented?) behavior

extensions/pdh/PerformanceDataMonitor.cpp
adamdebreceni: although `std::invalid_argument` makes sense, would throwing an `Exception(PROCESS_SCHEDULE_EXCEPTION)` make it more consistent with other processors?

extensions/pdh/PerformanceDataMonitor.cpp
martinzink: ValueParser unfortunetly doesnt (yet) support int8_t only uint64_t, int64_t, uint32_t.
We could ofcourse use int64_t, then cast that into int8_t. If thats better?

extensions/pdh/PerformanceDataMonitor.cpp
martinzink: that conflicts with the PerformanceDataMonitor::OutputFormat enum class

extensions/pdh/PerformanceDataMonitor.cpp
adamdebreceni: yes, we could go for that, and throw on out-of-range as well

extensions/pdh/PerformanceDataMonitor.cpp
adamdebreceni: it makes sense then

extensions/pdh/PerformanceDataMonitor.cpp
martinzink: changed it in [87a9d18](https://github.com/martinzink/nifi-minifi-cpp/commit/87a9d18114347303156fd987ac482949fba47630)

extensions/pdh/PerformanceDataMonitor.cpp
martinzink: changed it in [0844651](https://github.com/martinzink/nifi-minifi-cpp/commit/084465104f3b63c29485638adb1c62d2b3b7e0eb)

extensions/pdh/PerformanceDataCounter.h
```diff
@@ -18,6 +18,7 @@
 
 #include <string>
 #include "rapidjson/document.h"
+#include "utils/OptionalUtils.h"
```
fgerlits, line 21: this `#include` is not used

extensions/pdh/PerformanceDataMonitor.cpp
```diff
@@ -249,32 +265,54 @@ void PerformanceDataMonitor::addCustomPDHCountersFromProperty(const std::string&
...
+    } else if (output_format_string == PRETTY_JSON_FORMAT_STR || output_format_string == COMPACT_JSON_FORMAT_STR) {
       output_format_ = OutputFormat::JSON;
+      pretty_output_ = output_format_string == PRETTY_JSON_FORMAT_STR;
     } else {
-      logger_->log_error("Invalid OutputFormat, defaulting to JSON");
-      output_format_ = OutputFormat::JSON;
+      throw Exception(PROCESS_SCHEDULE_EXCEPTION, "Invalid PerformanceDataMonitor Output Format: " + output_format_string);
     }
   }
+  logger_->log_trace("OutputFormat is configured to be %s %s", pretty_output_ ? "pretty" : "compact", output_format_ == OutputFormat::JSON ? "JSON" : "OpenTelemtry");
```
fgerlits, line 297: typo: OpenTelemtry -> OpenTelemetry

extensions/pdh/PerformanceDataMonitor.cpp
```diff
@@ -45,9 +45,13 @@ core::Property PerformanceDataMonitor::CustomPDHCounters(
 core::Property PerformanceDataMonitor::OutputFormatProperty(
     core::PropertyBuilder::createProperty("Output Format")->
     withDescription("Format of the created flowfiles")->
-    withAllowableValue<std::string>(JSON_FORMAT_STR)->
-    withAllowableValue(OPEN_TELEMETRY_FORMAT_STR)->
-    withDefaultValue(JSON_FORMAT_STR)->build());
+    withAllowableValues<std::string>({ PRETTY_JSON_FORMAT_STR, COMPACT_JSON_FORMAT_STR, PRETTY_OPEN_TELEMETRY_FORMAT_STR, COMPACT_OPEN_TELEMETRY_FORMAT_STR })->
```
fgerlits, line 48: I would prefer two separate properties: `Output Format = JSON | OPEN_TELEMETRY` and `Output Style = PRETTY | COMPACT`, as that would make the code simpler, and I think the interface would be clearer, too.

extensions/pdh/tests/PerformanceDataMonitorTests.cpp
```diff
@@ -136,8 +136,9 @@ TEST_CASE("PerformanceDataMonitorCustomPDHCountersTest", "[performancedatamonito
```
fgerlits, line 136: I have seen the assertion at line 127 fail seemingly randomly in some CI jobs (I think they were all Windows 2017 jobs).  Any idea why that happens?

extensions/pdh/PerformanceDataMonitor.cpp
```diff
@@ -45,9 +45,13 @@ core::Property PerformanceDataMonitor::CustomPDHCounters(
...
+    withDefaultValue(PRETTY_JSON_FORMAT_STR)->build());
+
+core::Property PerformanceDataMonitor::DecimalPlaces(
+  core::PropertyBuilder::createProperty("Round to decimal places")->
+  withDescription("The number of decimal places to round the values to (blank for no rounding)")->
+  withDefaultValue("3")->build());
```
szaszm, line 54: If there is a default value, doesn't it mean it's impossible to have a blank value? Because it would just get replaced with the default.

martinzink: I was able to set empty string during testing with c2 protocoll, but changed it to blank nontheless thanks

extensions/pdh/tests/PerformanceDataMonitorTests.cpp
martinzink: yeah, I already talked about that with @lordgamez , my guess is that the first flowfile can be invalid due to no wait time between the onSchedule and the onTrigger
Will definetly fix that, however I wouldnt delay this PR with that

extensions/pdh/PerformanceDataMonitor.cpp
martinzink: changed it

extensions/pdh/PerformanceDataMonitor.cpp
martinzink: fixed it

extensions/pdh/PerformanceDataCounter.h
martinzink: good catch, thanks fixed it

extensions/pdh/tests/PerformanceDataMonitorTests.cpp
fgerlits: Agreed, it should be in a separate PR if it is not an obvious/tiny change.

extensions/pdh/PerformanceDataMonitor.cpp
```diff
@@ -249,32 +273,70 @@ void PerformanceDataMonitor::addCustomPDHCountersFromProperty(const std::string&
...
+      throw Exception(PROCESS_SCHEDULE_EXCEPTION, "Invalid PerformanceDataMonitor Output Compactness: " + output_compactness_string);
+    }
+  }
+
+  logger_->log_trace("OutputFormat is configured to be %s %s", pretty_output_ ? "pretty" : "compact", output_format_ == OutputFormat::JSON ? "JSON" : "OpenTelemetry");
+}
+
+void PerformanceDataMonitor::setupDecimalPlacesFromProperties(const std::shared_ptr<core::ProcessContext>& context) {
+  std::string decimal_places_str;
+  if (context->getProperty(DecimalPlaces.getName(), decimal_places_str) && decimal_places_str == "") {
```
fgerlits, line 320: I think this should be
```suggestion
  if (!context->getProperty(DecimalPlaces.getName(), decimal_places_str) || decimal_places_str == "") {
```
although I don't know if `getProperty()` can return false when there is a default.

adamdebreceni: I don't think it can, what are the benefits of using the empty string instead of leaving it without a default?

adamdebreceni (approved):

fgerlits (approved):

szaszm (approved):

PR #1085, opened 2021-05-27
MINIFICPP-1568 Retry PDH processor run to fix test flakiness

lordgamez:
According to @martinzink due to a timing issue the first trigger of the `PerformanceDataMonitor` processor may not contain all the requested counter data, only the following triggers do. Adding the possibility to retry the processor trigger in case the first one fails should remove the flakiness cause by this.

------------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/pdh/tests/PerformanceDataMonitorTests.cpp
+++ b/extensions/pdh/tests/PerformanceDataMonitorTests.cpp
@@ -20,6 +20,7 @@
 #include <vector>
 #include <set>
 #include <fstream>
+#include <functional>
 
 #include "TestBase.h"
 #include "processors/PutFile.h"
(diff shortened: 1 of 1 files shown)


martinzink (approved):
LGTM, this should fix the flakyness
10 tries is maybe overkill, but couldnt hurt either

szaszm:
What is the timing issue? Is it unfeasible to eliminate it instead of a workaround in the test?

lordgamez:
> What is the timing issue? Is it unfeasible to eliminate it instead of a workaround in the test?

The timing issue is not in our code, but comes from the Windows OS. After creating the counter we have to let some time pass in order for the OS to gather counter data to be returned when we run the query in the `onTrigger` call. In the first 1-2 calls no data is available yet so the assertions may fail. I hope I'm not mistaken @martinzink may comment further on this.

martinzink:
> What is the timing issue? Is it unfeasible to eliminate it instead of a workaround in the test?

The problem is that the processor opens the windows pdh query in the onschedule function, and collects the data in the ontrigger function.
If there is no usable data in the query then it wont be included in the results.
From microsofts documentation
https://docs.microsoft.com/en-us/windows/win32/perfctrs/collecting-performance-data
Its recommended to wait one second between data collection.

So the onTriggers calls which happen right after the onschedule could lack some fields.

Maybe a cleaner workaround would be if we could schedule the processor separatetly wait a bit than start the triggering, but not sure how big that change would be.

szaszm:
Ok, I see. It sounds like it would be complicated to ensure that onTrigger results in actual data.

szaszm (approved):

arpadboda (approved):

PR #1090, opened 2021-05-31
MINIFICPP-1561 - Allow rocksdb encryption

adamdebreceni:
Depends on #1071 
---

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/rocksdb-repos/encryption/RocksDbEncryptionProvider.cpp
```diff
@@ -0,0 +1,123 @@
...
+namespace minifi {
+namespace core {
+namespace repository {
+
+using utils::crypto::Bytes;
+using utils::crypto::Aes256EcbCipher;
+
+namespace {
+
+class AES256BlockCipher final : public rocksdb::BlockCipher {
```
szaszm, line 35: [`systemd::libwrapper::DlopenJournal`](https://github.com/apache/nifi-minifi-cpp/blob/main/extensions/systemd/libwrapper/DlopenWrapper.cpp#L35) also started out with static linkage. The reason I changed it was because debuggers have a hard time showing symbols that are not there and this can make debugging much harder. I think I had to step through asm instructions at one point to debug something.

If you think having type info for these types while debugging is worth the overhead, consider moving the classes out of the anonymous namespace.

libminifi/test/TestBase.h
```diff
@@ -445,6 +445,13 @@ class TestController {
     return dir;
   }
 
+  template<size_t N>
+  utils::Path createTempDirectory(const char (&format)[N]) {
```
lordgamez, line 449: This simplifies things a bit :+1:

lordgamez (approved):

extensions/rocksdb-repos/encryption/RocksDbEncryptionProvider.cpp
```diff
@@ -0,0 +1,123 @@
...
+    return Aes256EcbCipher::BLOCK_SIZE;
+  }
+
+  bool equals(const AES256BlockCipher& other) const {
+    return cipher_impl_.equals(other.cipher_impl_);
+  }
+
+  rocksdb::Status Encrypt(char *data) override;
+
+  rocksdb::Status Decrypt(char *data) override;
```
szaszm, line 56: I prefer using separate buffers for input and output, unless we can realize significant efficiency gains by reusing mutable memory.
How do we ensure that the buffer size is sufficient for the output? Plaintext and cipher size can be different in either direction.
Tip: use `gsl::span` for the input

libminifi/include/utils/crypto/ciphers/Aes256Ecb.h
```diff
@@ -0,0 +1,69 @@
...
+
+  bool equals(const Aes256EcbCipher& other) const;
+
+ private:
+  template<typename ...Args>
+  static void handleError(Args&& ...args) {
+    std::string error_msg = core::logging::format_string(-1, "%s", std::forward<Args>(args)...);
+    logger_->log_error("%s", error_msg);
+    throw CipherError(error_msg);
+  }
```
szaszm, line 60: This seems to just print the passed first argument string as is, but it's passed a format string in the constructor. If this function is supposed to expect format strings, line 57 shouldn't have the `"%s"` arguments. Consider making this function take a c string and handle the formatting in the constructor.

extensions/rocksdb-repos/FlowFileRepository.cpp
```diff
@@ -220,17 +240,21 @@ void FlowFileRepository::initialize_repository() {
...
   }
-  rocksdb::Checkpoint *checkpoint;
   // delete any previous copy
-  if (utils::file::FileUtils::delete_dir(checkpoint_dir_) >= 0 && opendb->NewCheckpoint(&checkpoint).ok()) {
-    if (checkpoint->CreateCheckpoint(checkpoint_dir_).ok()) {
+  if (utils::file::FileUtils::delete_dir(checkpoint_dir_) >= 0) {
+    rocksdb::Checkpoint* checkpoint = nullptr;
+    rocksdb::Status checkpoint_status = opendb->NewCheckpoint(&checkpoint);
+    if (checkpoint_status.ok()) {
+      checkpoint_status = checkpoint->CreateCheckpoint(checkpoint_dir_);
```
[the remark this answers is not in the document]
adamdebreceni: done

libminifi/test/TestBase.h
adamdebreceni: if only there wasn't now three different ways to create a temporary directory (now 2 in TestController, and 1 standalone in TestUtils.h) but I felt like removing all the single-use `format` variables in all the tests deserves its own PR

lordgamez: Yes I thought the same, it should definitely be in a separate PR.

libminifi/test/rocksdb-tests/EncryptionTests.cpp
```diff
@@ -0,0 +1,108 @@
...
+    });
+    REQUIRE(success);
+    REQUIRE(utils::verifyLogLinePresenceInPollTime(
+        std::chrono::seconds{5},
+        "Successfully opened checkpoint database at '" + checkpoint_dir_.str() + "'"));
+    std::set<std::shared_ptr<core::FlowFile>> expired;
+    auto flowfile = container_->poll(expired);
+    REQUIRE(expired.empty());
+    REQUIRE(flowfile);
+    REQUIRE(flowfile->getAttribute("my little pony") == "my horse is amazing");
```
szaszm, line 106: Can we have an assertion to the ciphertext value in the repository? It would be a great proof that everything is indeed encrypted on disk and not just working.

libminifi/test/TestBase.h
adamdebreceni: https://issues.apache.org/jira/browse/MINIFICPP-1600

extensions/rocksdb-repos/FlowFileRepository.cpp
```diff
@@ -220,17 +240,21 @@ void FlowFileRepository::initialize_repository() {
...
       checkpoint_ = std::unique_ptr<rocksdb::Checkpoint>(checkpoint);
-      logger_->log_trace("Created checkpoint directory");
+      logger_->log_trace("Created checkpoint in directory '%s'", checkpoint_dir_);
     } else {
-      logger_->log_trace("Could not create checkpoint. Corrupt?");
+      logger_->log_error("Could not create checkpoint: %s", checkpoint_status.ToString());
     }
   } else
-    logger_->log_trace("Could not create checkpoint directory. Not properly deleted?");
+    logger_->log_error("Could not delete existing checkpoint directory '%s'", checkpoint_dir_);
```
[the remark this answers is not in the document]
adamdebreceni: @lordgamez had the same concern, done in commit [3db64f](https://github.com/apache/nifi-minifi-cpp/pull/1090/commits/3db64f47d25f9067fd23ade4e1100a666c542efa)

extensions/rocksdb-repos/DatabaseContentRepository.cpp
```diff
@@ -42,14 +43,31 @@ bool DatabaseContentRepository::initialize(const std::shared_ptr<minifi::Configu
...
+    encryption_opts.database = directory_;
+    encryption_opts.encryption_key_name = ENCRYPTION_KEY_NAME;
+    auto env = createEncryptingEnv(utils::crypto::EncryptionManager{configuration->getHome()}, encryption_opts);
+    if (env) {
+      logger_->log_info("Using encrypted DatabaseContentRepository");
+    } else {
+      logger_->log_info("Using plaintext DatabaseContentRepository");
+    }
+    return env;
+  }();
```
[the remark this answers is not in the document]
adamdebreceni: done in [9d07c8e](https://github.com/apache/nifi-minifi-cpp/pull/1090/commits/9d07c8e9c534a87ab31acd5d7dbf843ce047da04) (also for other applicable occurrences)

extensions/rocksdb-repos/encryption/RocksDbEncryptionProvider.cpp
```diff
@@ -0,0 +1,123 @@
...
+    return "AES256BlockCipher";
+  }
+
+  size_t BlockSize() override {
+    return Aes256EcbCipher::BLOCK_SIZE;
+  }
+
+  bool equals(const AES256BlockCipher& other) const {
+    return cipher_impl_.equals(other.cipher_impl_);
+  }
```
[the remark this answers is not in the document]
adamdebreceni: I forgot to rename it, renamed it to `hasEqualKey` in [7793d6f](https://github.com/apache/nifi-minifi-cpp/pull/1090/commits/7793d6faf7460f37f69cd2576c8ad2979ac16bd3)

libminifi/test/rocksdb-tests/EncryptionTests.cpp
adamdebreceni: as I understand encryption does not happen on the key-value pair level, but on the file level, so there is no way to get the encrypted value (or key)

szaszm: ok, I see.

extensions/rocksdb-repos/encryption/RocksDbEncryptionProvider.cpp
adamdebreceni: the interface is not my design, `rocksdb::BlockCipher` is the base class for this cipher, and there both `Encrypt` and `Decrypt` have the comment: `"Length of data is equal to BlockSize()."`

libminifi/include/utils/crypto/ciphers/Aes256Ecb.h
adamdebreceni: indeed this would have misbehaved, fixed in [09e2a1e0](https://github.com/apache/nifi-minifi-cpp/pull/1090/commits/09e2a1e0686a320b4e29264b5ff830b0cc74d471)

extensions/rocksdb-repos/encryption/RocksDbEncryptionProvider.cpp
adamdebreceni: done, for future reference do you remember which debugger struggled?

szaszm: I was using CLion, I think with the LLDB backend.

libminifi/src/utils/crypto/ciphers/Aes256Ecb.cpp
```diff
@@ -0,0 +1,122 @@
...
+    handleError("Could not create cipher context");
+  }
+
+  if (1 != EVP_EncryptInit_ex(ctx.get(), EVP_aes_256_ecb(), nullptr, encryption_key_.data(), nullptr)) {
+    handleError("Could not initialize encryption cipher context");
+  }
+
+  if (1 != EVP_CIPHER_CTX_set_padding(ctx.get(), 0)) {
+    handleError("Could not disable padding for cipher");
+  }
```
szaszm, line 63: Why do you want to disable padding? It pads so that the output is always a multiple of the block size. In theory we should always get one BLOCK_SIZE output for one BLOCK_SIZE input, so it shouldn't matter.

extensions/rocksdb-repos/encryption/RocksDbEncryptionProvider.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+ private:
+  std::unique_ptr<Env> env_;
+  std::shared_ptr<AES256BlockCipher> cipher_;
+};
+
+std::shared_ptr<logging::Logger> AES256BlockCipher::logger_ = logging::LoggerFactory<AES256BlockCipher>::getLogger();
+
+std::shared_ptr<rocksdb::Env> createEncryptingEnv(const utils::crypto::EncryptionManager& manager, const DbEncryptionOptions& options) {
+  auto cipher_impl = manager.createAes256EcbCipher(options.encryption_key_name);
```
szaszm, line 77: Can you explain why you chose ECB encryption mode? According to wikipedia, it's not recommended for for use in cryptographic protocols. https://en.wikipedia.org/wiki/Block_cipher_mode_of_operation#Electronic_codebook_(ECB)

Block cipher modes also don't ensure data integrity and I don't know whether rocksdb has built-in data authentication mechanisms.

extensions/rocksdb-repos/FlowFileRepository.cpp
```diff
@@ -220,17 +231,24 @@ void FlowFileRepository::initialize_repository() {
...
+  rocksdb::Status checkpoint_status = opendb->NewCheckpoint(&checkpoint);
+  if (!checkpoint_status.ok()) {
+    logger_->log_error("Could not create checkpoint object: %s", checkpoint_status.ToString());
+    return;
+  }
+  checkpoint_status = checkpoint->CreateCheckpoint(checkpoint_dir_);
+  if (!checkpoint_status.ok()) {
+    logger_->log_error("Could not initialize checkpoint: %s", checkpoint_status.ToString());
+    return;
+  }
```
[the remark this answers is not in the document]
adamdebreceni: done in [0e20dfe8](https://github.com/apache/nifi-minifi-cpp/pull/1090/commits/0e20dfe88485cade5969d1e42769d3c3ea02c0ea)

libminifi/src/utils/crypto/ciphers/Aes256Ecb.cpp
```diff
@@ -0,0 +1,122 @@
...
+    handleError("Invalid key length %zu bytes, expected %zu bytes", encryption_key_.size(), static_cast<size_t>(KEY_SIZE));
+  }
+}
+
+Bytes Aes256EcbCipher::generateKey() {
+  unsigned char key[KEY_SIZE];
+  if (1 != RAND_bytes(key, KEY_SIZE)) {
+    handleError("Couldn't generate key");
+  }
+  return Bytes(key, key + KEY_SIZE);
```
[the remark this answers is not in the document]
adamdebreceni: there is no `RAND_priv_bytes` neither on my machine nor on libressl's main, added error handling in [cfa43ced6](https://github.com/apache/nifi-minifi-cpp/pull/1090/commits/cfa43ced6793e2f7be94f8d92c92f84d712a8989)

libminifi/src/utils/crypto/ciphers/Aes256Ecb.cpp
```diff
@@ -0,0 +1,122 @@
...
+}
+
+void Aes256EcbCipher::encrypt(unsigned char *data) const {
+  EVP_CIPHER_CTX_ptr ctx(EVP_CIPHER_CTX_new(), EVP_CIPHER_CTX_free);
+  if (!ctx) {
+    handleError("Could not create cipher context");
+  }
+
+  if (1 != EVP_EncryptInit_ex(ctx.get(), EVP_aes_256_ecb(), nullptr, encryption_key_.data(), nullptr)) {
```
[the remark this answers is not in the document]
adamdebreceni: we are using the `CTREncryptionProvider`, which wraps our block cipher and provides the appropriate iv and counter per block

libminifi/src/utils/crypto/ciphers/Aes256Ecb.cpp
adamdebreceni: in theory yes, in practice I found that on BLOCK_SIZE input we get 2*BLOCK_SIZE output because `EVP_EncryptFinal_ex` interprets 'no more data' as 'please pad this to a block' for some reason, `EVP_EncryptUpdate` provides no guarantees of full encryption of all incoming data, so unfortunately we cannot drop the `EVP_EncryptFinal_ex` call

libminifi/include/utils/crypto/ciphers/Aes256Ecb.h
```diff
@@ -0,0 +1,69 @@
...
+
+class Aes256EcbCipher {
+  static std::shared_ptr<core::logging::Logger> logger_;
+ public:
+  static constexpr size_t BLOCK_SIZE = 16;
+  static constexpr size_t KEY_SIZE = 32;
+
+  explicit Aes256EcbCipher(Bytes encryption_key);
+  void encrypt(unsigned char* data) const;
+  void decrypt(unsigned char* data) const;
```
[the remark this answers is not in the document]
adamdebreceni: done in [cfa43ced](https://github.com/apache/nifi-minifi-cpp/pull/1090/commits/cfa43ced6793e2f7be94f8d92c92f84d712a8989)

extensions/rocksdb-repos/encryption/RocksDbEncryptionProvider.cpp
adamdebreceni: based on the interface of `BlockAccessCipherStream` (which an `EncryptionProvider` should produce) I am under the impression that authenticated encryption is not yet supported by `rocksdb`

```
  // Encrypt one or more (partial) blocks of data at the file offset.
  // Length of data is given in dataSize.
  virtual Status Encrypt(uint64_t fileOffset, char* data, size_t dataSize);

  // Decrypt one or more (partial) blocks of data at the file offset.
  // Length of data is given in dataSize.
  virtual Status Decrypt(uint64_t fileOffset, char* data, size_t dataSize);
```

libminifi/src/utils/crypto/ciphers/Aes256Ecb.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+std::shared_ptr<core::logging::Logger> Aes256EcbCipher::logger_{core::logging::LoggerFactory<Aes256EcbCipher>::getLogger()};
+
+Aes256EcbCipher::Aes256EcbCipher(Bytes encryption_key) : encryption_key_(std::move(encryption_key)) {
+  if (encryption_key_.size() != KEY_SIZE) {
+    handleError("Invalid key length %zu bytes, expected %zu bytes", encryption_key_.size(), static_cast<size_t>(KEY_SIZE));
+  }
+}
+
+Bytes Aes256EcbCipher::generateKey() {
```
fgerlits, line 53: we already have `utils::crypto::randomBytes()` in EncryptionUtils which does the same thing (but instead of libressl it uses libsodium, which is better according to @alopresto)

libminifi/test/rocksdb-tests/EncryptionTests.cpp
```diff
@@ -0,0 +1,108 @@
...
+};
+
+TEST_CASE_METHOD(FFRepoFixture, "FlowFileRepository creates checkpoint and loads flowfiles") {
+  SECTION("Without encryption") {
+    // pass
+  }
+  SECTION("With encryption") {
+    utils::file::FileUtils::create_dir((home_ / "conf").str());
+    std::ofstream{(home_ / "conf" / "bootstrap.conf").str()}
+      << static_cast<const char*>(FlowFileRepository::ENCRYPTION_KEY_NAME) << "="
```
fgerlits, line 78: why is this cast needed?  `FlowFileRepository::ENCRYPTION_KEY_NAME` is already a `const char*`

extensions/rocksdb-repos/database/RocksDbInstance.cpp
```diff
@@ -99,6 +112,7 @@ utils::optional<OpenRocksDb> RocksDbInstance::open(const std::string& column, co
       return utils::nullopt;
     }
     gsl_Expects(db_instance);
+    db_options_patch_ = db_options_patch;
```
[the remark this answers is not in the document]
adamdebreceni: added comment why we must store this object in [b13b5b7d](https://github.com/apache/nifi-minifi-cpp/pull/1090/commits/b13b5b7d24b681c15e15850d36ce2922b5f2b9e9)

extensions/rocksdb-repos/database/RocksDbUtils.h
```diff
@@ -38,19 +38,14 @@ class Writable {
...
+    if (!comparator(target_.*member, value)) {
       target_.*member = value;
       is_modified_ = true;
     }
   }
 
-  template<typename Transformer, typename F>
-  void transform(F T::* member) {
-    set(member, Transformer::transform(target_.*member));
-  }
```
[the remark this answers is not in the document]
adamdebreceni: yes, removed in [b13b5b7](https://github.com/apache/nifi-minifi-cpp/pull/1090/commits/b13b5b7d24b681c15e15850d36ce2922b5f2b9e9)

libminifi/src/utils/crypto/ciphers/Aes256Ecb.cpp
adamdebreceni: changed it to use `utils::crypto::randomBytes` in [b13b5b7d](https://github.com/apache/nifi-minifi-cpp/pull/1090/commits/b13b5b7d24b681c15e15850d36ce2922b5f2b9e9)

libminifi/test/rocksdb-tests/EncryptionTests.cpp
adamdebreceni: since we have an rvalue ref stream `operator<<` will resolve to a forwarding template function
```
template <class _Stream, class _Tp>
inline _LIBCPP_INLINE_VISIBILITY
typename enable_if
<
    !is_lvalue_reference<_Stream>::value &&
    is_base_of<ios_base, _Stream>::value,
    _Stream&&
>::type
operator<<(_Stream&& __os, const _Tp& __x)
{
    __os << __x;
    return _VSTD::move(__os);
}
```

which takes the argument by reference, which makes the static member [odr-used](https://en.cppreference.com/w/cpp/language/definition#ODR-use), adding a `static_cast` makes an lvalue-to-rvalue conversion which evaluates to a constant expression, and no longer odr-used, from `c++17` we won't need this cast

fgerlits (approved):

martinzink (approved):

extensions/rocksdb-repos/encryption/RocksDbEncryptionProvider.cpp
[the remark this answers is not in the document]
szaszm: Since `database_` is only used for logging, I still think that what we now call `hasEqualKey` is semantically an equality comparison.

libminifi/src/utils/crypto/ciphers/Aes256Ecb.cpp
szaszm: Could you add a comment with this info above these lines? This is useful context for the future reader IMO.

extensions/rocksdb-repos/encryption/RocksDbEncryptionProvider.cpp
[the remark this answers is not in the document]
adamdebreceni: changed them to `operator==` for the ciphers, but left the `hasEqualKey` for the `EncryptingEnv` in [2b211f45](https://github.com/apache/nifi-minifi-cpp/pull/1090/commits/2b211f4547de1e60e4b37bf8c86480e54d723d67)

libminifi/src/utils/crypto/ciphers/Aes256Ecb.cpp
adamdebreceni: added comment in [2b211f45](https://github.com/apache/nifi-minifi-cpp/pull/1090/commits/2b211f4547de1e60e4b37bf8c86480e54d723d67)

extensions/rocksdb-repos/encryption/RocksDbEncryptionProvider.cpp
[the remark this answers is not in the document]
szaszm: ok, that makes sense

libminifi/src/utils/crypto/ciphers/Aes256Ecb.cpp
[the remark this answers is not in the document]
szaszm: I've found CTREncryptionProvider in the rocksdb code, but couldn't find where we specify that we use it in this PR. Is it always used with rocksdb?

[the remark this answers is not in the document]
adamdebreceni: no we explicitly create one in `createEncryptingEnv`

extensions/rocksdb-repos/encryption/RocksDbEncryptionProvider.cpp
szaszm: Back to the first question, you mentioned CTR (a.k.a counter mode encryption) in a different answer. This makes me think that we are just using ECB functions to encrypt the nonce + counter as a block, but not really using ECB mode encryption. Am I correct with this assumption?

Here's the rocksdb CTR function: https://github.com/facebook/rocksdb/commit/51778612c9cf4cb842eed10f270ec0ed29ff22f1#diff-31e5f5f565771d096323f879c644c2792db9bf42018b12ca0441f09b2c876b39R803

Second (authentication): I agree, I can't see anything related to message authentication codes in the above commit or the latest rocksdb env/env_encryption* source files. This is a pity, because an attacker can easily flip bits in the plaintext by just flipping the corresponding bits in the ciphertext.

libminifi/include/utils/crypto/ciphers/Aes256Ecb.h
```diff
@@ -0,0 +1,73 @@
...
+namespace minifi {
+namespace utils {
+namespace crypto {
+
+class CipherError : public Exception {
+ public:
+  explicit CipherError(const std::string& error_msg) : Exception(ExceptionType::GENERAL_EXCEPTION, error_msg) {}
+};
+
+class Aes256EcbCipher {
```
szaszm, line 40: This and any ECB-related class needs a comment that it is unsafe to use for anything except with a CTR/counter mode implementation like the one in rocksdb.

szaszm (commented):
This feature needs documentation. Could you extend CONFIGURE.md with a guide on how to use rocksdb encryption?

extensions/rocksdb-repos/encryption/RocksDbEncryptionProvider.cpp
adamdebreceni: ECB mode would be `ciphertext := encrypt(plaintext)`, whereas CTR is `ciphertext := plantext XOR encrypt(concat(iv, counter))`, I could remove the `"Ecb"` from the `"Aes256EcbCipher"` as these modes do not really describe the cipher but the way they are used for encryption, but I feel like it prompts (as prompted in you) careful consideration when used in the future (as it should)

I share your disappointment for the lack of authentication, that is why we had to use the openssl library instead of the libsodium, as the latter only provides authenticated encryption

libminifi/include/utils/crypto/ciphers/Aes256Ecb.h
adamdebreceni: added documentation and comments on safety in [650a76f](https://github.com/apache/nifi-minifi-cpp/pull/1090/commits/650a76f0ecbf23f6f32e54d51b9f769192ff532c)

szaszm (approved):

PR #1094, opened 2021-06-02
MINIFICPP-1576 Allow build of all extensions in docker and cleanup docker files

lordgamez:
- Add and cleanup dependencies in docker image required by extensions
- Fix extension builds without glibc (for alpine base image)
- Cleanup .dockerignore file
- Avoid extension rebuild in CPack preinstall phase while building docker image
- Merge docker image and minimal docker image build phases
- Fix dockerfile linter issues

### For all changes:
     in the commit message?

### For code changes:

### Note:


CMakeLists.txt
```diff
@@ -583,8 +583,11 @@ add_subdirectory(main)
 add_subdirectory(nanofi)
 add_dependencies(nanofi minifiexe)
 
-add_subdirectory(encrypt-config)
-add_dependencies(encrypt-config minifi)
+option(DISABLE_ENCRYPT_CONFIG "Disables build of encrypt-config binary." OFF)
+if (NOT DISABLE_ENCRYPT_CONFIG)
```
szaszm, line 587: There was a discussion about double negatives in the ConsumeJournald PR. It should apply here, too. https://github.com/apache/nifi-minifi-cpp/pull/1044#discussion_r622005451

cmake/DockerConfig.cmake
```diff
@@ -65,10 +65,19 @@ add_custom_target(
...
+        -c ENABLE_LIBRDKAFKA=ON
+        -c ENABLE_AWS=ON
+        -c DISABLE_CONTROLLER=ON
+        -c DISABLE_SCRIPTING=ON
+        -c DISABLE_PYTHON_SCRIPTING=ON
+        -c DISABLE_ENCRYPT_CONFIG=ON
+        -c AWS_ENABLE_UNITY_BUILD=OFF
         -c DOCKER_BASE_IMAGE=${DOCKER_BASE_IMAGE}
         -c BUILD_NUMBER=${BUILD_NUMBER}
+        -c CMAKE_BUILD_TYPE=MinSizeRel
```
szaszm, line 80: I don't think a minimal image should mean size-optimized. I'd rather have inlining unless it's about as embedded as an arduino.

thirdparty/openwsman/openwsman.patch
```diff
@@ -79,3 +79,39 @@ diff -rupN orig/src/lib/wsman-soap.c patched/src/lib/wsman-soap.c
...
+
+diff -rupN orig/src/lib/u/lock.c patched/src/lib/u/lock.c
+--- orig/src/lib/u/lock.c	2021-05-31 13:44:43.992941115 +0200
++++ patched/src/lib/u/lock.c	2021-05-31 12:00:21.972733061 +0200
+@@ -50,7 +50,7 @@
+ extern int pthread_mutexattr_settype(pthread_mutexattr_t *attr, int type);
+ #endif
+ 
+-#if defined (__SVR4) && defined (__sun)
++#if (defined (__SVR4) && defined (__sun)) || !defined(__GLIBC__)
```
szaszm, line 91: why?

thirdparty/openwsman/openwsman.patch
```diff
@@ -79,3 +79,39 @@ diff -rupN orig/src/lib/wsman-soap.c patched/src/lib/wsman-soap.c
...
+
+diff -rupN orig/include/u/lock.h patched/include/u/lock.h
+--- orig/include/u/lock.h	2021-05-31 13:44:43.992941115 +0200
++++ patched/include/u/lock.h	2021-05-31 12:00:30.792726402 +0200
+@@ -2,7 +2,7 @@
+ #ifndef LOCKING_H
+ #define LOCKING_H
+ 
+-#if defined (__FreeBSD__)  || defined (__OpenBSD__) || defined (__NetBSD__) || defined (__APPLE__)
++#if defined (__FreeBSD__)  || defined (__OpenBSD__) || defined (__NetBSD__) || defined (__APPLE__) || !defined(__GLIBC__)
```
szaszm, line 113: why?

cmake/DockerConfig.cmake
lordgamez: Actually the minimal image was at least party about size optimization. It was meant for use cases in cloud environments where MiNiFi is used as a sidecar for every Kubernetes pod for log collection. In this case we only wanted to have a very small image and only include what is necessary for this use case as sidecards should be minimal in size if they are used in every pod. It's not a huge difference if we use the Release build type instead of MinSizeRel as it's 33MB or 39.2MB image though. What would be the benefit of the Release build in this use case?

thirdparty/openwsman/openwsman.patch
lordgamez: PTHREAD_MUTEX_RECURSIVE_NP is a non-portable glibc mutex and our docker base distro alpine does not have glibc only musl is available. On these systems only PTHREAD_MUTEX_RECURSIVE mutex is available that's why we redefine the non-portable symbol. An issue with a similar solution: https://github.com/godotengine/godot/issues/31555

thirdparty/openwsman/openwsman.patch
lordgamez: Same as above

CMakeLists.txt
lordgamez: Thanks, updated in d6f56f9a2a7808d459cb72afc7da5f2f091fd710

.dockerignore
```diff
@@ -1,14 +1,75 @@
...
+libminifi/src/agent/agent_version.cpp
+
+**/assemblies
+**/CMakeCache.txt
+**/CMakeFiles
+**/CMakeScripts
+**/cmake_install.cmake
+**/install_manifest.txt
+**/CTestTestfile.cmake
+**/cmake-build-debug
```
fgerlits, line 73: this is already covered by line 39

fgerlits (approved):

.dockerignore
lordgamez: Removed in 2c3a22a89c66a830ed70851a6128e9026872367a

cmake/DockerConfig.cmake
szaszm: I agree with the first part. By size-optimized, I meant `-Os`. In some cases it gives up inlining, which would enable many other optimizations, loop unrolling and a bunch of other compiler optimizations that increases the size of the binary. I would only go that route if there is a constraint forcing us to do so.

thirdparty/openwsman/openwsman.patch
szaszm: Could you add this answer to the top of the patch file? You can add random text above the diff command line, like in thirdparty/bzip2/bzip2.patch

lordgamez: Good idea, added in 2192050b5ae17cbad8c4c700fb072e53c615c644

cmake/DockerConfig.cmake
lordgamez: Thanks, I see what you mean. As we don't have specific size requirements and as I think under 39.2MB size is still okay, I changed the minimal image to use the `Release` build for those benefits in 2192050b5ae17cbad8c4c700fb072e53c615c644

docker/Dockerfile
```diff
@@ -16,45 +16,75 @@
...
+ARG DISABLE_CIVET=OFF
+ARG DISABLE_EXPRESSION_LANGUAGE=OFF
+ARG DISABLE_ROCKSDB=OFF
+ARG DISABLE_LIBARCHIVE=OFF
+ARG DISABLE_LZMA=OFF
+ARG DISABLE_BZIP2=OFF
+ARG DISABLE_SCRIPTING=OFF
+ARG DISABLE_PYTHON_SCRIPTING=OFF
+ARG DISABLE_CONTROLLER=OFF
+ARG CMAKE_BUILD_TYPE=Release
```
szaszm, line 62: Is it intentional that the systemd extension is missing? I assume WEL and PDH are missing because they are windows-only. A comment about these would be nice.

szaszm (approved):

docker/Dockerfile
lordgamez: It was not intentional, as only the previously present arguments were copied from the build phase here, so I suppose it was missing beforehand. After considering it I explicitly turned the SYSTEMD extension off as it wouldn't be of use inside and alpine container and removed the switch for WEL as well in 523df830a0509fbdc67fb98dd4d5d28f97d9db68

PR #1093, opened 2021-06-02
MINIFICPP-1578 Upgrade OS versions in docker builds

fgerlits:
https://issues.apache.org/jira/browse/MINIFICPP-1578

Upgrade OS versions: centos 7 -> 8, debian stretch -> buster, fedora 29 -> 34, ubuntu xenial -> focal.

---

Thank you for submitting a contribution to Apache NiFi - MiNiFi C++.

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [x] Is there a JIRA ticket associated with this PR? Is it referenced
     in the commit message?

- [x] Does your PR title start with MINIFICPP-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.

- [x] Has your PR been rebased against the latest commit within the target branch (typically main)?

- [x] Is your initial contribution a single, squashed commit?

### For code changes:
- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?
- [ ] If applicable, have you updated the LICENSE file?
- [ ] If applicable, have you updated the NOTICE file?

### For documentation related changes:
- [ ] Have you ensured that format looks appropriate for the output in which it is rendered?

### Note:
Please ensure that once the PR is submitted, you check GitHub Actions CI results for build issues and submit an update to your PR as soon as possible.

PR #1098, opened 2021-06-04
MINIFICPP-1581: set cmake policy CMP0096 to NEW to preserve the leadi…

martinzink:
…ng zeros in project version

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,8 @@
 # under the License.
 #
 
-cmake_minimum_required(VERSION 3.11)
+cmake_minimum_required(VERSION 3.16)
+cmake_policy(SET CMP0096 NEW) # policy to preserve the leading zeros in PROJECT_VERSION_{MAJOR,MINOR,PATCH,TWEAK}
 cmake_policy(SET CMP0065 OLD) # default export policy, required for self-dlopen
 project(nifi-minifi-cpp VERSION 0.10.0)
 set(PROJECT_NAME "nifi-minifi-cpp")
--- a/docker/debian/Dockerfile
+++ b/docker/debian/Dockerfile
@@ -28,7 +28,9 @@ ENV MINIFI_BASE_DIR /opt/minifi
 ENV MINIFI_HOME $MINIFI_BASE_DIR/nifi-minifi-cpp-$MINIFI_VERSION
 ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64/
 
-RUN apt-get update && apt-get install -y openjdk-11-jdk openjdk-11-source libpython3.7-dev sudo git maven libarchive13 cmake
+RUN echo "deb http://deb.debian.org/debian buster-backports main" | tee -a /etc/apt/sources.list \
+    && apt-get update && apt-get install -y openjdk-11-jdk openjdk-11-source libpython3.7-dev sudo git maven libarchive13 libtool \ 
+    && apt-get -t buster-backports install -y cmake
 
 RUN mkdir -p $MINIFI_BASE_DIR
 COPY . ${MINIFI_BASE_DIR}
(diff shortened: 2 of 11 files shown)


szaszm:
libminifi and probably nanofi are separate projects with their corresponding root CMakeLists.txt files. You may want to apply the change there as well.

martinzink:
Unrelated ci failures 
https://issues.apache.org/jira/browse/MINIFICPP-1583
https://issues.apache.org/jira/browse/MINIFICPP-1582

fgerlits:
I thought you were going to replace occurrences of `${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH}` with `${PROJECT_VERSION}`, which would solve the original issue without forcing an upgrade to CMake 3.16.  Did that plan fail?

martinzink:
> I thought you were going to replace occurrences of `${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH}` with `${PROJECT_VERSION}`, which would solve the original issue without forcing an upgrade to CMake 3.16. Did that plan fail?

Yeah, that didnt fail. I just found out that all of our targets use newer cmake anyways.
Except for debian but were using this backported (3.16 version) prior to #1093 as well.

I still think that we could use the PROJECT_VERSION variable exclusively but not sure if leaving these separate variables inconsistent is a good idea in the long term (maybe some tool uses them?)

fgerlits (approved):

docker/debian/Dockerfile
```diff
@@ -28,7 +28,9 @@ ENV MINIFI_BASE_DIR /opt/minifi
 ENV MINIFI_HOME $MINIFI_BASE_DIR/nifi-minifi-cpp-$MINIFI_VERSION
 ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64/
 
-RUN apt-get update && apt-get install -y openjdk-11-jdk openjdk-11-source libpython3.7-dev sudo git maven libarchive13 cmake
+RUN echo "deb http://deb.debian.org/debian buster-backports main" | tee -a /etc/apt/sources.list \
+    && apt-get update && apt-get install -y openjdk-11-jdk openjdk-11-source libpython3.7-dev sudo git maven libarchive13 libtool \
```
szaszm, line 32: Why are we installing openjdk-11 while setting JAVA_HOME to openjdk-8 three lines above?

CMakeLists.txt
```diff
@@ -17,7 +17,8 @@
 # under the License.
 #
 
-cmake_minimum_required(VERSION 3.11)
+cmake_minimum_required(VERSION 3.16)
```
szaszm, line 20: When raising cmake or other requirements, it needs to be documented in the README.

docker/debian/Dockerfile
martinzink: Not sure. I just modified the cmake source. It was changed in #1093 , and we are still using and installing java 8 in ubuntu for example.
@fgerlits what should be here?

fgerlits: Java 8 was not available in Centos 8, so I changed it to Java 11, but forgot to update the JAVA_HOME line.  I would just update that to Java 11, too.

martinzink: updated in https://github.com/apache/nifi-minifi-cpp/pull/1098/commits/ca4eaa48a97ac2da56937811589de8d3c8a02161

CMakeLists.txt
martinzink: updated in https://github.com/apache/nifi-minifi-cpp/pull/1098/commits/b20daf04ede9c0d02c1934b75a00da6254cc2e1f

szaszm (approved):

arpadboda (approved):

PR #1101, opened 2021-06-07
MINIFICPP-1557 Fix CPU usage percentage calculation and extend verification

lordgamez:
`std::thread::hardware_concurrency` can return with 0 if the value is not well defined or not computable, which can cause zero division in the calculation of the CPU usage. The total time difference calculations can also be zero that can also cause zero division. This happened quite often in github's virtual environments, that resulted in the heartbeat payload JSON to be truncated, and the invalid JSON caused the test to fail transiently. 

The test is also updated to include more information of the parsing failure and to avoid any timing issues when the server is shut down.

-------------------------------------------------------------------------------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/http-curl/tests/HTTPHandlers.h
+++ b/extensions/http-curl/tests/HTTPHandlers.h
@@ -27,6 +27,7 @@
 #include "CivetStream.h"
 #include "io/CRCStream.h"
 #include "rapidjson/document.h"
+#include "rapidjson/error/en.h"
 #include "HTTPUtils.h"
 #include "ServerAwareHandler.h"
 #include "utils/gsl.h"
@@ -422,11 +423,15 @@ class HeartbeatHandler : public ServerAwareHandler {
 
   void verify(struct mg_connection *conn) {
     auto post_data = readPayload(conn);
+    if (!isServerRunning()) {
+      return;
+    }
     if (!IsNullOrEmpty(post_data)) {
       rapidjson::Document root;
-      rapidjson::ParseResult ok = root.Parse(post_data.data(), post_data.size());
-      assert(ok);
-      (void)ok;  // unused in release builds
+      rapidjson::ParseResult result = root.Parse(post_data.data(), post_data.size());
+      if (!result) {
+        throw std::runtime_error("JSON parse error: " + std::string(rapidjson::GetParseError_En(result.Code())) + "\n JSON data: " + post_data);
+      }
       std::string operation = root["operation"].GetString();
       if (operation == "heartbeat") {
         handleHeartbeat(root, conn);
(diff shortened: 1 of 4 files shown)


libminifi/src/utils/ProcessCpuUsageTracker.cpp
```diff
@@ -126,7 +127,7 @@ double ProcessCpuUsageTracker::getProcessCpuUsageBetweenLastTwoQueries() const {
   uint64_t sys_cpu_times_diff = sys_cpu_times_ - previous_sys_cpu_times_;
   uint64_t user_cpu_times_diff = user_cpu_times_ - previous_user_cpu_times_;
   double percent = static_cast<double>(sys_cpu_times_diff + user_cpu_times_diff) / static_cast<double>(cpu_times_diff);
-  percent = percent / std::thread::hardware_concurrency();
+  percent = percent / (std::max)(static_cast<uint32_t>(1), std::thread::hardware_concurrency());
```
szaszm, line 130: `uint32_t{1}` is shorter and avoids the cast.

extensions/http-curl/tests/HTTPHandlers.h
```diff
@@ -422,11 +423,15 @@ class HeartbeatHandler : public ServerAwareHandler {
...
+      return;
+    }
     if (!IsNullOrEmpty(post_data)) {
       rapidjson::Document root;
-      rapidjson::ParseResult ok = root.Parse(post_data.data(), post_data.size());
-      assert(ok);
-      (void)ok;  // unused in release builds
+      rapidjson::ParseResult result = root.Parse(post_data.data(), post_data.size());
+      if (!result) {
+        throw std::runtime_error("JSON parse error: " + std::string(rapidjson::GetParseError_En(result.Code())) + "\n JSON data: " + post_data);
```
szaszm, line 433: `fmt::format` or `StringUtils::join_pack` would be nicer IMO. `fmt::format` allows nice format string. If you don't like that, `StringUtils::join_pack` can avoid intermediate allocation and deallocation of temporaries.

lordgamez: Updated in a199f6907d09938d30af9fe21b65ffc0e5116cb2

libminifi/src/utils/ProcessCpuUsageTracker.cpp
lordgamez: Updated in a199f6907d09938d30af9fe21b65ffc0e5116cb2

szaszm (approved):

martinzink (approved):

arpadboda (approved):

fgerlits:
We also divide by `hardware_concurrency()` in `libminifi/test/unit/CpuUsageTest.cpp`; I think a fix for that could be included in this PR.

lordgamez:
> We also divide by `hardware_concurrency()` in `libminifi/test/unit/CpuUsageTest.cpp`; I think a fix for that could be included in this PR.

Added fix in cbc639977e11570075f894cf346840f1f7c486a3

fgerlits (approved):

adamdebreceni (approved):

PR #1107, opened 2021-06-11
MINIFICPP-1338 - Handle processor state persistence through the session

adam-markovics:

### For all changes:
     in the commit message?

### For code changes:

### Note:


cmake/BuildTests.cmake
```diff
@@ -151,11 +151,13 @@ SET(INT_TEST_COUNT 0)
 FOREACH(testfile ${INTEGRATION_TESTS})
   get_filename_component(testfilename "${testfile}" NAME_WE)
   add_executable("${testfilename}" "${TEST_DIR}/integration/${testfile}")
+  target_wholearchive_library("${testfilename}" minifi-standard-processors)
```
szaszm, line 154: What changed that makes this link necessary?

extensions/sql/processors/QueryDatabaseTable.cpp
```diff
@@ -220,7 +219,9 @@ void QueryDatabaseTable::initializeMaxValues(core::ProcessContext &context) {
     logger_->log_info("Found no stored state");
   } else {
     if (!loadMaxValuesFromStoredState(stored_state)) {
+      state_manager_->beginTransaction();
       state_manager_->clear();
+      state_manager_->commit();
```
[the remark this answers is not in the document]
adam-markovics: From now on state transactions begin with beginTransaction and end with either commit or rollback. In onTrigger beginTransaction and commit (or rollback) are called automatically by the creation and commit/rollback calls on ProcessSession. Processor::onTrigger wraps the calling of the derived processor's onTrigger (different signature) and manages the session.
This method above is called from onSchedule. For onSchedule there is no session, so it has to be manual.

cmake/BuildTests.cmake
adam-markovics: UnorderedMapKeyValueStoreService (or some other state storage) has to be registered, otherwise we can't use it. Registration is done at static initialization time using the REGISTER_RESOURCE macro. It would be optimized out if not linked and test would fail.

szaszm: And do we need whole archive, wouldn't normal linking be enough? I think static initialization would be done even with normal linking, but I'm not 100% sure.

adam-markovics: My experience is that it would be optimized out. Also seen this in another project. Without this line the test would fail (it did until I came up with this). Maybe it also depends on linkers and their settings.
https://stackoverflow.com/questions/7327741/initialisation-of-static-object-when-linking-against-a-static-library

extensions/windows-event-log/tests/BookmarkTests.cpp
```diff
@@ -40,7 +40,10 @@ std::unique_ptr<Bookmark> createBookmark(TestPlan &test_plan,
                                          const utils::Identifier &uuid = IdGenerator::getIdGenerator()->generate()) {
   const auto state_manager = test_plan.getStateManagerProvider()->getCoreComponentStateManager(uuid);
   const auto logger = test_plan.getLogger();
-  return utils::make_unique<Bookmark>(channel, L"*", "", uuid, false, state_manager, logger);
+  state_manager->beginTransaction();
+  auto bookmark = utils::make_unique<Bookmark>(channel, L"*", "", uuid, false, state_manager, logger);
+  state_manager->commit();
```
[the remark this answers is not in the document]
szaszm: Could we somehow make this easier to use correctly and hard/impossible to use incorrectly? I'm thinking about transaction objects having state mutating capabilities, and the state manager only handing them out. onTrigger could get one as an argument.

cmake/BuildTests.cmake
szaszm: I think we have UnorderedMapKeyValueStoreService working on main without this change. I wonder how was it working before this change?

extensions/windows-event-log/tests/BookmarkTests.cpp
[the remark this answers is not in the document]
adam-markovics: If you want to make a change to the state (set or clear) without being in a transaction, your call will fail, returning false.
I don't think it is possible to auto-commit outside of onTrigger. It is the user's responsibility to decide when/how often/if to commit or rollback. I cannot think of a way that is universally applicable everywhere.
Transaction objects wouldn't solve this problem, and they would break preexisting interface of state manager, affecting all users, even in onTriggers.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -332,7 +334,6 @@ void ConsumeWindowsEventLog::onSchedule(const std::shared_ptr<core::ProcessConte
 bool ConsumeWindowsEventLog::commitAndSaveBookmark(const std::wstring &bookmark_xml, const std::shared_ptr<core::ProcessSession> &session) {
   {
     const TimeDiff time_diff;
-    session->commit();
```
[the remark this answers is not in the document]
adam-markovics: Done.

cmake/BuildTests.cmake
adam-markovics: In which executable? This one affects integration tests.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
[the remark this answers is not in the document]
szaszm: For the record: This was a bad idea, I mistakenly identified the first block as the function body. We discussed the solution in private.

extensions/windows-event-log/tests/BookmarkTests.cpp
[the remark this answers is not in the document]
adam-markovics: A conclusion was made that state changes would be autocommit (SQL-like) when outside of a transaction (such as in onTrigger). A transaction would begin and at the end of the call it would finish with commit. If users want to, they can still begin and commit/rollback a transaction manually if they wish so.

cmake/BuildTests.cmake
adam-markovics: The same whole archive linking is done for unit tests of standard processors, nanofi tests and many other tests.
The main minifiexe binary is also linking it through linking all the extensions (including standard processors) in main/CMakeLists.txt:72:
`target_wholearchive_library(minifiexe ${EXTENSION})`

adam-markovics: Line was moved after discussion in private.

libminifi/include/controllers/keyvalue/AbstractCoreComponentStateManagerProvider.h
```diff
@@ -66,8 +76,11 @@ class AbstractCoreComponentStateManagerProvider : public std::enable_shared_from
   virtual bool removeImpl(const utils::Identifier& key) = 0;
   virtual bool persistImpl() = 0;
 
-  virtual std::string serialize(const core::CoreComponentState& kvs);
-  bool deserialize(const std::string& serialized, core::CoreComponentState& kvs);
+ private:
+  void removeFromCache(utils::Identifier id);
+
+  std::mutex mutex;
```
lordgamez, line 82: Use "_" suffix for member

libminifi/include/controllers/keyvalue/AbstractCoreComponentStateManagerProvider.h
```diff
@@ -34,29 +34,39 @@ namespace controllers {
...
+      CLEAR
+    };
+
     std::shared_ptr<AbstractCoreComponentStateManagerProvider> provider_;
     utils::Identifier id_;
     bool state_valid_;
     core::CoreComponentState state_;
+    bool transactionInProgress_;
+    ChangeType changeType_;
+    core::CoreComponentState stateToSet_;
```
lordgamez, line 69: Use snake_case_ format for member variables

libminifi/include/core/ProcessContext.h
```diff
@@ -362,6 +369,8 @@ class ProcessContext : public controller::ControllerServiceLookup, public core::
   controller::ControllerServiceProvider* controller_service_provider_;
   // state manager provider
   std::shared_ptr<core::CoreComponentStateManagerProvider> state_manager_provider_;
+  // state manager
+  std::shared_ptr<CoreComponentStateManager> stateManager_;
```
lordgamez, line 373: Use snake_case_ here as well (valid for any further member variables in the PR), also I think these member comments are unnecessary as they do not contain any useful info.

lordgamez (commented):
Looks mostly okay, just some minor comments

extensions/coap/tests/CoapIntegrationBase.h
```diff
@@ -37,7 +37,7 @@ int ssl_enable(void *, void *) {
 
 class CoapIntegrationBase : public IntegrationBase {
  public:
-  explicit CoapIntegrationBase(uint64_t waitTime = 5000)
+  explicit CoapIntegrationBase(std::chrono::milliseconds waitTime = std::chrono::milliseconds(5000))
```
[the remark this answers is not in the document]
adam-markovics: Done.

libminifi/include/controllers/keyvalue/AbstractCoreComponentStateManagerProvider.h
adam-markovics: Done.

extensions/windows-event-log/tests/BookmarkTests.cpp
```diff
@@ -41,7 +41,8 @@ std::unique_ptr<Bookmark> createBookmark(TestPlan &test_plan,
                                          const utils::Identifier &uuid = IdGenerator::getIdGenerator()->generate()) {
   const auto state_manager = test_plan.getStateManagerProvider()->getCoreComponentStateManager(uuid);
   const auto logger = test_plan.getLogger();
-  return std::make_unique<Bookmark>(channel, L"*", "", uuid, false, state_manager, logger);
+  auto bookmark = std::make_unique<Bookmark>(channel, L"*", "", uuid, false, state_manager, logger);
+  return bookmark;
```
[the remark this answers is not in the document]
adam-markovics: Honestly, I don't remember. I think there were other commits where there were operations before returning, and they weren't completely cleaned up. I will revert and test it.

libminifi/include/controllers/keyvalue/AbstractCoreComponentStateManagerProvider.h
adam-markovics: Done.

libminifi/include/core/ProcessContext.h
adam-markovics: Done.

lordgamez (approved):

martinzink (approved):
LGTM :+1:

PR #1123, opened 2021-07-01
MINIFICPP-1579 Add config file checksums to the heartbeat

fgerlits:
https://issues.apache.org/jira/browse/MINIFICPP-1579

Created a new ResponseNode type, ConfigurationChecksums, which can be added to `nifi.c2.root.classes` (not included by default) to produce a new section in the heartbeat JSON message which shows the SHA256 checksums of minifi.properties and config.yml.

The new section looks like this:
```
          "agentClass": "Test-Checksum",
          "identifier": "Test-Checksum-001"
      },
+     "configurationChecksums": {
+         "SHA256": {
+             "config.yml": "634efcf2af76db967f4189c6318b77f03f7cae5bb1c3de1200a10014d0fc0ca4",
+             "minifi.properties": "efc629585c4a2207d9e7f129150a49f2dc6cfd1ae8338788b30921984da2b6c6"
+         }
+     },
      "deviceInfo": {
          "systemInfo": {
```

The line containing the agent identifier is excluded from the checksum calculation of `minifi.properties` to make sure that agents in the same class have the same checksum if all other settings are the same.

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/core/FlowConfiguration.h
+++ b/libminifi/include/core/FlowConfiguration.h
@@ -40,6 +40,8 @@
 #include "io/StreamFactory.h"
 #include "core/state/nodes/FlowInformation.h"
 #include "utils/file/FileSystem.h"
+#include "utils/ChecksumCalculator.h"
+#include "utils/gsl.h"
 #include "utils/OptionalUtils.h"
 
 namespace org {
@@ -92,6 +94,7 @@ class FlowConfiguration : public CoreComponent {
         logger_->log_error("Couldn't find config file \"%s\".", *path);
         config_path_ = path;
       }
+      checksum_calculator_.setFileLocation(*config_path_);
     }
 
     // it is okay if this has already been called
@@ -162,6 +165,8 @@ class FlowConfiguration : public CoreComponent {
     }
   }
 
+  gsl::not_null<utils::ChecksumCalculator*> getChecksumCalculator() { return gsl::make_not_null(&checksum_calculator_); }
+
  protected:
   void registerResource(const std::string &resource_function) {
     core::ClassLoader::getDefaultClassLoader().registerResource("", resource_function);
@@ -185,8 +190,8 @@ class FlowConfiguration : public CoreComponent {
   std::shared_ptr<io::StreamFactory> stream_factory_;
   std::shared_ptr<Configure> configuration_;
   std::shared_ptr<state::response::FlowVersion> flow_version_;
-
   std::shared_ptr<utils::file::FileSystem> filesystem_;
+  utils::ChecksumCalculator checksum_calculator_;
 
  private:
   std::shared_ptr<logging::Logger> logger_;
--- a/libminifi/include/core/FlowConfiguration.h
+++ b/libminifi/include/core/FlowConfiguration.h
@@ -41,7 +41,6 @@
 #include "core/state/nodes/FlowInformation.h"
 #include "utils/file/FileSystem.h"
 #include "utils/ChecksumCalculator.h"
-#include "utils/gsl.h"
 #include "utils/OptionalUtils.h"
 
 namespace org {
@@ -165,7 +164,7 @@ class FlowConfiguration : public CoreComponent {
     }
   }
 
-  gsl::not_null<utils::ChecksumCalculator*> getChecksumCalculator() { return gsl::make_not_null(&checksum_calculator_); }
+  utils::ChecksumCalculator& getChecksumCalculator() { return checksum_calculator_; }
 
  protected:
   void registerResource(const std::string &resource_function) {
--- a/libminifi/include/core/state/nodes/ConfigurationChecksums.h
+++ b/libminifi/include/core/state/nodes/ConfigurationChecksums.h
@@ -1,0 +1,56 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <string>
+#include <vector>
+
+#include "core/Resource.h"
+#include "core/state/nodes/MetricsBase.h"
+#include "utils/ChecksumCalculator.h"
+#include "utils/gsl.h"
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace state {
+namespace response {
+
+class ConfigurationChecksums : public ResponseNode {
+ public:
+  ConfigurationChecksums() = default;
+  explicit ConfigurationChecksums(const std::string& name, const utils::Identifier& uuid = {}) : ResponseNode(name, uuid) {}
+
+  void addChecksumCalculator(gsl::not_null<utils::ChecksumCalculator*> checksum_calculator);
+
+  std::string getName() const override { return "configurationChecksums"; }
+  std::vector<SerializedResponseNode> serialize() override;
+
+ private:
+  std::vector<gsl::not_null<utils::ChecksumCalculator*>> checksum_calculators_;
+};
+
+REGISTER_RESOURCE(ConfigurationChecksums, "Node part of an AST that defines checksums of configuration files in the C2 protocol");
+
+}  // namespace response
+}  // namespace state
+}  // namespace minifi
+}  // namespace nifi
+}  // namespace apache
+}  // namespace org
--- a/libminifi/include/core/state/nodes/ConfigurationChecksums.h
+++ b/libminifi/include/core/state/nodes/ConfigurationChecksums.h
@@ -37,7 +37,7 @@ class ConfigurationChecksums : public ResponseNode {
   ConfigurationChecksums() = default;
   explicit ConfigurationChecksums(const std::string& name, const utils::Identifier& uuid = {}) : ResponseNode(name, uuid) {}
 
-  void addChecksumCalculator(gsl::not_null<utils::ChecksumCalculator*> checksum_calculator);
+  void addChecksumCalculator(utils::ChecksumCalculator& checksum_calculator);
 
   std::string getName() const override { return "configurationChecksums"; }
   std::vector<SerializedResponseNode> serialize() override;
--- a/libminifi/include/utils/ChecksumCalculator.h
+++ b/libminifi/include/utils/ChecksumCalculator.h
@@ -1,0 +1,57 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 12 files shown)


libminifi/test/unit/ChecksumCalculatorTests.cpp
```diff
@@ -0,0 +1,101 @@
...
+      "nifi.c2.agent.identifier=Test-222\n"
+      "nifi.c2.agent.heartbeat.period=10000\n");
+
+  utils::ChecksumCalculator checksum_calculator_1;
+  checksum_calculator_1.setFileLocation(file_location_1);
+  utils::ChecksumCalculator checksum_calculator_2;
+  checksum_calculator_2.setFileLocation(file_location_2);
+
+  REQUIRE(checksum_calculator_1.getChecksum() == checksum_calculator_2.getChecksum());
+}
```
lordgamez, line 101: Some additional simple tests could be added for error cases of `getChecksum` and `getFileName` if the file location is not set.

libminifi/src/utils/ChecksumCalculator.cpp
```diff
@@ -0,0 +1,98 @@
...
+  if (!input_file.is_open()) {
+    throw std::runtime_error(StringUtils::join_pack("Could not open config file '", file_location, "' to compute the checksum: ", std::strerror(errno)));
+  }
+
+  crypto_hash_sha256_state state;
+  crypto_hash_sha256_init(&state);
+
+  std::string line;
+  while (std::getline(input_file, line)) {
+    if (StringUtils::startsWith(line, AGENT_IDENTIFIER_KEY)) {
```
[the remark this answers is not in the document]
fgerlits: added a comment in https://github.com/apache/nifi-minifi-cpp/pull/1123/commits/13494a5ce0551e2bef85ade283d6d524ad502e7e

libminifi/test/unit/ChecksumCalculatorTests.cpp
fgerlits: added tests of the error cases in https://github.com/apache/nifi-minifi-cpp/pull/1123/commits/f6868eb85a9f286c6229984d42e03ecd95dd0e86

lordgamez (approved):

libminifi/include/utils/ChecksumCalculator.h
```diff
@@ -0,0 +1,57 @@
...
+
+  void setFileLocation(const std::string& file_location);
+  std::string getFileName() const;
+  std::string getChecksum();
+  void invalidateChecksum();
+
+ private:
+  static std::string computeChecksum(const std::string& file_location);
+
+  utils::optional<std::string> file_location_;
```
arpadboda, line 44: What's the purpose of optionals here?
Maybe I miss something trivial, but I would expect checking the path before making the decision to compute the checksum of the given file.

arpadboda (commented):
Mostly LGTM, nice tests, one minor question.

libminifi/include/utils/ChecksumCalculator.h
fgerlits: `Properties` and `FlowConfiguration` both contain a default-constructed `ChecksumCalculator`, but initially these are empty, so all these optionals are `nullopt`.  Setting the file location and computing the checksum only happens if the user enables it by adding `ConfigurationChecksums` to `nifi.c2.root.classes`.

martinzink (commented):
LGTM, only a minor question.
Shouldnt we document this feature somewhere?
I understand why its not on by default, but maybe a commented line should be added in a minifi.properties just to hint at this new feature (similar to the HeartbeatLogger)

libminifi/include/utils/ChecksumCalculator.h
arpadboda: Okay, that makes perfect sense, thanks!

arpadboda (approved):

martinzink (approved):

libminifi/src/utils/ChecksumCalculator.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+namespace org {
+namespace apache {
+namespace nifi {
+namespace minifi {
+namespace utils {
+
+void ChecksumCalculator::setFileLocation(const std::string& file_location) {
+  file_location_ = file_location;
+  file_name_ = utils::file::get_child_path(file_location);
```
adamdebreceni, line 40: should we invalidate the checksum here?

libminifi/src/utils/ChecksumCalculator.cpp
```diff
@@ -0,0 +1,96 @@
...
+    // skip lines containing the agent identifier, so agents in the same class will have the same checksum
+    if (StringUtils::startsWith(line, AGENT_IDENTIFIER_KEY)) {
+      continue;
+    }
+    if (!input_file.eof()) {  // eof() means we have just read the last line, which was not terminated by a newline
+      line.append("\n");
+    }
+    crypto_hash_sha256_update(&state, reinterpret_cast<const unsigned char*>(line.data()), line.size());
+  }
+  if (input_file.bad()) {
```
adamdebreceni, line 82: do we care about `fail()`?

libminifi/src/utils/ChecksumCalculator.cpp
fgerlits: Yes, we do, good catch, thanks!  (We don't currently call `setFileLocation()` twice on the same object, but someone may do that in the future.)  Fixed in https://github.com/apache/nifi-minifi-cpp/pull/1123/commits/7369c4998a7b56ac88a38a2b9dc4ded508193dff.

libminifi/src/utils/ChecksumCalculator.cpp
fgerlits: No, because the failbit will be set on the last line if it does not end with a newline, and in all real error cases, the badbit will be set.  See https://gehrcke.de/2011/06/reading-files-in-c-using-ifstream-dealing-correctly-with-badbit-failbit-eofbit-and-perror/ for details.

adamdebreceni (approved):

libminifi/include/core/FlowConfiguration.h
```diff
@@ -162,6 +165,8 @@ class FlowConfiguration : public CoreComponent {
     }
   }
 
+  gsl::not_null<utils::ChecksumCalculator*> getChecksumCalculator() { return gsl::make_not_null(&checksum_calculator_); }
```
szaszm, line 168: Why not return a reference? `gsl::not_null<utils::ChecksumCalculator*>` is only useful in case of non-static data members in my experience, to support move semantics and similar mutations.

libminifi/include/core/state/nodes/ConfigurationChecksums.h
```diff
@@ -0,0 +1,56 @@
...
+namespace minifi {
+namespace state {
+namespace response {
+
+class ConfigurationChecksums : public ResponseNode {
+ public:
+  ConfigurationChecksums() = default;
+  explicit ConfigurationChecksums(const std::string& name, const utils::Identifier& uuid = {}) : ResponseNode(name, uuid) {}
+
+  void addChecksumCalculator(gsl::not_null<utils::ChecksumCalculator*> checksum_calculator);
```
szaszm, line 40: Same here, I would take a reference and convert it to `gsl::not_null<utils::ChecksumCalculator*>` only when storing in the vector. `std::reference_wrapper` is a possible alternative storage type that represents the same concept but without null checking in the constructor.

libminifi/include/utils/ChecksumCalculator.h
```diff
@@ -0,0 +1,57 @@
...
+
+class ChecksumCalculator {
+ public:
+  static constexpr const char* CHECKSUM_TYPE = "SHA256";
+  static constexpr size_t LENGTH_OF_HASH_IN_BYTES = 32;
+
+  void setFileLocation(const std::string& file_location);
+  std::string getFileName() const;
+  std::string getChecksum();
+  void invalidateChecksum();
```
szaszm, line 39: Did you consider `stat`-ing the files on each call and recalculating the checksum based on the last modified timestamp? This would fix the case when the watched files were modified outside of minifi. An alternative faster, but way harder and more complex approach would be subscribing to filesystem events like `inotify` (Linux)/`FindFirstChangeNotification`(Windows)/`fsevents`(Mac OS X), and recalculating the hashes or invalidating the cache on change events.

szaszm (commented):
I would prefer a simplified design where caching and checksum calculation are separate. I imagine checksum calculation as a function call to something like `utils::crypto::sha256(utils::file::readAll(filename))`. I also recognize that this is harder to write and the current version is fine, especially given the existing number of approvals.
In other words, ChecksumCalculator violates the Single Responsibility Principle by having the responsibility of maintaining a manually invalidated cache and calculating the checksum of files. Separating the cache responsibility would allow for a reusable class that can cache anything in a certain way. (Manually invalidated single entry in this case.)

Typically classes that are named -Manager, -Handler, -er/-or are smells and make me look for design problems, because those tend to cover operations, that are better modelled as functions, instead of entities.

edit: The original checksum function suggestion also had an SRP violation

libminifi/include/utils/ChecksumCalculator.h
fgerlits: I decided not to do that because minifi does not automatically re-read the configuration if it is changed on the disk.  We want to send the checksum of the configuration files as currently used by minifi; this can only change at startup and when minifi itself modifies it in response to a C2 command.

For example, if minifi.properties is modified but the change is reverted before the next minifi restart, sending a modified checksum would be a false alert.

libminifi/include/utils/ChecksumCalculator.h
```diff
@@ -0,0 +1,57 @@
...
+namespace nifi {
+namespace minifi {
+namespace utils {
+
+class ChecksumCalculator {
+ public:
+  static constexpr const char* CHECKSUM_TYPE = "SHA256";
+  static constexpr size_t LENGTH_OF_HASH_IN_BYTES = 32;
+
+  void setFileLocation(const std::string& file_location);
```
szaszm, line 36: The class implementation could be simplified by requiring a valid file location at all times (setting the member in the constructor). If the usage side needs late initialization, it could rely on an optional or unique_ptr.

libminifi/include/utils/ChecksumCalculator.h
szaszm: That's fair, thanks

libminifi/include/utils/ChecksumCalculator.h
fgerlits: `ChecksumCalculator` would be simpler, but its usage would be more complex.  In both Properties.h and FlowConfiguration.h, instead of `{ return checksum_calculator_; }`, we would need to check if the optional is valid, and construct the object if it isn't.

libminifi/include/core/FlowConfiguration.h
fgerlits: Yes, good idea, that makes things simpler.

libminifi/include/utils/ChecksumCalculator.h
szaszm: We still return a useless object before it's initialized, it would just be a null pointer or other representation of "nothing" instead of a useless `ChecksumCalculator` that throws on calling `getChecksum`.

fgerlits:
> [@szaszm rephrased] Rewrite the whole thing to fix the SRP violation

I think it will be fairly easy to refactor this later, if we add another feature which requires calculating a checksum.

szaszm:
That's certainly an option. I also didn't want to discount the value of this feature with my criticism. If you want it merged like this, I won't stand in the way.

libminifi/include/utils/ChecksumCalculator.h
fgerlits: I could change the `if (!...) { throw ...}` to `gsl_Expects`.  That would express the logic more clearly, but then these lines could no longer be unit tested, as the test would terminate instead of throwing.  Do you think that would be better?

szaszm: I'm fine with not testing invalid calls if the preconditions are documented. We don't need to test what happens on `gsl_Expects(false)`.

By "the getter could have a precondition that the appropriate member must be initialized before calling it", I meant the getter of `checksum_calculator_` in FlowConfiguration.h and Properties.h, but I'm not against a similar change in `ChecksumCalculator` either.

fgerlits: Done (at least partially) in https://github.com/apache/nifi-minifi-cpp/pull/1123/commits/e83ec43a61ce4cf1e4bbac98451a8c8f20ba0982.

szaszm (approved):

PR #1130, opened 2021-07-12
MINIFICPP-1605 Always refresh AWS credentials through default credentials chain

lordgamez:
Due to a time skew problem when using temporary AWS credentials like instance profiles the credentials may expire on AWS side even though it can be non-expired when checked on client side. Because of this when using default credentials chain we should let the AWS SDK handle the caching and renewing the credentials and not cache manually on client side.

Jira ticket: https://issues.apache.org/jira/browse/MINIFICPP-1605

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/test/aws-tests/AWSCredentialsServiceTest.cpp
+++ b/libminifi/test/aws-tests/AWSCredentialsServiceTest.cpp
@@ -45,6 +45,18 @@ class AWSCredentialsServiceTestAccessor {
   std::shared_ptr<core::controller::ControllerServiceNode> aws_credentials_service;
 };
 
+namespace {
+
+void setEnvironmentCredentials(const std::string& key, const std::string& secret_key) {
+  #ifdef WIN32
+  _putenv_s("AWS_ACCESS_KEY_ID", key.c_str());
+  _putenv_s("AWS_SECRET_ACCESS_KEY", secret_key.c_str());
+  #else
+  setenv("AWS_ACCESS_KEY_ID", key.c_str(), 1);
+  setenv("AWS_SECRET_ACCESS_KEY", secret_key.c_str(), 1);
+  #endif
+}
+
 TEST_CASE_METHOD(AWSCredentialsServiceTestAccessor, "Test expired credentials are refreshed", "[credentialRefresh]") {
   plan->setProperty(aws_credentials_service, "Access Key", "key");
   plan->setProperty(aws_credentials_service, "Secret Key", "secret");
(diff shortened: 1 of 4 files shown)


libminifi/test/aws-tests/AWSCredentialsServiceTest.cpp
```diff
@@ -65,3 +77,28 @@ TEST_CASE_METHOD(AWSCredentialsServiceTestAccessor, "Test expired credentials ar
...
+  setEnvironmentCredentials("key2", "secret2");
+
+  // Check for credential refresh
+  REQUIRE(aws_credentials_impl->getAWSCredentials());
+  REQUIRE(aws_credentials_impl->getAWSCredentials()->GetAWSAccessKeyId() == "key2");
+  REQUIRE(aws_credentials_impl->getAWSCredentials()->GetAWSSecretKey() == "secret2");
+  REQUIRE_FALSE(aws_credentials_impl->getAWSCredentials()->IsExpired());
+}
+
+}
```
szaszm, line 104:
```suggestion
}  // namespace
```

szaszm (approved):
LGTM after the linter error is fixed

libminifi/test/aws-tests/AWSCredentialsServiceTest.cpp
lordgamez: Fixed in c1a18bb5c2c60662e095ff43b63b2b0e53bd8635

arpadboda (approved):

adamdebreceni:
as I see in `AWSCredentialsProvider::getAWSCredentials()` the `use_default_credentials_` branch falls through if we cannot acquire credentials from the default provider chain, is that intentional?

adamdebreceni (commented):
as I see in `AWSCredentialsProvider::getAWSCredentials()` the `use_default_credentials_` branch falls through if we cannot acquire credentials from the default provider chain, is that intentional?

lordgamez:
> as I see in `AWSCredentialsProvider::getAWSCredentials()` the `use_default_credentials_` branch falls through if we cannot acquire credentials from the default provider chain, is that intentional?

Yes, in that case we still try to check if we can find any other set credentials like the access key and secret key from properties or the credentials file as a backup.

adamdebreceni:
> > as I see in `AWSCredentialsProvider::getAWSCredentials()` the `use_default_credentials_` branch falls through if we cannot acquire credentials from the default provider chain, is that intentional?
> 
> Yes, in that case we still try to check if we can find any other set credentials like the access key and secret key from properties or the credentials file as a backup.

and should we? I mean if I explicitly request the usage of default credentials shouldn't other user provided options be ignored?

lordgamez:
> > > as I see in `AWSCredentialsProvider::getAWSCredentials()` the `use_default_credentials_` branch falls through if we cannot acquire credentials from the default provider chain, is that intentional?
> > 
> > 
> > Yes, in that case we still try to check if we can find any other set credentials like the access key and secret key from properties or the credentials file as a backup.
> 
> and should we? I mean if I explicitly request the usage of default credentials shouldn't other user provided options be ignored?

I see your point, but if the user explicitly sets the `Access Key` and the default credentials chain does not have any credentials to provide, should that be ignored as well? I think it could be a bigger problem for the user if they forget to turn off the default credential chain option and then after setting the credentials manually, they get an error that no credentials are available. It could take a long time to debug why those credentials were ignored.

I look at the `AWSCredentialsProvider` as more of a priority list, where if the default credential chain is set that should be the priority, but if that is not available then the other options should be checked as well.

adamdebreceni (approved):

PR #1134, opened 2021-07-20
MINIFICPP-1608 upgrade pybind11 to 2.6.2

szaszm:
Due to this error in C++20 mode on AppleClang: https://github.com/szaszm/nifi-minifi-cpp/runs/3058420988#step:6:15174
And it doesn't hurt to upgrade dependencies from time to time anyway.

https://issues.apache.org/jira/browse/MINIFICPP-1608

---

### For all changes:
     in the commit message?

### For code changes:

### Note:


thirdparty/pybind11/include/pybind11/attr.h
```diff
@@ -12,7 +12,7 @@
 
 #include "cast.h"
 
-NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
+PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
```
adamdebreceni, line 15: could we make this dependency into an external project? there are a few dependencies that are directly committed into `thirdparty`, and I feel like we should minimize their number, this upgrade would be a great excuse to do so with pybind11

szaszm: What do you think about FetchContent? I prefer it over ExternalProject, because ExternalProject downloads sources during build while FetchContent does it during configure, which means if you're using an IDE, headers may not be present and include paths may be broken before the first build. It also results in simpler cmake code IMO.

Done in a2ac1f3, but I couldn't reuse pybind11's own cmake files. They look reasonable, but I just kept getting linker errors that I was unable to solve.

cmake/Pybind11.cmake
```diff
@@ -0,0 +1,32 @@
...
+FetchContent_Declare(pybind11_src
+    URL      https://github.com/pybind/pybind11/archive/refs/tags/v2.7.0.tar.gz
+    URL_HASH SHA256=6cd73b3d0bf3daf415b5f9b87ca8817cc2e2b64c275d65f9500250f9fee1677e
+)
+FetchContent_GetProperties(pybind11_src)
+if (NOT pybind11_src_POPULATED)
+    FetchContent_Populate(pybind11_src)
+    set(PYBIND11_INCLUDE_DIR "${pybind11_src_SOURCE_DIR}/include" CACHE STRING "" FORCE)
+    add_library(pybind11 INTERFACE IMPORTED)
+    target_sources(pybind11 INTERFACE ${PYBIND11_INCLUDE_DIR}/pybind11/pybind11.h)
```
adamdebreceni, line 29: why do we need this line?

thirdparty/pybind11/include/pybind11/attr.h
adamdebreceni: looks clean, thank you

cmake/Pybind11.cmake
szaszm: It's not necessary. I don't know what it does aside from throwing an early error when you try to use the target and the file is not there.

szaszm: > In the above example, note that .h header files were specified as sources too, not just the .cpp implementation files. Headers listed as sources don’t get compiled directly on their own, but the effect of adding them is for the benefit of IDE generators like Visual Studio, Xcode, Qt Creator, etc. This causes those headers to be listed in the project’s file list within the IDE, even if no source file refers to it via #include. This can make those headers easier to find during development and potentially aid things like refactoring functionality, etc.

from https://crascit.com/2016/01/31/enhanced-source-file-handling-with-target_sources/

lordgamez (approved):
LGTM

adamdebreceni (approved):

LICENSE
```diff
@@ -470,15 +470,8 @@ CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
...
-You are under no obligation whatsoever to provide any bug fixes, patches, or
-upgrades to the features, functionality or performance of the source code
-("Enhancements") to anyone; however, if you choose to make your Enhancements
-available either publicly, or directly to the author of this software, without
-imposing a separate written license agreement for such Enhancements, then you
-hereby grant the following license: a non-exclusive, royalty-free perpetual
-license to install, use, modify, prepare derivative works, incorporate into
-other computer software, distribute, and sublicense such enhancements or
-derivative works thereof, in binary and source code form.
+Please also refer to the file .github/CONTRIBUTING.md, which clarifies licensing of
```
fgerlits, line 473: we only have a `CONTRIB.md` -- did you mean to move and rename it?

fgerlits (approved):

martinzink (approved):

PR #1135, opened 2021-07-20
MINIFICPP-1609 Handle bash in the same way as patch

fgerlits:
When building Minifi with Azure on Windows, `cmake/BundledAzureSdkCpp.cmake` assumed that `bash` was on the PATH.  My local Git for Windows installation includes `bash.exe`, but it is not on the PATH.  Also, my Windows machine contains a `bash.exe` in `C:\Windows\System32`, which does not work the same way as Git Bash, so that had to be excluded.

https://issues.apache.org/jira/browse/MINIFICPP-1609

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -174,6 +174,9 @@ function(append_third_party_passthrough_args OUTPUT EXTERNALPROJECT_CMAKE_ARGS)
 	set(${OUTPUT} ${EXTERNALPROJECT_CMAKE_ARGS} PARENT_SCOPE)
 endfunction()
 
+# Find bash executable
+find_package(Bash REQUIRED)
+
 # Find patch executable
 find_package(Patch REQUIRED)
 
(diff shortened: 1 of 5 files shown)


szaszm (commented):
I welcome the change if it fixes such an issue, but I wonder why the bash version from your System32 didn't work. These commands are pretty straightforward and I would assume any bash version should be able to execute them.

I was recently thinking about an alternative approach: Make a cmake function to generate those complex and repetitive patch commands. This approach could solve this issue by making it able to generate a command for cmd.exe when used on windows. What do you think about this?

fgerlits:
Yes, a CMake function would be better.  If you have something in mind, then go for it, but I don't know how much time we want to spend on this, as the current approach (is ugly but) works.

lordgamez (approved):

szaszm:
I was not able to escape a cmd.exe command in cmake and have it properly end up in the generated VS solution, so let's keep the current version. `"`s mysteriously disappear or get escaped to `\"`, depending on the position, while `\"` ends up on the command line verbatim, so I gave up.

szaszm (approved):

martinzink (approved):

PR #1136, opened 2021-07-21
MINIFICPP-1610 fix PublishKafka::notifyStop + reduce PublishKafkaOnScheduleTests runtime ~30s -> ~1.5s

szaszm:
https://issues.apache.org/jira/browse/MINIFICPP-1610

With MINIFICPP-1351, I introduced a connection lock before the message lock to avoid a race condition on closing the connection. It went before the message lock, because when both are taken, connection lock has to come first to avoid a deadlock.
This caused notifyStop to wait until the messages in flight timed out and onTrigger stopped before being able to take the connection lock, then the message lock, then interrupt the messages, thus defeating the interruption.

This issue is about fixing the situation by first taking the message lock without the connection lock, interrupting the messages, then releasing the lock and taking the connection lock (and getting it after onTrigger terminated early, thanks to the interruptions) and terminating the connection.

Aside from that, I made an effort to reduce the runtime of PublishKafkaOnScheduleTests when no local kafka nodes are running. It tries to connect to one and waits for timeout, but the test doesn't need an actual running broker, this is just a side effect. Reduced timeouts, the runtime went from ~30sec to ~1.5sec.

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/librdkafka/PublishKafka.cpp
+++ b/extensions/librdkafka/PublishKafka.cpp
@@ -553,11 +553,18 @@ void PublishKafka::onSchedule(const std::shared_ptr<core::ProcessContext> &conte
 void PublishKafka::notifyStop() {
   logger_->log_debug("notifyStop called");
   interrupted_ = true;
-  std::lock_guard<std::mutex> conn_lock(connection_mutex_);
-  std::lock_guard<std::mutex> lock(messages_mutex_);
-  for (auto& messages : messages_set_) {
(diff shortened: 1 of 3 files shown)


lordgamez (approved):
Thanks for the fix and the clear explanation in the comment.

fgerlits (approved):
LGTM

martinzink (approved):

PR #1137, opened 2021-07-21
MINIFICPP-1607 Add AttributesToJson processor

lordgamez:
Implementing `AttributesToJson` processor that generates a JSON representation of the input content attributes and writes them either in the `JSONAttributes` attribute or as the flowfile content.

https://issues.apache.org/jira/browse/MINIFICPP-1607

-----------------------------------------------------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/docker/test/integration/features/attributes_to_json.feature
+++ b/docker/test/integration/features/attributes_to_json.feature
@@ -4,7 +4,7 @@ Feature: Writing attribute data using AttributesToJSON processor
 
   Scenario: Write selected attribute data to file
     Given a GetFile processor with the "Input Directory" property set to "/tmp/input"
-    And a file with filename "test_file.log" and content "test_data%" is present in "/tmp/input"
+    And a file with filename "test_file.log" and content "test_data" is present in "/tmp/input"
     And a AttributesToJSON processor with the "Attributes List" property set to "filename,invalid"
     And the "Destination" property of the AttributesToJSON processor is set to "flowfile-content"
     And the "Null Value" property of the AttributesToJSON processor is set to "true"
--- a/docker/test/integration/features/attributes_to_json.feature
+++ b/docker/test/integration/features/attributes_to_json.feature
@@ -1,0 +1,15 @@
+Feature: Writing attribute data using AttributesToJSON processor
+  Background:
+    Given the content of "/tmp/output" is monitored
+
+  Scenario: Write selected attribute data to file
+    Given a GetFile processor with the "Input Directory" property set to "/tmp/input"
+    And a file with filename "test_file.log" and content "test_data%" is present in "/tmp/input"
+    And a AttributesToJSON processor with the "Attributes List" property set to "filename,invalid"
+    And the "Destination" property of the AttributesToJSON processor is set to "flowfile-content"
+    And the "Null Value" property of the AttributesToJSON processor is set to "true"
+    And a PutFile processor with the "Directory" property set to "/tmp/output"
+    And the "success" relationship of the GetFile processor is connected to the AttributesToJSON
+    And the "success" relationship of the AttributesToJSON processor is connected to the PutFile
+    When the MiNiFi instance starts up
+    Then a flowfile with the content "{"filename":"test_file.log","invalid":null}" is placed in the monitored directory in less than 10 seconds
--- a/docker/test/integration/features/attributes_to_json.feature
+++ b/docker/test/integration/features/attributes_to_json.feature
@@ -12,4 +12,4 @@ Feature: Writing attribute data using AttributesToJSON processor
     And the "success" relationship of the GetFile processor is connected to the AttributesToJSON
     And the "success" relationship of the AttributesToJSON processor is connected to the PutFile
     When the MiNiFi instance starts up
-    Then a flowfile with the content "{"filename":"test_file.log","invalid":null}" is placed in the monitored directory in less than 10 seconds
+    Then a flowfile with the JSON content "{"filename":"test_file.log","invalid":null}" is placed in the monitored directory in less than 10 seconds
--- a/extensions/standard-processors/processors/AttributesToJSON.cpp
+++ b/extensions/standard-processors/processors/AttributesToJSON.cpp
@@ -86,7 +86,7 @@ void AttributesToJSON::onSchedule(core::ProcessContext* context, core::ProcessSe
   attribute_list_ = utils::StringUtils::split(attributes, ",");
   context->getProperty(AttributesRegularExpression.getName(), attributes_regular_expression_str_);
   if (!attributes_regular_expression_str_.empty()) {
-    attributes_regular_expression_ = std::regex(attributes_regular_expression_str_);
+    attributes_regular_expression_ = utils::Regex(attributes_regular_expression_str_);
   }
   context->getProperty(Destination.getName(), destination_);
   context->getProperty(IncludeCoreAttributes.getName(), include_core_attributes_);
@@ -97,11 +97,11 @@ bool AttributesToJSON::isCoreAttributeToBeFiltered(const std::string& attribute)
   return !include_core_attributes_ && CORE_ATTRIBUTES.find(attribute) != CORE_ATTRIBUTES.end();
 }
 
-bool AttributesToJSON::matchesAttributeRegex(const std::string& attribute) const {
-  return attributes_regular_expression_str_.empty() || std::regex_search(attribute, attributes_regular_expression_);
+bool AttributesToJSON::matchesAttributeRegex(const std::string& attribute) {
+  return attributes_regular_expression_str_.empty() || attributes_regular_expression_.match(attribute);
 }
 
-void AttributesToJSON::addAttributeToJson(rapidjson::Document& document, const std::string& key, const std::string& value) const {
+void AttributesToJSON::addAttributeToJson(rapidjson::Document& document, const std::string& key, const std::string& value) {
   if (isCoreAttributeToBeFiltered(key)) {
     logger_->log_debug("Core attribute '%s' will not be included in the attributes JSON.", key);
     return;
@@ -118,7 +118,7 @@ void AttributesToJSON::addAttributeToJson(rapidjson::Document& document, const s
   document.AddMember(json_key, json_val, document.GetAllocator());
 }
 
-std::string AttributesToJSON::buildAttributeJsonData(std::map<std::string, std::string>&& attributes) const {
+std::string AttributesToJSON::buildAttributeJsonData(std::map<std::string, std::string>&& attributes) {
   auto root = rapidjson::Document(rapidjson::kObjectType);
   if (!attribute_list_.empty()) {
     for (const auto& attribute : attribute_list_) {
--- a/extensions/standard-processors/processors/AttributesToJSON.cpp
+++ b/extensions/standard-processors/processors/AttributesToJSON.cpp
@@ -108,7 +108,7 @@ std::unordered_set<std::string> AttributesToJSON::getAttributesToBeWritten(const
 
   if (attributes_regular_expression_) {
     for (const auto& [key, value] : flowfile_attributes) {
-      if (!isCoreAttributeToBeFiltered(key) && std::regex_search(key, attributes_regular_expression_.value())) {
+      if (!isCoreAttributeToBeFiltered(key) && std::regex_match(key, attributes_regular_expression_.value())) {
         attributes.insert(key);
       }
     }
--- a/extensions/standard-processors/processors/AttributesToJSON.cpp
+++ b/extensions/standard-processors/processors/AttributesToJSON.cpp
@@ -67,8 +67,7 @@ const core::Property AttributesToJSON::NullValue(
     ->withDefaultValue<bool>(false)
     ->build());
 
-core::Relationship AttributesToJSON::Success("success", "Successfully converted attributes to JSON");
-core::Relationship AttributesToJSON::Failure("failure", "Failed to convert attributes to JSON");
+core::Relationship AttributesToJSON::Success("success", "All FlowFiles received are routed to success");
 
 void AttributesToJSON::initialize() {
   setSupportedProperties({
@@ -78,7 +77,7 @@ void AttributesToJSON::initialize() {
     IncludeCoreAttributes,
     NullValue
   });
-  setSupportedRelationships({Success, Failure});
+  setSupportedRelationships({Success});
 }
 
 void AttributesToJSON::onSchedule(core::ProcessContext* context, core::ProcessSessionFactory* /*sessionFactory*/) {
@@ -89,7 +88,7 @@ void AttributesToJSON::onSchedule(core::ProcessContext* context, core::ProcessSe
   if (!attributes_regular_expression_str_.empty()) {
     attributes_regular_expression_ = utils::Regex(attributes_regular_expression_str_);
   }
-  destination_ = utils::parsePropertyWithAllowableValuesOrThrow(*context, Destination.getName(), DESTINATIONS);
+  write_to_attribute_ = utils::parsePropertyWithAllowableValuesOrThrow(*context, Destination.getName(), DESTINATIONS) == "flowfile-attribute";
   context->getProperty(IncludeCoreAttributes.getName(), include_core_attributes_);
   context->getProperty(NullValue.getName(), null_value_);
 }
@@ -144,23 +143,15 @@ void AttributesToJSON::onTrigger(core::ProcessContext* /*context*/, core::Proces
   }
 
   auto json_data = buildAttributeJsonData(flow_file->getAttributes());
-  if (destination_ == "flowfile-attribute") {
+  if (write_to_attribute_) {
     logger_->log_debug("Writing the following attribute data to JSONAttributes attribute: %s", json_data);
     session->putAttribute(flow_file, "JSONAttributes", json_data);
     session->transfer(flow_file, Success);
-  } else if (destination_ == "flowfile-content") {
+  } else {
     logger_->log_debug("Writing the following attribute data to flowfile: %s", json_data);
     AttributesToJSON::WriteCallback callback(json_data);
-    try {
-      session->write(flow_file, &callback);
-      session->transfer(flow_file, Success);
-    } catch(const std::exception&) {
-      logger_->log_error("Failed to write attributes to flow file!");
-      session->transfer(flow_file, Failure);
-    }
-  } else {
-    logger_->log_error("Unimplemented destination was set in AttributesToJSON's Destination property");
-    session->transfer(flow_file, Failure);
+    session->write(flow_file, &callback);
+    session->transfer(flow_file, Success);
   }
 }
 
--- a/extensions/standard-processors/processors/AttributesToJSON.cpp
+++ b/extensions/standard-processors/processors/AttributesToJSON.cpp
@@ -24,6 +24,7 @@
 #include "rapidjson/writer.h"
 #include "utils/StringUtils.h"
 #include "utils/ProcessorConfigUtils.h"
+#include "range/v3/algorithm/find.hpp"
 
 namespace org {
 namespace apache {
@@ -67,7 +68,7 @@ const core::Property AttributesToJSON::NullValue(
     ->withDefaultValue<bool>(false)
     ->build());
 
-core::Relationship AttributesToJSON::Success("success", "All FlowFiles received are routed to success");
+const core::Relationship AttributesToJSON::Success("success", "All FlowFiles received are routed to success");
 
 void AttributesToJSON::initialize() {
   setSupportedProperties({
@@ -94,10 +95,11 @@ void AttributesToJSON::onSchedule(core::ProcessContext* context, core::ProcessSe
 }
 
 bool AttributesToJSON::isCoreAttributeToBeFiltered(const std::string& attribute) const {
-  return !include_core_attributes_ && core_attributes_.find(attribute) != core_attributes_.end();
+  const auto& special_attributes = core::SpecialFlowAttribute::getSpecialFlowAttributes();
+  return !include_core_attributes_ && ranges::find(special_attributes, attribute) != ranges::end(special_attributes);
 }
 
-std::unordered_set<std::string> AttributesToJSON::getAttributesToBeWritten(const std::map<std::string, std::string>& flowfile_attributes) const {
+std::unordered_set<std::string> AttributesToJSON::getAttributesToBeWritten(core::FlowFile::AttributeMap* flowfile_attributes) const {
   std::unordered_set<std::string> attributes;
 
   for (const auto& attribute : attribute_list_) {
@@ -107,7 +109,7 @@ std::unordered_set<std::string> AttributesToJSON::getAttributesToBeWritten(const
   }
 
   if (attributes_regular_expression_) {
-    for (const auto& [key, value] : flowfile_attributes) {
+    for (const auto& [key, value] : *flowfile_attributes) {
       if (!isCoreAttributeToBeFiltered(key) && std::regex_match(key, attributes_regular_expression_.value())) {
         attributes.insert(key);
       }
@@ -126,16 +128,17 @@ void AttributesToJSON::addAttributeToJson(rapidjson::Document& document, const s
   document.AddMember(json_key, json_val, document.GetAllocator());
 }
 
-std::string AttributesToJSON::buildAttributeJsonData(std::map<std::string, std::string>&& flowfile_attributes) {
+std::string AttributesToJSON::buildAttributeJsonData(core::FlowFile::AttributeMap* flowfile_attributes) {
   auto root = rapidjson::Document(rapidjson::kObjectType);
 
   if (!attribute_list_.empty() || attributes_regular_expression_) {
     auto attributes_to_write = getAttributesToBeWritten(flowfile_attributes);
     for (const auto& key : attributes_to_write) {
-      addAttributeToJson(root, key, flowfile_attributes[key]);
+      auto it = flowfile_attributes->find(key);
+      addAttributeToJson(root, key, it == flowfile_attributes->end() ? "" : it->second);
     }
   } else {
-    for (const auto& [key, value] : flowfile_attributes) {
+    for (const auto& [key, value] : *flowfile_attributes) {
       if (!isCoreAttributeToBeFiltered(key)) {
         addAttributeToJson(root, key, value);
       }
@@ -154,7 +157,7 @@ void AttributesToJSON::onTrigger(core::ProcessContext* /*context*/, core::Proces
     return;
   }
 
-  auto json_data = buildAttributeJsonData(flow_file->getAttributes());
+  auto json_data = buildAttributeJsonData(flow_file->getAttributesPtr());
   if (write_destination_ == WriteDestination::FLOWFILE_ATTRIBUTE) {
     logger_->log_debug("Writing the following attribute data to JSONAttributes attribute: %s", json_data);
     session->putAttribute(flow_file, "JSONAttributes", json_data);
--- a/extensions/standard-processors/processors/AttributesToJSON.cpp
+++ b/extensions/standard-processors/processors/AttributesToJSON.cpp
@@ -103,14 +103,12 @@ std::unordered_set<std::string> AttributesToJSON::getAttributesToBeWritten(core:
   std::unordered_set<std::string> attributes;
 
   for (const auto& attribute : attribute_list_) {
-    if (!isCoreAttributeToBeFiltered(attribute)) {
-      attributes.insert(attribute);
-    }
+    attributes.insert(attribute);
   }
 
   if (attributes_regular_expression_) {
     for (const auto& [key, value] : *flowfile_attributes) {
-      if (!isCoreAttributeToBeFiltered(key) && std::regex_match(key, attributes_regular_expression_.value())) {
+      if (std::regex_match(key, attributes_regular_expression_.value())) {
         attributes.insert(key);
       }
     }
--- a/extensions/standard-processors/processors/AttributesToJSON.cpp
+++ b/extensions/standard-processors/processors/AttributesToJSON.cpp
@@ -29,8 +29,6 @@ namespace nifi {
 namespace minifi {
 namespace processors {
 
-const std::set<std::string> AttributesToJSON::DESTINATIONS({"flowfile-attribute", "flowfile-content"});
-
 const core::Property AttributesToJSON::AttributesList(
   core::PropertyBuilder::createProperty("Attributes List")
     ->withDescription("Comma separated list of attributes to be included in the resulting JSON. "
@@ -49,8 +47,8 @@ const core::Property AttributesToJSON::Destination(
     ->withDescription("Control if JSON value is written as a new flowfile attribute 'JSONAttributes' or written in the flowfile content. "
                       "Writing to flowfile content will overwrite any existing flowfile content.")
     ->isRequired(true)
-    ->withDefaultValue<std::string>("flowfile-attribute")
-    ->withAllowableValues<std::string>(DESTINATIONS)
+    ->withDefaultValue<std::string>(toString(WriteDestination::FLOWFILE_ATTRIBUTE))
+    ->withAllowableValues<std::string>(WriteDestination::values())
     ->build());
 
 const core::Property AttributesToJSON::IncludeCoreAttributes(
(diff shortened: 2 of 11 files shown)


extensions/standard-processors/processors/AttributesToJSON.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+void AttributesToJSON::initialize() {
+  setSupportedProperties({
+    AttributesList,
+    AttributesRegularExpression,
+    Destination,
+    IncludeCoreAttributes,
+    NullValue
+  });
+  setSupportedRelationships({Success});
```
arpadboda, line 80: This processor has a failure relationship in NiFi.
Are you sure there is no such case where we can't provide proper json representation?

extensions/standard-processors/processors/AttributesToJSON.cpp
```diff
@@ -0,0 +1,162 @@
...
+}
+
+void AttributesToJSON::onSchedule(core::ProcessContext* context, core::ProcessSessionFactory* /*sessionFactory*/) {
+  std::string attributes;
+  context->getProperty(AttributesList.getName(), attributes);
+  attribute_list_ = utils::StringUtils::split(attributes, ",");
+  context->getProperty(AttributesRegularExpression.getName(), attributes_regular_expression_str_);
+  if (!attributes_regular_expression_str_.empty()) {
+    attributes_regular_expression_ = utils::Regex(attributes_regular_expression_str_);
```
arpadboda, line 89: If think we can already rely on std::regex, even if the c++20 PR is not yet merged. @szaszm can you confirm?

Moreover std::regex can throw in case of invalid regex. Which is fair as a throw in onschedule is going to cause a retry later, but I guess the error message is not going to be too talkative.

extensions/standard-processors/processors/AttributesToJSON.cpp
```diff
@@ -0,0 +1,162 @@
...
+  if (!matchesAttributeRegex(key)) {
+    logger_->log_debug("Attribute '%s' does not match the set regex, therefore it will not be included in the attributes JSON.", key);
+    return;
+  }
+  rapidjson::Value json_key(key.c_str(), document.GetAllocator());
+  rapidjson::Value json_val;
+  if (!value.empty() || !null_value_) {
+    json_val.SetString(value.c_str(), document.GetAllocator());
+  }
+  document.AddMember(json_key, json_val, document.GetAllocator());
```
arpadboda, line 118: What happens in case there is a spec char in the value (such as "{" or ":")? 
Do we expect RapidJson library to escape or shall we (if we can)?

arpadboda (commented):
Mostly looks good, added minor questions

extensions/standard-processors/processors/AttributesToJSON.cpp
szaszm: Yes, std::regex is part of the language since C++11, and the reason we couldn't use it before was supporting GCC 4.8, which has only partial support for C++11. Now we have dropped support for anything older than GCC 8, so we are good with regex.

extensions/standard-processors/processors/AttributesToJSON.cpp
lordgamez: I actually had a failure relationship, but it seemed to be forced and unnecessary in hindsight so I removed it in 74e8946b280c840c7c00e739f92bc5271ab69b11. NiFi has only one case where it transfers flowfile to failure and it's due to the used thirdparty JSON library's possible exception when writing the JSON object, but we do not have that here.

extensions/standard-processors/processors/AttributesToJSON.cpp
lordgamez: RapidJson escapes these characters by default, there is also a test case for this:
```
TEST_CASE_METHOD(AttributesToJSONTestFixture, "Move selected attributes with special characters to a flowfile attribute", "[AttributesToJSONTests]") {
```

extensions/standard-processors/processors/AttributesToJSON.cpp
lordgamez: Yes, when I opened this PR the GCC 4.8 support was still there, but I'll update this now.

lordgamez: Updated in 78d69d19c56f1eaba7718b0253acdeba672f46f4

extensions/standard-processors/processors/AttributesToJSON.cpp
arpadboda: Makes sense, thanks.

arpadboda (approved):

extensions/standard-processors/processors/AttributesToJSON.cpp
```diff
@@ -0,0 +1,162 @@
...
+  auto root = rapidjson::Document(rapidjson::kObjectType);
+  if (!attribute_list_.empty()) {
+    for (const auto& attribute : attribute_list_) {
+      addAttributeToJson(root, attribute, attributes[attribute]);
+    }
+  } else {
+    for (const auto& kvp : attributes) {
+      addAttributeToJson(root, kvp.first, kvp.second);
+    }
+  }
```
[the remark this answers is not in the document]
lordgamez: That makes sense, I think I misunderstood the meaning of the description of the regex property: "This property can be used in combination with the attributes list property.". I checked the NiFi code and the result is the union of both the regex matching attributes and the selected attribute list. Thanks for pointing it out, I will have a fix for this.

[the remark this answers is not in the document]
fgerlits: Yeah, I agree the description is not clear.  It may be worth adding a clarifying sentence to it, even if NiFi doesn't have one.

extensions/standard-processors/tests/unit/AttributesToJSONTests.cpp
```diff
@@ -0,0 +1,194 @@
...
+    writer.String(json.c_str(), json.size());
+    return buffer.GetString();
+  }
+
+  std::vector<std::string> getOutputFileContents() {
+    std::vector<std::string> file_contents;
+
+    auto callback = [&file_contents](const std::string& path, const std::string& filename) -> bool {
+      std::ifstream is(path + utils::file::FileUtils::get_separator() + filename, std::ifstream::binary);
+      std::string file_content((std::istreambuf_iterator<char>(is)), std::istreambuf_iterator<char>());
```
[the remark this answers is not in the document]
lordgamez: It's needed because of a vexing-parse warning where the `std::istreambuf_iterator<char>(is)` could be interpreted as a parameter declaration. Here is an explanation of this: https://stackoverflow.com/questions/7007817/a-confusing-detail-about-the-most-vexing-parse

libminifi/include/core/FlowFile.h
```diff
@@ -310,6 +311,10 @@ struct SpecialFlowAttribute {
   static const std::string ALTERNATE_IDENTIFIER;
   // Flow identifier
   static const std::string FLOW_ID;
+
+  static std::unordered_set<std::string> getSpecialFlowAttributes() {
+    return {PATH, ABSOLUTE_PATH, FILENAME, UUID, priority, MIME_TYPE, DISCARD_REASON, ALTERNATE_IDENTIFIER, FLOW_ID};
+  }
```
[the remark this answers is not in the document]
lordgamez: I'm not sure it's a good idea as there are currently 69 occurrences of the `SpecialFlowAttribute` in the codebase and it would take a lot of changes. My other problem with it is that the main use case of the `SpecialFlowAttribute`s is using them as strings while getting or setting attributes so the enums would always have to be converted to strings which would make their usage harder.

extensions/standard-processors/tests/unit/AttributesToJSONTests.cpp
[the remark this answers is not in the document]
fgerlits: OK, thanks.  I would probably use `{}` instead of the extra `()`:
```c++
       std::string file_content(std::istreambuf_iterator<char>{is}, std::istreambuf_iterator<char>{});
```
but it's fine as it is, too.

libminifi/include/core/FlowFile.h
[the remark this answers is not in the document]
fgerlits: OK, makes sense.  I thought it was already an enum, just not a smart one, but you're right, it's not.

extensions/standard-processors/processors/AttributesToJSON.h
```diff
@@ -0,0 +1,103 @@
...
+class AttributesToJSON : public core::Processor {
+ public:
+  static const std::set<std::string> DESTINATIONS;
+
+  explicit AttributesToJSON(const std::string& name, const utils::Identifier& uuid = {})
+      : core::Processor(name, uuid),
+        logger_(logging::LoggerFactory<AttributesToJSON>::getLogger()),
+        core_attributes_(core::SpecialFlowAttribute::getSpecialFlowAttributes()) {
+  }
+  static constexpr char const* ProcessorName = "AttributesToJSON";
```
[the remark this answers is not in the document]
lordgamez: Updated in 4e8f82bc2e1f7997a7d51dab9ff413fc0a1d0e13

extensions/standard-processors/processors/AttributesToJSON.h
```diff
@@ -0,0 +1,103 @@
...
+  bool matchesAttributeRegex(const std::string& attribute);
+  void addAttributeToJson(rapidjson::Document& document, const std::string& key, const std::string& value);
+  std::string buildAttributeJsonData(std::map<std::string, std::string>&& attributes);
+
+  std::shared_ptr<logging::Logger> logger_;
+  const std::unordered_set<std::string> core_attributes_;
+  std::vector<std::string> attribute_list_;
+  std::string attributes_regular_expression_str_;
+  std::regex attributes_regular_expression_;
+  bool write_to_attribute_ = true;
```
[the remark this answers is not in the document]
lordgamez: Updated in 4e8f82bc2e1f7997a7d51dab9ff413fc0a1d0e13

extensions/standard-processors/processors/AttributesToJSON.h
```diff
@@ -0,0 +1,103 @@
...
+  bool isCoreAttributeToBeFiltered(const std::string& attribute) const;
+  bool matchesAttributeRegex(const std::string& attribute);
+  void addAttributeToJson(rapidjson::Document& document, const std::string& key, const std::string& value);
+  std::string buildAttributeJsonData(std::map<std::string, std::string>&& attributes);
+
+  std::shared_ptr<logging::Logger> logger_;
+  const std::unordered_set<std::string> core_attributes_;
+  std::vector<std::string> attribute_list_;
+  std::string attributes_regular_expression_str_;
+  std::regex attributes_regular_expression_;
```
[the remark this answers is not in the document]
lordgamez: Updated in 4e8f82bc2e1f7997a7d51dab9ff413fc0a1d0e13

extensions/standard-processors/processors/AttributesToJSON.cpp
```diff
@@ -0,0 +1,162 @@
...
+}
+
+std::string AttributesToJSON::buildAttributeJsonData(std::map<std::string, std::string>&& attributes) {
+  auto root = rapidjson::Document(rapidjson::kObjectType);
+  if (!attribute_list_.empty()) {
+    for (const auto& attribute : attribute_list_) {
+      addAttributeToJson(root, attribute, attributes[attribute]);
+    }
+  } else {
+    for (const auto& kvp : attributes) {
```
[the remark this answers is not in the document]
lordgamez: Updated in 4e8f82bc2e1f7997a7d51dab9ff413fc0a1d0e13

extensions/standard-processors/processors/AttributesToJSON.cpp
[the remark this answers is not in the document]
lordgamez: Fixed the funtionality and the description in 1b513c74851d9f35ae42efd2d5788d04d7087536. Added additional tests and refactored the tests to ignore the order of the json parameters.

extensions/standard-processors/tests/unit/AttributesToJSONTests.cpp
```diff
@@ -0,0 +1,194 @@
...
+    update_attribute_->setDynamicProperty("my_attribute", "my_value");
+    update_attribute_->setDynamicProperty("other_attribute", "other_value");
+    update_attribute_->setDynamicProperty("empty_attribute", "");
+
+    std::fstream file;
+    std::stringstream ss;
+    ss << dir_ << utils::file::FileUtils::get_separator() << TEST_FILE_NAME;
+    file.open(ss.str(), std::ios::out);
+    file << TEST_FILE_CONTENT;
+    file.close();
```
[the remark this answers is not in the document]
lordgamez: Updated in 4e8f82bc2e1f7997a7d51dab9ff413fc0a1d0e13

extensions/standard-processors/tests/unit/AttributesToJSONTests.cpp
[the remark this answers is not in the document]
lordgamez: Thanks it looks better this way, updated in 4e8f82bc2e1f7997a7d51dab9ff413fc0a1d0e13

extensions/standard-processors/tests/unit/AttributesToJSONTests.cpp
```diff
@@ -0,0 +1,194 @@
...
+  plan_->setProperty(attribute_to_json_, org::apache::nifi::minifi::processors::AttributesToJSON::NullValue.getName(), "true");
+  test_controller_.runSession(plan_);
+  auto file_contents = getOutputFileContents();
+
+  REQUIRE(file_contents.size() == 1);
+  REQUIRE(file_contents[0].size() == TEST_FILE_CONTENT.size());
+  REQUIRE(LogTestController::getInstance().contains("key:JSONAttributes value:{\"my_attribute\":\"my_value\",\"non_existent_attribute\":null,\"empty_attribute\":null}"));
+}
+
+TEST_CASE_METHOD(AttributesToJSONTestFixture, "All non-existent or empty attributes shall be written as null in JSON", "[AttributesToJSONTests]") {
```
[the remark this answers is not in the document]
lordgamez: Updated in 4e8f82bc2e1f7997a7d51dab9ff413fc0a1d0e13

extensions/standard-processors/tests/unit/AttributesToJSONTests.cpp
```diff
@@ -0,0 +1,194 @@
...
+  std::shared_ptr<core::Processor> logattribute_;
+  std::shared_ptr<core::Processor> putfile_;
+};
+
+TEST_CASE_METHOD(AttributesToJSONTestFixture, "Move all attributes to a flowfile attribute", "[AttributesToJSONTests]") {
+  test_controller_.runSession(plan_);
+  auto file_contents = getOutputFileContents();
+
+  REQUIRE(file_contents.size() == 1);
+  REQUIRE(file_contents[0].size() == TEST_FILE_CONTENT.size());
```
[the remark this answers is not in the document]
lordgamez: Updated in 4e8f82bc2e1f7997a7d51dab9ff413fc0a1d0e13

extensions/standard-processors/tests/unit/AttributesToJSONTests.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+TEST_CASE_METHOD(AttributesToJSONTestFixture, "JSON attributes are written in flowfile", "[AttributesToJSONTests]") {
+  plan_->setProperty(attribute_to_json_, org::apache::nifi::minifi::processors::AttributesToJSON::Destination.getName(), "flowfile-content");
+  test_controller_.runSession(plan_);
+  std::string expected_content = "{\"absolute.path\":" + escapeJson(dir_ + utils::file::FileUtils::get_separator() + TEST_FILE_NAME) + ",\"empty_attribute\":\"\",\"filename\":" + escapeJson(TEST_FILE_NAME) + ",\"flow.id\":\"test\",\"my_attribute\":\"my_value\",\"other_attribute\":\"other_value\",\"path\":" + escapeJson(dir_ + utils::file::FileUtils::get_separator()) + "}";  // NOLINT
+
+  auto file_contents = getOutputFileContents();
+
+  REQUIRE(file_contents.size() == 1);
+  REQUIRE(file_contents[0].size() == expected_content.size());
```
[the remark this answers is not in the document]
lordgamez: Updated in 4e8f82bc2e1f7997a7d51dab9ff413fc0a1d0e13

extensions/standard-processors/processors/AttributesToJSON.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+  for (const auto& attribute : attribute_list_) {
+    if (!isCoreAttributeToBeFiltered(attribute)) {
+      attributes.insert(attribute);
+    }
+  }
+
+  if (attributes_regular_expression_) {
+    for (const auto& [key, value] : flowfile_attributes) {
+      if (!isCoreAttributeToBeFiltered(key) && std::regex_search(key, attributes_regular_expression_.value())) {
```
fgerlits, line 111: I think NiFi matches the regex against the whole attribute name, not only part of if.  I would change `regex_search` to `regex_match` so we do the same.

extensions/standard-processors/tests/unit/AttributesToJSONTests.cpp
```diff
@@ -0,0 +1,293 @@
...
+    update_attribute_->setDynamicProperty("my_attribute", "my_value");
+    update_attribute_->setDynamicProperty("other_attribute", "other_value");
+    update_attribute_->setDynamicProperty("empty_attribute", "");
+
+    utils::putFileToDir(dir_, TEST_FILE_NAME, TEST_FILE_CONTENT);
+  }
+
+  void assertJSONAttributesFromLog(const std::unordered_map<std::string, std::optional<std::string>>& expected_attributes) {
+    auto match = LogTestController::getInstance().matchesRegex("key:JSONAttributes value:(.*)").value();
+    assertAttributes(expected_attributes, match[1].str());
```
fgerlits, line 67: How clear is the exception thrown when the optional is null?  I think an explicit REQUIRE might give a better diagnostic:
```suggestion
    auto match = LogTestController::getInstance().matchesRegex("key:JSONAttributes value:(.*)");
    REQUIRE(match);
    assertAttributes(expected_attributes, (*match)[1].str());
```

libminifi/test/TestBase.h
```diff
@@ -164,6 +165,31 @@ class LogTestController {
...
+    auto start = std::chrono::system_clock::now();
+    bool found = false;
+    bool timed_out = false;
+    std::regex matcher_regex(regex_str);
+    std::smatch match;
+    do {
+      std::string str = log_output.str();
+      found = std::regex_search(str, match, matcher_regex);
+      auto now = std::chrono::system_clock::now();
+      timed_out = std::chrono::duration_cast<std::chrono::milliseconds>(now - start) > std::chrono::duration_cast<std::chrono::milliseconds>(timeout);
```
fgerlits, line 183: I don't think these `duration_cast`s are needed; `std::chrono::duration`s constructed with different units can be compared.
```suggestion
      timed_out = (now - start > timeout);
```

docker/test/integration/features/attributes_to_json.feature
```diff
@@ -0,0 +1,15 @@
+Feature: Writing attribute data using AttributesToJSON processor
+  Background:
+    Given the content of "/tmp/output" is monitored
+
+  Scenario: Write selected attribute data to file
+    Given a GetFile processor with the "Input Directory" property set to "/tmp/input"
+    And a file with filename "test_file.log" and content "test_data%" is present in "/tmp/input"
```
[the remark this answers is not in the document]
lordgamez: I think this line was just copied from a test where `%` was testing special characters, but it is not relevant here. Removed in dc9f6aa438e986a8ae6d1bacb85ca4fc677cb175

extensions/standard-processors/processors/AttributesToJSON.cpp
lordgamez: You are right, updated the code and the tests in dc9f6aa438e986a8ae6d1bacb85ca4fc677cb175

extensions/standard-processors/tests/unit/AttributesToJSONTests.cpp
lordgamez: Updated in dc9f6aa438e986a8ae6d1bacb85ca4fc677cb175

libminifi/test/TestBase.h
lordgamez: Updated in dc9f6aa438e986a8ae6d1bacb85ca4fc677cb175

fgerlits (approved):

libminifi/include/core/FlowFile.h
```diff
@@ -310,6 +311,10 @@ struct SpecialFlowAttribute {
   static const std::string ALTERNATE_IDENTIFIER;
   // Flow identifier
   static const std::string FLOW_ID;
+
+  static std::unordered_set<std::string> getSpecialFlowAttributes() {
+    return {PATH, ABSOLUTE_PATH, FILENAME, UUID, priority, MIME_TYPE, DISCARD_REASON, ALTERNATE_IDENTIFIER, FLOW_ID};
```
szaszm, line 316: We could use a static array of `const char*` or `std::string_view`, referring to the static strings from above. I want to avoid the overhead of hash tables is possible, but preferably the allocations and the need to store a copy in each processor instance as well.
The static object below is just a flat array of a set of pointers to static storage and sizes, about 144 bytes in size.
```suggestion
  static const auto& getSpecialFlowAttributes() {
    static const std::array<std::string_view, 9> list{
      PATH, ABSOLUTE_PATH, FILENAME, UUID, priority, MIME_TYPE, DISCARD_REASON, ALTERNATE_IDENTIFIER, FLOW_ID
    };
    return list;
```

extensions/standard-processors/processors/AttributesToJSON.h
```diff
@@ -0,0 +1,106 @@
...
+ public:
+  // Supported Properties
+  static const core::Property AttributesList;
+  static const core::Property AttributesRegularExpression;
+  static const core::Property Destination;
+  static const core::Property IncludeCoreAttributes;
+  static const core::Property NullValue;
+
+  // Supported Relationships
+  static core::Relationship Success;
```
szaszm, line 52: This could (and IMHO should) be `const`.

martinzink (approved):

extensions/standard-processors/processors/AttributesToJSON.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+  for (const auto& attribute : attribute_list_) {
+    if (!isCoreAttributeToBeFiltered(attribute)) {
+      attributes.insert(attribute);
+    }
+  }
+
+  if (attributes_regular_expression_) {
+    for (const auto& [key, value] : flowfile_attributes) {
+      if (!isCoreAttributeToBeFiltered(key) && std::regex_match(key, attributes_regular_expression_.value())) {
```
adamdebreceni, line 111: as I see, in the nifi implementation the regex based match does not check if it is a core attribute or not

extensions/standard-processors/processors/AttributesToJSON.cpp
```diff
@@ -0,0 +1,174 @@
...
+  return buffer.GetString();
+}
+
+void AttributesToJSON::onTrigger(core::ProcessContext* /*context*/, core::ProcessSession* session) {
+  auto flow_file = session->get();
+  if (!flow_file) {
+    return;
+  }
+
+  auto json_data = buildAttributeJsonData(flow_file->getAttributes());
```
adamdebreceni, line 157: as the ``attributes_` in the `FlowFile` is of type `FlatMap`, `getAttributes` constructs a new `std::map` for backwards compatibility, we should avoid this if we can IMO

extensions/standard-processors/processors/AttributesToJSON.cpp
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  return attributes;
+}
+
+void AttributesToJSON::addAttributeToJson(rapidjson::Document& document, const std::string& key, const std::string& value) {
+  rapidjson::Value json_key(key.c_str(), document.GetAllocator());
+  rapidjson::Value json_val;
+  if (!value.empty() || !null_value_) {
+    json_val.SetString(value.c_str(), document.GetAllocator());
```
adamdebreceni, line 124: the nifi implementation (despite what the documentation says) writes `null` only if the attribute does not exist and null value is requested, it does not seem to write `null` when the value of the attribute is an empty string

extensions/standard-processors/processors/AttributesToJSON.cpp
```diff
@@ -0,0 +1,174 @@
...
+  }
+  document.AddMember(json_key, json_val, document.GetAllocator());
+}
+
+std::string AttributesToJSON::buildAttributeJsonData(std::map<std::string, std::string>&& flowfile_attributes) {
+  auto root = rapidjson::Document(rapidjson::kObjectType);
+
+  if (!attribute_list_.empty() || attributes_regular_expression_) {
+    auto attributes_to_write = getAttributesToBeWritten(flowfile_attributes);
```
adamdebreceni, line 133: we could return an optional set from `getAttributesToBeWritten` and do the "else branch" when that is `nullopt`, this way we wouldn't need to recheck the condition that makes the result "valid" outside of that method

extensions/standard-processors/processors/AttributesToJSON.h
```diff
@@ -0,0 +1,106 @@
...
+  std::shared_ptr<logging::Logger> logger_;
+  const std::unordered_set<std::string> core_attributes_;
+  std::vector<std::string> attribute_list_;
+  std::optional<std::regex> attributes_regular_expression_;
+  WriteDestination write_destination_;
+  bool include_core_attributes_ = true;
+  bool null_value_ = false;
+};
+
+REGISTER_RESOURCE(AttributesToJSON, "Generates a JSON representation of the input FlowFile Attributes. "
```
adamdebreceni, line 99: note: after the dynamic extension PR is merged these should migrate to the source files, no need to move it now, but remind me in case I forget to do so on rebase :)

libminifi/test/TestBase.h
```diff
@@ -164,6 +165,31 @@ class LogTestController {
     return found;
   }
 
+  std::optional<std::smatch> matchesRegex(const std::string &regex_str,
```
adamdebreceni, line 168: as this is a single use function, it would be better IMO if we moved it into the `AttributesToJSONTests.cpp` if possible, thus other test wouldn't have to parse `<regex>` (or move the impl to the `TestBase.cpp` and return a `vector<string>`, but that would sacrifice some semantic power in the name of build-time)

extensions/standard-processors/processors/AttributesToJSON.cpp
lordgamez: That's true, but should we follow that functionality? It's not really intuitive and seems more like a bug, that when we set the `Include Core Attributes` to false it does not work the same way if the attribute matches the `Attributes List` or the `Attributes Regular Expression` property. IMO if the `Include Core Attributes` is set, then those attributes should not be included.

extensions/standard-processors/processors/AttributesToJSON.cpp
lordgamez: It's again seems to me like a bug on the NiFi side, either in the documentation or in the code. Which one would make more sense?

extensions/standard-processors/processors/AttributesToJSON.cpp
lordgamez: Could you elaborate a bit more please? I'm not sure what do you mean by rechecking the result of `getAttributesToBeWritten`. When would it be invalid?

libminifi/test/TestBase.h
lordgamez: The `matchesRegex` depends on the `log_output` member of the `LogTestController` to retrieve and match the log output, and there is no way currently to get the `log_output` outside of the class. We could check how it influences the build times of other tests, but if it's not that big impact, I would keep it as a utility for other use cases as well, and would not expose the `log_output`.

libminifi/include/core/FlowFile.h
lordgamez: Good idea, updated in e2893bb98eac22b2e9199c5354b33d143494edd5

extensions/standard-processors/processors/AttributesToJSON.h
lordgamez: Fixed in e2893bb98eac22b2e9199c5354b33d143494edd5

extensions/standard-processors/processors/AttributesToJSON.cpp
```diff
@@ -94,10 +95,11 @@ void AttributesToJSON::onSchedule(core::ProcessContext* context, core::ProcessSe
 }
 
 bool AttributesToJSON::isCoreAttributeToBeFiltered(const std::string& attribute) const {
-  return !include_core_attributes_ && core_attributes_.find(attribute) != core_attributes_.end();
+  const auto& special_attributes = core::SpecialFlowAttribute::getSpecialFlowAttributes();
+  return !include_core_attributes_ && ranges::find(special_attributes, attribute) != ranges::end(special_attributes);
 }
 
-std::unordered_set<std::string> AttributesToJSON::getAttributesToBeWritten(const std::map<std::string, std::string>& flowfile_attributes) const {
+std::unordered_set<std::string> AttributesToJSON::getAttributesToBeWritten(core::FlowFile::AttributeMap* flowfile_attributes) const {
```
szaszm, line 102: Having a pointer parameter that we're not checking for null before dereferencing is a recipe for disaster. Can we change it to const ref? We don't seem to modify the AttributeMap.

extensions/standard-processors/processors/AttributesToJSON.cpp
adamdebreceni: I feel like excluding the core attributes only makes sense when we specify neither the list nor the regex, if the user explicitly requests the attribute by name or regex, I think we should not exclude core ones

extensions/standard-processors/processors/AttributesToJSON.cpp
adamdebreceni: I would support writing `null` only for non-existing attributes, attributes with an empty string value could(should?) have different semantics from non-existing attributes

extensions/standard-processors/processors/AttributesToJSON.cpp
lordgamez: True, I checked with the NiFi team and they said that the `Include Core Attributes` should only control whether the attributes are included by default, the regex or attribute list should override this.

extensions/standard-processors/processors/AttributesToJSON.cpp
adamdebreceni: currently when `attribute_list_.empty() && !attributes_regular_expression_` the return value of `getAttributesToBeWritten` (which then is an empty set) is considered "invalid" in the sense that we fallback on enumerating all the attributes on the flowfile, but of course an empty set could also mean that e.g. the regex did not match any of the attributes, so we currently cannot decide on the return value of the `getAttributesToBeWritten` if we should fallback or not

if we returned an optional from `getAttributesToBeWritten` the return value alone could signal the need for fallback

extensions/standard-processors/processors/AttributesToJSON.cpp
lordgamez: I asked the NiFi team about this and they said the same, the empty string values should not be modified. I'll update according to this.

libminifi/test/TestBase.h
adamdebreceni: measured an unrelated test to see if this method makes any difference, fortunately the difference is negligible (~50ms) so I think we can keep it as is

lordgamez: Thanks for the effort :)

extensions/standard-processors/processors/AttributesToJSON.cpp
lordgamez: Good point, we can safely change it as the pointer is just the returned address of the internal member. Updated in effc866d914fb3ccff4f70a7a0ebcea6305403fc

extensions/standard-processors/processors/AttributesToJSON.cpp
lordgamez: Thanks for the explanation it is clear now, updated in effc866d914fb3ccff4f70a7a0ebcea6305403fc

extensions/standard-processors/processors/AttributesToJSON.cpp
lordgamez: Updated in effc866d914fb3ccff4f70a7a0ebcea6305403fc

extensions/standard-processors/processors/AttributesToJSON.cpp
lordgamez: Updated in 16ebbb4cc0bf6256200ee79c755133dec8ae09e2. Thanks for pointing out these discrepancies!

adamdebreceni (approved):

szaszm (approved):

PR #1121, opened 2021-06-29
MINIFICPP-1596 Refactor docker tests by removing unnecessary clusters

lordgamez:
Issues addressed in this PR regarding integration tests:

- The main problem was the misuse of the `Cluster` concept in the codebase causing several issues
- `Cluster` and `Container` concepts were mixed up in the codebase which made understanding the test code harder
- In several test scenarios multiple clusters were created with the same network, which made the cleanup cumbersome and fail transiently
- Some container deployments required creating new named clusters that were unnecessary and made the test scenario descriptions more complex
- The `SingleNodeDockerCluster` class had way too many responsibilities handling attributes corresponding to specific containers
- There were some redundancies in the test step definitions
- Custom images are built separately for every test case

Changes made to resolve the issues:

- The `SingleNodeDockerCluster` (single node referring to the intent of running the containers on a single compute node) is now defined as a set of `Containers` with a single Docker network connecting them
- All test scenarios now have a single cluster which makes the cleanup much simpler and no multiple named clusters in the test steps are needed
- Introducing the container concept in the code with the `Container` class from which all used containers are inherited from defining their own attributes, deployment functions and so on. This moves container specific functionality from the `SingleNodeDockerCluster` and simplifies the deployment of the cluster
- Removing the redundancies from the step descriptions which simplifies the test scenarios
- Adding ImageStore to cache built custom images and reuse them in additional testcases, additionally NiFi and MiNiFi configuration files are attached as volumes instead of building them in the image
- Fixing some timing issues to lower runtimes

Jira ticket: https://issues.apache.org/jira/browse/MINIFICPP-1596
This should also fix the following Jira regarding transient failures in cleanup: https://issues.apache.org/jira/browse/MINIFICPP-1516

--------------------------------------------------------------------------------
Thank you for submitting a contribution to Apache NiFi - MiNiFi C++.

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [ ] Is there a JIRA ticket associated with this PR? Is it referenced
     in the commit message?

- [ ] Does your PR title start with MINIFICPP-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.

- [ ] Has your PR been rebased against the latest commit within the target branch (typically main)?

- [ ] Is your initial contribution a single, squashed commit?

### For code changes:
[...]

PR #1141, opened 2021-07-26
MINIFICPP-1479-Rework integration tests for HashContent

aminadinari19:
https://issues.apache.org/jira/browse/MINIFICPP-1479

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/docker/test/integration/MiNiFi_integration_test_driver.py
+++ b/docker/test/integration/MiNiFi_integration_test_driver.py
@@ -272,3 +272,11 @@ class MiNiFi_integration_test():
     def check_query_results(self, cluster_name, query, number_of_rows, timeout_seconds):
         cluster = self.acquire_cluster(cluster_name)
         assert cluster.check_query_results(query, number_of_rows, timeout_seconds)
+
+    def check_minifi_log_contents(self, line):
+        line_found = False
+        for cluster in self.clusters.values():
+            if cluster.get_engine() == "minifi-cpp":
+                line_found = cluster.wait_for_app_logs(line, 60)
+        assert line_found
+
--- a/docker/test/integration/MiNiFi_integration_test_driver.py
+++ b/docker/test/integration/MiNiFi_integration_test_driver.py
@@ -279,4 +279,3 @@ class MiNiFi_integration_test():
             if cluster.get_engine() == "minifi-cpp":
                 line_found = cluster.wait_for_app_logs(line, 60)
         assert line_found
-
--- a/docker/test/integration/features/hashcontent.feature
+++ b/docker/test/integration/features/hashcontent.feature
@@ -1,0 +1,40 @@
+Feature: Hash value is added to Flowfiles by HashContent processor
+  In order to avoid duplication of content of Flowfiles
+  As a user of MiNiFi
+  I need to have HashContent processor to calculate and add hash value
+
+  Background:
+    Given the content of "/tmp/output" is monitored
+
+  Scenario Outline: HashContent adds hash attribute to flowfiles
+    Given a GetFile processor with the "Input Directory" property set to "/tmp/input"
+    And a file with the content <content> is present in "/tmp/input"
+    And a HashContent processor with the "Hash Attribute" property set to "hash"
+    And the "Hash Algorithm" property of the HashContent processor is set to "<hash_algorithm>"
+    And a LogAttribute processor
+    And the "success" relationship of the GetFile processor is connected to the HashContent
+    And the "success" relationship of the HashContent processor is connected to the LogAttribute
+    When the MiNiFi instance starts up
+    Then the minifi log contains "key:hash value:<hash_value>"
+
+    Examples:
+      | content  | hash_algorithm | hash_value                                                       |
+      | "test"   | MD5            | 098F6BCD4621D373CADE4E832627B4F6                                 |
+      | "test"   | SHA1           | A94A8FE5CCB19BA61C4C0873D391E987982FBBD3                         |
+      | "test"   | SHA256         | 9F86D081884C7D659A2FEAA0C55AD015A3BF4F1B2B0B822CD15D6C15B0F00A08 |
+      | "coffee" | MD5            | 24EB05D18318AC2DB8B2B959315D10F2                                 |
+      | "coffee" | SHA1           | 44213F9F4D59B557314FADCD233232EEBCAC8012                         |
+      | "coffee" | SHA256         | 37290D74AC4D186E3A8E5785D259D2EC04FAC91AE28092E7620EC8BC99E830AA |
+
+
+  Scenario: HashContent fails for an empty file if 'fail on empty' property is set to true
+    Given a GetFile processor with the "Input Directory" property set to "/tmp/input"
+    And an empty file is present in "/tmp/input"
+    And a HashContent processor with the "Hash Attribute" property set to "hash"
+    And the "Hash Algorithm" property of the HashContent processor is set to "MD5"
+    And the "Fail on empty" property of the HashContent processor is set to "true"
+    And a PutFile processor with the "Directory" property set to "/tmp/output"
+    And the "success" relationship of the GetFile processor is connected to the HashContent
+    And the "failure" relationship of the HashContent processor is connected to the PutFile
+    When the MiNiFi instance starts up
+    Then at least one empty flowfile is placed in the monitored directory in less than 10 seconds
--- a/docker/test/integration/features/hashcontent.feature
+++ b/docker/test/integration/features/hashcontent.feature
@@ -19,11 +19,8 @@ Feature: Hash value is added to Flowfiles by HashContent processor
 
     Examples:
       | content  | hash_algorithm | hash_value                                                       |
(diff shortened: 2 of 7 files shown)


docker/test/integration/features/hashcontent.feature
```diff
@@ -0,0 +1,40 @@
...
+    Given a GetFile processor with the "Input Directory" property set to "/tmp/input"
+    And an empty file is present in "/tmp/input"
+    And a HashContent processor with the "Hash Attribute" property set to "hash"
+    And the "Hash Algorithm" property of the HashContent processor is set to "MD5"
+    And the "Fail on empty" property of the HashContent processor is set to "true"
+    And a PutFile processor with the "Directory" property set to "/tmp/output"
+    And the "success" relationship of the GetFile processor is connected to the HashContent
+    And the "failure" relationship of the HashContent processor is connected to the PutFile
+    When the MiNiFi instance starts up
+    Then at least one empty flowfile is placed in the monitored directory in less than 10 seconds
```
szaszm, line 40: I think testing this in either c++ or behave is enough, we don't need to test the same thing twice. If the test is instantaneous (say <0.1 sec) in c++, then I would keep it there, otherwise keep the behave version.

aminadinari19: Ahan okay. Is it better to keep one scenario where (fail on empty = true) in behave and other one in unit test cause here I don't really check the other case?

szaszm: Yes, that sounds like a good idea.

docker/test/integration/MiNiFi_integration_test_driver.py
```diff
@@ -266,3 +266,10 @@ def check_azure_storage_server_data(self, cluster_name, object_data):
     def wait_for_kafka_consumer_to_be_registered(self, cluster_name):
         cluster = self.acquire_cluster(cluster_name)
         assert cluster.wait_for_kafka_consumer_to_be_registered()
+
+    def check_minifi_log_contents(self, line):
+        line_found = False
+        for cluster in self.clusters.values():
+            if cluster.get_engine() == "minifi-cpp":
+                line_found = cluster.wait_for_app_logs(line, 60)
+        assert line_found
```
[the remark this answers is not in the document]
lordgamez: Currently it would be better to have the cluster_name, although there may be still multiple minifi containers in a single cluster. After #1121 is merged it will be much simpler as there will only be a single cluster and you can specify the container you want to get the logs from.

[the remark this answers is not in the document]
fgerlits: I think it's OK to leave it like this for now, as it will need to be modified when resolving the conflict with #1121, whichever PR goes in first.

extensions/standard-processors/tests/unit/HashContentTest.cpp
```diff
@@ -149,17 +149,31 @@ TEST_CASE("TestingFailOnEmptyProperty", "[HashContentPropertiesCheck]") {
                                                                      core::Relationship("success", "description"), true);
   plan->setProperty(md5processor, org::apache::nifi::minifi::processors::HashContent::HashAttribute.getName(), MD5_ATTR);
   plan->setProperty(md5processor, org::apache::nifi::minifi::processors::HashContent::HashAlgorithm.getName(), "MD5");
-  plan->setProperty(md5processor, org::apache::nifi::minifi::processors::HashContent::FailOnEmpty.getName(), "true");
+  std::set<core::Relationship> relationships;
+  relationships.insert(core::Relationship("success", "description"));
+  relationships.insert(core::Relationship("failure", "description"));
+  md5processor->setAutoTerminatedRelationships(relationships);
```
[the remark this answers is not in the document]
aminadinari19: The failure relationship had to be auto terminated or else the test would end abruptly if the flowfile was routed to failure. I also think the initializer is nicer. Changing it now :)

docker/test/integration/features/hashcontent.feature
fgerlits: I wouldn't mind keeping both, as they test the same thing in slightly different ways, eg. the missing return statement bug was only caught by the integration test.  The unit test takes 0.01 second to run, including all test cases.

szaszm: Not sure why that's the case, but if they are testing different things, then I agree that keeping them both is ok.

extensions/standard-processors/tests/unit/HashContentTest.cpp
```diff
@@ -149,17 +149,31 @@ TEST_CASE("TestingFailOnEmptyProperty", "[HashContentPropertiesCheck]") {
...
 
+  SECTION("with an empty file and fail on empty property set to false") {
+  plan->setProperty(md5processor, org::apache::nifi::minifi::processors::HashContent::FailOnEmpty.getName(), "false");
+
   plan->runNextProcessor();
   plan->runNextProcessor();
 
-  REQUIRE(LogTestController::getInstance().contains("Failure as flow file is empty"));
-}
+  REQUIRE(LogTestController::getInstance().contains("attempting read"));
```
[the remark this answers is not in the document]
szaszm: Now that this is reverted, the indentation is wrong again.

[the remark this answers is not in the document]
aminadinari19: Thanks for noticing. Fixed it finally. :)

szaszm (approved):

fgerlits (approved):

docker/test/integration/features/hashcontent.feature
```diff
@@ -0,0 +1,40 @@
...
+    Then the minifi log contains "key:hash value:<hash_value>"
+
+    Examples:
+      | content  | hash_algorithm | hash_value                                                       |
+      | "test"   | MD5            | 098F6BCD4621D373CADE4E832627B4F6                                 |
+      | "test"   | SHA1           | A94A8FE5CCB19BA61C4C0873D391E987982FBBD3                         |
+      | "test"   | SHA256         | 9F86D081884C7D659A2FEAA0C55AD015A3BF4F1B2B0B822CD15D6C15B0F00A08 |
+      | "coffee" | MD5            | 24EB05D18318AC2DB8B2B959315D10F2                                 |
+      | "coffee" | SHA1           | 44213F9F4D59B557314FADCD233232EEBCAC8012                         |
+      | "coffee" | SHA256         | 37290D74AC4D186E3A8E5785D259D2EC04FAC91AE28092E7620EC8BC99E830AA |
```
lordgamez, line 27: I'm not sure we need to check with both `test` and `coffee` content with each hash algorithm, I don't think it makes a difference, but it doubles the execution time. Maybe we could have different content for each algorithm just to prove it works with different file contents so 3 examples would be enough.

aminadinari19: I agree. Thanks for pointing it out. Changed it now :)

extensions/standard-processors/tests/unit/HashContentTest.cpp
```diff
@@ -147,17 +147,30 @@ TEST_CASE("TestingFailOnEmptyProperty", "[HashContentPropertiesCheck]") {
                                                                      core::Relationship("success", "description"), true);
   plan->setProperty(md5processor, org::apache::nifi::minifi::processors::HashContent::HashAttribute.getName(), MD5_ATTR);
   plan->setProperty(md5processor, org::apache::nifi::minifi::processors::HashContent::HashAlgorithm.getName(), "MD5");
-  plan->setProperty(md5processor, org::apache::nifi::minifi::processors::HashContent::FailOnEmpty.getName(), "true");
+
+  using processors::HashContent;
```
adamdebreceni, line 151: we could move this `using` to the beginning of the test case and get rid of all the fully qualified `HashContent` references in this test

adamdebreceni (approved):

docker/test/integration/MiNiFi_integration_test_driver.py
```diff
@@ -272,3 +272,11 @@ def wait_for_kafka_consumer_to_be_registered(self, cluster_name):
...
         cluster = self.acquire_cluster(cluster_name)
         assert cluster.check_query_results(query, number_of_rows, timeout_seconds)
+
+    def check_minifi_log_contents(self, line):
+        line_found = False
+        for cluster in self.clusters.values():
+            if cluster.get_engine() == "minifi-cpp":
+                line_found = cluster.wait_for_app_logs(line, 60)
+        assert line_found
+
```
adamdebreceni, line 282: one of the builds fail as this contains an empty line at the end of file

szaszm (changes requested):
I'd prefer this not to be merged until at least the linter issue is fixed.

docker/test/integration/MiNiFi_integration_test_driver.py
aminadinari19: Thank-you for pointing it out. Fixed it :)

szaszm (approved):

lordgamez (approved):

PR #1142, opened 2021-07-26
MINIFICPP-1425 Proceed to partial C++20 support

szaszm:
squashed:
commit baa3f502d8fdc6a395983295c8f4187c088765d5
Author: Marton Szasz <szaszm@apache.org>
Date:   Sun Jul 25 14:07:11 2021 +0200

    patch sol2 to not fail in C++20 mode

    on mac xcode/AppleClang, due to const char*/const char8_t*
    incompatibility and the fact that u8"string" literals are now
    const char8*.

    error: invalid conversion from 'const char8_t*' to 'const char*'
    (and friends)

commit 8d6476be877d14e64db25bbdc71adc30b5fc9dd6
Author: Marton Szasz <szaszm@apache.org>
Date:   Wed Jul 21 07:00:26 2021 +0200

    back to C++20

---

### For all changes:
     in the commit message?

### For code changes:

### Note:


thirdparty/sol2-2.20.0/sol.hpp
```diff
@@ -1719,7 +1719,8 @@ namespace sol {
 #else
 			is_specialization_of<meta::unqualified_t<T>, basic_string_view>,
 #endif
```
adamdebreceni, line 1721: as with pybind11, could we fetch the sources configuration time?

szaszm: I'd rather not change it now, in the context of this PR. Maybe on the next upgrade?

adamdebreceni: I am OK with doing it in another PR, I am also OK with not making the change if this is a rarely touched dependency (it seems to be), although I feel that having a separate patch file helps with upgrades/tracking changes

szaszm: I agree that it would make the changes more visible, although I tried my best by adding comments where I patched the source.
Changing this to ExternalProject would be less than ideal, because the header would not be available before the first build.
Because of this I would go for FetchContent, but due to its different syntax, I would most likely need to manually invoke the patch command at the right times instead of letting cmake take care of it. This is somewhat unknown territory and I would need to do more research, that's why I prefer not to do it now.

adamdebreceni: makes sense

lordgamez (approved):

fgerlits (approved):

martinzink (approved):

adamdebreceni (approved):

PR #1143, opened 2021-07-26
MINIFICPP-1522 Log property name while parsing config in case of property validation failure

lordgamez:
Previously when parsing a single property value failed due to the failure of a custom validator there was only a generic error message:
```
[2021-03-02 16:35:37.377] [org::apache::nifi::minifi::core::YamlConfiguration] [error] Invalid yaml configuration file
[2021-03-02 16:35:37.377] [main] [error] Failed to load configuration due to exception: General Operation: Cannot convert invalid value
```
The message did not indicate which property had an invalid value, which made it hard for the user to find the source of the failure. With this change the property name is logged in case of validator failure to pinpoint the source of the problem:
```
[2021-07-26 16:53:31.190] [org::apache::nifi::minifi::core::YamlConfiguration] [error] Invalid value was set for property 'Max Poll Time'
[2021-07-26 16:53:31.190] [org::apache::nifi::minifi::core::YamlConfiguration] [error] Invalid yaml configuration file
[2021-07-26 16:53:31.190] [main] [error] Failed to load configuration due to exception: General Operation: Cannot convert invalid value
```

https://issues.apache.org/jira/browse/MINIFICPP-1522

--------------------------------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/standard-processors/tests/unit/YamlConfigurationTests.cpp
+++ b/extensions/standard-processors/tests/unit/YamlConfigurationTests.cpp
@@ -18,6 +18,7 @@
 
 #include <map>
 #include <memory>
+#include <chrono>
 #include "core/repository/VolatileContentRepository.h"
 #include "core/ProcessGroup.h"
 #include "core/RepositoryFactory.h"
@@ -25,6 +26,8 @@
 #include "TailFile.h"
 #include "TestBase.h"
 #include "utils/TestUtils.h"
+
+using namespace std::chrono_literals;  // NOLINT using namespace directive is required for literals
 
 TEST_CASE("Test YAML Config Processing", "[YamlConfiguration]") {
   TestController test_controller;
@@ -149,11 +152,9 @@ Provenance Reporting:
     REQUIRE(uuid);
     REQUIRE(!rootFlowConfig->findProcessorByName("TailFile")->getUUIDStr().empty());
     REQUIRE(1 == rootFlowConfig->findProcessorByName("TailFile")->getMaxConcurrentTasks());
-    REQUIRE(
-        core::SchedulingStrategy::TIMER_DRIVEN == rootFlowConfig->findProcessorByName("TailFile")->getSchedulingStrategy());
-    REQUIRE(1 == rootFlowConfig->findProcessorByName("TailFile")->getMaxConcurrentTasks());
+    REQUIRE(core::SchedulingStrategy::TIMER_DRIVEN == rootFlowConfig->findProcessorByName("TailFile")->getSchedulingStrategy());
     REQUIRE(1 * 1000 * 1000 * 1000 == rootFlowConfig->findProcessorByName("TailFile")->getSchedulingPeriodNano());
-    REQUIRE(std::chrono::seconds(30) == rootFlowConfig->findProcessorByName("TailFile")->getPenalizationPeriod());
+    REQUIRE(30s == rootFlowConfig->findProcessorByName("TailFile")->getPenalizationPeriod());
     REQUIRE(1 * 1000 == rootFlowConfig->findProcessorByName("TailFile")->getYieldPeriodMsec());
     REQUIRE(0 == rootFlowConfig->findProcessorByName("TailFile")->getRunDurationNano());
 
@@ -478,7 +479,7 @@ NiFi Properties Overrides: {}
   REQUIRE(core::SchedulingStrategy::TIMER_DRIVEN == rootFlowConfig->findProcessorByName("TailFile")->getSchedulingStrategy());
   REQUIRE(1 == rootFlowConfig->findProcessorByName("TailFile")->getMaxConcurrentTasks());
   REQUIRE(1 * 1000 * 1000 * 1000 == rootFlowConfig->findProcessorByName("TailFile")->getSchedulingPeriodNano());
-  REQUIRE(std::chrono::seconds(30) == rootFlowConfig->findProcessorByName("TailFile")->getPenalizationPeriod());
+  REQUIRE(30s == rootFlowConfig->findProcessorByName("TailFile")->getPenalizationPeriod());
   REQUIRE(1 * 1000 == rootFlowConfig->findProcessorByName("TailFile")->getYieldPeriodMsec());
   REQUIRE(0 == rootFlowConfig->findProcessorByName("TailFile")->getRunDurationNano());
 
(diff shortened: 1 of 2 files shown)


extensions/standard-processors/tests/unit/YamlConfigurationTests.cpp
```diff
@@ -37,160 +37,186 @@ TEST_CASE("Test YAML Config Processing", "[YamlConfiguration]") {
...
+
+    REQUIRE(rootFlowConfig);
+    REQUIRE(rootFlowConfig->findProcessorByName("TailFile"));
+    utils::Identifier uuid = rootFlowConfig->findProcessorByName("TailFile")->getUUID();
+    REQUIRE(uuid);
+    REQUIRE(!rootFlowConfig->findProcessorByName("TailFile")->getUUIDStr().empty());
+    REQUIRE(1 == rootFlowConfig->findProcessorByName("TailFile")->getMaxConcurrentTasks());
+    REQUIRE(
+        core::SchedulingStrategy::TIMER_DRIVEN == rootFlowConfig->findProcessorByName("TailFile")->getSchedulingStrategy());
+    REQUIRE(1 == rootFlowConfig->findProcessorByName("TailFile")->getMaxConcurrentTasks());
```
szaszm, line 154: Duplicate condition

extensions/standard-processors/tests/unit/YamlConfigurationTests.cpp
```diff
@@ -37,160 +37,186 @@ TEST_CASE("Test YAML Config Processing", "[YamlConfiguration]") {
...
+    REQUIRE(1 * 1000 * 1000 * 1000 == rootFlowConfig->findProcessorByName("TailFile")->getSchedulingPeriodNano());
+    REQUIRE(std::chrono::seconds(30) == rootFlowConfig->findProcessorByName("TailFile")->getPenalizationPeriod());
```
szaszm, line 156: Consider using standard literal suffixes. I also recognize that this is old code just reformatted, so I'm 100% fine with not touching these parts.

extensions/standard-processors/tests/unit/YamlConfigurationTests.cpp
```diff
@@ -37,160 +37,186 @@ TEST_CASE("Test YAML Config Processing", "[YamlConfiguration]") {
...
+    Properties:
+        Retry Attribute: ""
+Connections: []
+Remote Processing Groups: []
+Provenance Reporting:
+      )";
+
+    std::istringstream configYamlStream(CONFIG_YAML_EMPTY_RETRY_ATTRIBUTE);
+    REQUIRE_THROWS_AS(yamlConfig.getYamlRoot(configYamlStream), utils::internal::InvalidValueException);
+    REQUIRE(LogTestController::getInstance().contains("Invalid value was set for property 'Retry Attribute'"));
```
szaszm, line 218: Can we log the value itself?

extensions/standard-processors/tests/unit/YamlConfigurationTests.cpp
lordgamez: Updated in 37f146fc203f51adc62044c8c9e2fc11f358c63b

extensions/standard-processors/tests/unit/YamlConfigurationTests.cpp
lordgamez: Updated in 37f146fc203f51adc62044c8c9e2fc11f358c63b

extensions/standard-processors/tests/unit/YamlConfigurationTests.cpp
lordgamez: It would require to cast the `PropertyValue` type to a `string` which also runs the custom validator and it would also throw an exception in this case.

szaszm: Ouch, this sounds like tech debt.

szaszm (approved):

arpadboda (approved):
The change itself looks good to me, but I think the commit message is misleading: this change is about processor property values, not attributes. :)

lordgamez:
> The change itself looks good to me, but I think the commit message is misleading: this change is about processor property values, not attributes. :)

That's true, renamed it.

extensions/standard-processors/tests/unit/YamlConfigurationTests.cpp
adamdebreceni: but don't we have access to the raw string value at the point of log?
```
const std::string rawValueString = propertyValueNode.as<std::string>();
```

libminifi/src/core/yaml/YamlConfiguration.cpp
```diff
@@ -764,8 +764,15 @@ void YamlConfiguration::parseSingleProperty(const std::string& propertyName, con
   core::Property myProp(propertyName, "", "");
   processor->getProperty(propertyName, myProp);
   const PropertyValue coercedValue = getValidatedProcessorPropertyForDefaultTypeInfo(myProp, propertyValueNode);
```
adamdebreceni, line 766: what about this `getValidatedProcessorPropertyForDefaultTypeInfo` function? it seems to log errors on exception, in which case does it not log, when it should?

extensions/standard-processors/tests/unit/YamlConfigurationTests.cpp
lordgamez: You are right, we do have access to that, first I thought that it's the same conversion, but actually this is done through the `YAML::Node` class without running the validator and the other one is done through the `PropertyValue` class. 

The only problem I see with this is that the original problem was a validation failure due to defining a time period with quotation marks which caused a validation error: `Max Poll Time: "6 sec"`
If we log the value it looks like this: `[error] Invalid value '6 sec' was set for property 'Max Poll Time'` which could be misleading as it does not show what the problem with this value is. I'm not sure if we should log that value due to this.

libminifi/src/core/yaml/YamlConfiguration.cpp
lordgamez: That function does the cast through `yaml-cpp` lib's `YAML::Node` class and it does not run our custom validators for those properties, those are done through the `PropertyValue` class.

adamdebreceni: oh now I see, so even though `PropertyValue` contains a `validator_`, assignment in itself does not execute it, it only happens in `setValue` (which is called from `processor->setProperty`)

extensions/standard-processors/tests/unit/YamlConfigurationTests.cpp
adamdebreceni: uh tough one, should we maybe log the processor we encountered the error then? or can the user infer which processor the error originated from? (maybe we log something like `creating processor XY`?)

lordgamez: I like the idea of logging the processor name that would pinpoint the source as there could be duplicate property names in different processors (or controller services). Updated in a601b3e4cab33790af5c241157cf6d2ec25523ec

adamdebreceni (approved):

PR #1146, opened 2021-07-29
MINIFICPP-1617 Update CONTRIB.md with current practice

szaszm:
- divergence from the Google Style Guide
- recommendation to follow the C++ Core Guidelines

This is largely based on my observation of existing practice to help new contributors. If you don't agree with any of the changes, would like to add or remove something, please leave a comment to discuss the issue.

Preview the rendered output [here](https://github.com/szaszm/nifi-minifi-cpp/blob/MINIFICPP-1617/CONTRIB.md).

I want to gather as many opinions on the issues as possible and aim for only adding any rules when there is consensus supporting them. The benefit of each rule should outweigh the potential loss of turning away contributors who don't agree with it. I think we were relatively liberal before and that's the right way to go.

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/CONTRIB.md
+++ b/CONTRIB.md
@@ -19,7 +19,8 @@ We welcome all contributions to Apache MiNiFi. All new files must include a copy
 To make development easier, we've included the linter for the Google Style guide. Google provides an Eclipse formatter
 for their style guide. It is located
 [here](https://github.com/google/styleguide/blob/gh-pages/eclipse-cpp-google-style.xml).
-New contributions are expected to follow the [Google C++ Style Guide][https://google.github.io/styleguide/cppguide.html], except for the following points:
+New contributions are expected to follow the 
+[Google C++ Style Guide][https://google.github.io/styleguide/cppguide.html], except for the following points:
 - Use .cpp extension for implementation files
 - Use lowerCamelCase for functions, including accessors/mutators
 - Use UPPER_SNAKE_CASE for constants
@@ -53,6 +54,15 @@ it doesn't contradict any of the Google Style Guide or the above exceptions.
 
 C++ is a powerful language and "with great power comes great responsibility". Please aim for simple, readable and
 maintainable solutions and avoid footguns. Be open for respectful debate in pull request reviews.
+
+Shell script files shall follow the guidelines and best practices defined by the [shellcheck](https://github.com/koalaman/shellcheck) analysis tool.
+New contributions are expected to pass the shellcheck analysis as part of the verification process.
+If a shellcheck requested change is unfeasable it shall be disabled on per-line basis and will be subjected to review.
+For more information on an issue please check the [shellcheck wiki page](https://github.com/koalaman/shellcheck/wiki).
+You can run shellcheck by invoking the shellcheck cmake target, e.g. `make shellcheck`.
+
+Python script files shall follow the PEP8 guidelines and best practices. The project includes [flake8](https://flake8.pycqa.org/en/latest/) checks
+as part of the verification process, that is applied to all new contributions.
 
 ## Issues and Pull Requests
 
--- a/CONTRIB.md
+++ b/CONTRIB.md
@@ -31,7 +31,7 @@ New contributions are expected to follow the Google Style Guide, except for the 
 - Using-directives (`using namespace foo`) are discouraged, except for user-defined literal namespaces
 - Some patterns in the codebase rely on objects with static storage duration without being trivially destructible and
   initialized with a constant expression. It's OK to use these.
-- Operator overloading and user-defined literal suffixes are OK
+- User-defined literal suffixes are OK
 - Public mutable data members are allowed
 - Inline function definition is OK
 - Rvalue references, exceptions and RTTI are allowed
--- a/CONTRIB.md
+++ b/CONTRIB.md
@@ -19,7 +19,7 @@ We welcome all contributions to Apache MiNiFi. All new files must include a copy
 To make development easier, we've included the linter for the Google Style guide. Google provides an Eclipse formatter
 for their style guide. It is located
 [here](https://github.com/google/styleguide/blob/gh-pages/eclipse-cpp-google-style.xml).
-New contributions are expected to follow the Google Style Guide, except for the following points:
+New contributions are expected to follow the [Google C++ Style Guide][https://google.github.io/styleguide/cppguide.html], except for the following points:
 - Use .cpp extension for implementation files
 - Use lowerCamelCase for functions, including accessors/mutators
 - Use UPPER_SNAKE_CASE for constants
--- a/CONTRIB.md
+++ b/CONTRIB.md
@@ -20,7 +20,7 @@ To make development easier, we've included the linter for the Google Style guide
 for their style guide. It is located
 [here](https://github.com/google/styleguide/blob/gh-pages/eclipse-cpp-google-style.xml).
 New contributions are expected to follow the 
-[Google C++ Style Guide][https://google.github.io/styleguide/cppguide.html], except for the following points:
+[Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html), except for the following points:
 - Use .cpp extension for implementation files
 - Use lowerCamelCase for functions, including accessors/mutators
 - Use UPPER_SNAKE_CASE for constants
--- a/CONTRIB.md
+++ b/CONTRIB.md
@@ -25,7 +25,7 @@ New contributions are expected to follow the Google Style Guide, except for the 
 - Use UPPER_SNAKE_CASE for constants
 - Filenames are typically class names or a description of the contents in UpperCamelCase
 - If a class is imitating something from STL, boost or similar, then STL-style lower_snake_case is used for naming the
-  class. UpperSnakeCase is used for most classes, in line with the Google Style Guide.
+  class. UpperCamelCase is used for most classes, in line with the Google Style Guide.
 - Prefer `#pragma once` over include guards
 - Forward declarations are OK
 - Using-directives (`using namespace foo`) are discouraged, except for user-defined literal namespaces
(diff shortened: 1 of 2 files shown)


CONTRIB.md
```diff
@@ -15,24 +15,59 @@
...
+To make development easier, we've included the linter for the Google Style guide. Google provides an Eclipse formatter
+for their style guide. It is located
+[here](https://github.com/google/styleguide/blob/gh-pages/eclipse-cpp-google-style.xml).
+New contributions are expected to follow the Google Style Guide, except for the following points:
+- Use .cpp extension for implementation files
+- Use lowerCamelCase for functions, including accessors/mutators
+- Use UPPER_SNAKE_CASE for constants
+- Filenames are typically class names or a description of the contents in UpperCamelCase
+- If a class is imitating something from STL, boost or similar, then STL-style lower_snake_case is used for naming the
+  class. UpperSnakeCase is used for most classes, in line with the Google Style Guide.
```
fgerlits, line 28: typo:
```suggestion
  class. UpperCamelCase is used for most classes, in line with the Google Style Guide.
```

CONTRIB.md
```diff
@@ -15,24 +15,59 @@
...
+- Prefer `#pragma once` over include guards
+- Forward declarations are OK
+- Using-directives (`using namespace foo`) are discouraged, except for user-defined literal namespaces
+- Some patterns in the codebase rely on objects with static storage duration without being trivially destructible and
+  initialized with a constant expression. It's OK to use these.
+- Operator overloading and user-defined literal suffixes are OK
```
fgerlits, line 34: Suggested addition:
```suggestion
- Operator overloading and user-defined literal suffixes are OK, but use them sparingly and only where they make the code more readable
```

CONTRIB.md
```diff
@@ -15,24 +15,59 @@
...
+- Public mutable data members are allowed
+- Inline function definition is OK
+- Rvalue references, exceptions and RTTI are allowed
+- Use gsl::narrow and gsl::narrow_cast in addition to standard casts. The codebase doesn't use abseil.
+- We are more liberal regarding the use of `auto`. The Google Style Guide only allows using it when it makes the code
+  clearer. In MiNiFi C++, it's up to the personal preferences of the contributor.
+- Template metaprogramming is OK as long as the usage is clear.
```
fgerlits, line 41: Suggested addition:
```suggestion
- Template metaprogramming is OK where it is necessary, as long as the usage is clear.  Prefer `if constexpr` to `std::enable_if`.
```

CONTRIB.md
```diff
@@ -15,24 +15,59 @@
...
+- Static data members are capitalized according to lower_snake_case, in contrast to the Google Style Guide, which uses
```
fgerlits, line 42: I think our current usage of static (non-constant) data members is mixed, both with trailing `_` and without.  I would prefer having the trailing `_` in new code, as I would rather distinguish static members from local variables than from non-static members.

CONTRIB.md
szaszm: I don't want to discourage their usage. Using operator-overloading even for syntactic sugar is fine by my standards, it makes great interfaces like [Clara](https://github.com/catchorg/Clara) possible. And UDLs can also be helpful, see Literals.h in our codebase or the [units](https://github.com/nholthaus/units) library. Of course they can be misused, but I'd rather catch those during reviews than to potentially prevent their benefits from being utilized because of the discouraging wording.

CONTRIB.md
szaszm: Did you mean `static_assert` instead of `if constexpr`? Or `std::conditional` instead of `std::enable_if`? Because they serve different purposes.
I don't want to discourage TMP too much. Of course `constexpr` is preferred for values, but TMP allows some great interfaces and libraries to exist. One such example is the standard ranges library.

CONTRIB.md
szaszm: I'll remove the line then. That should mean fallback to the Google Style Guide.

CONTRIB.md
fgerlits: I was thinking of examples like https://www.bfilipek.com/2018/03/ifconstexpr.html, where `if constexpr` can simplify code which would have been written using `std::enable_if` before.

CONTRIB.md
fgerlits: Operator overloading can be nice, especially if done well in a library.  But it can easily be done poorly, and it does make navigating the code harder.  So I would like to lightly discourage it, as above, but I'd be interested in what others think.

lordgamez: I'm not a fan of operator overloading, but it is because of my own bad experiences. But I think my own bad experiences shouldn't discourage anyone who would create an excellent interface through operator overloading. I wouldn't add the `but use them sparingly` part, and I would rather say `but only use them if they do not make the code less readable`. In that case if we catch a similar problem on review we can point to this part of the CONTRIB.md.

szaszm: What do you think about not adding anything here and adding a sentence in the text part, emphasizing simplicity and readability? Like "Contributions should aim for simple and readable code".

lordgamez: I'm okay with that.

CONTRIB.md
szaszm: I think the "Prefer" part is a very specific use case and falls under the general goal of simple code anyway. I agree with the proposition, but wouldn't want to include it in the doc.

CONTRIB.md
szaszm: done in 3fd87be

CONTRIB.md
szaszm: done in 3fd87be

CONTRIB.md
szaszm: I came up with a slightly different version in 3fd87be. I'm hoping for to reach consensus, let me know if you like it or not.

CONTRIB.md
szaszm: I came up with a slightly different version in 3fd87be. I'm hoping for to reach consensus, let me know if you like it or not.

fgerlits: Yes, I like the new version.

CONTRIB.md
fgerlits: I would still prefer some indication in this section that operator overloading can easily result in those later-mentioned footguns, but I'm OK with the current version if everyone else is.

szaszm: How about removing the Operator overloading part, falling back on the [style guide](https://google.github.io/styleguide/cppguide.html#Operator_Overloading), which says "Overload operators judiciously"? It also disallows overloading &&, ||, , (comma), and unary &, which seems fine to me, and there is always a possible fallback to this in extra special cases: "It's ok to diverge from any of the rules with a good enough reason.".
```suggestion
- User-defined literal suffixes are OK
```

CONTRIB.md
szaszm: Let's leave this open so that future reviewers can see and join the bikeshedding.

CONTRIB.md
```diff
@@ -15,24 +15,60 @@
...
-the linter for the Google Style guide. Google provides an Eclipse formatter for their style
-guide. It is located [here](https://github.com/google/styleguide/blob/gh-pages/eclipse-cpp-google-style.xml).
-New contributions are expected to follow the Google style guide when it is reasonable.
-Additionally, all new files must include a copy of the Apache License Header.
-
+We welcome all contributions to Apache MiNiFi. All new files must include a copy of the Apache License Header.
+To make development easier, we've included the linter for the Google Style guide. Google provides an Eclipse formatter
+for their style guide. It is located
+[here](https://github.com/google/styleguide/blob/gh-pages/eclipse-cpp-google-style.xml).
+New contributions are expected to follow the Google Style Guide, except for the following points:
```
martinzink, line 22: We could link to the actual google style guide here.
```suggestion
New contributions are expected to follow the [Google C++ Style Guide][https://google.github.io/styleguide/cppguide.html], except for the following points:
```

martinzink (commented):
Nice work. :+1:  There is a Contributing part in README.md as well I think that could be just a link to this file to eliminate the need to update both of them.
And just include the shellscript, python part (which is only mentioned there atm) here.

szaszm:
> Nice work. +1 There is a Contributing part in README.md as well I think that could be just a link to this file to eliminate the need to update both of them.
> And just include the shellscript, python part (which is only mentioned there atm) here.

Moved the info over to CONTRIB.md in f8413dc. Also added a reference to `make shellcheck`, as I think that's the easiest way to invoke it.

CONTRIB.md
```diff
@@ -15,24 +15,70 @@
...
-guide. It is located [here](https://github.com/google/styleguide/blob/gh-pages/eclipse-cpp-google-style.xml).
-New contributions are expected to follow the Google style guide when it is reasonable.
-Additionally, all new files must include a copy of the Apache License Header.
-
+We welcome all contributions to Apache MiNiFi. All new files must include a copy of the Apache License Header.
+To make development easier, we've included the linter for the Google Style guide. Google provides an Eclipse formatter
+for their style guide. It is located
+[here](https://github.com/google/styleguide/blob/gh-pages/eclipse-cpp-google-style.xml).
+New contributions are expected to follow the 
+[Google C++ Style Guide][https://google.github.io/styleguide/cppguide.html], except for the following points:
```
martinzink, line 23: Now I see I used wrong syntax for the link in my suggestion. sry
```suggestion
[Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html), except for the following points:
```

martinzink (approved):

CONTRIB.md
```diff
@@ -15,24 +15,70 @@
...
+
+It's ok to diverge from any of the rules with a good enough reason. We recommend following the C++ Core Guidelines, when
+it doesn't contradict any of the Google Style Guide or the above exceptions.
+
+C++ is a powerful language and "with great power comes great responsibility". Please aim for simple, readable and
+maintainable solutions and avoid footguns. Be open for respectful debate in pull request reviews.
+
+Shell script files shall follow the guidelines and best practices defined by the [shellcheck](https://github.com/koalaman/shellcheck) analysis tool.
+New contributions are expected to pass the shellcheck analysis as part of the verification process.
+If a shellcheck requested change is unfeasable it shall be disabled on per-line basis and will be subjected to review.
```
fgerlits, line 60: typo:
```suggestion
If a shellcheck requested change is unfeasible it shall be disabled on per-line basis and will be subjected to review.
```

fgerlits (approved):

lordgamez (approved):

PR #1149, opened 2021-08-03
MINIFICPP-1038 - Update paho.mqtt.c to a released version

adam-markovics:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/cmake/BundledPahoMqttC.cmake
+++ b/cmake/BundledPahoMqttC.cmake
@@ -32,6 +32,7 @@ function(use_bundled_pahomqttc SOURCE_DIR BINARY_DIR)
     set(PAHOMQTTC_CMAKE_ARGS ${PASSTHROUGH_CMAKE_ARGS}
             "-DCMAKE_INSTALL_PREFIX=${BINARY_DIR}/thirdparty/paho.mqtt.c-install"
             -DPAHO_BUILD_STATIC=TRUE
+            -DPAHO_BUILD_SHARED=FALSE
             -DPAHO_ENABLE_TESTING=FALSE)
     if (OPENSSL_OFF)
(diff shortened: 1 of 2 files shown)


cmake/BundledPahoMqttC.cmake
```diff
@@ -45,7 +45,7 @@ function(use_bundled_pahomqttc SOURCE_DIR BINARY_DIR)
     ExternalProject_Add(
             paho.mqtt.c-external
             GIT_REPOSITORY "https://github.com/eclipse/paho.mqtt.c.git"
-            GIT_TAG "6aa07f575bc9369402b1b252fd280373f8d585ef" # This is not a tagged/released version, but the exact one that was in our repo.
+            GIT_TAG "3b7ae6348bc917d42c04efa962e4868c09bbde9f" # Once tagged as v1.3.9 release, but tags may move
```
szaszm, line 48: Tags can't move AFAIK, so I'd go with the tag name for simplicity

adam-markovics: They can move like this: delete tag (`git tag -d tagname`), put the same tag somewhere else.
From Professional CMake by Craig Scott on ExternalProject_Add (chapter 27.1.1):
"While GIT_TAG does support these different choices, it should be noted that only a commit hash is truly unambiguous.
With git, the commit referenced by a branch or tag name can move over time, so using them does not guarantee a repeatable build."

szaszm: okay. I still think moving release tags would be an ahole move from any project, but I'm fine with leaving it as is if that's what you prefer.

szaszm (approved):

fgerlits (approved):

lordgamez (approved):

PR #1151, opened 2021-08-04
MINIFICPP-1621 fix docker focal build

szaszm:
When upgrading compilers, I broke make u20. Since now we rely on the ubuntu toolchain ppa, we need the add-apt-repository command, and it's not available in ubuntu docker images. We need to install it before calling the bootstrap script in docker.

I also forgot to adapt it like I did with the other docker jobs, due to it not showing up in the CI.

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/docker/focal/Dockerfile
+++ b/docker/focal/Dockerfile
@@ -28,23 +28,22 @@ ENV MINIFI_BASE_DIR /opt/minifi
 ENV MINIFI_HOME $MINIFI_BASE_DIR/nifi-minifi-cpp-$MINIFI_VERSION
 ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64/
 
-RUN apt update
-RUN DEBIAN_FRONTEND="noninteractive" apt install -y openjdk-8-jdk openjdk-8-source python3.9-dev sudo git maven autogen autoconf automake cmake software-properties-common
+RUN apt update \
+    && DEBIAN_FRONTEND="noninteractive" apt install -y openjdk-8-jdk openjdk-8-source python3.9-dev sudo git maven autogen autoconf automake cmake software-properties-common \
+    && mkdir -p "${MINIFI_BASE_DIR}"
 
-RUN mkdir -p $MINIFI_BASE_DIR
 COPY . ${MINIFI_BASE_DIR}
 
 
 FROM build_deps AS release
 ARG ENABLE_JNI
 
-# Run bootstrap
-RUN cd $MINIFI_BASE_DIR && ./bootstrap.sh -t
-
 ENV CC gcc-11
 ENV CXX g++-11
 
-# Build
-RUN cd $MINIFI_BASE_DIR/build \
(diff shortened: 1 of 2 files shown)


docker/focal/Dockerfile
```diff
@@ -29,19 +29,22 @@ ENV MINIFI_HOME $MINIFI_BASE_DIR/nifi-minifi-cpp-$MINIFI_VERSION
 ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64/
 
 RUN apt update
-RUN DEBIAN_FRONTEND="noninteractive" apt install -y openjdk-8-jdk openjdk-8-source python3.9-dev sudo git maven autogen autoconf automake cmake
+RUN DEBIAN_FRONTEND="noninteractive" apt install -y openjdk-8-jdk openjdk-8-source python3.9-dev sudo git maven autogen autoconf automake cmake software-properties-common
 
 RUN mkdir -p $MINIFI_BASE_DIR
```
lordgamez, line 34: These commands could be merged into a single RUN command

docker/focal/Dockerfile
```diff
@@ -29,19 +29,22 @@ ENV MINIFI_HOME $MINIFI_BASE_DIR/nifi-minifi-cpp-$MINIFI_VERSION
...
-	&& make -j$(nproc) package
+
+# Run bootstrap
+RUN cd $MINIFI_BASE_DIR && ./bootstrap.sh -t
+
+ENV CC gcc-11
+ENV CXX g++-11
+
+# Build
+RUN cd $MINIFI_BASE_DIR/build \
```
lordgamez, line 48: ENV variables could be moved up and the two RUN commands merged into a single one

docker/focal/Dockerfile
szaszm: Would it be better that way?

lordgamez: Usually it's recommended to minimize the layer count and as every command in a Dockerfile creates a new layer it is good practice to merge multiple RUN commands to have less layers.

szaszm: Does this include ENV and COPY, or only RUN?

lordgamez: Only RUN commands can be chained in a bash like manner with `&&` ENV and COPY don't really have a similar option

szaszm: I mean do they also introduce separate layers? Should I keep DEBIAN_FRONTEND as a prefix to the command or separate it to an ENV command?

lordgamez: But if the question was regarding the creation of the layers, yes all of those create new layers.

lordgamez: I think keeping it as a prefix is better.

lordgamez (approved):

fgerlits (approved):

adamdebreceni (approved):

adamdebreceni:
would it make sense to add a focal build job to the ci, so we get immediate feedback if it breaks?

szaszm:
Both of our non-docker ubuntu jobs are running on that distro, just not in docker. I'd rather not duplicate this coverage or move it to docker, to keep resource usage in check.

PR #1152, opened 2021-08-04
MINIFICPP-1593 ProcFsMonitor to monitor /proc pseduo filesystem

martinzink:
to capture resource consumption metrics on linux

### For all changes:
     in the commit message?

### For code changes:

### Note:


adamdebreceni:
IMO we should wait to introduce new extension until the dynamic library refactor is merged

martinzink:
> IMO we should wait to introduce new extensions until the dynamic library refactor is merged

Sure thing, I'll will rebase and update when https://github.com/apache/nifi-minifi-cpp/pull/1138 is merged

martinzink:
I will keep this in draft until I rebase againts  #1138

szaszm:
Would it be feasible to create something that supports both windows and linux? PDH and this solve the same problem on different platforms in an incompatible way, but usually we try to write multiplatform code that works similarly on all supported platforms.

martinzink:
> Would it be feasible to create something that supports both windows and linux? PDH and this solve the same problem on different platforms in an incompatible way, but usually we try to write multiplatform code that works similarly on all supported platforms.

In my opition (unfortunately) this problem is too OS specific to be generalized. Windows and Linux has different terminology for these performance metrics (e.g. they handle memory quite differently. they count cpu cycles differently etc)
The basics which can be generalized are already included in the heartbeat https://issues.apache.org/jira/browse/MINIFICPP-1504
The other features of the PerformanceDataMonitor (you can subsribe to any windows specific counter) dont really overlap with this processor.
If we really wanted to create a platformindependent processor we would have to come up with new nonstandard terminology and map the windows performance metrics to linux ones, and vica-versa.

The windows version tries to emulate the Windows Performance Monitor, while this tries to emulate the how top/htop works on linux. (both these tools and their terminology are widely known by sysadmins)
I couldnt find any program that has close to the same feature set as these and would work both on windows and linux.

extensions/procfs/CpuStat.h
```diff
@@ -0,0 +1,140 @@
...
+  static constexpr const char IDLE_PERCENT_STR[] = "idle time %";
+  static constexpr const char IO_WAIT_PERCENT_STR[] = "io wait time %";
+  static constexpr const char IRQ_PERCENT_STR[] = "irq time %";
+  static constexpr const char SOFT_IRQ_PERCENT_STR[] = "soft irq %";
+  static constexpr const char STEAL_PERCENT_STR[] = "steal time %";
+  static constexpr const char GUEST_PERCENT_STR[] = "guest time %";
+  static constexpr const char GUEST_NICE_PERCENT_STR[] = "guest nice time %";
+
+  CpuStatPeriod(const CpuStatPeriod& src) = default;
+  CpuStatPeriod(CpuStatPeriod&& src) noexcept = default;
```
szaszm, line 113: same here

extensions/procfs/MemInfo.h
```diff
@@ -0,0 +1,51 @@
...
+  MemInfo(MemInfo&& src) noexcept = default;
+  static std::optional<MemInfo> parseMemInfoFile(std::ifstream& mem_info_file);
+
+  void addToJson(rapidjson::Value& body, rapidjson::Document::AllocatorType& alloc) const;
+
+  uint64_t getTotalMemory() const { return memory_total_; }
+  uint64_t getFreeMemory() const { return memory_free_; }
+  uint64_t getAvailableMemory() const { return memory_available_; }
+  uint64_t getTotalSwap() const { return swap_total_; }
+  uint64_t getFreeSwap() const { return swap_free_; }
```
szaszm, line 42: Please consider adding `[[nodiscard]]` and `noexcept` to all getters.

extensions/procfs/CpuStat.h
```diff
@@ -0,0 +1,140 @@
...
+  static constexpr const char IO_WAIT_STR[] = "io wait time";
+  static constexpr const char IRQ_STR[] = "irq time";
+  static constexpr const char SOFT_IRQ_STR[] = "soft irq time";
+  static constexpr const char STEAL_STR[] = "steal time";
+  static constexpr const char GUEST_STR[] = "guest time";
+  static constexpr const char GUEST_NICE_STR[] = "guest nice time";
+
+  CpuStat(const CpuStat& src) = default;
+  CpuStat(CpuStat&& src) noexcept = default;
+
```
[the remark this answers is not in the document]
martinzink: The rework removed these classes

extensions/procfs/CpuStat.h
martinzink: The rework removed these classes

extensions/procfs/MemInfo.h
martinzink: good idea, I've added it in https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/f606e41ffaf49145aefa8ff3c21fe0108a42e270

extensions/procfs/DiskStat.h
```diff
@@ -0,0 +1,152 @@
...
+  static constexpr const char READS_COMPLETED_STR[] = "Reads Completed";
+  static constexpr const char READS_MERGED_STR[] = "Reads Merged";
+  static constexpr const char SECTORS_READ_STR[] = "Sectors Read";
+  static constexpr const char WRITES_COMPLETED_STR[] = "Writes Completed";
+  static constexpr const char WRITES_MERGED_STR[] = "Writes Merged";
+  static constexpr const char SECTORS_WRITTEN_STR[] = "Sectors Written";
+  static constexpr const char IOS_IN_PROGRESS_STR[] = "IO-s in progress";
+
+  DiskStat(DiskStatData disk_stat_data, std::chrono::time_point<std::chrono::system_clock> time_point) : data_(disk_stat_data), time_point_(time_point) {}
+  void addToJson(rapidjson::Value& body, rapidjson::Document::AllocatorType& alloc) const;
```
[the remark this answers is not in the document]
martinzink: I reworked the PR based on this, it made the whole thing a lot more simple and hopefully easier to read.
https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/f606e41ffaf49145aefa8ff3c21fe0108a42e270

extensions/procfs/CpuStat.h
```diff
@@ -0,0 +1,140 @@
...
+  uint64_t getIrq() const { return irq_; }
+  uint64_t getSoftIrq() const { return soft_irq_; }
+  uint64_t getSteal() const { return steal_; }
+  uint64_t getGuest() const { return guest_; }
+  uint64_t getGuestNice() const { return guest_nice_; }
+
+  uint64_t getIdleAll() const { return idle_ + io_wait_; }
+  uint64_t getSystemAll() const { return system_ + irq_ + soft_irq_; }
+  uint64_t getVirtAll() const { return guest_ + guest_nice_; }
+  uint64_t getTotal() const { return user_ + nice_ + getSystemAll() + getIdleAll() + steal_ + getVirtAll(); }
```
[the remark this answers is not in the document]
martinzink: good idea, added them in https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/f606e41ffaf49145aefa8ff3c21fe0108a42e270

extensions/procfs/CpuStat.cpp
```diff
@@ -0,0 +1,116 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "CpuStat.h"
+
+namespace org::apache::nifi::minifi::procfs {
+
+std::optional<CpuStatData> CpuStatData::parseCpuStatLine(std::istringstream& iss) {
```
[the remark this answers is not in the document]
martinzink: good idea, changed it in https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/f606e41ffaf49145aefa8ff3c21fe0108a42e270

extensions/procfs/CPPLINT.cfg
```diff
@@ -0,0 +1 @@
+filter=-runtime/int
```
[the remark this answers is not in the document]
martinzink: good catch must be some leftover, anyway removed it in https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/f606e41ffaf49145aefa8ff3c21fe0108a42e270

libminifi/test/SingleInputTestController.h
```diff
@@ -34,9 +34,11 @@ class SingleInputTestController : public TestController {
   {}
 
   std::unordered_map<core::Relationship, std::vector<std::shared_ptr<core::FlowFile>>>
-  trigger(const std::string_view input_flow_file_content, std::unordered_map<std::string, std::string> input_flow_file_attributes = {}) {
-    const auto new_flow_file = createFlowFile(input_flow_file_content, std::move(input_flow_file_attributes));
-    input_->put(new_flow_file);
+  trigger(const std::optional<std::string_view> input_flow_file_content, std::unordered_map<std::string, std::string> input_flow_file_attributes = {}) {
+    if (input_flow_file_content) {
```
szaszm, line 38: `SingleInputTestController` was not designed to be used without an input. (It's in the name.) If you want to reuse it without an input, please refactor it to extract and customize the input generation part.

martinzink: My bad, I thought the SingleInput referenced that there is a single incoming connection(which can be empty) and not that there is a single input flowfile.

Looking at the code the main feature seems to be there is a single processors in the flow which's connections are made available for testing and not really the number of input flowfiles, how about I rename it to `SingleProcessorTestController` something like this https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/4048d55e52833862a4b49da97e8135f02e9db4e4?

What do you think?

extensions/procfs/ProcFsSerialization.h
```diff
@@ -0,0 +1,164 @@
...
+                                    const std::chrono::duration<double> all_cpu_time,
+                                    std::invocable<const char(&)[], const uint64_t> auto uint64_t_serializer,
+                                    std::invocable<const char(&)[], const std::string_view&> auto string_serializer,
+                                    std::invocable<const char(&)[], const double> auto double_serializer) {
+  gsl_Expects(all_cpu_time > 0ms);
+  gsl_Expects(process_stat_start.getComm() == process_stat_end.getComm());
+  gsl_Expects(process_stat_end.getCpuTime() >= process_stat_start.getCpuTime());
+  auto cpu_time_diff = process_stat_end.getCpuTime()-process_stat_end.getCpuTime();
+  string_serializer("COMM", process_stat_start.getComm());
+  uint64_t_serializer("RES", process_stat_start.getMemory());
```
fgerlits, line 160: I think
```suggestion
  uint64_t_serializer("RES", process_stat_end.getMemory());
```
would be more natural

fgerlits (commented):
I'm half way through, second half coming up later.

extensions/procfs/ProcFsSerialization.h
```diff
@@ -0,0 +1,164 @@
...
+  serializer("Transmit fifo errors/sec", net_dev_data.getTransmitFifoErrors()/duration.count());
+  serializer("Transmit collisions/sec", net_dev_data.getTransmitCollisions()/duration.count());
+  serializer("Transmit carrier losses/sec", net_dev_data.getTransmitCarrierLosses()/duration.count());
+  serializer("Compressed Packets Transmitted/sec", net_dev_data.getCompressedPacketsTransmitted()/duration.count());
+}
+
+void SerializeProcessStat(const ProcessStat& process_stat,
+                          std::invocable<const char(&)[], const uint64_t> auto uint64_t_serializer,
+                          std::invocable<const char(&)[], const std::string_view&> auto string_serializer) {
+  string_serializer("COMM", process_stat.getComm());
```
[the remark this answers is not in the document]
martinzink: Not sure how standard that is.
I've got the naming scheme from https://man7.org/linux/man-pages/man5/proc.5.html
this is under `/proc/[pid]/stat` section

extensions/procfs/CpuStat.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+namespace org::apache::nifi::minifi::extensions::procfs {
+
+std::optional<CpuStatData> CpuStatData::parseCpuStatLine(std::istream& iss) {
+  CpuStatData data;
+  iss >> data.user_ >> data.nice_ >> data.system_ >> data.idle_ >> data.io_wait_ >> data.irq_ >> data.soft_irq_ >> data.steal_ >> data.guest_ >> data.guest_nice_;
+  if (iss.fail())
+    return std::nullopt;
+  data.user_ -= data.guest_;  // Guest time is already accounted in usertime
+  data.nice_ -= data.guest_nice_;
```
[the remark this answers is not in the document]
martinzink: Thats an excellent question. I did this because I wanted to reproduce what htop reports.
https://github.com/htop-dev/htop/blob/main/linux/LinuxProcessList.c#L1964-L1965

extensions/procfs/ProcFsSerialization.h
```diff
@@ -0,0 +1,163 @@
...
+  uint64_t_serializer("RES", process_stat.getMemory());
+  uint64_t_serializer("CPUTIME", process_stat.getCpuTime().count());
+}
+
+void SerializeNormalizedProcessStat(const ProcessStat& process_stat_start,
+                                    const ProcessStat& process_stat_end,
+                                    const std::chrono::duration<double> all_cpu_time,
+                                    std::invocable<const char(&)[], const uint64_t> auto uint64_t_serializer,
+                                    std::invocable<const char(&)[], const std::string_view&> auto string_serializer,
+                                    std::invocable<const char(&)[], const double> auto double_serializer) {
```
szaszm, line 153: I think a single serializer with an overloaded or templated call operator would be nicer.
You can use the widely used `overloaded` utility:
```

File: extensions/procfs/ProcessStat.cpp [hunk 13]
```diff
@@ -0,0 +1,54 @@
...
+  ProcessStatData process_stat_data;
+  stat_stream >> process_stat_data.pid_;
+
+  std::string stat_stream_string = stat_stream.str();
+  size_t comm_start = stat_stream_string.find_first_of('(');
+  size_t comm_end = stat_stream_string.find_last_of(')');
+  if (comm_start == std::string::npos  || comm_end == std::string::npos)
+    return std::nullopt;
+  process_stat_data.comm_ = stat_stream_string.substr(comm_start+1, comm_end-comm_start-1);
+  stat_stream.seekg(comm_end+2);
```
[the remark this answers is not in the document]
martinzink: comm fields can contain whitespaces, so this wouldn't work unfortunately

libminifi/test/SingleProcessorTestController.h
```diff
@@ -54,6 +51,12 @@ class SingleInputTestController : public TestController {
     return result;
   }
 
+  auto trigger(const std::string_view input_flow_file_content, std::unordered_map<std::string, std::string> input_flow_file_attributes = {}) {
+    const auto new_flow_file = createFlowFile(input_flow_file_content, std::move(input_flow_file_attributes));
+    input_->put(new_flow_file);
+    return trigger();
+  }
```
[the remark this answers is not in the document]
martinzink: Good idea, changed it in https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/84e464c7b9501293616bcae170c4a12a710f7a6d#diff-a019248b6c700d2a0bc69681c6ea011e5afe07861a32a021f992375eeda11847R54

extensions/procfs/DiskStat.cpp
```diff
@@ -0,0 +1,70 @@
...
+  std::string disk_name;
+  iss >> disk_stat_data.major_device_number_ >> disk_stat_data.minor_device_number_ >> disk_name >> disk_stat_data.monotonic_increasing_members_.reads_completed_
+      >> disk_stat_data.monotonic_increasing_members_.reads_merged_ >> disk_stat_data.monotonic_increasing_members_.sectors_read_
+      >> disk_stat_data.monotonic_increasing_members_.milliseconds_spent_reading_ >> disk_stat_data.monotonic_increasing_members_.writes_completed_
+      >> disk_stat_data.monotonic_increasing_members_.writes_merges_ >> disk_stat_data.monotonic_increasing_members_.sectors_written_
+      >> disk_stat_data.monotonic_increasing_members_.milliseconds_spent_reading_ >> disk_stat_data.monotonic_increasing_members_.writes_completed_
+      >> disk_stat_data.monotonic_increasing_members_.writes_merges_ >> disk_stat_data.monotonic_increasing_members_.sectors_written_
+      >> disk_stat_data.monotonic_increasing_members_.milliseconds_spent_writing_ >> disk_stat_data.ios_in_progress_ >> disk_stat_data.ios_in_progress_
+      >> disk_stat_data.monotonic_increasing_members_.milliseconds_spent_io_ >> disk_stat_data.monotonic_increasing_members_.weighted_milliseconds_spent_io_;
```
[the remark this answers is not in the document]
martinzink: wow awesome catch, fixed it in https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/84e464c7b9501293616bcae170c4a12a710f7a6d#diff-9e8cddde6f61ed0902194bc13cdfbd1d5e6635c96874a12cba648e2d5d1ae43fR44-R57

extensions/procfs/CpuStat.h
```diff
@@ -0,0 +1,68 @@
...
+#include <string>
+
+#include "SystemClockDuration.h"
+
+using namespace std::literals::chrono_literals;
+
+namespace org::apache::nifi::minifi::extensions::procfs {
+
+class CpuStatData {
+  CpuStatData() = default;
```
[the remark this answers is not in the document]
martinzink: makes sense, changed it in  https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/84e464c7b9501293616bcae170c4a12a710f7a6d

extensions/procfs/MemInfo.h
```diff
@@ -0,0 +1,46 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <string>
```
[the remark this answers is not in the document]
martinzink: fixed it in https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/84e464c7b9501293616bcae170c4a12a710f7a6d#diff-6fed75c9f3cfbff3e0321e0e3d12e62ab961bfc77fc3f38ebcc758a9507cf35fR20

extensions/procfs/NetDev.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+namespace org::apache::nifi::minifi::extensions::procfs {
+
+std::optional<NetDevData> NetDevData::parseNetDevLine(std::istream& iss) {
+  NetDevData net_dev_data;
+  iss >> net_dev_data.bytes_received_ >> net_dev_data.packets_received_ >> net_dev_data.errs_received_ >> net_dev_data.drop_errors_received_
+      >> net_dev_data.fifo_errors_received_ >> net_dev_data.frame_errors_received_ >> net_dev_data.compressed_packets_received_
+      >> net_dev_data.multicast_frames_received_>> net_dev_data.bytes_transmitted_ >> net_dev_data.packets_transmitted_
+      >> net_dev_data.errs_transmitted_ >> net_dev_data.drop_errors_transmitted_ >> net_dev_data.fifo_errors_transmitted_
+      >> net_dev_data.collisions_transmitted_ >> net_dev_data.carrier_losses_transmitted_ >> net_dev_data.compressed_packets_transmitted_;
```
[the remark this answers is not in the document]
martinzink: makes sense, changed it in  https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/84e464c7b9501293616bcae170c4a12a710f7a6d#diff-bb8cccc4b43a558560608fd5680317d7f8585c7d18d6f1b6b4c84f3036020667R26-R43

extensions/procfs/ProcFs.cpp
```diff
@@ -0,0 +1,102 @@
...
+}
+
+std::optional<MemInfo> ProcFs::getMemInfo() const {
+  auto mem_info_file_path = root_path_ / MEMINFO_FILE;
+  std::ifstream mem_info_file(mem_info_file_path);
+  return MemInfo::parseMemInfoFile(mem_info_file);
+}
+
+std::unordered_map<std::string, NetDevData> ProcFs::getNetDevs() const {
+  std::unordered_map<std::string, NetDevData>net_devs;
```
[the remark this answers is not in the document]
martinzink: thanks :+1: 
https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/84e464c7b9501293616bcae170c4a12a710f7a6d#diff-2b6e743856f10f8108cb9791fda43ac483fc3e7b8a8b6d2d4bbf488fe057b299R73

 (maybe the linter gets confused with the template parameters?)

extensions/procfs/ProcFs.cpp
```diff
@@ -0,0 +1,102 @@
...
+  std::string line;
+  std::getline(stat_file, line);
+  std::getline(stat_file, line);
+  while (std::getline(stat_file, line)) {
+    std::istringstream iss(line);
+    std::string entry_name;
+    iss >> entry_name;
+    if (iss.fail())
+      continue;
+    entry_name.pop_back();
```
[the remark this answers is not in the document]
martinzink: makes sense, I've added the comment in https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/84e464c7b9501293616bcae170c4a12a710f7a6d#diff-bb8cccc4b43a558560608fd5680317d7f8585c7d18d6f1b6b4c84f3036020667R47

extensions/procfs/ProcFs.cpp
```diff
@@ -0,0 +1,102 @@
...
+    if (auto net_dev_data = NetDevData::parseNetDevLine(iss))
+      net_devs.emplace(entry_name, *net_dev_data);
```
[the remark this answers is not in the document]
martinzink: Good idea :+1: , i've added log_errors in https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/84e464c7b9501293616bcae170c4a12a710f7a6d#diff-2b6e743856f10f8108cb9791fda43ac483fc3e7b8a8b6d2d4bbf488fe057b299R38

extensions/procfs/ProcFsJsonSerialization.h
```diff
@@ -0,0 +1,179 @@
...
+
+namespace org::apache::nifi::minifi::extensions::procfs {
+
+namespace details {
+class UInt64Serializer {
+ public:
+  UInt64Serializer(rapidjson::Value& json, rapidjson::Document::AllocatorType& alloc) :
+      json_(json), alloc_(alloc) {
+  }
+  void operator()(const char(&key)[], const uint64_t value) {
```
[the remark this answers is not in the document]
martinzink: Yeah, it was leftover from when the literals where static constexpr members of the classes (it didnt work with rapidjson otherwise).
But entirely unnecessary now, so changed it in https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/84e464c7b9501293616bcae170c4a12a710f7a6d#diff-8a4db62b36fb7340e48e062700fa2ae5b9c3851563ef91e310a51b84ba0e44c4R35

extensions/procfs/ProcFsSerialization.h
```diff
@@ -0,0 +1,164 @@
...
+                           std::invocable<const char(&)[], const uint64_t> auto serializer) {
+  serializer("Major Device Number", disk_stat_data.getMajorDeviceNumber());
+  serializer("Minor Device Number", disk_stat_data.getMinorDeviceNumber());
+  serializer("Reads Completed", disk_stat_data.getReadsCompleted());
+  serializer("Reads Merged", disk_stat_data.getReadsMerged());
+  serializer("Sectors Read", disk_stat_data.getSectorsRead());
+  serializer("Writes Completed", disk_stat_data.getWritesCompleted());
+  serializer("Writes Merged", disk_stat_data.getWritesMerged());
+  serializer("Sectors Written", disk_stat_data.getSectorsWritten());
+  serializer("IO-s in progress", disk_stat_data.getIosInProgress());
```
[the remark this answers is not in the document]
martinzink: agree, changed it in  https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/84e464c7b9501293616bcae170c4a12a710f7a6d#diff-cb54b002ac8341fabeaff2d9dbbf25274c8a38a9bcb0e9e3a61e67b76f373dbdR75

extensions/procfs/ProcFsSerialization.h
martinzink: Good idea, reworked it in https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/84e464c7b9501293616bcae170c4a12a710f7a6d#diff-8a4db62b36fb7340e48e062700fa2ae5b9c3851563ef91e310a51b84ba0e44c4R30

szaszm (approved):

extensions/procfs/processors/ProcFsMonitor.cpp
```diff
@@ -0,0 +1,283 @@
...
+
+void ProcFsMonitor::setupDecimalPlacesFromProperties(const core::ProcessContext& context) {
+  std::string decimal_places_str;
+  if (!context.getProperty(DecimalPlaces.getName(), decimal_places_str) || decimal_places_str.empty()) {
+    decimal_places_ = std::nullopt;
+    return;
+  }
+
+  if (auto decimal_places = context.getProperty<int64_t>(DecimalPlaces)) {
+    if (*decimal_places > std::numeric_limits<uint8_t>::max())
```
fgerlits, line 161: we should probably throw if `*decimal_places` is `<= 0`, too, especially since `0` sounds like a reasonable value, but RapidJSON doesn't allow it: https://github.com/apache/nifi-minifi-cpp/blob/05560bf9bcf2aa3f476fcac016efd41083c93c09/thirdparty/rapidjson-48fbd8cd202ca54031fe799db2ad44ffa8e77c13/include/rapidjson/internal/dtoa.h#L217

extensions/procfs/ProcFsSerialization.h
```diff
@@ -0,0 +1,183 @@
...
+                                    const ProcessStat& process_stat_end,
+                                    const std::chrono::duration<double> all_cpu_time,
+                                    Serializer serializer) {
+  gsl_Expects(all_cpu_time > 0ms);
+  gsl_Expects(process_stat_start.getComm() == process_stat_end.getComm());
+  gsl_Expects(process_stat_end.getCpuTime() >= process_stat_start.getCpuTime());
+  auto cpu_time_diff = process_stat_end.getCpuTime()-process_stat_end.getCpuTime();
+  serializer("COMM", process_stat_start.getComm());
+  serializer("RES", process_stat_end.getMemory());
+  serializer("CPU%", cpu_time_diff/all_cpu_time);
```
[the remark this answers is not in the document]
martinzink: They are both chrono::durations so the different base is not a problem, chrono works it out in the background.

e.g. cpu_time_diff (102) / all_cpu_time (1.0216666666480403) -> 0.99836867864789169

(however i did find a serious typo in line 177 while checking this out)

extensions/procfs/processors/ProcFsMonitor.cpp
martinzink: good idea, changed it in https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/491d4746d43fcf6da3f82d4e1b0708aab89c44b2#diff-d776bd6998730fd9d2699bfd677cabc117092a8692c9b1a73392e7ba5310a529R160-R169

extensions/procfs/tests/MemInfoTests.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include "Catch.h"
+#include "ProcFs.h"
+
+using org::apache::nifi::minifi::extensions::procfs::ProcFs;
+using org::apache::nifi::minifi::extensions::procfs::MemInfo;
+
+TEST_CASE("ProcFSTest meminfo test with mock", "[procfmeminfomocktest]") {
+  ProcFs proc_fs("./mockprocfs_t0");
+  auto mem_info = proc_fs.getMemInfo();
+  REQUIRE(mem_info);
```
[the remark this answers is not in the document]
martinzink: Good idea, added checks in https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/491d4746d43fcf6da3f82d4e1b0708aab89c44b2

extensions/procfs/tests/CPUStatTests.cpp
```diff
@@ -0,0 +1,85 @@
...
+  percentage += cpu_stat.getGuest()/cpu_stat.getTotal();
+  percentage += cpu_stat.getGuestNice()/cpu_stat.getTotal();
+  REQUIRE(percentage == Approx(1.0));
+}
+
+TEST_CASE("ProcFSTest stat test with mock", "[procfsstatmockabsolutetest]") {
+  ProcFs proc_fs_t0("./mockprocfs_t0");
+  auto cpu_stats_t0 = proc_fs_t0.getCpuStats();
+  REQUIRE(cpu_stats_t0.size() == 13);
+  for (auto& [cpu_name, cpu_stat] : cpu_stats_t0) {
```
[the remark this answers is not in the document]
martinzink: :+1: changed them in  https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/491d4746d43fcf6da3f82d4e1b0708aab89c44b2

extensions/procfs/tests/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+    target_include_directories(${testfilename} PRIVATE BEFORE "${CMAKE_SOURCE_DIR}/extensions/procfs")
+    target_include_directories(${testfilename} PRIVATE BEFORE "${CMAKE_SOURCE_DIR}/extensions/procfs/procfs_classes")
+    target_include_directories(${testfilename} PRIVATE BEFORE "${CMAKE_SOURCE_DIR}/libminifi/test/")
+    target_include_directories(${testfilename} PRIVATE BEFORE "${CMAKE_SOURCE_DIR}/extensions/standard-processors")
+    createTests("${testfilename}")
+    target_link_libraries(${testfilename} ${CATCH_MAIN_LIB})
+    target_link_libraries(${testfilename} minifi-procfs)
+    target_link_libraries(${testfilename} minifi-standard-processors)
+    MATH(EXPR PROCFS_TEST_COUNT "${PROCFS_TEST_COUNT}+1")
+    add_test(NAME "${testfilename}" COMMAND "${testfilename}" "${MOCK_PROC_FS_PATH}")
```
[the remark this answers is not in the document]
martinzink: seems leftover but honestly dont remember, removed it in https://github.com/apache/nifi-minifi-cpp/commit/491d4746d43fcf6da3f82d4e1b0708aab89c44b2#diff-9586e0249e1560c57cdaa95f082a03ce2090afe5e902d02a5b1ebfcb84b0177cR35

extensions/procfs/tests/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+
+file(GLOB PROCFS_TESTS  "*.cpp")
+
+SET(PROCFS_TEST_COUNT 0)
+FOREACH(testfile ${PROCFS_TESTS})
+    get_filename_component(testfilename "${testfile}" NAME_WE)
+    add_executable("${testfilename}" "${testfile}")
+    target_include_directories(${testfilename} PRIVATE BEFORE "${CMAKE_SOURCE_DIR}/extensions/standard-processors")
+    target_include_directories(${testfilename} PRIVATE BEFORE "${CMAKE_SOURCE_DIR}/extensions/procfs")
+    target_include_directories(${testfilename} PRIVATE BEFORE "${CMAKE_SOURCE_DIR}/extensions/procfs/procfs_classes")
```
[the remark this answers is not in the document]
martinzink: :+1: good catch, i've removed it in 
https://github.com/apache/nifi-minifi-cpp/commit/491d4746d43fcf6da3f82d4e1b0708aab89c44b2#diff-9586e0249e1560c57cdaa95f082a03ce2090afe5e902d02a5b1ebfcb84b0177cL28 
https://github.com/apache/nifi-minifi-cpp/commit/491d4746d43fcf6da3f82d4e1b0708aab89c44b2#diff-048c6dce71b00b809f5e462322b4412cf46855b25d4630c0e407e017f63f73cbL22

extensions/procfs/processors/ProcFsMonitor.h
```diff
@@ -0,0 +1,131 @@
...
+  std::optional<uint8_t> decimal_places_;
+  std::shared_ptr<core::logging::Logger> logger_ = core::logging::LoggerFactory<ProcFsMonitor>::getLogger();
+
+  ProcFs proc_fs_;
+
+  std::unordered_map<std::string, CpuStatData> last_cpu_stats_;
+  std::unordered_map<std::string, NetDevData> last_net_devs_;
+  std::unordered_map<std::string, DiskStatData> last_disk_stats_;
+  std::unordered_map<pid_t, ProcessStat> last_process_stats_;
+  std::optional<std::chrono::steady_clock::time_point> last_trigger;
```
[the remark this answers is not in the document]
martinzink: good catch, fixed it in https://github.com/apache/nifi-minifi-cpp/commit/491d4746d43fcf6da3f82d4e1b0708aab89c44b2#diff-d776bd6998730fd9d2699bfd677cabc117092a8692c9b1a73392e7ba5310a529R281

extensions/procfs/processors/ProcFsMonitor.cpp
```diff
@@ -0,0 +1,282 @@
...
+}
+
+void ProcFsMonitor::processCPUInformation(const std::unordered_map<std::string, CpuStatData>& current_cpu_stats,
+                                          rapidjson::Value& body,
+                                          rapidjson::Document::AllocatorType& alloc) {
+  if (!cpu_stats_are_valid(current_cpu_stats))
+    return;
+
+  rapidjson::Value cpu_root{rapidjson::kObjectType};
+  for (auto& [cpu_name, cpu_stat] : current_cpu_stats) {
```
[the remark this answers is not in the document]
martinzink: sure I've changed this and all other loops to const auto& https://github.com/apache/nifi-minifi-cpp/commit/491d4746d43fcf6da3f82d4e1b0708aab89c44b2

extensions/procfs/processors/ProcFsMonitor.cpp
```diff
@@ -0,0 +1,282 @@
...
+    return;
+  }
+
+  if (auto decimal_places = context.getProperty<int64_t>(DecimalPlaces)) {
+    if (*decimal_places > std::numeric_limits<uint8_t>::max())
+      throw Exception(PROCESS_SCHEDULE_EXCEPTION, "ProcFsMonitor Decimal Places is out of range");
+    decimal_places_ = *decimal_places;
+  }
+  if (decimal_places_)
+    logger_->log_trace("Rounding is enabled with %d decimal places", decimal_places_.value());
```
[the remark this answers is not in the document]
martinzink: reworked this part in https://github.com/apache/nifi-minifi-cpp/commit/491d4746d43fcf6da3f82d4e1b0708aab89c44b2#diff-d776bd6998730fd9d2699bfd677cabc117092a8692c9b1a73392e7ba5310a529R160-R169

extensions/procfs/processors/ProcFsMonitor.cpp
```diff
@@ -0,0 +1,282 @@
+/**
+ * @file GenerateFlowFile.cpp
```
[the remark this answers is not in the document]
martinzink: 🤦 https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/666dca50c2b610e4413bd0543510c1bead6f22f9

extensions/procfs/processors/ProcFsMonitor.cpp
```diff
@@ -0,0 +1,283 @@
...
+  if (!context.getProperty(DecimalPlaces.getName(), decimal_places_str) || decimal_places_str.empty()) {
+    decimal_places_ = std::nullopt;
+    return;
+  }
+
+  try {
+    uint32_t decimal_places = std::stoul(decimal_places_str);
+    if (decimal_places > std::numeric_limits<uint8_t>::max() || decimal_places == 0)
+      throw std::out_of_range("");
+    decimal_places_ = decimal_places;
```
fgerlits, line 161: this will give unexpected results if `long` is 64 bits and `decimal_places_str` is something like 2^32 + 3, and although that's not going to happen, I would change this to
```suggestion
    const auto decimal_places = std::stoul(decimal_places_str);
    if (decimal_places > std::numeric_limits<uint8_t>::max() || decimal_places == 0) {
      throw Exception(PROCESS_SCHEDULE_EXCEPTION, "ProcFsMonitor Decimal Places property is zero or too large");
    }
    decimal_places_ = gsl::narrow<uint8_t>(decimal_places);
```

extensions/procfs/ProcFsSerialization.h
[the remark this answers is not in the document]
fgerlits: nice, I missed the typo on line 177 :smiley: 

OK, so `chrono::duration` is smarter than I thought; but then we are missing a factor of 100, aren't we?  If `cpu_time_diff` is 23 seconds (2300 ticks) and `all_cpu_time` is 100 seconds, then "CPU%" should be "23", but right now it is "0.23".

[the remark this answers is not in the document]
martinzink: you are right :+1:, in my mind anything relative is between 0-1, but since I named it CPU% then it should be 0-100 https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/4691fa9cfff94471f19b85180e634642fad4a736

fgerlits (approved):

extensions/procfs/CpuStat.cpp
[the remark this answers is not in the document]
fgerlits: Yeah, both ways are reasonable.  But since this is a ProcFsMonitor, not an HTopMonitor, I think we should either output the same numbers as in `/proc/stat`, or document the difference from `/proc/stat` in PROCESSORS.md.

fgerlits:
Can you add `ENABLE_PROCFS` to `cmake/DockerConfig.cmake` and `docker/Dockerfile`, so we can include the new extension in docker builds, please?

fgerlits (changes requested):
Can you add `ENABLE_PROCFS` to `cmake/DockerConfig.cmake` and `docker/Dockerfile`, so we can include the new extension in docker builds, please?

extensions/procfs/SystemClockDuration.h
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+#include <sys/param.h>
```
fgerlits, line 19: `HZ` is defined in `asm-generic/param.h`, which is (indirectly) included from `sys/param.h` in Ubuntu 20.04, but not in the alpine:3.13 docker image.  Including `asm-generic/param.h` directly works on both platforms.
```suggestion
#include <asm-generic/param.h>
```

docker/Dockerfile
```diff
@@ -123,7 +124,7 @@ RUN cmake -DSTATIC_BUILD= -DSKIP_TESTS=true -DENABLE_ALL="${ENABLE_ALL}" -DENABL
     -DENABLE_TEST_PROCESSORS="${ENABLE_TEST_PROCESSORS}" -DDISABLE_EXPRESSION_LANGUAGE="${DISABLE_EXPRESSION_LANGUAGE}" -DDISABLE_ROCKSDB="${DISABLE_ROCKSDB}" \
     -DDISABLE_LIBARCHIVE="${DISABLE_LIBARCHIVE}" -DDISABLE_LZMA="${DISABLE_LZMA}" -DDISABLE_BZIP2="${DISABLE_BZIP2}" \
     -DENABLE_SCRIPTING="${ENABLE_SCRIPTING}" -DDISABLE_PYTHON_SCRIPTING="${DISABLE_PYTHON_SCRIPTING}" -DENABLE_LUA_SCRIPTING="${ENABLE_LUA_SCRIPTING}" \
-    -DENABLE_KUBERNETES="${ENABLE_KUBERNETES}" \
+    -DENABLE_KUBERNETES="${ENABLE_KUBERNETES}" -DENABLE_PROCFS="${ENABLE_RPOCFS} "\
```
fgerlits, line 127: typo:
```suggestion
    -DENABLE_KUBERNETES="${ENABLE_KUBERNETES}" -DENABLE_PROCFS="${ENABLE_PROCFS} "\
```

extensions/procfs/ProcFs.h
```diff
@@ -0,0 +1,64 @@
...
+ public:
+  explicit ProcFs(std::filesystem::path path = DEFAULT_ROOT_PATH)
+      : root_path_(std::move(path)) {
+  }
+
+  [[nodiscard]] std::unordered_map<pid_t, ProcessStat> getProcessStats() const;
+  [[nodiscard]] std::unordered_map<std::string, CpuStatData> getCpuStats() const;
+  [[nodiscard]] std::optional<MemInfo> getMemInfo() const;
+  [[nodiscard]] std::unordered_map<std::string, NetDevData> getNetDevs() const;
+  [[nodiscard]] std::unordered_map<std::string, DiskStatData> getDiskStats() const;
```
fgerlits, line 55: I don't suppose these JSON files will be read by humans much, but in the few cases when that happens, it would be friendlier to those humans to use ordered `map`s instead of `unordered_map`s, so the CPUs, network interfaces, process IDs etc are not in random order.

Lexicographic ordering is not perfect, as we'll have `cpu0, cpu1, cpu11, cpu2, cpu3...`, but it is better than random.

szaszm: If that's a concert, I'd rather go with vector of pairs, to keep the original order. There's also `utils::FlatMap` which does the same AFAIK.

martinzink: I went with a mix of your ideas.
When it comes to CPUStat, DiskStat, NetDev the original order makes sense and the number of entries are expected to be low so I went with FlatMap.
With ProcessStats the original order is already random (due to std::filesystem::directory_iterator) and I can image scenarios where the number of processes go high where the linear lookup of FlatMap could hurt performance in a measurable way, so I went with std::map.

What do you think? https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/d04a5cd15e513d41c8dc9d6ad19a959ddaee0642

extensions/procfs/CpuStat.cpp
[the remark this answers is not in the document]
martinzink: I see your point, changed it in https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/e4ba2c310d44114a90d4b4d1830a99aa63a14b75

extensions/procfs/ProcFs.h
fgerlits: I like that the orders will be correct, better than with a lexicographic order.

I am a bit bothered by the fact that every time we insert an element into a FlatMap, we do a (linear) lookup first, so inserting N elements takes O(N^2) time.  On the other hand, these Ns are small, so maybe it's not a big issue.  Using a `vector<pair<...>>` instead of a FlatMap would solve this problem, but would make the code less nice.  I would probably go with `vector<pair<...>>` anyway, but it's up to you.

martinzink: I changed the FlatMaps to vector of pairs in https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/da1e32e1bec73705ccad155130947615bb70f804
I also refactored a bit so the processXYZInformation functions are not as overwhelming anymore.

extensions/procfs/ProcFsJsonSerialization.h
```diff
@@ -0,0 +1,156 @@
...
+                       rapidjson::Document::AllocatorType& alloc) {
+  rapidjson::Value disk_key(disk_name.c_str(), disk_name.length(), alloc);
+  disk_root.AddMember(disk_key.Move(), rapidjson::kObjectType, alloc);
+  rapidjson::Value& disk_json = disk_root[disk_name.c_str()];
+  SerializeDiskStatData(disk_stat,
+                        details::Serializer(disk_json, alloc));
+}
+
+void addDiskStatPerSecToJson(const std::string& disk_name,
+                             const DiskStatData disk_stat,
```
fgerlits, line 86: I think this is a typo:
```suggestion
                             const DiskStatData& disk_stat,
```
(same at line 111)

libminifi/include/utils/FlatMap.h
```diff
@@ -147,6 +147,22 @@ class FlatMap{
     return data_.rbegin()->second;
   }
 
+  const V& at(const K& key) const {
```
fgerlits, line 150: These new `FlatMap` member functions are not used in this PR any longer, so I would remove them and their tests.  But they are probably going to be useful later, so if you want to keep them, that's fine, too.

extensions/procfs/ProcFs.h
fgerlits: looks good, thanks!

libminifi/include/utils/FlatMap.h
martinzink: I think we should keep them since they are already written and they make the class more in line with the C++20 std::map/std::unordered_map interface.

extensions/procfs/tests/ProcFsMonitorTests.cpp
```diff
@@ -0,0 +1,157 @@
...
+    }
+  }
+
+  SECTION("Relative without wait") {
+    test_controller_.plan->setProperty(proc_fs_monitor, ProcFsMonitor::ResultRelativenessProperty.getName(), "Relative");
+    const auto& result1 = test_controller_.trigger();
+    REQUIRE(result1.at(ProcFsMonitor::Success).size() == 1);
+    const auto& result2 = test_controller_.trigger();
+    REQUIRE(result2.at(ProcFsMonitor::Success).size() == 1);
+  }
```
fgerlits, line 156: I don't think "Relative without wait" is a useful test case, as the total elapsed CPU time can be zero, and we divide by it in `SerializeNormalizedCPUStat()` (as happened on the last CI run).  I would remove this test case.

martinzink: On the contrary. I made this test so it makes sure we dont divide by zero. It turns out I removed a check during the last refactor.
I've fixed the issue https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/6248890ce7c068ee8fa4eab52598b18a933b388c, I think we can keep this.

fgerlits (approved):

PROCESSORS.md
```diff
@@ -1235,7 +1236,8 @@ In the list below, the names of required properties appear in bold. Any other pr
 |Predefined Groups||CPU<br>IO<br>Disk<br>Network<br>Memory<br>System<br>Process|Comma separated list from the allowable values, to monitor multiple common Windows Performance counters related to these groups. (e.g. "CPU,Network")|
 |Custom PDH Counters|||Comma separated list of Windows Performance Counters to monitor. (e.g. "\\System\\Threads,\\Process(*)\\ID Process")|
 |**Output Format**|JSON|JSON<br>OpenTelemetry|The output format of the new flowfile|
-|Round to decimal places|3|integers|The number of decimal places to round the values to (blank for no rounding)|
+|**OutputCompactness**|Pretty|Pretty<br>Compact|The output format of the new flowfile|
```
adamdebreceni, line 1239: missing space in the property name?

extensions/procfs/NetDev.cpp
```diff
@@ -0,0 +1,73 @@
...
+      >> net_dev_data.errs_transmitted_
+      >> net_dev_data.drop_errors_transmitted_
+      >> net_dev_data.fifo_errors_transmitted_
+      >> net_dev_data.collisions_transmitted_
+      >> net_dev_data.carrier_losses_transmitted_
+      >> net_dev_data.compressed_packets_transmitted_;
+  if (iss.fail())
+    return std::nullopt;
+  if (!entry_name.empty())
+    entry_name.pop_back();  // remove the ':' from the end of 'eth0:' etc
```
adamdebreceni, line 47: we could `gsl_Expect` if the last char is really a colon

extensions/procfs/ProcessStat.cpp
adamdebreceni, line 38: why `comm_end + 2`? is there a whitespace we want to skip after the `)`?

extensions/procfs/processors/ProcFsMonitor.cpp
```diff
@@ -0,0 +1,386 @@
...
+}
+}  // namespace
+
+void ProcFsMonitor::onTrigger(core::ProcessContext*, core::ProcessSession* session) {
+  gsl_Expects(session);
+  std::shared_ptr<core::FlowFile> flowFile = session->create();
+  if (!flowFile) {
+    logger_->log_error("Failed to create flowfile!");
+    yield();
+    return;
```
adamdebreceni, line 101: I am conflicted about this, some processors check if `create` succeeds some don't, I'm leaning towards assuming that it does, what do you think?

extensions/procfs/ProcessStat.cpp
martinzink: yeah there is a mandatory space after the `)`, but you are right this would work just as good with `comm_end+1` (because istream skips the whitespaces)
Should I change it?

adamdebreceni: we can keep it, we could add a comment here

extensions/procfs/processors/ProcFsMonitor.h
```diff
@@ -0,0 +1,135 @@
...
+class ProcFsMonitor : public core::Processor {
+ public:
+  explicit ProcFsMonitor(const std::string& name, utils::Identifier uuid = utils::Identifier())
+      : Processor(name, uuid) {
+  }
+  ProcFsMonitor(const ProcFsMonitor&) = delete;
+  ProcFsMonitor(ProcFsMonitor&&) = delete;
+  ProcFsMonitor& operator=(const ProcFsMonitor&) = delete;
+  ProcFsMonitor& operator=(ProcFsMonitor&&) = delete;
+  ~ProcFsMonitor() override = default;
```
adamdebreceni, line 47: I think the base class CoreComponent has these members deleted

extensions/procfs/tests/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+    target_include_directories(${testfilename} PRIVATE BEFORE "${CMAKE_SOURCE_DIR}/extensions/standard-processors")
+    createTests("${testfilename}")
+    target_link_libraries(${testfilename} ${CATCH_MAIN_LIB})
+    target_link_libraries(${testfilename} minifi-procfs)
+    target_link_libraries(${testfilename} minifi-standard-processors)
+    MATH(EXPR PROCFS_TEST_COUNT "${PROCFS_TEST_COUNT}+1")
+    add_test(NAME "${testfilename}" COMMAND "${testfilename}")
+ENDFOREACH()
+file(COPY "${CMAKE_SOURCE_DIR}/extensions/procfs/tests/mockprocfs_t0" DESTINATION "${CMAKE_BINARY_DIR}/extensions/procfs/tests/")
+file(COPY "${CMAKE_SOURCE_DIR}/extensions/procfs/tests/mockprocfs_t1" DESTINATION "${CMAKE_BINARY_DIR}/extensions/procfs/tests/")
```
adamdebreceni, line 38: are these test resources? sftp tests and script tests copy such resources to somewhere under `${CMAKE_BINARY_DIR}/bin/`

extensions/procfs/tests/CPUStatTests.cpp
```diff
@@ -0,0 +1,105 @@
...
+  REQUIRE(cpu_stats_t1.size() == 13);
+  for (const auto& [cpu_name, cpu_stat] : cpu_stats_t1) {
+    cpu_stat_period_total_should_be_one(cpu_stat);
+  }
+
+  for (size_t i = 0; i < cpu_stats_t1.size(); ++i) {
+    const auto& cpu_name_t0 = cpu_stats_t0[i].first;
+    const auto& cpu_stat_t0 = cpu_stats_t0[i].second;
+    const auto& cpu_name_t1 = cpu_stats_t1[i].first;
+    const auto& cpu_stat_t1 = cpu_stats_t1[i].second;
```
adamdebreceni, line 73: we could use structured binding here as well

extensions/procfs/tests/DiskStatTests.cpp
```diff
@@ -0,0 +1,73 @@
...
+  CHECK(disk_stats_t0[3].second.getMillisecondsSpentIo() == 6150);
+  CHECK(disk_stats_t0[3].second.getWeightedMillisecondsSpentIo() == 15052);
+
+
+  ProcFs proc_fs_t1("./mockprocfs_t1");
+  auto disk_stats_t1 = proc_fs_t1.getDiskStats();
+  REQUIRE(disk_stats_t1.size() == 17);
+
+  for (const auto& disk_stat_t0 : disk_stats_t0) {
+    auto disk_stat_t1 = std::find_if(disk_stats_t1.begin(), disk_stats_t1.end(), [&disk_stat_t0](auto& it){return it.first == disk_stat_t0.first;});
```
adamdebreceni, line 50: note: I think `it` here is a reference to the item not an iterator (I try to reserve `it` to name an iterator, others might not have this convention)

extensions/procfs/tests/MemInfoTests.cpp
```diff
@@ -0,0 +1,39 @@
...
+  CHECK(mem_info->getTotalMemory() == 32895000LL*1024);
+  CHECK(mem_info->getFreeMemory() == 9870588LL*1024);
+  CHECK(mem_info->getAvailableMemory() == 22167288LL*1024);
+  CHECK(mem_info->getTotalSwap() == 9227464LL*1024);
+  CHECK(mem_info->getFreeSwap() == 9188320LL*1024);
```
adamdebreceni, line 32: note: we have a `Literals.h` which has a `_KiB`

extensions/procfs/tests/ProcFsMonitorTests.cpp
```diff
@@ -0,0 +1,157 @@
...
+      document.Parse(content.c_str());
+      REQUIRE(document.IsObject());
+      // First trigger has not enough information for relative output
+      CHECK_FALSE(document.HasMember("CPU"));
+      CHECK_FALSE(document.HasMember("Disk"));
+      CHECK_FALSE(document.HasMember("Network"));
+      CHECK_FALSE(document.HasMember("Process"));
+      CHECK(document.HasMember("Memory"));
+    }
+    sleep(1);
```
adamdebreceni, line 90: is this the linux `sleep` syscall? could we use `std::this_thread::sleep_for` here?

adamdebreceni: is this in seconds? could we lower the wait time?

martinzink: good idea, changed it in https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/f06938c35f0943a66c4bd1d7b895a02da8c0a83f

extensions/procfs/tests/MemInfoTests.cpp
martinzink: good idea, changed it in https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/f06938c35f0943a66c4bd1d7b895a02da8c0a83f

extensions/procfs/tests/DiskStatTests.cpp
martinzink: makes sense, changed it in https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/f06938c35f0943a66c4bd1d7b895a02da8c0a83f

extensions/procfs/tests/CPUStatTests.cpp
martinzink: changed these in in https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/f06938c35f0943a66c4bd1d7b895a02da8c0a83f

extensions/procfs/processors/ProcFsMonitor.h
martinzink: thanks that's good to know, less boilerplate
removed them in https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/df0ddde8ad2fa559dee75a3c51752a6e0d2f0eb0

adamdebreceni (approved):

extensions/procfs/tests/CMakeLists.txt
martinzink: good idea, changed it in https://github.com/apache/nifi-minifi-cpp/pull/1152/commits/6d1985dea3427458200b68e5082df02b15e8d3c9

PR #1153, opened 2021-08-04
MINIFICPP-1622 Fix filename update of CompressContent in case of tar.gz

lordgamez:
In case the "Encapsulate in TAR" and "Update Filename" properties were set the filename property was only updated with the added ".gz" extension. Updated to add ".tar.gz" extension instead and to expect ".tar.gz" extension to be removed in case of decompression with the same properties set.

https://issues.apache.org/jira/browse/MINIFICPP-1622

---------------------------------------------------------------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/libarchive/CompressContent.cpp
+++ b/extensions/libarchive/CompressContent.cpp
@@ -25,6 +25,7 @@
 #include <set>
 #include "core/ProcessContext.h"
 #include "core/ProcessSession.h"
+#include "utils/StringUtils.h"
 
 namespace org {
 namespace apache {
@@ -61,6 +62,8 @@ core::Property CompressContent::BatchSize(
 
 core::Relationship CompressContent::Success("success", "FlowFiles will be transferred to the success relationship after successfully being compressed or decompressed");
 core::Relationship CompressContent::Failure("failure", "FlowFiles will be transferred to the failure relationship if they fail to compress/decompress");
+
+const std::string CompressContent::TAR_EXT = ".tar";
 
 const std::map<std::string, CompressContent::CompressionFormat> CompressContent::compressionFormatMimeTypeMap_{
   {"application/gzip", CompressionFormat::GZIP},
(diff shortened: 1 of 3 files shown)


extensions/libarchive/CompressContent.cpp
```diff
@@ -191,14 +194,20 @@ void CompressContent::processFlowFile(const std::shared_ptr<core::FlowFile>& flo
...
         fileName = fileName + fileExtension;
         session->putAttribute(result, core::SpecialFlowAttribute::FILENAME, fileName);
       }
     } else {
       session->removeAttribute(result, core::SpecialFlowAttribute::MIME_TYPE);
       if (updateFileName_) {
-        if (fileName.size() >= fileExtension.size() && fileName.compare(fileName.size() - fileExtension.size(), fileExtension.size(), fileExtension) == 0) {
+        if (utils::StringUtils::endsWith(fileName, fileExtension)) {
           fileName = fileName.substr(0, fileName.size() - fileExtension.size());
+          if (encapsulateInTar_ && utils::StringUtils::endsWith(fileName, TAR_EXT)) {
```
adamdebreceni, line 208: using the `encapsulateInTar_` during decompression seems strange, shouldn't we infer the need for "de-tar-ing" only from the extension?

lordgamez: Currently the "de-tar-ing" depends on the `Encapsulate in TAR` property in the processor, so if that is not set in decompression only the "gunzip" operation will be run on the data and the end content will still be encapsulated in tar. That is why I followed this same process in the filename changes.

adamdebreceni: oh I see, that makes sense

adam-markovics (approved):

adamdebreceni (approved):

fgerlits (approved):

PR #1158, opened 2021-08-06
MINIFICPP-1616 Create PutAzureDatalakeStorage processor

lordgamez:
- Updated Azure SDK to version 12.2.0
- Removed now unused dependencies of Azure SDK
- Added PutAzureDatalakeStorage for uploading files to Azure Data Lake Gen2 storage

Unfortunately Azure Data Lake Storage is [not supported](https://github.com/Azure/Azurite/issues/553) by the Azurite docker image, so the system testing was only done manually.

https://issues.apache.org/jira/browse/MINIFICPP-1616

-----------------------------------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/azure/processors/PutAzureDataLakeStorage.cpp
```diff
@@ -0,0 +1,167 @@
...
+    }
+  } else if (result.result_code == storage::UploadResultCode::FAILURE) {
+    logger_->log_error("Failed to upload file '%s/%s' to filesystem '%s' on Azure Data Lake storage", params->directory_name, params->filename, params->file_system_name);
+    session->transfer(flow_file, Failure);
+  } else {
+    session->putAttribute(flow_file, "azure.filesystem", params->file_system_name);
+    session->putAttribute(flow_file, "azure.directory", params->directory_name);
+    session->putAttribute(flow_file, "azure.filename", params->filename);
+    session->putAttribute(flow_file, "azure.primaryUri", result.primary_uri);
+    session->putAttribute(flow_file, "azure.length", std::to_string(result.length));
```
fgerlits, line 142: Is it useful to have this `length` field in the `result` struct?  Can its value be different from `flow_file->getSize()`?

extensions/azure/storage/AzureDataLakeStorageClient.cpp
```diff
@@ -0,0 +1,55 @@
...
+  if (!params.directory_name.empty()) {
+    directory_client.CreateIfNotExists();
+  }
+  return directory_client.GetFileClient(params.filename);
+}
+
+bool AzureDataLakeStorageClient::createFile(const PutAzureDataLakeStorageParameters& params) {
+  auto file_client = getFileClient(params);
+  auto response = file_client.CreateIfNotExists();
+  return response.Value.Created;
```
fgerlits, line 46: Is it possible to detect why the file creation failed?  I would expect three different return values here: Success, File Already Exists, and Other Error.  In the last case we don't want to ignore the error, whatever the Conflict Resolution Strategy is.

libminifi/test/azure-tests/PutAzureDataLakeStorageTests.cpp
```diff
@@ -0,0 +1,272 @@
...
+    // Build MiNiFi processing graph
+    plan_ = test_controller_.createPlan();
+    auto mock_data_lake_storage_client = std::make_unique<MockDataLakeStorageClient>();
+    mock_data_lake_storage_client_ptr_ = mock_data_lake_storage_client.get();
+    put_azure_data_lake_storage_ = std::shared_ptr<minifi::azure::processors::PutAzureDataLakeStorage>(
+      new minifi::azure::processors::PutAzureDataLakeStorage("PutAzureDataLakeStorage", utils::Identifier(), std::move(mock_data_lake_storage_client)));
+    auto input_dir = test_controller_.createTempDirectory();
+    std::ofstream input_file_stream(input_dir + utils::file::FileUtils::get_separator() + GETFILE_FILE_NAME);
+    input_file_stream << TEST_DATA;
+    input_file_stream.close();
```
fgerlits, line 108: you could use `putFileToDir` (in TestUtils.h) for this

.github/workflows/ci.yml
```diff
@@ -35,7 +35,7 @@ jobs:
           export LDFLAGS="-L/usr/local/opt/flex/lib"
           export CPPFLAGS="-I/usr/local/opt/flex/include"
           # CPPFLAGS are not recognized by cmake, so we have to force them to CFLAGS and CXXFLAGS to have flex 2.6 working
-          ./bootstrap.sh -e -t && cd build  && cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_FLAGS="${CPPFLAGS} ${CFLAGS}" -DCMAKE_CXX_FLAGS="${CPPFLAGS} ${CXXFLAGS}" -DENABLE_LUA_SCRIPTING=ON -DENABLE_SQL=ON -DUSE_REAL_ODBC_TEST_DRIVER=ON -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_RULE_MESSAGES=OFF -DSTRICT_GSL_CHECKS=AUDIT -DFAIL_ON_WARNINGS=ON .. && cmake --build . --parallel 4
+          ./bootstrap.sh -e -t && cd build  && cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_FLAGS="${CPPFLAGS} ${CFLAGS}" -DCMAKE_CXX_FLAGS="${CPPFLAGS} ${CXXFLAGS}" -DENABLE_LUA_SCRIPTING=ON -DENABLE_SQL=ON -DUSE_REAL_ODBC_TEST_DRIVER=ON -DENABLE_AZURE=ON -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_RULE_MESSAGES=OFF -DSTRICT_GSL_CHECKS=AUDIT -DFAIL_ON_WARNINGS=ON .. && cmake --build . --parallel 4
```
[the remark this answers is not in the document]
lordgamez: It is already present in the ubuntu and windows builds and was only missing in the MacOS build, so I added there as well.

extensions/azure/storage/AzureDataLakeStorageClient.cpp
lordgamez: It is done in the AzureDataLakeStorage class. If an error occurs it throws an exception from the SDK, but if the file already exists it does not throw any errors, only sets the `response.Value.Created` value false. In AzureDataLakeStorage::uploadFile we return these 3 possible values represented in the UploadResultCode enum. I wanted to minimize the functionality of the AzureDataLakeStorageClient as it is mocked in the tests, so it should mostly be a wrapper of the Azure SDK client.

extensions/azure/processors/AzureStorageProcessor.h
```diff
@@ -0,0 +1,51 @@
...
+#include <string>
+
+#include "core/Property.h"
+#include "core/Processor.h"
+#include "core/logging/Logger.h"
+#include "core/logging/LoggerConfiguration.h"
+
+namespace org::apache::nifi::minifi::azure::processors {
+
+class AzureStorageProcessor : public core::Processor {
```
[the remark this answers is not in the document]
lordgamez: I'm usually not a fan of using `Base` suffixes in inheritance, but in this case I think it's okay to avoid confusion. Updated in c8257e48b454819ecef4cc2f527f2457a5f81b89

extensions/azure/processors/AzureStorageProcessor.h
```diff
@@ -0,0 +1,51 @@
...
+ public:
+  // Supported Properties
+  static const core::Property AzureStorageCredentialsService;
+
+  explicit AzureStorageProcessor(const std::string& name, const minifi::utils::Identifier& uuid, const std::shared_ptr<logging::Logger>& logger)
```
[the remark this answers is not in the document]
lordgamez: Updated in c8257e48b454819ecef4cc2f527f2457a5f81b89

extensions/azure/processors/PutAzureDataLakeStorage.h
```diff
@@ -0,0 +1,106 @@
...
+    (FAIL_FLOW, "fail"),
+    (REPLACE_FILE, "replace"),
+    (IGNORE_REQUEST, "ignore")
+  )
+
+  explicit PutAzureDataLakeStorage(const std::string& name, const minifi::utils::Identifier& uuid = minifi::utils::Identifier())
+    : PutAzureDataLakeStorage(name, uuid, nullptr) {
+  }
+
+  ~PutAzureDataLakeStorage() override = default;
```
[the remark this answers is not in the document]
lordgamez: Updated in c8257e48b454819ecef4cc2f527f2457a5f81b89

extensions/azure/processors/PutAzureDataLakeStorage.cpp
lordgamez: I suppose we do not need it, especially as it is part of the input. Updated in c8257e48b454819ecef4cc2f527f2457a5f81b89

libminifi/test/azure-tests/PutAzureDataLakeStorageTests.cpp
lordgamez: Updated in c8257e48b454819ecef4cc2f527f2457a5f81b89

fgerlits (approved):

extensions/azure/processors/AzureStorageProcessorBase.h
```diff
@@ -0,0 +1,51 @@
...
+
+#pragma once
+
+#include <memory>
+#include <string>
+
+#include "core/Property.h"
+#include "core/Processor.h"
+#include "core/logging/Logger.h"
+#include "core/logging/LoggerConfiguration.h"
```
adamdebreceni, line 29: is `LoggerConfiguration.h` needed?

extensions/azure/processors/PutAzureDataLakeStorage.cpp
```diff
@@ -0,0 +1,167 @@
...
+  if (params.filename.empty() && (!flow_file->getAttribute("filename", params.filename) || params.filename.empty())) {
+    logger_->log_error("No File Name is set and default object key 'filename' attribute could not be found!");
+    return std::nullopt;
+  }
+
+  return params;
+}
+
+void PutAzureDataLakeStorage::onTrigger(const std::shared_ptr<core::ProcessContext>& context, const std::shared_ptr<core::ProcessSession>& session) {
+  logger_->log_debug("PutAzureDataLakeStorage onTrigger");
```
adamdebreceni, line 105: should we log at `trace` here? I see that all S3/Azure processors use the debug level in `onTrigger`

extensions/azure/processors/PutAzureDataLakeStorage.cpp
```diff
@@ -0,0 +1,167 @@
...
+    return;
+  }
+
+  const auto params = buildUploadParameters(context, flow_file);
+  if (!params) {
+    session->transfer(flow_file, Failure);
+    return;
+  }
+
+  PutAzureDataLakeStorage::ReadCallback callback(flow_file->getSize(), azure_data_lake_storage_, *params, logger_);
```
adamdebreceni, line 118: do we need to pass in the `flow_file->getSize()`, can't we rely on `stream->size()`?

adamdebreceni: I realize that in theory one resource stream could contain the contents of multiple flowfiles (that is why we seek to `flow_file->getOffset()`

extensions/azure/processors/PutAzureDataLakeStorage.cpp
```diff
@@ -0,0 +1,167 @@
...
+}
+
+int64_t PutAzureDataLakeStorage::ReadCallback::process(const std::shared_ptr<io::BaseStream>& stream) {
+  std::vector<uint8_t> buffer;
+  int read_ret = stream->read(buffer, flow_size_);
+  if (io::isError(read_ret)) {
+    return -1;
+  }
+
+  result_ = azure_data_lake_storage_.uploadFile(params_, buffer.data(), flow_size_);
```
adamdebreceni, line 163: here we might have read less than `flow_size_`

extensions/azure/storage/AzureDataLakeStorageClient.cpp
```diff
@@ -0,0 +1,55 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "AzureDataLakeStorageClient.h"
+
+namespace org::apache::nifi::minifi::azure::storage {
+
+void AzureDataLakeStorageClient::resetClientIfNeeded(const std::string& connection_string, const std::string& file_system_name) {
```
adamdebreceni, line 25: it seems to me that a single `AzureDataLakeStorage` instance is contained in the processors, two parallel threads of which might call `resetClientIfNeeded` simultaneously without locks, with possibly different `file_system_name`

extensions/azure/processors/PutAzureDataLakeStorage.cpp
lordgamez: I suppose it's due to the underlying buffer's size? What would be the most secure way here to read? If we read the data to a temporary buffer, can we know what's the appropriate buffer size for our case?

extensions/azure/storage/AzureDataLakeStorageClient.cpp
lordgamez: Every processor has its own `AzureDataLakeStorage` instance containing separate clients for each processor, there should be no problems with concurrency.

lordgamez: Do you think we should implement a mechanism to process the requests in parallel until a client reset is needed (then wait for the ongoing parallel requests to finish and reset the client continuing to process the queued requests) or sequential processing with a lock is sufficient?

extensions/azure/processors/PutAzureBlobStorage.cpp
```diff
@@ -90,14 +81,13 @@ const core::Relationship PutAzureBlobStorage::Failure("failure", "Unsuccessful o
 
 void PutAzureBlobStorage::initialize() {
   // Set the supported properties
-  setSupportedProperties({
+  updateSupportedProperties({
```
arpadboda, line 84: Why not set? 
Update would make sense to me in case the base class would already set some properties, but that doesn't seem to happen as far as I see.

arpadboda (commented):
Mostly LGTM

extensions/azure/processors/PutAzureBlobStorage.cpp
lordgamez: The base class `AzureStorageProcessorBase` does set the `AzureStorageCredentialsService` property beforehand in its constructor.

extensions/azure/storage/AzureDataLakeStorageClient.cpp
adamdebreceni: in light of the recent discussions it is okay if we restrict the processor to a single thread by a mutex

extensions/azure/processors/PutAzureDataLakeStorage.cpp
```diff
@@ -0,0 +1,167 @@
...
+  context->getProperty(FileName, params.filename, flow_file);
+  if (params.filename.empty() && (!flow_file->getAttribute("filename", params.filename) || params.filename.empty())) {
+    logger_->log_error("No File Name is set and default object key 'filename' attribute could not be found!");
+    return std::nullopt;
+  }
+
+  return params;
+}
+
+void PutAzureDataLakeStorage::onTrigger(const std::shared_ptr<core::ProcessContext>& context, const std::shared_ptr<core::ProcessSession>& session) {
```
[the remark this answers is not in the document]
adam-markovics: https://issues.apache.org/jira/browse/MINIFICPP-1566

extensions/azure/processors/PutAzureDataLakeStorage.cpp
```diff
@@ -0,0 +1,167 @@
...
+  if (!params) {
+    session->transfer(flow_file, Failure);
+    return;
+  }
+
+  PutAzureDataLakeStorage::ReadCallback callback(flow_file->getSize(), azure_data_lake_storage_, *params, logger_);
+  session->read(flow_file, &callback);
+  auto result = callback.getResult();
+  if (result.result_code == storage::UploadResultCode::FILE_ALREADY_EXISTS) {
+    gsl_Expects(conflict_resolution_strategy_ != FileExistsResolutionStrategy::REPLACE_FILE);
```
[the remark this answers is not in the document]
lordgamez: In this case if the `conflict_resolution_strategy_` is set to `REPLACE_FILE` it should never return a result that `FILE_ALREADY_EXISTS` as it should be overwritten. This condition should be an invariant here and it is a sign of faulty functionality if it is not fulfilled. Any possible handling of this case seem right to me.

extensions/azure/storage/AzureDataLakeStorage.h
```diff
@@ -0,0 +1,57 @@
...
+class AzureDataLakeStorage {
+ public:
+  AzureDataLakeStorage();
+  explicit AzureDataLakeStorage(std::unique_ptr<DataLakeStorageClient> data_lake_storage_client);
+
+  storage::UploadDataLakeStorageResult uploadFile(const storage::PutAzureDataLakeStorageParameters& params, const uint8_t* buffer, std::size_t buffer_size);
+
+ private:
+  std::shared_ptr<logging::Logger> logger_{logging::LoggerFactory<AzureDataLakeStorage>::getLogger()};
+  std::unique_ptr<DataLakeStorageClient> data_lake_storage_client_;
```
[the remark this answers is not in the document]
lordgamez: Updated in f02f86cd63fd4d08e8a125d469ada637bd7635fe

extensions/azure/processors/PutAzureDataLakeStorage.cpp
```diff
@@ -0,0 +1,167 @@
...
+    params_(params),
+    logger_(std::move(logger)) {
+}
+
+int64_t PutAzureDataLakeStorage::ReadCallback::process(const std::shared_ptr<io::BaseStream>& stream) {
+  std::vector<uint8_t> buffer;
+  int read_ret = stream->read(buffer, flow_size_);
+  if (io::isError(read_ret)) {
+    return -1;
+  }
```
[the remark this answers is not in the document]
lordgamez: Updated in f02f86cd63fd4d08e8a125d469ada637bd7635fe

extensions/azure/processors/AzureStorageProcessorBase.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+std::string AzureStorageProcessorBase::getConnectionStringFromControllerService(const std::shared_ptr<core::ProcessContext> &context) const {
+  std::string service_name;
+  if (!context->getProperty(AzureStorageCredentialsService.getName(), service_name) || service_name.empty()) {
+    return "";
+  }
+
+  std::shared_ptr<core::controller::ControllerService> service = context->getControllerService(service_name);
+  if (nullptr == service) {
+    logger_->log_error("Azure Storage credentials service with name: '%s' could not be found", service_name.c_str());
```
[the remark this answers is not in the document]
lordgamez: Updated in f02f86cd63fd4d08e8a125d469ada637bd7635fe

extensions/azure/processors/PutAzureBlobStorage.cpp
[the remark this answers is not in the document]
lordgamez: Updated in f02f86cd63fd4d08e8a125d469ada637bd7635fe

cmake/BundledAzureSdkCpp.cmake
```diff
@@ -54,7 +53,7 @@ function(use_bundled_libazure SOURCE_DIR BINARY_DIR)
     ExternalProject_Add(
             azure-sdk-cpp-external
             GIT_REPOSITORY "https://github.com/Azure/azure-sdk-for-cpp.git"
-            GIT_TAG "azure-storage-blobs_12.0.0-beta.7"
+            GIT_TAG "azure-storage-files-datalake_12.2.0"
```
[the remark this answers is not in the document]
lordgamez: Updated in f02f86cd63fd4d08e8a125d469ada637bd7635fe

extensions/azure/processors/PutAzureDataLakeStorage.cpp
lordgamez: Updated in f02f86cd63fd4d08e8a125d469ada637bd7635fe

extensions/azure/processors/AzureStorageProcessorBase.h
lordgamez: Updated in f02f86cd63fd4d08e8a125d469ada637bd7635fe

extensions/azure/processors/PutAzureDataLakeStorage.cpp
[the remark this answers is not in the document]
lordgamez: Added lock and comment in latest commit.

extensions/azure/processors/PutAzureBlobStorage.cpp
lordgamez: Updated in f02f86c

extensions/azure/processors/PutAzureDataLakeStorage.cpp
lordgamez: Added check for read size in c569c2d49a1aacfc4bbf84f88da5a7588a203324

extensions/azure/processors/PutAzureDataLakeStorage.cpp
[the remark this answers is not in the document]
szaszm: Okay, it looks like I didn't fully understand this part of the code. Thanks for the clarification.

libminifi/test/azure-tests/PutAzureDataLakeStorageTests.cpp
```diff
@@ -0,0 +1,271 @@
...
+    LogTestController::getInstance().setDebug<processors::UpdateAttribute>();
+    LogTestController::getInstance().setDebug<processors::LogAttribute>();
+    LogTestController::getInstance().setTrace<minifi::azure::processors::PutAzureDataLakeStorage>();
+
+    // Build MiNiFi processing graph
+    plan_ = test_controller_.createPlan();
+    auto mock_data_lake_storage_client = std::make_unique<MockDataLakeStorageClient>();
+    mock_data_lake_storage_client_ptr_ = mock_data_lake_storage_client.get();
+    put_azure_data_lake_storage_ = std::shared_ptr<minifi::azure::processors::PutAzureDataLakeStorage>(
+      new minifi::azure::processors::PutAzureDataLakeStorage("PutAzureDataLakeStorage", utils::Identifier(), std::move(mock_data_lake_storage_client)));
```
szaszm, line 105: Use `std::make_shared`

lordgamez: `std::make_shared` does not work in this case as the constructor is private and `std::make_shared` does not have permission to call it.

extensions/azure/processors/PutAzureDataLakeStorage.cpp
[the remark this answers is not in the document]
lordgamez: I tried to minimize the lock to the code block where the concurrency could occur which is currently the Azure storage calls where the parameters could change and the client has to be reset. I suppose later if the allowed threads feature is implemented the scope will be widened to the whole onTrigger, but is there any benefit for that currently?

extensions/azure/storage/DataLakeStorageClient.h
```diff
@@ -0,0 +1,41 @@
...
+  std::string file_system_name;
+  std::string directory_name;
+  std::string filename;
+  bool replace_file = false;
+};
+
+class DataLakeStorageClient {
+ public:
+  virtual bool createFile(const PutAzureDataLakeStorageParameters& params) = 0;
+  virtual std::string uploadFile(const PutAzureDataLakeStorageParameters& params, const uint8_t* buffer, std::size_t buffer_size) = 0;
```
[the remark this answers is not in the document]
lordgamez: Good idea, added in faaa33597ae6e8e0649079b3b33aea5cb3ad2d53

extensions/azure/processors/PutAzureDataLakeStorage.cpp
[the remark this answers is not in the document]
szaszm: I wasn't sure if session->get is thread-safe, but I guess it must be as it's used in processors that support concurrency.

szaszm (approved):

adamdebreceni (approved):

PR #1163, opened 2021-08-18
MINIFICPP-1545 - MQTT keeps rebuilding after clean incremental build

adam-markovics:

### For all changes:
     in the commit message?

### For code changes:

### Note:


cmake/PahoMqttC.cmake
```diff
@@ -0,0 +1,58 @@
...
+if (OPENSSL_OFF)
+    set(PAHO_WITH_SSL OFF CACHE BOOL "" FORCE)
+else()
+    set(PAHO_WITH_SSL ON CACHE BOOL "" FORCE)
+endif()
+
+FetchContent_Declare(
+    paho.mqtt.c-external
+    GIT_REPOSITORY "https://github.com/eclipse/paho.mqtt.c.git"
+    GIT_TAG "3b7ae6348bc917d42c04efa962e4868c09bbde9f" # Once tagged as v1.3.9 release, but tags may move
```
szaszm, line 34: It would be better to download the release tarball. Download + extract is faster than git clone.

cmake/PahoMqttC.cmake
```diff
@@ -0,0 +1,58 @@
...
+)
+
+FetchContent_MakeAvailable(paho.mqtt.c-external)
+
+# Set dependencies and target to link to
+if (NOT OPENSSL_OFF)
+    # workaround
```
szaszm, line 41: What issue are you working around here?

cmake/BundledPahoMqttC.cmake
```diff
@@ -1,78 +0,0 @@
...
-    else()
-        include(GNUInstallDirs)
-        set(BYPRODUCT "${CMAKE_INSTALL_LIBDIR}/libpaho-mqtt3cs.a")
-    endif()
-
-    # Set build options
-    set(PAHOMQTTC_CMAKE_ARGS ${PASSTHROUGH_CMAKE_ARGS}
-            "-DCMAKE_INSTALL_PREFIX=${BINARY_DIR}/thirdparty/paho.mqtt.c-install"
-            -DPAHO_BUILD_STATIC=TRUE
-            -DPAHO_BUILD_SHARED=FALSE
```
szaszm, line 35: Any way to build a static library instead of shared?

cmake/PahoMqttC.cmake
```diff
@@ -0,0 +1,58 @@
...
+    if(TARGET paho-mqtt3cs)
+        # this is selected for the first configure run
+        add_library(paho.mqtt.c ALIAS paho-mqtt3cs)
+    else()
+        # this is selected for further configure runs
+        add_library(paho.mqtt.c ALIAS paho-mqtt3cs-static)
```
szaszm, line 47: Are we building both static and shared libraries now? Previously it was static only.

cmake/BundledPahoMqttC.cmake
adam-markovics: We are building static, not shared, the two lines above say exactly that.

szaszm: But these are removed lines, and the new replacement code doesn't say that.

cmake/PahoMqttC.cmake
adam-markovics: It was still static only anyways, paho-mqtt3cs is another name for the static library, that is created on first configure runs. Static library has a different output name than the target.

cmake/PahoMqttC.cmake
adam-markovics: In a phase during development, it was an issue, but it is not present anymore, so I simplified this. The issue was that for first configure paho-mqtt3c had to be linked, while later it was paho-mqtt3c-static, both referring to the same. Static library has a different output name than the target.

cmake/PahoMqttC.cmake
adam-markovics: Done.

cmake/BundledPahoMqttC.cmake
adam-markovics: The replacement code is this:
`set(PAHO_BUILD_STATIC ON CACHE BOOL "" FORCE)
set(PAHO_BUILD_SHARED OFF CACHE BOOL "" FORCE)`

szaszm (approved):

lordgamez (approved):

arpadboda:
What's the reason of removing the patch? We don't need it anymore?

adam-markovics:
> What's the reason of removing the patch? We don't need it anymore?

Yes, that's the reason. Patch step is also removed.

arpadboda (approved):

martinzink (approved):

PR #1164, opened 2021-08-23
MINIFICPP-1633 Add ccache to docker CI builds

lordgamez:
After checking the docker image build times of the same code in CI the time of the build went from 43m to 5m using the ccache. 

https://github.com/lordgamez/nifi-minifi-cpp/actions/runs/1158614270 - 43m build time
https://github.com/lordgamez/nifi-minifi-cpp/actions/runs/1158939622 - 5m build time

Unfortunately no viable way is present to use non-readonly bind mounts for docker builds, so the ccache has to be sent with the build context to the docker daemon if we want to have it available for the time of the build.

https://issues.apache.org/jira/browse/MINIFICPP-1633

### For all changes:
     in the commit message?

### For code changes:

### Note:


docker/Dockerfile
```diff
@@ -98,11 +99,15 @@ ENV MINIFI_VERSION ${MINIFI_VERSION}
 
 # Setup minificpp user
 RUN addgroup -g ${GID} ${USER} && adduser -u ${UID} -D -G ${USER} -g "" ${USER} && \
-    install -d -o ${USER} -g ${USER} ${MINIFI_BASE_DIR}
+    install -d -o ${USER} -g ${USER} ${MINIFI_BASE_DIR} && \
+    ln -s $(which ccache) /usr/local/bin/gcc && \
+    ln -s $(which ccache) /usr/local/bin/g++
```
szaszm, line 104: This shouldn't be needed. Our CMakeLists.txt (line 141) is looking for ccache and using it if available, even if the compiler binaries are not symlinks to ccache itself.

docker/DockerBuild.sh
```diff
@@ -133,14 +136,28 @@ if [ -n "${BUILD_NUMBER}" ]; then
...
+  DOCKER_COMMAND="${DOCKER_BUILD_START} ${BUILD_ARGS} \
+                -f ${DOCKERFILE} \
+                --target build \
+                -t \
+                minifi_build .."
+  echo "Build image Docker Command: 'DOCKER_BUILDKIT=1 ${DOCKER_COMMAND}'"
+  DOCKER_BUILDKIT=1 ${DOCKER_COMMAND}
+  container_id=$(docker run --rm -d  minifi_build sh -c "while true; do sleep 1; done")
+  mkdir -p "${DOCKER_CCACHE_DUMP_LOCATION}"
+  docker cp "${container_id}:/home/minificpp/.ccache/." "${DOCKER_CCACHE_DUMP_LOCATION}"
```
szaszm, line 156: Shouldn't we volume mount the host $HOME/.ccache to the container instead? We could avoid the sleep hack and the copy this way.

lordgamez: Unfortunately there is no way to mount a host directory in build time (it's due to a principle that it would ruin the reproducibility of the image). The only way that would be possible is to create a build image, and run the build commands in the container while mounting the host ccache. But that would also mean we could not use multi-stage builds, so the built binaries would have to be copied out manually and added to the final image separately.

lordgamez: The problem with mounting is that is causes permission errors, because the userid of the host differ from the userid used in the container and if we mount the directory the files are copied to mounted directory with the userid and groupid of the container user (and we would also have to change the permissions of the directory to be able to copy there). With `docker cp` we don't have this problem.

docker/Dockerfile
lordgamez: Updated in 34bc56c2cbef5932e2025ea09eec6f5b906af595

szaszm (approved):

arpadboda (approved):

fgerlits (approved):

PR #1165, opened 2021-09-02
MINIFICPP-1634 remove namespace aliases from Core.h

szaszm:
and adapt code that depends on them by using local aliases or
sufficiently qualified names.

The change started from an effort to move implementations from headers
to source files, so some unrelated refactoring is included, at least
until someone objects.

Long PR, but should be easy to review due to mostly repetitive trivial changes with localized effects only.

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/controllers/UpdatePolicyControllerService.h
+++ b/libminifi/include/controllers/UpdatePolicyControllerService.h
@@ -76,7 +76,7 @@ class UpdatePolicyControllerService : public core::controller::ControllerService
 
  private:
   bool persist_updates_ = false;
-  std::unique_ptr<state::UpdatePolicy> policy_ = std::make_unique<state::UpdatePolicy(false);
+  std::unique_ptr<state::UpdatePolicy> policy_ = std::make_unique<state::UpdatePolicy>(false);
   std::shared_ptr<core::logging::Logger> logger_ = core::logging::LoggerFactory<UpdatePolicyControllerService>::getLogger();
 };
 


libminifi/src/utils/crypto/EncryptionManager.cpp
```diff
@@ -32,11 +32,7 @@ namespace minifi {
 namespace utils {
 namespace crypto {
 
-#ifdef WIN32
-constexpr const char* DEFAULT_NIFI_BOOTSTRAP_FILE = "\\conf\\bootstrap.conf";
```
fgerlits, line 36: Why did you change this?  If we want to change it, should DEFAULT_NIFI_PROPERTIES_FILE be changed similarly, too?  Either way, I don't think this change should be part of this pull request.

libminifi/include/utils/file/FileManager.h
```diff
@@ -17,18 +17,10 @@
 #ifndef LIBMINIFI_INCLUDE_UTILS_FILE_FILEMANAGER_H_
 #define LIBMINIFI_INCLUDE_UTILS_FILE_FILEMANAGER_H_
 
+#include <filesystem>
```
fgerlits, line 20: I think these `std::filesystem`-related changes should be in a separate pull request.

fgerlits (commented):
Full disclosure: I mostly just scrolled through this, and only looked more closely at a few files selected at random, but with this method, the changes looked good. :)

Thanks for the cleanup!

libminifi/src/utils/crypto/EncryptionManager.cpp
szaszm: This was because I changed concat_path to use std::filesystem::path /=, and concatenating "\conf\bootstrap.conf" (an absolute path) to anything results in this second part overwriting the path of the first. This is indeed a behavior change that would ideally not be part of this PR. The reason it is is that I started with this change in the same working directory and I wanted to avoid separating the changes if I could get away leaving them as they are.

szaszm: I would still prefer not separating them, so unless you insist, I would prefer to have these two independent changes reviewed together. Of course this only works if all changes are acceptable.

fgerlits: I definitely don't want `DEFAULT_NIFI_PROPERTIES_FILE == "\\conf\\minifi.properties"` and `DEFAULT_NIFI_BOOTSTRAP_FILE == "./conf/bootstrap.conf"` on Windows, so if one of them changes, the other should, too.

I would prefer this change to be part of a separate, `std::filesystem`-related PR, but if it is too much work to separate them, I am going to approve it is at is (after the properties file change).

extensions/standard-processors/processors/RouteOnAttribute.h
```diff
@@ -39,9 +39,8 @@ namespace processors {
 
 class RouteOnAttribute : public core::Processor {
  public:
-  RouteOnAttribute(const std::string& name, const utils::Identifier& uuid = {}) // NOLINT
-      : core::Processor(name, uuid),
-        logger_(logging::LoggerFactory<RouteOnAttribute>::getLogger()) {
+  explicit RouteOnAttribute(const std::string& name, const utils::Identifier& uuid = {}) // NOLINT
```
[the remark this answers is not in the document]
szaszm: It's no longer needed, I've removed it during the latest rebase

libminifi/src/utils/crypto/EncryptionManager.cpp
szaszm: I've removed all filesystem-related changes in the latest rebase

PR #1170, opened 2021-09-08
MINIFICPP-1618 Create the ReplaceText processor

fgerlits:
Processor based on https://nifi.apache.org/docs/nifi-docs/components/org.apache.nifi/nifi-standard-nar/1.14.0/org.apache.nifi.processors.standard.ReplaceText/index.html

---

### For all changes:
     in the commit message?

### For code changes:

### Note:


docker/test/integration/MiNiFi_integration_test_driver.py
```diff
@@ -104,7 +104,7 @@ def generate_input_port_for_remote_process_group(remote_process_group, name):
         return input_port_node
 
     def add_test_data(self, path, test_data, file_name=str(uuid.uuid4())):
-        self.docker_directory_bindings.put_file_to_docker_path(self.test_id, path, file_name, test_data.encode('utf-8'))
+        self.docker_directory_bindings.put_file_to_docker_path(self.test_id, path, file_name, test_data.replace('\\n', '\n').encode('utf-8'))
```
adamdebreceni, line 107: note: I have also touched this part for the same reason (use escaped `\t`, `\n`, ... chars) in [1168](https://github.com/apache/nifi-minifi-cpp/pull/1168) (same with the file content validation)

extensions/standard-processors/processors/ReplaceText.h
```diff
@@ -0,0 +1,102 @@
...
+  std::string search_value_;
+  std::regex search_regex_;
+  std::string replacement_value_;
+  uint64_t maximum_buffer_size_ = 0;
+  EvaluationModeType evaluation_mode_ = EvaluationModeType::LINE_BY_LINE;
+  LineByLineEvaluationModeType line_by_line_evaluation_mode_ = LineByLineEvaluationModeType::ALL;
+  ReplacementStrategyType replacement_strategy_ = ReplacementStrategyType::REGEX_REPLACE;
+};
+
+REGISTER_RESOURCE(ReplaceText, "Updates the content of a FlowFile by replacing parts of it using various replacement strategies.");
```
adamdebreceni, line 100: note: after the dynamic PR, resource registration should happen in cpp files

libminifi/src/utils/LineByLineInputOutputStreamCallback.cpp
```diff
@@ -0,0 +1,129 @@
...
+  }
+  current_line_ = next_line_;
+  next_line_ = input_line;
+  total_bytes_read_ += current_read;
+  return gsl::narrow<int64_t>(current_read);
+}
+
+size_t LineByLineInputOutputStreamCallback::readLine(io::InputStream& stream, std::string& line) {
+  if (current_pos_ == end_pos_) {
+    const auto current_bytes_read = fillBuffer(stream);
```
adamdebreceni, line 72: the way it is implemented in RouteText is to check if the input stream supports access to its underlying buffer (BufferStream and the default RocksDbStream do) and only read into a buffer as a fallback

libminifi/src/utils/LineByLineInputOutputStreamCallback.cpp
```diff
@@ -0,0 +1,129 @@
...
+    ++pos;
+  }
+  if (pos < end_pos_ && buffer_[pos] == '\n') {
+    ++pos;
+    line = start_of_line + std::string{buffer_.begin() + current_pos_, buffer_.begin() + pos};
+    current_pos_ = pos;
+    return line.size();
+  }
+
+  logger_->log_error("Found a line longer than the max buffer size (%zu bytes)", buffer_.size());
```
adamdebreceni, line 110: it seems we might process longer than `buffer size` lines, as we might append the end of the previous buffer fill

docker/test/integration/MiNiFi_integration_test_driver.py
fgerlits: OK, I'll revert these bits if your PR is merged first.

extensions/standard-processors/processors/ReplaceText.h
fgerlits: OK, I'll move it after #1138 is merged.

libminifi/src/utils/LineByLineInputOutputStreamCallback.cpp
fgerlits: thanks, I'll take a look

libminifi/src/utils/LineByLineInputOutputStreamCallback.cpp
fgerlits: That's correct, lines up to 2 times the buffer size may be accepted, but only lines <= the buffer size are guaranteed to be accepted.  Do you think that is a problem?  Or do you think we need to make the error message more verbose?

libminifi/src/utils/LineByLineInputOutputStreamCallback.cpp
```diff
@@ -0,0 +1,129 @@
...
+  const auto current_bytes_read = fillBuffer(stream);
+  if (io::isError(current_bytes_read)) {
+    return io::STREAM_ERROR;
+  } else if (current_bytes_read == 0) {
+    line = start_of_line;
+    return line.size();
+  }
+
+  pos = current_pos_;
+  while (pos < end_pos_ && buffer_[pos] != '\n') {
```
adamdebreceni, line 101: unfortunately it seems nifi has a different concept of "line", it  uses the `LineDemarcator` which

1. considers all `\n` as end-of-line
2. considers all `\r` followed by a non-`\n` as end-of-line
3. considers a non-empty (!) tail after the last end-of-line as a line

(it's quite confusing) 

this was implemented in RouteText to be in-line with nifi, but of course we could decide to change them to use a different segmentation logic

fgerlits: Just speculating, but that may have been done in NiFi to deal with the old MacOS line endings of bare `\r`.  Since these line endings are not used by any computer in use today, I think using `\n`, which covers Windows, Linux and MacOS X+, is fine.

libminifi/src/utils/LineByLineInputOutputStreamCallback.cpp
adamdebreceni: I guess I was just making an observation aloud :)

libminifi/src/utils/LineByLineInputOutputStreamCallback.cpp
fgerlits: (3) is handled correctly (the `\n`-less last line is treated as a line, is processed and output without a `\n`), **except** in the edge case where the last line doesn't end in a `\n` **and** it straddles a buffer boundary, which causes an error (flow file routed to failure).  Thanks for pointing this out, I'll fix it.

fgerlits: the bug in (3) is fixed by https://github.com/apache/nifi-minifi-cpp/pull/1170/commits/4df70bdec67d7f91182e10b01e6df646acf5e85a

extensions/standard-processors/processors/ReplaceText.cpp
```diff
@@ -0,0 +1,359 @@
...
+      return;
+  }
+
+  throw Exception{PROCESSOR_EXCEPTION, utils::StringUtils::join_pack("Unsupported ", EvaluationMode.getName(), ": ", evaluation_mode_.toString())};
+}
+
+void ReplaceText::readSearchValueProperty(const std::shared_ptr<core::ProcessContext>& context, const std::shared_ptr<core::FlowFile>& flow_file) {
+  bool found_search_value;
+  if (replacement_strategy_ == ReplacementStrategyType::REGEX_REPLACE) {
+    found_search_value = context->getProperty(SearchValue.getName(), search_value_);
```
adamdebreceni, line 159: won't writing members (`search_value_`, `search_regex_`) interferes with other threads of this same processor?

extensions/standard-processors/processors/ReplaceText.cpp
```diff
@@ -0,0 +1,359 @@
...
+      std::copy(it, input.end(), std::back_inserter(output));
+      it = input.end();
+    }
+  } while (it != input.end());
+
+  return std::string{output.begin(), output.end()};
+}
+
+std::string ReplaceText::applySubstituteVariables(const std::string& input, const std::shared_ptr<core::FlowFile>& flow_file) const {
+  static const std::regex PLACEHOLDER{R"(\$\{([^}]+)\})"};
```
adamdebreceni, line 322: I see that nifi also uses a hand-rolled variable substitution here, but the attribute substitution already happened when we queried the property and the ExpressionLanguage is enabled, right? it seems to me that standard processors have the EL extension as dependency, so we might be able to rely on that here as well

extensions/standard-processors/processors/ReplaceText.cpp
fgerlits: Good point, `onTrigger()` was not thread-safe.  Fixed in https://github.com/apache/nifi-minifi-cpp/pull/1170/commits/86eb3f776005301a6507ca7b27e5d28736d5923a.

extensions/standard-processors/processors/ReplaceText.cpp
fgerlits: That was my original plan, too, but the expression language substitution logic is hidden inside the expression-language extension, and I didn't think exposing it should be part of this PR.  Also, full expression language support was not a requirement.  We could do this later in a separate PR.

libminifi/src/utils/LineByLineInputOutputStreamCallback.cpp
fgerlits: Does this mean that if we use RocksDB (ie practically always), then the whole flow file is always loaded in memory?  I wanted to avoid that in line-by-line mode, but I may have been worrying about a non-existent case.

Do you think there is value in keeping this implementation which supports large flow files being read one (potentially small) chunk at a time?  If not, we should probably remove the whole max buffer size property, which would make things simpler.

extensions/standard-processors/processors/ReplaceText.h
fgerlits: done

extensions/standard-processors/processors/ReplaceText.cpp
```diff
@@ -0,0 +1,342 @@
...
+                                                         "This includes both FlowFiles that had text replaced and those that did not.");
+const core::Relationship ReplaceText::Failure("failure", "FlowFiles that could not be updated are routed to this relationship.");
+
+ReplaceText::ReplaceText(const std::string& name, const utils::Identifier& uuid)
+  : core::Processor(name, uuid),
+    logger_(logging::LoggerFactory<ReplaceText>::getLogger()) {
+}
+
+core::annotation::Input ReplaceText::getInputRequirement() const {
+  return core::annotation::Input::INPUT_REQUIRED;
```
adamdebreceni, line 83: all other `getInputRequirement` overrides are implemented in the headers, which I kind of agree with, as its implementation (the return value) is something inherent to the processor class (like the relationships and properties)

fgerlits: yup, makes sense

done in https://github.com/apache/nifi-minifi-cpp/pull/1170/commits/fead017bb2845854247d04c64d51882b273aba5a

libminifi/include/io/StreamPipe.h
```diff
@@ -42,6 +42,11 @@ class OutputStreamCallback {
   virtual ~OutputStreamCallback() = default;
   virtual int64_t process(const std::shared_ptr<io::BaseStream>& stream) = 0;
 };
+class InputOutputStreamCallback {
+ public:
+  virtual ~InputOutputStreamCallback() = default;
+  virtual int64_t process(const std::shared_ptr<io::BaseStream>& input, const std::shared_ptr<io::BaseStream>& output) = 0;
+};
```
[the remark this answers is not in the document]
fgerlits: Is this what you mean: https://github.com/fgerlits/nifi-minifi-cpp/commit/cdfbca12e6e2a49f93f286f411cecd91a0bce1bf?  I will do it if you explain why it would be better.

extensions/standard-processors/processors/ReplaceText.cpp
```diff
@@ -0,0 +1,339 @@
...
+  }
+  if (found_replacement_value) {
+    logger_->log_debug("the %s property is set to %s", ReplacementValue.getName(), parameters.replacement_value_);
+  } else {
+    throw Exception{PROCESSOR_EXCEPTION, utils::StringUtils::join_pack("Missing required property: ", ReplacementValue.getName())};
+  }
+
+  if (evaluation_mode_ == EvaluationModeType::LINE_BY_LINE) {
+    const auto [chomped_value, line_ending] = utils::StringUtils::chomp(parameters.replacement_value_);
+    parameters.replacement_value_ = chomped_value;
```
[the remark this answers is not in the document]
fgerlits: done in https://github.com/apache/nifi-minifi-cpp/pull/1170/commits/ace5e48f1697ad97efb54693b2536bf06664917a

extensions/standard-processors/processors/ReplaceText.cpp
```diff
@@ -0,0 +1,339 @@
...
+    found_search_value = context->getProperty(SearchValue.getName(), parameters.search_value_);
+  } else {
+    found_search_value = context->getProperty(SearchValue, parameters.search_value_, flow_file);
+  }
+  if (found_search_value) {
+    logger_->log_debug("the %s property is set to %s", SearchValue.getName(), parameters.search_value_);
+    if (replacement_strategy_ == ReplacementStrategyType::REGEX_REPLACE) {
+      parameters.search_regex_ = std::regex{parameters.search_value_};
+    }
+  }
```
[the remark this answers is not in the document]
fgerlits: That would complicate the code considerably, as we would need to read the Search Value either in `onSchedule` or in `onTrigger`, depending on whether we are using Regex Replace or not.

I have run this in Google Benchmark:
```c++
#include <benchmark/benchmark.h>
#include <regex>
#include <string>

static void SimpleRegex(benchmark::State& state) {
  std::string input = ".*foo?";
  for (auto _ : state) {
    std::regex my_regex{input};
    benchmark::DoNotOptimize(my_regex);
  }
}
BENCHMARK(SimpleRegex);

static void ComplicatedRegex(benchmark::State& state) {
  std::string input = "(.)[ompli]{5}\\1[ae]t[ae]d";
  for (auto _ : state) {
    std::regex my_regex{input};
    benchmark::DoNotOptimize(my_regex);
  }
}
BENCHMARK(ComplicatedRegex);

BENCHMARK_MAIN();
```
and the simple regex took 0.6 microseconds to compile, the complicated one took 77 microseconds (on gcc 11.1 with `-O3`).  I think that is an acceptable price for simpler code.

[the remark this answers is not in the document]
szaszm: That's fair, I expected them to be in the magnitude of tens of milliseconds. Thanks for doing the measurement.

libminifi/include/io/StreamPipe.h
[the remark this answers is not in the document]
fgerlits: As discussed, let's keep it as it is for now, and possibly change all three (read, write, readWrite) functions together in a separate PR.

libminifi/src/utils/LineByLineInputOutputStreamCallback.cpp
```diff
@@ -0,0 +1,72 @@
...
+    return status;
+  }
+
+  std::size_t total_bytes_written_ = 0;
+  bool is_first_line = true;
+  readLine();
+  do {
+    readLine();
+    std::string output_line = callback_(*current_line_, is_first_line, isLastLine());
+    output->write(reinterpret_cast<const uint8_t *>(output_line.data()), output_line.size());
```
adamdebreceni, line 42: should we check the return value of `write` here?

libminifi/include/utils/StringUtils.h
```diff
@@ -75,7 +75,12 @@ class StringUtils {
 
   static std::string toLower(std::string_view str);
 
-  // Trim String utils
+  /**
+   * Strips the line ending (\n or \r\n) from the end of the input line.
+   * @param input_line
+   * @return (stripped line, line ending) pair
+   */
+  static std::pair<std::string, std::string> chomp(const std::string& input_line);
```
adamdebreceni, line 83: we seem to migrate our StringUtils to operate on `std::string_view` wherever possible, we might benefit from writing the new ones on the same note

libminifi/src/utils/LineByLineInputOutputStreamCallback.cpp
fgerlits: Good catch, thanks!  Fixed in https://github.com/apache/nifi-minifi-cpp/pull/1170/commits/9a5b33d91cca7316a42b4b4c26498de622833742

libminifi/include/utils/StringUtils.h
fgerlits: That would require modifying `endsWith()` (and, for consistency, `endsWithIgnoreCase()` and `startsWith()`), which would introduce more conflicts with #1168.  If #1168 is merged before this, and it includes the change to `endsWith()`, then I will change `chomp()` similarly.

extensions/standard-processors/processors/ReplaceText.cpp
```diff
@@ -0,0 +1,339 @@
...
+      return parameters.replacement_value_ + line_ending;
+
+    case ReplacementStrategyType::SUBSTITUTE_VARIABLES:
+      return applySubstituteVariables(chomped_input, flow_file) + line_ending;
+  }
+
+  throw Exception{PROCESSOR_EXCEPTION, utils::StringUtils::join_pack("Unsupported ", ReplacementStrategy.getName(), ": ", replacement_strategy_.toString())};
+}
+
+std::string ReplaceText::applyLiteralReplace(const std::string& input, const Parameters& parameters) const {
```
szaszm, line 278: `applyLiteralReplace` could be `static`, or a free utility function in an anonymous namespace.

fgerlits: Yes it could, but should it?  We would be marking this function as different from those around it, when the difference is only "this function doesn't currently log anything", which is a minor implementation detail.  The only advantage I can see of adding "static" would be to get rid of the clang-tidy warning, but I don't think we should let clang-tidy push us around like that. :)

szaszm: I see this as an unnecessary dependency just for subjective aesthetics. I also don't want to block the PR because of this minor detail if you disagree.

I like following clang-tidy suggestions unless I have a good argument against it.

fgerlits: I think pretty patterns are useful, too, as they improve readability.  Also, I suspect it won't take long before cpplint and clang-tidy disagree about something and then we'll be stuck.  But okay: https://github.com/apache/nifi-minifi-cpp/pull/1170/commits/4b1476c18104f4fa9de3994c9b366c284177e779.

szaszm: I don't want to force my opinion on you, so feel free to revert the change and close this thread if I didn't convince you. I think minor issues like this are well within the author's discretion.

szaszm (approved):

PR #1171, opened 2021-09-10
MINIFICPP-1638 Optimize catch and cmake usage

szaszm:
- removed a few unnecessary `add_dependencies` calls in cmake: add_dependencies(A B) will wait for B to be built before starting to build A. In most cases, we only want to link them together, but we can build them in parallel.
- reduced the number of catch2 main instances (TUs where CATCH_CONFIG_MAIN is defined). Compiling such a TU takes about 30-40sec. We already have a target with a catch2 main that we can link to test binaries, so I changed a few of them to link instead of instantiating another main. There were also cases where the tests were both linked to catch2 main and had CATCH_CONFIG_MAIN themselves, so I removed the unnecessary CATCH_CONFIG_MAIN definitions.
- Opted out of some catch2 features to reduce compilation time: https://github.com/catchorg/Catch2/blob/v2.13.6/docs/configuration.md#other-toggles

rough measurements of `make -j64` with a wide range of extensions enabled:

main:   
3:54
real    3m54.001s
user    127m7.528s
sys     6m44.876s

this PR: 
2:57  (20% faster)
real    2m57.236s
user    128m48.359s
sys     6m50.702s

this PR + #1165:
2:48  (30% faster)
real    2m48.448s
user    107m30.460s
sys     6m0.179s

Less parallel builds might benefit less from the changes, especially removal of add_dependencies, that only affected parallelization opportunities.

---

### For all changes:
     in the commit message?

### For code changes:

### Note:


adam-markovics:
A very welcome change!

fgerlits:
I have tried building this, and on my computer (Intel i7-8850H CPU @ 2.60GHz, 6 cores x2 hyperthreading), the new code takes longer to compile.  I have no idea why.

This is what I did:
- check out the code
- remove the build directory, create a new build directory, `cd build`
- `cmake -DSTRICT_GSL_CHECKS=AUDIT -DCMAKE_BUILD_TYPE=Debug .. && make -j12`
- `time make -j12 clean all`
- `ccache -C; time make -j12 clean all`

On the main branch, the second (cached) build took 3:42, the third (uncached) one 10:45.
On the PR branch, the same numbers were 4:04 and 10:48.

szaszm:
I did some more measurements in the same environment, but this time I was building with `-j12` and using your cmake command line.

My environment: 
- Gentoo amd64 
- system-wide -O3 and LTO, including GCC
- AMD Ryzen Threadripper 3970X @ 3.7GHz, 32 cores/64 threads

One modification I made was `ccache -C` between the cmake call and the first unmeasured cache warmup build.

The new results:
```
compile-time branch:
warm ccache: 1:53 1:36 1:37 (median -11%)
cold ccache: 4:26 4:34 4:30 (median -7%)

main branch:
warm ccache: 1:49 1:51 1:48
cold ccache: 4:51 4:52 4:49
```

There is some variability in the results, and I think your numbers are affected by that. I can do some more measurements on a laptop later.

fgerlits:
Yes, I agree on the variability.  My results don't really show that the compilation on my computer on the compile-time branch is slower, but it does show that it is roughly the same as on the main branch.  I will do some more measurements, too, and if the code change speeds up the compilation for other people, and doesn't slow it down for me, then I'm fine with it.  It would be good if others could run some measurements, too.

szaszm:
I measured on a laptop. The tests didn't show a significant decrease in compilation times there, so it looks like your conclusion was not just random noise.

My numbers on a 6c/12t laptop:
```
main branch:
warm ccache: 1:26 1:26 1:25
cold ccache: 6:57 6:55 6:57

compile-time branch:
warm ccache: 1:25 1:25 1:25
cold ccache: 6:52 6:54 6:52
```

I think it's still worth pursuing these changes, but the compilation time decrease is not as clear cut as I initially expected.

fgerlits (approved):

lordgamez (approved):

PR #1172, opened 2021-09-13
MQTT tests updated

adam-markovics:

### For all changes:
     in the commit message?

### For code changes:

### Note:


docker/test/integration/features/mqtt.feature
```diff
@@ -0,0 +1,66 @@
+Feature: Sending data to using MQTT streaming platform using PublishMQTT
```
lordgamez, line 1: The word 'using' is not needed here

docker/test/integration/minifi/core/DockerTestCluster.py
```diff
@@ -54,13 +55,16 @@ def get_app_log(self, container_name):
 
         return container.status, None
 
-    def wait_for_app_logs(self, container_name, log_entry, timeout_seconds, count=1):
+    def wait_for_app_logs(self, container_name, log_entry, timeout_seconds, count=1, use_regex=False):
```
lordgamez, line 58: I would prefer having a separate function for log matching and regex matching, the surrounding code could may be extracted as a decorator.

docker/test/integration/minifi/core/ImageStore.py
```diff
@@ -23,21 +24,28 @@ def cleanup(self):
             self.client.images.remove(image.id, force=True)
 
     def get_image(self, container_engine):
+        if container_engine in self.images:
```
lordgamez, line 27: Good improvement. With this change I think we should also rename the `__get_.*_image` functions to `__build_.*image`

docker/test/integration/minifi/core/MqttBrokerContainer.py
```diff
@@ -0,0 +1,23 @@
+import logging
+from .Container import Container
+
+
+class MqttBrokerContainer(Container):
+    def __init__(self, name, vols, network, image_store):
+        super().__init__(name, 'mqtt-broker', vols, network, image_store)
+
+    def get_startup_finished_log_entry(self):
+        return "mosquitto version 2.0.11 running"
```
lordgamez, line 10: Could this be changed somehow not to be version dependent?

docker/test/integration/resources/mqtt_broker/mosquitto_test.conf
```diff
@@ -0,0 +1,874 @@
+# Config file for mosquitto
```
lordgamez, line 1: There seems to be only 3 lines not commented out in this configuration. Could we only leave those in this test config? I think it would be more clear at glance what is actually set.

docker/test/integration/minifi/core/MqttBrokerContainer.py
adam-markovics: With regex matching I think. Would you second that?
Currently the output of starting mosquitto in verbose:
```
adam@amarkovics-Linux:~/work/minifi-cpp/build$ mosquitto -v
1632119285: mosquitto version 2.0.12 starting
1632119285: Using default config.
1632119285: Starting in local only mode. Connections will only be possible from clients running on this machine.
1632119285: Create a configuration file which defines a listener to allow remote access.
1632119285: For more details see https://mosquitto.org/documentation/authentication-methods/
1632119285: Opening ipv4 listen socket on port 1883.
1632119285: Opening ipv6 listen socket on port 1883.
1632119285: mosquitto version 2.0.12 running
```

docker/test/integration/resources/mqtt_broker/mosquitto_test.conf
adam-markovics: This config is a slightly modified version of the base image, eclipse-mosquitto:2.0.11. Maybe I could apply some small diff on that instead in the Dockerfile in a RUN section.

lordgamez: That should be okay as well, or a `sed` command if that's a very trivial change.

docker/test/integration/minifi/core/MqttBrokerContainer.py
lordgamez: If that's not a huge effort to implement regex check here, that would okay.

docker/test/integration/features/mqtt.feature
```diff
@@ -0,0 +1,66 @@
...
+    And the "success" relationship of the GetFile processor is connected to the PublishMQTT
+    And the "success" relationship of the PublishMQTT processor is connected to the PutFile
+
+    And an MQTT broker is set up in correspondence with the PublishMQTT
+
+    When both instances start up
+    Then a flowfile with the content "test" is placed in the monitored directory in less than 60 seconds
+    And the MQTT broker has 1 log line matching "Received PUBLISH from .*testtopic.*\\(4 bytes\\)"
+
+  Scenario: A MiNiFi instance tries to transfer data to a non-existent MQTT broker
```
fgerlits, line 23: Is the test file not routed to failure because PublishMQTT will throw in `onSchedule()`?  If so, then the name of this scenario could be better, eg. "If the MQTT broker is not available, the PublishMQTT processor will not start".

docker/test/integration/features/mqtt.feature
```diff
@@ -0,0 +1,66 @@
...
+    And a PutFile processor with the "Directory" property set to "/tmp/output"
+    And "ConsumeMQTT" processor is a start node
+    And the "success" relationship of the ConsumeMQTT processor is connected to the PutFile
+
+    And an MQTT broker is set up in correspondence with the PublishMQTT and ConsumeMQTT
+
+    When both instances start up
+    Then a flowfile with the content "test" is placed in the monitored directory in less than 60 seconds
+    And the MQTT broker has 1 log line matching "Received PUBLISH from .*testtopic.*\\(4 bytes\\)"
+    And the MQTT broker has 1 log line matching "Received SUBSCRIBE from"
```
fgerlits, line 66: missing newline at the end

docker/test/integration/features/mqtt.feature
```diff
@@ -0,0 +1,66 @@
...
+    And a file with the content "test" is present in "/tmp/input"
+    And a PublishMQTT processor set up to communicate with an MQTT broker instance
+    And a PutFile processor with the "Directory" property set to "/tmp/output"
+    And the "success" relationship of the GetFile processor is connected to the PublishMQTT
+    And the "success" relationship of the PublishMQTT processor is connected to the PutFile
+
+    When the MiNiFi instance starts up
+
+    Then no files are placed in the monitored directory in 30 seconds of running time
+    And an MQTT broker is deployed in correspondence with the PublishMQTT
```
fgerlits, line 46: moving the blank line may make this test clearer:
```suggestion
    When the MiNiFi instance starts up
    Then no files are placed in the monitored directory in 30 seconds of running time

    And an MQTT broker is deployed in correspondence with the PublishMQTT
```

docker/test/integration/minifi/core/DockerTestCluster.py
```diff
@@ -54,13 +55,16 @@ def get_app_log(self, container_name):
...
         wait_start_time = time.perf_counter()
         while (time.perf_counter() - wait_start_time) < timeout_seconds:
             logging.info('Waiting for app-logs `%s` in container `%s`', log_entry, container_name)
             status, logs = self.get_app_log(container_name)
-            if logs is not None and count <= logs.decode("utf-8").count(log_entry):
-                return True
+            if logs is not None:
+                if not use_regex and count <= logs.decode("utf-8").count(log_entry):
+                    return True
+                elif use_regex and len(re.findall(log_entry, logs.decode("utf-8"))) <= count:
```
fgerlits, line 66: this should be `>=`

(I would also change the non-regex case from `count <= something` to `something >= count` as I think that would read better)

docker/test/integration/minifi/core/ImageStore.py
```diff
@@ -12,6 +12,7 @@ class ImageStore:
     def __init__(self):
         self.client = docker.from_env()
         self.images = dict()
+        self.test_dir = os.environ['PYTHONPATH'].split(':')[-1]  # Based on DockerVerify.sh
```
fgerlits, line 15: This looks brittle, as someone may add another directory to the end of `PYTHONPATH` in the future.  It would be better if eg. DockerVerify.sh exported a dedicated `TEST_DIRECTORY` environment variable which we could read here.

EDIT: I can see this same code is already used in a few other places; I would change those, too.

docker/test/integration/steps/steps.py
```diff
@@ -546,3 +554,14 @@ def step_impl(context, query, number_of_rows, timeout_seconds):
 @then("the Minifi logs contain the following message: \"{log_message}\" in less than {duration}")
 def step_impl(context, log_message, duration):
     context.test.check_minifi_log_contents(log_message, timeparse(duration))
+
+# MQTT
+@then("the MQTT broker has 1 log line matching \"{log_pattern}\"")
```
fgerlits, line 559: this sounds like the test will only pass if there is exactly 1 matching log line, but in fact we check for at least 1 matching log line
```suggestion
@then("the MQTT broker has a log line matching \"{log_pattern}\"")
```

docker/test/integration/features/mqtt.feature
adam-markovics: No, it is because PublishMQTT will do nothing and wait for the broker to appear. Then it will forward it. I'm open to discussion if this is a desired behavior.

docker/test/integration/steps/steps.py
```diff
@@ -556,9 +556,9 @@ def step_impl(context, log_message, duration):
     context.test.check_minifi_log_contents(log_message, timeparse(duration))
 
 # MQTT
```
lordgamez, line 558: This line fails flake8 test locally because 2 blank lines are expected:
`steps.py:559:1: E302 expected 2 blank lines, found 1`

Somehow it does not fail in the CI which should be checked.

adam-markovics: I checked before submitting and nothing failed for me either. Strange. My local version is:
```
adam@amarkovics-Linux:~/work/minifi-cpp/build$ flake8 --version
3.7.9 (mccabe: 0.6.1, pycodestyle: 2.5.0, pyflakes: 2.1.1) CPython 3.8.10 on Linux
```

lordgamez: Must be the version difference (in CI as well) because my local version is 3.8.4

lordgamez (approved):

docker/test/integration/features/mqtt.feature
fgerlits: OK, I get it -- if the processor can't connect to the broker, it won't even pick up any flow files from the queue.  I would change the title of this scenario to better explain this, e.g. "If the MQTT broker does not exist, then no flow files are processed".

This approach could be improved (e.g. with a configurable timeout), but definitely not in this pull request.

PR #1173, opened 2021-09-13
MINIFICPP-1458-Register and test hidden InvokeHTTP properties

aminadinari19:

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/http-curl/tests/unit/InvokeHTTPTests.cpp
```diff
@@ -306,3 +291,97 @@ TEST_CASE("HTTPTestsPostNoResourceClaim", "[httptest1]") {
   REQUIRE(true == LogTestController::getInstance().contains("Exiting because method is POST"));
   LogTestController::getInstance().reset();
 }
+
+TEST_CASE("HTTPTestsPenalizeNoRetry", "[httptest1]") {
+  TestController testController;
+  using processors::InvokeHTTP;
+
+  std::shared_ptr<core::ContentRepository>
+  content_repo = std::make_shared<core::repository::VolatileContentRepository>();
```
szaszm, line 300: This is a single statement. It should be on one line in this case, but in general it can be separated with a new line + continuation indentation (+4 spaces or aligned) if it's too long.

extensions/http-curl/tests/unit/InvokeHTTPTests.cpp
```diff
@@ -306,3 +291,97 @@ TEST_CASE("HTTPTestsPostNoResourceClaim", "[httptest1]") {
...
+  std::shared_ptr<TestRepository> repo = std::make_shared<TestRepository>();
+
+  LogTestController::getInstance().setTrace<org::apache::nifi::minifi::processors::GenerateFlowFile>();
+  LogTestController::getInstance().setTrace<org::apache::nifi::minifi::processors::ListenHTTP>();
+  LogTestController::getInstance().setTrace<org::apache::nifi::minifi::processors::InvokeHTTP>();
+  LogTestController::getInstance().setTrace<minifi::core::ProcessSession>();
+
+  std::shared_ptr<core::Processor>
+  listenhttp = std::make_shared<org::apache::nifi::minifi::processors::ListenHTTP>("listenhttp");
```
szaszm, line 309: Same here

extensions/http-curl/tests/unit/InvokeHTTPTests.cpp
```diff
@@ -306,3 +291,97 @@ TEST_CASE("HTTPTestsPostNoResourceClaim", "[httptest1]") {
...
+  }
+}
+
+TEST_CASE("HTTPTestsPutResponseBodyinAttribute", "[httptest1]") {
+  TestController testController;
+  using processors::InvokeHTTP;
+  std::string url = "http://localhost:8681/testytesttest";
+
+  std::shared_ptr<core::ContentRepository>
+  content_repo = std::make_shared<core::repository::VolatileContentRepository>();
```
szaszm, line 355: same here

extensions/http-curl/tests/unit/InvokeHTTPTests.cpp
```diff
@@ -306,3 +291,97 @@ TEST_CASE("HTTPTestsPostNoResourceClaim", "[httptest1]") {
...
+  std::shared_ptr<TestRepository> repo = std::make_shared<TestRepository>();
+
+  LogTestController::getInstance().setTrace<org::apache::nifi::minifi::processors::GenerateFlowFile>();
+  LogTestController::getInstance().setTrace<org::apache::nifi::minifi::processors::ListenHTTP>();
+  LogTestController::getInstance().setTrace<org::apache::nifi::minifi::processors::InvokeHTTP>();
+  LogTestController::getInstance().setTrace<minifi::core::ProcessSession>();
+
+  std::shared_ptr<core::Processor>
+  listenhttp = std::make_shared<org::apache::nifi::minifi::processors::ListenHTTP> ("listenhttp");
```
szaszm, line 364: same here

docker/test/integration/features/http.feature
```diff
@@ -8,7 +8,6 @@ Feature: Sending data using InvokeHTTP to a receiver using ListenHTTP
 
   Scenario: A MiNiFi instance transfers data to another MiNiFi instance with message body
     Given a GetFile processor with the "Input Directory" property set to "/tmp/input"
-    And the "Keep Source File" property of the GetFile processor is set to "true"
```
szaszm, line 11: Are these removed because the input files are not used anymore and just to keep things simple? Or was there an actual problem due to processing the same file multiple time? Or what is the reason for this change?

extensions/http-curl/processors/InvokeHTTP.cpp
```diff
@@ -443,6 +447,7 @@ void InvokeHTTP::route(const std::shared_ptr<core::FlowFile> &request, const std
   } else {
     if (request != nullptr) {
       if (penalize_no_retry_) {
+        logger_->log_debug("Flowfile has been penalized");
         session->penalize(request);
       }
```
szaszm, line 452: ProcessSession::penalize already produces a log message, so I think this is unnecessary.

extensions/http-curl/tests/unit/InvokeHTTPTests.cpp
aminadinari19: Okay I will change it. For some reason, it keeps happening on its own. Is there any way to make sure it doesn't happen anymore?

docker/test/integration/features/http.feature
aminadinari19: There was no problem but it seemed unnecessary so the change was made to just keep it simple :)

extensions/http-curl/processors/InvokeHTTP.cpp
aminadinari19: This was added to check whether the flowfile is being penalized in case of no retry relationship. Maybe the wording of the log message could be improved as it is used to test for PenalizeOnNoRetry property?

extensions/http-curl/tests/unit/InvokeHTTPTests.cpp
szaszm: Maybe your editor/IDE is enforcing some line length requirements? In CLion, it's under Settings/Editor/Code Style/{Hard Wrap At,Wrap on typing}.

extensions/http-curl/processors/InvokeHTTP.cpp
szaszm: You could check for that log message instead. This is how it's produced: (ProcessSession.cpp:202)
```
  logging::LOG_INFO(logger_) << "Penalizing " << flow->getUUIDStr() << " for " << penalization_period.count() << "ms at " << process_context_->getProcessorNode()->getName();
```

Try using `LogTestController::matchesRegex` to match this message. Something like this (untested) regex might work: `"Penalizing ([0-9a-f]+-)+[0-9a-f]+ for [0-9]+ms at invokehttp"`

docker/test/integration/features/http.feature
```diff
@@ -23,7 +22,6 @@ Feature: Sending data using InvokeHTTP to a receiver using ListenHTTP
 
   Scenario: A MiNiFi instance sends data through a HTTP proxy and another one listens
     Given a GetFile processor with the "Input Directory" property set to "/tmp/input"
-    And the "Keep Source File" property of the GetFile processor is set to "true"
```
lordgamez, line 26: As far as I remember these were kept here to avoid flakiness. It could happen that the due to some environment issues the first HTTP request times out or fails in that case after the next GetFile invocation can still succeed and the test case succeeds. If this is removed the GetFile only runs once.

PR #1178, opened 2021-09-22
MINIFICPP-1643 Add Managed Identity support for Azure processors

lordgamez:
- Adding `Use Managed Identity Credentials` to `AzureStorageCredentialsService` and `PutAzureBlobStorage` properties to support managed identities
- `AzureBlobStorage` can now support empty `Blob` property by using the filename attribute as a default
- Azure processors now use the following credential priority list if they are set: Managed Identity, Connection String property, any other credential properties
- PutAzureBlobStorage prioritizes credentials set in the `AzureStorageCredentialsService` over the ones set in processor properties

https://issues.apache.org/jira/browse/MINIFICPP-1643

--------------------------------------------------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/azure/processors/PutAzureBlobStorage.cpp
```diff
@@ -138,38 +149,80 @@ void PutAzureBlobStorage::onSchedule(const std::shared_ptr<core::ProcessContext>
...
 }
 
-std::string PutAzureBlobStorage::getConnectionString(
+std::optional<storage::AzureStorageCredentials> PutAzureBlobStorage::getCredentials(
     const std::shared_ptr<core::ProcessContext> &context,
     const std::shared_ptr<core::FlowFile> &flow_file) const {
-  auto connection_string = getAzureConnectionStringFromProperties(context, flow_file);
-  if (!connection_string.empty()) {
-    return connection_string;
+  auto credentialsValid = [](const storage::AzureStorageCredentials& credentials) {
```
fgerlits, line 205: this could be a member function in `AzureStorageCredentials` (which could then be used in `PutAzureDataLakeStorage::onSchedule`, too)

extensions/azure/storage/AzureBlobStorage.cpp
```diff
@@ -23,61 +23,39 @@
...
-  connection_string_ = connection_string;
-  container_name_ = container_name;
-  logger_->log_debug("Client has been reset with new credentials");
-  container_client_ = std::make_unique<Azure::Storage::Blobs::BlobContainerClient>(Azure::Storage::Blobs::BlobContainerClient::CreateFromConnectionString(connection_string, container_name));
+AzureBlobStorage::AzureBlobStorage(std::unique_ptr<BlobStorageClient> blob_storage_client)
+  : blob_storage_client_(blob_storage_client ? std::move(blob_storage_client) : std::make_unique<AzureBlobStorageClient>()) {
 }
 
-void AzureBlobStorage::createContainer() {
+std::optional<bool> AzureBlobStorage::createContainerIfNotExists(const PutAzureBlobStorageParameters& params) {
```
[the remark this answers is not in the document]
lordgamez: Currently we only care if there was an error or not while trying to create the container, but I would keep the possibility to check if the container already existed or not if we want to differentiate later.

CONTROLLERS.md
```diff
@@ -56,8 +56,9 @@ default values, and whether a property supports the NiFi Expression Language.
 
 | Name | Default Value | Allowable Values | Expression Language Supported? | Description |
 | - | - | - | - | - |
-|Storage Account Name||||The storage account name.|
-|Storage Account Key||||The storage account key. This is an admin-like password providing access to every container in this account. It is recommended one uses Shared Access Signature (SAS) token instead for fine-grained control with policies.|
-|SAS Token||||Shared Access Signature token. Specify either SAS Token (recommended) or Account Key.|
+|Storage Account Name|||Yes|The storage account name.|
+|Storage Account Key|||Yes|The storage account key. This is an admin-like password providing access to every container in this account. It is recommended one uses Shared Access Signature (SAS) token instead for fine-grained control with policies.|
+|SAS Token|||Yes|Shared Access Signature token. Specify either SAS Token (recommended) or Account Key together with Storage Account Key if Managed Identity is not used.|
```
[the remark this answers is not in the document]
lordgamez: Updated the descriptions in 184b900f78ed72755f3d0e1876cc69d694da7a9e

extensions/azure/controllerservices/AzureStorageCredentialsService.cpp
```diff
@@ -22,53 +22,69 @@
...
 }
 
 void AzureStorageCredentialsService::onEnable() {
-  getProperty(StorageAccountName.getName(), credentials_.storage_account_name);
-  getProperty(StorageAccountKey.getName(), credentials_.storage_account_key);
-  getProperty(SASToken.getName(), credentials_.sas_token);
-  getProperty(CommonStorageAccountEndpointSuffix.getName(), credentials_.endpoint_suffix);
-  getProperty(ConnectionString.getName(), credentials_.connection_string);
+  std::string value;
+  if (getProperty(StorageAccountName.getName(), value)) {
```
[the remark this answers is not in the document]
lordgamez: You are right, I'm not sure why it was used in AWS controller service and also documented that way. I removed all signs of expression language support for controller services in 184b900f78ed72755f3d0e1876cc69d694da7a9e

extensions/azure/storage/AzureStorageCredentials.h
```diff
@@ -21,53 +21,31 @@
...
+  void setEndpontSuffix(const std::string& endpoint_suffix);
+  void setConnectionString(const std::string& connection_string);
+  void setUseManagedIdentityCredentials(bool use_managed_identity_credentials);
+
+  std::string getStorageAccountName() const;
+  std::string getEndpointSuffix() const;
+  bool getUseManagedIdentityCredentials() const;
+  std::string buildConnectionString() const;
+
+  bool operator==(const AzureStorageCredentials& other);
```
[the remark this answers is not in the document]
lordgamez: Updated in 184b900f78ed72755f3d0e1876cc69d694da7a9e

extensions/azure/processors/PutAzureBlobStorage.cpp
lordgamez: Added in 184b900f78ed72755f3d0e1876cc69d694da7a9e

extensions/azure/storage/AzureBlobStorage.cpp
```diff
@@ -23,61 +23,39 @@
...
-  if (connection_string == connection_string_ && container_name_ == container_name) {
-    logger_->log_debug("Client credentials have not changed, no need to reset client");
-    return;
-  }
-  connection_string_ = connection_string;
-  container_name_ = container_name;
-  logger_->log_debug("Client has been reset with new credentials");
-  container_client_ = std::make_unique<Azure::Storage::Blobs::BlobContainerClient>(Azure::Storage::Blobs::BlobContainerClient::CreateFromConnectionString(connection_string, container_name));
+AzureBlobStorage::AzureBlobStorage(std::unique_ptr<BlobStorageClient> blob_storage_client)
+  : blob_storage_client_(blob_storage_client ? std::move(blob_storage_client) : std::make_unique<AzureBlobStorageClient>()) {
```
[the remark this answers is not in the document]
lordgamez: Added in 184b900f78ed72755f3d0e1876cc69d694da7a9e

extensions/azure/storage/AzureBlobStorageClient.cpp
```diff
@@ -0,0 +1,65 @@
...
+namespace org::apache::nifi::minifi::azure::storage {
+
+void AzureBlobStorageClient::resetClientIfNeeded(const AzureStorageCredentials &credentials, const std::string &container_name) {
+  if (container_client_ && credentials == credentials_ && container_name == container_name_) {
+    logger_->log_debug("Client credentials have not changed, no need to reset client");
+    return;
+  }
+
+  if (credentials.getUseManagedIdentityCredentials()) {
+    logger_->log_debug("Client has been reset with new managed identity credentials.");
```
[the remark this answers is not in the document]
lordgamez: I could not check all internal implementation so it could be possible, I moved the logs to the end in 184b900f78ed72755f3d0e1876cc69d694da7a9e

extensions/azure/storage/AzureDataLakeStorageClient.cpp
```diff
@@ -20,19 +20,28 @@
...
+#include "azure/identity.hpp"
+
 namespace org::apache::nifi::minifi::azure::storage {
 
-void AzureDataLakeStorageClient::resetClientIfNeeded(const std::string& connection_string, const std::string& file_system_name) {
-  if (client_ == nullptr || connection_string_ != connection_string || file_system_name_ != file_system_name) {
-    client_ = std::make_unique<Azure::Storage::Files::DataLake::DataLakeFileSystemClient>(
-      Azure::Storage::Files::DataLake::DataLakeFileSystemClient::CreateFromConnectionString(connection_string, file_system_name));
+void AzureDataLakeStorageClient::resetClientIfNeeded(const AzureStorageCredentials& credentials, const std::string& file_system_name) {
+  if (client_ == nullptr || credentials_ != credentials || file_system_name_ != file_system_name) {
```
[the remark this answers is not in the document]
lordgamez: Updated in 184b900f78ed72755f3d0e1876cc69d694da7a9e

extensions/azure/storage/DataLakeStorageClient.h
```diff
@@ -20,24 +20,27 @@
...
   bool replace_file = false;
 };
 
-class DataLakeStorageClient {
+class DataLakeStorageClient : public AzureStorageClient {
  public:
   virtual bool createFile(const PutAzureDataLakeStorageParameters& params) = 0;
   virtual std::string uploadFile(const PutAzureDataLakeStorageParameters& params, gsl::span<const uint8_t> buffer) = 0;
-  virtual ~DataLakeStorageClient() {}
+  virtual ~DataLakeStorageClient() = default;
```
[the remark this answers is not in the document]
lordgamez: Removed in 184b900f78ed72755f3d0e1876cc69d694da7a9e

libminifi/test/azure-tests/PutAzureBlobStorageTests.cpp
```diff
@@ -283,24 +278,75 @@ TEST_CASE_METHOD(PutAzureBlobStorageTestsFixture, "Test credentials settings", "
...
   }
+
+  SECTION("Account name and managed identity are used in properties") {
+    plan->setProperty(put_azure_blob_storage, "Storage Account Name", STORAGE_ACCOUNT_NAME);
+    plan->setProperty(put_azure_blob_storage, "Use Managed Identity Credentials", "true");
+    test_controller.runSession(plan, true);
+    REQUIRE(getFailedFlowFileContents().size() == 0);
+    auto passed_params = mock_blob_storage_ptr->getPassedParams();
+    REQUIRE(passed_params.credentials.buildConnectionString().empty());
+    REQUIRE(passed_params.credentials.getStorageAccountName() == STORAGE_ACCOUNT_NAME);
```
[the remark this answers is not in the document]
lordgamez: Changed it in both Blob and Data Lake storage tests with some assertion order changes to be more logical in 184b900f78ed72755f3d0e1876cc69d694da7a9e

extensions/azure/processors/PutAzureBlobStorage.cpp
```diff
@@ -179,53 +232,43 @@ void PutAzureBlobStorage::onTrigger(const std::shared_ptr<core::ProcessContext>
...
   }
 
   std::optional<storage::UploadBlobResult> upload_result;
   {
     // TODO(lordgamez): This can be removed after maximum allowed threads are implemented. See https://issues.apache.org/jira/browse/MINIFICPP-1566
+    // When used in multithreaded environment make sure to use the azure_storage_mutex_ to lock the wrapper so the
+    // client is not reset with different configuration while another thread is using it.
     std::lock_guard<std::mutex> lock(azure_storage_mutex_);
-    createAzureStorageClient(connection_string, container_name);
     if (create_container_) {
```
martinzink, line 242: onSchedule sets this create_container_ member but onSchedule is not protected by the azure_storage_mutex_
Could that cause unexpected behaviour in multithreaded enviroment?

extensions/azure/processors/PutAzureDataLakeStorage.cpp
```diff
@@ -72,19 +72,26 @@ void PutAzureDataLakeStorage::initialize() {
...
+  if (!credentials) {
     throw Exception(PROCESS_SCHEDULE_EXCEPTION, "Azure Storage Credentials Service property missing or invalid");
   }
 
+  if ((!credentials->getUseManagedIdentityCredentials() && credentials->buildConnectionString().empty()) ||
+      (credentials->getUseManagedIdentityCredentials() && credentials->getStorageAccountName().empty())) {
+    throw Exception(PROCESS_SCHEDULE_EXCEPTION, "Azure Storage Credentials Service properties are not set or invalid");
+  }
+
+  credentials_ = *credentials;
```
martinzink, line 84: This onSchedule also sets members but isnt protected by the mutex

martinzink (commented):
Looks good to me only a few questions regarding the mutex usage.

extensions/azure/processors/PutAzureBlobStorage.cpp
lordgamez: It should not be a problem as onSchedule is only run once and the onTrigger calls that could be running in parallel may only read that variable. The mutex is used for protecting the Azure client when multiple onTrigger calls could reset the client with different parameters while running in parallel.

extensions/azure/processors/PutAzureDataLakeStorage.cpp
lordgamez: Same as before.

extensions/azure/processors/PutAzureBlobStorage.cpp
martinzink: Yeah I see your point, however onSchedule is on the public interface (so in theory anyone could call that) and the overhead from the mutex would be non-existent (onSchedule is rarely called) so I would still prefer if it was protected as well, purely for object oriented principles. But I am not hell bent on that.

lordgamez: That's true, in theory anyone could call it, but we usually handle onSchedule to be only run once before onTrigger calls, and we do not handle the possible concurrency in any other processor either. Also as the feature which will remove the mutexes from onTriggers and instead will mark the processors if they can be run in parallel or not is already in progress I would not add additional mutex usages to it.

martinzink (approved):

fgerlits (approved):

extensions/azure/processors/PutAzureBlobStorage.cpp
```diff
@@ -138,38 +149,75 @@ void PutAzureBlobStorage::onSchedule(const std::shared_ptr<core::ProcessContext>
...
+  logger_->log_debug("No valid Azure credentials are set in credentials controller service, checking properties...");
+
+  auto property_creds = getAzureCredentialsFromProperties(context, flow_file);
+  if (property_creds.isValid()) {
+    logger_->log_debug("Azure credentials read from properties!");
+    return property_creds;
   }
 
-  return getConnectionStringFromControllerService(context);
+  logger_->log_error("No valid Azure credentials are set in credentials controller service nor in properties!");
```
adamdebreceni, line 219: we already log an error after calling this method and it returns a nullopt at line 182

extensions/azure/processors/PutAzureBlobStorage.cpp
```diff
@@ -138,38 +149,75 @@ void PutAzureBlobStorage::onSchedule(const std::shared_ptr<core::ProcessContext>
...
 
-std::string PutAzureBlobStorage::getConnectionString(
+std::optional<storage::AzureStorageCredentials> PutAzureBlobStorage::getCredentials(
     const std::shared_ptr<core::ProcessContext> &context,
     const std::shared_ptr<core::FlowFile> &flow_file) const {
-  auto connection_string = getAzureConnectionStringFromProperties(context, flow_file);
-  if (!connection_string.empty()) {
-    return connection_string;
+  auto controller_service_creds = getCredentialsFromControllerService(context);
+  if (controller_service_creds && controller_service_creds->isValid()) {
```
adamdebreceni, line 206: I feel like we should differentiate between "nullopt because the user didn't set the Azure Storage Credentials Service property" and between the "user set something (explicitly requesting the service) but did a typo or added a different service by accident", in the latter case I don't think that the fallback to property credentials is the way to go

adamdebreceni: a default constructed AzureStorageCredentials seems to be inValid, that could be returned in the second case

extensions/azure/processors/PutAzureBlobStorage.cpp
lordgamez: Removed this log in favor of the previous one in c9965a1249b9fae617c6a4467d1ec2596b915158

extensions/azure/processors/PutAzureBlobStorage.cpp
lordgamez: Good point, added a change to fail in case invalid parameters are set in the credentials service in c9965a1249b9fae617c6a4467d1ec2596b915158

adamdebreceni: we might need to touch `getCredentialsFromControllerService` as well, to return not a nullopt but an invalid `AzureStorageCredentials` after it observes that the `Azure Storage Credentials Service` is set but subsequent attempts to get the service fail

lordgamez: I'm not sure if it's worth differentiating if the provided service name is invalid or empty. From the user's point of view the result will be the same and you will get a specific error message which was the case here. Do you think it's a problem that in both cases we check if the property credentials are set?

adamdebreceni: right now it seems to me, that if the user set the `Azure Storage Credentials Service` property but there is no such service or it is not of the appropriate type we return `std::nullopt`, and we go on trying to create the credentials from the properties despite the user explicitly providing the service name

lordgamez: Yes I was just conflicted if that was actually a bad thing. But I suppose technically it could happen that the service name is wrong and the property credentials are set to some valid parameters and the processor does not use the credentials it was intended to use and works regardless. I added a result code to the function in 2c5a10d585fabbd6f2127ad9a0814dcb4bcee719. I think that more explicitly shows the cause of failure.

extensions/azure/processors/PutAzureBlobStorage.h
```diff
@@ -77,11 +77,8 @@ class PutAzureBlobStorage final : public AzureStorageProcessorBase {
         return -1;
       }
 
-      result_ = blob_storage_wrapper_.uploadBlob(blob_name_, buffer.data(), flow_size_);
-      if (!result_) {
-        return read_ret;
-      }
-      return result_->length;
+      result_ = azure_blob_storage_.uploadBlob(params_, gsl::span<const uint8_t>{buffer.data(), flow_size_});
```
szaszm, line 80: Specifying the span template argument should be unnecessary. CTAD may work, but `gsl::make_span` definitely does.

extensions/azure/storage/AzureStorageCredentials.h
```diff
@@ -21,53 +21,32 @@
...
+  void setSasToken(const std::string& sas_token);
+  void setEndpontSuffix(const std::string& endpoint_suffix);
+  void setConnectionString(const std::string& connection_string);
+  void setUseManagedIdentityCredentials(bool use_managed_identity_credentials);
+
+  std::string getStorageAccountName() const;
+  std::string getEndpointSuffix() const;
+  bool getUseManagedIdentityCredentials() const;
+  std::string buildConnectionString() const;
+  bool isValid() const;
```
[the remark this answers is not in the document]
lordgamez: I only provided getter for the `use_managed_identity_credentials_` member and the `storage_account_name_` and `endpoint_suffix_` members to be retrievable if managed identity is used. The goal was that every other members should only be settable and if managed identity is not used only the full connectionstring should be built and retrieved with the `buildConnectionString` member (the built connection string is not stored in the `connection_string_` member, it is a separate member for overriding credentials if needed).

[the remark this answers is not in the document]
szaszm: Okay, that makes sense. Thanks.

extensions/azure/processors/PutAzureBlobStorage.h
lordgamez: Updated in 32d6542825310892eed3e5648dd55bc5b0d66257

extensions/azure/storage/AzureStorageClient.h
```diff
@@ -0,0 +1,34 @@
...
+
+namespace org::apache::nifi::minifi::azure::storage {
+
+class AzureStorageClient {
+ public:
+  virtual ~AzureStorageClient() = default;
+
+ protected:
+  AzureStorageCredentials credentials_;
+};
```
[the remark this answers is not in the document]
lordgamez: Updated in 32d6542825310892eed3e5648dd55bc5b0d66257

extensions/azure/storage/BlobStorageClient.h
```diff
@@ -24,40 +24,23 @@
...
+namespace org::apache::nifi::minifi::azure::storage {
+
+struct PutAzureBlobStorageParameters {
+  AzureStorageCredentials credentials;
+  std::string container_name;
+  std::string blob_name;
 };
 
-class BlobStorage {
+class BlobStorageClient : public AzureStorageClient {
```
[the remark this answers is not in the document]
lordgamez: Updated in 32d6542825310892eed3e5648dd55bc5b0d66257

extensions/azure/storage/AzureBlobStorage.cpp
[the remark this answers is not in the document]
lordgamez: Updated in 32d6542825310892eed3e5648dd55bc5b0d66257

extensions/azure/storage/AzureStorageClient.h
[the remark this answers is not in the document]
szaszm: I think `credentials_` is not even needed in `BlobStorageClient`, as it's just an interface class. If you agree, I suggest pushing it further down to the actual implementation classes.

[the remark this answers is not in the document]
lordgamez: I agree, it's better to keep it as an interface without any members, moved the member in b2847320d940d1a53fe5070b749dbd35b4a9d64c

extensions/azure/storage/AzureBlobStorage.h
```diff
@@ -24,33 +24,29 @@
 #include <string>
 #include <vector>
 
-#include "BlobStorage.h"
+#include "BlobStorageClient.h"
 #include "azure/storage/blobs.hpp"
 #include "core/logging/Logger.h"
 #include "core/logging/LoggerConfiguration.h"
+#include "gsl/gsl-lite.hpp"
```
szaszm, line 31: When introducing gsl-lite to the codebase, I created `"utils/gsl.h"` with the intention of it being the common entry point from minifi to gsl, including a namespace alias of `::org::apache::nifi::minifi::gsl` to `::gsl_lite`, following [gsl-lite's own guidelines for libraries](https://github.com/gsl-lite/gsl-lite#using-gsl-lite-in-libraries). I also introduced one gsl-related utility in the same file, but I'm open to moving that elsewhere.

Consider using `"utils/gsl.h"` and the namespace alias to access gsl-lite functionality.

lordgamez: Thanks, I didn't know about this purpose of the header, I replaced the old ones in 1ff3b235bc9a137a979a98f4c8bf3c9565676a39

szaszm (approved):

PR #1182, opened 2021-09-29
DRAFT: Refactor *StreamCallback to be function objects

szaszm:
Done with a limited number of extensions until I can gather feedback. I propose that we get rid of the *StreamCallback base classes and just use std::function to type erase the callback types where necessary. The reason is simlicity of implementation and usage.

The code on the PR branch is tested with these cmake flags: `-DFORCE_COLORED_OUTPUT=ON -DAWS_ENABLE_UNITY_BUILD=OFF -DASAN_BUILD=OFF -DENABLE_PYTHON=ON -DENABLE_OPS=ON -DENABLE_JNI=OFF -DENABLE_OPC=OFF -DENABLE_COAP=ON -DENABLE_GPS=ON -DENABLE_MQTT=ON -DENABLE_LIBRDKAFKA=ON -DENABLE_SENSORS=OFF -DENABLE_USB_CAMERA=ON -DENABLE_AWS=ON -DENABLE_SFTP=ON -DENABLE_OPENWSMAN=OFF -DENABLE_BUSTACHE=OFF -DENABLE_OPENCV=OFF -DENABLE_TENSORFLOW=OFF -DENABLE_SQL=ON -DENABLE_PCAP=OFF -DENABLE_LUA_SCRIPTING=ON -DENABLE_NANOFI=ON -DENABLE_AZURE=ON`

- `InputStreamCallback` and `OutputStreamCallback` are now aliases of `std::function<int64_t(const std::shared_ptr<io::BaseStream>&)>`.
- `ProcessSession::read` and friends now take `const InputStreamCallback&` and `const OutputStreamCallback&`. This means value semantics of the function object, due to `std::function`. If the callback needs to preserve state after the call, it needs to be passed with `std::ref(callback)`, which wraps it in a `std::reference_wrapper<T>`.
- Introduced `ProcessSession::readBuffer` that returns the status as usual and the buffer as `std::vector<std::byte>`. If you think I should use `uint8_t` or `char` or `unsigned char` or anything else, feel free to open a discussion. (here or PM) Also introduced `ProcessSession::appendBuffer`.
- Replaced about a dozen callback classes with calls to `readBuffer`/`writeBuffer`/`appendBuffer`.
- Replaced simple callback classes with inline lambdas or stream pipes.
- Converted stream pipes to use stream references instead of `shared_ptr`s. The motivation was to convert `SiteToSiteClient` to use stream pipes, but that didn't work out in the end and I reverted the `SiteToSiteClient`. I can revert this change if desired.

---

### For all changes:
     in the commit message?

### For code changes:

### Note:


fgerlits:
I like this.  Just two comments for discussion:
1. using a `const&` {Input,Output}StreamCallback parameter, which makes the `std::ref()` necessary, does not feel natural to me.  If I was to write new code with a stateful callback, I would not know I'd need `std::ref()`, and then spend a fair bit of time trying to find out why my code doesn't work.  I would prefer a universal reference parameter, but I don't like having to put `read()` etc in the header.  Maybe someone has a better idea?
2. [[predictable]] I would put the stream `shared_ptr` to `&` change in a separate pull request.

szaszm:
I'll link them, but there is no issue for this yet. This is mostly here just to gather feedback.

szaszm:
Closing this for now, will reopen when I have something final-ish ready

PR #1191, opened 2021-10-04
MINIFICPP-1566 - Annotate maximum allowed threads for processors

adam-markovics:

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/src/core/Processor.cpp
```diff
@@ -385,17 +385,20 @@ void Processor::validateAnnotations() const {
...
       if (hasIncomingConnections()) {
         throw Exception(PROCESS_SCHEDULE_EXCEPTION, "INPUT_FORBIDDEN was specified for the processor, but there are incoming connections");
       }
-      return;
     }
   }
+
+  if (isSingleThreaded() && max_concurrent_tasks_ > 1) {
+    throw Exception(PROCESS_SCHEDULE_EXCEPTION, "Processor can not be run in parallel, its \"max concurrent tasks\" value is too high. It must be set to 1.");
+  }
```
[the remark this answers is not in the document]
adam-markovics: If it was two functions, then here the new one would have to be called as well:
```
void ProcessGroup::verify() const {
  for (const auto& processor : processors_) {
    processor->validateAnnotations();
    processor->validateThreads(); // like this
  }
}
```
Do you agree with that?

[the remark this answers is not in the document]
szaszm: I was thinking about a `Processor::validateAnnotations` that calls private `validateInputRequirements` and `validateThreads`.

extensions/openwsman/processors/SourceInitiatedSubscriptionListener.cpp
```diff
@@ -435,7 +435,6 @@ bool SourceInitiatedSubscriptionListener::Handler::handleSubscriptionManager(str
   ws_xml_set_ns(subscription, XML_NS_CUSTOM_SUBSCRIPTION, "m");
 
   // Body/EnumerationResponse/Items/Subscription/Version
-  std::lock_guard<std::mutex> lock(processor_.mutex_);
```
szaszm, line 438: This is in an http request handler. Are you certain that civetweb (http server) doesn't use multiple threads that need synchronization on concurrent incoming requests? I don't know it's internals, but I think this might be an issue if it does.

adam-markovics: It seems to me that the mutex was there because of the member `subscribers_`, and is not related to civetweb. But because I can't tell for sure, and because this processor is unused (AFAIK) I will put it back, it won't do harm. From commit history it seems like nobodoy else touched the processor seriously other than the initial commit from bakaid.

szaszm: My theory is that civetweb calls the handler on a thread that is not in our normal thread pool. This means if the processor is running and an http requests are handled at the same time, or multiple http requests are being handled (on civet threads), then `subscribers_` is accessed concurrently, which is a data race.

szaszm (approved):

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -188,7 +188,6 @@ ConsumeWindowsEventLog::ConsumeWindowsEventLog(const std::string& name, const ut
 }
 
 void ConsumeWindowsEventLog::notifyStop() {
-  std::lock_guard<std::mutex> lock(on_trigger_mutex_);
   logger_->log_trace("start notifyStop");
   bookmark_.reset();
```
adamdebreceni, line 192: could an `onTrigger` be running when this `notifyStop` is called? if it could, the mutex protects the bookmark so we might not be able to remove it

PR #1192, opened 2021-10-04
MINIFICPP-1645 Only include each processor or controller service in the heartbeat once

fgerlits:
https://issues.apache.org/jira/browse/MINIFICPP-1645

`RocksDbPersistableKeyValueStoreService` is registered in the class loader under two different names, so we can refer to it from `minifi.properties` using different capitalization: eg. "RocksDBPersistableKeyValueStoreService" is accepted, too.  This is nice, but a side effect is that it used to be included in the heartbeat manifest twice.

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/http-curl/tests/C2LogHeartbeatTest.cpp
+++ b/extensions/http-curl/tests/C2LogHeartbeatTest.cpp
@@ -26,7 +26,12 @@
 #include "protocols/RESTSender.h"
 #include "HTTPIntegrationBase.h"
 #include "HTTPHandlers.h"
-#include "range/v3/all.hpp"
+#include "range/v3/action/sort.hpp"
+#include "range/v3/action/unique.hpp"
+#include "range/v3/range/conversion.hpp"
+#include "range/v3/view/filter.hpp"
+#include "range/v3/view/split.hpp"
+#include "range/v3/view/transform.hpp"
 #include "utils/IntegrationTestUtils.h"
 #include "utils/StringUtils.h"
 
--- a/extensions/http-curl/tests/C2LogHeartbeatTest.cpp
+++ b/extensions/http-curl/tests/C2LogHeartbeatTest.cpp
@@ -17,14 +17,18 @@
  */
 
 #undef NDEBUG
+
 #include "TestBase.h"
+
 #include "c2/C2Agent.h"
+#include "c2/HeartbeatLogger.h"
 #include "protocols/RESTProtocol.h"
 #include "protocols/RESTSender.h"
 #include "HTTPIntegrationBase.h"
 #include "HTTPHandlers.h"
+#include "range/v3/all.hpp"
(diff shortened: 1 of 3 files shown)


libminifi/include/core/ClassLoader.h
```diff
@@ -24,10 +24,10 @@
 #include <string>
 #include <map>
 #include <memory>
-#include "utils/StringUtils.h"
+
 #include "core/Core.h"
-#include "io/BufferStream.h"
 #include "ObjectFactory.h"
+#include "range/v3/all.hpp"
```
szaszm, line 30: Same here. Interestingly, the difference is still about 0.6 seconds despite using less utilities.

```
#include "range/v3/action/sort.hpp"
#include "range/v3/action/unique.hpp"
```

szaszm (approved):
+1 with or without the suggestions

lordgamez (approved):

extensions/http-curl/tests/C2LogHeartbeatTest.cpp
```diff
@@ -17,14 +17,18 @@
...
+
 #include "TestBase.h"
+
 #include "c2/C2Agent.h"
+#include "c2/HeartbeatLogger.h"
 #include "protocols/RESTProtocol.h"
 #include "protocols/RESTSender.h"
 #include "HTTPIntegrationBase.h"
 #include "HTTPHandlers.h"
+#include "range/v3/all.hpp"
```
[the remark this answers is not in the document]
fgerlits: Yeah, 0.6s (times many files, as I expect we'll use it all over the place) off the compilation time is a good thing.  I've changed the includes in https://github.com/apache/nifi-minifi-cpp/pull/1192/commits/6d76f4aca20e5a868ef43ef357ccd7b981442dd4.

martinzink (approved):

libminifi/include/core/ClassLoader.h
```diff
@@ -93,18 +94,18 @@ class ClassLoader {
...
       }
     }
     for (const auto& factory : loaded_factories_) {
       if (factory.second->getGroupName() == group) {
-        classes.push_back(factory.second->getClassName());
+        class_names.push_back(factory.second->getClassName());
       }
     }
-    return classes;
+    return std::move(class_names) | ranges::actions::sort | ranges::actions::unique;
```
adamdebreceni, line 108: I think we should move this to the source file (and its dependency on ranges) as many source files include ClassLoader.h transitively through Resource.h

fgerlits: makes sense; done in https://github.com/apache/nifi-minifi-cpp/pull/1192/commits/112db412cba844cc8014d76c2108434ef34d016d

adamdebreceni (approved):

PR #1211, opened 2021-11-09
MINIFICPP-1676 - Verify build id of extension before load

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/cmake/Extensions.cmake
+++ b/cmake/Extensions.cmake
@@ -22,10 +22,29 @@ define_property(GLOBAL PROPERTY EXTENSION-OPTIONS
 
 set_property(GLOBAL PROPERTY EXTENSION-OPTIONS "")
 
+set(extension-build-info-file "${CMAKE_CURRENT_BINARY_DIR}/ExtensionBuildInfo.cpp")
+file(GENERATE OUTPUT ${extension-build-info-file}
+    CONTENT "\
+    #ifdef BUILD_ID_VARIABLE_NAME\n\
+    extern const char* const BUILD_ID_VARIABLE_NAME = \"__EXTENSION_BUILD_IDENTIFIER_BEGIN__${BUILD_IDENTIFIER}__EXTENSION_BUILD_IDENTIFIER_END__\";\n\
+    #else\n\
+    static_assert(false, \"BUILD_ID_VARIABLE_NAME is not defined\");\n\
+    #endif\n")
+
+function(get_build_id_variable_name extension-name output)
+  string(REPLACE "-" "_" result ${extension-name})
+  string(APPEND result "_build_identifier")
+  set("${output}" "${result}" PARENT_SCOPE)
+endfunction()
+
 macro(register_extension extension-name)
   get_property(extensions GLOBAL PROPERTY EXTENSION-OPTIONS)
   set_property(GLOBAL APPEND PROPERTY EXTENSION-OPTIONS ${extension-name})
-  target_compile_definitions(${extension-name} PRIVATE "MODULE_NAME=${extension-name}")
+  get_build_id_variable_name(${extension-name} build-id-variable-name)
+  target_sources(${extension-name} PRIVATE ${extension-build-info-file})
+  target_compile_definitions(${extension-name}
+      PRIVATE "MODULE_NAME=${extension-name}"
+      PRIVATE "BUILD_ID_VARIABLE_NAME=${build-id-variable-name}")
   set_target_properties(${extension-name} PROPERTIES
           ENABLE_EXPORTS True
           POSITION_INDEPENDENT_CODE ON)
--- a/cmake/Extensions.cmake
+++ b/cmake/Extensions.cmake
@@ -27,7 +27,7 @@ file(GENERATE OUTPUT ${extension-build-info-file}
     CONTENT "\
     #include \"utils/Export.h\"\n\
     #ifdef BUILD_ID_VARIABLE_NAME\n\
-    EXTENSIONAPI const char* const BUILD_ID_VARIABLE_NAME = \"__EXTENSION_BUILD_IDENTIFIER_BEGIN__${BUILD_IDENTIFIER}__EXTENSION_BUILD_IDENTIFIER_END__\";\n\
+    EXTENSIONAPI extern const char* const BUILD_ID_VARIABLE_NAME = \"__EXTENSION_BUILD_IDENTIFIER_BEGIN__${BUILD_IDENTIFIER}__EXTENSION_BUILD_IDENTIFIER_END__\";\n\
     #else\n\
     static_assert(false, \"BUILD_ID_VARIABLE_NAME is not defined\");\n\
     #endif\n")
--- a/cmake/Extensions.cmake
+++ b/cmake/Extensions.cmake
@@ -41,6 +41,7 @@ macro(register_extension extension-name)
   get_property(extensions GLOBAL PROPERTY EXTENSION-OPTIONS)
   set_property(GLOBAL APPEND PROPERTY EXTENSION-OPTIONS ${extension-name})
   get_build_id_variable_name(${extension-name} build-id-variable-name)
+  set_source_files_properties(${extension-build-info-file} PROPERTIES GENERATED TRUE)
   target_sources(${extension-name} PRIVATE ${extension-build-info-file})
   target_compile_definitions(${extension-name}
       PRIVATE "MODULE_NAME=${extension-name}"
--- a/cmake/Extensions.cmake
+++ b/cmake/Extensions.cmake
@@ -25,8 +25,9 @@ set_property(GLOBAL PROPERTY EXTENSION-OPTIONS "")
 set(extension-build-info-file "${CMAKE_CURRENT_BINARY_DIR}/ExtensionBuildInfo.cpp")
 file(GENERATE OUTPUT ${extension-build-info-file}
     CONTENT "\
+    #include \"utils/Export.h\"\n\
     #ifdef BUILD_ID_VARIABLE_NAME\n\
-    extern const char* const BUILD_ID_VARIABLE_NAME = \"__EXTENSION_BUILD_IDENTIFIER_BEGIN__${BUILD_IDENTIFIER}__EXTENSION_BUILD_IDENTIFIER_END__\";\n\
+    EXTENSIONAPI const char* const BUILD_ID_VARIABLE_NAME = \"__EXTENSION_BUILD_IDENTIFIER_BEGIN__${BUILD_IDENTIFIER}__EXTENSION_BUILD_IDENTIFIER_END__\";\n\
     #else\n\
     static_assert(false, \"BUILD_ID_VARIABLE_NAME is not defined\");\n\
     #endif\n")
--- a/libminifi/include/core/extension/Utils.h
+++ b/libminifi/include/core/extension/Utils.h
@@ -20,6 +20,10 @@
 #include <optional>
 #include <functional>
 #include <chrono>
+#include <utility>
+#include <memory>
+#include <string>
+
 #include "utils/gsl.h"
 #include "utils/file/FileView.h"
 #include "utils/StringUtils.h"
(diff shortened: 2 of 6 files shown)


lordgamez (approved):

cmake/Extensions.cmake
```diff
@@ -22,10 +22,31 @@ define_property(GLOBAL PROPERTY EXTENSION-OPTIONS
...
 
+set(extension-build-info-file "${CMAKE_CURRENT_BINARY_DIR}/ExtensionBuildInfo.cpp")
+file(GENERATE OUTPUT ${extension-build-info-file}
+    CONTENT "\
+    #include \"utils/Export.h\"\n\
+    #ifdef BUILD_ID_VARIABLE_NAME\n\
+    EXTENSIONAPI extern const char* const BUILD_ID_VARIABLE_NAME = \"__EXTENSION_BUILD_IDENTIFIER_BEGIN__${BUILD_IDENTIFIER}__EXTENSION_BUILD_IDENTIFIER_END__\";\n\
+    #else\n\
+    static_assert(false, \"BUILD_ID_VARIABLE_NAME is not defined\");\n\
+    #endif\n")
```
szaszm, line 33: I think this is a hack. We shouldn't have to open the extension as a binary file, search for a prefix and extract the ID as text. Why don't extensions define a function that takes a build identifier and returns whether the extension is compatible with the agent or not? The agent could call this function after dlopen, before the extension's init.

libminifi/include/core/extension/Utils.h
```diff
@@ -0,0 +1,113 @@
...
+  std::string name;
+  std::filesystem::path dir;
+  std::string filename;
+
+  [[nodiscard]]
+  bool verify(const std::shared_ptr<logging::Logger>& logger) const {
+    auto path = getFullPath();
+    Timer timer{[&] (int ms) {
+      logger->log_error("Verification for '%s' took %d ms", path.string(), ms);
+    }};
```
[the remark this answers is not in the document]
adamdebreceni: the Timer has been removed as it was a development util and was left in by accident

cmake/Extensions.cmake
adamdebreceni: by the time we would call the verification function in the dll, the damage could very well be done, as there are numerous static objects (loggers, processors) created at load-time registering themselves, calling into libminifi

libminifi/test/unit/ExtensionVerificationTests.cpp
```diff
@@ -0,0 +1,98 @@
...
+#else
+const std::string extension_file = "libextension.so";
+#endif
+
+struct Fixture : public TestController {
+  Fixture() {
+    extension_ = std::filesystem::path(createTempDirectory()) / extension_file;
+  }
+  std::filesystem::path extension_;
+  static inline std::shared_ptr<logging::Logger> logger_{core::logging::LoggerFactory<Fixture>::getLogger()};
```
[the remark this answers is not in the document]
adamdebreceni: moved the utils in this test into an anonymous namespace

libminifi/include/core/extension/Utils.h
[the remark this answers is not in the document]
szaszm: I think it's useful to measure and log these times. Probably not as errors, but rather debug level logs, but I think this is useful data. I have observed extension loading taking seconds on windows, so I'm in favor of more measurement around extensions.

martinzink (approved):

libminifi/include/core/extension/Utils.h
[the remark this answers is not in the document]
adamdebreceni: restored the timer, is the date library among our thirdparty dependencies?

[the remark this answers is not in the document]
szaszm: Yes, it's already used by the expression language and systemd extensions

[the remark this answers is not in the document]
szaszm: After reading the current version, now I think it's good enough without the date library, and maybe it's not such a good idea to include it here, just to use a library function instead of a few extra characters of formatting.

cmake/Extensions.cmake
```diff
@@ -22,10 +22,31 @@ define_property(GLOBAL PROPERTY EXTENSION-OPTIONS
 
 set_property(GLOBAL PROPERTY EXTENSION-OPTIONS "")
 
+set(extension-build-info-file "${CMAKE_CURRENT_BINARY_DIR}/ExtensionBuildInfo.cpp")
+file(GENERATE OUTPUT ${extension-build-info-file}
+    CONTENT "\
+    #include \"utils/Export.h\"\n\
+    #ifdef BUILD_ID_VARIABLE_NAME\n\
+    EXTENSIONAPI extern const char* const BUILD_ID_VARIABLE_NAME = \"__EXTENSION_BUILD_IDENTIFIER_BEGIN__${BUILD_IDENTIFIER}__EXTENSION_BUILD_IDENTIFIER_END__\";\n\
```
szaszm, line 30: I think using `extern` here is incorrect, since it only declares, but doesn't define a variable. That is, unless the variable is defined somewhere. How it ends up in the binary is not clear to me, it shouldn't be there.

libminifi/include/utils/file/FileView.h
```diff
@@ -0,0 +1,184 @@
...
+class FileView {
+  static std::string getLastError() {
+#ifdef WIN32
+    return std::system_category().message(GetLastError());
+#else
+    return std::system_category().message(errno);
+#endif
+  }
+#ifdef WIN32
+  struct FileHandle {
```
szaszm, line 48: In this file, there is a lot of unnecessary complexity just to search in a memory-mapped file instead of using fstream to do conventional read. If there is another reason for this, then could you explain it? If not, could you convert the verification code to normal buffered file io code and remove these mmap wrappers?

adamdebreceni: `std::ifstream` is an input iterator and not a forward iterator, so `std::search` does not operate on it, we could read the file chunk by chunk but then we would have to cover edge cases like the id appearing on the boundary, tried wrapping the `std::ifstream` to accommodate it for `std::search` but it took tens of seconds with the constant seeks, the memory-mapped solution runs in tens of milliseconds, also we could use the Boyer-Moore searcher

I think it is best to use standard library utilities whenever possible instead of hand-rolling our own search implementation

szaszm: In general I agree about using the standard library, but here the cost is about 150 lines of code. I want to avoid that bloat if possible, even if that takes a few extra lines on the search side.

szaszm: To be clear, my objection is about code bloat and more maintenance burden, not memory-mapped io itself. If you want to use memory mapped io, having a third party platform-independent lib doing that would be an improvement IMO. [mio](https://github.com/mandreyel/mio) looks like a good candidate.

adamdebreceni: I took a stab at a custom buffering solution but it quickly felt like reimplementing the paging logic the memory mapping gives us (just less performant) so I would stick with the memory mapped solution

regarding including a thirdparty in order to avoid bloat seems nice, but it is harder to modify (fix) and we are at the mercy of the owners for maintenance (in the library you linked the owner commented in one of the issues that they can't work on this project for now), since the current solution is less than 200 lines and essentially a wrapper around platform utils, and is tested on every application start, and is currently used in a single place, I feel like we should wait with replacing it with a thirdparty until we have multiple uses for it and/or more extensive requirements, like specifying offset/size

cmake/Extensions.cmake
adamdebreceni: could you expand on this? should this be split into a separate declaration with `extern` and a definition? I found that without `extern` this const variable would have internal linkage and stripped from the shared library

szaszm: Normally `extern` is used do declare a global variable in a header that is then defined in some implementation file without violation ODR. Otherwise a global variable declaration would also be a definition. I don't understand why having an initializer here is even legal c++, since it should go to the separate definition, not the declaration.

According to [this](https://stackoverflow.com/a/57900212) stackoverflow answer, having `extern` here is the same as not having it, because the initializer makes a definition and all globals have external linkage by default.

What do you think about having a global function instead that returns a `const char*` to the magic string? I think that shouldn't be stripped from the binary even with LTO.

adamdebreceni: only non-const variables are global by default, const ones have internal linkage unless declared extern

szaszm: Okay, I missed that. Now it all makes sense, nevermind the comment.

libminifi/include/utils/file/FileView.h
szaszm: I committed a reasonably fast solution without mmap. Most cases take <5ms on my machine, while the worst cases take around 40ms. I removed the log lines, because they would have required a much more complex solution, and it's probably not worth so much complexity. 
f263e07

libminifi/include/core/extension/Utils.h
```diff
@@ -0,0 +1,111 @@
...
+  explicit Timer(Callback cb): cb_(std::move(cb)) {}
+  ~Timer() {
+    cb_(std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - start_));
+  }
+  std::chrono::steady_clock::time_point start_{std::chrono::steady_clock::now()};
+  Callback cb_;
+};
+
+struct LibraryDescriptor {
+  std::string name;
```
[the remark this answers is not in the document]
adamdebreceni: added

libminifi/include/core/extension/Utils.h
```diff
@@ -0,0 +1,111 @@
...
+#else
+  const std::string extension = ".so";
+#endif
+
+#ifdef WIN32
+  const std::string prefix = "";
+#else
+  const std::string prefix = "lib";
+#endif
+  std::string filename = path.filename().string();
```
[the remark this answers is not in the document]
adamdebreceni: done

libminifi/include/core/extension/Utils.h
```diff
@@ -0,0 +1,111 @@
...
+
+  [[nodiscard]]
+  std::filesystem::path getFullPath() const {
+    return dir / filename;
+  }
+};
+
+std::optional<LibraryDescriptor> asDynamicLibrary(const std::filesystem::path& path) {
+#if defined(WIN32)
+  const std::string extension = ".dll";
```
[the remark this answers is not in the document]
adamdebreceni: made it static and a string_view

libminifi/include/core/extension/Utils.h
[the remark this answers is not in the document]
adamdebreceni: so I did have a reason :) reverted

libminifi/src/utils/file/FileUtils.cpp
```diff
@@ -48,6 +48,44 @@ uint64_t computeChecksum(const std::string &file_name, uint64_t up_to_position)
   return checksum;
 }
 
+bool contains(const std::filesystem::path& file_path, std::string_view string) {
```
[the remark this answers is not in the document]
szaszm: That was my addition, so renamed it to `text_to_search`.

libminifi/include/utils/file/FileView.h
szaszm: Does the :+1: mean that you agree with the changes and want to keep them as submitted? If there are any objections, we can continue to discuss here or on slack.

adamdebreceni: yes, we can to keep it as submitted

szaszm (approved):

adamdebreceni:
since I cannot approve my own pull request, please treat this comment as an 'approve' for commits made by @szaszm

adam-markovics (approved):

libminifi/include/core/extension/Utils.h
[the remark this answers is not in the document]
adamdebreceni: you are right, and since we don't really need to expose the `Timer`'s internals, made it into a `class`

fgerlits (approved):

PR #1214, opened 2021-11-10
MINIFICPP-1683 Revert github actions caching changes

lordgamez:
Recently caching was changed to only keep a single cache key for each branch and update the cache content for that key in every CI build. This was done to save space so it would take longer for the older branch caches (especially main) to be evicted due to the 5GB cache space per repository. Unfortunately github cache actions does not support updating the cache content for a key because caches are immutable:
- https://github.com/actions/cache/issues/171
- https://github.com/actions/cache/issues/109

https://issues.apache.org/jira/browse/MINIFICPP-1683

------------------------------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,9 +15,10 @@ jobs:
         uses: actions/cache@v2
         with:
           path: ${{ env.CCACHE_DIR }}
-          key: macos-xcode12.0-ccache-${{github.ref}}
+          key: macos-xcode12.0-ccache-${{github.ref}}-${{github.sha}}
           restore-keys: |
-            macos-xcode12.0-ccache-
+            macos-xcode12.0-ccache-${{github.ref}}-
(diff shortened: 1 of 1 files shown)


szaszm:
Could we only cache builds on main and always restore the cache of main, regardless of the branch? It should be relatively close to most branches, so we would get the most benefit.

lordgamez:
> Could we only cache builds on main and always restore the cache of main, regardless of the branch? It should be relatively close to most branches, so we would get the most benefit.

Unfortunately I was not able to save the cache only specifically for the main branch. The problem is that there seems to be no option to do this as we have to define a key for every job that uses caching and if the key is not found it will be saved. Only the restore keys can be specified in a priority order.

szaszm (approved):

adamdebreceni (approved):

fgerlits (approved):

PR #1218, opened 2021-11-29
MINIFICPP-1692 - TLSSocket hangs when no more data is available

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/cmake/BuildTests.cmake
+++ b/cmake/BuildTests.cmake
@@ -101,6 +101,7 @@ target_include_directories(${CATCH_MAIN_LIB} SYSTEM BEFORE PRIVATE "${CMAKE_SOUR
 SET(TEST_RESOURCES ${TEST_DIR}/resources)
 
 GETSOURCEFILES(UNIT_TESTS "${TEST_DIR}/unit/")
+GETSOURCEFILES(TLS_UNIT_TESTS "${TEST_DIR}/unit/tls/")
 GETSOURCEFILES(NANOFI_UNIT_TESTS "${NANOFI_TEST_DIR}")
 GETSOURCEFILES(INTEGRATION_TESTS "${TEST_DIR}/integration/")
 
@@ -114,6 +115,18 @@ FOREACH(testfile ${UNIT_TESTS})
   add_test(NAME "${testfilename}" COMMAND "${testfilename}" WORKING_DIRECTORY ${TEST_DIR})
 ENDFOREACH()
 message("-- Finished building ${UNIT_TEST_COUNT} unit test file(s)...")
+
+if (NOT OPENSSL_OFF)
+  SET(UNIT_TEST_COUNT 0)
+  FOREACH(testfile ${TLS_UNIT_TESTS})
+    get_filename_component(testfilename "${testfile}" NAME_WE)
+    add_executable("${testfilename}" "${TEST_DIR}/unit/tls/${testfile}")
+    createTests("${testfilename}")
+    MATH(EXPR UNIT_TEST_COUNT "${UNIT_TEST_COUNT}+1")
+    add_test(NAME "${testfilename}" COMMAND "${testfilename}" "${TEST_RESOURCES}/" WORKING_DIRECTORY ${TEST_DIR})
+  ENDFOREACH()
+  message("-- Finished building ${UNIT_TEST_COUNT} TLS unit test file(s)...")
+endif()
 
 if(NOT WIN32 AND ENABLE_NANOFI)
   SET(UNIT_TEST_COUNT 0)
--- a/libminifi/src/io/tls/TLSSocket.cpp
+++ b/libminifi/src/io/tls/TLSSocket.cpp
@@ -436,7 +436,7 @@ size_t TLSSocket::read(uint8_t *buf, size_t buflen) {
       sslStatus = SSL_get_error(fd_ssl, status);
     } while (status < 0 && sslStatus == SSL_ERROR_WANT_READ);
 
-    if (status < 0)
+    if (status <= 0)
       break;
 
     buflen -= gsl::narrow<size_t>(status);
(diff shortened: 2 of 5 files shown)


lordgamez (approved):

adam-markovics (approved):

libminifi/src/io/tls/TLSSocket.cpp
```diff
@@ -434,9 +434,9 @@ size_t TLSSocket::read(uint8_t *buf, size_t buflen) {
       const auto ssl_read_size = gsl::narrow<int>(std::min(buflen, gsl::narrow<size_t>(std::numeric_limits<int>::max())));
       status = SSL_read(fd_ssl, buf, ssl_read_size);
       sslStatus = SSL_get_error(fd_ssl, status);
-    } while (status < 0 && sslStatus == SSL_ERROR_WANT_READ);
+    } while (status <= 0 && sslStatus == SSL_ERROR_WANT_READ);
```
szaszm, line 437: Wouldn't this just keep us inside the inner loop while SSL_read returns 0, which I assume happens when there is no more data available?

adamdebreceni: I found no connection between the result of `SSL_read` and if retry is possible, from [openssl](https://www.openssl.org/docs/man1.1.1/man3/SSL_read.html) regarding the return value:

> Old documentation indicated a difference between 0 and -1, and that -1 was retryable. You should instead call SSL_get_error() to find out if it's retryable.

it does not even seem specify that a return value of '0' indicates no more data, just that <= 0 is a return state

adamdebreceni: if truly no more data can be read, the `SSL_get_error` would return `SSL_ERROR_ZERO_RETURN` ([ref](https://www.openssl.org/docs/man1.1.1/man3/SSL_get_error.html))

szaszm (approved):

martinzink (approved):

libminifi/test/unit/tls/TLSStreamTests.cpp
```diff
@@ -0,0 +1,82 @@
...
+    throw std::logic_error("Specify the key directory");
+  }
+  std::filesystem::path key_dir(argv[1]);
+
+  LogTestController::getInstance().setTrace<minifi::io::Socket>();
+  LogTestController::getInstance().setTrace<minifi::io::TLSSocket>();
+  LogTestController::getInstance().setTrace<minifi::io::TLSServerSocket>();
+  LogTestController::getInstance().setTrace<minifi::io::TLSContext>();
+
+  auto server = std::make_unique<SimpleSSLTestServer>(TLSv1_2_server_method(), 0, key_dir);
```
adamdebreceni, line 55: I tried to use the `TSLServerSocket` but couldn't trigger the observed behavior to verify the fix, so I factored out the `SimpleSSLTestServer` and used that one instead, now the test should fail when the fix is reverted but pass otherwise

martinzink (approved):
LGTM, kudos for the test :+1:

cmake/BuildTests.cmake
```diff
@@ -101,6 +101,7 @@ target_include_directories(${CATCH_MAIN_LIB} SYSTEM BEFORE PRIVATE "${CMAKE_SOUR
 SET(TEST_RESOURCES ${TEST_DIR}/resources)
 
 GETSOURCEFILES(UNIT_TESTS "${TEST_DIR}/unit/")
+GETSOURCEFILES(TLS_UNIT_TESTS "${TEST_DIR}/unit/tls/")
```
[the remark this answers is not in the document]
adamdebreceni: I wanted to do that first but this test needs access to the test resources which other tests get through command line arguments, moreover to access said resource path we need to have our own main, which conflicts with linking to `CATCH_MAIN_LIB`, moreover these tests use relative source file names (`add_executable("${testfilename}" "${TEST_DIR}/unit/${testfile}")`) but `GETSOURCEFILES` do not expose to what path we would like to have the relative path (`FILE(GLOB children RELATIVE ${curdir} ${curdir}/*)`) so we would need to either expose that or use `file(GLOB ...)` here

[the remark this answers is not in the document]
szaszm: okay, that's reasonable

szaszm (approved):

libminifi/test/SimpleSSLTestServer.h
```diff
@@ -0,0 +1,130 @@
...
+  static SocketDescriptor createSocket(int port) {
+    struct sockaddr_in addr;
+
+    addr.sin_family = AF_INET;
+    addr.sin_port = htons(port);
+    addr.sin_addr.s_addr = htonl(INADDR_ANY);
+
+    SocketDescriptor socket_descriptor = socket(AF_INET, SOCK_STREAM, 0);
+    assert(socket_descriptor >= 0);
+    assert(bind(socket_descriptor, (struct sockaddr*)&addr, sizeof(addr)) >= 0);
```
szaszm, line 125: This fails on my machine as well, not just the CI. I think we need to generate a random port ourselves here and try to `bind` until we find one that works.

szaszm (changes requested):
Sorry for going back on my vote, but I found this issue while doing my pre-merge tests.

libminifi/test/SimpleSSLTestServer.h
adamdebreceni: which test fails?

szaszm: Now that I checked again, this doesn't seem to be new, TLSClientSocketSupportedProtocolsTest keeps failing even on the current main. Given the fact that this issue was not caused by your change, do you prefer to have this merged or to change the logic anyway to avoid this issue?

adamdebreceni: I added a commit that hopefully solves this problem by choosing a random port

szaszm:
Forgot to repost a +1, but the change resolved the test issue, so I merged the PR.

PR #1219, opened 2021-11-29
MINIFICPP-1691: PutSplunkHTTP and QuerySplunkIndexingStatus

martinzink:
MiNiFi C++ implementation of the processors added to NiFi in https://issues.apache.org/jira/browse/NIFI-7801 

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/splunk/PutSplunkHTTP.cpp
```diff
@@ -0,0 +1,180 @@
...
+    context->yield();
+    return;
+  }
+  auto flow_file = gsl::not_null(std::move(ff));
+
+  utils::HTTPClient client(getUrl() + getEndpoint(*context, flow_file), getSSLContextService(*context));
+  setHeaders(client);
+
+  std::unique_ptr<utils::ByteInputCallBack> payload_callback = std::make_unique<utils::ByteInputCallBack>();
+  std::unique_ptr<utils::HTTPUploadCallback> payload_callback_obj = std::make_unique<utils::HTTPUploadCallback>();
```
szaszm, line 166: Don't repeat the type after `make_unique`, use `auto` instead.
```suggestion
  const auto payload_callback = std::make_unique<utils::ByteInputCallBack>();
  const auto payload_callback_obj = std::make_unique<utils::HTTPUploadCallback>();
```

extensions/splunk/PutSplunkHTTP.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+  return addAttributesFromClientResponse(flow_file, client) && client.getResponseCode() == 200;
+}
+
+void setFlowFileAsPayload(core::ProcessSession& session,
+                                         core::ProcessContext& context,
+                                         utils::HTTPClient& client,
+                                         const gsl::not_null<std::shared_ptr<core::FlowFile>>& flow_file,
+                                         const std::unique_ptr<utils::ByteInputCallBack>& payload_callback,
+                                         const std::unique_ptr<utils::HTTPUploadCallback>& payload_callback_obj) {
```
szaszm, line 137: This function only needs references to the actual object. No need to force the caller to use unique_ptr, just use normal references instead.
```suggestion
                                         utils::ByteInputCallBack& payload_callback,
                                         utils::HTTPUploadCallback& payload_callback_obj) {
```

extensions/splunk/PutSplunkHTTP.cpp
```diff
@@ -0,0 +1,180 @@
...
+  updateSupportedProperties({Source, SourceType, Host, Index, ContentType});
+}
+
+void PutSplunkHTTP::onSchedule(const std::shared_ptr<core::ProcessContext>& context, const std::shared_ptr<core::ProcessSessionFactory>& sessionFactory) {
+  SplunkHECProcessor::onSchedule(context, sessionFactory);
+}
+
+
+namespace {
+std::optional<std::string> getContentType(core::ProcessContext& context, const gsl::not_null<std::shared_ptr<core::FlowFile>>& flow_file) {
```
szaszm, line 74: A const ref to a flow file should be enough here. If not, then getAttribute needs fixing.
```suggestion
std::optional<std::string> getContentType(core::ProcessContext& context, const core::FlowFile& flow_file) {
```

docker/Dockerfile
```diff
@@ -113,8 +114,8 @@ RUN cmake -DSTATIC_BUILD= -DSKIP_TESTS=true -DENABLE_ALL="${ENABLE_ALL}" -DENABL
     -DENABLE_COAP="${ENABLE_COAP}" -DENABLE_SQL="${ENABLE_SQL}" -DENABLE_MQTT="${ENABLE_MQTT}" -DENABLE_PCAP="${ENABLE_PCAP}" \
     -DENABLE_LIBRDKAFKA="${ENABLE_LIBRDKAFKA}" -DENABLE_SENSORS="${ENABLE_SENSORS}" -DENABLE_USB_CAMERA="${ENABLE_USB_CAMERA}" \
     -DENABLE_TENSORFLOW="${ENABLE_TENSORFLOW}" -DENABLE_AWS="${ENABLE_AWS}" -DENABLE_BUSTACHE="${ENABLE_BUSTACHE}" -DENABLE_SFTP="${ENABLE_SFTP}" \
-    -DENABLE_OPENWSMAN="${ENABLE_OPENWSMAN}" -DENABLE_AZURE="${ENABLE_AZURE}" -DENABLE_NANOFI=${ENABLE_NANOFI} -DENABLE_SYSTEMD=OFF \
```
szaszm, line 116: extra space added here

docker/test/integration/minifi/core/SplunkContainer.py
```diff
@@ -0,0 +1,26 @@
...
+            return
+
+        logging.info('Creating and running Splunk docker container...')
+        self.client.containers.run(
+            self.image_store.get_image(self.get_engine()),
+            detach=True,
+            name=self.name,
+            network=self.network.name,
+            environment=[
+                "SPLUNK_START_ARGS=--accept-license",
```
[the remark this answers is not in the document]
martinzink: They have different license terms based on the type we are using. 
https://docs.splunk.com/Documentation/Splunk/8.2.4/Admin/TypesofSplunklicenses
The whole eula can be found here: https://www.splunk.com/eula/sii/1.4


I changed this section in https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/b79afdea234a890818fa39c6a5ded4fb5fc117d8 so we explicity accept the free license which can be used to for these kinds of tests(we only use this container for the integration tests) more about it here: https://docs.splunk.com/Documentation/Splunk/8.2.4/Admin/TypesofSplunklicenses#Free_license

extensions/splunk/PutSplunkHTTP.h
```diff
@@ -0,0 +1,54 @@
...
+
+#include <string>
+#include <memory>
+
+#include "SplunkHECProcessor.h"
+#include "utils/gsl.h"
+
+namespace org::apache::nifi::minifi::extensions::splunk {
+
+class PutSplunkHTTP final : public SplunkHECProcessor {
```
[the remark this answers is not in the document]
martinzink: fixed in https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/25e4878262c6ec80237339cd7388e7caab830ea0#diff-fffe34ee91301ca41f9e7bc23592fde7ae3205eb7e6e7c0c4e796694b5babac6R51-R54

libminifi/include/utils/TimeUtil.h
```diff
@@ -37,6 +37,24 @@ namespace minifi {
 namespace utils {
 namespace timeutils {
 
+/**
+ * Converts the time point to the elapsed time since epoch
+ * @returns TimeUnit since epoch
+ */
+template<typename TimeUnit, typename TimePoint>
+uint64_t getTimeStamp(const TimePoint& time_point) {
```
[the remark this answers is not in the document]
martinzink: nice catch, changed this and the usages in https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/25e4878262c6ec80237339cd7388e7caab830ea0#diff-4a76905d55704437ae0a7a4ee434a73f057c105c97bdf8756f5747b4fbc65e9fR45

extensions/splunk/SplunkHECProcessor.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+  if (!context->getProperty(Token.getName(), token_))
+    throw Exception(PROCESS_SCHEDULE_EXCEPTION, "Failed to get Token");
+
+  if (!context->getProperty(SplunkRequestChannel.getName(), request_channel_))
+    throw Exception(PROCESS_SCHEDULE_EXCEPTION, "Failed to get SplunkRequestChannel");
+}
+
+std::string SplunkHECProcessor::getUrl() const {
+  return hostname_ + ":" + port_;
```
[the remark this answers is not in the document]
martinzink: changed it in https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/25e4878262c6ec80237339cd7388e7caab830ea0#diff-fffe34ee91301ca41f9e7bc23592fde7ae3205eb7e6e7c0c4e796694b5babac6R56

extensions/splunk/PutSplunkHTTP.cpp
martinzink: changed it in https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/25e4878262c6ec80237339cd7388e7caab830ea0#diff-2633ef573b024e894869a6a974a55671c3468db3eff99e4cdc646a081a700efdR145

extensions/splunk/PutSplunkHTTP.cpp
martinzink: good idea changed it in https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/25e4878262c6ec80237339cd7388e7caab830ea0#diff-2633ef573b024e894869a6a974a55671c3468db3eff99e4cdc646a081a700efdR73

docker/Dockerfile
martinzink: fixed in https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/25e4878262c6ec80237339cd7388e7caab830ea0#diff-f34da55ca08f1a30591d8b0b3e885bcc678537b2a9a4aadea4f190806b374ddcR117

extensions/splunk/QuerySplunkIndexingStatus.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+const core::Relationship QuerySplunkIndexingStatus::Failure("failure",
+    "A FlowFile is transferred to this relationship when the acknowledgement was not successful due to errors during the communication, "
+    "or if the flowfile was missing the acknowledgement id");
+
+void QuerySplunkIndexingStatus::initialize() {
+  SplunkHECProcessor::initialize();
+  setSupportedRelationships({Acknowledged, Unacknowledged, Undetermined, Failure});
+  updateSupportedProperties({MaximumWaitingTime, MaxQuerySize});
+}
```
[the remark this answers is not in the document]
martinzink: Changed it in 

https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/25e4878262c6ec80237339cd7388e7caab830ea0#diff-a2db2ff59dd1ebf5f1e3e053781c55708df909e3009ee21015da3fe04218def4R66 

https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/25e4878262c6ec80237339cd7388e7caab830ea0#diff-2633ef573b024e894869a6a974a55671c3468db3eff99e4cdc646a081a700efdR64

docker/test/integration/minifi/core/SplunkContainer.py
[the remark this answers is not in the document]
szaszm: This is non-free software, but I think it's fine as long as we are not shipping any of it.

extensions/splunk/tests/PutSplunkHTTPTests.cpp
```diff
@@ -0,0 +1,78 @@
...
+  read_from_failure->setAutoTerminatedRelationships({ReadFromFlowFileTestProcessor::Success});
+
+  plan->setProperty(put_splunk_http, PutSplunkHTTP::Hostname.getName(), "localhost");
+  plan->setProperty(put_splunk_http, PutSplunkHTTP::Port.getName(), mock_splunk_hec.getPort());
+  plan->setProperty(put_splunk_http, PutSplunkHTTP::Token.getName(), MockSplunkHEC::TOKEN);
+  plan->setProperty(put_splunk_http, PutSplunkHTTP::SplunkRequestChannel.getName(), "a12254b4-f481-435d-896d-3b6033eabe58");
+
+  write_to_flow_file->setContent("foobar");
+
+  SECTION("Happy path") {
```
lordgamez, line 54: Could we add some additional tests for checking non-default endpoint parameters? Also checking if the content type is set correctly with the default and a non-default value. I think with this great MockSplunkHEC we have a lot of options :)

extensions/splunk/SplunkHECProcessor.h
```diff
@@ -0,0 +1,65 @@
...
+
+  explicit SplunkHECProcessor(const std::string& name, const utils::Identifier& uuid = {})
+      : Processor(name, uuid) {
+  }
+  ~SplunkHECProcessor() override = default;
+
+  void initialize() override;
+  void onSchedule(const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSessionFactory> &sessionFactory) override;
+
+  bool isSingleThreaded() const override {
```
lordgamez, line 47: What's the reason for not support multithreading for this processor?

extensions/splunk/tests/QuerySplunkIndexingStatusTests.cpp
```diff
@@ -0,0 +1,123 @@
...
+    std::this_thread::sleep_for(std::chrono::milliseconds(100ms));
+    test_controller.runSession(plan);
+
+    CHECK(read_from_failure->numberOfFlowFilesRead() == 0);
+    CHECK(read_from_undetermined->numberOfFlowFilesRead() == 1);
+    CHECK(read_from_unacknowledged->numberOfFlowFilesRead() == 0);
+    CHECK(read_from_acknowledged->numberOfFlowFilesRead() == 0);
+  }
+
+  SECTION("Querying not indexed old id") {
```
lordgamez, line 99: Could we add a test for verifying the MaxQuerySize?

extensions/splunk/QuerySplunkIndexingStatus.cpp
```diff
@@ -0,0 +1,193 @@
...
+  }
+  rapidjson::StringBuffer buffer;
+  rapidjson::Writer<rapidjson::StringBuffer> writer(buffer);
+  payload.Accept(writer);
+  return buffer.GetString();
+}
+
+void getIndexingStatusFromSplunk(utils::HTTPClient& client, std::unordered_map<uint64_t, FlowFileWithIndexStatus>& undetermined_flow_files) {
+  rapidjson::Document response;
+  if (client.submit()) {
```
lordgamez, line 127: I would go for an early here, I just don't like long ifs :)

libminifi/include/utils/TimeUtil.h
```diff
@@ -37,6 +37,24 @@ namespace minifi {
...
+}
+
+/**
+ * Converts the time since epoch into a time point
+ * @returns the time point matching the input timestamp
+ */
+template<typename TimeUnit, typename ClockType>
+std::chrono::time_point<ClockType> getTimePoint(uint64_t timestamp) {
+  return std::chrono::time_point<ClockType>() + TimeUnit(timestamp);
+}
```
szaszm, line 56: I criticized these in your other PR, which would apply here, too. If you think that they are useful, feel free to reintroduce them there, but if they don't improve the readability, you may want to consider removing them.

extensions/splunk/PutSplunkHTTP.cpp
```diff
@@ -0,0 +1,179 @@
...
+  payload_callback_obj.pos = 0;
+  client.appendHeader("Content-Length", std::to_string(flow_file->getSize()));
+
+  client.setUploadCallback(&payload_callback_obj);
+  client.setSeekFunction(&payload_callback_obj);
+
+  auto content_type = getContentType(context, *flow_file);
+  if (content_type.has_value())
+    client.setContentType(content_type.value());
+}
```
szaszm, line 148: This if statement looks a bit confusing with the multiple closing braces after it. Consider enclosing the if body in braces, or you can use optional map as well.

extensions/splunk/PutSplunkHTTP.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+  return addAttributesFromClientResponse(flow_file, client) && client.getResponseCode() == 200;
+}
+
+void setFlowFileAsPayload(core::ProcessSession& session,
+                                         core::ProcessContext& context,
+                                         utils::HTTPClient& client,
+                                         const gsl::not_null<std::shared_ptr<core::FlowFile>>& flow_file,
+                                         utils::ByteInputCallBack& payload_callback,
+                                         utils::HTTPUploadCallback& payload_callback_obj) {
```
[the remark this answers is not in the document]
martinzink: good catch, I gotta be more careful with renaming things in the last minute :+1: 
fixed in https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/a0e996b4844c92dd3d2cc4d70955ad4ea5697f31#diff-2633ef573b024e894869a6a974a55671c3468db3eff99e4cdc646a081a700efdR130-R134

extensions/splunk/PutSplunkHTTP.cpp
```diff
@@ -0,0 +1,179 @@
...
+    context->yield();
+    return;
+  }
+  auto flow_file = gsl::not_null(std::move(ff));
+
+  utils::HTTPClient client(getNetworkLocation() + getEndpoint(*context, flow_file), getSSLContextService(*context));
+  setHeaders(client);
+
+  auto payload_callback = std::make_unique<utils::ByteInputCallBack>();
+  auto payload_callback_obj = std::make_unique<utils::HTTPUploadCallback>();
```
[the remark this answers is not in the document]
martinzink: :+1: nothing, it was due to copy paste from InvokeHTTP...
fixed in https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/a0e996b4844c92dd3d2cc4d70955ad4ea5697f31#diff-2633ef573b024e894869a6a974a55671c3468db3eff99e4cdc646a081a700efdR162-R163

extensions/splunk/tests/QuerySplunkIndexingStatusTests.cpp
```diff
@@ -0,0 +1,123 @@
...
+#include "processors/UpdateAttribute.h"
+#include "ReadFromFlowFileTestProcessor.h"
+#include "WriteToFlowFileTestProcessor.h"
+#include "utils/TimeUtil.h"
+
+using QuerySplunkIndexingStatus = org::apache::nifi::minifi::extensions::splunk::QuerySplunkIndexingStatus;
+using ReadFromFlowFileTestProcessor = org::apache::nifi::minifi::processors::ReadFromFlowFileTestProcessor;
+using WriteToFlowFileTestProcessor = org::apache::nifi::minifi::processors::WriteToFlowFileTestProcessor;
+using UpdateAttribute = org::apache::nifi::minifi::processors::UpdateAttribute;
+using namespace std::chrono_literals;  // NOLINT(build/namespaces)
```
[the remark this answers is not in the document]
martinzink: fixed in https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/a0e996b4844c92dd3d2cc4d70955ad4ea5697f31#diff-84834c115fc3bbd69a5501ad9715f71721ee1fb6fa3b87dcdd107807ce004da2R33

extensions/splunk/tests/QuerySplunkIndexingStatusTests.cpp
martinzink: sure thing, added it in https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/e130d7dcf4f90e19cd36c46a5742a1f123d18eb7

extensions/splunk/tests/PutSplunkHTTPTests.cpp
martinzink: sure thing, added it in e130d7d

libminifi/include/utils/TimeUtil.h
martinzink: sure thing :+1: , I removed these in https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/a0e996b4844c92dd3d2cc4d70955ad4ea5697f31#diff-4a76905d55704437ae0a7a4ee434a73f057c105c97bdf8756f5747b4fbc65e9fL40-L57

extensions/splunk/tests/PutSplunkHTTPTests.cpp
martinzink: I'll add tests for the content_type aswell :eyes:

extensions/splunk/PutSplunkHTTP.cpp
martinzink: good idea, changed it in https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/a0e996b4844c92dd3d2cc4d70955ad4ea5697f31#diff-2633ef573b024e894869a6a974a55671c3468db3eff99e4cdc646a081a700efdR143

docker/test/integration/MiNiFi_integration_test_driver.py
```diff
@@ -60,6 +60,12 @@ def start_kafka_broker(self):
         self.cluster.deploy('kafka-broker')
         assert self.wait_for_container_startup_to_finish('kafka-broker')
 
+    def start_splunk(self):
+        self.cluster.acquire_container('splunk', 'splunk')
+        self.cluster.deploy('splunk')
+        assert self.wait_for_container_startup_to_finish('splunk')
+        assert self.cluster.enable_hec_indexer('splunk', 'splunk_hec_token')
```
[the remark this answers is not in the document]
lordgamez: It could be avoided if we would keep the source file and repeatedly rerun the flow until it succeeds, but maybe this solution makes the test case more stable and reliable. I'm okay with keeping it like this.

extensions/splunk/tests/PutSplunkHTTPTests.cpp
martinzink: Added tests for checking the content-type 
it was actually bugged due to a typo so thanks for suggesting the test :+1: 
https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/ee52021fd2a471abb9504282a4bddf2591186cb6

extensions/splunk/QuerySplunkIndexingStatus.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+#include <unordered_map>
+#include <utility>
+
+#include "SplunkAttributes.h"
+
+#include "core/Resource.h"
+#include "client/HTTPClient.h"
+#include "utils/HTTPClient.h"
+#include "utils/TimeUtil.h"
```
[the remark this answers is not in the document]
szaszm: I mean the include in all 3 cases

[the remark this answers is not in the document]
martinzink: removed it in https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/5021417766c08d4c98026e26dae85155ef28e409

extensions/splunk/PutSplunkHTTP.cpp
```diff
@@ -0,0 +1,177 @@
...
+#include <vector>
+#include <utility>
+
+#include "SplunkAttributes.h"
+
+#include "core/Resource.h"
+#include "utils/StringUtils.h"
+#include "client/HTTPClient.h"
+#include "utils/HTTPClient.h"
+#include "utils/TimeUtil.h"
```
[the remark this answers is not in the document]
martinzink: removed it in https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/5021417766c08d4c98026e26dae85155ef28e409

extensions/splunk/tests/QuerySplunkIndexingStatusTests.cpp
```diff
@@ -0,0 +1,134 @@
...
+#include <chrono>
+
+#include "QuerySplunkIndexingStatus.h"
+#include "MockSplunkHEC.h"
+#include "SplunkAttributes.h"
+#include "TestBase.h"
+#include "processors/UpdateAttribute.h"
+#include "ReadFromFlowFileTestProcessor.h"
+#include "WriteToFlowFileTestProcessor.h"
+#include "utils/TimeUtil.h"
```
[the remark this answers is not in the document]
martinzink: removed it in https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/5021417766c08d4c98026e26dae85155ef28e409

docker/test/integration/features/splunk.feature
```diff
@@ -0,0 +1,24 @@
+Feature: Sending data to Splunk HEC using PutSplunkHTTP
+
+  Background:
+    Given the content of "/tmp/output" is monitored
+
+  Scenario: A MiNiFi instance transfers data to a Splunk HEC
```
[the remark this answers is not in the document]
martinzink: Good idea, I've added a docker test where SSL is enabled on both minifi and splunk in https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/2cec29069f5d2f4e4bbb062d38eaf6742483ff00

docker/requirements.txt
```diff
@@ -6,3 +6,4 @@ confluent-kafka==1.7.0
 PyYAML==5.4.1
 m2crypto==0.37.1
 watchdog==2.1.2
+pyopenssl
```
lordgamez, line 9: Please specify version number of the package to avoid possible future problems with newer versions.

lordgamez (approved):
LGTM, thanks! Only 1 minor comment

docker/requirements.txt
martinzink: specified it in https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/e77a2891da3ba880f1a24b8a74757d572dc2ccde

szaszm (approved):

PROCESSORS.md
```diff
@@ -50,7 +50,9 @@
 - [PutS3Object](#puts3object)
 - [PutSFTP](#putsftp)
 - [PutSQL](#putsql)
+- [PutSplunkHTTP](#putsplunkhttp)
```
adamdebreceni, line 53: nitpick: this might not be in alphabetical order

bstrp_functions.sh
```diff
@@ -445,6 +446,7 @@ read_feature_options(){
     x) ToggleFeature AZURE_ENABLED ;;
     y) ToggleFeature SYSTEMD_ENABLED ;;
     z) ToggleFeature NANOFI_ENABLED ;;
+    aa) ToggleFeature SPLUNK_ENABLED ;;
```
adamdebreceni, line 449: we should update the usage message printed (`Enter choice [ A - Z or 1-7 ]`) to reflect the addition of `aa` option

extensions/splunk/PutSplunkHTTP.cpp
```diff
@@ -0,0 +1,176 @@
...
+    ->supportsExpressionLanguage(true)->build());
+
+const core::Property PutSplunkHTTP::ContentType(core::PropertyBuilder::createProperty("Content Type")
+    ->withDescription("The media type of the event sent to Splunk. If not set, \"mime.type\" flow file attribute will be used. "
+                      "In case of neither of them is specified, this information will not be sent to the server.")
+    ->supportsExpressionLanguage(true)->build());
+
+
+const core::Relationship PutSplunkHTTP::Success("success", "FlowFiles that are sent successfully to the destination are sent to this relationship.");
+const core::Relationship PutSplunkHTTP::Failure("failure", "FlowFiles that failed to send to the destination are sent to this relationship.");
```
adamdebreceni, line 60: nitpick: shouldn't this be "FlowFiles that failed to be sent to ..."? (I'm not really sure about this)

extensions/splunk/PutSplunkHTTP.cpp
```diff
@@ -0,0 +1,176 @@
...
+  if (parse_result.IsError())
+    return false;
+
+  if (response_json.HasMember("code") && response_json["code"].IsInt())
+    flow_file.addAttribute(SPLUNK_RESPONSE_CODE, std::to_string(response_json["code"].GetInt()));
+  else
+    result = false;
+
+  if (response_json.HasMember("ackId") && response_json["ackId"].IsUint64())
+    flow_file.addAttribute(SPLUNK_ACK_ID, std::to_string(response_json["ackId"].GetUint64()));
```
adamdebreceni, line 114: not sure if it is intentional but `addAttribute` only adds the attribute if it does not exist on the flowfile

extensions/splunk/PutSplunkHTTP.cpp
```diff
@@ -0,0 +1,176 @@
...
+    parameters.push_back("sourcetype=" + prop_value);
+  }
+  if (context.getProperty(PutSplunkHTTP::Source, prop_value, flow_file)) {
+    parameters.push_back("source=" + prop_value);
+  }
+  if (context.getProperty(PutSplunkHTTP::Host, prop_value, flow_file)) {
+    parameters.push_back("host=" + prop_value);
+  }
+  if (context.getProperty(PutSplunkHTTP::Index, prop_value, flow_file)) {
+    parameters.push_back("index=" + prop_value);
```
adamdebreceni, line 93: as we don't sanitize the parameters they should not contain reserved characters, we should either document this, or check ourselves and throw an error (possibly through an URIComponentValidator, according to [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) non-reserved characters are `A-Z a-z 0-9 - _ . ! ~ * ' ( )`)

extensions/splunk/QuerySplunkIndexingStatus.cpp
```diff
@@ -0,0 +1,191 @@
...
+#include "rapidjson/document.h"
+#include "rapidjson/stringbuffer.h"
+#include "rapidjson/writer.h"
+
+namespace org::apache::nifi::minifi::extensions::splunk {
+
+const core::Property QuerySplunkIndexingStatus::MaximumWaitingTime(core::PropertyBuilder::createProperty("Maximum Waiting Time")
+    ->withDescription("The maximum time the processor tries to acquire acknowledgement confirmation for an index, from the point of registration. "
+                      "After the given amount of time, the processor considers the index as not acknowledged and transfers the FlowFile to the \"unacknowledged\" relationship.")
+    ->withDefaultValue("1 hour")->isRequired(true)->build());
```
adamdebreceni, line 39: we could specify the validator with `withDefaultValue<core::TimePeriodValue>("1 hour")` and get feedback on a malformed property at yml parsing time

extensions/splunk/QuerySplunkIndexingStatus.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+const core::Property QuerySplunkIndexingStatus::MaxQuerySize(core::PropertyBuilder::createProperty("Maximum Query Size")
+    ->withDescription("The maximum number of acknowledgement identifiers the outgoing query contains in one batch. "
+                      "It is recommended not to set it too low in order to reduce network communication.")
+    ->withDefaultValue("1000")->isRequired(true)->build());
```
adamdebreceni, line 44: we could specify the type with `withDefaultValue<uint64_t>(1000)` here as well

extensions/splunk/QuerySplunkIndexingStatus.cpp
```diff
@@ -0,0 +1,191 @@
...
+}
+
+void QuerySplunkIndexingStatus::onSchedule(const std::shared_ptr<core::ProcessContext>& context, const std::shared_ptr<core::ProcessSessionFactory>& sessionFactory) {
+  gsl_Expects(context && sessionFactory);
+  SplunkHECProcessor::onSchedule(context, sessionFactory);
+  std::string max_wait_time_str;
+  if (context->getProperty(MaximumWaitingTime.getName(), max_wait_time_str)) {
+    core::TimeUnit unit;
+    uint64_t max_wait_time;
+    if (core::Property::StringToTime(max_wait_time_str, max_wait_time, unit) && core::Property::ConvertTimeUnitToMS(max_wait_time, unit, max_wait_time)) {
```
adamdebreceni, line 75: I know this is how it's done in lot of places, but I think if we specify the validator we can directly query the milliseconds with `getProperty<uint64_t>(MaximumWaitingTime)`

extensions/splunk/QuerySplunkIndexingStatus.cpp
```diff
@@ -0,0 +1,191 @@
...
+    auto flow = session.get();
+    if (flow == nullptr)
+      break;
+    std::optional<std::string> splunk_ack_id_str = flow->getAttribute(SPLUNK_ACK_ID);
+    if (!splunk_ack_id_str.has_value()) {
+      session.transfer(flow, QuerySplunkIndexingStatus::Failure);
+      continue;
+    }
+    uint64_t splunk_ack_id = std::stoull(splunk_ack_id_str.value());
+    undetermined_flow_files.emplace(std::make_pair(splunk_ack_id, gsl::not_null(std::move(flow))));
```
adamdebreceni, line 107: should we handle if there are multiple flowfiles with the same ack_id? (currently I think we do not specify what should happen to the abandoned flowfile and the whole session will get rolled back)

extensions/splunk/SplunkHECProcessor.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+namespace org::apache::nifi::minifi::extensions::splunk {
+
+const core::Property SplunkHECProcessor::Hostname(core::PropertyBuilder::createProperty("Hostname")
+    ->withDescription("The ip address or hostname of the Splunk server.")
+    ->isRequired(true)->build());
+
+const core::Property SplunkHECProcessor::Port(core::PropertyBuilder::createProperty("Port")
+    ->withDescription("The HTTP Event Collector HTTP Port Number.")
+    ->withDefaultValue("8088")->isRequired(true)->build());
```
adamdebreceni, line 30: we could specify the `PortValidator` here

extensions/splunk/tests/MockSplunkHEC.h
```diff
@@ -0,0 +1,214 @@
...
+ public:
+  static constexpr const char* TOKEN = "Splunk 822f7d13-2b70-4f8c-848b-86edfc251222";
+
+  static inline std::vector<uint64_t> indexed_events = {0, 1};
+
+  explicit MockSplunkHEC(std::string port) : port_(std::move(port)) {
+    std::vector<std::string> options;
+    options.emplace_back("listening_ports");
+    options.emplace_back(port_);
+    server_.reset(new CivetServer(options, &callbacks_, &logger_));
```
adamdebreceni, line 185: I think we need to initialize the civet library to safely use it like in TestServer.h

martinzink: good idea, fixed in https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/412fa041455d782eedc52bdfb24c680ae7e56808#diff-22bed17b8af51ae7e6dba78a65e7a6c2a0e66fe55431bf0d27ee8508055afaa6R176-R193

extensions/splunk/SplunkHECProcessor.cpp
martinzink: :+1:  added it in https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/412fa041455d782eedc52bdfb24c680ae7e56808#diff-8464594c14ae65a1b12e66a4dbe2e5afa51be34981fc286fcd9aac9bb5dee1f1R30

extensions/splunk/QuerySplunkIndexingStatus.cpp
martinzink: thats a valid point, in that case I think every flowfile with duplicated ack_id should be transferred to failure (because even if the server acknowledges we cant be sure which one is it)
https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/412fa041455d782eedc52bdfb24c680ae7e56808#diff-a2db2ff59dd1ebf5f1e3e053781c55708df909e3009ee21015da3fe04218def4R109-R119

extensions/splunk/PutSplunkHTTP.cpp
martinzink: Good point, I restructed these so the HTTPClient can escape these parameters. https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/412fa041455d782eedc52bdfb24c680ae7e56808#diff-2633ef573b024e894869a6a974a55671c3468db3eff99e4cdc646a081a700efdR84-R93

extensions/splunk/PutSplunkHTTP.cpp
martinzink: To be honest this was a shameless copy paste from the [nifi](https://github.com/apache/nifi/blob/main/nifi-nar-bundles/nifi-splunk-bundle/nifi-splunk-processors/src/main/java/org/apache/nifi/processors/splunk/PutSplunkHTTP.java#L133) version
But I agree, we should fix this https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/412fa041455d782eedc52bdfb24c680ae7e56808#diff-2633ef573b024e894869a6a974a55671c3468db3eff99e4cdc646a081a700efdR60

bstrp_functions.sh
martinzink: I am not sure what would be the correct way to phrase this...
How about this? https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/7a0401fdaf024ac39ec0b6f5cd2527c5c98762b4

PROCESSORS.md
martinzink: Reordered it in https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/412fa041455d782eedc52bdfb24c680ae7e56808#diff-fd2410931e7fdc4bf8b3ce23f5f7a27c7aacdf9337320626d86d806448c90b9bL52

extensions/splunk/PutSplunkHTTP.cpp
martinzink: Good catch, it wasnt intentional. I very well can image situations where we wanna retry after a failed PutSplunkHTTP.
fixed it in https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/412fa041455d782eedc52bdfb24c680ae7e56808#diff-2633ef573b024e894869a6a974a55671c3468db3eff99e4cdc646a081a700efdR122-R123

extensions/splunk/QuerySplunkIndexingStatus.cpp
martinzink: :+1:  done https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/412fa041455d782eedc52bdfb24c680ae7e56808#diff-a2db2ff59dd1ebf5f1e3e053781c55708df909e3009ee21015da3fe04218def4R40

extensions/splunk/QuerySplunkIndexingStatus.cpp
martinzink: :+1: done https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/412fa041455d782eedc52bdfb24c680ae7e56808#diff-a2db2ff59dd1ebf5f1e3e053781c55708df909e3009ee21015da3fe04218def4R45

extensions/splunk/QuerySplunkIndexingStatus.cpp
martinzink: The TimePeriodValue is being refactored in #1225 
I think this will merge sooner than that, how about I fix this issue there?

extensions/splunk/PutSplunkHTTP.cpp
adamdebreceni: curl providing an utility is neat, but according to the [documentation](https://curl.se/libcurl/c/curl_easy_escape.html) the `char*` returned should be freed using `curl_free` (instead we initialize an `std::string` with it and leak it) could you fix that as well?

extensions/splunk/QuerySplunkIndexingStatus.cpp
adamdebreceni: if we have 3 flowfiles with the same id, the first one is emplaced in the map, the second one is detected routed to failure and the id is placed in the vector, the third one is detected routed to failure and the id is placed in the vector
iterating through the vector encounters the same id twice, and the `std::unordered_map<>::at` will throw (as we have already erased the element)

extensions/splunk/tests/MockSplunkHEC.h
adamdebreceni: moving the `CivetLibrary` into its own header under `libminifi/test` might have its merits but I do not insist on it

extensions/splunk/PutSplunkHTTP.cpp
martinzink: good catch, since it was already provided by our HTTPClient class I didnt think about this
fixed it and also added tests (so it could be checked with valgrind)
https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/fe26bccb340e9ac50052111656f3c0df4006812d

extensions/splunk/QuerySplunkIndexingStatus.cpp
martinzink: :man_facepalming: you are absolutely right, replaced the vector with unordered_set, and added a test where this scenario is also checked
https://github.com/apache/nifi-minifi-cpp/pull/1219/commits/fe26bccb340e9ac50052111656f3c0df4006812d

extensions/http-curl/client/HTTPClient.cpp
```diff
@@ -212,7 +212,10 @@ void HTTPClient::setContentType(std::string content_type) {
 }
 
 std::string HTTPClient::escape(std::string string_to_escape) {
-  return curl_easy_escape(http_session_, string_to_escape.c_str(), gsl::narrow<int>(string_to_escape.length()));
+  char* escaped_chars = curl_easy_escape(http_session_, string_to_escape.c_str(), gsl::narrow<int>(string_to_escape.length()));
+  std::string escaped_string(escaped_chars);
+  curl_free(escaped_chars);
+  return escaped_string;
```
szaszm, line 218: In theory string's constructor can throw which would result in leaking `escaped_chars`. Consider using `unique_ptr` with a custom curl deleter.
```suggestion
  struct curl_deleter { void operator()(void* p) noexcept { curl_free(p); } };
  std::unique_ptr<char, curl_deleter> escaped_chars{curl_easy_escape(http_session_, string_to_escape.c_str(), gsl::narrow<int>(string_to_escape.length()))};
  std::string escaped_string(escaped_chars.get());
  return escaped_string;
```

adamdebreceni (approved):

PR #1220, opened 2021-11-29
MINIFICPP-1694 Remove false positive error log messages

lordgamez:
Several users have encountered a few error messages while debugging unrelated issues. These error messages were shown during startup and did not actually signal any problems, but they were misleading enough to suspect some relation to the issues being debugged. These error messages were removed or changed to warnings with some additional information to be more elaborate about their nature.

- For controllers both class and type fields are accepted without any error messages
- When a configuration file does not exist like the default bootstrap file, instead of the previous `load configure file failed` we get a warning with the message `Configuration file './conf/bootstrap.conf' does not exist, so it could not be loaded.`
- When a rocksdb key cannot be retrieved when it is probably not initialized yet we only get a warning

https://issues.apache.org/jira/browse/MINIFICPP-1694

----------------------------------------------------------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/core/yaml/CheckRequiredField.h
+++ b/libminifi/include/core/yaml/CheckRequiredField.h
@@ -19,6 +19,7 @@
 
 #include <string>
 #include <memory>
+#include <vector>
 
 #include "core/logging/LoggerConfiguration.h"
 #include "yaml-cpp/yaml.h"
@@ -31,6 +32,7 @@ namespace core {
 namespace yaml {
 
 bool isFieldPresent(const YAML::Node *yamlNode, const std::string &fieldName);
+std::string buildErrorMessage(const YAML::Node *yamlNode, const std::vector<std::string>& alternateFieldNames, const std::string &yamlSection = "");
 
 /**
  * This is a helper function for verifying the existence of a required
@@ -51,7 +53,23 @@ bool isFieldPresent(const YAML::Node *yamlNode, const std::string &fieldName);
  *                               not present in 'yamlNode'
  */
 void checkRequiredField(
-    const YAML::Node *yamlNode, const std::string &fieldName, const std::shared_ptr<logging::Logger>& logger, const std::string &yamlSection = "", const std::string &errorMessage = "");
+    const YAML::Node *yamlNode, const std::string &fieldName, const std::shared_ptr<logging::Logger>& logger, const std::string &yamlSection = "", std::string errorMessage = "");
+
+template<typename T>
+T parseRequiredField(const YAML::Node* yamlNode,
+    const std::vector<std::string>& alternateNames, const std::shared_ptr<logging::Logger>& logger,
+    const std::string &yamlSection, std::function<T(const YAML::Node&)> parser, std::string errorMessage = "") {
+  for (const auto& name : alternateNames) {
+    if (yaml::isFieldPresent(yamlNode, name)) {
+      return parser((*yamlNode)[name]);
+    }
+  }
+  if (errorMessage.empty()) {
+    errorMessage = buildErrorMessage(yamlNode, alternateNames, yamlSection);
+  }
+  logger->log_error(errorMessage.c_str());
+  throw std::invalid_argument(errorMessage);
+}
 
 }  // namespace yaml
 }  // namespace core
--- a/libminifi/include/core/yaml/CheckRequiredField.h
+++ b/libminifi/include/core/yaml/CheckRequiredField.h
@@ -30,6 +30,8 @@ namespace minifi {
 namespace core {
 namespace yaml {
 
+bool isFieldPresent(const YAML::Node *yamlNode, const std::string &fieldName);
+
 /**
  * This is a helper function for verifying the existence of a required
  * field in a YAML::Node object. If the field is not present, an error
--- a/libminifi/include/core/yaml/CheckRequiredField.h
+++ b/libminifi/include/core/yaml/CheckRequiredField.h
@@ -31,8 +31,8 @@ namespace minifi {
 namespace core {
 namespace yaml {
 
-bool isFieldPresent(const YAML::Node *yamlNode, const std::string &fieldName);
-std::string buildErrorMessage(const YAML::Node *yamlNode, const std::vector<std::string>& alternateFieldNames, const std::string &yamlSection = "");
+bool isFieldPresent(const YAML::Node &yaml_node, std::string_view field_name);
+std::string buildErrorMessage(const YAML::Node &yaml_node, const std::vector<std::string> &alternate_field_names, std::string_view yaml_section = "");
 
 /**
  * This is a helper function for verifying the existence of a required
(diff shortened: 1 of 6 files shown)


fgerlits (approved):
Thanks for fixing this!  Those non-error errors were annoying.

libminifi/src/core/yaml/YamlConfiguration.cpp
```diff
@@ -529,14 +529,20 @@ void YamlConfiguration::parseControllerServices(const YAML::Node& controllerServ
...
-      } catch (const std::invalid_argument &) {
-        yaml::checkRequiredField(&controllerServiceNode, "type", logger_, CONFIG_YAML_CONTROLLER_SERVICES_KEY);
+      } else {
+        const YAML::Node name_node = controllerServiceNode.as<YAML::Node>()["name"];
+        std::string err_msg =
+          name_node ?
+            "Unable to parse configuration file for component named '" + name_node.as<std::string>() + "' as required field 'class' or 'type' is missing" :
+            "Unable to parse configuration file as required field 'class' or 'type' is missing";
+        err_msg += " [in '" + std::string(CONFIG_YAML_CONTROLLER_SERVICES_KEY) + "' section of configuration file]";
+        yaml::checkRequiredField(&controllerServiceNode, "type", logger_, CONFIG_YAML_CONTROLLER_SERVICES_KEY, err_msg);
```
adamdebreceni, line 541: it seems we have duplicated the error message from `checkRequiredField`, should we maybe create a `checkRequiredField` overload that takes a list of alternative keys (here {"class", "type"})? we could even add a "parser" parameter, and use it as

```
template<typename T>
T parseRequiredField(const YAML::Node* yamlNode, const std::vector<std::string>& alternateNames, const std::shared_ptr<logging::Logger>& logger, const std::string &yamlSection, std::function<T(const YAML::Node&)> parser, const std::string& errorMessage = "") {...}

std::string type = parseRequiredField(&controllerServiceNode, {"class", "type"}, logger_, CONFIG_YAML_CONTROLLER_SERVICES_KEY, [] (const YAML::Node& node) {return node.as<std::string>();});
```

lordgamez: Good idea, added in 31e4679c2f41a3f2a788132ca34a77fedf5d9ec7

adamdebreceni (approved):

libminifi/include/core/yaml/CheckRequiredField.h
```diff
@@ -49,7 +53,23 @@ namespace yaml {
...
+    if (yaml::isFieldPresent(yamlNode, name)) {
+      return parser((*yamlNode)[name]);
+    }
+  }
+  if (errorMessage.empty()) {
+    errorMessage = buildErrorMessage(yamlNode, alternateNames, yamlSection);
+  }
+  logger->log_error(errorMessage.c_str());
+  throw std::invalid_argument(errorMessage);
+}
```
szaszm, line 72: 1. Most of this logic could go to the implementation file, since the only thing requiring it to be a template is the `parser` call. Extract the logic to something like `const YAML::Node& getRequiredField(const YAML::Node& yaml_node, gsl::span<std::string_view> alternate_names, logging::Logger& logger, std::string_view yaml_section, std::string_view error_message = {});`
2. The passed strings could be string_view instead to not force allocations for all strings that are passed in. The collection could be a `gsl::span` to avoid the allocation from `std::vector`.
3. The pointer parameter doesn't seem to be nullable, so it should be a reference instead.
4. If it's a template anyway, there's no need to type-erase the `parser` function
5. If logging a string, use `"%s"`, so that the logging doesn't misbehave when the error message happens to contain format specifiers.
6. One of log or throw should normally be enough, so the error handling (catch) code can decide whether to log the error or not. Maybe it could recover.
I changed the parameter names to snake_case in the suggestion, because I like them better, but it's mixed in the codebase, so I don't insist on changing that.
```suggestion
// in the cpp file
const YAML::Node& getRequiredField(const YAML::Node& yaml_node, gsl::span<std::string_view> alternate_names, logging::Logger& logger, std::string_view yaml_section, std::string error_message) {
  for (const auto& name : alternate_names) {
    if (yaml::isFieldPresent(yaml_node, name)) {
      return parser((*yaml_node)[name]);
    }
  }
  if (error_message.empty()) {
    error_message = buildErrorMessage(yaml_node, alternate_names, yaml_section);
  }
  // logger->log_error("%s", error_message);
  throw std::invalid_argument(error_message);
}

// in the header
const YAML::Node& getRequiredField(const YAML::Node& yaml_node, gsl::span<std::string_view> alternate_names, logging::Logger& logger, std::string_view yaml_section, std::string error_message = {});


template<typename T, typename F>
auto parseRequiredField(const YAML::Node& yaml_node, gsl::span<std::string_view> alternate_names, logging::Logger& logger, std::string_view yaml_section, F parser, std::string error_message = {})
    -> std::enable_if_t<std::is_convertible_v<decltype(parser(std::declval<const YAML::Node&>())), T>, T> {
  return parser(getRequiredField(yaml_node, alternate_names, logger, yaml_section, std::move(error_message)));
}

// or alternatively automatically deduce the return type based on the parser function
template<typename F>
auto parseRequiredField(const YAML::Node& yaml_node, gsl::span<std::string_view> alternate_names, logging::Logger& logger, std::string_view yaml_section, F parser, std::string error_message = {})
    /* optional explicit return type: -> decltype(parser(std::declval<const YAML::Node&>())) */ {
  return parser(getRequiredField(yaml_node, alternate_names, logger, yaml_section, std::move(error_message)));
}
```

adamdebreceni: I think we might as well drop the `parseRequiredField` and keep the `getRequiredField`, we wouldn't need to template anything and could move all implementations to the cpp file

`getRequiredField(...).as<std::string>()` even looks better than `parseRequiredField(..., [] (const auto& node) {return node.as<std::string>()}` IMO

libminifi/src/core/yaml/CheckRequiredField.cpp
```diff
@@ -27,28 +28,48 @@ namespace minifi {
...
+}
+
+std::string getRequiredField(const YAML::Node &yaml_node, const std::vector<std::string> &alternate_names, std::string_view yaml_section, std::string_view error_message) {
+  for (const auto& name : alternate_names) {
+    if (yaml::isFieldPresent(yaml_node, name)) {
+      return yaml_node[name].as<std::string>();
+    }
+  }
+  if (error_message.empty()) {
+    std::invalid_argument(buildErrorMessage(yaml_node, alternate_names, yaml_section));
```
szaszm, line 70: I think you meant to `throw` this.

libminifi/include/core/yaml/CheckRequiredField.h
lordgamez: Thanks for your comments, I appreciate the effort you put into them. I updated most of the points listed, unfortunately I could not avoid allocating the vector used for the alternate name so I didn't use `gsl::span` and also out implementation of `StringUtils::join` does not work with `std::string_view`s so I had to fallback to strings.

libminifi/src/core/yaml/CheckRequiredField.cpp
lordgamez: Fixed in e6d61d1d6b803061bc6dd60ec9800109e789efda

szaszm (approved):

libminifi/src/core/yaml/CheckRequiredField.cpp
```diff
@@ -27,28 +28,48 @@ namespace minifi {
...
+    if (error_message.empty()) {
+      error_message = buildErrorMessage(yaml_node, std::vector<std::string>{std::string(field_name)}, yaml_section);
     }
-    logger->log_error(errMsg.c_str());
-    throw std::invalid_argument(errMsg);
+    throw std::invalid_argument(error_message);
+  }
+}
+
+std::string getRequiredField(const YAML::Node &yaml_node, const std::vector<std::string> &alternate_names, std::string_view yaml_section, std::string error_message) {
```
adamdebreceni, line 63: I think we should return the node ref instead, converting it into a string seems something the caller should decide to do

szaszm: I think it's fine as a string if it's only ever used as a string, but I'm fine with returning a node ref as well.

adamdebreceni: the only place I could find that doesn't use a required field as a string is the `Input Ports` in RPGs, so I'm fine with leaving it to return a string until more use-cases arise

adamdebreceni: oh yeah, sorry `Node::operator[]` returns a `Node` not a ref to some preconstructed object, obv the ref return type doesn't work in this case

PR #1222, opened 2021-12-02
MINIFICPP-1695 Fix execution of native python processors

lordgamez:
PR https://github.com/apache/nifi-minifi-cpp/pull/1126 fixed the execution of ExecutePythonProcessor processor, but unfortunately broke the execution of native python processors. The ExecutePythonProcessor implements a solution now that can host the native python processors but can also be run individually with a script property parameter.

------------------------------------------------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/script/python/ExecutePythonProcessor.cpp
```diff
@@ -54,35 +54,62 @@ core::Relationship ExecutePythonProcessor::Success("success", "Script successes"
...
-    ScriptBody,
-    ModuleDirectory
-  });
-  setAcceptAllProperties();
-  setSupportedRelationships({
-    Success,
-    Failure
-  });
-}
+  if (getProperties().empty()) {
```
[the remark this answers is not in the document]
lordgamez: This is needed because in case of native python processors initialization is actually called twice. Once normally as for every other processor and once before that in the `PythonObjectFactory` when creating the native processor. In the first call the `ScriptFile` property is set to be used for the native python processor, which can provide description and initialization for the processor still in the `onInitialize` phase. Unfortunately the operation of declaring the properties is not idempotent, in this case it clears the value of the `ScriptFile` on the second `onInitialize` run that's why the check was needed.

szaszm (approved):

extensions/script/python/ExecutePythonProcessor.cpp
```diff
@@ -50,39 +50,72 @@ core::Property ExecutePythonProcessor::ModuleDirectory(core::PropertyBuilder::cr
...
-
   appendPathForImportModules();
-  loadScript();
 
-  if (script_to_exec_.empty()) {
-    throw std::runtime_error("Neither Script Body nor Script File is available to execute");
+  try {
+    loadScript();
+  } catch(const std::runtime_error&) {
+    logger_->log_warn("Could not load python script while initializing. In case of non-native python processor this is normal and will be done in the schedule phase.");
```
adamdebreceni, line 83: what makes a non-native processor fail here but succeed in the schedule?

lordgamez: Normally in the `ExecutePythonProcessor`'s initialize phase the values of the properties are not yet available like in any other processor. However when we use a native python processor, after the properties are initialized the `PythonObjectFactory` sets the `ScriptFile` property's value to the python processor's file path and calls the `initialize` again and the script will be loaded.

adamdebreceni: and a native processor's `onSchedule` won't be called, an initialization in `onSchedule` would be too late for them?

lordgamez: `onSchedule` runs for that as well, but as we need to call the native processor's `describe` before the `onSchedule` so we can provide manifest our processor identity on C2 as well.

adamdebreceni: I see, that makes sense, could you add this reason as a comment?

lordgamez: Sure, added comment in e32ad12483fe66bc7759237495e9922737a3958b

adamdebreceni (approved):

martinzink (approved):

martinzink:
https://github.com/apache/nifi-minifi-cpp/commit/ca8861edc26f2417b03076b9e961177644509d82

PR #1225, opened 2021-12-10
MINIFICPP-1688: When storing time durations we should use std::chrono…

martinzink:
… instead of simple integers

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/sftp/processors/ListSFTP.cpp
```diff
@@ -870,7 +871,7 @@ void ListSFTP::listByTrackingEntities(
 
   time_t now = time(nullptr);
   uint64_t min_timestamp_to_list = (!initial_listing_complete_ && entity_tracking_initial_listing_target_ == ENTITY_TRACKING_INITIAL_LISTING_TARGET_ALL_AVAILABLE)
```
lordgamez, line 873: Maybe we can write a separate Jira ticket to replace uint64_t timestamp representations like this to std::chrono in this file.

extensions/standard-processors/processors/GetFile.h
```diff
@@ -114,7 +114,7 @@ class GetFile : public core::Processor, public state::response::MetricsNodeSourc
   explicit GetFile(const std::string& name, const utils::Identifier& uuid = {})
       : Processor(name, uuid),
         metrics_(std::make_shared<GetFileMetrics>()),
-        last_listing_time_(0) {
+        last_listing_time_() {
```
lordgamez, line 117: I think this can be removed

extensions/standard-processors/tests/CPPLINT.cfg
```diff
@@ -0,0 +1 @@
+filter=-build/namespaces
```
lordgamez, line 1: As per my previous comment maybe we could go with `build/namespaces_literals` here as well.

libminifi/include/sitetosite/RawSocketProtocol.h
```diff
@@ -75,9 +75,9 @@ class RawSiteToSiteClient : public sitetosite::SiteToSiteClient {
     peer_ = std::move(peer);
     _batchSize = 0;
     _batchCount = 0;
-    _batchDuration = 0;
-    _batchSendNanos = 5000000000;  // 5 seconds
-    _timeOut = 30000;  // 30 seconds
+    _batchDuration = std::chrono::seconds(0);
+    _batchSendNanos = std::chrono::seconds(5);
+    _timeOut = std::chrono::seconds(30);  // 30 seconds
```
lordgamez, line 80: The comment is redundant now, it can be removed.

lordgamez (commented):
Thanks for this change, the code is much more clear now with the explicit timestamps. LGTM only a few minor comments.

extensions/libarchive/BinFiles.cpp
```diff
@@ -118,13 +118,9 @@ void BinFiles::onSchedule(core::ProcessContext *context, core::ProcessSessionFac
...
-  std::string maxBinAgeStr;
-  if (context->getProperty(MaxBinAge.getName(), maxBinAgeStr)) {
-    core::TimeUnit unit;
-    if (core::Property::StringToTime(maxBinAgeStr, val64, unit) && core::Property::ConvertTimeUnitToMS(val64, unit, val64)) {
-      this->binManager_.setBinAge(val64);
-      logger_->log_debug("BinFiles: MaxBinAge [%" PRIu64 "]", val64);
-    }
+  if (auto max_bin_age = context->getProperty<core::TimePeriodValue>(MaxBinAge)) {
+    this->binManager_.setBinAge(max_bin_age->getMilliseconds());
+    logger_->log_debug("BinFiles: MaxBinAge [%" PRIu64 "]", val64);
```
adamdebreceni, line 123: this log is out of sync, we no longer seem use the `val64` to store the max bin age

libminifi/include/core/Repository.h
```diff
@@ -56,17 +56,19 @@ namespace core {
 
 #define REPOSITORY_DIRECTORY "./repo"
 #define MAX_REPOSITORY_STORAGE_SIZE (10*1024*1024)  // 10M
-#define MAX_REPOSITORY_ENTRY_LIFE_TIME (600000)  // 10 minute
-#define REPOSITORY_PURGE_PERIOD (2500)  // 2500 msec
+constexpr std::chrono::milliseconds MAX_REPOSITORY_ENTRY_LIFE_TIME = std::chrono::minutes(10);
+constexpr std::chrono::milliseconds REPOSITORY_PURGE_PERIOD = std::chrono::milliseconds(2500);
```
szaszm, line 60:
```suggestion
using namespace std::literals;
constexpr auto MAX_REPOSITORY_ENTRY_LIFE_TIME = 10m;
constexpr auto REPOSITORY_PURGE_PERIOD = 2500ms;
```

extensions/libarchive/BinFiles.h
```diff
@@ -154,7 +154,7 @@ class BinManager {
   void setMinEntries(uint32_t entries) {
     minEntries_ = {entries};
   }
-  void setBinAge(uint64_t age) {
+  void setBinAge(std::chrono::milliseconds age) {
     binAge_ = {age};
```
adamdebreceni, line 158: this kind of assignment was used to protect against accidental narrowing conversions, since we use `std::chrono::milliseconds` now I think they are unnecessary

extensions/sftp/processors/ListSFTP.cpp
```diff
@@ -965,25 +966,18 @@ void ListSFTP::onTrigger(const std::shared_ptr<core::ProcessContext> &context, c
...
-      /* The default is 3 hours */
-      entity_tracking_time_window = 3 * 3600 * 1000;
-      logger_->log_error("Entity Tracking Time Window attribute is invalid");
-    }
+  if (auto entity_tracking_window = context->getProperty<core::TimePeriodValue>(EntityTrackingTimeWindow)) {
+    entity_tracking_time_window = entity_tracking_window->getMilliseconds();
   } else {
-    /* The default is 3 hours */
-    entity_tracking_time_window = 3 * 3600 * 1000;
+    logger_->log_error("Entity Tracking Time Window attribute is invalid");
```
szaszm, line 980: I think `getProperty` returns `false` when the property is not required and missing. Invalid values cause exceptions. This log message seems to be invalid, because this property is not required, so it's not an error to have it be empty and thus `getProperty` returning `false` is valid.

extensions/sftp/processors/ListSFTP.cpp
```diff
@@ -359,21 +360,21 @@ bool ListSFTP::filterFile(const std::string& parent_path, const std::string& fil
 
   /* Age */
   time_t now = time(nullptr);
-  uint64_t file_age = (now - attrs.mtime) * 1000;
+  std::chrono::milliseconds file_age = std::chrono::seconds(now - attrs.mtime);
```
szaszm, line 363:
```c++
  std::chrono::milliseconds file_age = std::chrono::file_clock::now() - std::chrono::seconds(attrs.mtime);
```

extensions/sftp/processors/ListSFTP.cpp
```diff
@@ -261,20 +263,19 @@ void ListSFTP::onSchedule(const std::shared_ptr<core::ProcessContext> &context,
...
-  if (context->getProperty(MaximumFileAge.getName(), value)) {
-    core::TimeUnit unit;
-    if (!core::Property::StringToTime(value, maximum_file_age_, unit) || !core::Property::ConvertTimeUnitToMS(maximum_file_age_, unit, maximum_file_age_)) {
-      logger_->log_error("Maximum File Age attribute is invalid");
-    }
+
+  if (auto maximum_file_age = context->getProperty<core::TimePeriodValue>(MaximumFileAge)) {
+    maximum_file_age_ = maximum_file_age->getMilliseconds();
+  } else {
+    logger_->log_error("Maximum File Age attribute is missing or invalid");
```
szaszm, line 276: An exception could happen that couldn't happen before: `getProperty` throws on invalid value, and the conversion got pushed into `getProperty`. An exception thrown from `onSchedule` prevents the processor from starting, while the old behavior was to log an error and proceed with whatever default was set in the member. Please keep the old behavior if the property is invalid.

extensions/openwsman/processors/SourceInitiatedSubscriptionListener.cpp
```diff
@@ -248,9 +248,9 @@ bool SourceInitiatedSubscriptionListener::loadState() {
   return true;
 }
 
-std::string SourceInitiatedSubscriptionListener::Handler::millisecondsToXsdDuration(int64_t milliseconds) {
+std::string SourceInitiatedSubscriptionListener::Handler::millisecondsToXsdDuration(std::chrono::milliseconds milliseconds) {
   char buf[1024];
-  snprintf(buf, sizeof(buf), "PT%" PRId64 ".%03" PRId64 "S", milliseconds / 1000, milliseconds % 1000);
+  snprintf(buf, sizeof(buf), "PT%" PRId64 ".%03" PRId64 "S", milliseconds.count() / 1000, milliseconds.count() % 1000);
```
adamdebreceni, line 253: the requirement for `std::chrono::milliseconds::rep` is that it is a signed integer type of at least 45 bits, on most hardware this probably means `int64_t` but it might not be, a cast would help ensure that `PRId64` is the correct format specifier

extensions/rocksdb-repos/ProvenanceRepository.h
```diff
@@ -83,12 +83,11 @@ class ProvenanceRepository : public core::Repository, public std::enable_shared_
     }
     logger_->log_debug("MiNiFi Provenance Max Partition Bytes %d", max_partition_bytes_);
     if (config->get(Configure::nifi_provenance_repository_max_storage_time, value)) {
-      core::TimeUnit unit;
-      if (core::Property::StringToTime(value, max_partition_millis_, unit)) {
-        core::Property::ConvertTimeUnitToMS(max_partition_millis_, unit, max_partition_millis_);
-      }
+      auto max_partition = utils::timeutils::StringToDuration<std::chrono::milliseconds>(value);
+      if (max_partition.has_value())
```
adamdebreceni, line 87: we could use the declaration form of the if statement here as well

extensions/sftp/processors/ListSFTP.cpp
```diff
@@ -359,21 +360,21 @@ bool ListSFTP::filterFile(const std::string& parent_path, const std::string& fil
...
   time_t now = time(nullptr);
-  uint64_t file_age = (now - attrs.mtime) * 1000;
+  std::chrono::milliseconds file_age = std::chrono::seconds(now - attrs.mtime);
   if (file_age < minimum_file_age_) {
     logger_->log_debug("Ignoring \"%s/%s\" because it is younger than the Minimum File Age: %ld ms < %lu ms",
         parent_path.c_str(),
         filename.c_str(),
-        file_age,
-        minimum_file_age_);
+        file_age.count(),
```
adamdebreceni, line 368: `%ld` might not be the right one here (it seems it never was)

libminifi/src/provenance/Provenance.cpp
```diff
@@ -45,13 +45,13 @@ const char *ProvenanceEventRecord::ProvenanceEventTypeStr[REPLAY + 1] = { "CREAT
 
 ProvenanceEventRecord::ProvenanceEventRecord(ProvenanceEventRecord::ProvenanceEventType event, std::string componentId, std::string componentType)
     : core::SerializableComponent(core::getClassName<ProvenanceEventRecord>()),
-      _entryDate(0),
-      _lineageStartDate(0),
+      _entryDate(std::chrono::time_point<std::chrono::system_clock>()),
```
adamdebreceni, line 48: is there any advantage using this over `std::chrono::system_clock::time_point`? the latter is not much shorter so I don't insist on changing it, just asking

szaszm: We could also just delete the line and let the default constructor be called automatically.

libminifi/src/RemoteProcessorGroupPort.cpp
```diff
@@ -167,17 +167,12 @@ void RemoteProcessorGroupPort::onSchedule(const std::shared_ptr<core::ProcessCon
...
-      logger_->log_debug("%s attribute is invalid, so default value of %s will be used", idleTimeout.getName(),
-                         idleTimeout.getValue());
-      if (!core::Property::getTimeMSFromString(idleTimeout.getValue(), idleTimeoutVal)) {
-        assert(false);  // Couldn't parse our default value
-      }
+    if (auto idle_timeout = context->getProperty<core::TimePeriodValue>(idleTimeout)) {
+      idle_timeout_ = idle_timeout->getMilliseconds();
+    } else {
+      logger_->log_debug("%s attribute is invalid, so default value of %s will be used", idleTimeout.getName(), idleTimeout.getDefaultValue());
+      idle_timeout_ = static_cast<core::TimePeriodValue>(idleTimeout.getDefaultValue().to_string()).getMilliseconds();
```
adamdebreceni, line 174: this `static_cast` feels "off", although at a glance I couldn't find anything in the coding guidelines that would advise against it, might even advise for it: 
> Use static_cast as the equivalent of a C-style cast that does value conversion

extensions/sftp/processors/ListSFTP.cpp
martinzink: this wouldnt work because the rhs would be a `time_point<file_clock>` and not a `duration`, and afaik the attrs.mtime is measured from unix epoch and since the epoch of the file_clock is not necessarily the same as the unix epoch, the `time_point<file_clock>::time_since_epoch` could return something completely different from what we want.

we could create a `time_point<file_clock>` from the attrs.mtime but for that we would need a platform specific helper function (because the file_clock interface is quite different on msvc, libc++ and on stdlibc++) it could be done with something similar to this (only in reverse):
https://github.com/rustammendel/nifi-minifi-cpp/pull/1/files#diff-796b46848475896d2982dcb681e0f547bac6aec3c98da80e8c03139562092604R89-R97

or we could stick to using system_clock (which has the unix epoch)
`auto file_age = std::chrono::system_clock::now() - std::chrono::system_clock::from_time_t(attrs.mtime);`

szaszm: Good point. I would go for `system_clock` in that case, for simplicity. I think that these timestamps are just system timestamps anyway.

extensions/sftp/processors/ListSFTP.cpp
martinzink: This one calls the
```
template<typename T = std::string>
std::enable_if_t<std::is_default_constructible<T>::value, std::optional<T>>
getProperty(const Property& property) const {
  T value;
  try {
    if (!getProperty(property.getName(), value)) return std::nullopt;
  } catch (const utils::internal::ValueException&) {
    return std::nullopt;
  }
  return value;
}
```
which doesnt throw just returns a nullopt if the value is invalid

szaszm: Sorry, I assumed that all of them throw on validation errors.

libminifi/include/utils/TimeUtil.h
```diff
@@ -197,6 +193,100 @@ inline bool getDateTimeStr(int64_t unix_timestamp, std::string& date_time_str) {
...
+}
+
+
+inline std::chrono::milliseconds getMillisecondsSinceUnixEpoch(const std::chrono::time_point<std::chrono::system_clock> time_point) {
+  return std::chrono::duration_cast<std::chrono::milliseconds>(time_point.time_since_epoch());
+}
+
+inline std::chrono::time_point<std::chrono::system_clock> fromMillisecondsSinceUnixEpoch(const std::chrono::milliseconds milliseconds_since_unix_epoch) {
+  return std::chrono::time_point<std::chrono::system_clock>() + milliseconds_since_unix_epoch;
+}
```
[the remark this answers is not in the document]
martinzink: I wanted these to explicitly state that we are converting to/from unix_epoch milliseconds timestamp to avoid ambiguity, because before c++20 system_clock wasn't neccessarly tied to unix_epoch (now it is and we are also verifying this in TimeUtilTests `TEST_CASE("Test system_clock epoch", "[systemclockepoch]")` but with this function this knowledge is not required by the users and becomes an implementation detail)

But I am not hard set on this either so I can change if thats more readable.

libminifi/src/RemoteProcessorGroupPort.cpp
martinzink: you are right, I think it was due to copy paste but yeah entirely unneccesary removed it in https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/78f8658fde802c73a5965dd734a544acbc37dca3 https://github.com/martinzink/nifi-minifi-cpp/commit/78f8658fde802c73a5965dd734a544acbc37dca3#diff-11bc2ef16726b961c333f05cf658ab5a88b7331722598d5d0921029e3cc47f89R174

libminifi/src/provenance/Provenance.cpp
martinzink: Removed these and other unneccesary initializations from the ctors in https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/78f8658fde802c73a5965dd734a544acbc37dca3

I prefer the `std::chrono::system_clock::time_point` aswell, its more readable thanks for idea @adamdebreceni, changed all references of time_point<x_clock> to x_clock::time_point in https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/78f8658fde802c73a5965dd734a544acbc37dca3

extensions/sftp/processors/ListSFTP.cpp
martinzink: changed all debug logs to use the `PRId64` and `static_cast<int64_t>` in https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/78f8658fde802c73a5965dd734a544acbc37dca3 and https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/1fcbcc05ce27c9f7386c8f5b3a4a5267593c8fa1

extensions/openwsman/processors/SourceInitiatedSubscriptionListener.cpp
martinzink: changed all printf like usages to use the `PRId64` and `static_cast<int64_t>` in https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/78f8658fde802c73a5965dd734a544acbc37dca3 and https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/1fcbcc05ce27c9f7386c8f5b3a4a5267593c8fa1

extensions/rocksdb-repos/ProvenanceRepository.h
martinzink: good idea, changed in https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/1fcbcc05ce27c9f7386c8f5b3a4a5267593c8fa1

extensions/sftp/processors/ListSFTP.cpp
martinzink: changed to `auto file_age = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now() - std::chrono::system_clock::from_time_t(attrs.mtime));` (milliseconds because it will be printed out later) in https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/78f8658fde802c73a5965dd734a544acbc37dca3

extensions/sftp/processors/ListSFTP.cpp
martinzink: good catch, changed back to original behaviour so missing property doesnt cause error logs in https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/78f8658fde802c73a5965dd734a544acbc37dca3

libminifi/include/core/Repository.h
martinzink: good idea, changed them to auto in https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/78f8658fde802c73a5965dd734a544acbc37dca3, I've left out the using namespace std::literals; because this is a header

extensions/rocksdb-repos/ProvenanceRepository.h
```diff
@@ -38,8 +38,8 @@ namespace provenance {
 
 #define PROVENANCE_DIRECTORY "./provenance_repository"
 #define MAX_PROVENANCE_STORAGE_SIZE (10*1024*1024)  // 10M
-#define MAX_PROVENANCE_ENTRY_LIFE_TIME (60000)  // 1 minute
-#define PROVENANCE_PURGE_PERIOD (2500)  // 2500 msec
+constexpr std::chrono::milliseconds MAX_PROVENANCE_ENTRY_LIFE_TIME = std::chrono::minutes(1);
+constexpr std::chrono::milliseconds PROVENANCE_PURGE_PERIOD = std::chrono::milliseconds(2500);
```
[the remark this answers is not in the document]
martinzink: good idea, changed them to auto in https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/78f8658fde802c73a5965dd734a544acbc37dca3, I've left out the using namespace std::literals; because this is a header

extensions/rocksdb-repos/FlowFileRepository.h
```diff
@@ -51,9 +51,9 @@ namespace repository {
 #define FLOWFILE_CHECKPOINT_DIRECTORY "./flowfile_checkpoint"
 #endif
 #define MAX_FLOWFILE_REPOSITORY_STORAGE_SIZE (10*1024*1024)  // 10M
-#define MAX_FLOWFILE_REPOSITORY_ENTRY_LIFE_TIME (600000)  // 10 minute
-#define FLOWFILE_REPOSITORY_PURGE_PERIOD (2000)  // 2000 msec
-#define FLOWFILE_REPOSITORY_RETRY_INTERVAL_INCREMENTS (500)  // msec
+constexpr std::chrono::milliseconds MAX_FLOWFILE_REPOSITORY_ENTRY_LIFE_TIME = std::chrono::minutes(10);
+constexpr std::chrono::milliseconds FLOWFILE_REPOSITORY_PURGE_PERIOD = std::chrono::seconds(2);
+constexpr std::chrono::milliseconds FLOWFILE_REPOSITORY_RETRY_INTERVAL_INCREMENTS = std::chrono::milliseconds(500);
```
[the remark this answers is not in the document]
martinzink: good idea, changed them to auto in https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/78f8658fde802c73a5965dd734a544acbc37dca3, I've left out the using namespace std::literals; because this is a header

libminifi/include/utils/ValueParser.h
```diff
@@ -27,6 +27,7 @@
 #include <type_traits>
 #include <limits>
 #include <algorithm>
+#include <chrono>
```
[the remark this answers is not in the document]
martinzink: removed it in https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/78f8658fde802c73a5965dd734a544acbc37dca3

extensions/mqtt/processors/AbstractMQTTProcessor.h
```diff
@@ -130,8 +128,8 @@ class AbstractMQTTProcessor : public core::Processor {
   MQTTClient_deliveryToken delivered_token_;
   std::string uri_;
   std::string topic_;
-  int64_t keepAliveInterval_;
-  int64_t connectionTimeOut_;
+  std::chrono::seconds keepAliveInterval_{60};
+  std::chrono::seconds connectionTimeOut_{30};
```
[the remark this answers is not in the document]
martinzink: wow good catch, I've changed so now we store ms, and explicitly cast to seconds and ms where it is used in https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/78f8658fde802c73a5965dd734a544acbc37dca3

extensions/libarchive/BinFiles.h
martinzink: changed it in https://github.com/martinzink/nifi-minifi-cpp/commit/78f8658fde802c73a5965dd734a544acbc37dca3#diff-495a1240beddf26f5b8b605419131e85beb584bc830c457c7db3cdb9cbcc92c1R158

extensions/libarchive/BinFiles.cpp
martinzink: fixed in https://github.com/martinzink/nifi-minifi-cpp/commit/78f8658fde802c73a5965dd734a544acbc37dca3#diff-8437e347b195a4944072645eb29b284c9575430a08c714d7cbb9115a86933315R123

libminifi/include/sitetosite/RawSocketProtocol.h
martinzink: you are right, removed it in https://github.com/martinzink/nifi-minifi-cpp/commit/78f8658fde802c73a5965dd734a544acbc37dca3#diff-c1de4968ef4bf87b88a7ac2062a5abd667a9311affd42e8752a2531172e3357eR80

extensions/standard-processors/tests/CPPLINT.cfg
martinzink: with the filter `build/namespaces_literals ` in the root CPPLINT.cfg this file became obsolete deleted it in https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/78f8658fde802c73a5965dd734a544acbc37dca3

extensions/librdkafka/ConsumeKafka.cpp
```diff
@@ -25,6 +25,8 @@
 #include "utils/ProcessorConfigUtils.h"
 #include "utils/gsl.h"
 
+using namespace std::chrono_literals;  // NOLINT(build/namespaces)
```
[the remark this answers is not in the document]
martinzink: awesome thanks, changed all references in https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/78f8658fde802c73a5965dd734a544acbc37dca3

extensions/sftp/processors/ListSFTP.cpp
martinzink: Good idea, I created a Jira so we can collect these leftover todo-s https://issues.apache.org/jira/browse/MINIFICPP-1700

extensions/standard-processors/processors/GetFile.h
martinzink: removed it in https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/78f8658fde802c73a5965dd734a544acbc37dca3#diff-6cb7e1be41456997448531e124f3d2ae5e4bc94d24c528fc1a7d2f38eb9fb685R116

libminifi/include/utils/TimeUtil.h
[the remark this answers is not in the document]
szaszm: I think we can assume that the unix epoch is the system_clock epoch, but I didn't know that it wasn't universal before C++20.
For me, `duration_cast<milliseconds>(time_point.time_since_epoch())` reads as "the time since epoch, in milliseconds", which is about the same as the function name, that's why I didn't see the added value.
Similarly `time_point{} + milliseconds_since_epoch` reads as "epoch + milliseconds since epoch", so to me it's obvious what's happening, and it's better described by the expression than the function name IMO.

If you disagree and see additional value here, then I won't stand in the way. I leave my arguments here, but you should make the call.

extensions/openwsman/processors/SourceInitiatedSubscriptionListener.cpp
szaszm: `static_cast` works, but wrapping the number in a braced init style conversion (i.e. `int64_t{milliseconds.count()}`) is better IMO, because if if ever becomes wider than `int64_t`, then that will result in a compiler error. (Which is better than silently doing narrowing conversion IMO.)

martinzink: good idea, changed to the suggested format in https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/078ef637a1043f6e41d9d383833620cf6202f7e5

libminifi/include/utils/TimeUtil.h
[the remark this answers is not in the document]
martinzink: I see your point and agree, removed these helper functions in https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/078ef637a1043f6e41d9d383833620cf6202f7e5

lordgamez (approved):

libminifi/include/utils/TimeUtil.h
```diff
@@ -197,6 +193,91 @@ inline bool getDateTimeStr(int64_t unix_timestamp, std::string& date_time_str) {
...
+
+  return details::cast_to_matching_unit<TargetDuration,
+    std::chrono::nanoseconds,
+    std::chrono::microseconds,
+    std::chrono::milliseconds,
+    std::chrono::seconds,
+    std::chrono::minutes,
+    std::chrono::hours,
+    std::chrono::days>(unit, value);
+}
```
szaszm, line 279: Consider making this (and all of the helpers) `constexpr`. It would allow you to test it with `static_assert`. It might have to take a `std::string_view` instead of `std::string`, depending on the level of C++20 support our compilers implement.

libminifi/include/utils/TimeUtil.h
```diff
@@ -197,6 +193,91 @@ inline bool getDateTimeStr(int64_t unix_timestamp, std::string& date_time_str) {
...
+  ((result = cast_if_unit_matches<TargetDuration, T>(unit, value)) || ...);
+  return result;
+}
+
+inline bool get_unit_and_value(const std::string& input, std::string& unit, int64_t& value) {
+  std::stringstream input_stream(input);
+  input_stream >> value >> unit;
+  std::transform(unit.begin(), unit.end(), unit.begin(), ::tolower);
+  return !input_stream.fail();
+}
```
szaszm, line 260: I would prefer a stricter parsing logic than istream.
For example: find the space, before that it's the value, afterwards it's the unit. I think this would be equivalent with the old one.

libminifi/include/utils/TimeUtil.h
martinzink: Good idea, however I didnt find any good constexpr way to cast a string to int (std::stoi, std::from_chars are not constexpr).

Of course we could implement our own ([related stackoverflow question)](https://stackoverflow.com/questions/25195176/how-do-i-convert-a-c-string-to-a-int-at-compile-time) but I am not sure if it worth it.

szaszm: Hmm, I didn't think of this. Nevermind then, thanks for checking.

libminifi/src/c2/C2Agent.cpp
```diff
@@ -161,12 +161,9 @@ void C2Agent::configure(const std::shared_ptr<Configure> &configure, bool reconf
...
 
   if (configure->get("nifi.c2.agent.heartbeat.period", "c2.agent.heartbeat.period", heartbeat_period)) {
-    core::TimeUnit unit;
-
     try {
-      int64_t schedulingPeriod = 0;
-      if (core::Property::StringToTime(heartbeat_period, schedulingPeriod, unit) && core::Property::ConvertTimeUnitToMS(schedulingPeriod, unit, schedulingPeriod)) {
-        heart_beat_period_ = schedulingPeriod;
+      if (auto heartbeat_period_ms = utils::timeutils::StringToDuration<std::chrono::milliseconds>(heartbeat_period)) {
+        heart_beat_period_ = heartbeat_period_ms->count();
```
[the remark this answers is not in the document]
martinzink: done https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/9f0c3752c6a328e005bb22a566eceb406c7fe348

libminifi/src/utils/Id.cpp
```diff
@@ -42,6 +42,7 @@
 #endif
 
 #include "utils/StringUtils.h"
+#include "utils/TimeUtil.h"
```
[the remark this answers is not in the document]
martinzink: good catch removed and another one like this in https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/9f0c3752c6a328e005bb22a566eceb406c7fe348

extensions/mqtt/controllerservice/MQTTControllerService.h
```diff
@@ -293,8 +289,8 @@ class MQTTControllerService : public core::controller::ControllerService {
   MQTTClient client_;
   std::string uri_;
   std::string topic_;
-  int64_t keepAliveInterval_;
-  int64_t connectionTimeOut_;
+  std::chrono::milliseconds keepAliveInterval_{0};
+  std::chrono::milliseconds connectionTimeOut_{0};
```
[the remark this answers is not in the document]
martinzink: good idea, changed it in https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/9f0c3752c6a328e005bb22a566eceb406c7fe348

libminifi/test/unit/CpuUsageTest.cpp
```diff
@@ -23,39 +23,39 @@
 #include "utils/ProcessCpuUsageTracker.h"
 #include "../TestBase.h"
 
-void busySleep(int duration_ms, std::chrono::milliseconds& start_ms, std::chrono::milliseconds& end_ms, const std::chrono::system_clock::time_point& origin) {
-  start_ms = std::chrono::duration_cast<std::chrono::milliseconds> (std::chrono::system_clock::now() - origin);
-  end_ms = std::chrono::duration_cast<std::chrono::milliseconds> (std::chrono::system_clock::now() - origin);
-  while (end_ms-start_ms < std::chrono::milliseconds(duration_ms)) {
-    end_ms = std::chrono::duration_cast<std::chrono::milliseconds> (std::chrono::system_clock::now() - origin);
+using namespace std::literals::chrono_literals;
+using namespace std::chrono;
```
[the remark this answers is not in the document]
martinzink: fixed it in https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/9f0c3752c6a328e005bb22a566eceb406c7fe348

libminifi/include/utils/TimeUtil.h
martinzink: space is not required (e.g. 5s is valid), but I agree with you and sstream is quite slow compared to other methods so I reverted this to the original implementation in https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/9f0c3752c6a328e005bb22a566eceb406c7fe348

extensions/mqtt/controllerservice/MQTTControllerService.h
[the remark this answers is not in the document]
szaszm: I see the changes in other files, but this particular one is not changed in the commit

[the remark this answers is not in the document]
martinzink: you are right I do seem to have skipped this, probably because it is not used anywhere :confused: 
i've looked at MQTTControllerService and it has a bunch of properties and matching members that are not used anywhere as far as I could tell.
from BrokerURL, ClientID, CleanSession, UserName, Password, KeepLiveInterval, ConnectionTimeout, QOS Topic, SecurityProtocol only BrokerURL and ClientID is used in onEnable

Am i missing something ?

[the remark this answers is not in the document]
szaszm: it is indeed unused, so it's probably best to remove it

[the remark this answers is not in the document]
martinzink: I will create a separate ticket and investigate, because the MQTTControllerService::reconnect relies on some of these uninitialized members and the AbstractMQTTProcessor and MQTTControllerService have a lot of duplicate codes.
I wouldnt be suprised if there are some bugs here or there.

Anyways I fixed the timeout typos throughout the code in https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/63a26de4232196b77a7a366ad95f99a3a62fc598

fgerlits (commented):
LGTM, added some nitpicky suggestions.

libminifi/include/utils/TimeUtil.h
```diff
@@ -197,6 +193,106 @@ inline bool getDateTimeStr(int64_t unix_timestamp, std::string& date_time_str) {
...
+namespace details {
+
+template<class Duration>
+bool unit_matches(const std::string&) {
+  return false;
+}
+
+template<>
+inline bool unit_matches<std::chrono::nanoseconds>(const std::string& unit) {
+  return unit == "ns" || unit == "nano" || unit == "nanos" || unit == "nanoseconds";
```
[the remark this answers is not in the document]
martinzink: sure thing, https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/80ebd14a79cbffc9287b7a8c260abe3510db4d74#diff-3f1741daa1a9506c6303aeb6642d79146ec5ee6ee9469a2b04c4cf73a2ffd50dR165

libminifi/include/sitetosite/Peer.h
```diff
@@ -155,12 +158,11 @@ class SiteToSitePeer : public org::apache::nifi::minifi::io::BaseStream {
...
         port_(port),
-        timeout_(30000),
-        yield_expiration_(0),
+        timeout_(std::chrono::seconds(30)),
         logger_(core::logging::LoggerFactory<SiteToSitePeer>::getLogger()) {
     url_ = "nifi://" + host_ + ":" + std::to_string(port_);
-    yield_expiration_ = 0;
-    timeout_ = 30000;  // 30 seconds
+    yield_expiration_ = std::chrono::system_clock::time_point();
+    timeout_ = std::chrono::seconds(30);  // 30 seconds
```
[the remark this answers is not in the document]
martinzink: fixed https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/80ebd14a79cbffc9287b7a8c260abe3510db4d74#diff-3f1741daa1a9506c6303aeb6642d79146ec5ee6ee9469a2b04c4cf73a2ffd50dR165

libminifi/include/core/ProcessorConfig.h
```diff
@@ -32,8 +32,8 @@ namespace core {
 
 #define DEFAULT_SCHEDULING_STRATEGY "TIMER_DRIVEN"
 #define DEFAULT_SCHEDULING_PERIOD_STR "1 sec"
-#define DEFAULT_SCHEDULING_PERIOD_MILLIS 1000
-#define DEFAULT_RUN_DURATION 0
+constexpr std::chrono::milliseconds DEFAULT_SCHEDULING_PERIOD_MILLIS{1000};
+constexpr std::chrono::nanoseconds DEFAULT_RUN_DURATION{0};
 #define DEFAULT_MAX_CONCURRENT_TASKS 1
 #define DEFAULT_YIELD_PERIOD_SECONDS 1
```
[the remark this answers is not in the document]
martinzink: :+1:  done https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/80ebd14a79cbffc9287b7a8c260abe3510db4d74#diff-f2a42efc99a3aa3a32fef6dc7b543b9d6181889937b4c368070753700a7b5cc5R38

extensions/windows-event-log/CollectorInitiatedSubscription.cpp
```diff
@@ -39,6 +39,8 @@
 #pragma comment(lib, "wevtapi.lib")
 #pragma comment(lib, "Wecapi.lib")
 
+using namespace std::chrono_literals;  // NOLINT(build/namespaces)
```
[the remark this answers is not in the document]
martinzink: good idea, changed it in https://github.com/apache/nifi-minifi-cpp/pull/1225/commits/80ebd14a79cbffc9287b7a8c260abe3510db4d74#diff-d3b46e6cb41aa26d0fa4816a424dd9c3de38244da6918403333e0192906b5b9aR35

szaszm (approved):

fgerlits (approved):

adamdebreceni (approved):

PR #1226, opened 2021-12-13
MINIFICPP-1673: ProcessSession::write/ProcessSession::append should o…

martinzink:
…nly operate on flowfiles marked as new/modified

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/src/core/ProcessSession.cpp
+++ b/libminifi/src/core/ProcessSession.cpp
@@ -212,6 +212,10 @@ void ProcessSession::transfer(const std::shared_ptr<core::FlowFile> &flow, Relat
 }
 
 void ProcessSession::write(const std::shared_ptr<core::FlowFile> &flow, OutputStreamCallback *callback) {
+  gsl_Expects(_updatedFlowFiles.contains(flow->getUUID())
+              || _addedFlowFiles.contains(flow->getUUID())
+              || std::any_of(_clonedFlowFiles.begin(), _clonedFlowFiles.end(), [&flow](const auto& cloned_flow_file) { return flow == cloned_flow_file; }));
+
   std::shared_ptr<ResourceClaim> claim = content_session_->create();
 
   try {
@@ -255,6 +259,10 @@ void ProcessSession::writeBuffer(const std::shared_ptr<core::FlowFile>& flow_fil
 }
 
 void ProcessSession::append(const std::shared_ptr<core::FlowFile> &flow, OutputStreamCallback *callback) {
+  gsl_Expects(_updatedFlowFiles.contains(flow->getUUID())
+              || _addedFlowFiles.contains(flow->getUUID())
+              || std::any_of(_clonedFlowFiles.begin(), _clonedFlowFiles.end(), [&flow](const auto& cloned_flow_file) { return flow == cloned_flow_file; }));
+
   std::shared_ptr<ResourceClaim> claim = flow->getResourceClaim();
   if (!claim) {
     // No existed claim for append, we need to create new claim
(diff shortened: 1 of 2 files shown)


libminifi/src/core/ProcessSession.cpp
```diff
@@ -212,6 +212,10 @@ void ProcessSession::transfer(const std::shared_ptr<core::FlowFile> &flow, Relat
 }
 
 void ProcessSession::write(const std::shared_ptr<core::FlowFile> &flow, OutputStreamCallback *callback) {
+  gsl_Expects(_updatedFlowFiles.contains(flow->getUUID())
+              || _addedFlowFiles.contains(flow->getUUID())
+              || std::any_of(_clonedFlowFiles.begin(), _clonedFlowFiles.end(), [&flow](const auto& cloned_flow_file) { return flow == cloned_flow_file; }));
+
```
szaszm, line 218: This check looks like it's on the more expensive side. Consider changing it to [`gsl_ExpectsAudit`](https://github.com/gsl-lite/gsl-lite#contract-checking-configuration-macros) to remove the check from non-CI builds that don't define `gsl_CONFIG_CONTRACT_CHECKING_AUDIT`. (same in append)

martinzink: good idea, changed it in https://github.com/apache/nifi-minifi-cpp/pull/1226/commits/300532be1a9347d2926c139185bbf46edb619102

szaszm (approved):

libminifi/src/core/ProcessSession.cpp
szaszm: Consider extracting the lambda to a variable to avoid the error with clang. It will start working once they implement the C++20 feature "Lambdas in unevaluated contexts".
This is the disabled implementation of the macro:
```
# define gsl_ELIDE_( x )  static_assert( ::std::is_constructible<bool, decltype( x )>::value, "argument of contract check must be convertible to bool" )
```

libminifi/src/core/ProcessSession.cpp
```diff
@@ -212,6 +212,10 @@ void ProcessSession::transfer(const std::shared_ptr<core::FlowFile> &flow, Relat
 }
 
 void ProcessSession::write(const std::shared_ptr<core::FlowFile> &flow, OutputStreamCallback *callback) {
+  gsl_ExpectsAudit(_updatedFlowFiles.contains(flow->getUUID())
```
adamdebreceni, line 215: as we provide access to the session to custom code through the script extension, this feels like an invariant we should always check, turning the `std::map`s (and the `_clonedFlowFiles` vector) into `std::unordered_map`s would make the check an `O(1)` operation

szaszm: I prefer the narrow contract, but if you decide to have a wide contract, then please use if + throw or something similar to implement it, not gsl_Expects (which is meant to declare preconditions).

adamdebreceni: we wrap the session in a `PyProcessSession` which could handle the check if it had access to the internals of the `ProcessSession` (not probable) or had its own flowfile registry (would be duplicating the session's internal registry)

szaszm: I would not be against a public member function that can check if a flow file can be written to. I don't know a good name though. Maybe `bool canBeWrittenTo(const core::FlowFile&) const noexcept;`? If we did that, `PyProcessSession` could check that and `write`/`append` could have that as their precondition.

adam-markovics (approved):

lordgamez (approved):

PR #1227, opened 2021-12-14
MINIFICPP-1222 Remove engine queue for ExecutePythonProcessor

lordgamez:
Originally the Jira ticket was about reowrking the engine queue to only allow the `maximum concurrent tasks` number of script engines to run at the same time. The problem with that was that even if the script engines were executed concurrently they use the same python interpreter with the global interpreter lock in place so no concurrency can take place.

The only possibility for concurrency to use sub-interpreters which would be also problematic. First of all pybind11 does not support an interface for sub-interpreters, it is only available through the CPython API, but the documentation also warns about the caveats of the sub-interpreters here: https://pybind11.readthedocs.io/en/stable/advanced/embedding.html
Secondly using the CPython API for sub-interpreters would require us to always know which thread we are running on and create separate sub-interpreters and thread states for each thread (in this case it would possibly not even allow us to run the python scripts from the main thread anymore). Managing the thread states and the sub-interpreters would be cumbersome in our architecture.

Because all of the above problems I removed the python script queue from the ExecutePythonProcessor which will run the python scripts on a single thread.

--------------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/script/python/ExecutePythonProcessor.cpp
+++ b/extensions/script/python/ExecutePythonProcessor.cpp
@@ -79,8 +79,6 @@ void ExecutePythonProcessor::initialize() {
     return;
   }
 
-  python_logger_ = core::logging::LoggerFactory<ExecutePythonProcessor>::getAliasedLogger(getName());
-
   getProperty(ModuleDirectory.getName(), module_directory_);
   appendPathForImportModules();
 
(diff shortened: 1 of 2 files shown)


szaszm (approved):

martinzink:
Looks good, but [ExecuteScript](https://github.com/apache/nifi-minifi-cpp/blob/main/extensions/script/ExecuteScript.cpp#L93) also uses engine queues (for both python and lua) so this problem should be there for ExecuteScript while in python mode.

On the other hand lua doesnt have a global interpreter lock and ExecuteScript doesnt really allow stateful scripts, so the engine queue should theoretically work there (everything is contained in the sol::state member of LuaScriptEngine.h), but then I guess the original problem detailed in https://issues.apache.org/jira/browse/MINIFICPP-1222 is still present in ExecuteScript Lua mode.

lordgamez:
> Looks good, but [ExecuteScript](https://github.com/apache/nifi-minifi-cpp/blob/main/extensions/script/ExecuteScript.cpp#L93) also uses engine queues (for both python and lua) so this problem should be there for ExecuteScript while in python mode.
> 
> On the other hand lua doesnt have a global interpreter lock and ExecuteScript doesnt really allow stateful scripts, so the engine queue should theoretically work there (everything is contained in the sol::state member of LuaScriptEngine.h), but then I guess the original problem detailed in https://issues.apache.org/jira/browse/MINIFICPP-1222 is still present in ExecuteScript Lua mode.

That's a good point, I think that should also be fixed with a solution that allows us to have concurrency with Lua executions with the set limit of maximum concurrent tasks, but handling python executions differently by only having a single script engine for that. I created a Jira ticket for that rework: https://issues.apache.org/jira/browse/MINIFICPP-1706

martinzink (approved):

PR #1228, opened 2021-12-21
MINIFICPP-1703 ExecuteScript error handling fix, more docs and tests

martinzink:
- ExecuteScript, ExecutePythonScript now lets exceptions through
- LuaScriptEngine now throws on errors
- Enabled TestExecuteScriptProcessorWithLuaScript in CI
- Added more examples and tests

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/script/python/PythonScriptEngine.h
```diff
@@ -57,7 +57,7 @@ struct Interpreter {
 
 Interpreter *getInterpreter();
 
-class PythonScriptEngine : public script::ScriptEngine {
+class __attribute__((visibility("default"))) PythonScriptEngine : public script::ScriptEngine {
```
szaszm, line 60: This may not compile on msvc, try doing it with a platform-specific macro.

szaszm (commented):
This is awesome, thanks.

extensions/script/ExecuteScript.cpp
```diff
@@ -79,80 +79,72 @@ void ExecuteScript::onSchedule(core::ProcessContext *context, core::ProcessSessi
...
+      engine->eval(script_body_);
+    } else if (!script_file_.empty()) {
+      engine->evalFile(script_file_);
+    } else {
+      throw std::runtime_error("Neither Script Body nor Script File is available to execute");
+    }
+  }
+
+  if (script_engine_ == "python") {
 #ifdef PYTHON_SUPPORT
```
arpadboda, line 129: Wouldn't it be easier to check if engine is valid (not nullptr) here?
In case python support is disabled, I think the error above at engine init is already thrown.

extensions/script/ExecuteScript.cpp
```diff
@@ -79,80 +79,72 @@ void ExecuteScript::onSchedule(core::ProcessContext *context, core::ProcessSessi
...
+    logger_->log_debug("Using available %s script engine instance", script_engine_);
+  } else {
+    logger_->log_info("Creating new %s script instance", script_engine_);
+    logger_->log_info("Approximately %d %s script instances created for this processor",
+                      script_engine_q_.size_approx(),
+                      script_engine_);
 
-      if (script_engine_ == "python") {
+    if (script_engine_ == "python") {
 #ifdef PYTHON_SUPPORT
```
arpadboda, line 102: I know it's old code, but I feel like these ifdefs are unnecessary here. 
Python being enabled in the build could also be checked in OnSchedule. In case it's not, the processor can simply thrown an exception there, preventing ontrigger to be called ever.

arpadboda (commented):
Most looks good, added some improvement ideas.

extensions/script/ExecuteScript.cpp
martinzink: I agree 👍
This is quite old code, and I think it could really use some refactor
(we could replace these template functions and ifdefs with some good old polymorphism, add allowable values for the ScriptEngine Property etc..)

Originally I didnt wanna inflate this PR with these refactors, how about a separete PR aiming at refactoring and modernizing the ExecuteScript?

arpadboda: Makes sense to me :)

arpadboda (approved):

extensions/script/python/PythonScriptEngine.h
szaszm: It looks like we are not testing this extension on msvc. I'm fine with ignoring the error until we actually want to support this on windows.

szaszm (approved):
Not sure if you want to add more changes, but so far I'm +1.

lordgamez (approved):

PR #1230, opened 2022-01-03
MINIFICPP-1705 Upgrade and fix compilation of OpenCV

lordgamez:
Loading the OpenCV extension failed due to zlib library not linked to opencv-core library. Upgraded OpenCV to the latest version 4.5.5 and linked our bundled zlib to the opencv-core lib.

https://issues.apache.org/jira/browse/MINIFICPP-1705

### For all changes:
     in the commit message?

### For code changes:

### Note:


szaszm (approved):

cmake/BundledOpenCV.cmake
```diff
@@ -138,10 +134,15 @@ function(use_bundled_opencv SOURCE_DIR BINARY_DIR)
     set_target_properties(OPENCV::libpng PROPERTIES IMPORTED_LOCATION "${OPENCV_BYPRODUCT_DIR}/${CMAKE_INSTALL_LIBDIR}/${THIRDPARTY_DIR}${PREFIX}libpng${THIRDPARTY_SUFFIX}")
     add_dependencies(OPENCV::libpng opencv-external)
 
+    add_library(OPENCV::libopenjp2 STATIC IMPORTED)
```
adamdebreceni, line 137: are there some PRs incoming that verify this fix?

fgerlits (approved):

cmake/BundledOpenCV.cmake
lordgamez: OpenCV is compiled in the CI so the upgrade should be okay but I only tested the loading of the library manually. The proper way would be to have some test coverage for the OpenCV extension, but currently I'm not aware of anyone using it so it is not a priority. I created a ticket for this in Jira so we can add it in the future https://issues.apache.org/jira/browse/MINIFICPP-1709

adamdebreceni (approved):

cmake/BundledOpenCV.cmake
```diff
@@ -138,10 +134,15 @@ function(use_bundled_opencv SOURCE_DIR BINARY_DIR)
...
+    set_target_properties(OPENCV::libopenjp2 PROPERTIES IMPORTED_LOCATION "${OPENCV_BYPRODUCT_DIR}/${CMAKE_INSTALL_LIBDIR}/${THIRDPARTY_DIR}${PREFIX}libopenjp2${THIRDPARTY_SUFFIX}")
```
szaszm, line 138: I got this error:
```
make[2]: *** No rule to make target 'thirdparty/opencv-install/lib64/opencv4/3rdparty/liblibopenjp2.a', needed by 'bin/libminifi-opencv.so'.  Stop.
make[1]: *** [CMakeFiles/Makefile2:6131: extensions/opencv/CMakeFiles/minifi-opencv.dir/all] Error 2
make: *** [Makefile:166: all] Error 2
```

Removing the "lib" part after `${PREFIX}` didn't help, but it changed the missing target from "liblibopenjp2.a" to "libopenjp2.a". I don't know what's the problem.

This was a clean build, only reusing FetchContent dependencies, but this uses ExternalProject, so I think it shouldn't be an issue. Any ideas?

szaszm (changes requested):

cmake/BundledOpenCV.cmake
lordgamez: Could you check the openjp2 library on your system? Is it available under the install path and if it is what is its build path?

szaszm: it's not under 'thirdparty/opencv-install/lib64/opencv4/3rdparty/liblibopenjp2.a' and it's not installed on my system elsewhere

lordgamez: That looks to be right, maybe it is a ordering issue with the compilation of the targets when compiling them on multiple threads. Could you check if the issue persists if the `add_dependencies(OPENCV::libopencv-imgcodecs opencv-external)` line is expanded with the linked targets?

szaszm: It's not a build order issue. Running on a single thread or repeated running on a lot of threads didn't solve the issue. Added all of the targets from `target_link_libraries` to `add_dependencies` for `imgcodecs`, but it didn't help either. For some reason, the library is not built at all, at least in my environment.

szaszm: I think I just found the issue: we need to pass `-DBUILD_OPENJPEG=ON` to the opencv cmake build for it to build the library. I don't know why it worked anywhere else.

lordgamez: Sorry I misread your comment, I thought you wrote that it's under the mentioned directory. It's quite strange, what OS are you building it on? Maybe I can try to reproduce it in a docker image. Also can you check if you have the same issue when building the library separately from https://github.com/opencv/opencv.git on the 4.5.5 branch?

lordgamez: > I think I just found the issue: we need to pass `-DBUILD_OPENJPEG=ON` to the opencv cmake build for it to build the library. I don't know why it worked anywhere else.

Great, thanks! I'll add this to the PR

szaszm: Do we actually need this new dependency though? Why did you add openjp2? I originally assumed it's just a new upstream dependency, but it seems to be just another alternative jpeg codec library.

lordgamez: It just didn't compile without it, failing with linkage issues.

lordgamez: Removed libopenjp2 dependency in e0c9e963a92e77ca72a62930f151e094851141d7

szaszm: Summary (for the record, we continued in PM):
OpenCV changed their JPEG2000 codec from Jasper to OpenJPEG. We didn't have Jasper enabled previously, but to disable the new codec, we need to use a new option: `-DWITH_OPENJPEG=OFF`. This one defaults to ON, while the corresponding `-DBUILD_OPENJPEG` option defaulted to OFF, at least on my machine. Either including the openjp2 target (this is the openjpeg library name) or disabling `WITH_OPENJPEG` makes opencv_imgcodecs correctly link to our opencv extension.
We decided not to add JPEG2000 support for now and disable the new codec, too.

szaszm (approved):

PR #1231, opened 2022-01-04
MINIFICPP-1707 - Speed up linter

adam-markovics:
Speedup is 19 to 5 seconds on my computer with using all cores:
`make -j12 linter`

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/cmake/Extensions.cmake
+++ b/cmake/Extensions.cmake
@@ -115,7 +115,7 @@ macro(register_extension_linter target-name)
     get_property(extensions GLOBAL PROPERTY EXTENSION-LINTERS)
     set_property(GLOBAL APPEND PROPERTY EXTENSION-LINTERS "${target-name}")
     add_custom_target(${target-name}
-    COMMAND python ${CMAKE_SOURCE_DIR}/thirdparty/google-styleguide/run_linter.py -q -i ${CMAKE_CURRENT_LIST_DIR}/)
+    COMMAND python3 ${CMAKE_SOURCE_DIR}/thirdparty/google-styleguide/run_linter.py -q -i ${CMAKE_CURRENT_LIST_DIR}/)
   endif()
 endmacro()
 
--- a/thirdparty/google-styleguide/run_linter.py
+++ b/thirdparty/google-styleguide/run_linter.py
@@ -35,13 +35,11 @@ def cpplint_main_wrapper(file_list):
     except SystemExit as err:
         return err.code
 
-    return 1
-
 
 if __name__ == '__main__':
     # break up list_of_files to ~equal chunks
     chunk_num = multiprocessing.cpu_count()
-    chunk_size = math.ceil((len(list_of_files) / chunk_num))
+    chunk_size = math.ceil(len(list_of_files) / chunk_num)
     chunks = []
     chunk_begin = 0
     chunk_end = chunk_size
--- a/thirdparty/google-styleguide/run_linter.py
+++ b/thirdparty/google-styleguide/run_linter.py
@@ -2,7 +2,6 @@ import argparse
 import multiprocessing
 import os
 import cpplint
-from multiprocessing import Pool
 import sys
 import math
 
@@ -51,7 +50,7 @@ if __name__ == '__main__':
         chunk_end += chunk_size
         if chunk_end > len(list_of_files):
             chunk_end = len(list_of_files)
-    pool = Pool(processes=chunk_num)
+    pool = multiprocessing.Pool(processes=chunk_num)
     # execute in parallel
     map_res = pool.map(cpplint_main_wrapper, chunks)
     pool.close()
--- a/thirdparty/google-styleguide/run_linter.py
+++ b/thirdparty/google-styleguide/run_linter.py
@@ -35,21 +35,26 @@ def cpplint_main_wrapper(file_list):
         return err.code
 
 
-if __name__ == '__main__':
-    # break up list_of_files to ~equal chunks
-    chunk_num = multiprocessing.cpu_count()
-    chunk_size = math.ceil(len(list_of_files) / chunk_num)
-    chunks = []
+def create_chunks(chunk_size, items):
+    """breaks up items to equal chunks, with last being shorter if items are not evenly divisible by chunk_size"""
+    chunk_list = []
     chunk_begin = 0
     chunk_end = chunk_size
     for chunk_cnt in range(chunk_num):
-        chunks.append(list_of_files[chunk_begin:chunk_end])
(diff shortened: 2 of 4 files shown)


thirdparty/google-styleguide/run_linter.py
```diff
@@ -9,18 +13,50 @@
...
-cpplint.main(arg_list + list_of_files)
+
+def cpplint_main_wrapper(file_list):
+    try:
+        cpplint.main(arg_list + file_list)
+        return 0
+    except SystemExit as err:
+        return err.code
+
+    return 1
```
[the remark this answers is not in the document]
adam-markovics: What happens if there is an exception that is not SystemExit?

thirdparty/google-styleguide/run_linter.py
```diff
@@ -9,18 +13,50 @@
...
+
+
+if __name__ == '__main__':
+    # break up list_of_files to ~equal chunks
+    chunk_num = multiprocessing.cpu_count()
+    chunk_size = math.ceil((len(list_of_files) / chunk_num))
```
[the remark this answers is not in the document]
adam-markovics: True, fixing.

thirdparty/google-styleguide/run_linter.py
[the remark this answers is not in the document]
lordgamez: The exception will be uncaught and the process will terminate with an error code. It may be the same 1 error code.

cmake/Extensions.cmake
```diff
@@ -115,7 +115,7 @@ macro(register_extension_linter target-name)
     get_property(extensions GLOBAL PROPERTY EXTENSION-LINTERS)
     set_property(GLOBAL APPEND PROPERTY EXTENSION-LINTERS "${target-name}")
     add_custom_target(${target-name}
-    COMMAND python ${CMAKE_SOURCE_DIR}/thirdparty/google-styleguide/run_linter.py -q -i ${CMAKE_CURRENT_LIST_DIR}/)
+    COMMAND python3 ${CMAKE_SOURCE_DIR}/thirdparty/google-styleguide/run_linter.py -q -i ${CMAKE_CURRENT_LIST_DIR}/)
```
szaszm, line 118: Is there a mainstream OS where `python` doesn't refer to python 3.x?

thirdparty/google-styleguide/run_linter.py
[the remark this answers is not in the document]
adam-markovics: Okay, then I remove this line.

thirdparty/google-styleguide/run_linter.py
```diff
@@ -1,6 +1,10 @@
 import argparse
+import multiprocessing
 import os
 import cpplint
+from multiprocessing import Pool
```
szaszm, line 5: Only one of line 2 or line 5 should be enough.

cmake/Extensions.cmake
adam-markovics: Yes, for example Ubuntu 20.04.

thirdparty/google-styleguide/run_linter.py
adam-markovics: Done.

lordgamez (approved):

szaszm (approved):

martinzink (approved):

adamdebreceni (approved):

thirdparty/google-styleguide/run_linter.py
```diff
@@ -9,18 +12,48 @@
...
+    chunk_begin = 0
+    chunk_end = chunk_size
+    for chunk_cnt in range(chunk_num):
+        chunks.append(list_of_files[chunk_begin:chunk_end])
+        chunk_begin += chunk_size
+        if chunk_begin >= len(list_of_files):
+            break
+        chunk_end += chunk_size
+        if chunk_end > len(list_of_files):
+            chunk_end = len(list_of_files)
```
fgerlits, line 52: this is quite ugly; would it be overkill to install `toolz`? then we could do
```suggestion
    from toolz import partition_all
    
    chunks = list(partition_all(chunk_num, list_of_files))
```

fgerlits, line 52: this is quite ugly; would it be overkill to install `toolz`? then we could do
```suggestion
    from toolz import partition_all
    
    chunks = list(partition_all(chunk_size, list_of_files))
```

adam-markovics: Seems like a good idea, but
"If the length of seq is not evenly divisible by n, the final tuple is dropped if pad is not specified, or filled to length n by pad"
https://toolz.readthedocs.io/en/latest/api.html#toolz.itertoolz.partition
I don't think we should drop files from being linted, and my algorithm does not.
But I'm open to suggestions. Another question is, how would I have to specify the installation of this module? And is it worth it for 10 lines?

szaszm: Would this mean that in order to run the linter, we all need to install this package system-wide? If so, it may not be worth the extra effort required to simplify this code. One solution could be using a virtualenv in the build directory, but this also sounds more complex than maintaining 10 extra lines of code. What do you think?

fgerlits: `pip/pip3` installs the packages to the current user's home directory, so not quite system-wide, but yeah, maybe not worth it

yes, https://toolz.readthedocs.io/en/latest/api.html#toolz.itertoolz.partition is not what we need; I suggested https://toolz.readthedocs.io/en/latest/api.html#toolz.itertoolz.partition_all which allows the last chunk to be smaller

how about factoring out the chunking code; I think that would improve the readability without introducing a new dependency:
```suggestion
def create_chunks(chunk_size, items):    # move to before if __name__ == '__main__':
    chunks = []
    chunk_begin = 0
    chunk_end = chunk_size
    for chunk_cnt in range(chunk_num):
        chunks.append(items[chunk_begin:chunk_end])
        chunk_begin += chunk_size
        if chunk_begin >= len(list_of_files):
            break
        chunk_end += chunk_size
        if chunk_end > len(list_of_files):
            chunk_end = len(list_of_files)

chunks = create_chunks(chunk_size, list_of_files)
```

adam-markovics: I committed a slightly modified version. Do you agree with that one?

fgerlits: yes, looks good, thanks!

fgerlits (approved):

PR #1232, opened 2022-01-04
MINIFICPP-1706 Rework script engine management in ExecuteScript processor

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-1706

------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/script/ExecuteScript.h
```diff
@@ -28,17 +28,83 @@
 
 #include "ScriptEngine.h"
 #include "ScriptProcessContext.h"
+#include "python/PythonScriptEngine.h"
```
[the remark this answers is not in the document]
lordgamez: Good catch! I added additional compile flag checks to be consistent with the previous Lua and Python support checks in 108a39ccc89cdf6d229b19f5a80545ecb925c993

extensions/script/ExecuteScript.cpp
```diff
@@ -89,40 +113,30 @@ void ExecuteScript::onTrigger(const std::shared_ptr<core::ProcessContext> &conte
...
-    }
+  }
+
+  if (engine == nullptr) {
+    throw std::runtime_error("No script engine available");
+  }
+
+  if (!script_body_.empty()) {
+    engine->eval(script_body_);
+  } else if (!script_file_.empty()) {
```
adam-markovics, line 140: If we have both script_body_ and script_file_ (erroneously), will we just silently execute script_body_, without telling the user about the misconfiguration?

extensions/script/ExecuteScript.cpp
```diff
@@ -52,6 +52,28 @@ core::Property ExecuteScript::ModuleDirectory("Module Directory",  // NOLINT
...
+    : max_engine_count_(max_engine_count),
+      engine_factory_(engine_factory),
+      logger_(logger) {
+}
+
+void ScriptEngineQueue::returnScriptEngine(std::shared_ptr<script::ScriptEngine>&& engine) {
+  const std::lock_guard<std::mutex> lock(queue_mutex_);
+  if (engine_queue_.size_approx() < max_engine_count_) {
+    logger_->log_debug("Releasing [%p] script engine", engine.get());
+    engine_queue_.enqueue(engine);
```
adam-markovics, line 71: std::move could be used

extensions/script/ExecuteScript.cpp
lordgamez: Good point, I added additional checks and tests for misconfiguration use cases in fb9b4e78d3762de139ca065b3044f572ca399a14

extensions/script/ExecuteScript.cpp
lordgamez: Updated in 268d10998075d418fa7fca9fbd3d318996bda6bd

extensions/script/tests/TestExecuteScriptProcessorWithPythonScript.cpp
```diff
@@ -27,7 +27,47 @@
...
+TEST_CASE("Test both script body and script file set", "[executescriptMisconfiguration]") {
+  TestController testController;
+  auto plan = testController.createPlan();
+
+  auto executeScript = plan->addProcessor("ExecuteScript", "executeScript");
+
+  plan->setProperty(executeScript, minifi::processors::ExecuteScript::ScriptEngine.getName(), "python");
+  plan->setProperty(executeScript, minifi::processors::ExecuteScript::ScriptFile.getName(), "/path/to/script.py");
+  plan->setProperty(executeScript, minifi::processors::ExecuteScript::ScriptBody.getName(), R"(
+    function onTrigger(context, session)
```
adam-markovics, line 62: I think Python code would suit better here. Just to make sure the exception is not coming because of the wrong code syntax.

lordgamez: Thanks! Fixed in a1eae0e46cf88802e7e158e235cec6bffe801a0b

adam-markovics (approved):

extensions/script/ExecuteScript.h
```diff
@@ -28,17 +28,95 @@
...
+
 namespace processors {
 
+class ScriptEngineFactory {
+ public:
+  ScriptEngineFactory(core::Relationship& success, core::Relationship& failure, std::shared_ptr<core::logging::Logger> logger);
+
+  template<typename T>
+  std::shared_ptr<T> createEngine() const {
+    auto engine = std::make_shared<T>();
```
szaszm, line 52: I think it would make sense to add some restriction, e.g. that `T` is derived from `ScriptEngine`.
```suggestion
  template<typename T>
  std::enable_if_t<std::is_base_of_v<script::ScriptEngine, T>, std::shared_ptr<T>> createEngine() const {
    auto engine = std::make_shared<T>();
```

extensions/script/ExecuteScript.h
```diff
@@ -28,17 +28,95 @@
...
+  core::Relationship& failure_;
+  std::shared_ptr<core::logging::Logger> logger_;
+};
+
+class ScriptEngineQueue {
+ public:
+  ScriptEngineQueue(uint8_t max_engine_count, ScriptEngineFactory& engine_factory, std::shared_ptr<core::logging::Logger> logger);
+
+  template<typename T>
+  std::shared_ptr<script::ScriptEngine> getScriptEngine() {
```
szaszm, line 72: How is it enforced that the returned script engine is of type `T`, especially if it's taken out of the queue? It looks like the queue can contain any kind of script engine and this function just takes out whatever is at the end.

extensions/script/ExecuteScript.h
lordgamez: Good idea, added restriction in e1583ef65af13edcd0d5b07da6eee5ea84ba3cbd

extensions/script/ExecuteScript.h
lordgamez: Good point, there is no point of having different types of script engines in a queue. Added a class level template parameter for the engine type and similar type restriction to `createEngine` in e1583ef65af13edcd0d5b07da6eee5ea84ba3cbd

szaszm: The error of the docker integration test job is related, check it out! https://github.com/apache/nifi-minifi-cpp/runs/4761689961?check_suite_focus=true#step:4:8518

lordgamez: My bad I didn't have the LUA_SUPPORT flag enabled when compiled locally, the compilation issues should be fixed in 9c96d09addaa6516382a2d171f002e14f4151099

extensions/script/tests/CMakeLists.txt
```diff
@@ -30,6 +30,7 @@ endif()
 
 if (ENABLE_LUA_SCRIPTING)
 	file(GLOB EXECUTESCRIPT_LUA_TESTS "TestExecuteScriptProcessorWithLuaScript.cpp" "LuaScriptEngineTests.cpp")
+	include_directories(${LUA_INCLUDE_DIR})
```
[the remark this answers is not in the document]
lordgamez: Thanks for the tip, I wasn't sure if there's a better way to have those includes available for all test targets without duplication, but making it PUBLIC seems to work, changed it in 5914deef48db52f59a26dcf432e2af8262841001

szaszm (approved):

martinzink (approved):

PR #1234, opened 2022-01-06
MINIFICPP-1699 Add kubernetes-client/c library dependency

fgerlits:
https://issues.apache.org/jira/browse/MINIFICPP-1699

This is just the inclusion of the C Kubernetes client library for now, actual functionality will come later.  The `minifi-kubernetes` executable is just a POC to test that it compiles and links, I will remove it later and replace it with the actual extension lib.

---

### For all changes:
     in the commit message?

### For code changes:

### Note:


thirdparty/kubernetes-client-c/remove-findpackage.patch
```diff
@@ -0,0 +1,12 @@
...
+index 93ea00e..1e064b8 100644
+--- a/kubernetes/PreTarget.cmake
++++ b/kubernetes/PreTarget.cmake
+@@ -41,6 +41,3 @@ list(APPEND HDRS
+         websocket/kube_exec.h
+         include/generic.h
+         include/utils.h)
+-
+-find_package(libwebsockets CONFIG REQUIRED)
+-find_package(yaml CONFIG REQUIRED)
```
szaszm, line 11: Consider adding dummy find modules under `cmake/libwebsockets/dummy` and `cmake/yaml/dummy`. See the existing ones for example. These directories are added to the cmake module path in our root CMakeLists.txt. This approach would make find_package work.

szaszm (approved):

cmake/BundledLibreSSL.cmake
```diff
@@ -60,6 +60,7 @@ function(use_libre_ssl SOURCE_DIR BINARY_DIR)
     # Set variables
     set(OPENSSL_FOUND "YES" CACHE STRING "" FORCE)
     set(OPENSSL_INCLUDE_DIR "${LIBRESSL_BIN_DIR}/include" CACHE STRING "" FORCE)
+    set(OPENSSL_INCLUDE_DIRS "${OPENSSL_INCLUDE_DIR}" CACHE STRING "" FORCE)  # workaround for libwebsockets
```
lordgamez, line 63: If some thirdparties require this maybe OPENSSL_INCLUDE_DIRS should be added to FindOpenSSL.cmake as well.

extensions/kubernetes/KubernetesClientPOC.cpp
```diff
@@ -0,0 +1,123 @@
...
+        list_ForEach(listEntry, pod_list->items) {
+            pod = static_cast<v1_pod_t *>(listEntry->data);
+            printf("\tThe pod name: %s\n", pod->metadata->name);
+            v1_container_t *container = NULL;
+        listEntry_t *containerEntry = NULL;
+        list_ForEach(containerEntry, pod->spec->containers) {
+        container = static_cast<v1_container_t *>(containerEntry->data);
+        printf("\tThe container name: %s\n", container->name);
+            }
+    }
```
lordgamez, line 60: The indentation is a bit off in these lines. I also think that we should change the indentation of the file from 4 to 2 spaces to be aligned with our style guide.

thirdparty/kubernetes-client-c/remove-findpackage.patch
fgerlits: I think adding `Find*.cmake` files would only work for `find_package` in `MODULE` mode, and this is `CONFIG` mode, so CMake is looking for `yamlConfig.cmake` and `libwebsockets-config.cmake`.  These are, in fact, generated by CMake (at `build/_deps/yaml-build/CMakeFiles/yamlConfig.cmake` in the case of `libyaml`), but I haven't found a way of making CMake look for it there.  This issue: https://gitlab.kitware.com/cmake/cmake/-/issues/21687 may be about adding this capability to CMake, but it has not been resolved.

cmake/BundledLibreSSL.cmake
fgerlits: I think this is more a bug in `libwebsockets`.  It should be reading `OPENSSL_INCLUDE_DIR`, but (in the case when OpenSSL is part of the project being compiled and not installed on the system) it looks for `OPENSSL_INCLUDE_DIRS` instead.

A better solution may be to patch `libwebsockets` to correctly look for `OPENSSL_INCLUDE_DIR` (and submit the patch upstream).  Do you think it's worth doing that?

thirdparty/kubernetes-client-c/remove-findpackage.patch
szaszm: According to the `find_package` docs, you could set `<PackageName>_ROOT` (`libwebsockets_ROOT` and `yaml_ROOT` in this case) to this directory and it would be found by `find_package`. The issue is about doing this automatically in cmake.

extensions/kubernetes/KubernetesClientPOC.cpp
fgerlits: fixed in https://github.com/apache/nifi-minifi-cpp/pull/1234/commits/000894de4fe336b62711c75dda76d0120e9a38a2

cmake/BundledLibreSSL.cmake
szaszm: If you are willing to do that, I think it's a great idea to contribute this improvement to libwebsockets.

lordgamez: I'm not sure if it's a bug or not because I saw that in case of other 3rd party libraries both *INCLUDE_DIR and *INCLUDE_DIRS variables are defined being interchangeable. That's why I thought this might be the case here too, so it would worth defining both if any other library would use either of them. If it's actually a bug I think this workaround is okay it's not worth patching it.

szaszm: According to [this](https://discourse.cmake.org/t/findboost-difference-between-boost-include-dir-and-boost-include-dirs/855), INCLUDE_DIR is its own include dir while INCLUDE_DIRS is all of the include dirs needed to use the project, at least in case of Boost. [They are identical](https://github.com/Kitware/CMake/blob/master/Modules/FindBoost.cmake#L2351), though, so I'm not sure if that's actually the case.

fgerlits: I think it's very likely to be a bug.  Suggestion: I'll leave this as it is for now and submit a patch to `libwebsockets`.  If they accept it, and a new version with the fix becomes availabe, I will upgrade `libwebsockets` and remove this workaround.

lordgamez: I'm okay with that, that should clear all the ambiguity eventually.

lordgamez (approved):

thirdparty/kubernetes-client-c/remove-findpackage.patch
fgerlits: Thanks, I did not know about the `*_ROOT` variable.  I have tried to make this work, but
 * I still need to patch the kubernetes lib to set the [CMP0074](https://cmake.org/cmake/help/latest/policy/CMP0074.html) policy to NEW, and
 * the generated `yamlConfig.cmake` and `libwebsockets-config.cmake` files don't quite work: `yamlConfig.cmake` refers to a `yamlTargets.cmake` file which does not exist, and `libwebsockets-config.cmake` tries to `include()` `LibwebsocketsTargets.cmake`, which does exist, but CMake says it "was generated by the export() command.  It may not be used as the argument to the include() command.  Use ALIAS targets instead to refer to targets by alternative names."

I'm sure all these problems could be solved, but at this point I think the least messy solution is to leave this as it is.

martinzink (approved):

cmake/BundledLibreSSL.cmake
fgerlits: I have opened this PR: https://github.com/warmcat/libwebsockets/pull/2535.  I haven't got any response yet, but I'll keep an eye on it.

adam-markovics (approved):

PR #1237, opened 2022-01-12
MINIFICPP-1224 Add module directory support to script execution

lordgamez:
Adding `Module Directory` property support in `ExecuteScript` and `ExecutePythonPorcessor` processors. The module paths are added to the executed script's module paths to find the used lua modules and python packages.

https://issues.apache.org/jira/browse/MINIFICPP-1224

-------------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/script/ScriptEngine.h
```diff
@@ -41,7 +43,14 @@ class ScriptEngine {
...
   virtual void evalFile(const std::string &file_name) = 0;
 
+  void setModuleDirectories(std::vector<std::string>&& module_dirs) {
+    module_directories_ = std::move(module_dirs);
+  }
+
   virtual ~ScriptEngine() = default;
+
+ protected:
+  std::vector<std::string> module_directories_;
```
szaszm, line 53: This should probably be a vector of `std::filesystem::path` for stronger typing.

extensions/script/ExecuteScript.cpp
```diff
@@ -116,6 +121,9 @@ void ExecuteScript::onTrigger(const std::shared_ptr<core::ProcessContext> &conte
       throw std::runtime_error("No script engine available");
     }
 
+    if (module_directory_.size()) {
+      engine->setModuleDirectories(utils::StringUtils::splitAndTrimRemovingEmpty(module_directory_, ","));
```
szaszm, line 125: Not sure if you want to support weird filenames, just want to point out that filenames can start/end with spaces or contain commas.

extensions/script/lua/LuaScriptEngine.cpp
```diff
@@ -65,15 +65,32 @@ LuaScriptEngine::LuaScriptEngine()
 
 void LuaScriptEngine::eval(const std::string &script) {
   try {
-    lua_.script(script, sol::script_throw_on_error);
+    if (!module_directories_.empty()) {
+      auto appended_script = script;
+      for (const auto& module_dir : module_directories_) {
+        appended_script = "package.path = package.path .. \";" + module_dir + "/?.lua\"\n" + script;
+      }
+      lua_.script(appended_script, sol::script_throw_on_error);
```
szaszm, line 73: Please extract this logic and reuse it in `evalFile`

lordgamez: Updated in c5bc09b98061f1d2295e38e4b1d63795c3eafdfd

extensions/script/ScriptEngine.h
lordgamez: I'm all for strong types, but I'm not sure if it's worth here as all these paths are used only in string concatenation so we would only convert these to paths just to convert them back to strings when used. Do you see any greater benefits for using path types?

extensions/script/ExecuteScript.cpp
lordgamez: I think regarding spaces, it would be a more common mistake to leave a space at the beginning, the end or between comma-separated entries than using a file path with actual space in it. I'm not sure if it's worth the effort supporting commas in file paths here, if it would ever be a use case maybe it can be requested.

extensions/script/ScriptEngine.h
szaszm: If it's that simple, we can leave it as is.

szaszm (approved):

extensions/script/ExecuteScript.cpp
```diff
@@ -123,6 +128,10 @@ void ExecuteScript::onTrigger(const std::shared_ptr<core::ProcessContext> &conte
     throw std::runtime_error("No script engine available");
   }
 
+  if (module_directory_.size()) {
```
adamdebreceni, line 131: should we use `!module_directory_.empty()`?

(or better yet `std::optional<std::string>`, although I see that this member is not a new addition)

lordgamez: Updated module directory to be optional in f6be6d51339473f54b18c103fdef812b47c759ee

extensions/librdkafka/ConsumeKafka.cpp
```diff
@@ -211,7 +211,7 @@ void ConsumeKafka::onSchedule(core::ProcessContext* context, core::ProcessSessio
   context->getProperty(DuplicateHeaderHandling.getName(), duplicate_header_handling_);
 
   headers_to_add_as_attributes_ = utils::listFromCommaSeparatedProperty(*context, HeadersToAddAsAttributes.getName());
-  max_poll_records_ = gsl::narrow<std::size_t>(utils::getOptionalUintProperty(*context, MaxPollRecords.getName()).value_or(DEFAULT_MAX_POLL_RECORDS));
+  max_poll_records_ = gsl::narrow<std::size_t>(utils::getOptionalProperty<uint64_t>(*context, MaxPollRecords.getName()).value_or(DEFAULT_MAX_POLL_RECORDS));
```
adamdebreceni, line 214: `ProcessContext::getProperty<T>(const Property&)` already returns an `std::optional<T>`, does `getOptionalProperty` behave differently?

lordgamez: I wasn't aware of this, in that case this is not needed at all, removed it in ddfa1b976649599300457af5d5b4100f948d6a43

adamdebreceni (approved):

martinzink (approved):

PR #1238, opened 2022-01-12
MINIFICPP-1718 - Handle case sensitive column names

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/sql/data/SQLIdentifier.h
```diff
@@ -19,18 +19,40 @@
...
 
-namespace org {
-namespace apache {
-namespace nifi {
-namespace minifi {
-namespace utils {
+namespace org::apache::nifi::minifi::sql {
 
-std::vector<std::string> inputStringToList(const std::string& str);
+class SQLIdentifier {
```
szaszm, line 26: I suggest renaming this to something like `SqlColumnName` to avoid confusion. "Identifier" typically refers to a primary key in the relational database context. "Column" is not the most precise term (that would be "attribute"), but it's the most intuitive.
And I suggest avoiding full-caps acronyms in CamelCase, because it makes it more difficult to determine word boundaries (e.g. `IXMLHTTPRequest`) and to programmatically convert between conventions (e.g. `s_q_l_identifier` instead of `sql_identifier`). It's also recommended in the [google style guide](https://google.github.io/styleguide/cppguide.html#General_Naming_Rules).

docker/test/integration/features/sql.feature
```diff
@@ -32,6 +32,21 @@ Feature: Executing SQL operations from MiNiFi-C++
...
+    Given a GenerateFlowFile processor with the "File Size" property set to "0B"
+    And a UpdateAttribute processor with the "sql.args.1.value" property set to "ApPlE"
+    And the "sql.args.2.value" property of the UpdateAttribute processor is set to "BaNaNa"
+    And a ExecuteSQL processor with the "SQL select query" property set to "SELECT * FROM test_table2 WHERE "tExT_Col" = ? OR "tExT_Col" = ? ORDER BY int_col DESC"
+    And the "Output Format" property of the ExecuteSQL processor is set to "JSON"
+    And a PutFile processor with the "Directory" property set to "/tmp/output"
+    And the "success" relationship of the GenerateFlowFile processor is connected to the UpdateAttribute
+    And the "success" relationship of the UpdateAttribute processor is connected to the ExecuteSQL
+    And the "success" relationship of the ExecuteSQL processor is connected to the PutFile
+    And an ODBCService is setup up for ExecuteSQL with the name "ODBCService" and connection string "Driver={PostgreSQL ANSI};Server=postgresql-server;Port=5432;Database=postgres;Uid=postgres;Pwd=password;"
```
lordgamez, line 45: This step could be refactored to have the name and the connection string hardcoded in the step implementation. I don't think it's going to change in any use case and we can remove the duplication this way.

docker/test/integration/features/sql.feature
```diff
@@ -32,6 +32,21 @@ Feature: Executing SQL operations from MiNiFi-C++
     When all instances start up
     Then at least one flowfile with the content '[{"int_col":2,"text_col":"banana"},{"int_col":1,"text_col":"apple"}]' is placed in the monitored directory in less than 120 seconds
 
+  Scenario: A MiNiFi instance can query to test table containing mixed case column names with ExecuteSQL processor
+    Given a GenerateFlowFile processor with the "File Size" property set to "0B"
+    And a UpdateAttribute processor with the "sql.args.1.value" property set to "ApPlE"
+    And the "sql.args.2.value" property of the UpdateAttribute processor is set to "BaNaNa"
+    And a ExecuteSQL processor with the "SQL select query" property set to "SELECT * FROM test_table2 WHERE "tExT_Col" = ? OR "tExT_Col" = ? ORDER BY int_col DESC"
```
lordgamez, line 39: We could add a comment that PostgreSQL's quoted columns mark case sensitive queries just to be clear.

extensions/sql/data/SQLIdentifier.cpp
```diff
@@ -16,30 +16,26 @@
...
 
-#include "Utils.h"
+#include "SQLIdentifier.h"
 
-#include <vector>
-#include <string>
+namespace org::apache::nifi::minifi::sql {
 
-#include "utils/StringUtils.h"
+SQLIdentifier::SQLIdentifier(std::string str) {
```
lordgamez, line 23: As it is a self-contained entity I would add some simple unit tests for it.

extensions/sql/data/SQLIdentifier.h
adamdebreceni: a quick googling seems to suggest [otherwise](https://www.google.com/search?q=sql+identifier)

adamdebreceni: as for the capitalization, all other SQL entities in this extension have all-caps "SQL" in them, so I would stick to it for consistency

szaszm: fair. What do you think about `SQLColumnIdentifier`? Or do you intend this to be applicable to table and other identifiers as well?

docker/test/integration/features/sql.feature
adamdebreceni: moved the string into the step

docker/test/integration/features/sql.feature
adamdebreceni: added a comment

extensions/sql/data/SQLIdentifier.cpp
adamdebreceni: added unit tests

lordgamez (approved):

extensions/sql/data/SQLIdentifier.h
adamdebreceni: I think we can constrain it to columns for now, renamed it to `SQLColumnIdentifier` (if we need to use something like this for tables/others we can revisit this)

szaszm (approved):

martinzink (approved):

PR #1243, opened 2022-01-18
MINIFICPP-1696 Clean up the condition checking in behavior-based tests

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-1696

---------------------------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/docker/test/integration/MiNiFi_integration_test_driver.py
+++ b/docker/test/integration/MiNiFi_integration_test_driver.py
@@ -118,57 +118,57 @@ class MiNiFi_integration_test():
     def add_file_system_observer(self, file_system_observer):
         self.file_system_observer = file_system_observer
 
-    def check_for_no_files_generated(self, timeout_seconds):
+    def check_for_no_files_generated(self, wait_time_in_seconds):
         output_validator = NoFileOutPutValidator()
         output_validator.set_output_dir(self.file_system_observer.get_output_dir())
-        self.check_output(timeout_seconds, output_validator, 1)
+        self.__check_output_after_time_period(wait_time_in_seconds, output_validator)
 
     def check_for_single_file_with_content_generated(self, content, timeout_seconds):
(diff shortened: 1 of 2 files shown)


docker/test/integration/MiNiFi_integration_test_driver.py
```diff
@@ -118,57 +118,57 @@ def rm_out_child(self):
...
     def check_for_an_empty_file_generated(self, timeout_seconds):
         output_validator = EmptyFilesOutPutValidator()
         output_validator.set_output_dir(self.file_system_observer.get_output_dir())
-        self.check_output(timeout_seconds, output_validator, 1)
+        self.__check_output(timeout_seconds, output_validator, 1)
 
-    def check_output_force_wait(self, timeout_seconds, output_validator):
-        time.sleep(timeout_seconds)
-        self.validate(output_validator)
+    def __check_output_after_time_period(self, wait_time_in_seconds, output_validator):
```
szaszm, line 161: Did you mean to make these "private"? According to some random internet sources, a double underscore prefix makes the interpreter use name mangling to avoid subclasses overriding these, but I see no subclasses here. The usual way to say "private" is a single underscore prefix, which seems more appropriate here, but let me know if I'm wrong.

lordgamez: I usually use the double underscore for making a member "private" because of the mangling to make it harder for a user of the class to call it. I was not aware that it is only used for subclasses not to override these. In my understanding it was usually to use double underscore for private, single for protected and none for public members and I would default to using "private" members if no subclasses are there to use them.

szaszm: Okay, this sounds reasonable. Thanks for clarifying.

szaszm (approved):

adam-markovics (approved):

martinzink (approved):

PR #1244, opened 2022-01-18
MINIFICPP-1727 Add ENABLE_LUA_SCRIPTING flag to docker

martinzink:

### For all changes:
     in the commit message?

### For code changes:

### Note:


docker/Dockerfile
```diff
@@ -159,6 +163,7 @@ RUN addgroup -g ${GID} ${USER} && adduser -u ${UID} -D -G ${USER} -g "" ${USER}
     if [ "$ENABLE_JNI" = "ON" ]; then apk add --no-cache openjdk8-jre-base; fi && \
     if [ "$ENABLE_PCAP" = "ON" ]; then apk add --no-cache libpcap; fi && \
     if [ "$ENABLE_USB_CAMERA" = "ON" ]; then apk add --no-cache libpng libusb; fi && \
+    if [ "$ENABLE_LUA_SCRIPTING" = "ON" ]; then apk add --no-cache lua; fi && \
```
lordgamez, line 166: I think we should also add a check for ENABLE_SCRIPTING flag here. Also just a thought but maybe we could also include the ENABLE_ALL check here as well for all of these optional packages.

docker/Dockerfile
```diff
@@ -159,6 +163,7 @@ RUN addgroup -g ${GID} ${USER} && adduser -u ${UID} -D -G ${USER} -g "" ${USER}
...
     if [ "$ENABLE_OPENCV" = "ON" ] || [ "$ENABLE_BUSTACHE" = "ON" ]; then apk add --no-cache boost; fi && \
     if { [ "$ENABLE_PYTHON" = "ON" ] || [ "$ENABLE_SCRIPTING" = "ON" ]; } && [ -z "$DISABLE_PYTHON_SCRIPTING" ]; then apk add --no-cache python3-dev; fi
```
lordgamez, line 168: Not strictly related to this PR, but python3-dev should not be needed here, only the python3 package. Also ENABLE_PYTHON check can be removed here as it turns out that it's unrelated to the python scripting extension. (It's an unused nanofi based extension.)

docker/Dockerfile
```diff
@@ -61,6 +61,7 @@ ARG DISABLE_LZMA=OFF
 ARG DISABLE_BZIP2=OFF
 ARG ENABLE_SCRIPTING=ON
```
lordgamez, line 62: Should we have scripting enabled by default? I just realized this and because the DISABLE_PYTHON_SCRIPTING flag is off by default, it automatically installs python packages which adds 80MBs to the image. I'm not sure if scripting should be enabled. (If it's needed by the users of the published minimal image then it should be enabled in that config)

martinzink: I am not sure about this, the minimal image explicitly disables this though
https://github.com/apache/nifi-minifi-cpp/blob/main/cmake/DockerConfig.cmake#L77-L78

lordgamez: In that case as that is our released docker image I think we can disable it by default. It shouldn't have any impact and I don't think it should be included if not specified.

docker/Dockerfile
martinzink: I added the ENABLE_SCRIPTING check in https://github.com/apache/nifi-minifi-cpp/pull/1244/commits/c222bf84612ca4ff4bd85c6cf5dc79b03cbc1fed
i would skip the ENABLE_ALL for now because it is not that consistent (e.g. its not even considered when building the minifi-scripting library https://github.com/apache/nifi-minifi-cpp/blob/main/extensions/script/CMakeLists.txt#L47 ) 
I think we should refactor this whole ENABLE_ALL flag in a different ticket

docker/Dockerfile
martinzink: good idea changed it in https://github.com/apache/nifi-minifi-cpp/pull/1244/commits/c222bf84612ca4ff4bd85c6cf5dc79b03cbc1fed

docker/Dockerfile
lordgamez: That's a good point, I agree.

docker/Dockerfile
szaszm: I prefer disabled by default. We do the same in cmake because libpython is a problematic dependency for shipping binary packages. It changes soname every minor version, which breaks compatibility with any other version.

martinzink: changed it in https://github.com/apache/nifi-minifi-cpp/pull/1244/commits/cf4fbe4c44e764f775278784a06ee2aef6893b84

lordgamez (approved):

szaszm (approved):

adamdebreceni (approved):

martinzink:
All changes have been incorporated into https://github.com/apache/nifi-minifi-cpp/pull/1232

PR #1251, opened 2022-01-27
MINIFICPP-1636 - Use std::filesystem instead of boost and platform specific implementations

rustammendel:

### For all changes:
     in the commit message?

- [] Is your initial contribution a single, squashed commit?

### For code changes:

### Note:

### DIFF
--- a/docker/DockerVerify.sh
+++ b/docker/DockerVerify.sh
@@ -44,11 +44,11 @@ echo "Installing test dependencies..." 1>&2
 # hint include/library paths if homewbrew is in use
 if brew list 2> /dev/null | grep openssl > /dev/null 2>&1; then
   echo "Using homebrew paths for openssl" 1>&2
-  LDFLAGS="-L$(brew --prefix openssl)/lib"
+  LDFLAGS="-L$(brew --prefix openssl@1.1)/lib"
   export LDFLAGS
-  CFLAGS="-I$(brew --prefix openssl)/include"
+  CFLAGS="-I$(brew --prefix openssl@1.1)/include"
   export CFLAGS
-  SWIG_FEATURES="-cpperraswarn -includeall -I$(brew --prefix openssl)/include"
+  SWIG_FEATURES="-cpperraswarn -includeall -I$(brew --prefix openssl@1.1)/include"
   export SWIG_FEATURES
 fi
 
--- a/extensions/script/python/PythonCreator.h
+++ b/extensions/script/python/PythonCreator.h
@@ -121,7 +121,7 @@ class PythonCreator : public minifi::core::CoreComponent {
     }
 
     for (const auto &path : pathOrFiles) {
-      utils::file::FileUtils::addFilesMatchingExtension(logger_, path, ".py", classpaths_);
+      utils::file::addFilesMatchingExtension(logger_, path, ".py", classpaths_);
     }
   }
 
--- a/extensions/script/python/PythonCreator.h
+++ b/extensions/script/python/PythonCreator.h
@@ -125,16 +125,15 @@ class PythonCreator : public minifi::core::CoreComponent {
     }
   }
 
-  std::string getPackage(const std::string& basePath, const std::string& pythonscript) {
+  std::string getPackage(const std::string &basePath, const std::string &pythonscript) {
     const auto script_directory = getPath(pythonscript);
     const auto loc = script_directory.find_first_of(basePath);
     if (loc != 0 || script_directory.size() <= basePath.size()) {
       return "";
     }
-    const auto python_dir = script_directory.substr(basePath.length() + 1);
-    auto python_package = python_dir.substr(0, python_dir.find_last_of("/\\"));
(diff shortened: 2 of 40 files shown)


libminifi/include/utils/file/FileUtils.h
```diff
@@ -25,14 +25,7 @@
...
 
-#ifdef USE_BOOST
-#include <dirent.h>
-#include <boost/filesystem.hpp>
-#include <boost/system/error_code.hpp>
-#ifndef WIN32
-#include <sys/stat.h>
-#endif
-#else
+#ifndef USE_BOOST
```
szaszm, line 28: We can now remove this check and just assume that it's not defined. We will eventually get rid of all boost code from libminifi.

docker/DockerVerify.sh
```diff
@@ -44,11 +44,11 @@ echo "Installing test dependencies..." 1>&2
...
 if brew list 2> /dev/null | grep openssl > /dev/null 2>&1; then
   echo "Using homebrew paths for openssl" 1>&2
-  LDFLAGS="-L$(brew --prefix openssl)/lib"
+  LDFLAGS="-L$(brew --prefix openssl@1.1)/lib"
   export LDFLAGS
-  CFLAGS="-I$(brew --prefix openssl)/include"
+  CFLAGS="-I$(brew --prefix openssl@1.1)/include"
   export CFLAGS
-  SWIG_FEATURES="-cpperraswarn -includeall -I$(brew --prefix openssl)/include"
+  SWIG_FEATURES="-cpperraswarn -includeall -I$(brew --prefix openssl@1.1)/include"
```
szaszm, line 51: Why is this version pin? Was something not working on mac?

szaszm (commented):
Thank you for your contribution, this is an awesome and very welcome change. I only have 2 small comments, which is super rare in a 500 line diff, especially in a first contribution.

docker/DockerVerify.sh
rustammendel: Yes SWIG libs have incompatibility with openssl@3, it was failing on local.

libminifi/include/utils/file/FileUtils.h
rustammendel: I have tried removing it but that failed windows build. I will try to optimize imports by moving relevant includes to ifdef WIN32 scope

szaszm (approved):

martinzink (approved):
Thanks for your contribution. This PR certainly makes FileUtils more readable than before. :+1:

extensions/sftp/tests/PutSFTPTests.cpp
```diff
@@ -504,7 +504,9 @@ TEST_CASE_METHOD(PutSFTPTestsFixture, "PutSFTP set mtime", "[PutSFTP]") {
   testController.runSession(plan, true);
 
   testFile("nifi_test/tstFile1.ext", "content 1");
-  testModificationTime("nifi_test/tstFile1.ext", 3000000000LL);
+  using namespace std::chrono;  // NOLINT(build/namespaces)
```
lordgamez, line 507: nitpick: I would use `std::literals::chrono_literals` namespace here instead as we allow that in the linter and use the fully qualified namespaces for the type declaration.

lordgamez (approved):
LGTM, thanks for this change!

extensions/script/python/PythonCreator.h
```diff
@@ -121,41 +122,34 @@ class PythonCreator : public minifi::core::CoreComponent {
...
-    const auto python_dir = script_directory.substr(basePath.length() + 1);
-    auto python_package = python_dir.substr(0, python_dir.find_last_of("/\\"));
-    utils::StringUtils::replaceAll(python_package, "/", ".");
-    utils::StringUtils::replaceAll(python_package, "\\", ".");
+    const auto python_package_path = std::filesystem::relative(pythonscript, basePath).parent_path();
+    std::vector<std::string> path_elements{python_package_path.begin(), python_package_path.end()};
+    std::string python_package = minifi::utils::StringUtils::join(".", path_elements);
     if (python_package.length() > 1 && python_package.at(0) == '.') {
       python_package = python_package.substr(1);
     }
```
fgerlits, line 138: I don't think this is needed any longer, as none of the path elements can be empty, and `StringUtils::join()` is not going to put a separator at the front.

rustammendel: image
on my local it does

fgerlits: That is very strange; I get "h2o.h2o3.mojo".  There is probably a bug somewhere, but it was most likely already there before this change, so we don't have to fix it now.

adam-markovics (approved):

fgerlits (approved):

szaszm:
I copied the name and email address from the last commit on your branch when merging, but GitHub seems to associate it with a different account of yours. I hope it's not a problem.

PR #1259, opened 2022-02-08
MINIFICPP-1748 - Modify log properties through the c2 protocol

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/http-curl/tests/C2PropertiesUpdateTests.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+static std::string properties_file =
+    "nifi.property.one=tree\n"
+    "nifi.c2.agent.protocol.class=RESTSender\n"
+    "nifi.c2.enable=true\n"
+    "nifi.c2.agent.class=test\n"
+    "nifi.c2.agent.heartbeat.period=100\n";
+
+static std::string log_properties_file =
+    "logger.root=INFO,ostream\n";
```
lordgamez, line 99: Minor: These could be constants

libminifi/include/properties/Configure.h
```diff
@@ -42,12 +44,23 @@ class Configure : public Configuration, public core::AgentIdentificationProvider
...
   void setFallbackAgentIdentifier(const std::string& id);
 
+  using Configuration::set;
+  void set(const std::string& key, const std::string& value, PropertyChangeLifetime lifetime) override;
+  bool commitChanges() override;
+
+  // WARNING! a test utility
+  void setLoggerProperties(std::shared_ptr<core::logging::LoggerProperties> new_properties) {
+    logger_properties_ = new_properties;
+  }
```
lordgamez, line 54: If this is only for testing could this be made private and declare the test class be a friend class?

libminifi/include/properties/Properties.h
```diff
@@ -29,16 +29,23 @@
...
+  PERSISTENT  // the changed value will be written to the source file
+};
+
 class Properties {
   struct PropertyValue {
-    std::string value;
-    bool changed;
+    std::string persisted_value;
+    std::string active_value;
+    bool changed;  // persisted_value should be written to disk if requested
```
lordgamez, line 48: I would add a default `false` not to leave it uninitialized to be fool proof.

libminifi/include/properties/Configure.h
adamdebreceni: done

libminifi/include/properties/Properties.h
adamdebreceni: done

libminifi/include/core/Core.h
```diff
@@ -80,13 +81,16 @@ static inline std::string getClassName() {
...
-  const std::string clazz = "class ";
-  auto haz_clazz = adjusted_name.find(clazz);
-  if (haz_clazz == 0)
-    adjusted_name = adjusted_name.substr(clazz.length(), adjusted_name.length() - clazz.length());
-  return adjusted_name;
+  std::string_view name = typeid(T).name();
+  const std::string_view class_prefix = "class ";
+  const std::string_view struct_prefix = "struct ";
+
+  if (name.find(class_prefix) == 0) {
```
martinzink, line 88: Since C++20 we can use [starts_with](https://en.cppreference.com/w/cpp/string/basic_string_view/starts_with) which is a bit more readable

martinzink (commented):
LGTM, only a couple minor suggestions

libminifi/include/core/Core.h
lordgamez: We also have a `startsWith` function in the StringUtils, maybe in the future in a separate jira ticket that can be removed in favor of the c++20 variant.

extensions/http-curl/tests/EmpyFlow.h
```diff
@@ -0,0 +1,28 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+constexpr const char* empty_flow = R"(
```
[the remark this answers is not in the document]
adamdebreceni: renamed

libminifi/include/core/Core.h
adamdebreceni: last time I tried to use `std::string_view::starts_with`, I recall that some of our platforms did not support it, for now added `StringUtils::startsWith` so we could bulk replace it as @lordgamez suggested

martinzink (approved):

libminifi/include/properties/Properties.h
```diff
@@ -29,16 +29,23 @@
...
+  PERSISTENT  // the changed value will be written to the source file
+};
+
 class Properties {
   struct PropertyValue {
-    std::string value;
-    bool changed;
+    std::string persisted_value;
+    std::string active_value;
+    bool changed{false};  // persisted_value should be written to disk if requested
```
fgerlits, line 48: I would rename this to something like `need_to_persist_new_value`, and then the comment is no longer needed

libminifi/src/core/logging/LoggerConfiguration.cpp
```diff
@@ -359,6 +323,57 @@ void LoggerConfiguration::initializeCompression(const std::lock_guard<std::mutex
...
+    } catch (const std::invalid_argument &) {
+    } catch (const std::out_of_range &) {
+    }
+  }
+
+  std::lock_guard<std::mutex> guard(sink_map_mtx);
+  auto it = rotating_file_sinks.find(file_name);
+  if (it != rotating_file_sinks.end()) {
+    return it->second;
+  }
```
fgerlits, line 371: What is the reason for this new caching?  It looks a bit weird that we do it for the rolling appender but not for the other appender types.  Is it possible that some update will not get applied because of a cached old value?

libminifi/include/properties/Properties.h
adamdebreceni: renamed

libminifi/src/core/logging/LoggerConfiguration.cpp
adamdebreceni: added comment: in the spdlog implementation I found no guarantee that multiple sinks to the same directory won't interfere with each other (it might be guaranteed at the OS level though), also found this https://github.com/gabime/spdlog/discussions/2151

adamdebreceni: this also might be relevant: https://github.com/gabime/spdlog/issues/243

> If you want multiple loggers to share the same file, they must share the same sink. Otherwise bad things will happen as you have noticed.

adamdebreceni: and this indeed means that some updates will not take effect until restart, e.g. changing the rotating file count/size

fgerlits: Thanks, makes sense.  But I don't think the new comment is clear; something like
```
// According to spdlog docs, if two loggers write to the same file, they must use the same sink object.
// Note that some logging configuration changes will not take effect until MiNiFi is restarted.
```
would be better.

adamdebreceni: changed comment

fgerlits (approved):

PR #1255, opened 2022-02-07
MINIFICPP-1750 Fix C2 clear operation with corecomponentstate operand

lordgamez:
Previously the `clear corecomponentstate` operation was using the operand `corecomponentstate` as the query name of the component to clear instead of the arguments where the actual component names are passed. The component names are now extracted from the argument pairs (corecomponent1=<component name 1>, corecomponent2=<component name 2>) where all component names are queried and cleared.

https://issues.apache.org/jira/browse/MINIFICPP-1750

--------------------------------------------------------------
Thank you for submitting a contribution to Apache NiFi - MiNiFi C++.

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [ ] Is there a JIRA ticket associated with this PR? Is it referenced
     in the commit message?

- [ ] Does your PR title start with MINIFICPP-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.

- [ ] Has your PR been rebased against the latest commit within the target branch (typically main)?

- [ ] Is your initial contribution a single, squashed commit?

### For code changes:
- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?
- [ ] If applicable, have you updated the LICENSE file?
- [ ] If applicable, have you updated the NOTICE file?

### For documentation related changes:
- [ ] Have you ensured that format looks appropriate for the output in which it is rendered?

### Note:
Please ensure that once the PR is submitted, you check GitHub Actions CI results for build issues and submit an update to your PR as soon as possible.

PR #1261, opened 2022-02-09
MINIFICPP-1752 Remove the catch2 dependency from our the test infrast…

martinzink:
…ructure

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/encrypt-config/tests/ConfigFileEncryptorTests.cpp
+++ b/encrypt-config/tests/ConfigFileEncryptorTests.cpp
@@ -23,6 +23,7 @@
 #include "ConfigFileEncryptor.h"
 
 #include "TestBase.h"
+#include "Catch.h"
 
 using org::apache::nifi::minifi::encrypt_config::ConfigFile;
 using org::apache::nifi::minifi::encrypt_config::encryptSensitivePropertiesInFile;
--- a/encrypt-config/tests/ConfigFileTests.cpp
+++ b/encrypt-config/tests/ConfigFileTests.cpp
@@ -25,6 +25,7 @@
 #include "utils/gsl.h"
 
 #include "TestBase.h"
+#include "Catch.h"
 #include "utils/file/FileUtils.h"
 
 namespace org {
--- a/extensions/civetweb/tests/ListenHTTPTests.cpp
+++ b/extensions/civetweb/tests/ListenHTTPTests.cpp
@@ -23,6 +23,7 @@
 #include <iostream>
 
 #include "TestBase.h"
+#include "Catch.h"
 
 #include "utils/file/FileUtils.h"
 #include "processors/GetFile.h"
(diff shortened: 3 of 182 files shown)


lordgamez (commented):
[In another PR](https://github.com/apache/nifi-minifi-cpp/pull/1255) I had to change `libminifi/include/utils/TestUtils.h` because I found that it also has a dependency to catch2 with a single `REQUIRE` call, maybe that could also be removed as part of this PR.

martinzink:
> [In another PR](https://github.com/apache/nifi-minifi-cpp/pull/1255) I had to change `libminifi/include/utils/TestUtils.h` because I found that it also has a dependency to catch2 with a single `REQUIRE` call, maybe that could also be removed as part of this PR.

Good idea, https://github.com/apache/nifi-minifi-cpp/pull/1261/commits/be3f276dc4005e6c3218486f01c9dd126e84bee6

szaszm:
The windows job failure seems to be related

szaszm:
And another question: does this change benefit the codebase aside from the upcoming GCS extension? E.g. do we save compilation time?
If not, maybe it would be better for GCS to not rely on these utilities, or only depend on a limited set of them. I don't expect many too many dependencies in the future that mandate their own test framework.

martinzink:
> And another question: does this change benefit the codebase aside from the upcoming GCS extension? E.g. do we save compilation time? If not, maybe it would be better for GCS to not rely on these utilities, or only depend on a limited set of them, and keep the extra complexity of dealing with Google's BS inside the extension. I don't expect many too many dependencies in the future that mandate their own test framework.

I dont think that there is measurable difference if any. But I do think this improves the codebase because it removes an unneccessary direct dependency.
The utilities in TestBase.h and SingleInputTestController.h dont rely on Catch2 so this change is really an _include what you use_ refactor, because I think our catch tests should include the catch headers and not get them indirectly from TestBase.h

Also without TestBase.h I think the quality of the GCS tests would greatly decrease (or bring in some code duplication)
Besides the GCS extension we could also use gtest to create [death tests](https://google.github.io/googletest/advanced.html#death-tests) (catch2 doesnt support those yet iirc)

Overall I might agree that the upside is not that great, but I dont really see a downside to this change either.

fgerlits (approved):

lordgamez (approved):

adamdebreceni (approved):

szaszm (approved):

szaszm:
During a pre-merge check, I got this compilation error:
```
/home/szaszm/nifi-minifi-cpp-3/extensions/expression-language/tests/ProcessContextExprTests.cpp:54:10: error: expected constructor, destructor, or type conversion before ‘(’ token
   54 | TEST_CASE("ProcessContextExpr can update existing processor properties", "[setProperty][getProperty]") {
      |          ^
make[2]: *** [extensions/expression-language/tests/CMakeFiles/ProcessContextExprTests.dir/build.make:76: extensions/expression-language/tests/CMakeFiles/ProcessContextExprTests.dir/ProcessContextExprTests.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:9470: extensions/expression-language/tests/CMakeFiles/ProcessContextExprTests.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```
Could you fix it? We just need another Catch.h include in ProcessContextExprTests.cpp, but I don't want to change files in a merge commit.

PR #1270, opened 2022-02-24
MINIFICPP-1734 Add integration test for Kubernetes log collection

fgerlits:
https://issues.apache.org/jira/browse/MINIFICPP-1734

Integration tests for PR #1257.

---

### For all changes:
     in the commit message?

### For code changes:

### Note:


docker/test/integration/minifi/core/DockerTestCluster.py
```diff
@@ -34,6 +34,15 @@ def get_stdout_encoding():
         return encoding
 
     def get_app_log(self, container_name):
+        type = self.containers[container_name].type()
+        if type == 'docker container':
+            return self.__get_app_log_from_docker_container(container_name)
+        elif type == 'direct':
+            return self.containers[container_name].get_app_log()
+        else:
+            raise Exception("Unexpected container type '%s'" % type)
```
lordgamez, line 43: I would either change the method name from `type()` to `log_source()` or the type `direct` to `kind cluster container` or something similar to be more clearer, because `direct` doesn't really name a type. Also maybe a class could be added with const members for these literals to be used from a single place like

```
class ContainerType:
  DOCKER_CONTAINER = "docker container"
  KIND_CLUSTER_CONTAINER = "kind cluster container"
```

docker/test/integration/minifi/core/ImageStore.py
```diff
@@ -84,6 +87,18 @@ def __build_minifi_cpp_image(self):
 
         return self.__build_image(dockerfile)
 
+    def __build_simple_minifi_cpp_image_with_root(self):
```
lordgamez, line 90: Because of the specific tag, this may need to be renamed to `__build_minifi_cpp_image_for_kind_proxy_with_root` or something similar to contain the intent of use.

docker/test/integration/features/kubernetes.feature
```diff
@@ -0,0 +1,43 @@
+Feature: TailFile can collect logs from Kubernetes pods
```
adamdebreceni, line 1: I see that we do not really add the apache license text in files in the docker directory, could this have unintended consequences?

docker/test/integration/minifi/core/KindProxy.py
```diff
@@ -0,0 +1,112 @@
...
+        for namespace in namespaces:
+            self.__wait_for_default_service_account(namespace)
+
+        self.__create_objects_of_type(resources_directory, 'pod')
+        self.__create_objects_of_type(resources_directory, 'clusterrole')
+        self.__create_objects_of_type(resources_directory, 'clusterrolebinding')
+
+    def __wait_for_default_service_account(self, namespace):
+        for _ in range(120):
+            if subprocess.run(['docker', 'exec', 'kind-control-plane', 'kubectl', '-n', namespace, 'get', 'serviceaccount', 'default']).returncode == 0:
```
lordgamez, line 87: It may be better to instantiate a python docker client to get the container and call `exec_run` on it to execute commands inside the container.

docker/test/integration/minifi/core/KindProxy.py
```diff
@@ -0,0 +1,112 @@
...
+            config_file.write(kind_config.encode('utf-8'))
+
+    def start_cluster(self):
+        subprocess.run([self.kind_binary_path, 'delete', 'cluster'])
+
+        if subprocess.run([self.kind_binary_path, 'create', 'cluster', '--config=' + self.kind_config_path]).returncode != 0:
+            raise Exception("Could not start the kind cluster")
+
+    def load_docker_image(self, image_store):
+        image_store.get_image('minifi-cpp-in-kubernetes')
```
lordgamez, line 67: I think the image name should be a parameter, we could load other images as well if we want to test them with minifi in kubernetes later

docker/test/integration/minifi/core/KubernetesRunner.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+import logging
+import os
+import shutil
+from .KindProxy import KindProxy
+from .MinifiContainer import MinifiContainer
+
+
+class KubernetesRunner(MinifiContainer):
```
lordgamez, line 24: I think this should be renamed to reflect that this runs a minifi container inside a kubernetes environment, KubernetesRunner seems too generic opposed to the function of the class.

adam-markovics (approved):

docker/test/integration/features/kubernetes.feature
fgerlits: I'm not sure.  I have added a license header to new files containing code, but not to new files containing test data.  I think these feature files are test data.

Do you think we need to add a header to all files, or just that feature files are code and not test data, so should have a header?  There is no technical problem with adding headers to feature files; lines starting with # are ignored.

docker/test/integration/minifi/core/DockerTestCluster.py
fgerlits: Part of the problem is that we have our own `minifi.core.Container` (and derived classes), and the docker client's `docker.models.Container`, and we call both "container".

The difference here is whether we get the logs from the `docker.models.Container` contained in the `minifi.core.Container`, or directly from the `minifi.core.Container`, using its `get_app_log()` method.  This could be used more generally than Kubernetes or Kind, so "kind cluster container" is too specific.

docker/test/integration/minifi/core/KindProxy.py
fgerlits: Do you know how to redirect a file to the stdin of `exec_run`?  This is needed in `__create_objects_of_type()`.

docker/test/integration/minifi/core/KindProxy.py
fgerlits: Do you mean `'minifi-cpp-kubernetes'` should be (the value of) a parameter of the `load_docker_image()` function?  Or of the `KindProxy` constructor?  I don't think either of those would be useful right now, and we could make the change later when needed.

szaszm (commented):
I didn't see the open threads when pressing "Approve". I don't have any comments, but please wait until the open comments are resolved before merging.

docker/test/integration/minifi/core/ImageStore.py
fgerlits: I have moved the tagging to KindProxy in https://github.com/apache/nifi-minifi-cpp/pull/1270/commits/34bbb4f50d6dda4ced4a0a105611f4d6a99d9089, so now this function does what its name suggests.

docker/test/integration/minifi/core/KubernetesRunner.py
fgerlits: I have renamed the class to `MinifiAsPodInKubernetesCluster` in https://github.com/apache/nifi-minifi-cpp/pull/1270/commits/26d204d8759b5f9b5a61e9157d4bffbac217d9d1.

docker/test/integration/minifi/core/DockerTestCluster.py
fgerlits: I have renamed `type()` to `log_source()` and created an enum-like class for the values as you suggested (https://github.com/apache/nifi-minifi-cpp/pull/1270/commits/4f416b5c2a955c7ed6e7af6aa7dbdb911e4ecccf).

docker/test/integration/minifi/core/KindProxy.py
lordgamez: Unfortunately I didn't find any similar functionality in `exec_run`'s documentation, I would suggest using docker copy to copy that file inside the container, but if you would like to keep it as it is due to this I'm okay with that as well.

docker/test/integration/minifi/core/KindProxy.py
fgerlits: I have moved minifi-specific code from `KindProxy` to `MinifiAsPodInKubernetesCluster` in https://github.com/apache/nifi-minifi-cpp/pull/1270/commits/5ccee646c882a3ac9d97acc2655617a7824f0667

docker/test/integration/minifi/core/KindProxy.py
fgerlits: I have changed `subprocess.run()` to `docker_client.exec_run()` where I could in https://github.com/apache/nifi-minifi-cpp/pull/1270/commits/57741586adabccb1706313c2f353e20c769270f7.

Unfortunately, there is no API for `docker cp` in the Python client library.  I didn't like the recommended workaround of "create a tar archive, use `put_archive()` to copy it to the container, then use `exec_run()` to untar it inside", so I left that as `subprocess.run()`.

docker/test/integration/features/kubernetes.feature
fgerlits: I have added a license header to all new or modified `.py`/`.feature` files in https://github.com/apache/nifi-minifi-cpp/pull/1270/commits/b912ce6ca9ae4728c3db8dae7fc89814725d011f, and created a Jira to update the rest: [MINIFICPP-1764](https://issues.apache.org/jira/browse/MINIFICPP-1764)

docker/test/integration/minifi/core/KindProxy.py
lordgamez: That workaround sounds way too complicated to use, this is a better solution at the moment, thanks.

docker/test/integration/features/kubernetes.feature
adamdebreceni: thanks!

adamdebreceni (approved):

lordgamez:
It might be worth investigating how to restructure the test setups to initialize and delete the kind cluster on feature level instead of test scenario level. It takes a lot of time (about one minute as I experienced) to set the cluster up and with the increase of the number of test cases it would be cumbersome to do this for every test scenario. It may be better to set up a single kind cluster on feature level, do a cleanup after every scenario, and only delete the cluster after all the test scenarios in the feature are finished. It doesn't have to be a part of this PR, but we should create a Jira ticket to investigate this possibility in the future.

lordgamez (approved):

fgerlits:
> It might be worth investigating how to restructure the test setups to initialize and delete the kind cluster on feature level instead of test scenario level. It takes a lot of time (about one minute as I experienced) to set the cluster up and with the increase of the number of test cases it would be cumbersome to do this for every test scenario. It may be better to set up a single kind cluster on feature level, do a cleanup after every scenario, and only delete the cluster after all the test scenarios in the feature are finished. It doesn't have to be a part of this PR, but we should create a Jira ticket to investigate this possibility in the future.

https://issues.apache.org/jira/browse/MINIFICPP-1765

martinzink (approved):

PR #1271, opened 2022-03-01
MINIFICPP-1763 - Move extension inclusion logic into the extensions

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:


CONTRIB.md
```diff
@@ -103,22 +103,29 @@ are contributing a custom Processor or Controller Service, the mechanism to regi
...
+if (NOT ENABLE_GPS)
+  return()
+endif()
+
+#
+# extension definition goes here
+#
+
+# at the end we should announce our extension
+register_extension(minifi-gps "GPS EXTENSIONS" GPS-EXTENSION "Enables LibGPS Functionality and the GetGPS processor." "${TEST_DIR}/gps-tests")
```
fgerlits, line 123: It would be better to pick an example where the extension name and the extension guard are the same, as I think that is better practice.

extensions/systemd/CMakeLists.txt
```diff
@@ -17,12 +17,19 @@
 # under the License.
 #
 
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+  option(ENABLE_SYSTEMD "Enables the systemd extension." ON)
+endif()
```
fgerlits, line 22: I would keep this part in the root `CMakeLists.txt` file, together with the other `option(...)` commands

libminifi/include/core/ConfigurableComponent.h
```diff
@@ -56,6 +56,9 @@ class ConfigurableComponent {
   ConfigurableComponent& operator=(const ConfigurableComponent &other) = delete;
   ConfigurableComponent& operator=(ConfigurableComponent &&other) = delete;
 
+  template<typename T = std::string, typename = std::enable_if_t<std::is_default_constructible_v<T>>>
+  std::optional<T> getProperty(const Property& property) const;
```
fgerlits, line 60: why is this in this PR?

CONTRIB.md
adamdebreceni: do you mean the `EXTENSION` vs `EXTENSIONS`?

extensions/systemd/CMakeLists.txt
adamdebreceni: agreed, at first I wanted to move (and did move) everything including `option(...)`s but then moved them back, this one was duplicated here by accident, removed it

libminifi/include/core/ConfigurableComponent.h
adamdebreceni: I tried to avoid opening a PR for these lines, should I remove this?

CONTRIB.md
fgerlits: yes

libminifi/include/core/ConfigurableComponent.h
fgerlits: what is it for?
you could add a commit with an explanation of this change, so whoever will merge the PR can include it in the squashed commit message

adamdebreceni: added commit message

CONTRIB.md
adamdebreceni: changed it to kubernetes extension

fgerlits (approved):

extensions/http-curl/CMakeLists.txt
```diff
@@ -17,6 +17,12 @@
 # under the License.
 #
 
+if (NOT ((DISABLE_CURL STREQUAL "OFF" OR NOT DISABLE_CURL) AND NOT DISABLE_CIVET))
```
szaszm, line 20:
```suggestion
if (NOT ((NOT DISABLE_CURL OR NOT DISABLE_CURL) AND NOT DISABLE_CIVET))
```

This should be about the same, expect for accepting "no" and 0 as well.

libminifi/include/core/ConfigurableComponent.h
szaszm: Please use `requires` instead of `enable_if`. We can't yet use standard library concepts (like [`std::default_initializable`](https://en.cppreference.com/w/cpp/concepts/default_initializable)) because of xcode, but we can use requires declarations with type traits.

And I think T should be movable, too, so that we can move the value to the result object.

adamdebreceni: removed it

extensions/http-curl/CMakeLists.txt
adamdebreceni: changed it

szaszm (approved):

martinzink (approved):

lordgamez (approved):

PR #1280, opened 2022-03-09
MINIFICPP-1770 Add C2 NO_OPERATION response if property update is NoOp

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-1770

------------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/http-curl/tests/HTTPHandlers.h
+++ b/extensions/http-curl/tests/HTTPHandlers.h
@@ -800,11 +800,7 @@ class C2AcknowledgeHandler : public ServerAwareHandler {
       if (root["operationState"].IsObject() && root["operationState"].HasMember("state")) {
         std::lock_guard<std::mutex> guard(apply_count_mtx_);
         auto result_state = root["operationState"]["state"].GetString();
-        if (apply_count_.find(result_state) != apply_count_.end()) {
-          ++apply_count_[result_state];
-        } else {
-          apply_count_[result_state] = 1;
-        }
+        ++apply_count_[result_state];
       }
     }
 
--- a/extensions/http-curl/tests/HTTPHandlers.h
+++ b/extensions/http-curl/tests/HTTPHandlers.h
@@ -792,20 +792,40 @@ class C2AcknowledgeHandler : public ServerAwareHandler {
     rapidjson::Document root;
     root.Parse(req.data(), req.size());
     if (root.IsObject() && root.HasMember("operationId")) {
-      std::lock_guard<std::mutex> guard(mtx_);
+      std::lock_guard<std::mutex> guard(ack_operations_mtx_);
       acknowledged_operations_.insert(root["operationId"].GetString());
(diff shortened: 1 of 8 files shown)


libminifi/src/Configure.cpp
```diff
@@ -30,6 +30,14 @@ namespace nifi {
 namespace minifi {
 
 bool Configure::get(const std::string& key, std::string& value) const {
+  const std::string_view log_prefix = "nifi.log.";
```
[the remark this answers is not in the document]
lordgamez: Good idea, updated in 50839c9804f63ec6a588c765821dce6b170b3af0

szaszm (approved):

extensions/http-curl/tests/HTTPHandlers.h
```diff
@@ -792,20 +792,40 @@ class C2AcknowledgeHandler : public ServerAwareHandler {
     rapidjson::Document root;
     root.Parse(req.data(), req.size());
     if (root.IsObject() && root.HasMember("operationId")) {
-      std::lock_guard<std::mutex> guard(mtx_);
+      std::lock_guard<std::mutex> guard(ack_operations_mtx_);
```
adam-markovics, line 795: With locking two mutexes, you always have to be careful. Is it possible to have another thread locking them in the reverse order than here? So, first `apply_count_mtx_` and then `ack_operations_mtx_`? That could result in a deadlock. It's possible by calling `isAcknowledged` first and `getApplyCount` second, for example.

szaszm: `ack_operations_mtx_` is released before locking `apply_count_mtx_`, so they are never locked at the same time. Since only one is held at a time, it's not possible that this function and another thread are waiting for each other.

adam-markovics (approved):

extensions/http-curl/tests/HTTPHandlers.h
```diff
@@ -792,20 +792,40 @@ class C2AcknowledgeHandler : public ServerAwareHandler {
...
       acknowledged_operations_.insert(root["operationId"].GetString());
     }
+
+    if (root.IsObject() && root.HasMember("operationState")) {
+      if (root["operationState"].IsObject() && root["operationState"].HasMember("state")) {
+        std::lock_guard<std::mutex> guard(apply_count_mtx_);
+        auto result_state = root["operationState"]["state"].GetString();
+        if (apply_count_.find(result_state) != apply_count_.end()) {
```
adamdebreceni, line 803: note: `++apply_count[result_state]` would achieve the same result

lordgamez: Good call! Updated in 4198ad6599cd8ae739342b5376347c60329d53b6

adamdebreceni (approved):

PR #1281, opened 2022-03-09
MINIFICPP-1769 Add propertyValue to agent manifest metadata

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-1769

-----------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/include/properties/ConfigurationReader.h
```diff
@@ -0,0 +1,32 @@
...
+#include <string>
+
+namespace org::apache::nifi::minifi {
+
+class ConfigurationReader {
+ public:
+  virtual bool get(const std::string& key, std::string& value) const = 0;
+  virtual bool get(const std::string& key, const std::string& alternate_key, std::string& value) const = 0;
+  virtual std::optional<std::string> get(const std::string& key) const = 0;
+};
```
szaszm, line 30: I think `std::optional<std::string> get(std::string_view key)` would cover all use cases. Alternate keys could be handled with `utils::orElse`. Consider removing this interface and replacing it with a `std::function<std::optional<std::string>(std::string_view)>` or templates with a similar callback signature, in the name of simplicity.

If you decide to keep it like this, please add a virtual destructor.

lordgamez: I decided to keep the interface for now. It's more consistent with the other setters of the agent information nodes and in my opinion it's cleaner to see what that member represents.

szaszm: I mean all I'm seeing in AgentInformation is setters for random objects, and I don't think it makes a difference if that object is a custom class versus a `std::function`, a setter is a setter.

lordgamez: I was referring to classes like the `AgentIdentificationProvider` or the `StateMonitor` where those classes are also implemented as separate interfaces to be later set in one of the agent information nodes. I think these adhere more to the interface segregation principle and especially in case of the FlowController where there are already way too many responsibilities they could be later extracted to separate classes and composed instead of inheriting from them.

**BUT** I rechecked this case specifically and as you mentioned we only need a single overload of the get functions and I'm not sure if this "read-only configuration" interface will be reused in the future anywhere else, so after these considerations I replaced it with a single std::function in c2544df895ed1fac5ba6d5b50e27ecea4e073c3e

szaszm: Great, thanks! One more little nitpick: is there a reason for using `const std::string&` over `std::string_view`? Unless you're working with only `std::string` objects, `std::string_view` can eliminate the need for temporary allocations when passing C strings.

lordgamez: The reason was the signature of the `get` function of the `Configure` class which expects `const std::string& key` as its parameter.

szaszm (approved):

extensions/http-curl/tests/C2VerifyHeartbeatAndStop.cpp
```diff
@@ -55,7 +40,7 @@ class VerifyC2Heartbeat : public VerifyC2Base {
 
   void runAssertions() override {
     using org::apache::nifi::minifi::utils::verifyLogLinePresenceInPollTime;
-    assert(verifyLogLinePresenceInPollTime(std::chrono::milliseconds(wait_time_),
+    assert(verifyLogLinePresenceInPollTime(std::chrono::milliseconds(120000),
```
adamdebreceni, line 43: what is the reason behind this hardcoded wait time?

lordgamez: Thanks for catching this, I only set this while testing something and I forgot about it. Fixed in a064ba2a3bab0dbe4ded393a4d52dd5315c3f976

adamdebreceni (approved):

fgerlits (approved):

extensions/http-curl/tests/C2ClearCoreComponentStateTest.cpp
```diff
@@ -65,7 +65,9 @@ class VerifyC2ClearCoreComponentState : public VerifyC2Base {
 
 class ClearCoreComponentStateHandler: public HeartbeatHandler {
  public:
-  explicit ClearCoreComponentStateHandler(std::atomic_bool& component_cleared_successfully) : component_cleared_successfully_(component_cleared_successfully) {
+  explicit ClearCoreComponentStateHandler(std::atomic_bool& component_cleared_successfully, std::shared_ptr<minifi::Configure> configuration)
+    : HeartbeatHandler(std::move(configuration)),
+      component_cleared_successfully_(component_cleared_successfully) {
```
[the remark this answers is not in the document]
lordgamez: We are asserting on the variable while waiting until the timeout for its value to return true on line 49.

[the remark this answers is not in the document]
adam-markovics: Oh, I see. I just opened my IDE which said it was unused and I believed it.

adam-markovics (approved):

PR #1284, opened 2022-03-17
MINIFICPP-1776 - Parallel heartbeat and flow update can cause crash

adam-markovics:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/http-curl/tests/C2ClearCoreComponentStateTest.cpp
+++ b/extensions/http-curl/tests/C2ClearCoreComponentStateTest.cpp
@@ -51,10 +51,21 @@ class VerifyC2ClearCoreComponentState : public VerifyC2Base {
 
  protected:
   void updateProperties(minifi::FlowController& flow_controller) override {
-    dynamic_cast<minifi::state::ProcessorController*>(flow_controller.getComponents("TailFile1")[0])
-        ->getProcessor()->setProperty(minifi::processors::TailFile::FileName, test_file_1_);
-    dynamic_cast<minifi::state::ProcessorController*>(flow_controller.getComponents("TailFile2")[0])
-        ->getProcessor()->setProperty(minifi::processors::TailFile::FileName, test_file_2_);
+    size_t componentIdx = 0;
+    auto setFileName = [&componentIdx] (const std::string& fileName, minifi::state::StateController* component){
+      if (componentIdx == 0) {
+        auto* processor = dynamic_cast<minifi::state::ProcessorController*>(component)->getProcessor();
+        processor->setProperty(minifi::processors::TailFile::FileName, fileName);
+      }
+      ++componentIdx;
+    };
+
+    flow_controller.executeOnComponents("TailFile1",
+      [&](minifi::state::StateController* component) {setFileName(test_file_1_, component);});
+    // reset index counter
+    componentIdx = 0;
+    flow_controller.executeOnComponents("TailFile2",
+      [&](minifi::state::StateController* component) {setFileName(test_file_2_, component);});
   }
 
   TestController testController;
--- a/extensions/http-curl/tests/C2ClearCoreComponentStateTest.cpp
+++ b/extensions/http-curl/tests/C2ClearCoreComponentStateTest.cpp
@@ -51,20 +51,14 @@ class VerifyC2ClearCoreComponentState : public VerifyC2Base {
 
  protected:
   void updateProperties(minifi::FlowController& flow_controller) override {
-    size_t componentIdx = 0;
-    auto setFileName = [&componentIdx] (const std::string& fileName, minifi::state::StateController* component){
-      if (componentIdx == 0) {
-        auto* processor = dynamic_cast<minifi::state::ProcessorController*>(component)->getProcessor();
-        processor->setProperty(minifi::processors::TailFile::FileName, fileName);
-      }
-      ++componentIdx;
+    auto setFileName = [] (const std::string& fileName, minifi::state::StateController* component){
+      auto* processor = dynamic_cast<minifi::state::ProcessorController*>(component)->getProcessor();
+      processor->setProperty(minifi::processors::TailFile::FileName, fileName);
     };
 
-    flow_controller.executeOnComponents("TailFile1",
+    flow_controller.executeOnComponent("TailFile1",
       [&](minifi::state::StateController* component) {setFileName(test_file_1_, component);});
-    // reset index counter
-    componentIdx = 0;
-    flow_controller.executeOnComponents("TailFile2",
+    flow_controller.executeOnComponent("TailFile2",
       [&](minifi::state::StateController* component) {setFileName(test_file_2_, component);});
   }
 
--- a/extensions/http-curl/tests/C2ClearCoreComponentStateTest.cpp
+++ b/extensions/http-curl/tests/C2ClearCoreComponentStateTest.cpp
@@ -51,15 +51,15 @@ class VerifyC2ClearCoreComponentState : public VerifyC2Base {
 
  protected:
   void updateProperties(minifi::FlowController& flow_controller) override {
-    auto setFileName = [] (const std::string& fileName, minifi::state::StateController* component){
-      auto* processor = dynamic_cast<minifi::state::ProcessorController*>(component)->getProcessor();
+    auto setFileName = [] (const std::string& fileName, minifi::state::StateController& component){
+      auto* processor = dynamic_cast<minifi::state::ProcessorController&>(component).getProcessor();
       processor->setProperty(minifi::processors::TailFile::FileName, fileName);
     };
 
     flow_controller.executeOnComponent("TailFile1",
-      [&](minifi::state::StateController* component) {setFileName(test_file_1_, component);});
+      [&](minifi::state::StateController& component) {setFileName(test_file_1_, component);});
     flow_controller.executeOnComponent("TailFile2",
-      [&](minifi::state::StateController* component) {setFileName(test_file_2_, component);});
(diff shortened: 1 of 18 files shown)


extensions/http-curl/tests/C2ClearCoreComponentStateTest.cpp
```diff
@@ -51,10 +51,21 @@ class VerifyC2ClearCoreComponentState : public VerifyC2Base {
 
  protected:
   void updateProperties(minifi::FlowController& flow_controller) override {
-    dynamic_cast<minifi::state::ProcessorController*>(flow_controller.getComponents("TailFile1")[0])
-        ->getProcessor()->setProperty(minifi::processors::TailFile::FileName, test_file_1_);
-    dynamic_cast<minifi::state::ProcessorController*>(flow_controller.getComponents("TailFile2")[0])
-        ->getProcessor()->setProperty(minifi::processors::TailFile::FileName, test_file_2_);
+    size_t componentIdx = 0;
+    auto setFileName = [&componentIdx] (const std::string& fileName, minifi::state::StateController* component){
```
lordgamez, line 55: Could you please elaborate on the purpose of the `componentIdx` here? What I see that during the test in both cases it is called with the value 0, why the check and the reset of the counter is needed?

adam-markovics: It's going to be removed in a future commit.

adam-markovics: Not present anymore.

lordgamez (approved):

libminifi/include/FlowController.h
```diff
@@ -234,7 +233,11 @@ class FlowController : public core::controller::ForwardingControllerServiceProvi
   std::chrono::steady_clock::time_point start_time_;
 
  private:
-  void getProcessorController(const std::string& name, std::vector<state::StateController*>& controllerVec,
+  std::vector<state::StateController*> getAllComponents();
+
+  state::StateController* getComponent(const std::string &name);
+
+  void getProcessorController(const std::string& name, state::StateController*& controller,
                               const std::function<std::unique_ptr<state::ProcessorController>(core::Processor&)>& controllerFactory);
```
fgerlits, line 241: this could return the `controller` instead of using the slightly strange-looking `state::StateController*&` out parameter:
```suggestion
  state::StateController* getProcessorController(const std::string& name,
                              const std::function<std::unique_ptr<state::ProcessorController>(core::Processor&)>& controllerFactory);
```

libminifi/src/FlowController.cpp
```diff
@@ -438,6 +440,20 @@ std::shared_ptr<state::response::ResponseNode> FlowController::getAgentManifest(
...
+  for (auto* component: getAllComponents()) {
+    func(component);
+  }
+}
+
+void FlowController::executeOnComponent(const std::string &name, std::function<void(state::StateController*)> func) {
+  std::lock_guard<std::recursive_mutex> lock(mutex_);
+  if (auto* component = getComponent(name); component != nullptr) {
+    func(component);
+  }
```
fgerlits, line 454: maybe add an error log if the component is not found?  I think that should never happen, and a log could be useful if it does

libminifi/src/FlowController.cpp
```diff
@@ -523,14 +537,14 @@ void FlowController::getAllProcessorControllers(std::vector<state::StateControll
...
 }
 
-void FlowController::getProcessorController(const std::string& name, std::vector<state::StateController*>& controllerVec,
+void FlowController::getProcessorController(const std::string& name, state::StateController*& controller,
                                           const std::function<std::unique_ptr<state::ProcessorController>(core::Processor&)>& controllerFactory) {
   auto* processor = root_->findProcessorByName(name);
-  auto& controller = processor_to_controller_[processor->getUUID()];
-  if (!controller) {
-    controller = controllerFactory(*processor);
+  auto& foundController = processor_to_controller_[processor->getUUID()];
```
fgerlits, line 543: we should check if `processor` is null before dereferencing it

libminifi/src/FlowController.cpp
```diff
@@ -523,14 +537,14 @@ void FlowController::getAllProcessorControllers(std::vector<state::StateControll
...
+  if (!foundController) {
+    foundController = controllerFactory(*processor);
   }
-  controllerVec.push_back(controller.get());
+  controller = foundController.get();
```
fgerlits, line 547: `controller` is dangling: `controllerFactory` returns a `unique_ptr` which gets destroyed at the end of the function.

I think we should add the newly created `ProcessorController` to the `processor_to_controller_` map, both here and in `getAllProcessorControllers`.

libminifi/src/c2/ControllerSocketProtocol.cpp
```diff
@@ -180,14 +178,18 @@ void ControllerSocketProtocol::initialize(core::controller::ControllerServicePro
...
+              components.emplace_back(component->getComponentName(), component->isRunning());
+            });
             io::BufferStream resp;
             resp.write(&head, 1);
-            const auto size_ = gsl::narrow<uint16_t>(update_sink_->getAllComponents().size());
-            resp.write(size_);
-            for (const auto &component : update_sink_->getAllComponents()) {
-              resp.write(component->getComponentName());
-              resp.write(component->isRunning() ? "true" : "false");
+            resp.write(gsl::narrow<uint16_t>(gsl::narrow<uint16_t>(components.size())));
```
fgerlits, line 187: one `gsl_narrow` is enough :)
```suggestion
            resp.write(gsl::narrow<uint16_t>(components.size()));
```

libminifi/src/FlowController.cpp
adam-markovics: It's not dangling, because `foundController` is a reference to the found value in the map. The created controller is going to be stored there.

libminifi/src/FlowController.cpp
adam-markovics: Done.

libminifi/src/FlowController.cpp
adam-markovics: Done.

szaszm: You can even leave out the second part, and it means the same.
```
if (auto* const component = getComponent(name)) {
  func(component);
}
```

libminifi/include/FlowController.h
adam-markovics: Done.

libminifi/src/FlowController.cpp
fgerlits: Yes, you're right, sorry.

libminifi/src/FlowController.cpp
```diff
@@ -506,31 +513,36 @@ std::map<std::string, std::unique_ptr<io::InputStream>> FlowController::getDebug
...
-void FlowController::getProcessorController(const std::string& name, std::vector<state::StateController*>& controllerVec,
-                                          const std::function<std::unique_ptr<state::ProcessorController>(core::Processor&)>& controllerFactory) {
+state::StateController* FlowController::getProcessorController(const std::string& name, const std::function<std::unique_ptr<state::ProcessorController>(core::Processor&)>& controllerFactory) {
   auto* processor = root_->findProcessorByName(name);
-  auto& controller = processor_to_controller_[processor->getUUID()];
-  if (!controller) {
-    controller = controllerFactory(*processor);
+  if (processor == nullptr) {
+    logger_->log_error("Could not get processor controller for requested name \"%s\", because processor was not found either", name);
+  }
```
fgerlits, line 538: logging and then crashing is an improvement, but not crashing is even better:
```suggestion
  if (processor == nullptr) {
    logger_->log_error("Could not get processor controller for requested name \"%s\", because processor was not found either", name);
    return nullptr;
  }
```

libminifi/src/FlowController.cpp
```diff
@@ -438,30 +432,43 @@ std::shared_ptr<state::response::ResponseNode> FlowController::getAgentManifest(
...
+  std::lock_guard<std::recursive_mutex> lock(mutex_);
+  for (auto* component: getAllComponents()) {
+    func(component);
+  }
+}
+
+void FlowController::executeOnComponent(const std::string &name, std::function<void(state::StateController*)> func) {
+  std::lock_guard<std::recursive_mutex> lock(mutex_);
+  if (auto* component = getComponent(name); component != nullptr) {
+    func(component);
```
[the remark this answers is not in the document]
adam-markovics: This will result in a large change in all usages, but I'll do it.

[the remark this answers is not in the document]
szaszm: thanks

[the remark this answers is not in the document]
adam-markovics: Done.

libminifi/src/FlowController.cpp
```diff
@@ -438,30 +432,43 @@ std::shared_ptr<state::response::ResponseNode> FlowController::getAgentManifest(
   return agentInfo;
 }
 
+void FlowController::executeOnAllComponents(std::function<void(state::StateController*)> func) {
+  std::lock_guard<std::recursive_mutex> lock(mutex_);
+  for (auto* component: getAllComponents()) {
+    func(component);
+  }
```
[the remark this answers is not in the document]
adam-markovics: Done.

libminifi/src/FlowController.cpp
adam-markovics: Done.

fgerlits (approved):

szaszm (approved):

PR #1286, opened 2022-03-24
MINIFICPP-1782 Upgrade and verify interworking with latest NiFi version

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-1782

----------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/docker/test/integration/minifi/core/ImageStore.py
+++ b/docker/test/integration/minifi/core/ImageStore.py
@@ -14,7 +14,6 @@
 # limitations under the License.
 
 
-from .NifiContainer import NifiContainer
 from .MinifiContainer import MinifiContainer
 import logging
 import tarfile
@@ -49,8 +48,6 @@ class ImageStore:
             image = self.__build_minifi_cpp_image_with_provenance_repo()
         elif container_engine == "http-proxy":
             image = self.__build_http_proxy_image()
-        elif container_engine == "nifi":
-            image = self.__build_nifi_image()
         elif container_engine == "postgresql-server":
             image = self.__build_postgresql_server_image()
         elif container_engine == "kafka-broker":
@@ -132,19 +129,6 @@ class ImageStore:
                     echo 'http_port {proxy_port}' >> /etc/squid/squid.conf
                 ENTRYPOINT ["/sbin/entrypoint.sh"]
                 """.format(base_image='sameersbn/squid:3.5.27-2', proxy_username='admin', proxy_password='test101', proxy_port='3128'))
-
-        return self.__build_image(dockerfile)
-
-    def __build_nifi_image(self):
-        dockerfile = dedent(r"""\
-                FROM {base_image}
-                USER root
-                RUN sed -i -e 's/^\(nifi.remote.input.host\)=.*/\1={name}/' {nifi_root}/conf/nifi.properties
-                RUN sed -i -e 's/^\(nifi.remote.input.socket.port\)=.*/\1=5000/' {nifi_root}/conf/nifi.properties
-                USER nifi
-                """.format(name='nifi',
-                           base_image='apache/nifi:' + NifiContainer.NIFI_VERSION,
-                           nifi_root=NifiContainer.NIFI_ROOT))
 
         return self.__build_image(dockerfile)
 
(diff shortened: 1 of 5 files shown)


docker/test/integration/minifi/core/NifiContainer.py
```diff
@@ -54,11 +55,12 @@ def deploy(self):
         logging.info('Creating and running nifi docker container...')
         self.__create_config()
         self.client.containers.run(
-            self.image_store.get_image(self.get_engine()),
+            "apache/nifi:" + NifiContainer.NIFI_VERSION,
```
adamdebreceni, line 58: was this change left in by accident? all other containers seem to use the former version

docker/test/integration/minifi/core/NifiContainer.py
```diff
@@ -23,12 +23,13 @@
 
 
 class NifiContainer(FlowContainer):
-    NIFI_VERSION = '1.7.0'
+    NIFI_VERSION = '1.15.3'
```
adamdebreceni, line 26: is this the latest version? should we offer some level of backwards compatibility?

docker/test/integration/minifi/core/NifiContainer.py
lordgamez: This change was on purpose as the newest image declares the config directory as `VOLUME` so it cannot be changed in build only in runtime which is done that way now. Because of this there is no need to do anything in build time we can use the docker image as it is. Not all containers need a changed base image, for example the Azure image is done this way too.

docker/test/integration/minifi/core/NifiContainer.py
lordgamez: Yes this is the latest version currently. We have backwards compatibility as it is working with the older image as well, the changes done only serve the requirements of the newer image, so it should work with the older version as well. Or did you think about something else regarding backwards compatibility?

adamdebreceni: I'm not sure what the lowest version of nifi we aim to support, but IMO there should be some tests that use that one

lordgamez: Well the previous version we used was 4 years old, and if we work with that, then I don't think we have a lot to worry about :) I think these site to site and invokehttp use cases we test did not really change since then. If they change with a newer version later then I think we can introduce a test with an older version to verify both.

szaszm: Let's aim for the latest version and avoid breaking interop unless there is a good reason.

szaszm: I would wait for 1.16.0 to be out and update to that

lordgamez: Updated to 1.16.0 in bd25b5058b3cae4a727e2cacb914bd10266e1b83

szaszm: It's not out yet. I assume someone discovered issues in the background and we're going to see another RC.

lordgamez: It's strange then that the apache/nifi:1.16.0 image was already released on dockerhub 3 days ago. Anyway we can wait for the official release for the merge, if the image changes it will be pulled again automatically due to the hash change.

szaszm: fair, that's good enough for me

szaszm (approved):

adamdebreceni (approved):

docker/test/integration/minifi/core/ImageStore.py
```diff
@@ -135,19 +132,6 @@ def __build_http_proxy_image(self):
 
         return self.__build_image(dockerfile)
 
-    def __build_nifi_image(self):
-        dockerfile = dedent(r"""\
-                FROM {base_image}
-                USER root
-                RUN sed -i -e 's/^\(nifi.remote.input.host\)=.*/\1={name}/' {nifi_root}/conf/nifi.properties
-                RUN sed -i -e 's/^\(nifi.remote.input.socket.port\)=.*/\1=5000/' {nifi_root}/conf/nifi.properties
```
fgerlits, line 143: is `nifi.remote.input.socket.port` no longer needed?

lordgamez: I don't think it was even used before this change, as port 8080 was used for site to site tests.

fgerlits (approved):

martinzink (approved):

PR #1287, opened 2022-03-25
MINIFICPP-1785 Fix CWEL attributes to be valid in HTTP requests

lordgamez:
When ConsumeWindowsEventLogs processor creates a flowfile and that is forwarded by the InvokeHTTP processor the attributes are written as HTTP header. HTTP headers do not allow spaces in the header names, so due to the `Timezone name` and `Timezone offset` attributes the requests would always result in code 400. These attributes are changed to avoid this issue.

https://issues.apache.org/jira/browse/MINIFICPP-1785

-------------------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/windows-event-log/ConsumeWindowsEventLog.cpp
+++ b/extensions/windows-event-log/ConsumeWindowsEventLog.cpp
@@ -712,8 +712,8 @@ void ConsumeWindowsEventLog::putEventRenderFlowFileToSession(const EventRender& 
       session.write(flowFile, &wc);
     }
     session.putAttribute(flowFile, core::SpecialFlowAttribute::MIME_TYPE, mimeType);
-    session.putAttribute(flowFile, "Timezone name", timezone_name_);
-    session.putAttribute(flowFile, "Timezone offset", timezone_offset_);
+    session.putAttribute(flowFile, "timezone.name", timezone_name_);
+    session.putAttribute(flowFile, "timezone.offset", timezone_offset_);
     session.getProvenanceReporter()->receive(flowFile, provenanceUri_, getUUIDStr(), "Consume windows event logs", 0ms);
     session.transfer(flowFile, Success);
   };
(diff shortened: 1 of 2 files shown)


adam-markovics (approved):

szaszm (approved):

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -712,8 +712,8 @@ void ConsumeWindowsEventLog::putEventRenderFlowFileToSession(const EventRender&
       session.write(flowFile, &wc);
     }
     session.putAttribute(flowFile, core::SpecialFlowAttribute::MIME_TYPE, mimeType);
-    session.putAttribute(flowFile, "Timezone name", timezone_name_);
-    session.putAttribute(flowFile, "Timezone offset", timezone_offset_);
+    session.putAttribute(flowFile, "timezone.name", timezone_name_);
+    session.putAttribute(flowFile, "timezone.offset", timezone_offset_);
```
adamdebreceni, line 716: IMO it is the responsibility of InvokeHTTP to adjust the attribute keys to create a valid HTTP header, CWEL should not concern itself with how its output might be used

szaszm: I thought the same at first, but then most other attributes have this format of dot separated lowercase words, so it's improving consistency anyway.

adamdebreceni: I think this change has merit, but I also think that as it currently stands, it does not address the root cause of the issue, note that we also append all attributes of the "Data" xml node as flowfile attributes (granted attribute names probably cannot contain spaces)

lordgamez: You are right that that CWEL should not concern itself with other processors, but as windows event logs forwarded through InvokeHTTP is a use case that is being implemented by one of our users this should be fixed, also as Marton mentioned our attribute names should be consistent. The root cause is that the HTTP headers do not allow these kinds of attributes and we cannot really do much about that other than adapting our code for this scenario. If we find other instances like this I think they should also be corrected, but I don't think we have any other known issues of this.

szaszm (changes requested):
Let's wait with merging until we reach consensus

szaszm (approved):
This and #1321 together adequately handle ass of the issues IMO. I'm +1 on this, since it makes attributes names consistent with other processors. This will break backwards compatibility, but at this stage of the project, we shouldn't let this stop innovation.

adamdebreceni (approved):

PR #1292, opened 2022-03-31
MINIFICPP-1792 Fix build with cmake 3.23.0

lordgamez:
According to the CMake [GNUInstallDirs](https://cmake.org/cmake/help/v3.23/module/GNUInstallDirs.html) documentation `CMAKE_INSTALL_LIBDIR` can contain the multiarch-tuple on Debian besides the previously used `lib`/`lib64` values. On contrary to earlier versions this is now specific to the case when `CMAKE_INSTALL_PREFIX` is set to `/`, `/usr`, or `/usr/local`. As `/usr/local` is the default value, in our build this will be used setting the `CMAKE_INSTALL_LIBDIR` to for example `lib/x86_64-linux-gnu` which causes a contradiction with the library paths of the external projects built where `CMAKE_INSTALL_PREFIX` is passed for their install paths thus it will be set to either `lib` or `lib64`.

As this change caused problems in other projects as well, it will be reverted in the next cmake release (without reverting the documentation change), but reintroduced later in after a discussion according to these:
https://gitlab.kitware.com/cmake/cmake/-/issues/23365
https://gitlab.kitware.com/cmake/cmake/-/merge_requests/7127

This change should fix the current problem and avoid any problems later if the change is reintroduced. It should also fix similar problems listed in ARM compilations: https://issues.apache.org/jira/browse/MINIFICPP-1756

https://issues.apache.org/jira/browse/MINIFICPP-1792

----------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


martinzink (approved):

szaszm (approved):
Thanks for the fix. I tried compiling again on ARM, this time on an old Raspberry Pi 2B and it seems to fail only because of a minor rocksdb issue: https://github.com/facebook/rocksdb/issues/8609

adamdebreceni (approved):

PR #1294, opened 2022-04-01
MINIFICPP-1771: Reworked ListenSyslog

martinzink:

### For all changes:
     in the commit message?

### For code changes:

### Note:


docker/test/integration/steps/steps.py
```diff
@@ -621,6 +626,11 @@ def step_impl(context, num_flowfiles, duration):
     context.test.check_for_num_files_generated(int(num_flowfiles), timeparse(duration))
 
 
+@then("at least one flowfile is placed in the monitored directory in less than {duration}")
+def step_impl(context, duration):
+    context.test.check_for_num_files_generated(1, timeparse(duration))
```
adamdebreceni, line 631: doesn't this look for exactly one file?

extensions/standard-processors/processors/ListenSyslog.cpp
```diff
@@ -17,318 +14,283 @@
...
-          if ((uint64_t) (recvlen + getEventQueueByteSize()) <= static_cast<uint64_t>(_recvBufSize)) {
-            uint8_t *payload = new uint8_t[recvlen];
-            memcpy(payload, _buffer, recvlen);
-            putEvent(payload, recvlen);
-          }
-          ++it;
-        }
-      }
-    }
+  if (!context->getProperty(MaxBatchSize.getName(), max_batch_size_) || max_batch_size_ <= 0) {
```
adamdebreceni, line 88: I think having set `withDefaultValue<uint64_t>` the `UNSIGNED_LONG_VALIDATOR` is used and will reject this branch

extensions/standard-processors/processors/ListenSyslog.cpp
```diff
@@ -17,318 +14,283 @@
...
-      bp = b;
-    }
-    c = *bp++;
-    *bufptr++ = c;
-    if (c == '\n') {
-      *bufptr = '\n';
-      return bufptr - bufx + 1;
-    }
+  if (!context->getProperty(ParseMessages.getName(), parse_messages_)) {
+    parse_messages_ = ParseMessages.getDefaultValue();
```
adamdebreceni, line 94: isn't the default value returned when a property is not set?

martinzink: Didnt know that, but you are right.
Changed it in https://github.com/apache/nifi-minifi-cpp/pull/1294/commits/027fea1d65bd423940edd81d598cef68946e2202#diff-a22ae2c37869a184bd936ac719f98ae8bbdb1b8d2ad5b3e6fc48ea9207fe530bR88-R93

extensions/standard-processors/processors/ListenSyslog.cpp
martinzink: :+1: I've changed it to a custom validator https://github.com/apache/nifi-minifi-cpp/pull/1294/commits/027fea1d65bd423940edd81d598cef68946e2202#diff-a22ae2c37869a184bd936ac719f98ae8bbdb1b8d2ad5b3e6fc48ea9207fe530bR43

docker/test/integration/steps/steps.py
martinzink: you are right :+1: changed it to a range validator https://github.com/apache/nifi-minifi-cpp/pull/1294/commits/027fea1d65bd423940edd81d598cef68946e2202#diff-dce84c359f2f7128e501f5a322d8f6ac3325c2b471844f2c42cbef9abca58185R631

adamdebreceni (approved):

PROCESSORS.md
```diff
@@ -1002,27 +1002,52 @@ In the list below, the names of required properties appear in bold. Any other pr
...
-|Parse Messages|false||Indicates if the processor should parse the Syslog messages. If set to false, each outgoing FlowFile will only.|
-|Port|514||The port for Syslog communication|
-|Protocol|UDP|UDP<br>TCP<br>|The protocol for Syslog communication.|
-|Receive Buffer Size|65507 B||The size of each buffer used to receive Syslog messages.|
+| Name                      | Default Value | Allowable Values | Description                                                                                                                                                                                          |
+|---------------------------|---------------|------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Listening Port            | 514           |                  | The port for Syslog communication.                                                                                                                                                                   |
```
arpadboda, line 1017: I think a warning here would be useful about the required root priviliges in case of using the default port.

extensions/standard-processors/processors/ListenSyslog.cpp
```diff
@@ -17,318 +14,272 @@
...
-      }
-      _clientSockets.clear();
-      if (_serverSocket > 0) {
-        close(_serverSocket);
-        _serverSocket = 0;
-      }
-    }
+  Protocol protocol = Protocol(ProtocolProperty.getDefaultValue());
+  if (!context->getProperty(ProtocolProperty.getName(), protocol)) {
+    logger_->log_error("Missing or invalid Protocol: defaulting to %s", protocol.toString());
```
arpadboda, line 97: Not sure if this can occur anyhow given the validator, but in case it does, it should throw

extensions/standard-processors/processors/ListenSyslog.cpp
```diff
@@ -17,318 +14,272 @@
...
-    int retval;
-    fds = _readfds;
-    tv.tv_sec = 0;
-    // 100 msec
-    tv.tv_usec = 100000;
-    retval = select(_maxFds + 1, &fds, NULL, NULL, &tv);
-    if (retval < 0)
+  int port = Port.getDefaultValue();
+  if (!context->getProperty(Port.getName(), port)) {
+    logger_->log_error("Missing or invalid Port: defaulting to %s", port);
```
arpadboda, line 102: The same concerns here as in case of the protocol

PROCESSORS.md
martinzink: Good idea, I've changed it in https://github.com/apache/nifi-minifi-cpp/pull/1294/commits/251865e006f1e53e45c33fdaa184a9259f7bf13d

docker/test/integration/minifi/core/ImageStore.py
```diff
@@ -181,6 +185,12 @@ def __build_mqtt_broker_image(self):
     def __build_splunk_image(self):
         return self.__build_image_by_path(self.test_dir + "/resources/splunk-hec", 'minifi-splunk')
 
+    def __build_syslog_udp_client_image(self):
+        return self.__build_image_by_path(self.test_dir + "/resources/syslog-client/udp", 'syslog-udp-client')
+
+    def __build_syslog_tcp_client_image(self):
+        return self.__build_image_by_path(self.test_dir + "/resources/syslog-client/tcp", 'syslog-tcp-client')
```
lordgamez, line 192: Do these need to be built as new images? Looking at the Dockerfiles it seems to me that we could use the ubuntu image and define the entrypoint parameter with the command in the image. This way we could remove the images, and have only 1 container with 2 different commands for TCP and UDP.

extensions/standard-processors/processors/ListenSyslog.cpp
martinzink: Good point. I checked and since these are required properties the context->getProperty will already throw if it can't find the property.
So these branches are unreachable. I removed them in https://github.com/apache/nifi-minifi-cpp/pull/1294/commits/251865e006f1e53e45c33fdaa184a9259f7bf13d#diff-a22ae2c37869a184bd936ac719f98ae8bbdb1b8d2ad5b3e6fc48ea9207fe530bR95-R99

extensions/standard-processors/processors/ListenSyslog.cpp
martinzink: Good point. I checked and since these are required properties the context->getProperty will already throw if it can't find the property.
So these branches are unreachable. I removed them in https://github.com/apache/nifi-minifi-cpp/pull/1294/commits/251865e006f1e53e45c33fdaa184a9259f7bf13d#diff-a22ae2c37869a184bd936ac719f98ae8bbdb1b8d2ad5b3e6fc48ea9207fe530bR95-R99

docker/test/integration/resources/syslog-client/tcp/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM ubuntu:latest
```
lordgamez, line 1: Please use a specific version instead of latest as the latest tag could refer to another version later.

docker/test/integration/minifi/core/ImageStore.py
martinzink: Good idea. I've changed this in https://github.com/apache/nifi-minifi-cpp/pull/1294/commits/26053fde5f0a6d4ed055ee98558d9653fd89254a

docker/test/integration/resources/syslog-client/tcp/Dockerfile
martinzink: Good idea. I've changed this in https://github.com/apache/nifi-minifi-cpp/pull/1294/commits/26053fde5f0a6d4ed055ee98558d9653fd89254a

docker/test/integration/minifi/core/ImageStore.py
martinzink: Replaced it with a simpler version, based on our conversation :+1:  https://github.com/apache/nifi-minifi-cpp/pull/1294/commits/29d3690e8bd030a1315f6c72fbac0f1a3d0c2cc8

arpadboda (approved):

PR #1296, opened 2022-04-06
MINIFICPP-1794 Remove sensitive properties from agent manifest

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-1794

----------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/src/core/state/nodes/SupportedOperations.cpp
+++ b/libminifi/src/core/state/nodes/SupportedOperations.cpp
@@ -70,7 +70,7 @@ SupportedOperations::Metadata SupportedOperations::buildUpdatePropertiesMetadata
     if (ranges::find(sensitive_properties, config_property.name) != ranges::end(sensitive_properties)) {
       continue;
     }
-    if (!update_policy_controller_ || (update_policy_controller_ && update_policy_controller_->canUpdate(std::string(config_property.name)))) {
+    if (!update_policy_controller_ || update_policy_controller_->canUpdate(std::string(config_property.name))) {
       std::unordered_map<std::string, std::string> property;
       property.emplace("propertyName", config_property.name);
       property.emplace("validator", config_property.validator->getName());
--- a/libminifi/src/core/state/nodes/SupportedOperations.cpp
+++ b/libminifi/src/core/state/nodes/SupportedOperations.cpp
@@ -18,6 +18,7 @@
 
 #include "core/state/nodes/SupportedOperations.h"
 #include "core/Resource.h"
+#include "range/v3/algorithm/find.hpp"
 
 namespace org::apache::nifi::minifi::state::response {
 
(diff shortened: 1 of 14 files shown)


libminifi/src/core/state/nodes/SupportedOperations.cpp
```diff
@@ -62,11 +63,27 @@ void SupportedOperations::addProperty(SerializedResponseNode& properties, const
...
+  sensitive_properties.assign(Configuration::DEFAULT_SENSITIVE_PROPERTIES.begin(), Configuration::DEFAULT_SENSITIVE_PROPERTIES.end());
+  if (configuration_reader_) {
+    const auto additional_sensitive_props_list = configuration_reader_(Configuration::nifi_sensitive_props_additional_keys);
+    if (additional_sensitive_props_list) {
+      std::vector<std::string> additional_sensitive_properties = utils::StringUtils::split(*additional_sensitive_props_list, ",");
+      return Configuration::mergeProperties(sensitive_properties, additional_sensitive_properties);
+    }
+  }
+  return sensitive_properties;
+}
```
adamdebreceni, line 77: it seems to me, that there are three separate instances of building up the sensitive properties list, could we somehow unify these?

lordgamez: I was thinking about this, when I checked how these properties are used in the encrypt-config. There was one major difference though that the encrypt-config reads these properties from a ConfigFile object which is a PropertiesFile representation and not from a whereas we only have a function object available to read a single property from a Configuration object in the SupportedProperties.

As for the tests we also have a similar problem where we only pass a function object to the SupportedOperations to reach the configuration properties while in the tests we have the actual Configuration object available, so it's not that easy to generalize it, but I was also worried about to use the same code that we are testing in the test as well.

adamdebreceni: the least common denominator seems to be the function object, we could go for a static `Configuration::getSensitiveProperties(std::function<std::string(const std::string)> reader)`

lordgamez: That's a fair point, extracted the function in 17c2db229f70f3453489cc9e2a7a96a65422da15

libminifi/src/core/state/nodes/SupportedOperations.cpp
```diff
@@ -64,9 +65,12 @@ void SupportedOperations::addProperty(SerializedResponseNode& properties, const
...
   std::vector<std::unordered_map<std::string, std::string>> supported_config_updates;
-  for (const auto& config_property : minifi::Configuration::CONFIGURATION_PROPERTIES) {
-    if (!update_policy_controller_ ||
-        (update_policy_controller_ && update_policy_controller_->canUpdate(std::string(config_property.name)))) {
+  for (const auto& config_property : Configuration::CONFIGURATION_PROPERTIES) {
+    auto sensitive_properties = Configuration::getSensitiveProperties(configuration_reader_);
+    if (ranges::find(sensitive_properties, config_property.name) != ranges::end(sensitive_properties)) {
+      continue;
+    }
+    if (!update_policy_controller_ || (update_policy_controller_ && update_policy_controller_->canUpdate(std::string(config_property.name)))) {
```
martinzink, line 73:
```suggestion
    if (!update_policy_controller_ || update_policy_controller_->canUpdate(std::string(config_property.name))) {
```

The second checking of update_policy_controller_ seems unnecessary.

martinzink (commented):
LGTM, just a small suggestion

libminifi/src/core/state/nodes/SupportedOperations.cpp
lordgamez: Good point, updated in 8561e67e7453596a5158f2ab937fc724ec551974

martinzink (approved):

arpadboda (approved):

PR #1300, opened 2022-04-13
MINIFICPP-1675 Use regex.h instead of std::regex when using libstdc++

lordgamez:
Due to a bug in the libstdc++ implementation of std::regex it is unsafe to use std::regex_match and std::regex_search on large texts as the process can crash due to a stack overflow. In this PR std::regex usages are replaced with minifi::utils::Regex which uses regex.h if libstdc++ is used otherwise it uses std::regex.

There are a few exceptions:

- std::regex is still used for regex_replace calls
- AppendHostInfo still uses std::regex as POSIX regex grammar is not sufficient for easily filtering interface names, and there is no risk of running into the bug as the interface names are small
- In tests and mocks we still use std::regex as they are only run in our test environments with controlled inputs and sometimes use regex patterns not available in POSIX grammar

-----------------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/standard-processors/tests/unit/ExtractTextTests.cpp
```diff
@@ -180,3 +181,36 @@ TEST_CASE("Test usage of ExtractText in regex mode", "[extracttextRegexTest]") {
 
   LogTestController::getInstance().reset();
 }
+
+#ifndef NO_MORE_REGFREEE
```
szaszm, line 185: Is this `#ifndef` required? It should work in either case if I understand correctly.

lordgamez: Good catch, it was turned off while testing locally, I readded it in affce89dfb183b10598bf0e8a4351a0e92db3fb2

szaszm (approved):

PR #1310, opened 2022-04-20
MINIFICPP-1806 - Use boyer_moore for extension verification

adamdebreceni:
For a debug build the extension verification for `libcore-minifi.so` took ~1400 ms (which together with other dynamic libraries adds seconds to the startup time during development), with this change it takes ~100 ms.

Alternatives considered:
- boyer_moore + custom span concat: ~150 ms
- boyer_moore + ranges::concat: ~1700 ms
- boyer_moore + mmap: ~100 ms (the mmap wrapper would add significant complexity)

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/src/utils/file/FileUtils.cpp
+++ b/libminifi/src/utils/file/FileUtils.cpp
@@ -21,35 +21,9 @@
 
 #include <algorithm>
 #include <iostream>
-#include <version>
 
 #include "utils/Literals.h"
-
-#ifdef _LIBCPP_VERSION
-
-template<typename ...Args>
-static auto make_searcher(Args&& ...args) {
-  return std::default_searcher(std::forward<Args>(args)...);
-}
-
-#elif __cpp_lib_boyer_moore_searcher < 201603L
-#include <experimental/functional>
-
-template<typename ...Args>
-static auto make_searcher(Args&& ...args) {
-  return std::experimental::boyer_moore_searcher(std::forward<Args>(args)...);
-}
-
-#else
-#include <functional>
-
-template<typename ...Args>
-static auto make_searcher(Args&& ...args) {
-  return std::boyer_moore_searcher(std::forward<Args>(args)...);
-}
-
-#endif
-
+#include "utils/Searcher.h"
 
 namespace org {
 namespace apache {
@@ -82,7 +56,7 @@ bool contains(const std::filesystem::path& file_path, std::string_view text_to_s
   gsl_ExpectsAudit(std::filesystem::exists(file_path));
   std::array<char, 16_KiB> buf{};
 
-  auto searcher = make_searcher(text_to_search.begin(), text_to_search.end());
+  Searcher searcher(text_to_search.begin(), text_to_search.end());
 
   std::ifstream ifs{file_path, std::ios::binary};
   do {
--- a/libminifi/src/utils/file/FileUtils.cpp
+++ b/libminifi/src/utils/file/FileUtils.cpp
@@ -21,16 +21,35 @@
 
 #include <algorithm>
 #include <iostream>
+#include <version>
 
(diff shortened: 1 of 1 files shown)


martinzink (approved):

lordgamez:
I think the clang job should be investigated, the FileUtilTests failure might relate to this change:

```
double free or corruption (!prev)


99% tests passed, 1 tests failed out of 183

Total Test time (real) = 300.55 sec

The following tests FAILED:
	 27 - FileUtilsTests (Timeout)

```

adamdebreceni:
> I think the clang job should be investigated, the FileUtilTests failure might relate to this change:
> 
> ```
> double free or corruption (!prev)
> 
> 
> 99% tests passed, 1 tests failed out of 183
> 
> Total Test time (real) = 300.55 sec
> 
> The following tests FAILED:
> 	 27 - FileUtilsTests (Timeout)
> ```

yes it is currently being investigated, I will move it to draft until it is sorted out

szaszm (approved):

lordgamez (approved):

libminifi/src/utils/file/FileUtils.cpp
```diff
@@ -49,41 +52,21 @@ uint64_t computeChecksum(const std::string &file_name, uint64_t up_to_position)
...
-    ifs.read(right.data(), gsl::narrow<std::streamsize>(right.size()));
-    if (check_range(0, left.size())) return true;
+    std::copy(buf.end() - text_to_search.size(), buf.end(), buf.begin());
+    ifs.read(buf.data() + text_to_search.size(), buf.size() - text_to_search.size());
+    if (std::search(buf.begin(), buf.end(), searcher) != buf.end()) {
+      return true;
+    }
   } while (ifs);
-  return check_range(left.size(), left.size() + right.size());
+  return std::search(buf.begin(), buf.end(), searcher) != buf.end();
```
fgerlits, line 69: After the last chunk is read, if it is shorter than the capacity of the buffer, `buf` will contain bytes remaining from the previous chunk at the end.  Is it possible that e.g `text_to_search` is `"abcde"`, and it is found incorrectly because the end of the file is `"abc"`, and it is followed by `"de"` in the buffer because that happened to be there from the previous chunk?

adamdebreceni: indeed it could result in false positives, updated it in [18e826e](https://github.com/apache/nifi-minifi-cpp/pull/1310/commits/18e826eb034124eacfe13edc90f588e1f6c0589a)

libminifi/src/utils/file/FileUtils.cpp
```diff
@@ -49,41 +52,23 @@ uint64_t computeChecksum(const std::string &file_name, uint64_t up_to_position)
...
-      size_t j{};
-      for (j = 0; j < text_to_search.size(); ++j) {
-        if (charat(i + j) != text_to_search[j]) break;
-      }
-      if (j == text_to_search.size()) return true;
-    }
-    return false;
-  };
+  std::array<char, 16_KiB> buf{};
+  gsl::span<char> view;
```
fgerlits, line 58: we have `std::span` in C++20; or is that not supported on some of our platforms, yet?

adamdebreceni: it is not yet used in our codebase, if there is no technical reason behind this, I think we should dedicate a separate PR to replace all `gsl::span` with `std::span`

fgerlits: OK, I see there is a Jira for this already: https://issues.apache.org/jira/browse/MINIFICPP-1755.

fgerlits (approved):

PR #1315, opened 2022-04-21
MINIFICPP-1810 Provide logs to be read from docker logs

lordgamez:
We log to stderr by default in docker containers from now on and let the docker logging engine take care of the logs.

https://issues.apache.org/jira/browse/MINIFICPP-1810

------------------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


docker/Dockerfile
```diff
@@ -174,9 +174,12 @@ RUN addgroup -g ${GID} ${USER} && adduser -u ${UID} -D -G ${USER} -g "" ${USER}
...
+RUN mkdir ${MINIFI_HOME}/scripts
+COPY --from=build --chown=${USER}:${USER} ${MINIFI_BASE_DIR}/docker/scripts/start.sh ${MINIFI_HOME}/scripts
+RUN chmod +x ${MINIFI_HOME}/scripts/start.sh
 
 USER ${USER}
 WORKDIR ${MINIFI_HOME}
 
 # Start MiNiFi CPP in the foreground
-CMD ["./bin/minifi.sh", "run"]
+CMD ["./scripts/start.sh"]
```
szaszm, line 185: Instead of this script, we should configure the minifi logger to output logs to stdout, through `minifi-log.properties`. Can we do that in the context of the docker image?

szaszm: stderr would be even better if docker and k8s is capable of using that.

lordgamez: Good point, I checked with the NiFi team and it seems that they use this method due to legacy problems, but will be changed in the future, so we do not have any reason not to use stderr in a docker container by default for logging. I updated the default log properties in 38705267358e4e56f40a9a720dbcd68508a0a69f and we can use `docker logs` for every container in our test environment as well from now on.

szaszm (approved):

adam-markovics (approved):

arpadboda (approved):

PR #1320, opened 2022-04-28
MINIFICPP-1813 Stabilize AbsoluteTimeoutTest in CI on Windows

lordgamez:
Due to scheduling issues when too many tests are run parallel the read timeout limit can be reached before we reach the absolute timeout (3 times the read timeout).

https://issues.apache.org/jira/browse/MINIFICPP-1813

------------------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/http-curl/tests/AbsoluteTimeoutTest.cpp
+++ b/extensions/http-curl/tests/AbsoluteTimeoutTest.cpp
@@ -22,20 +22,14 @@
 #include "tests/TestServer.h"
 #include "HTTPHandlers.h"
 
+using namespace std::literals::chrono_literals;
+
 int main() {
   TestController controller;
 
   std::string port = "12324";
   std::string rootURI =  "/";
-  TimeoutingHTTPHandler handler({
-    std::chrono::milliseconds(500),
-    std::chrono::milliseconds(500),
-    std::chrono::milliseconds(500),
-    std::chrono::milliseconds(500),
-    std::chrono::milliseconds(500),
-    std::chrono::milliseconds(500),
-    std::chrono::milliseconds(500)
-  });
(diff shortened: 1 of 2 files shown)


extensions/http-curl/tests/AbsoluteTimeoutTest.cpp
```diff
@@ -22,20 +22,14 @@
...
-  TimeoutingHTTPHandler handler({
-    std::chrono::milliseconds(500),
-    std::chrono::milliseconds(500),
-    std::chrono::milliseconds(500),
-    std::chrono::milliseconds(500),
-    std::chrono::milliseconds(500),
-    std::chrono::milliseconds(500),
-    std::chrono::milliseconds(500)
-  });
+  TimeoutingHTTPHandler handler({35, 100ms});
```
szaszm, line 32: I would change this to have explicit type with parens instead of braces. Usually I prefer this style, but in this case, I want to make it clear that this is not the initializer_list contstructor of std::vector, but it's going to create 35 elements with the value 100ms. I got bitten by a similar thing when I tried to direct-list-initialize a string with N elements of a certain character, but it got converted to char and used the initializer_list constructor instead of the two parameter one. After changing to parens, it used the two parameter ctor.
```suggestion
  TimeoutingHTTPHandler handler(std::vector<std::chrono::milliseconds>(35, 100ms));
```

lordgamez: You are right, it's more explicit and clear this way, updated in 8497f10ca144a7d8053d83ae67c24942168e4540

szaszm (approved):

adamdebreceni (approved):

martinzink (approved):

PR #1326, opened 2022-05-05
MINIFICPP-1823 Fix absolute.path output attribute in ListFile

lordgamez:
Adjusted absolute.path attribute according to the output description and the FetchFile expectation.

https://issues.apache.org/jira/browse/MINIFICPP-1823

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/standard-processors/processors/ListFile.cpp
+++ b/extensions/standard-processors/processors/ListFile.cpp
@@ -142,39 +142,39 @@ void ListFile::onSchedule(const std::shared_ptr<core::ProcessContext> &context, 
 }
 
 bool ListFile::fileMatchesFilters(const ListedFile& listed_file) {
-  if (ignore_hidden_files_ && utils::file::FileUtils::is_hidden(listed_file.absolute_path)) {
-    logger_->log_debug("File '%s' is hidden so it will not be listed", listed_file.absolute_path);
+  if (ignore_hidden_files_ && utils::file::FileUtils::is_hidden(listed_file.full_file_path)) {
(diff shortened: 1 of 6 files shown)


extensions/standard-processors/processors/ListFile.cpp
```diff
@@ -234,18 +234,19 @@ void ListFile::onTrigger(const std::shared_ptr<core::ProcessContext> &context, c
   auto file_list = utils::file::FileUtils::list_dir_all(input_directory_, logger_, recurse_subdirectories_);
   for (const auto& [path, filename] : file_list) {
     ListedFile listed_file;
-    listed_file.absolute_path = (std::filesystem::path(path) / filename).string();
+    listed_file.full_file_path = (std::filesystem::path(path) / filename).string();
+    listed_file.absolute_path = path + utils::file::FileUtils::get_separator();
```
szaszm, line 238: I don't think absolute_path has to end with a separator.

szaszm: Nevermind, just checked the nifi docs and they do it the same way.

adamdebreceni:
note: it seems to me that both `TailFile` and `FetchS3Object` treat `"absolute.path"` as the path to the file itself, not to its directory

lordgamez:
> note: it seems to me that both `TailFile` and `FetchS3Object` treat `"absolute.path"` as the path to the file itself, not to its directory

Yes, that was the reason of this bug that I expected that it works the same way here, but NiFi also has this difference between processors.

PR #1328, opened 2022-05-06
MINIFICPP-1812 - Clean up Repository threads

adam-markovics:
https://issues.apache.org/jira/browse/MINIFICPP-1812

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/include/core/repository/VolatileRepository.h
```diff
@@ -39,31 +39,24 @@ namespace nifi {
...
-#pragma GCC diagnostic ignored "-Woverloaded-virtual"
-#endif
+
 /**
  * Flow File repository
  * Design: Extends Repository and implements the run function, using RocksDB as the primary substrate.
  */
-template<typename T>
-class VolatileRepository : public core::Repository {
+template<typename T, typename T_Repository = core::Repository>
```
lordgamez, line 47: Maybe you could rename the template arguments to KeyType and RepositoryType to be more clear

libminifi/src/core/RepositoryFactory.cpp
```diff
@@ -96,6 +96,62 @@ std::unique_ptr<core::ContentRepository> createContentRepository(const std::stri
...
+    } else {
+      throw std::runtime_error("Support for the provided configuration class could not be found");
+    }
+  } catch (const std::runtime_error &) {
+    if (fail_safe) {
+      return {};
+    }
+  }
+
+  throw std::runtime_error("Support for the provided configuration class could not be found");
```
lordgamez, line 152: We don't need to throw here if we just call a `throw` in the catch block rethrowing the original exception.

libminifi/src/core/RepositoryFactory.cpp
```diff
@@ -96,6 +96,62 @@ std::unique_ptr<core::ContentRepository> createContentRepository(const std::stri
...
+}
+
 } /* namespace core */
 } /* namespace minifi */
 } /* namespace nifi */
```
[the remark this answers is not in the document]
adam-markovics: Done.

libminifi/src/core/RepositoryFactory.cpp
```diff
@@ -96,6 +96,62 @@ std::unique_ptr<core::ContentRepository> createContentRepository(const std::stri
...
+  }
+
+  std::thread& getThread() override {
+    return thread_;
+  }
+
+  std::thread thread_;
+};
+
+std::unique_ptr<core::ThreadedRepository> createThreadedRepository(const std::string& configuration_class_name, bool fail_safe, const std::string& repo_name) {
```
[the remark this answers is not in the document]
adam-markovics: Done.

libminifi/src/core/RepositoryFactory.cpp
adam-markovics: Done.

libminifi/src/core/RepositoryFactory.cpp
```diff
@@ -96,6 +96,62 @@ std::unique_ptr<core::ContentRepository> createContentRepository(const std::stri
...
+    if (class_name_lc == "flowfilerepository" || class_name_lc == "volatileflowfilerepository") {
+      return_obj = instantiate<repository::VolatileFlowFileRepository>(repo_name);
+    } else if (class_name_lc == "provenancerepository" || class_name_lc == "volatileprovenancefilerepository") {
+      return_obj = instantiate<repository::VolatileProvenanceRepository>(repo_name);
+    } else if (class_name_lc == "nooprepository") {
+      return_obj = std::make_unique<core::NoOpThreadedRepository>(repo_name);
+    }
+    if (return_obj) {
+      return return_obj;
+    }
```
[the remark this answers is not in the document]
adam-markovics: Done.

libminifi/include/core/Repository.h
```diff
@@ -240,30 +205,20 @@ class Repository : public virtual core::SerializableComponent, public core::Trac
   std::map<std::string, core::Connectable*> containers_;
 
   std::map<std::string, core::Connectable*> connection_map_;
-  // Mutex for protection
-  std::mutex mutex_;
   // repository directory
```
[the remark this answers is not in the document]
adam-markovics: Done.

libminifi/include/core/Repository.h
```diff
@@ -240,30 +205,20 @@ class Repository : public virtual core::SerializableComponent, public core::Trac
...
   // max db size
   int64_t max_partition_bytes_;
   // purge period
   std::chrono::milliseconds purge_period_;
-  // thread
-  std::thread thread_;
-  // whether the monitoring thread is running for the repo while it was enabled
-  std::atomic<bool> running_;
-  // whether stop accepting provenace event
+  // whether stop accepting provenance event
```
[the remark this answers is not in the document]
adam-markovics: Done.

extensions/rocksdb-repos/FlowFileRepository.h
```diff
@@ -27,30 +27,25 @@
...
+constexpr auto FLOWFILE_REPOSITORY_DIRECTORY = ".\\flowfile_repository";
+constexpr auto FLOWFILE_CHECKPOINT_DIRECTORY = ".\\flowfile_checkpoint";
 #else
-#define FLOWFILE_REPOSITORY_DIRECTORY "./flowfile_repository"
-#define FLOWFILE_CHECKPOINT_DIRECTORY "./flowfile_checkpoint"
+constexpr auto FLOWFILE_REPOSITORY_DIRECTORY = "./flowfile_repository";
+constexpr auto FLOWFILE_CHECKPOINT_DIRECTORY = "./flowfile_checkpoint";
 #endif
-#define MAX_FLOWFILE_REPOSITORY_STORAGE_SIZE (10*1024*1024)  // 10M
+constexpr auto MAX_FLOWFILE_REPOSITORY_STORAGE_SIZE = 10*1024*1024;  // 10M
```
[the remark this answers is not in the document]
adam-markovics: Done.

lordgamez (approved):

libminifi/include/core/ThreadedRepository.h
```diff
@@ -0,0 +1,86 @@
...
+
+  // Starts repository monitor thread
+  void start() {
+    if (running_.exchange(true)) {
+      return;
+    }
+    if (purge_period_ <= std::chrono::milliseconds(0)) {
+      return;
+    }
+    getThread() = std::thread(&ThreadedRepository::run, this);
```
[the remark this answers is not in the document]
adamdebreceni: I don't think there is such a restriction, `&ThreadedRepository::run` is a pointer to member function which, together with the `this`, seem to be a suitable arguments for the `std::thread` constructor, the behind the scenes magic (most likely offset into the vtable) makes sure that when called the overridden method based on the objects dynamic type is called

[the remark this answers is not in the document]
szaszm: You're right. I didn't find anything in the standard that disallows this, and the member function pointer explanation in [this article](https://www.codeproject.com/Articles/7150/Member-Function-Pointers-and-the-Fastest-Possible) suggests that they keep track internally whether it's virtual or non-virtual and single or multiple inheritance.
I'd still replace it with a lambda, simply because that looks simpler to me, but it's OK as is.

libminifi/include/core/RepositoryFactory.h
```diff
@@ -49,10 +45,14 @@ std::unique_ptr<core::Repository> createRepository(const std::string& configurat
...
-}  // namespace nifi
-}  // namespace apache
-}  // namespace org
+/**
+ * Create a threaded repository represented by the configuration class name
+ * @param configuration_class_name configuration class name
+ * @param fail_safe determines whether or not to make the default class if configuration_class_name is invalid
+ * @param repo_name name of the repository
+ */
+std::unique_ptr<core::ThreadedRepository> createThreadedRepository(const std::string& configuration_class_name, const std::string& repo_name = "");
```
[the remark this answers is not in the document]
adam-markovics: `core::createRepository` has been deleted

libminifi/include/core/Repository.h
```diff
@@ -20,89 +20,77 @@
...
 
-#define REPOSITORY_DIRECTORY "./repo"
-#define MAX_REPOSITORY_STORAGE_SIZE (10*1024*1024)  // 10M
+constexpr auto REPOSITORY_DIRECTORY = "./repo";
+constexpr auto MAX_REPOSITORY_STORAGE_SIZE = 10_MiB;
 constexpr auto MAX_REPOSITORY_ENTRY_LIFE_TIME = std::chrono::minutes(10);
 constexpr auto REPOSITORY_PURGE_PERIOD = std::chrono::milliseconds(2500);
 
-class Repository : public virtual core::SerializableComponent, public core::TraceableResource {
+class Repository : public virtual core::SerializableComponent {
```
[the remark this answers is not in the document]
adam-markovics: I chose `initialize`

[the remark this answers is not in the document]
szaszm: Why not all of those that don't have a _real_ default implementation?

libminifi/include/core/RepositoryFactory.h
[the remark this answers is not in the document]
adam-markovics: They are merged now.

libminifi/include/core/Repository.h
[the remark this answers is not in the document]
adam-markovics: `ThreadedRepository` is gone, in `Repository` `run` and `getThread` are already pure virtual.

libminifi/include/core/Repository.h
```diff
@@ -218,56 +195,77 @@ class Repository : public virtual core::SerializableComponent, public core::Trac
...
+  // Stops repository monitor thread
+  void stop() {
+    if (!running_.exchange(false)) {
+      return;
+    }
+    if (getThread().joinable()) {
+      getThread().join();
+    }
+    logger_->log_debug("%s Repository monitor thread stop", name_);
+  }
```
szaszm, line 231: It's possible to start again on another thread after line 224 has finished, and we're waiting for 227 or 228. One should not be able to start the repository until stopping has finished. Same with start, you shouldn't be able to stop until starting has finished.
I suggest using a state enum with 4 values: starting, running, stopping, stopped.
Only transitions to the right and stopped -> starting should be possible.

libminifi/include/core/repository/VolatileContentRepository.h
```diff
@@ -68,55 +59,55 @@ class VolatileContentRepository :
...
+
+ private:
+  void run() override {
+  }
 
- protected:
-  virtual void start();
+  std::thread& getThread() override {
+    return thread_;
+  }
```
szaszm, line 107: This class doesn't need a thread. Can you think of a structure that doesn't require all repositories to have a thread?

adam-markovics: The previous one was like that. You asked to be like this: https://github.com/apache/nifi-minifi-cpp/pull/1328#discussion_r923437527

szaszm: You're right, I asked for it, because I thought that all repositories are "threaded". Obviously that's not the case. Additionally, I don't think `ThreadedRepository`, or the "threadedness" of a repository is something user code should be concerned with, so I don't say we should go back to the previous revision. But an empty/noop implementation is basically a violation of the [LSP](https://en.wikipedia.org/wiki/Liskov_substitution_principle).

My first thought would be to add a thread to every "threaded" repostitory, and implement the start/stop logic in all such leaf classes. That would go against [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). That could be solved by reintroducing `ThreadedRepository` as an implementation helper, **not exposed to user code**. This is still an anti-pattern, by reusing implementation through inheritance, but this seems to be the least evil, and if we can enforce no naming of this class outside of Repository implementations, then it's relatively contained and harmless IMO. I would do this. If you know better, please comment it, so we can discuss.

In general, you understand your code best. When I comment an idea, and you run into an issue with it (e.g. [LSP](https://en.wikipedia.org/wiki/Liskov_substitution_principle) violation in the last case), or I just comment something stupid, please ask or push back. Solving a small problem by making a bigger one is not good.

adam-markovics: Okay, I will partially revert to previous implementation of having a ThreadedRepository, but without exposing it anywhere. If it is possible. If I run into something that prevents it, I will update here.

adam-markovics: It looks like the the threadedness of a repository must be exposed, because for example [FlowController uses it](https://github.com/adam-markovics/nifi-minifi-cpp/blob/53fbb264cb00fbb4189a2fb1ae7978c6913d3a1c/libminifi/src/FlowController.cpp#L381)
A threadless repo should not have a `start` method, it would be worse in my opinion.

szaszm: I think a pure virtual `start` on `Repository` would be a valid approach, I could imagine repositories not needing a separate thread, but still needing start/stop logic, e.g. a synchronous DB repo. We don't have any of those as far as I'm aware, I'm just justifying keeping `start` in the `Repostory` base class. What do you think?

adam-markovics: This makes sense, so be it.

libminifi/include/core/Repository.h
adam-markovics: Done.

libminifi/include/core/repository/VolatileContentRepository.h
adam-markovics: Done.

libminifi/include/core/ThreadedRepository.h
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  bool initialize(const std::shared_ptr<Configure>& /*configure*/) override {
+    return true;
+  }
+
+  // Starts repository monitor thread
+  bool start() override {
+    // if STOPPED, turn to STARTING, otherwise return
+    RUNNING_STATE expected{RUNNING_STATE::STOPPED};
```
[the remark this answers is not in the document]
adam-markovics: Done.

szaszm (approved):

libminifi/include/core/repository/FileSystemRepository.h
```diff
@@ -25,49 +25,39 @@
...
 /**
  * FileSystemRepository is a content repository that stores data onto the local file system.
  */
 class FileSystemRepository : public core::ContentRepository, public core::CoreComponent {
  public:
   FileSystemRepository(std::string name = getClassName<FileSystemRepository>()) // NOLINT
-      : core::CoreComponent(name),
-        logger_(logging::LoggerFactory<FileSystemRepository>::getLogger()) {
+          : core::CoreComponent(name),
+            logger_(logging::LoggerFactory<FileSystemRepository>::getLogger()) {
```
szaszm, line 38: The old indentation was better here

adam-markovics: Done.

adamdebreceni (approved):

PR #1330, opened 2022-05-11
MINIFICPP-1832 Integrate first batch of clang-tidy checks in CI

lordgamez:
Compile project with clang-14 generating compile_commands.json and use it to analyze code with clang-tidy-14. Clang-tidy is run on every cpp file modified in the pull request. Only an initial batch of checks is implemented, other checks should be implemented as part of [MINIFICPP-1833](https://issues.apache.org/jira/browse/MINIFICPP-1833) or other further tickets.

https://issues.apache.org/jira/browse/MINIFICPP-1832

-----------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


thirdparty/rapidjson-48fbd8cd202ca54031fe799db2ad44ffa8e77c13/include/rapidjson/document.h
```diff
@@ -1236,7 +1236,7 @@ class GenericValue {
 
             // Use static buffer and placement-new to prevent destruction
             static char buffer[sizeof(GenericValue)];
-            return *new (buffer) GenericValue();
+            return *new (buffer) GenericValue();  // NOLINT(clang-analyzer-cplusplus.PlacementNew)
```
szaszm, line 1239: Can we exclude the thirdparty directory instead?

libminifi/test/unit/SocketTests.cpp
```diff
@@ -190,7 +190,7 @@ TEST_CASE("TestTLSContextCreation", "[TestSocket8]") {
     std::function<bool()> f_ex = createSocket;
     utils::Worker<bool> functor(f_ex, "id");
     std::future<bool> fut;
-    pool.execute(std::move(functor), fut);
+    pool.execute(std::move(functor), fut);  // NOLINT(bugprone-use-after-move)
```
szaszm, line 193: What kind of use-after-move is happening here? This looks perfectly valid to me, not reusing anything after move.

libminifi/src/controllers/SSLContextService.cpp
```diff
@@ -213,7 +213,8 @@ bool SSLContextService::addP12CertificateToSSLContext(SSL_CTX* ctx) const {
       core::logging::LOG_ERROR(logger_) << "Failed to set additional certificate from " << certificate_ << ", " << getLatestOpenSSLErrorString();
       return false;
     }
-    cacert.release();  // a successful SSL_CTX_add_extra_chain_cert() takes ownership of cacert
+    // a successful SSL_CTX_add_extra_chain_cert() takes ownership of cacert
+    cacert.release();  // NOLINT(bugprone-unused-return-value)
```
szaszm, line 217: Would casting the expression to void hide the warning? It might be an alternative to NOLINT.

extensions/procfs/CpuStat.h
```diff
@@ -34,7 +34,12 @@ class CpuStatData {
  public:
   static std::optional<CpuStatData> parseCpuStatLine(std::istream& iss);
 
-  auto operator<=>(const CpuStatData& rhs) const = default;
+  bool operator<=(const CpuStatData& rhs) const;
+  bool operator>(const CpuStatData& rhs) const;
+  bool operator>=(const CpuStatData& rhs) const;
+  bool operator<(const CpuStatData& rhs) const;
+  bool operator==(const CpuStatData& rhs) const;
+  bool operator!=(const CpuStatData& rhs) const;
```
szaszm, line 42: Why this over the spaceship operator?

extensions/openwsman/processors/SourceInitiatedSubscriptionListener.cpp
```diff
@@ -887,7 +887,7 @@ void SourceInitiatedSubscriptionListener::onSchedule(const std::shared_ptr<core:
 }
 
 void SourceInitiatedSubscriptionListener::notifyStop() {
-  server_.release();
+  server_.release();  // NOLINT(bugprone-unused-return-value)
```
szaszm, line 890: Isn't this a real bug? I think it should be reset, not release.

.clang-tidy
```diff
@@ -0,0 +1,2 @@
+Checks: 'boost*,bugprone*,cert*,portability*,-clang-analyzer-optin.cplusplus.VirtualCall,-bugprone-narrowing-conversions,-bugprone-implicit-widening-of-multiplication-result,-bugprone-macro-parentheses,-bugprone-easily-swappable-parameters,-bugprone-exception-escape,-cert-err58-cpp,-cert-err33-c'
```
szaszm, line 1: A few of the disabled ones make sense IMO. This is the list I'm using with CLion:
```
-*
bugprone-argument-comment
bugprone-assert-side-effect
bugprone-bad-signal-to-kill-thread
bugprone-branch-clone
bugprone-copy-constructor-init
bugprone-dangling-handle
bugprone-dynamic-static-initializers
bugprone-fold-init-type
bugprone-forward-declaration-namespace
bugprone-forwarding-reference-overload
bugprone-inaccurate-erase
bugprone-incorrect-roundings
bugprone-integer-division
bugprone-lambda-function-name
bugprone-macro-parentheses
bugprone-macro-repeated-side-effects
bugprone-misplaced-operator-in-strlen-in-alloc
bugprone-misplaced-pointer-arithmetic-in-alloc
bugprone-misplaced-widening-cast
bugprone-move-forwarding-reference
bugprone-multiple-statement-macro
bugprone-no-escape
bugprone-not-null-terminated-result
bugprone-parent-virtual-call
bugprone-posix-return
bugprone-reserved-identifier
bugprone-sizeof-container
bugprone-sizeof-expression
bugprone-spuriously-wake-up-functions
bugprone-string-constructor
bugprone-string-integer-assignment
bugprone-string-literal-with-embedded-nul
bugprone-suspicious-enum-usage
bugprone-suspicious-include
bugprone-suspicious-memory-comparison
bugprone-suspicious-memset-usage
bugprone-suspicious-missing-comma
bugprone-suspicious-semicolon
bugprone-suspicious-string-compare
bugprone-swapped-arguments
bugprone-terminating-continue
bugprone-throw-keyword-missing
bugprone-too-small-loop-variable
bugprone-undefined-memory-manipulation
bugprone-undelegated-constructor
bugprone-unhandled-self-assignment
bugprone-unused-raii
bugprone-unused-return-value
bugprone-use-after-move
bugprone-virtual-near-miss
cert-dcl21-cpp
cert-dcl58-cpp
cert-err34-c
cert-err52-cpp
cert-err58-cpp
cert-err60-cpp
cert-flp30-c
cert-msc50-cpp
cert-msc51-cpp
cert-str34-c
cppcoreguidelines-interfaces-global-init
cppcoreguidelines-narrowing-conversions
cppcoreguidelines-pro-type-member-init
cppcoreguidelines-pro-type-static-cast-downcast
cppcoreguidelines-slicing
google-default-arguments
google-explicit-constructor
google-runtime-operator
hicpp-exception-baseclass
hicpp-multiway-paths-covered
misc-misplaced-const
misc-new-delete-overloads
misc-no-recursion
misc-non-copyable-objects
misc-throw-by-value-catch-by-reference
misc-unconventional-assign-operator
misc-uniqueptr-reset-release
modernize-avoid-bind
modernize-concat-nested-namespaces
modernize-deprecated-headers
modernize-deprecated-ios-base-aliases
modernize-loop-convert
modernize-make-shared
modernize-make-unique
modernize-pass-by-value
modernize-raw-string-literal
modernize-redundant-void-arg
modernize-replace-auto-ptr
modernize-replace-disallow-copy-and-assign-macro
modernize-replace-random-shuffle
modernize-return-braced-init-list
modernize-shrink-to-fit
modernize-unary-static-assert
modernize-use-auto
modernize-use-bool-literals
modernize-use-emplace
modernize-use-equals-default
modernize-use-equals-delete
modernize-use-nodiscard
modernize-use-noexcept
modernize-use-nullptr
modernize-use-override
modernize-use-transparent-functors
modernize-use-uncaught-exceptions
mpi-buffer-deref
mpi-type-mismatch
openmp-use-default-none
performance-faster-string-find
performance-for-range-copy
performance-implicit-conversion-in-loop
performance-inefficient-algorithm
performance-inefficient-string-concatenation
performance-inefficient-vector-operation
performance-move-const-arg
performance-move-constructor-init
performance-no-automatic-move
performance-noexcept-move-constructor
performance-trivially-destructible
performance-type-promotion-in-math-fn
performance-unnecessary-copy-initialization
performance-unnecessary-value-param
portability-simd-intrinsics
readability-avoid-const-params-in-decls
readability-const-return-type
readability-container-size-empty
readability-convert-member-functions-to-static
readability-delete-null-pointer
readability-deleted-default
readability-inconsistent-declaration-parameter-name
readability-make-member-function-const
readability-misleading-indentation
readability-misplaced-array-index
readability-non-const-parameter
readability-redundant-control-flow
readability-redundant-declaration
readability-redundant-function-ptr-dereference
readability-redundant-smartptr-get
readability-redundant-string-cstr
readability-redundant-string-init
readability-simplify-subscript-expr
readability-static-accessed-through-instance
readability-static-definition-in-anonymous-namespace
readability-string-compare
readability-uniqueptr-delete-release
readability-use-anyofallof
```

thirdparty/rapidjson-48fbd8cd202ca54031fe799db2ad44ffa8e77c13/include/rapidjson/document.h
lordgamez: Rapidjson is set to be a system include so it should be exluded by clang-tidy. It does work for every other thirdparty, but somehow when a file uses this inlude clang-tidy still throws a warning about this. Unfortunately I could not find another way of excluding it and it may be due to a clang-tidy-14 bug because when I checked the same files that had this error with clang-tidy-10 I did not get any warnings.

libminifi/test/unit/SocketTests.cpp
lordgamez: I don't know either why it gives a warning, it seems okay to me too. I suspected that maybe the TEST_CASE macro does something I'm not aware of, but as it looked okay I only put the NOLINT comment there.

extensions/procfs/CpuStat.h
lordgamez: Clang compilation failed with a warning on the spaceship operator in this particular case.
```
/home/ggyimesi/projects/nifi-minifi-cpp-fork/extensions/procfs/CpuStat.h:37:8: error: explicitly defaulted three-way comparison operator is implicitly deleted [-Werror,-Wdefaulted-function-deleted]
  auto operator<=>(const CpuStatData& rhs) const = default;
       ^
/home/ggyimesi/projects/nifi-minifi-cpp-fork/extensions/procfs/CpuStat.h:57:23: note: defaulted 'operator<=>' is implicitly deleted because there is no viable comparison function for member 'user_'
  SystemClockDuration user_;
```

I'm not sure why it can't use it with the SystemClockDuration type.

.clang-tidy
lordgamez: The `cert-err58-cpp` check was disabled because it gave a warning for all `REGISTER_RESOURCE` macro usages and for all initialization of static processor properties (`initialization of '<Property or Processor_registrar>' with static storage duration may throw an exception that cannot be caught [cert-err58-cpp]`) and it didn't seem viable to change those.

The `bugprone-macro-parentheses` check was disabled because it gave a warning a huge number of times without any indication of the filename or the line where it was found:
```
warning: macro replacement list should be enclosed in parentheses [bugprone-macro-parentheses]
```

As for the other checks I created a seperate Jira ticket [MINIFICPP-1833](https://issues.apache.org/jira/browse/MINIFICPP-1833) to evaluate each check group to see which ones should be integrated and fix the existing issues before adding them to the CI. This jira ticket can still be broken up further if needed, because I think it can effect too many files at once to fix all the issues. But this list is quite useful as integration candidates, thanks!

libminifi/src/controllers/SSLContextService.cpp
lordgamez: Good idea, updated in 3b4102f32349a842f90c11b08189d9e79411eca3

extensions/openwsman/processors/SourceInitiatedSubscriptionListener.cpp
lordgamez: I think this is, good catch, updated in 3b4102f32349a842f90c11b08189d9e79411eca3

extensions/procfs/CpuStat.h
szaszm: This is a clang (or rather libc++) bug. I would rather have one manual spaceship operator implementation that lexicographically compares each member instead of writing all six manually, without any kind of strict ordering.

libminifi/test/unit/SocketTests.cpp
szaszm: If we get these false positives, I'd rather disable the check. This is clearly correct usage, and I don't want to have NOLINT everywhere.

lordgamez: It would be better, but this false positive only occurs in test code so I would not disable it as it can catch some real problems in production code

extensions/procfs/CpuStat.h
lordgamez: I agree, but if werror is turned on this fails the compilation. What would you suggest?

libminifi/test/unit/SocketTests.cpp
szaszm: OK, fair point.

extensions/libarchive/FocusArchiveEntry.cpp
```diff
@@ -176,15 +176,13 @@ int64_t FocusArchiveEntry::ReadCallback::operator()(const std::shared_ptr<io::Ba
   archive_read_support_filter_all(inputArchive);
 
   // Read each item in the archive
-  int res;
-
-  if ((res = archive_read_open(inputArchive, &data, ok_cb, read_cb, ok_cb))) {
+  if ((archive_read_open(inputArchive, &data, ok_cb, read_cb, ok_cb))) {
```
adam-markovics, line 179: Is double parentheses necessary?

extensions/procfs/CpuStat.h
adam-markovics: Does it even make sense to say one CpuStatData is greater than another? If it's only for map storage, equality operator and hash would be enough, and turning the map into an unordered one.

szaszm: I suggest using memberwise lexicographical comparison in the simplest possible way. AND'ing together the comparison of all members is not consistent: there can be values that are both less and greater than each other, depending on the expression used.

example comparator: https://godbolt.org/z/nz553fbxG

szaszm: I had some success with a simpler workaround: just define operator<=> when using libc++.
See the branch on my fork: https://github.com/szaszm/nifi-minifi-cpp/tree/MINIFICPP-1832
Take the last commit if you like it.

lordgamez: Thanks for the workaround! I added your commit to the PR 7c6ee4bbb1d0e7319821d4b5957fc41493b5f3b2

libminifi/src/FlowControlProtocol.cpp
```diff
@@ -280,11 +280,11 @@ int FlowControlProtocol::sendReportReq() {
   data = this->encode(data, hdr.msgType);
   data = this->encode(data, hdr.seqNumber);
   data = this->encode(data, hdr.status);
-  data = this->encode(data, hdr.payloadLen);
+  this->encode(data, hdr.payloadLen);
```
fgerlits, line 283: I think this should not have been changed: now `FLOW_YML_NAME` starts writing the buffer at the same point as where `hdr.payloadLen` started, overwriting it.

martinzink (approved):

szaszm (approved):

PR #1334, opened 2022-05-16
MINIFICPP-1803 Do not instantiate all classes at heartbeat

fgerlits:
https://issues.apache.org/jira/browse/MINIFICPP-1803

Processor and ControllerService properties, relationships and other attributes are now static, and can be read without instantiating the class.

There are 4 cases, represented by the new `ResourceType` enum:
* `Processor`: a subclass of `Processor`, used to be registered using `REGISTER_RESOURCE[_AS]`
* `ControllerService`: a subclass of `ControllerService`, used to be registered using `REGISTER_RESOURCE[_AS]`
* `InternalResource`: a subclass of either `Processor` or `ControllerService`, used to be registered using `REGISTER_INTERNAL_RESOUCE[_AS]`
* `DescriptionOnly`: used to be registered using `REGISTER_RESOURCE[_AS]`, but not a subclass of either `Processor` or `ControllerService`.

Ideally, `properties()` and `relationships()` should be `constexpr` variables instead of functions, too.  This will be done later as part of https://issues.apache.org/jira/browse/MINIFICPP-1825.

The manifest has mostly stayed identical, with these differences:

* minifi-coap:
  * CoapConnectorService is no longer included in the manifest.  Since this controller service is marked as 'internal resource', I think this is correct.

* minifi-jni:
  * ExecuteJavaProcessor was not included in the manifest, but now it is.  I don't know why it wasn't included earlier.
  * The descriptions of ExecuteJavaControllerService and JavaControllerService are slightly different: I have deleted a space from the end.

* minifi-mqtt:
  * ConvertHeartbeat, ConvertJSONAck, ConvertUpdate used to be included in the manifest, and they aren't any longer.  Since these processors are marked as 'internal resource', I think this is correct.

* minifi-script-extensions:
  * The ReloadOnScriptChange property was not included in the ExecutePythonProcessor section of the manifest.  This was a bug, fixed now.

* minifi-sensors:
  * This extension was missing from the manifest on main, not sure why.

* minifi-standard-processors:
  * The description of ExecuteProcess is slightly different: I have added 2 missing spaces.

* minifi-system:
  * The Battery Status Discharge property of LinuxPowerManagerService was missing from the old manifest.  This was a bug, fixed now.
  * The description of NetworkPrioritizerService is slightly diferent: fixed typo ouput -> output.

* minifi-test-processors:
  * The description of KamikazeProcessor is slightly diferent: fixed typo configration -> configuration.
  * Not sure if we should include this artifact in the manifest, but we did, so we still do.

---

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/include/agent/agent_docs.h
```diff
@@ -13,53 +13,117 @@
...
-    const std::map<std::string, std::string> &extensions = getDescriptions();
-    auto iff = extensions.find(feature);
-    if (iff != extensions.end()) {
-      value = iff->second;
-      return true;
-    } else {
-      return false;
+  static bool getDescription(const std::string &feature, std::string &value);
+
+  template<typename Class, ResourceType Type>
```
szaszm, line 82: The ResourceType could be autodetected IMO. Included some untested detection logic that will probably need some changes, but should at least demonstrate the idea:
```suggestion
  namespace detail {
  template<typename T>
  constexpr bool is_array = false;
  template<typename ValueType, size_t Size>
  constexpr bool is_array<std::array<ValueType, Size>> = true;
  template<typename T, typename ValueType>
  concept array_of = is_array<T> && std::is_same_v<typename T::value_type, ValueType>;
  
  template<typename T>
  concept has_description = requires() {
    { T::Description } -> const char*;
  };
  
  template<typename T>
  concept has_metadata = has_description<T> && requires() {
    { T::properties() } -> array_of<core::Property>;
    { T::SupportsDynamicProperties } -> bool;
  };
  
  template<typename T>
  concept processor = has_description<T> && std::derived_from<T, core::Processor>;
  
  template<typename T>
  concept controller_service = has_description<T> && std::derived_from<T, core::controller::ControllerService>;
  
  struct unknown_resource_type : std::exception {};
  
  template<typename T>
  constexpr ResourceType resource_type_of() {
    if constexpr (processor<T>) {
      return ResourceType::Processor;
    } else if constexpr (controller_service<T>) {
      return ResourceType::ControllerService;
    } else if constexpr (has_metadata<T>) {
      return ResourceType::InternalResource;
    } else if constepxr (has_description<T>) {
      return ResourceType::DescriptionOnly;
    } else {
      throw unknown_resource_type{};
    }
  }
  }  // namespace detail

  template<typename Class, ResourceType Type>
```

extensions/aws/controllerservices/AWSCredentialsService.h
```diff
@@ -53,6 +48,25 @@ class AWSCredentialsService : public core::controller::ControllerService {
...
+    return std::array{
+      UseDefaultCredentials,
+      AccessKey,
+      SecretKey,
+      CredentialsFile
+    };
+  }
+
+  EXTENSIONAPI static constexpr bool SupportsDynamicProperties = false;
+  EXTENSIONAPI static constexpr bool SupportsDynamicRelationships = false;
```
szaszm, line 67: There should be no need for `SupportsDynamicRelationships` for anything other than Processors. Services can't have relationships.

szaszm (commented):
I really like the refactoring that you've done here. I have a few suggestions, but I didn't review all of the extensions yet.

extensions/http-curl/protocols/RESTSender.cpp
```diff
@@ -172,10 +167,6 @@ C2Payload RESTSender::sendPayload(const std::string url, const Direction directi
   }
 }
 
-REGISTER_RESOURCE(RESTSender, "Encapsulates the restful protocol that is built upon C2Protocol.");
+REGISTER_RESOURCE(RESTSender, DescriptionOnly);
```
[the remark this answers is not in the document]
fgerlits: This is a bit of a mess, but it was already like that before this PR:
- `InternalResource` means it inherits from either `Processor` or `ControllerService`, but should not be included in the manifest.  These do not have a `Description` field.
- `DescriptionOnly` means it does not inherit from `Processor` or `ControllerService`, so it will not be included in the manifest, but it does have a `Description` field (which I don't think we use anywhere, but I'm not sure, so I did not want to remove it).

If these `Description`s are in fact not needed, then we could simplify this to:
- if it has a `Description`, then include it in the manifest, detect whether it is a `Processor` or `ControllerService`, and throw a compile error if it's neither;
- if it doesn't have a `Description`, then only register it in the class loader, and skip it in `agent_docs.h` etc completely.

Should we do this?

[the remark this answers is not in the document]
szaszm: I didn't realize there is a third variable, whether it's included in the manifest. With that in mind, I wouldn't do complete autodetection, but rather keep the explicit declaration of "internalness". Adding an "isInternal" member could work, but leaving it as it is is fine, too.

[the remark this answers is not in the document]
fgerlits: Let's leave it as it is for now; we can revisit this and clean it up later.

extensions/aws/controllerservices/AWSCredentialsService.h
fgerlits: Yeah, it didn't make much sense, but I didn't want to change the existing logic.

Currently, every controller service in the manifest has a `"supportsDynamicRelationships": false` line; should I remove that, or leave it hard-coded to false?

extensions/azure/processors/PutAzureBlobStorage.cpp
```diff
@@ -22,41 +22,14 @@
 
 #include "core/ProcessContext.h"
 #include "core/ProcessSession.h"
+#include "core/PropertyBuilder.h"
```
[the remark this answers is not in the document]
fgerlits: I left these in on purpose, as I thought I'd move the static definitions back from the `...StaticDefinitions.cpp` soon.  Since it looks like they will stay like that for a while, I have removed these includes in 416f85452eb9a640de42bc52c9619d459bc5e5cf

extensions/aws/controllerservices/AWSCredentialsService.h
fgerlits: Done in 3c2574ae73f8020181503d669161607601f861cf

The manifest will still have a `"supportsDynamicRelationships": false` line in each controller service block, as it did before this PR, I hope that's OK.  We can remove it later, after checking with C2 developers.

libminifi/include/controllers/keyvalue/AbstractAutoPersistingKeyValueStoreService.h
```diff
@@ -30,23 +29,17 @@
...
 class AbstractAutoPersistingKeyValueStoreService : virtual public PersistableKeyValueStoreService {
  public:
   explicit AbstractAutoPersistingKeyValueStoreService(const std::string& name, const utils::Identifier& uuid = {});
 
   ~AbstractAutoPersistingKeyValueStoreService() override;
 
-  MINIFIAPI static core::Property AlwaysPersist;
-  MINIFIAPI static core::Property AutoPersistenceInterval;
+  static constexpr const char* AlwaysPersistPropertyName = "Always Persist";
+  static constexpr const char* AutoPersistenceIntervalPropertyName = "Auto Persistence Interval";
```
adamdebreceni, line 41: what is the reason behind this change?

libminifi/src/core/Connectable.cpp
```diff
@@ -85,7 +69,7 @@ bool Connectable::isSupportedRelationship(const core::Relationship &relationship
   }
 }
 
-bool Connectable::setAutoTerminatedRelationships(const std::set<Relationship> &relationships) {
+bool Connectable::setAutoTerminatedRelationships(const std::vector<Relationship>& relationships) {
```
adamdebreceni, line 72: shouldn't we turn this into a template as well? (as `setSupportedProperties`)

libminifi/include/controllers/keyvalue/AbstractAutoPersistingKeyValueStoreService.h
fgerlits: This class has two derived classes, in two different libraries: `UnorderedMapPersistableKeyValueStoreService` in `libminifi` and `RocksDbPersistableKeyValueStoreService` in `rocksdb-repos`.  The static definitions of each inheritance hierarchy need to be in a single translation unit to ensure the correct order of initialization at startup, but that is not possible if they are in different libraries.  So instead of the two derived classes inheriting these two properties, they define them separately, but they both use these constants to indicate that they are really the same properties.

libminifi/src/core/Connectable.cpp
fgerlits: I changed `setAutoTerminatedRelationships` to a template in 529f7014bdaf548ebfd5198cd78cdd5ff9b5b136

but then I realized it doesn't need to be a template, and changed `setSupportedProperties`, `setSupportedRelationships` and `setAutoTerminatedRelationships` back to functions, with a `gsl::span` argument type, in cb2e0defeb0d19539590fe82cd6eb69e466ca2dc

let me know if you don't like cb2e0defeb0d19539590fe82cd6eb69e466ca2dc and want it changed back

adamdebreceni: I'm fine with `gsl::span` my only concern is from the caller's side we injected a number of `std::array` calls where previously we used an initializer list, do we need them or can we drop them now that the methods are not templates?

fgerlits: Unfortunately, initializer lists don't convert implicitly to `gsl::span`.  Nevertheless, we used to create a `std::vector` (with elements on the heap) and now we only create a `std::array` (with elements on the stack).

run_clang_tidy.sh
```diff
@@ -6,6 +6,9 @@ exit_code=0
 FILES=$1
 
 for changed_file in ${FILES}; do
+  if [[ "${changed_file}" =~ /pdh/ || "${changed_file}" =~ /windows-event-log/ ]]; then
+    continue
+  fi
   if [[ "${changed_file}" == *.cpp ]]; then
     clang-tidy-14 -warnings-as-errors=* -quiet -p build "${changed_file}"
```
fgerlits, line 13: I have added this as a quick-and dirty way of skipping the Windows-only extensions.  @lordgamez please let me know if you can think of a better way.

run_clang_tidy.sh
```diff
@@ -5,7 +5,14 @@ set -uo pipefail
...
+EXCLUDED_EXTENSIONS=("pdh", "windows-event-log", "tensorflow")
+
 for changed_file in ${FILES}; do
+  for extension in "${EXCLUDED_EXTENSIONS[@]}"; do
+    if [[ "${changed_file}" =~ extensions/${extension}/ ]]; then
+      continue 2
+    fi
+  done
   if [[ "${changed_file}" == *.cpp ]]; then
     clang-tidy-14 -warnings-as-errors=* -quiet -p build "${changed_file}"
```
fgerlits, line 17: I have added this as a quick-and dirty way of skipping the Windows-only extensions.  @lordgamez please let me know if you can think of a better way.

libminifi/src/core/Connectable.cpp
szaszm: You can overload them to accept both gsl::span and std::array, if you want.

fgerlits: I prefer `some_function(std::array{foo, bar})` to `some_function({foo, bar})`; I think it's clearer and it's also slightly more efficient.  The only downside is a few more characters, but I don't think that's a problem.

adamdebreceni (approved):

extensions/standard-processors/processors/ExtractText.h
```diff
@@ -36,26 +36,43 @@ namespace minifi {
 namespace processors {
 
 //! ExtractText Class
-class ExtractText : public core::Processor {
+class
+ExtractText : public core::Processor {
```
szaszm, line 40: I think this strange formatting was done by mistake.

fgerlits: yes it was, thanks for catching it -- fixed in f4f8c9a11f2fd379c3b5593441619ad547736246

extensions/script/ExecuteScript.h
```diff
@@ -137,13 +137,35 @@ class ExecuteScript : public core::Processor {
...
+  EXTENSIONAPI static constexpr const char* Description = "Executes a script given the flow file and a process session. "
+      "The script is responsible for handling the incoming flow file (transfer to SUCCESS or remove, e.g.) as well as "
+      "any flow files created by the script. If the handling is incomplete or incorrect, the session will be rolled back.Scripts must define an onTrigger function which accepts NiFi Context "
+      "and Property objects. For efficiency, scripts are executed once when the processor is run, then the onTrigger method is called for each incoming flowfile. This enables scripts to keep state "
+      "if they wish, although there will be a script context per concurrent task of the processor. In order to, e.g., compute an arithmetic sum based on incoming flow file information, set the "
+      "concurrent tasks to 1.";
+
+  EXTENSIONAPI static core::Property ScriptEngine;
+  EXTENSIONAPI static core::Property ScriptFile;
+  EXTENSIONAPI static core::Property ScriptBody;
```
martinzink, line 149: I noticed you fixed the constness for the properties you touched. Kudos for that. :+1: 
I think this could be made const aswell.

szaszm (approved):

martinzink (approved):

lordgamez (approved):

PR #1340, opened 2022-05-26
MINIFICPP-1829 Export metrics for use with Prometheus

lordgamez:
This feature enables the agent to expose metrics through a Prometheus metrics publisher to be scraped when configured

- Metric loading is extracted from C2 client to be reused
- C2 metrics are tested and fixed where needed
- C2 update configuration is tested to verify metric reloading
- Metrics are extended with calculateMetrics method for Prometheus or any future monitoring tool to be published with new Prometheus extension
- prometheus-cpp client is added for exposing metrics to Prometheus
- minifi.properties is extended to specify metrics publisher and published metric nodes
- Fixed serialization on C2 of empty QueueMetrics

https://issues.apache.org/jira/browse/MINIFICPP-1829
https://issues.apache.org/jira/browse/MINIFICPP-1849

----------------------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


Windows.md
```diff
@@ -67,6 +67,7 @@ After the build directory it will take optional parameters modifying the CMake c
 | /L | Enables Linter |
 | /O | Enables OpenCV |
 | /PDH | Enables Performance Monitor |
+| /P | Enables Prometheus |
```
adamdebreceni, line 70: shouldn't this be `PR`?

extensions/http-curl/tests/C2MetricsTest.cpp
```diff
@@ -0,0 +1,209 @@
...
+    return runtime_metrics.HasMember("deviceInfo") &&
+      runtime_metrics.HasMember("flowInfo") &&
+      runtime_metrics["flowInfo"].HasMember("versionedFlowSnapshotURI") &&
+      runtime_metrics["flowInfo"].HasMember("queues") &&
+      runtime_metrics["flowInfo"].HasMember("components") &&
+      runtime_metrics["flowInfo"]["queues"].HasMember("2438e3c8-015a-1000-79ca-83af40ec1997") &&
+      runtime_metrics["flowInfo"]["components"].HasMember("FlowController") &&
+      runtime_metrics["flowInfo"]["components"].HasMember("GetTCP") &&
+      runtime_metrics["flowInfo"]["components"].HasMember("LogAttribute");
+  }
```
adamdebreceni, line 136: we have a `verifyJSON` test util, could that be used here? (maybe with some enhancements? e.g. "__any__" matching all values if we only care about field existence)

extensions/prometheus/PrometheusExposerWrapper.cpp
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "PrometheusExposerWrapper.h"
+
+namespace org::apache::nifi::minifi::extensions::prometheus {
+
+PrometheusExposerWrapper::PrometheusExposerWrapper(uint32_t port)
+    : exposer_(std::to_string(port)) {
+  logger_->log_info("Started Prometheus metrics publisher on port %u", port);
```
adamdebreceni, line 23: `PRIu32` might be the appropriate format here

libminifi/include/FlowController.h
```diff
@@ -252,8 +255,7 @@ class FlowController : public core::controller::ForwardingControllerServiceProvi
   // Thread pool for schedulers
   utils::ThreadPool<utils::TaskRescheduleInfo> thread_pool_;
   std::map<utils::Identifier, std::unique_ptr<state::ProcessorController>> processor_to_controller_;
+  std::unique_ptr<state::MetricsPublisher> metrics_publisher_;
```
szaszm, line 258: The flow controller shouldn't own this service, or even depend on it. Each entity that's capable of exporting metrics should expose that on its public interface. The flow controller may provide some aggregation into a single structure if necessary, but it shouldn't do much else.
Let's have the MetricsPublisher depend on a metric source callback, that's bound to this aggregation member function of the flow controller, without each depending on the other.

Windows.md
lordgamez: Good catch, fixed in 4f53b2c73daee9f416fab8ad038404064002d0e8

extensions/http-curl/tests/C2MetricsTest.cpp
lordgamez: I checked it, but I think in this case it's not really viable. There are a lot of environment or flow state specific metrics that cannot be checked for and exact expected json and I think it's better to check on these specific fields separately.

libminifi/src/core/state/nodes/QueueMetrics.cpp
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "core/state/nodes/QueueMetrics.h"
+#include "core/Resource.h"
+
+namespace org::apache::nifi::minifi::state::response {
+
+REGISTER_RESOURCE(QueueMetrics, "Node part of an AST that defines queue metric information");
```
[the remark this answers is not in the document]
lordgamez: I think it's a legacy term used in all the metric node descriptions, not sure why, as in C2 the metric nodes are represented in kind of a tree format, but not an AST. I changed it to be a more generic "Metric node" in the descriptions in 4f53b2c73daee9f416fab8ad038404064002d0e8.

docker/test/integration/minifi/core/DockerTestCluster.py
```diff
@@ -283,3 +282,79 @@ def write_content_to_container(self, content, dst):
...
             with open(os.path.join(td, 'content.tar'), 'rb') as data:
                 return container.put_archive(os.path.dirname(dst_path), data.read())
+
+    def check_metric_class_on_prometheus(self, metric_class, timeout_seconds):
+        start_time = time.perf_counter()
+        while (time.perf_counter() - start_time) < timeout_seconds:
+            if self.verify_metric_class(metric_class):
+                return True
+            time.sleep(1)
+        return False
```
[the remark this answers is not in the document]
lordgamez: Good point, changed it in 4f53b2c73daee9f416fab8ad038404064002d0e8

METRICS.md
```diff
@@ -0,0 +1,153 @@
...
+	# in minifi.properties
+
+	nifi.metrics.publisher.class=PrometheusMetricsPublisher
+
+Currently PrometheusMetricsPublisher is the only available publisher in MiNiFi C++ which publishes metrics to a Prometheus server.
+To use the publisher a port should also be configured where the metrics will be available to be scraped through:
+
+	# in minifi.properties
+
+	nifi.metrics.publisher.port=9936
```
szaszm, line 47: This port should be a property of PrometheusMetricsPublisher, not all publishers in general.

extensions/prometheus/PublishedMetricGaugeCollection.cpp
```diff
@@ -0,0 +1,49 @@
...
+    ::prometheus::ClientMetric client_metric;
+    client_metric.label = ranges::views::transform(metric.labels, [](auto&& kvp) { return ::prometheus::ClientMetric::Label{kvp.first, kvp.second}; })
+      | ranges::to<std::vector<::prometheus::ClientMetric::Label>>;
+    client_metric.gauge = ::prometheus::ClientMetric::Gauge{metric.value};
+    collection.push_back({
+      .name = metric.name,
+      .help = "",
+      .type = ::prometheus::MetricType::Gauge,
+      .metric = { client_metric }
+    });
```
szaszm, line 44: I think it would help if metric names were prefixed with "minifi_" before serving the data to prometheus. It just makes finding them much easier if multiple programs with different metrics are used.

image

lordgamez: Most guidelines discouraged prefixes for metric names and I didn't want to use it because Prometheus already adds a `job` label for the metric with the `minifi` value that can be filtered for. But I checked the [Prometheus best practices](https://prometheus.io/docs/practices/naming/) and it says that the metrics should have an application name as a prefix to mark the domain (or sometimes it's not an application name but something general like `http`). In our case using `minifi_` prefix is a good idea I added it in 96431d07a867a26723d9d6c2cf71edc1d1dd0b17

lordgamez:
Removed the QueueMetrics class from default configuration in df6ae67603f4e75ecc45ab87d2acd2a401ff9cf6 due to the bug in C2 serialization described in https://issues.apache.org/jira/browse/MINIFICPP-1849

METRICS.md
lordgamez: If our goal is to support more ports for future publishers, wouldn't that be easier to use comma separated values for `nifi.metrics.publisher.port` for those publishers?

szaszm: It begs the question, which of those values belong to which publisher? What if a publisher doesn't use the network, or not through TCP/UDP? Or it periodically posts via HTTP, in which the URL already has the port?

We could declare a global port setting and have the publishers read from it if they want to, but it fundamentally belongs to the component that actually uses that port, if they need such a setting.

lordgamez: Updated in 2f43e4e6818e72cc2e6cd5a0e22d4f96962db9eb

lordgamez:
Added fix for [MINIFICPP-1849](https://issues.apache.org/jira/browse/MINIFICPP-1849) for serialization of empty QueueMetrics on C2 in 9c0f374283b8507f9940690547d267c188de0330. Readded QueueMetrics to the default configuration.

METRICS.md
```diff
@@ -0,0 +1,155 @@
...
+### QueueMetrics
+
+QueueMetrics is a system level metric that reports queue metrics for every connection in the flow.
+
+| Metric name          | Labels                                         | Description                                |
+|----------------------|------------------------------------------------|--------------------------------------------|
+| queue_data_size      | metric_class, connection_uuid, connection_name | Max queue size to apply back pressure      |
+| queue_data_size_max  | metric_class, connection_uuid, connection_name | Max queue data size to apply back pressure |
+| queue_size           | metric_class, connection_uuid, connection_name | Current queue size                         |
+| queue_size_max       | metric_class, connection_uuid, connection_name | Current queue data size                    |
```
szaszm, line 70: I've found this guide to metric naming: https://prometheus.io/docs/practices/naming/

By the way I think the descriptions are not in the correct order here.

libminifi/include/core/state/nodes/RepositoryMetrics.h
```diff
@@ -90,15 +87,18 @@ class RepositoryMetrics : public ResponseNode {
...
 
+  std::vector<PublishedMetric> calculateMetrics() override {
+    std::vector<PublishedMetric> metrics;
+    for (const auto& [_, repo] : repositories_) {
+      metrics.push_back({"is_running", (repo->isRunning() ? 1.0 : 0.0), {{"metric_class", getName()}, {"repository_name", repo->getName()}}});
+      metrics.push_back({"is_full", (repo->isFull() ? 1.0 : 0.0), {{"metric_class", getName()}, {"repository_name", repo->getName()}}});
+      metrics.push_back({"repository_size", static_cast<double>(repo->getRepoSize()), {{"metric_class", getName()}, {"repository_name", repo->getName()}}});
+    }
+    return metrics;
+  }
```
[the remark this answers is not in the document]
lordgamez: Could you elaborate a bit more please? What do you think is currently too Prometheus specific format here that should be changed and transformed later? Also what do you mean by unifying `serialize`? The serialized nodes are not used by Prometheus only on the C2 protocol.

libminifi/include/properties/Configuration.h
```diff
@@ -156,12 +156,17 @@ class Configuration : public Properties {
 
   static constexpr const char *nifi_asset_directory = "nifi.asset.directory";
 
+  // Metrics publisher options
+  static constexpr const char *nifi_metrics_publisher_class = "nifi.metrics.publisher.class";
+  static constexpr const char *nifi_metrics_publisher_prometheus_metrics_publisher_port = "nifi.metrics.publisher.PrometheusMetricsPublisher.port";
```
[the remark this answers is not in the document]
lordgamez: The problem is that all possible configuration options sent through C2 are retrieved from here, so if we want to advertise the prometheus options as well, they should be added here.

libminifi/include/core/state/ConnectionStore.h
```diff
@@ -0,0 +1,60 @@
...
+        {{"connection_uuid", connection->getUUIDStr()}, {"connection_name", connection->getName()}, {"metric_class", metric_class}}});
+    }
+
+    return metrics;
+  }
+
+  virtual ~ConnectionStore() = default;
+
+ protected:
+  std::map<std::string, minifi::Connection*> connections_;
```
szaszm, line 57: Consider using `std::unordered_map<utils::Identifier, Connection*>`.

Untested pseudocode to help with `std::hash` specialization for `Identifier`:
```
namespace std {
template<>
struct hash<Identifier> {
  size_t operator()(const Identifier& id) const noexcept {
    constexpr int slices = sizeof(Identifier) / sizeof(size_t);
    const auto combine = [](size_t& seed, size_t new_hash) {
      // from the boost hash_combine docs
      seed ^= new_hash + 0x9e3779b9 + (seed << 6) + (seed >> 2);
    };
    const auto get_slice = [](const Identifier& id, size_t idx) -> size_t {
      // Needs access to Identifier internals
      size_t result{};
      memcpy(&result, reinterpret_cast<const unsigned char*>(&id.data_) + idx * sizeof(size_t), sizeof(size_t));
      return result;
    };
    size_t hash = get_slice(id, 0);
    for (size_t i = 1; i < slices; ++i) {
      combine(hash, get_slice(id, i));
    };
    return hash;
  }
};
}  // namespace std

File: libminifi/include/properties/Configuration.h (hunk 11 above)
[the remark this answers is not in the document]
Comment (szaszm) [reply libminifi/include/properties/Configuration.h:161]: Fair point. Let's leave it as is, and maybe think about it later, if it starts to become a problem.

File: libminifi/include/core/state/nodes/RepositoryMetrics.h (hunk 10 above)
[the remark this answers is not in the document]
Comment (szaszm) [reply libminifi/include/core/state/nodes/RepositoryMetrics.h:98]: Now each metric class needs to have two different member functions (serialize and calculateMetrics) that perform essentially the same task: collect metrics and format them in a specific way. I would prefer to separate the "collect metrics" part from the formatting, and have 1 "collect metrics" function here, and one formatting function in each of C2Client and somewhere in the prometheus extension. I didn't think about the details yet, only have this high level idea.

File: METRICS.md (hunk 9 above)
Comment (szaszm) [reply METRICS.md:70]: I realize it would be hard to have metric sink-specific names without some type metadata for metrics, and I wouldn't want to touch the C2 protocol. Let's leave it as is for now, and think about changing the prometheus metrics names later, after some metrics-related refactoring.

Could you check the descriptions though? They are mixed up.

Comment (lordgamez) [reply METRICS.md:70]: Okay, I think we would have more input and would be able to generalize these things a bit easier if we had another monitoring tool support, that would make it easier to see what needs to be more generalized.

I see there is a mixup in the descriptions I'll fix it.

File: libminifi/src/c2/C2Client.cpp [hunk 13]
```diff
@@ -206,35 +130,13 @@ void C2Client::loadC2ResponseConfiguration(const std::string &prefix) {
...
-
+        loadNodeClasses(class_definitions, new_node);
       } else {
         std::string optionName = option + "." + name;
-        auto node = loadC2ResponseConfiguration(optionName, new_node);
+        loadC2ResponseConfiguration(optionName, new_node);
       }
 
-      std::lock_guard<std::mutex> guard{metrics_mutex_};
+      // We don't need to lock here we do it in the initializeResponseNodes
```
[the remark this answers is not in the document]
lordgamez: Updated in afe6e3ec99c2c21454fd840d031677a2b948f1bd

libminifi/include/core/state/ConnectionStore.h
lordgamez: Thanks! Updated in afe6e3ec99c2c21454fd840d031677a2b948f1bd

libminifi/include/core/state/nodes/RepositoryMetrics.h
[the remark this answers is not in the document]
szaszm: > Also moving the serialization and the metrics calculation to C2Client and the metrics publisher would require us to define parts of a newly introduced metric in 3 places, which would not be really intuitive.

I was thinking about metric-independent formatting in C2Client and the prometheus extension, so new metrics require only 1 code change, not 2 or 3.

I've added this to a followup refactoring Jira issue: https://issues.apache.org/jira/browse/MINIFICPP-1869

szaszm (approved):

martinzink (approved):

extensions/prometheus/PublishedMetricGaugeCollection.cpp
```diff
@@ -0,0 +1,49 @@
...
+  for (const auto& metric : metric_->calculateMetrics()) {
+    ::prometheus::ClientMetric client_metric;
+    client_metric.label = ranges::views::transform(metric.labels, [](auto&& kvp) { return ::prometheus::ClientMetric::Label{kvp.first, kvp.second}; })
+      | ranges::to<std::vector<::prometheus::ClientMetric::Label>>;
+    client_metric.gauge = ::prometheus::ClientMetric::Gauge{metric.value};
+    collection.push_back({
+      .name = "minifi_" + metric.name,
+      .help = "",
+      .type = ::prometheus::MetricType::Gauge,
+      .metric = { client_metric }
```
adamdebreceni, line 43: we could move `client_metric` here

libminifi/include/core/state/nodes/MetricsBase.h
```diff
@@ -108,27 +104,33 @@ class ObjectNode : public ResponseNode {
     nodes_.push_back(node);
   }
 
-  virtual std::string getName() const {
+  const std::vector<std::shared_ptr<ResponseNode>>& get_child_nodes() const {
```
adamdebreceni, line 107: I see that the casing is mixed in this class, IMO we should go with camelCase for new methods

extensions/prometheus/PublishedMetricGaugeCollection.cpp
lordgamez: Updated in 40b469f11033f229a9237afd0b3ba397cb2e316d

libminifi/include/core/state/nodes/MetricsBase.h
lordgamez: Removed `get_child_nodes` in 40b469f11033f229a9237afd0b3ba397cb2e316d because it is not used anymore.

libminifi/include/utils/Id.h
```diff
@@ -132,3 +134,26 @@ class NonRepeatingStringGenerator {
...
+    const auto get_slice = [](const org::apache::nifi::minifi::utils::Identifier& id, size_t idx) -> size_t {
+      size_t result{};
+      memcpy(&result, reinterpret_cast<const unsigned char*>(&id.data_) + idx * sizeof(size_t), sizeof(size_t));
+      return result;
+    };
+    size_t hash = get_slice(id, 0);
+    for (size_t i = 1; i < slices; ++i) {
+      combine(hash, get_slice(id, i));
+    }
+    return hash;
```
adamdebreceni, line 156: we could use `gsl::span` here, it would also verify convertibility
```suggestion
    for (size_t slice : gsl::span(id.data_).as_span<const size_t>()) {
      combine(hash, slice);
    }
```

szaszm: I don't understand what kind of convertibility you mean here. Using `gsl::span::as_span` is just a reinterpret_cast internally, which usually results in UB because of strict aliasing violation. It's fine to use it to view the bytes of the object representation, and a handful of other cases, but usually `memcpy` is the safer option.

Another alternative that I didn't think of at the time was to `const auto chunks = std::bit_cast<std::array<size_t, slices>>(id.data_);`, and iterate over the resulting array. This uses `memcpy` internally as well.

adamdebreceni: I was thinking about `sizeof(Identifier) % sizeof(size_t) == 0`, a simple `static_assert` will do

libminifi/src/FlowController.cpp
```diff
@@ -150,6 +154,7 @@ bool FlowController::applyConfiguration(const std::string &source, const std::st
     this->root_ = std::move(prevRoot);
```
adamdebreceni, line 154: won't unlocking at line 140 make it possible for this line to execute without a valid lock (due to an exception after the unlock)?

lordgamez: I think in theory you are right, it could be unlocked if the exception happens after the unlock. Although the exception check is mostly meant for the loading of the new configuration, because after the unlock we only write some logs and set a few configuration values that shouldn't really throw any exceptions. Just in case I added the additional lock in 61844c0a17630875f7db41fcd329bedae7ef9df6

lordgamez: After realizing that it could throw an exception if we lock multiple times I reverted this change. It shouldn't throw when it is unlocked so it should be okay.

adamdebreceni: what was the problem that required manual locking/unlocking?

lordgamez: As far as I remember it could cause a deadlock when the updating_ flag is checked from another thread and it's already set to false trying to acquire the lock afterwards, but it's still locked, so we have to unlock it before setting the update_ flag back to false.

adamdebreceni: we should restructure this method to avoid manual locking, the whole lock protected region could get its own block, we should move the logging block out of the try block to the very end of this method

```
  updating_ = true;

  {
    std::lock_guard<std::recursive_mutex> flow_lock(mutex_);
    stop();
    unload();
    controller_map_->clear();
    clearResponseNodes();
    if (metrics_publisher_) {
      metrics_publisher_->clearMetricNodes();
    }
    auto prevRoot = std::move(this->root_);
    this->root_ = std::move(newRoot);
    processor_to_controller_.clear();
    initialized_ = false;
    bool started = false;
    try {
      load(std::move(root_), true);
      flow_update_ = true;
      started = start() == 0;
    } catch (...) {
      this->root_ = std::move(prevRoot);
      load(std::move(this->root_), true);
      flow_update_ = true;
    }
  }

  updating_ = false;

  if (started) {
    auto flowVersion = flow_configuration_->getFlowVersion();
    if (flowVersion) {
      logger_->log_debug("Setting flow id to %s", flowVersion->getFlowId());
      configuration_->set(Configure::nifi_c2_flow_id, flowVersion->getFlowId());
      configuration_->set(Configure::nifi_c2_flow_url, flowVersion->getFlowIdentifier()->getRegistryUrl());
    } else {
      logger_->log_debug("Invalid flow version, not setting");
    }
  }

  return started;
```

lordgamez: It could happen when `executeOnAllComponents` or `executeOnComponent` was called from a metric while updating that's why the `updating_` check was added there.

lordgamez: Good idea, updated in b53169315c1ec8d6222e8e55efab8631c75325c5

PR #1348, opened 2022-06-09
MINIFICPP-1833 - Integrate misc and modernize clang-tidy checks in CI

lordgamez:
misc and modernize category checks of clang-tidy are fixed and added to CI. A few checks are only fixed in the changed files, as fixing them in the whole codebase would have blown up the PR significantly, but can be fixed over time when changing the problematic file in a PR. These checks are:

- modernize-concat-nested-namespaces
- modernize-use-auto
- modernize-use-nullptr
- modernize-pass-by-value

https://issues.apache.org/jira/browse/MINIFICPP-1833

----------------------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/test/unit/EnvironmentUtilsTests.cpp
```diff
@@ -117,6 +117,9 @@ TEST_CASE("multithreaded environment manipulation", "[getenv][setenv][unsetenv]"
               utils::Environment::unsetEnvironmentVariable(env_name.c_str());
               break;
             }
+          default: {
+            throw std::runtime_error("Operation value must be modulo 3");
```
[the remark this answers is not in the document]
lordgamez: The idea was throwing an unhandled runtime_error which will terminate by default eventually. What would be the benefit of calling terminate explicitly?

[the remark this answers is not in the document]
szaszm: If it terminates, then it's impossible to accidentally catch, log and proceed on a programming error. It would be best if it prevented compilation, but since this is a runtime condition, we can only detect the condition at runtime. Crashing the program can not be ignored, which should prevent releasing with a programming error.

libminifi/src/utils/Id.cpp
```diff
@@ -193,8 +189,8 @@ IdGenerator::~IdGenerator() = default;
 
 uint64_t IdGenerator::getDeviceSegmentFromString(const std::string& str, int numBits) const {
   uint64_t deviceSegment = 0;
-  for (size_t i = 0; i < str.length(); i++) {
-    unsigned char c = toupper(str[i]);
+  for (auto ch : str) {
+    unsigned char c = toupper(ch);
```
fgerlits, line 193: The argument of `toupper` needs to fit into an `unsigned char`.  We have `static_cast<unsigned char>` elsewhere, but a `gsl::narrow` may be better.

fgerlits: This is a hexadecimal digit coming from the `uid.minifi.device.segment` property in the config file, so terminating (not starting up) would be OK, but if we don't want to change the behavior in this refactoring PR, that's fine, too.

extensions/opc/src/opc.cpp
```diff
@@ -550,7 +546,7 @@ std::string OPCDateTime2String(UA_DateTime raw_date) {
 
   int sz = snprintf(charBuf.data(), charBuf.size(), "%02hu-%02hu-%04hu %02hu:%02hu:%02hu.%03hu", dts.day, dts.month, dts.year, dts.hour, dts.min, dts.sec, dts.milliSec);
 
-  return std::string(charBuf.data(), sz);
+  return {charBuf.data(), static_cast<std::size_t>(sz)};
```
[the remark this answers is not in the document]
lordgamez: Updated in da2bf0f8cc4fd4748eb075a83196de5d6fe77c66

extensions/coap/protocols/CoapC2Protocol.cpp
```diff
@@ -186,29 +181,30 @@ minifi::c2::Operation CoapProtocol::getOperation(int type) const {
       return minifi::c2::Operation::PAUSE;
     case 9:
       return minifi::c2::Operation::RESUME;
+    default:
+      return minifi::c2::Operation::ACKNOWLEDGE;
```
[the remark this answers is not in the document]
lordgamez: Updated in da2bf0f8cc4fd4748eb075a83196de5d6fe77c66

libminifi/src/c2/ControllerSocketProtocol.cpp
```diff
@@ -237,6 +233,8 @@ void ControllerSocketProtocol::initialize(core::controller::ControllerServicePro
           }
         }
         break;
+        default:
+          throw std::runtime_error("Unhandled operation: " + std::to_string(head));
```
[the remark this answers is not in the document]
lordgamez: Replaced with a log in da2bf0f8cc4fd4748eb075a83196de5d6fe77c66

extensions/coap/protocols/CoapC2Protocol.cpp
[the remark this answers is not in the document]
lordgamez: Good idea, thanks! Replaced `abort` in 40d67984d1fe1c036c5f78c32c2177a0a1deecb7

extensions/libarchive/BinFiles.cpp
```diff
@@ -97,8 +97,8 @@ void BinFiles::preprocessFlowFile(core::ProcessContext* /*context*/, core::Proce
 void BinManager::gatherReadyBins() {
   std::lock_guard < std::mutex > lock(mutex_);
   std::vector< std::string > emptyQueue;
-  for (std::map<std::string, std::unique_ptr<std::deque<std::unique_ptr<Bin>>> >::iterator it=groupBinMap_.begin(); it !=groupBinMap_.end(); ++it) {
-    std::unique_ptr < std::deque<std::unique_ptr<Bin>>>&queue = it->second;
+  for (auto& kvp : groupBinMap_) {
+    auto &queue = kvp.second;
```
szaszm, line 101: This is the perfect use case for structured bindings.
```suggestion
  for (auto& [key, queue] : groupBinMap_) {
```

lordgamez: Good catch, updated in c067c00af8bb65f077173a4398705e59aec59339

szaszm (approved):

libminifi/src/utils/Id.cpp
```diff
@@ -193,8 +190,8 @@ IdGenerator::~IdGenerator() = default;
 
 uint64_t IdGenerator::getDeviceSegmentFromString(const std::string& str, int numBits) const {
   uint64_t deviceSegment = 0;
-  for (size_t i = 0; i < str.length(); i++) {
-    unsigned char c = toupper(str[i]);
+  for (auto ch : str) {
+    auto c = gsl::narrow<unsigned char>(std::toupper(gsl::narrow<unsigned char>(ch)));
```
adamdebreceni, line 194: is `gsl::narrow` the right choice here? `StringUtils` uses implicit conversions ([cppreference](https://en.cppreference.com/w/cpp/string/byte/toupper))

adamdebreceni: the discussion mentions either `static_cast` or `gsl::narrow_cast`, `gsl::narrow` is different as it can throw, is that intentional?

szaszm: @fgerlits suggested `gsl::narrow` originally. It shouldn't matter, since this function is working with a restricted set of characters. I suggested `gsl::narrow_cast` or `static_cast`, but since it hex digits only, @fgerlits commented that terminating is OK in that case. I don't think we should enforce that here, but I don't care that much.

adamdebreceni (approved):

PR #1362, opened 2022-06-30
MINIFICPP-1830 Verify regex properties in processors in case it is set to empty

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-1830

---------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/librdkafka/PublishKafka.cpp
```diff
@@ -184,12 +184,12 @@ class ReadCallback {
...
   }
 
-  static rd_kafka_headers_unique_ptr make_headers(const core::FlowFile& flow_file, utils::Regex& attribute_name_regex) {
+  static rd_kafka_headers_unique_ptr make_headers(const core::FlowFile& flow_file, std::optional<utils::Regex>& attribute_name_regex) {
     const gsl::owner<rd_kafka_headers_t*> result{ rd_kafka_headers_new(8) };
     if (!result) { throw std::bad_alloc{}; }
 
     for (const auto& kv : flow_file.getAttributes()) {
-      if (utils::regexSearch(kv.first, attribute_name_regex)) {
+      if (attribute_name_regex && utils::regexSearch(kv.first, *attribute_name_regex)) {
```
szaszm, line 192: Wherever regexSearch is used over regexMatch, we should check if it really makes sense to use subsequence matching.

lordgamez: Good idea, updated in 2c70c9d7f88a3b7568408c2b45c44516b199ae37

libminifi/include/utils/HTTPUtils.h
```diff
@@ -31,14 +31,14 @@ so we convert localhost to our local hostname.
 inline bool parse_http_components(const std::string &url, std::string &port, std::string &scheme, std::string &path) {
 #ifdef WIN32
   auto hostname = (url.find(org::apache::nifi::minifi::io::Socket::getMyHostName()) != std::string::npos ? org::apache::nifi::minifi::io::Socket::getMyHostName() : "localhost");
-  std::string regex_str = "^(http|https)://(" + hostname + ":)([0-9]+)?(/.*)$";
+  std::string regex_str = "(http|https)://(" + hostname + ":)([0-9]+)?(/.*)$";
 #else
-  std::string regex_str = "^(http|https)://(localhost:)([0-9]+)?(/.*)$";
+  std::string regex_str = "(http|https)://(localhost:)([0-9]+)?(/.*)$";
```
szaszm, line 36: Please remove the trailing `$` sign / end of line match as well. It should no longer be necessary, just like `^`.

lordgamez: Good catch, removed in 4d1784c2a28e7e9982f71ba3c77d0fac4f88c0c4

extensions/civetweb/processors/ListenHTTP.cpp
```diff
@@ -108,10 +108,12 @@ void ListenHTTP::onSchedule(core::ProcessContext *context, core::ProcessSessionF
 
   bool randomPort = listeningPort == "0";
 
-  std::string authDNPattern;
+  std::string authDNPattern = ".*";
 
   if (context->getProperty(AuthorizedDNPattern.getName(), authDNPattern) && !authDNPattern.empty()) {
     logger_->log_debug("ListenHTTP using %s: %s", AuthorizedDNPattern.getName(), authDNPattern);
+  } else {
+    logger_->log_debug("Authorized DN Pattern not set or invalid, using default '%s' pattern", authDNPattern);
```
adamdebreceni, line 116: shouldn't we set `authDNPattern` here to the default `".*"`? I think the property could contain the empty string value and overwrite the default at line 113

lordgamez: Good catch, fixed in 780e0ce9d630ec0fd4214b676a2d19da5d840755

adamdebreceni (approved):

extensions/librdkafka/PublishKafka.cpp
```diff
@@ -184,12 +184,12 @@ class ReadCallback {
     });
   }
 
-  static rd_kafka_headers_unique_ptr make_headers(const core::FlowFile& flow_file, utils::Regex& attribute_name_regex) {
+  static rd_kafka_headers_unique_ptr make_headers(const core::FlowFile& flow_file, std::optional<utils::Regex>& attribute_name_regex) {
```
fgerlits, line 187: could this be `const optional<Regex>&`?  (also at line 241)

extensions/standard-processors/processors/GetFile.cpp
```diff
@@ -244,7 +244,7 @@ bool GetFile::fileMatchesRequestCriteria(std::string fullName, std::string name,
     return false;
 
   utils::Regex rgx(request.fileFilter);
-  if (!utils::regexSearch(name, rgx)) {
```
fgerlits, line 247: This is a breaking change, as shown by the unit test.  I think changing the default value of `FileFilter` to `.*` (which it really [almost] was, with `regexSearch`) would lessen its impact.

extensions/civetweb/processors/ListenHTTP.cpp
```diff
@@ -275,10 +277,10 @@ void ListenHTTP::processRequestBuffer(core::ProcessSession *session) {
   logger_->log_debug("ListenHTTP transferred %zu flow files from HTTP request buffer", flow_file_count);
 }
 
-ListenHTTP::Handler::Handler(std::string base_uri, core::ProcessContext *context, std::string &&auth_dn_regex, std::string &&header_as_attrs_regex)
+ListenHTTP::Handler::Handler(std::string base_uri, core::ProcessContext *context, std::string &&auth_dn_regex, std::optional<std::string> &&headers_as_attrs_regex)
     : base_uri_(std::move(base_uri)),
       auth_dn_regex_(std::move(auth_dn_regex)),
-      headers_as_attrs_regex_(std::move(header_as_attrs_regex)),
+      headers_as_attrs_regex_(std::move(headers_as_attrs_regex)),
       process_context_(context) {
```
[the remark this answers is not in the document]
lordgamez: Good point, I updated it to pass `std::nullopt` in case the property is not set or empty.

extensions/librdkafka/PublishKafka.cpp
lordgamez: Yes it can, updated in 07973bdddd5299402d6c0c7a2e71ff7e0fc39e67

extensions/standard-processors/processors/GetFile.cpp
lordgamez: You are right it's a breaking change, but the reason I changed it is to be aligned with the NiFi's implementation, to have the same default value and use the same regex matching.

lordgamez: I don't know how closely we want to stay with the NiFi implementation, but I agree it would be better to have `.*` as the default. I'm okay with both, I wonder what the others think about this.

extensions/civetweb/processors/ListenHTTP.cpp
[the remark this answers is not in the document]
fgerlits: The `move` at line 283 (now 284) is still not happening.  I would change this to
```c++
  handler_ = std::make_unique<Handler>(basePath, context, std::move(authDNPattern),
    headersAsAttributesPattern.empty() ? std::nullopt : std::make_optional<utils::Regex>(headersAsAttributesPattern));
```
and
```c++
ListenHTTP::Handler::Handler(std::string base_uri, core::ProcessContext *context, std::string &&auth_dn_regex, std::optional<utils::Regex> &&headers_as_attrs_regex)
    : base_uri_(std::move(base_uri)),
      auth_dn_regex_(std::move(auth_dn_regex)),
      headers_as_attrs_regex_(std::move(headers_as_attrs_regex)),
      process_context_(context) {
```

extensions/standard-processors/processors/GetFile.cpp
szaszm: I'm fine any of these: `.*`, `[^\\.].*`, `[^.].*`

extensions/civetweb/processors/ListenHTTP.cpp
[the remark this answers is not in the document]
lordgamez: Updated in 6204845653afff905ccac36e42fe0c5bf93fee82

fgerlits (approved):

martinzink (approved):

PR #1364, opened 2022-07-05
MINIFICPP-1878 Enable Ninja build in Windows builds

lordgamez:
Ninja build can take advantage of multiple threads better than msbuild. In my experience ninja build is significantly faster on a local environment as it uses all available cores for parallel build jobs. Unfortunately the ninja build does not seem to be faster in the CI environment probably due to the limited 2 threads available, so I reverted that commit.

https://issues.apache.org/jira/browse/MINIFICPP-1878

--------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


cmake/BundledCivetWeb.cmake
```diff
@@ -64,12 +61,11 @@ function(use_bundled_civetweb SOURCE_DIR BINARY_DIR)
     # Build project
     ExternalProject_Add(
             civetweb-external
-            URL "https://github.com/civetweb/civetweb/archive/v1.12.tar.gz"
-            URL_HASH "SHA256=8cab1e2ad8fb3e2e81fed0b2321a5afbd7269a644c44ed4c3607e0a212c6d9e1"
+            GIT_REPOSITORY "https://github.com/civetweb/civetweb.git"
+            GIT_TAG "4447b6501d5c568b4c6c0940eac801ec690b2250" # commit containing fix for MSVC issue https://github.com/civetweb/civetweb/issues/1024
```
szaszm, line 65: I would prefer to stay on HTTP and a release version, and keep the necessary patch.
Switching to FetchContent and keeping the git commit ref is a reasonable alternative, too. It doesn't make the commit reference any more stable and tested, but at least it helps with caching artifacts between builds, since one can specify an external FETCHCONTENT_BASE_DIR.

win_build_vs.bat
```diff
@@ -73,12 +73,17 @@ for %%x in (%*) do (
...
+    cmake -G %generator% -A %build_platform% -DINSTALLER_MERGE_MODULES=%installer_merge_modules% -DTEST_CUSTOM_WEL_PROVIDER=%test_custom_wel_provider% -DENABLE_SQL=%build_SQL% -DUSE_REAL_ODBC_TEST_DRIVER=%real_odbc% -DCMAKE_BUILD_TYPE_INIT=%cmake_build_type% -DCMAKE_BUILD_TYPE=%cmake_build_type% -DWIN32=WIN32 -DENABLE_LIBRDKAFKA=%build_kafka% -DENABLE_JNI=%build_jni% -DOPENSSL_OFF=OFF -DENABLE_COAP=%build_coap% -DENABLE_AWS=%build_AWS% -DENABLE_PDH=%build_PDH% -DENABLE_AZURE=%build_azure% -DENABLE_SFTP=%build_SFTP% -DENABLE_SPLUNK=%build_SPLUNK% -DENABLE_GCP=%build_GCP% -DENABLE_NANOFI=%build_nanofi% -DENABLE_OPENCV=%build_opencv% -DENABLE_PROMETHEUS=%build_prometheus% -DENABLE_ELASTICSEARCH=%build_ELASTIC% -DUSE_SHARED_LIBS=OFF -DDISABLE_CONTROLLER=ON  -DBUILD_ROCKSDB=ON -DFORCE_WINDOWS=ON -DUSE_SYSTEM_UUID=OFF -DDISABLE_LIBARCHIVE=OFF -DENABLE_SCRIPTING=OFF -DEXCLUDE_BOOST=ON -DENABLE_WEL=ON -DFAIL_ON_WARNINGS=OFF -DSKIP_TESTS=%skiptests% %strict_gsl_checks% %redist% -DENABLE_LINTER=%build_linter% "%scriptdir%" && msbuild /m nifi-minifi-cpp.sln /property:Configuration=%cmake_build_type% /property:Platform=%build_platform% && copy bin\%cmake_build_type%\minifi.exe main\
+)
```
szaszm, line 86: Could you avoid the duplication of this monster of a line? I'm thinking about introducing a build command variable earlier, and executing that instead of `&& ninja` or `&& msbuild`.
Hopefully something like this could work:
```suggestion
if [%generator%] EQU ["Ninja"] (
    set "buildcmd=ninja"
) else (
    set "buildcmd=msbuild /m nifi-minifi-cpp.sln /property:Configuration=%cmake_build_type% /property:Platform=%build_platform% && copy bin\%cmake_build_type%\minifi.exe main\"
)
cmake -G %generator% -DINSTALLER_MERGE_MODULES=%installer_merge_modules% -DTEST_CUSTOM_WEL_PROVIDER=%test_custom_wel_provider% -DENABLE_SQL=%build_SQL% -DUSE_REAL_ODBC_TEST_DRIVER=%real_odbc% -DCMAKE_BUILD_TYPE_INIT=%cmake_build_type% -DCMAKE_BUILD_TYPE=%cmake_build_type% -DWIN32=WIN32 -DENABLE_LIBRDKAFKA=%build_kafka% -DENABLE_JNI=%build_jni% -DOPENSSL_OFF=OFF -DENABLE_COAP=%build_coap% -DENABLE_AWS=%build_AWS% -DENABLE_PDH=%build_PDH% -DENABLE_AZURE=%build_azure% -DENABLE_SFTP=%build_SFTP% -DENABLE_SPLUNK=%build_SPLUNK% -DENABLE_GCP=%build_GCP% -DENABLE_NANOFI=%build_nanofi% -DENABLE_OPENCV=%build_opencv% -DENABLE_PROMETHEUS=%build_prometheus% -DENABLE_ELASTICSEARCH=%build_ELASTIC% -DUSE_SHARED_LIBS=OFF -DDISABLE_CONTROLLER=ON  -DBUILD_ROCKSDB=ON -DFORCE_WINDOWS=ON -DUSE_SYSTEM_UUID=OFF -DDISABLE_LIBARCHIVE=OFF -DENABLE_SCRIPTING=OFF -DEXCLUDE_BOOST=ON -DENABLE_WEL=ON -DFAIL_ON_WARNINGS=OFF -DSKIP_TESTS=%skiptests% %strict_gsl_checks% %redist% -DENABLE_LINTER=%build_linter% "%scriptdir%" && %buildcmd%
```

lordgamez: Good point, thanks, updated in d578834c8d66fac12537468c21b8887283a38045

win_build_vs.bat
```diff
@@ -73,12 +73,18 @@ for %%x in (%*) do (
...
+if [%generator%] EQU ["Ninja"] (
+    set "buildcmd=ninja"
+) else (
+    set "buildcmd=msbuild /m nifi-minifi-cpp.sln /property:Configuration=%cmake_build_type% /property:Platform=%build_platform% && copy bin\%cmake_build_type%\minifi.exe main\"
```
fgerlits, line 85: Is the `&& copy bin\%cmake_build_type%\minifi.exe main\` not needed when building with ninja?

win_build_vs.bat
```diff
@@ -73,12 +73,18 @@ for %%x in (%*) do (
     if [%%~x] EQU [/NONFREEUCRT] set "redist=-DMSI_REDISTRIBUTE_UCRT_NONASL=ON"
     if [%%~x] EQU [/L]           set build_linter=ON
     if [%%~x] EQU [/RO]          set real_odbc=ON
+    if [%%~x] EQU [/NJ]          set generator="Ninja"
```
fgerlits, line 76: I would prefer `/NINJA`

win_build_vs.bat
lordgamez: Good catch, it should be copied, updated in 5aa6d9a5942b642f49b0903ee80d1f74fc0ec82c

win_build_vs.bat
lordgamez: I think it's better to be more explicit, updated in 5aa6d9a5942b642f49b0903ee80d1f74fc0ec82c

cmake/CivetWeb.cmake
```diff
@@ -0,0 +1,42 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+include(FetchContent)
+
+if (NOT WIN32)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-error" CACHE STRING "" FORCE)
+  set(CMAKE_CFLAGS "${CMAKE_CFLAGS} -Wno-error" CACHE STRING "" FORCE)
+endif()
```
szaszm, line 23: Is this leaking outside of the flags of this module? I wouldn't want to accidentally disable Werror globally if we have `FAIL_ON_WARNINGS=ON`.

lordgamez: I'm not sure if they could affect other modules, but it could be possible that the targets defined after civetweb would be affected. To make sure it doesn't happen I updated to only set target specific flags in civetweb in 4e37ffca0fc66548cf89b4738a8f73a7f9a5645e

cmake/CivetWeb.cmake
```diff
@@ -0,0 +1,46 @@
...
+
+add_dependencies(civetweb-c-library OpenSSL::Crypto OpenSSL::SSL)
+add_dependencies(civetweb-cpp OpenSSL::Crypto OpenSSL::SSL)
+
+set(CIVETWEB_C_FLAGS "${CMAKE_CFLAGS} -Wno-error" CACHE STRING "" FORCE)
+set(CIVETWEB_CXX_LAGS "${CMAKE_CXX_FLAGS} -Wno-error" CACHE STRING "" FORCE)
+string(REPLACE " " ";" REPLACED_C_FLAGS ${CIVETWEB_C_FLAGS})
+string(REPLACE " " ";" REPLACED_CXX_FLAGS ${CIVETWEB_CXX_LAGS})
+target_compile_options(civetweb-c-library INTERFACE ${REPLACED_C_FLAGS})
+target_compile_options(civetweb-cpp INTERFACE ${REPLACED_CXX_FLAGS})
```
szaszm: This takes all of the global flags, appends `-Wno-error` and appends them to anything linking to civet. Two problems:
1. It's a bit excessive to append all the flags from C/CXXFLAGS again to civet. Anything before `-Wno-error` will be duplicated on the command line.
2. The flags should apply to civet itself, but not the dependencies. Use `PRIVATE` instead of `INTERFACE`.

lordgamez: Good point, fixed in e5b0aad87214253154820ee11f858fafa622e12d

adamdebreceni (approved):

martinzink (approved):

PR #1367, opened 2022-07-11
MINIFICPP-1822 - Add alert capability

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/src/core/logging/alert/AlertSink.cpp
```diff
@@ -0,0 +1,267 @@
...
+AlertSink::~AlertSink() {
+  {
+    std::lock_guard lock(mtx_);
+    running_ = false;
+    cv_.notify_all();
+  }
+  if (flush_thread_.joinable()) {
+    flush_thread_.join();
+  }
+}
```
fgerlits, line 186: do we need to delete `services_`, too?

libminifi/src/core/logging/alert/AlertSink.cpp
```diff
@@ -0,0 +1,267 @@
...
+  std::match_results<std::string_view::const_iterator> match;
+  std::string_view payload(msg.payload.data(), msg.payload.size());
+  if (!std::regex_match(payload.begin(), payload.end(), match, config_.filter)) {
+    return;
+  }
+  size_t hash = 0;
+  for (size_t idx = 1; idx < match.size(); ++idx) {
+    std::string_view submatch;
+    if (match[idx].first != match[idx].second) {
+      // TODO(adebreceni): std::string_view(It begin, It end) is not yet supported on all platforms
```
fgerlits, line 132: Can you add to the comment on which platform it is not supported yet, please?  That way, we'll know when we can remove the workaround.

libminifi/include/core/logging/alert/AlertSink.h
```diff
@@ -0,0 +1,109 @@
...
+    static LogBuffer allocate(size_t size);
+    LogBuffer commit();
+    [[nodiscard]]
+    size_t size() const;
+  };
+
+  class LiveLogSet {
+    std::chrono::milliseconds lifetime_{};
+    std::unordered_set<size_t> ignored_;
+    std::deque<std::pair<std::chrono::milliseconds, size_t>> ordered_;
```
fgerlits, line 69: I think these could be named better; maybe
```suggestion
    std::unordered_set<size_t> hashes_to_ignore_;
    std::deque<std::pair<std::chrono::milliseconds, size_t>> timestamped_hashes_;
```
?

Also, typedefing `size_t` to something like `Hash` could help with readability.

libminifi/src/core/logging/alert/AlertSink.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+  bool req_success = client->submit();
+
+  int64_t resp_code = client->getResponseCode();
+  const bool client_err = 400 <= resp_code && resp_code < 500;
+  const bool server_err = 500 <= resp_code && resp_code < 600;
+  if (client_err || server_err) {
+    logger_->log_error("Error response code '" "%" PRId64 "' from '%s'", resp_code, config_.url);
+  } else {
+    logger_->log_debug("Response code '" "%" PRId64 "' from '%s'", resp_code, config_.url);
```
fgerlits, line 228: This could be logged at `info` or `warning` level if the response code is not an error, but not a `2xx` success, either.

libminifi/include/utils/Id.h
```diff
@@ -141,18 +142,14 @@ struct hash<org::apache::nifi::minifi::utils::Identifier> {
   size_t operator()(const org::apache::nifi::minifi::utils::Identifier& id) const noexcept {
     static_assert(sizeof(org::apache::nifi::minifi::utils::Identifier) % sizeof(size_t) == 0);
     constexpr int slices = sizeof(org::apache::nifi::minifi::utils::Identifier) / sizeof(size_t);
```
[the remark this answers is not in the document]
adamdebreceni: does the `static_assert` at linke 143 not cover these cases?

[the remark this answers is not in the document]
fgerlits: ah yes, thanks, I somehow missed that, never mind

libminifi/include/core/logging/alert/AlertSink.h
```diff
@@ -0,0 +1,109 @@
...
+
+#include <deque>
+#include <mutex>
+#include <unordered_set>
+#include <regex>
+#include <utility>
+#include <string>
+#include <memory>
+
+#include "controllers/SSLContextService.h"
```
fgerlits, line 28: This include breaks the Windows build, because `SSLContextService.h` pulls in `wincrypt.h`, which `#define`s `X509_CERT_PAIR` as `(LPCSTR) 53`, but OpenSSL/LibreSSL has its own `X509_CERT_PAIR` struct which conflicts with this.

Specifically, in `ExtendedKeyUsage.cpp`, we first `#include <openssl/x509v3.h>`, then we `#include "core/logging/LoggerConfiguration.h --> AlertSink.h --> SSLContextService.h --> wincrypt.h`, which redefines `X509_CERT_PAIR`.

I think the best solution would be to move `#include "controllers/SSLContextService.h"` to `AlertSink.cpp`, and put a forward declaration in the header instead.

extensions/http-curl/tests/AlertTests.cpp
```diff
@@ -0,0 +1,148 @@
...
+    // time = 13
+
+    verifyLogsArrived({
+        "other <begin>one<end>"
+    });
+
+    return true;
+  };
+
+  harness.run(flow_file, dir.getPath());
```
fgerlits, line 147:
```suggestion
  harness.run(flow_file.string(), dir.getPath());
```
as `std::filesystem::path` doesn't convert implicitly to `std::string` on Windows.

extensions/http-curl/tests/AlertTests.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+  std::function<bool()> verify_;
+};
+
+TEST_CASE("Alert system forwards logs") {
+  auto clock = std::make_shared<utils::ManualClock>();
+  utils::timeutils::setClock(clock);
+
+  TempDirectory dir;
+  auto flow_file = std::filesystem::path(dir.getPath()) / "config.yml";
```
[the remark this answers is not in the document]
adamdebreceni: changed it

libminifi/src/core/logging/LoggerConfiguration.cpp
```diff
@@ -60,26 +62,28 @@ namespace org::apache::nifi::minifi::core::logging {
...
-    return spdlog::level::warn;
-  } else if (utils::StringUtils::equalsIgnoreCase(level_name, "error")) {
-    return spdlog::level::err;
-  } else if (utils::StringUtils::equalsIgnoreCase(level_name, "critical")) {
-    return spdlog::level::critical;
-  } else if (utils::StringUtils::equalsIgnoreCase(level_name, "off")) {
-    return spdlog::level::off;
+namespace internal {
+
+bool LoggerNamespace::findSink(std::function<bool(const std::shared_ptr<spdlog::sinks::sink>&)> filter) const {
```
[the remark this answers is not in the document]
adamdebreceni: changed it

extensions/http-curl/tests/AlertTests.cpp
```diff
@@ -0,0 +1,148 @@
...
+  harness.getConfiguration()->setHome(dir.getPath());
+
+  auto log_props = std::make_shared<logging::LoggerProperties>();
+  log_props->set("appender.alert1", "alert");
+  log_props->set("appender.alert1.url", harness.getC2RestUrl());
+  log_props->set("appender.alert1.filter", ".*<begin>(.*)<end>.*");
+  log_props->set("appender.alert1.rate.limit", "10 s");
+  log_props->set("appender.alert1.flush.period", "1 s");
+  log_props->set("logger.root", "INFO,alert1");
+  logging::LoggerConfiguration::getConfiguration().initialize(log_props);
```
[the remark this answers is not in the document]
adamdebreceni: added commented section to `conf/minifi-log.properties`

libminifi/src/core/logging/alert/AlertSink.cpp
adamdebreceni: nice catch

libminifi/src/core/logging/alert/AlertSink.cpp
adamdebreceni: added compiler version

libminifi/include/core/logging/alert/AlertSink.h
adamdebreceni: renamed

libminifi/src/core/logging/alert/AlertSink.cpp
adamdebreceni: added a new warning

libminifi/include/core/logging/alert/AlertSink.h
adamdebreceni: done

extensions/http-curl/tests/AlertTests.cpp
adamdebreceni: done

libminifi/include/utils/TimeUtil.h
```diff
@@ -70,6 +72,16 @@ class Clock {
  public:
   virtual ~Clock() = default;
   virtual std::chrono::milliseconds timeSinceEpoch() const = 0;
+  virtual std::chrono::milliseconds wait_until(std::condition_variable& cv, std::unique_lock<std::mutex>& lck, std::chrono::milliseconds time, const std::function<bool()>& pred) {
```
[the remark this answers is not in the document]
adamdebreceni: based on the comment above `Clock`, these classes' purpose is to be mocked, as it is not possible to safely wait when using a `Clock` (as it might be an instance of `ManualClock`) I think this new method has relevance to all users of `Clock`

libminifi/include/core/logging/alert/AlertSink.h
```diff
@@ -0,0 +1,109 @@
...
+
+class AlertSink : public spdlog::sinks::base_sink<std::mutex> {
+  struct Config {
+    std::string url;
+    std::optional<std::string> ssl_service_name;
+    int batch_size;
+    std::chrono::milliseconds flush_period;
+    std::chrono::milliseconds rate_limit;
+    int buffer_limit;
+    std::regex filter;
```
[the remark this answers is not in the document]
adamdebreceni: changed it to use `utils::Regex`, it has some shortcomings created a followup ticket [MINIFICPP-1903](https://issues.apache.org/jira/browse/MINIFICPP-1903)

lordgamez (approved):
LGTM, only some nitpicks

libminifi/include/utils/TestUtils.h
```diff
@@ -62,14 +63,45 @@ Identifier generateUUID() {
...
+  }
 
  private:
+  mutable std::mutex mtx_;
+  std::unordered_set<std::condition_variable*> cvs_;
   std::chrono::milliseconds time_{0};
 };
 
 
+
```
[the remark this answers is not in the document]
adamdebreceni: removed

libminifi/include/core/logging/alert/AlertSink.h
```diff
@@ -0,0 +1,112 @@
...
+    [[nodiscard]]
+    size_t size() const;
+  };
+
+  class LiveLogSet {
+    using Hash = size_t;
+    std::chrono::milliseconds lifetime_{};
+    std::unordered_set<Hash> hashes_to_ignore_;
+    std::deque<std::pair<std::chrono::milliseconds, Hash>> timestamped_hashes_;
+   public:
```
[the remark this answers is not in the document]
adamdebreceni: moved

libminifi/include/core/logging/alert/AlertSink.h
```diff
@@ -0,0 +1,112 @@
...
+    bool tryAdd(std::chrono::milliseconds now, Hash hash);
+    void setLifetime(std::chrono::milliseconds lifetime);
+  };
+
+ public:
+  // must be public for make_shared
+  AlertSink(Config config, std::shared_ptr<Logger> logger);
```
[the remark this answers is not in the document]
adamdebreceni: neat trick, but since there is a single `make_shared` that would utilize this, the raw pointer constructor of `std::shared_ptr` feels like more appropriate/simpler in this case

libminifi/src/core/logging/alert/AlertSink.cpp
```diff
@@ -0,0 +1,269 @@
...
+  return size_;
+}
+
+void AlertSink::LiveLogSet::setLifetime(std::chrono::milliseconds lifetime) {
+  lifetime_ = lifetime;
+}
+
+bool AlertSink::LiveLogSet::tryAdd(std::chrono::milliseconds now, size_t hash) {
+  auto limit = now - lifetime_;
+  while (!timestamped_hashes_.empty() && timestamped_hashes_.front().first < limit) {
```
adam-markovics, line 256: This is assuming that `timestamped_hashes_` is having the hashes ordered in a chronological order. I think it would be better to use a priority queue for that. Also, they could become invalid if `setLifeTime` is called between two `tryAdd`s and `lifetime_` is modified. But since I just see one call for it in the ` AlertSink` constructor, can this method be removed and `lifetime_` set from constructor only?

libminifi/include/utils/TimeUtil.h
```diff
@@ -70,6 +72,16 @@ class Clock {
...
+    auto now = timeSinceEpoch();
+    if (now < time) {
+      cv.wait_for(lck, time - now, [&] {
+        now = timeSinceEpoch();
+        return now >= time || pred();
+      });
+    }
+    return timeSinceEpoch();
```
adam-markovics, line 83: I think it would be better to tell in the return value that a timeout happened or the predicate returned true.

libminifi/src/utils/TimeUtil.cpp
```diff
@@ -0,0 +1,36 @@
...
+std::shared_ptr<Clock> getClock() {
+  std::lock_guard lock(global_clock_mtx);
+  return global_clock;
+}
+
+// test-only utility to specify what clock to use
+void setClock(std::shared_ptr<Clock> clock) {
+  std::lock_guard lock(global_clock_mtx);
+  global_clock = std::move(clock);
+}
```
szaszm, line 34: These globals can be avoided. Please use dependency injection.

libminifi/src/core/logging/alert/AlertSink.cpp
adamdebreceni: I don't remember what the purpose of `setLifeTime` was, removed it

libminifi/include/utils/Hash.h
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <cstddef>
+
+namespace org::apache::nifi::minifi::utils {
+
+// boost::hash_combine
```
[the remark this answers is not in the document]
adamdebreceni: updated the comment

libminifi/include/utils/Hash.h
```diff
@@ -0,0 +1,31 @@
...
+inline size_t hash_combine(size_t seed, size_t h) noexcept {
+  seed ^= h + 0x9e3779b9 + (seed << 6U) + (seed >> 2U);
+  return seed;
+}
```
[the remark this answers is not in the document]
adamdebreceni: rewrote it to the second version

libminifi/src/utils/TimeUtil.cpp
adamdebreceni: routing the clock through the logging system would be quite the hassle for something that has a test-only relevance, also being able to access this clock from anywhere makes sense

libminifi/include/core/logging/alert/AlertSink.h
```diff
@@ -0,0 +1,113 @@
...
+  };
+
+  struct Services {
+    std::shared_ptr<controllers::SSLContextService> ssl_service;
+    std::shared_ptr<AgentIdentificationProvider> agent_id;
+  };
+
+  struct LogBuffer {
+    size_t size_{0};
+    std::deque<std::pair<std::string, size_t>> data_;
```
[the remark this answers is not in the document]
adamdebreceni: changed it

libminifi/src/core/logging/alert/AlertSink.cpp
adam-markovics: Okay. Wouldn't you think a priority queue would better fit the role of `timestamped_hashes_`, so that it is always in a chronological order? Don't do it if it's too complicated, just an idea.

adamdebreceni: I think chronological order is already guaranteed as by default we use a steady clock

libminifi/include/utils/TimeUtil.h
adamdebreceni: yes it makes sense to be inline with the interface of `std::condition_variable::wait_until`, changed it to return a `bool`

adam-markovics (approved):

libminifi/src/core/logging/alert/AlertSink.cpp
```diff
@@ -0,0 +1,268 @@
...
+  };
+
+  config.batch_size = readPropertyOr(".batch.size", datasize_parser, "100 KB");
+  config.flush_period = readPropertyOr(".flush.period", TimePeriodValue::fromString, "5 s").getMilliseconds();
+  config.rate_limit = readPropertyOr(".rate.limit", TimePeriodValue::fromString, "10 min").getMilliseconds();
+  config.buffer_limit = readPropertyOr(".buffer.limit", datasize_parser, "1 MB");
+  config.level = readPropertyOr(".level", utils::parse_log_level, "trace");
+  config.ssl_service_name = logger_properties->getString(prop_name_prefix + ".ssl.context.service");
+
+  return std::shared_ptr<AlertSink>(new AlertSink(std::move(config), std::move(logger)));
```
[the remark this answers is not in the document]
adamdebreceni: my understanding is that the storage will be automatically freed up https://eel.is/c++draft/except.ctor#note-2
exception-safety of `std::shared_ptr<T>(new T{...})` was only of concern until C++17 before which a call like `f(std::shared_ptr<T>(new T{...}, g())` could indeed leak if `g()` throws between allocating in `new T{...}` and the constructor call of that object, from C++17 arguments are indeterminately sequenced so even this is not a concern anymore

libminifi/src/core/logging/alert/AlertSink.cpp
```diff
@@ -0,0 +1,268 @@
...
+        logger_->log_error("Service '%s' is not an SSLContextService", config_.ssl_service_name.value());
+        return;
+      }
+    } else {
+      logger_->log_error("Could not find service '%s'", config_.ssl_service_name.value());
+      return;
+    }
+  }
+
+  services.reset(services_.exchange(services.release()));
```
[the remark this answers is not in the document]
adamdebreceni: copy and move are automatically disabled as `AlertSink` has both `std::mutex` and `std::atomic` members, of course we could make this explicit, should we make it explicit?

fgerlits (approved):

libminifi/src/core/logging/alert/AlertSink.cpp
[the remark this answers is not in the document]
szaszm: I think this only applies to delegating constructors, where the delegating expression allocates via new.
Tested this scenario:  https://godbolt.org/z/zK6Wj6rTd

libminifi/src/core/logging/alert/AlertSink.cpp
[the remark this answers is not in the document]
szaszm: Yes, making this explicit would eliminate a similar doubt in future readers, so I think it's worth the extra couple of lines.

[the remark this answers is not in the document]
adamdebreceni: disabled copy/move ctor/assignment in [017d5bd](https://github.com/apache/nifi-minifi-cpp/pull/1367/commits/017d5bd474704cba5c2c82ed34915ab6555577db)

libminifi/src/core/logging/alert/AlertSink.cpp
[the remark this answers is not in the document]
szaszm: nevermind, shared_ptr constructor frees the ptr before throwing.

szaszm (approved):

szaszm:
pushed a rebase to main

szaszm:
force pushed the same thing to trigger github actions, because check didn't run for some reason

martinzink (approved):

PR #1349, opened 2022-06-13
MINIFICPP-1843 Implement PutElasticsearchJson

martinzink:
Thank you for submitting a contribution to Apache NiFi - MiNiFi C++.

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [ ] Is there a JIRA ticket associated with this PR? Is it referenced
     in the commit message?

- [ ] Does your PR title start with MINIFICPP-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.

- [ ] Has your PR been rebased against the latest commit within the target branch (typically main)?

- [ ] Is your initial contribution a single, squashed commit?

### For code changes:
- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?
- [ ] If applicable, have you updated the LICENSE file?
- [ ] If applicable, have you updated the NOTICE file?

### For documentation related changes:
- [ ] Have you ensured that format looks appropriate for the output in which it is rendered?

### Note:
Please ensure that once the PR is submitted, you check GitHub Actions CI results for build issues and submit an update to your PR as soon as possible.

PR #1369, opened 2022-07-13
MINIFICPP-1874 docker-verify should only run tests that are enabled

martinzink:
removed options duplication in cmake

### For all changes:
     in the commit message?

### For code changes:
- [X] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?
- [X] If applicable, have you updated the LICENSE file?
- [X] If applicable, have you updated the NOTICE file?

- [X] Have you ensured that format looks appropriate for the output in which it is rendered?

### Note:


docker/DockerVerify.sh
```diff
@@ -67,11 +67,11 @@ TEST_DIRECTORY="${docker_dir}/test/integration"
 export TEST_DIRECTORY
 
 # Add --no-logcapture to see logs interleaved with the test output
-BEHAVE_OPTS=(-f pretty --logging-level INFO --logging-clear-handlers --tags ~@no-ci)
```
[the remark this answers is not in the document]
martinzink: I moved the filtering to https://github.com/apache/nifi-minifi-cpp/pull/1369/files#diff-aebf957cd74e48986733a1597cb988264e869aa6709d4c6990e7f3e564e4e384R75

docker/Dockerfile
```diff
@@ -28,47 +28,7 @@ ARG GID=1000
 
 # PDH and WEL extensions and not listed as they are Windows specific
 # SYSTEMD extension is turned OFF explicitly as it has no use in an alpine container
```
lordgamez, line 30: These comments can be removed

docker/Dockerfile
```diff
@@ -28,47 +28,7 @@ ARG GID=1000
...
-ARG DISABLE_LZMA=OFF
-ARG DISABLE_BZIP2=OFF
-ARG ENABLE_SCRIPTING=OFF
-ARG DISABLE_PYTHON_SCRIPTING=
-ARG ENABLE_LUA_SCRIPTING=
-ARG ENABLE_KUBERNETES=OFF
-ARG ENABLE_PROCFS=OFF
-ARG ENABLE_PROMETHEUS=OFF
-ARG DISABLE_CONTROLLER=OFF
+ARG MINIFI_OPTIONS=""
```
lordgamez, line 31: Maybe we could add a comment what this argument should look like if someone would want to use this dockerfile individually without using the shell script or a make target.

extensions/sftp/tests/PutSFTPTests.cpp
```diff
@@ -506,7 +507,7 @@ TEST_CASE_METHOD(PutSFTPTestsFixture, "PutSFTP set mtime", "[PutSFTP]") {
 
   testFile("nifi_test/tstFile1.ext", "content 1");
   using namespace std::chrono;  // NOLINT(build/namespaces)
-  time_point<system_clock> modification_time = sys_days(January / 24 / 2065) + 5h + 20min;
+  system_clock::time_point modification_time = date::sys_days(date::January / 24 / 2065) + 5h + 20min;
```
lordgamez, line 510: How is this change tied to this PR?

docker/DockerVerify.sh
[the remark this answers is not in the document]
szaszm: My point is that there would be no harm in keeping the tag, even if unused. The added complexity is minimal, and there may be cases in the future where we need to disable individual (e.g. long running) tests, not the whole extension. It can also be readded, so it's fine if you prefer to still remove it.

extensions/sftp/tests/PutSFTPTests.cpp
martinzink: I found this during development, because it didnt compile on centos (usually this is not a problem because -DSKIP_TESTS is hardcoded to be ON)

docker/DockerVerify.sh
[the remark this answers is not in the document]
martinzink: The problem with the tag was that it influenced local runs as well. https://github.com/apache/nifi-minifi-cpp/pull/1349#discussion_r907500456
Due to this reason Id rather readd it later if we need it.

docker/Dockerfile
martinzink: nice catch, removed this in https://github.com/apache/nifi-minifi-cpp/pull/1369/commits/69bbe6d590f5608d809a28fff7aa062bfd80800b

extensions/sftp/tests/PutSFTPTests.cpp
lordgamez: Thanks for the info!

lordgamez (approved):

docker/bionic/Dockerfile
```diff
@@ -20,6 +20,8 @@
 FROM ubuntu:bionic AS build_deps
 LABEL maintainer="Apache NiFi <dev@nifi.apache.org>"
 
+# MINIFI_OPTIONS will be passed directly to cmake
+# use it to define cmake options (e.g. -DENABLE_AWS=ON -DENABLE_AZURE=ON)
```
adamdebreceni, line 24: I think this comment might be misplaced (also some other Dockerfiles)

martinzink: you are absolutely right :+1: 
fixed in https://github.com/apache/nifi-minifi-cpp/pull/1369/commits/6851250bc6ce26b4c7a0c5824e66cf8d7d16f923

adamdebreceni (approved):

PR #1373, opened 2022-07-19
MINIFICPP-1851 Collect pod metrics in Kubernetes

fgerlits:
https://issues.apache.org/jira/browse/MINIFICPP-1851

---

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/kubernetes/ApiClient.cpp
```diff
@@ -0,0 +1,53 @@
...
+    throw std::runtime_error("apiClient_create_with_base_path() failed");
+  }
+  return gsl::make_not_null(api_client);
+}
+
+}  // namespace
+
+namespace org::apache::nifi::minifi::kubernetes {
+
+ApiClient::ApiClient() : api_client_(createApiClient(&base_path_, &ssl_config_, &api_keys_)) {}
```
adamdebreceni, line 45: while this is safe in its current form, if the order of members were to change, this could break, I think either a comment in the header, that the order of members should not change (or that `api_client_` must come after others), or moving this initialization to the body of the constructor could help

extensions/kubernetes/ApiClient.h
```diff
@@ -0,0 +1,48 @@
...
+  ApiClient& operator=(ApiClient&&) = delete;
+  ApiClient& operator=(const ApiClient&) = delete;
+
+  [[nodiscard]] gsl::not_null<apiClient_t*> getClient() const noexcept { return api_client_; }
+
+ private:
+  char* base_path_ = nullptr;
+  sslConfig_t* ssl_config_ = nullptr;
+  list_t* api_keys_ = nullptr;
+  gsl::not_null<apiClient_t*> api_client_;
```
adamdebreceni, line 45: the objects behind these pointers seem to be managed by `ApiClient`, should we wrap then in `gsl::owner`?

extensions/kubernetes/ApiClient.cpp
fgerlits: Good point.  If it was in the body of the constructor, it couldn't be marked `gsl::not_null`, so I added a comment instead.  https://github.com/apache/nifi-minifi-cpp/pull/1373/commits/58fb3ddeb8e35322c17c8fb1d49ffc0ca57d90ac

extensions/kubernetes/ApiClient.h
fgerlits: I have added `gsl::owner` annotations in https://github.com/apache/nifi-minifi-cpp/pull/1373/commits/58fb3ddeb8e35322c17c8fb1d49ffc0ca57d90ac

extensions/kubernetes/MetricsFilter.h
```diff
@@ -0,0 +1,30 @@
...
+#include <functional>
+#include <string>
+
+#include "utils/expected.h"
+
+namespace org::apache::nifi::minifi::kubernetes::metrics {
+
+[[nodiscard]] nonstd::expected<std::string, std::string> filter(
+    const std::string& metrics_json,
+    const std::function<bool(const std::string& name_space, const std::string& pod_name, const std::string& container_name)>& filter_function);
```
adamdebreceni, line 28: I'm not sure how I feel about the filter function's type, specifically the three `std::string` parameters, given how we construct the lambda and how we forward to another function taking three `std::string`s (`KubernetesControllerService::matchesRegexFilters`), I think we could wrap this in some kind of `ContainerInfo` struct, it could also make it easier to later add other things we would like to filter on (maybe labels?)

fgerlits: Yes, I agree that is more readable.  Changed in 4e9944adebd8b6800d1e35dd00cb76ac85d00f7d.

adamdebreceni (approved):

lordgamez (approved):

extensions/kubernetes/MetricsApi.cpp
```diff
@@ -0,0 +1,47 @@
...
+}
+
+#include "ApiClient.h"
+
+namespace {
+struct genericClient_t_deleter { void operator()(genericClient_t* ptr) const noexcept { genericClient_free(ptr); } };
+using genericClient_unique_ptr = std::unique_ptr<genericClient_t, genericClient_t_deleter>;
+
+struct char_deleter { void operator()(char* ptr) const noexcept { free(ptr); } };
+using char_unique_ptr = std::unique_ptr<char, char_deleter>;
```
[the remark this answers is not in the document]
fgerlits: fixed in ebfa8b42bbebbac3c74adb2875c66ad003a458bb

extensions/kubernetes/MetricsFilter.cpp
```diff
@@ -0,0 +1,81 @@
...
+nonstd::expected<std::string, std::string> filter(const std::string& metrics_json, const std::function<bool(const kubernetes::ContainerInfo&)>& filter_function) {
+  rapidjson::Document document;
+  rapidjson::ParseResult parse_result = document.Parse<rapidjson::kParseStopWhenDoneFlag>(metrics_json.data());
+  if (parse_result.IsError()) {
+    return nonstd::make_unexpected(utils::StringUtils::join_pack("Error parsing the metrics received from the Kubernetes API at offset ",
+        std::to_string(parse_result.Offset()), ": ", rapidjson::GetParseError_En(parse_result.Code())));
+  }
+
+  if (!document.HasMember("items") || !document["items"].IsArray()) {
+    return nonstd::make_unexpected("Unexpected JSON received from the Kubernetes API: missing list of 'items'");
```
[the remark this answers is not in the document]
fgerlits: Since we only use this from one place, and the only thing we do with the error is log it, I would leave it as it is for now.  We can refine it later.

szaszm (approved):

PR #1374, opened 2022-07-21
MINIFICPP-1888 Move all extension builds to CentOS job

lordgamez:
With more extensions added to be turned on by default, the centos build targeting the default extensions takes longer to finish. To make sure we can build on both ubuntu and CentOS and to minimize the CI resources used, all extension builds that are available in the CentOS docker build are moved to the CentOS job and tests are run on CentOS and in the clang Ubuntu build.

https://issues.apache.org/jira/browse/MINIFICPP-1888

------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


szaszm:
I would rather start disabling a few extensions here and there. We should test the core everywhere, but testing everything everywhere might be more excessive than it's worth it.

I would disable AWS, GCP, Kafka and SQL on centos (and maybe on mac os), since they pull in dependency libs that take some time to compile.

lordgamez:
> I would rather start disabling a few extensions here and there. We should test the core everywhere, but testing everything everywhere might be more excessive than it's worth it.
> 
> I would disable AWS, GCP, Kafka, SQL and Civetweb on centos (and maybe on mac os), since they pull in dependency libs that take some time to compile.

I thought the reason for adding new default extensions was to include them in the centos build by default, but then I'm mistaken.

szaszm:
I did it in #1341, since I thought that these extensions are stable and well tested, there's not much reason a user wouldn't want to build and have them. Released convenience binaries should contain these. If a user doesn't want the functionality, they can just remove the extension .so/.dlls from their installation.

The CI is a different beast, since it's costing the ASF money per machine minute. We can use it, but we should make an effort to not overuse it, and testing extensions once or twice on _a_ linux is probably good enough coverage.

adam-markovics (approved):

fgerlits (approved):

.github/workflows/ci.yml
```diff
@@ -201,11 +199,19 @@ jobs:
...
+              -DENABLE_OPENWSMAN=ON -DENABLE_OPS=ON -DENABLE_PYTHON=ON -DENABLE_SENSORS=ON -DENABLE_SFTP=ON -DENABLE_SQL=ON -DENABLE_SYSTEMD=ON -DENABLE_TENSORFLOW=OFF \
+              -DENABLE_USB_CAMERA=ON -DENABLE_SCRIPTING=ON -DENABLE_LUA_SCRIPTING=ON -DENABLE_KUBERNETES=ON -DENABLE_GCP=ON -DENABLE_PROCFS=ON -DENABLE_PROMETHEUS=ON -DENABLE_ELASTICSEARCH=ON \
+              -DDOCKER_SKIP_TESTS=OFF -DDOCKER_BUILD_ONLY=ON -DDOCKER_CCACHE_DUMP_LOCATION=$HOME/.ccache .. && make centos
+      - id: test
+        run: docker run --rm apacheminificpp:$(docker images | grep apacheminificpp | grep centos | awk '{print $2}') bash -c 'cd /opt/minifi/build && make test ARGS="--timeout 300 -j8 --output-on-failure"'
   docker_integration_tests:
     name: "Docker integration tests"
     runs-on: ubuntu-20.04
-    timeout-minutes: 120
+    timeout-minutes: 180
```
fgerlits, line 214: we could increase the timeout on the MacOS build to 180 minutes, too, as I have seen timeouts on that job in my fork, too

docker/DockerBuild.sh
```diff
@@ -146,18 +149,31 @@ BUILD_ARGS+=("--build-arg" "UID=${UID_ARG}"
             "--build-arg" "GID=${GID_ARG}"
             "--build-arg" "MINIFI_VERSION=${MINIFI_VERSION}"
             "--build-arg" "DUMP_LOCATION=${DUMP_LOCATION}"
-            "--build-arg" "DISTRO_NAME=${DISTRO_NAME}")
+            "--build-arg" "DISTRO_NAME=${DISTRO_NAME}"
+            "--build-arg" "DOCKER_SKIP_TESTS=${DOCKER_SKIP_TESTS}")
 
-if [ -n "${DOCKER_CCACHE_DUMP_LOCATION}" ]; then
-  DOCKER_BUILDKIT=1 docker build "${BUILD_ARGS[@]}" -f ${DOCKERFILE} --target build -t minifi_build ..
+if [ -n "${DISTRO_NAME}" ]; then
```
fgerlits, line 155: why can't we (or why don't we want to) do the distro-based builds in the same way as the plain docker build?

docker/DockerBuild.sh
```diff
@@ -146,18 +149,31 @@ BUILD_ARGS+=("--build-arg" "UID=${UID_ARG}"
...
+else
+  if [ -n "${DOCKER_CCACHE_DUMP_LOCATION}" ]; then
+    DOCKER_BUILDKIT=1 docker build "${BUILD_ARGS[@]}" -f ${DOCKERFILE} --target build -t minifi_build ..
 
-DOCKER_BUILDKIT=0 docker build "${BUILD_ARGS[@]}" -f ${DOCKERFILE} -t apacheminificpp:"${TAG}" ..
+    container_id=$(docker run --rm -d minifi_build sh -c "while true; do sleep 1; done")
+    mkdir -p "${DOCKER_CCACHE_DUMP_LOCATION}"
+    docker cp "${container_id}:/home/minificpp/.ccache/." "${DOCKER_CCACHE_DUMP_LOCATION}"
+    docker rm -f "${container_id}"
+  fi
```
fgerlits, line 173: this is a large, almost-duplicate block -- is there any way of avoiding this?

docker/DockerBuild.sh
```diff
@@ -146,18 +149,31 @@ BUILD_ARGS+=("--build-arg" "UID=${UID_ARG}"
...
+  echo DOCKER_BUILDKIT=1 docker build "${BUILD_ARGS[@]}" -f ${DOCKERFILE} -t apacheminificpp:"${TAG}" ..
+  DOCKER_BUILDKIT=1 docker build "${BUILD_ARGS[@]}" -f ${DOCKERFILE} -t apacheminificpp:"${TAG}" ..
```
fgerlits, line 175: What does `DOCKER_BUILDKIT=0/1` do, and why it change here from 0 to 1?

docker/centos/Dockerfile
```diff
@@ -17,33 +17,47 @@
...
+
+USER ${USER}
 
+# Perform the build
+RUN cd $MINIFI_BASE_DIR && \
+    cd build && \
+    source /opt/rh/devtoolset-10/enable && \
+    export PATH=/usr/lib64/ccache${PATH:+:${PATH}} && \
+    cmake3 -DSTATIC_BUILD= -DSKIP_TESTS=${DOCKER_SKIP_TESTS} ${MINIFI_OPTIONS} -DAWS_ENABLE_UNITY_BUILD=OFF -DEXCLUDE_BOOST=ON -DCMAKE_BUILD_TYPE="${CMAKE_BUILD_TYPE}" .. && \
+    make -j "$(nproc)" package
```
fgerlits, line 63: why is `scl enable devtoolset-10` not needed any longer?

lordgamez: I use `source /opt/rh/devtoolset-10/enable` to have it enabled in this context.

docker/DockerBuild.sh
lordgamez: In the plain docker build we have multiple stages and use the `--target build` to create the build image then copy the ccache files from that image if needed, then continue with the second stage for the released image. In the other distro builds we don't have separate build and release images, because we don't use those images for running MiNiFi only get the built binaries from it.

fgerlits: Could we add stages to all Dockerfiles so we can use them in the same way?

lordgamez: I would rather not, stages are usually used if we need one or more helper images to create the final image, this way the layers of the helper stages only created as temporary layers that can be cleaned up afterwards. In the centos and other distro docker images we do not really have any logically distinguished stages and I would remove these stages in the focal, fedora, bionic Dockerfiles as well in the future.

docker/DockerBuild.sh
lordgamez: Extracted to a separate function in 6af554983dc9bfb2cdff69fd134f909374a876e9

.github/workflows/ci.yml
lordgamez: Updated in 87fd6d00ed72d55b17f0d682da239187ce415250

fgerlits (approved):

adam-markovics (approved):

adamdebreceni (approved):

PR #1377, opened 2022-07-27
MINIFICPP-1865 Integrate performance clang-tidy checks in CI

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-1865

-------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/http-curl/tests/unit/InvokeHTTPTests.cpp
+++ b/extensions/http-curl/tests/unit/InvokeHTTPTests.cpp
@@ -103,7 +103,7 @@ TEST_CASE("HTTPTestsWithNoResourceClaimPOST", "[httptest1]") {
   auto records = reporter->getEvents();
   auto record = session->get();
   REQUIRE(record == nullptr);
-  REQUIRE(records.size() == 0);
+  REQUIRE(records.empty());
 
   reporter = session->getProvenanceReporter();
 
@@ -117,7 +117,7 @@ TEST_CASE("HTTPTestsWithNoResourceClaimPOST", "[httptest1]") {
   session->commit();
   records = reporter->getEvents();
   // FIXME(fgerlits): this test is very weak, as `records` is empty
-  for (auto provEventRecord : records) {
+  for (const auto& provEventRecord : records) {
     REQUIRE(provEventRecord->getComponentType() == TestHTTPServer::PROCESSOR_NAME);
   }
 
@@ -197,7 +197,7 @@ TEST_CASE("HTTPTestsWithResourceClaimPOST", "[httptest1]") {
   auto records = reporter->getEvents();
   record = session->get();
   REQUIRE(record == nullptr);
-  REQUIRE(records.size() == 0);
+  REQUIRE(records.empty());
 
   listenhttp->incrementActiveTasks();
   listenhttp->setScheduledState(core::ScheduledState::RUNNING);
@@ -215,7 +215,7 @@ TEST_CASE("HTTPTestsWithResourceClaimPOST", "[httptest1]") {
   session2->commit();
   records = reporter->getEvents();
   // FIXME(fgerlits): this test is very weak, as `records` is empty
-  for (auto provEventRecord : records) {
+  for (const auto& provEventRecord : records) {
     REQUIRE(provEventRecord->getComponentType() == listenhttp->getName());
   }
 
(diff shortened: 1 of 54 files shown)


extensions/http-curl/tests/unit/InvokeHTTPTests.cpp
```diff
@@ -238,14 +238,14 @@ TEST_CASE("HTTPTestsPostNoResourceClaim", "[httptest1]") {
   auto records = plan->getProvenanceRecords();
   std::shared_ptr<core::FlowFile> record = plan->getCurrentFlowFile();
   REQUIRE(record == nullptr);
-  REQUIRE(records.size() == 0);
+  REQUIRE(records.empty());
```
adamdebreceni, line 241: I feel like these transformations in `REQUIRE` are not advantageous, as in the former case we get an error like "error: 3 != 0", while in this new case we get "<...> is false"

lordgamez: I don't think we really care about the number of events in this test. The real question here is that if got any events or not. So if we break it and get any events here the count doesn't really help with the debugging IMO.

extensions/standard-processors/tests/integration/TestExecuteProcess.cpp
```diff
@@ -87,17 +87,15 @@ int main(int /*argc*/, char ** /*argv*/) {
...
-  }
+  processor_workers.push_back(std::thread([processor, test_repo, &is_ready]() {
+    auto node = std::make_shared<core::ProcessorNode>(processor.get());
+    auto context = std::make_shared<core::ProcessContext>(node, nullptr, test_repo, test_repo);
+    context->setProperty(org::apache::nifi::minifi::processors::ExecuteProcess::Command, "sleep 0.5");
+    auto session = std::make_shared<core::ProcessSession>(context);
+    while (!is_ready.load(std::memory_order_relaxed)) {
+    }
+    processor->onTrigger(context.get(), session.get());
+  }));
```
adamdebreceni, line 98: I don't think these cases are interchangeable (I mean functionally sure they are), we should rework the whole test if we are sure that a single trigger is enough, I recon there was some idea behind testing multiple triggers, but for some reason was not materialized

libminifi/src/c2/C2Client.cpp
```diff
@@ -177,7 +177,7 @@ std::shared_ptr<state::response::ResponseNode> C2Client::loadC2ResponseConfigura
             std::static_pointer_cast<state::response::ObjectNode>(prev_node)->add_node(new_node);
           }
         } else {
-          std::string optionName = option + "." + name;
+          std::string optionName = std::string(option).append(".").append(name);
```
adamdebreceni, line 180: should we go with `StringUtils::join_pack` for such string concatenations?

extensions/standard-processors/tests/integration/TestExecuteProcess.cpp
lordgamez: I agree it should be reworked and rewritten in the future, I added a jira ticket for this: https://issues.apache.org/jira/browse/MINIFICPP-1893

libminifi/src/c2/C2Client.cpp
lordgamez: Good idea, updated in 3b7a34d4fbb31fbb01acfa3654723ea05239f0fa

adamdebreceni (approved):

martinzink (approved):

libminifi/test/unit/ExpectedTest.cpp
```diff
@@ -52,7 +52,7 @@ TEST_CASE("expected map", "[expected][map]") {
 
   {
     const nonstd::expected<int, int> e = 21;
-    auto ret = std::move(e) | utils::map(mul2);
+    auto ret = e | utils::map(mul2);
```
szaszm, line 55: These are meant to test utilities on rvalues. You can see the same tests above on lvalues. Please revert these changes, except for the last one.

libminifi/test/unit/IdTests.cpp
```diff
@@ -77,7 +77,7 @@ TEST_CASE("Test Generate Move", "[id]") {
 
   auto generated = generator->generate();
   auto str = generated.to_string();
-  utils::Identifier moved = std::move(generated);
+  utils::Identifier moved = generated;
```
szaszm, line 80: This change is also removing an intentional move.

szaszm (changes requested):

libminifi/src/c2/C2Agent.cpp
```diff
@@ -739,7 +739,7 @@ utils::TaskRescheduleInfo C2Agent::produce() {
         std::make_move_iterator(payload_batch.end()),
         [&] (C2Payload&& payload) {
           try {
-            C2Payload && response = protocol_.load()->consumePayload(std::move(payload));
+            C2Payload && response = protocol_.load()->consumePayload(payload);
```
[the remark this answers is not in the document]
lordgamez: Updated in 7d86623adbe9d1cf018a62e72536bce62cd4332d

libminifi/test/unit/ExpectedTest.cpp
lordgamez: Good catch, updated in 7d86623adbe9d1cf018a62e72536bce62cd4332d

libminifi/test/unit/IdTests.cpp
lordgamez: Updated in 7d86623adbe9d1cf018a62e72536bce62cd4332d

szaszm (approved):

PR #1381, opened 2022-08-04
MINIFICPP-1898 - Fix ListenHTTP test failure

adam-markovics:
https://issues.apache.org/jira/browse/MINIFICPP-1898

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/civetweb/processors/ListenHTTP.cpp
+++ b/extensions/civetweb/processors/ListenHTTP.cpp
@@ -410,6 +410,20 @@ bool ListenHTTP::Handler::handleHead(CivetServer* /*server*/, struct mg_connecti
   return true;
 }
 
+bool ListenHTTP::Handler::handlePut(CivetServer* /*server*/, struct mg_connection* conn) {
+  mg_printf(conn, "HTTP/1.1 405 Method Not Allowed\r\n"
+                  "Content-Type: text/html\r\n"
+                  "Content-Length: 0\r\n\r\n");
+  return true;
+}
+
+bool ListenHTTP::Handler::handleDelete(CivetServer* /*server*/, struct mg_connection* conn) {
+  mg_printf(conn, "HTTP/1.1 405 Method Not Allowed\r\n"
+                  "Content-Type: text/html\r\n"
+                  "Content-Length: 0\r\n\r\n");
+  return true;
+}
+
 void ListenHTTP::Handler::setResponseBody(const ResponseBody& response) {
   std::lock_guard<std::mutex> guard(uri_map_mutex_);
 
--- a/extensions/civetweb/processors/ListenHTTP.h
+++ b/extensions/civetweb/processors/ListenHTTP.h
@@ -114,6 +114,8 @@ class ListenHTTP : public core::Processor {
     bool handlePost(CivetServer *server, struct mg_connection *conn) override;
     bool handleGet(CivetServer *server, struct mg_connection *conn) override;
     bool handleHead(CivetServer *server, struct mg_connection *conn) override;
+    bool handlePut(CivetServer *server, struct mg_connection *conn) override;
+    bool handleDelete(CivetServer *server, struct mg_connection *conn) override;
 
     /**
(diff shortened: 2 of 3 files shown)


fgerlits (approved):
Can you put some explanation in either the PR description or the Jira, please?  Was the default implementation of these two HTTP handlers no-op in the older CivetWeb version, and changed to `assert(false)` in the new version?

adam-markovics:
CivetWeb's default implementations for request handlers (including PUT and DELETE) return error codes, indicating they didn't handle the request (returning `false`). Then control flow goes on a strange way, using `goto` to jump inside a function. Earlier version of the library didn't have an assert on this path, the new one does. It's saying "we have a handler, but we don't". So my answer to this is providing overriding handlers in the user code.

Also commented on the JIRA ticket.

fgerlits (changes requested):
Not strictly related, but there is a Clang Tidy check failure in `ListenHTTPTests.cpp`, line 183:
```
/home/runner/work/nifi-minifi-cpp/nifi-minifi-cpp/extensions/civetweb/tests/ListenHTTPTests.cpp:183:15:
error: the 'empty' method should be used to check for emptiness instead of comparing to an empty object [readability-container-size-empty,-warnings-as-errors]
      REQUIRE("" == response_body);
              ^~~~~~~~~~~~~~~~~~~
              response_body.empty()
```

I think this small fix could be included in this PR.  Either changing the check to `.empty()` or adding a `NOLINT` comment would be fine.

adam-markovics:
> Not strictly related, but there is a Clang Tidy check failure in `ListenHTTPTests.cpp`, line 183:
> 
> ```
> /home/runner/work/nifi-minifi-cpp/nifi-minifi-cpp/extensions/civetweb/tests/ListenHTTPTests.cpp:183:15:
> error: the 'empty' method should be used to check for emptiness instead of comparing to an empty object [readability-container-size-empty,-warnings-as-errors]
>       REQUIRE("" == response_body);
>               ^~~~~~~~~~~~~~~~~~~
>               response_body.empty()
> ```
> 
> I think this small fix could be included in this PR. Either changing the check to `.empty()` or adding a `NOLINT` comment would be fine.

Good Idea. I fixed it.

fgerlits (approved):

martinzink (approved):

szaszm:
While this makes the test green, ListenHTTPTests are also running awfully long now, for almost 2 minutes, regularly. Could you have a look into why is this the case? It used to take about 10 seconds (which is still on the longer side), or 1/10 of the current runtime, before the issue was introduced in 287fe85.

lordgamez (approved):

adam-markovics:
> While this makes the test green, ListenHTTPTests are also running awfully long now, for almost 2 minutes, regularly. Could you have a look into why is this the case? It used to take about 10 seconds (which is still on the longer side), or 1/10 of the current runtime, before the issue was introduced in [287fe85](https://github.com/apache/nifi-minifi-cpp/commit/287fe853822880c083728d8d03b00f9c5c1321ef).
> 
> edit: To be clear, I tested [287fe85](https://github.com/apache/nifi-minifi-cpp/commit/287fe853822880c083728d8d03b00f9c5c1321ef), the issue was introduced in the commit after that, [0bf8ce9](https://github.com/apache/nifi-minifi-cpp/commit/0bf8ce968bdebfa6867eb339049bfabca19e5df0).

It looks like this comes from the CivetWeb version upgrade. Just changing that makes ListenHTTPTests run faster. I'm looking if we can do anything on the user side.

adam-markovics:
This pull request doesn't make ListenHTTPTests run slower. On main it is the same time (1 minute 45 seconds) as with this PR, measured in release.
In debug, main fails very fast, while this PR runs for roughly the same time as in release, and it passes.
If we want to understand why it's running slower than 287fe85, it's something CivetWeb internal, and very not obvious to understand.
I think it is a separate issue and should be handled in a different PR if we dedicate resources to that.

PR #1392, opened 2022-08-17
MINIFICPP-1900 Make the (Python part of the) script extension compile on Windows

fgerlits:
https://issues.apache.org/jira/browse/MINIFICPP-1900

I have tested it with Python only, I'll take a look at Lua scripting later in a separate pull request.

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/script/python/PythonCreator.h
+++ b/extensions/script/python/PythonCreator.h
@@ -129,7 +129,8 @@ class PythonCreator : public minifi::core::CoreComponent {
       return "";
     }
     const auto python_package_path = std::filesystem::relative(pythonscript, basePath).parent_path();
-    std::vector<std::string> path_elements{python_package_path.begin(), python_package_path.end()};
+    std::vector<std::string> path_elements;
+    std::transform(python_package_path.begin(), python_package_path.end(), std::back_inserter(path_elements), [](const auto& path) { return path.string(); });
     std::string python_package = minifi::utils::StringUtils::join(".", path_elements);
     if (python_package.length() > 1 && python_package.at(0) == '.') {
       python_package = python_package.substr(1);
--- a/extensions/script/python/PythonCreator.h
+++ b/extensions/script/python/PythonCreator.h
@@ -130,6 +130,7 @@ class PythonCreator : public minifi::core::CoreComponent {
     }
     const auto python_package_path = std::filesystem::relative(pythonscript, basePath).parent_path();
     std::vector<std::string> path_elements;
+    path_elements.reserve(std::distance(python_package_path.begin(), python_package_path.end()));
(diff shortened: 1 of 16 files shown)


extensions/script/python/PythonCreator.h
```diff
@@ -129,7 +129,8 @@ class PythonCreator : public minifi::core::CoreComponent {
       return "";
     }
     const auto python_package_path = std::filesystem::relative(pythonscript, basePath).parent_path();
-    std::vector<std::string> path_elements{python_package_path.begin(), python_package_path.end()};
+    std::vector<std::string> path_elements;
+    std::transform(python_package_path.begin(), python_package_path.end(), std::back_inserter(path_elements), [](const auto& path) { return path.string(); });
```
szaszm, line 133:
```suggestion
    std::vector<std::string> path_elements;
    path_elements.reserve(std::distance(python_package_path.begin(), python_package_path.end()));
    std::transform(python_package_path.begin(), python_package_path.end(), std::back_inserter(path_elements), [](const auto& path) { return path.string(); });
```

libminifi/include/agent/agent_version.h
```diff
@@ -18,6 +18,8 @@
 #ifndef LIBMINIFI_INCLUDE_AGENT_AGENT_VERSION_H_
 #define LIBMINIFI_INCLUDE_AGENT_AGENT_VERSION_H_
 
+#undef COMPILER
+
```
szaszm, line 22: Why is this needed? Could you add a comment?

fgerlits: TBH, I don't know why it's needed.  Something somewhere `#define`s `COMPILER` to the MSVC version, but I couldn't find what and where.

I'll take another look and add the comment if I find out, but I didn't think `// something somewhere #defines COMPILE to the MSVC version on Windows` would be a useful comment.

szaszm: That's fine, thank you.

fgerlits: I have found the culprit (it's https://github.com/python/cpython/blob/3.10/PC/pyconfig.h) and added a comment.

extensions/script/python/PythonCreator.h
fgerlits: done in ce7b901964cb0c53365c40ba7d5f00fdd701665a

adam-markovics (approved):

szaszm (approved):

lordgamez (approved):

lordgamez:
The MacOS I job failure looks to be connected to this change, could you check what caused that failure?

lordgamez (approved):

PR #1398, opened 2022-08-19
MINIFICPP-1915 Fix clang-tidy for newly created branch

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-1915

-----------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -151,13 +151,20 @@ jobs:
         run: cd build && make flake8
       - id: files
         uses: Ana06/get-changed-files@v2.1.0
+        continue-on-error: true
       - name: clang-tidy
         run: |
+          if [[ -n "${{ steps.files.outputs.all }}" ]]; then
+            FILES="${{ steps.files.outputs.all }}"
+          else
+            git fetch origin main && git checkout -b main origin/main || true
+            FILES=$(git diff --name-only main ${{ github.sha }} --)
+          fi
           # https://stackoverflow.com/questions/58466701/clang-tidy-cant-locate-stdlib-headers
           sed -i -e 's/\/usr\/lib\/ccache\/clang++-14/\/lib\/llvm-14\/bin\/clang++/g' build/compile_commands.json
           sed -i -e 's/\/usr\/lib\/ccache\/clang-14/\/lib\/llvm-14\/bin\/clang/g' build/compile_commands.json
 
-          ./run_clang_tidy.sh "${{ steps.files.outputs.all }}"
+          ./run_clang_tidy.sh "${FILES}"
   centos:
     name: "centos"
     runs-on: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -151,13 +151,20 @@ jobs:
...
       - id: files
         uses: Ana06/get-changed-files@v2.1.0
+        continue-on-error: true
       - name: clang-tidy
         run: |
+          if [[ -n "${{ steps.files.outputs.all }}" ]]; then
+            FILES="${{ steps.files.outputs.all }}"
+          else
+            git fetch origin main && git checkout -b main origin/main || true
+            FILES=$(git diff --name-only main ${{ github.sha }} --)
```
adamdebreceni, line 161: so on first PR action run this will run clang-tidy for all changed files (relative to main) and then next runs will run clang-tidy for all files relative to the previous commit?

lordgamez: Yes, at least on push triggers.

adamdebreceni: would it make sense to always check relative to main? I'm not sure what the average time clang-tidy takes

.github/workflows/ci.yml
```diff
@@ -151,13 +151,20 @@ jobs:
         run: cd build && make flake8
       - id: files
         uses: Ana06/get-changed-files@v2.1.0
+        continue-on-error: true
       - name: clang-tidy
         run: |
+          if [[ -n "${{ steps.files.outputs.all }}" ]]; then
+            FILES="${{ steps.files.outputs.all }}"
+          else
+            git fetch origin main && git checkout -b main origin/main || true
```
adam-markovics, line 160: I think that checkout is unnecessary, we could run `git diff` right against `origin/main` after `git fetch` has run.

.github/workflows/ci.yml
lordgamez: Well it's relative to the number of cpp files changed, so it can be long if we do it for every single commit. I would stick with checking only the newly changed files per pushes, as we will run on every file eventually on PRs.

.github/workflows/ci.yml
lordgamez: 'origin/main' may not exist if we push a new branch on the apache repository, but you are right that we should not checkout the main and run the clang tidy checks that afterwards. I fixed it by re-checking out the original commit in fd61850c10ec6a4597d9d7e56ded80e1bcb76942

.github/workflows/ci.yml
adamdebreceni: makes sense

adamdebreceni (approved):

adam-markovics (approved):

fgerlits (approved):

martinzink (approved):

szaszm (approved):

PR #1411, opened 2022-08-29
MINIFICPP-1931 Stabilize ResourceQueueTests

lordgamez:
Due to `std::set` not being thread safe, manipulating it in the test from multiple threads could leave it in an inconsistent state. In some cases this resulted in a state where the set contained the 2 correct members, but the `size()` call returned 3.

https://issues.apache.org/jira/browse/MINIFICPP-1931

-----------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/test/unit/ResourceQueueTests.cpp
+++ b/libminifi/test/unit/ResourceQueueTests.cpp
@@ -32,11 +32,13 @@ TEST_CASE("maximum_number_of_creatable_resources", "[utils::ResourceQueue]") {
   std::shared_ptr<core::logging::Logger> logger_{core::logging::LoggerFactory<ResourceQueue<int>>::getLogger()};
   LogTestController::getInstance().setTrace<ResourceQueue<int>>();
 
+  std::mutex resources_created_mutex;
   std::set<int> resources_created;
 
   auto worker = [&](int value, const std::shared_ptr<ResourceQueue<int>>& resource_queue) {
     auto resource = resource_queue->getResource([value]{return std::make_unique<int>(value);});
     std::this_thread::sleep_for(10ms);
+    std::lock_guard<std::mutex> lock(resources_created_mutex);
     resources_created.emplace(*resource);
   };
 
--- a/libminifi/test/unit/ResourceQueueTests.cpp
+++ b/libminifi/test/unit/ResourceQueueTests.cpp
@@ -62,13 +62,13 @@ TEST_CASE("Resource limitation is not set to the resource queue", "[utils::Resou
   auto resource_queue = ResourceQueue<int>::create(std::nullopt, logger_);
   std::set<int> resources_created;
 
-  auto resource_one = resource_queue->getResource([]{return std::make_unique<int>(1);});
-  auto resource_two = resource_queue->getResource([]{return std::make_unique<int>(2);});
-  auto resource_three = resource_queue->getResource([]{return std::make_unique<int>(3);});
+  auto resource_wrapper_one = resource_queue->getResource([]{return std::make_unique<int>(1);});
+  auto resource_wrapper_two = resource_queue->getResource([]{return std::make_unique<int>(2);});
+  auto resource_wrapper_three = resource_queue->getResource([]{return std::make_unique<int>(3);});
 
-  resources_created.emplace(*resource_one);
-  resources_created.emplace(*resource_two);
-  resources_created.emplace(*resource_three);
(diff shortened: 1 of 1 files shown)


fgerlits (approved):
Thanks!  I asked @martinzink to do this, but fixing it instead of pestering someone else to fix it is also an option.  :-)

Also, @martinzink found another, rarer, race condition in the "No Maximum resources" section of the test, please talk to him about it.

lordgamez:
> Thanks! I asked @martinzink to do this, but fixing it instead of pestering someone else to fix it is also an option. :-)
> 
> Also, @martinzink found another, rarer, race condition in the "No Maximum resources" section of the test, please talk to him about it.

I was just there stabilizng tests when I saw this so I just took it :)

I talked to @martinzink and fixed the concurrency issue in 92dcdb522e440cd6aa910286f4f36a7ccd4aadaa

martinzink (approved):

libminifi/test/unit/ResourceQueueTests.cpp
```diff
@@ -53,23 +55,26 @@ TEST_CASE("maximum_number_of_creatable_resources", "[utils::ResourceQueue]") {
...
+  auto resource_three = resource_queue->getResource([]{return std::make_unique<int>(3);});
 
-    CHECK(!resources_created.empty());
-    CHECK(!LogTestController::getInstance().contains("Waiting for resource", 0ms));
-    CHECK(LogTestController::getInstance().contains("Number of instances: 3", 0ms));
-    CHECK(resources_created.size() <= 3);
-  }
+  resources_created.emplace(*resource_one);
+  resources_created.emplace(*resource_two);
+  resources_created.emplace(*resource_three);
```
fgerlits, line 73: I don't think we need the `resources_created` set, as the `resource_x` variables hold onto the wrappers already.

fgerlits (changes requested):
I thought adding a Comment review would undo the previous Approve, but apparently not.

libminifi/test/unit/ResourceQueueTests.cpp
```diff
@@ -28,19 +28,21 @@ using namespace std::literals::chrono_literals;
...
 
   auto worker = [&](int value, const std::shared_ptr<ResourceQueue<int>>& resource_queue) {
     auto resource = resource_queue->getResource([value]{return std::make_unique<int>(value);});
     std::this_thread::sleep_for(10ms);
+    std::lock_guard<std::mutex> lock(resources_created_mutex);
     resources_created.emplace(*resource);
   };
 
-  SECTION("Maximum 2 resources") {
+  SECTION("") {
```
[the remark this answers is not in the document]
lordgamez: Updated in 5f0605b2e6a9017c1768f37968c050b9e551f812

libminifi/test/unit/ResourceQueueTests.cpp
fgerlits: Maybe I'm missing something, but I don't see why we need to copy the resources into the set.  Would this not do the same?
```c++
TEST_CASE("Resource limitation is not set to the resource queue", "[utils::ResourceQueue]") {
  std::shared_ptr<core::logging::Logger> logger_{core::logging::LoggerFactory<ResourceQueue<int>>::getLogger()};
  LogTestController::getInstance().setTrace<ResourceQueue<int>>();
  LogTestController::getInstance().clear();

  auto resource_queue = ResourceQueue<int>::create(std::nullopt, logger_);

  auto resource_wrapper_one = resource_queue->getResource([]{return std::make_unique<int>(1);});
  auto resource_wrapper_two = resource_queue->getResource([]{return std::make_unique<int>(2);});
  auto resource_wrapper_three = resource_queue->getResource([]{return std::make_unique<int>(3);});

  CHECK(!LogTestController::getInstance().contains("Waiting for resource", 0ms));
  CHECK(LogTestController::getInstance().contains("Number of instances: 3", 0ms));
  CHECK(resources_created.size() == 3);
}
```

lordgamez: We put 3 different resources in the queue 1, 2 and 3, and if we add those to the set and the set size is 3 then we can be sure that none of the resources was reused, but new resources were created for every instance. I mean you are right that asserting on the `"Number of instances: 3"` is kind of the same, but I would not rely on log messages if not necessary.

fgerlits (approved):

adam-markovics (approved):

PR #1430, opened 2022-10-06
MINIFICPP-1922 Implement ListenUDP processor

martinzink:

### For all changes:
     in the commit message?

### For code changes:

### Note:


lordgamez (approved):

extensions/standard-processors/tests/unit/ListenUDPTests.cpp
```diff
@@ -0,0 +1,114 @@
...
+  SingleProcessorTestController controller{listen_udp};
+  LogTestController::getInstance().setTrace<ListenUDP>();
+  REQUIRE(listen_udp->setProperty(ListenUDP::Port, std::to_string(PORT)));
+  REQUIRE(listen_udp->setProperty(ListenUDP::MaxBatchSize, "2"));
+
+  controller.plan->scheduleProcessor(listen_udp);
+  REQUIRE(utils::sendUdpDatagram({"test_message_1"}, endpoint));
+  REQUIRE(utils::sendUdpDatagram({"another_message"}, endpoint));
+  ProcessorTriggerResult result;
+  REQUIRE(controller.triggerUntil({{ListenUDP::Success, 2}}, result, 300ms, 50ms));
```
adamdebreceni, line 61: it seems to me `triggerUntil` also accepts more flowfiles than expected, should we restrict this, or verify here the exact size of `result.at(ListenUDP::Success)`?

extensions/standard-processors/processors/ListenUDP.cpp
```diff
@@ -0,0 +1,83 @@
...
+        ->withType(core::StandardValidators::get().LISTEN_PORT_VALIDATOR)
+        ->isRequired(true)
+        ->build());
+
+const core::Property ListenUDP::MaxQueueSize(
+    core::PropertyBuilder::createProperty("Max Size of Message Queue")
+        ->withDescription("Maximum number of messages allowed to be buffered before processing them when the processor is triggered. "
+                          "If the buffer is full, the message is ignored. If set to zero the buffer is unlimited.")
+        ->withDefaultValue<uint64_t>(10000)
+        ->isRequired(true)
```
adam-markovics, line 38: If it has a default value, does it make sense to require it? If it is empty, it will fall back to default.

extensions/standard-processors/processors/ListenUDP.cpp
```diff
@@ -0,0 +1,83 @@
...
+        ->build());
+
+const core::Property ListenUDP::MaxBatchSize(
+    core::PropertyBuilder::createProperty("Max Batch Size")
+        ->withDescription("The maximum number of messages to process at a time.")
+        ->withDefaultValue<uint64_t>(500)
+        ->isRequired(true)
```
adam-markovics, line 45: Same as above.

extensions/standard-processors/processors/ListenUDP.h
```diff
@@ -0,0 +1,62 @@
...
+ * limitations under the License.
+ */
+#pragma once
+
+#include <memory>
+#include <string>
+
+#include "NetworkListenerProcessor.h"
+#include "core/logging/LoggerConfiguration.h"
+#include "utils/Enum.h"
```
adam-markovics, line 24: I don't see this being used.

extensions/standard-processors/tests/unit/ListenUDPTests.cpp
```diff
@@ -0,0 +1,114 @@
...
+#include "controllers/SSLContextService.h"
+#include "range/v3/algorithm/contains.hpp"
+
+using ListenUDP = org::apache::nifi::minifi::processors::ListenUDP;
+
+using namespace std::literals::chrono_literals;
+
+namespace org::apache::nifi::minifi::test {
+
+constexpr uint64_t PORT = 10256;
```
adam-markovics, line 32: Is this just some random number to hard wire? So would this test fail if some service is using this port? Do we get an easy-to-interpret error message then? Also, `uint16_t` is enough for ports.

extensions/standard-processors/processors/ListenUDP.cpp
arpadboda: I think in C2 perspective it makes sense

extensions/standard-processors/processors/ListenUDP.h
martinzink: good catch removed it in https://github.com/apache/nifi-minifi-cpp/pull/1430/commits/f7822f8282a3882520f88b437e291fff40066c4c#diff-4c39ae13ae374b8f23e28ecc9bc64b27953b919a0facdad5e5bc8e209a353529L24

extensions/standard-processors/processors/ListenUDP.cpp
martinzink: I think it helps in understanding the property, and NiFi does the same thing https://nifi.apache.org/docs/nifi-docs/components/org.apache.nifi/nifi-standard-nar/1.17.0/org.apache.nifi.processors.standard.ListenUDP/index.html (bold (required and has default value)

extensions/standard-processors/processors/ListenUDP.cpp
martinzink: Same as above. :)

extensions/standard-processors/processors/ListenUDP.cpp
```diff
@@ -0,0 +1,83 @@
...
+}
+
+void ListenUDP::onSchedule(const std::shared_ptr<core::ProcessContext>& context, const std::shared_ptr<core::ProcessSessionFactory>&) {
+  gsl_Expects(context);
+  startUdpServer(*context);
+}
+
+void ListenUDP::transferAsFlowFile(const utils::net::Message& message, core::ProcessSession& session) {
+  auto flow_file = session.create();
+  session.writeBuffer(flow_file, message.message_data);
```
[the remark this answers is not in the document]
martinzink: Considering the message length is limited by the maximum size of the UDP datagram (`A UDP datagram is carried in a single IP packet and is hence limited to a maximum payload of 65,507 bytes for IPv4 and 65,527 bytes for IPv6.`) , I think that won't be necessary.

extensions/standard-processors/tests/unit/ListenUDPTests.cpp
martinzink: Good idea, I've changed this (and the tests working with similar hard coded port numbers in https://github.com/apache/nifi-minifi-cpp/pull/1430/commits/f7822f8282a3882520f88b437e291fff40066c4c).

extensions/standard-processors/tests/unit/ListenUDPTests.cpp
martinzink: you are right, it makes sense I've added a check in https://github.com/apache/nifi-minifi-cpp/pull/1430/commits/f7822f8282a3882520f88b437e291fff40066c4c#diff-f4b79b10812ae57099687c8d9d2f7171ce2c686ed9b0f58a459be4711099a09bR62

adam-markovics (approved):

adamdebreceni (approved):

martinzink:
I've tried creating a test utility that creates a socket on port 0. https://github.com/apache/nifi-minifi-cpp/pull/1430/commits/f7822f8282a3882520f88b437e291fff40066c4c (so the OS assigns a random port to it), then close the socket and return the port to bind to.

Unfortunately this could leave the port on TIME_WAIT (which could be worked-around by setting reuse_address (which we may very well want to use besides this, so restarting the processor on the same port can work without hiccups), but due to race conditions sometimes the port is not even in the TIME_WAIT state when the processor wants to bind to it.

The proper solution would be to directly set the processor (or utility) to port 0. And get the listening port out after the socket/acceptor is already up and running, but this would require a larger refactor (both tests and prod code as well), since I am already in the process to rework these codes in [MINIFICPP-1979](https://issues.apache.org/jira/browse/MINIFICPP-1979), it would make sense to do this during that refactor.

@adam-markovics @arpadboda How about I revert the relevant parts from https://github.com/apache/nifi-minifi-cpp/pull/1430/commits/f7822f8282a3882520f88b437e291fff40066c4c and properly implement the feature in/after [MINIFICPP-1979](https://issues.apache.org/jira/browse/MINIFICPP-1979) I've already created a ticket for this [MINIFICPP-1985](https://issues.apache.org/jira/browse/MINIFICPP-1985)

adam-markovics (approved):

szaszm (approved):

PR #1439, opened 2022-10-20
MINIFICPP-1967 Add batch processing of lines in TailFile

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-1967

------------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


adam-markovics (approved):

PROCESSORS.md
```diff
@@ -2424,6 +2424,7 @@ In the list below, the names of required properties appear in bold. Any other pr
...
+| **Batch Size**             | 0                 |                                                        | Maximum number of lines to process in a single trigger. If set to 0 all new lines will be processed.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               |
```
szaszm, line 2427: I think it would make sense to combine this with another property that controls whether a batch is emitted as a single flow file or one flow file per line.

extensions/standard-processors/tests/unit/TailFileTests.cpp
```diff
@@ -1847,3 +1847,38 @@ TEST_CASE("TailFile can use an AttributeProviderService", "[AttributeProviderSer
...
+  std::stringstream state_file;
+  state_file << dir << utils::file::get_separator() << STATE_FILE;
+
+  plan->setProperty(tailfile, org::apache::nifi::minifi::processors::TailFile::FileName.getName(), temp_file.str());
+  plan->setProperty(tailfile, org::apache::nifi::minifi::processors::TailFile::Delimiter.getName(), "\n");
+  plan->setProperty(tailfile, org::apache::nifi::minifi::processors::TailFile::BatchSize.getName(), "10");
+
+  testController.runSession(plan);
+
+  REQUIRE(LogTestController::getInstance().contains("Logged 10 flow files"));
```
martinzink, line 1881: I think it would be better to utilize the https://github.com/apache/nifi-minifi-cpp/blob/main/libminifi/test/SingleProcessorTestController.h then you could directly check the number of flowfiles tailfile generates per trigger, instead of the indirect LogAttribute approach, and this test could be much simpler.

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -135,6 +135,13 @@ const core::Property TailFile::AttributeProviderService(
         ->asType<minifi::controllers::AttributeProviderService>()
         ->build());
 
+const core::Property TailFile::BatchSize(
+    core::PropertyBuilder::createProperty("Batch Size")
+        ->withDescription("Maximum number of lines to process in a single trigger. If set to 0 all new lines will be processed.")
```
adamdebreceni, line 140: should we change this to "number of flowfiles emitted" or similar? with the input delimiter the user could break the "flowfile == line" correspondence

PROCESSORS.md
lordgamez: If the delimiter already control how the lines are processed, is it okay if we only change the description as @adamdebreceni suggested?

extensions/standard-processors/tests/unit/TailFileTests.cpp
lordgamez: Good idea, updated in 24f9ab5c32083fea259b4b1e0889bd93f20e8cda

extensions/standard-processors/processors/TailFile.cpp
lordgamez: Good point, updated in 24f9ab5c32083fea259b4b1e0889bd93f20e8cda

PROCESSORS.md
szaszm: I would update the property name as well. This leaves the door open for a future property that controls how many lines are emitted in a single flow file, with a similar same name.

lordgamez: After reading the comments I don't think we can really talk about "lines" specifically in this processor, as the flow file contents are fully controlled by the delimiter property. So I'm not sure it would be a good idea to have a separate property that controls lines specifically. The `Batch Size` property name is consistently used in all processors to mark the number of flow files emitted at once by the processor so it should represent the same here.

szaszm: My bad, the Batch Size property indeed means the same in all other processors.

szaszm (approved):

martinzink (approved):

arpadboda (approved):
LGTM

PR #1442, opened 2022-10-20
MINIFICPP-1952 InvokeHTTP POST/PUT should support large files

fgerlits:
https://issues.apache.org/jira/browse/MINIFICPP-1952

When doing POST/PUT in InvokeHTTP, read the flow file contents in chunks during the upload instead of reading the whole thing at once into memory.

The change to `catch.hpp` is not part of this pull request, but of #1441.  That PR should be merged first, after which the `catch.hpp` change will disappear from this PR.

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/src/utils/BaseHTTPClient.cpp
+++ b/libminifi/src/utils/BaseHTTPClient.cpp
@@ -250,17 +250,15 @@ size_t HTTPUploadByteArrayInputCallback::setPosition(int64_t offset) {
 size_t HTTPUploadStreamContentsCallback::getDataChunk(char *data, size_t size) {
   logger_->log_trace("HTTPUploadStreamContentsCallback is asked for up to %zu bytes", size);
 
-  std::vector<std::byte> buffer(size);
-  size_t num_read = input_stream_->read(buffer);
+  gsl::span<char> buffer{data, size};
+  size_t num_read = input_stream_->read(buffer.as_span<std::byte>());
 
   if (io::isError(num_read)) {
     logger_->log_error("Error reading the input stream in HTTPUploadStreamContentsCallback");
     return 0;
   }
-  logger_->log_debug("HTTPUploadStreamContentsCallback is returning %zu bytes", num_read);
 
-  gsl_Expects(num_read <= size);
-  memcpy(data, buffer.data(), num_read);
+  logger_->log_debug("HTTPUploadStreamContentsCallback is returning %zu bytes", num_read);
   return num_read;
 }
 
--- a/libminifi/src/utils/BaseHTTPClient.cpp
+++ b/libminifi/src/utils/BaseHTTPClient.cpp
@@ -160,4 +160,118 @@ std::string URL::toString() const {
   }
 }
 
+/**
+ * Receive HTTP Response.
+ */
+size_t HTTPRequestResponse::receiveWrite(char *data, size_t size, size_t nmemb, void *p) {
+  try {
+    if (p == nullptr) {
+      return CALLBACK_ABORT;
+    }
+    auto *callback = static_cast<HTTPReadCallback *>(p);
+    if (callback->stop) {
+      return CALLBACK_ABORT;
+    }
+    callback->write(data, (size * nmemb));
+    return (size * nmemb);
+  } catch (...) {
+    return CALLBACK_ABORT;
+  }
+}
+
+/**
+ * Callback for post, put, and patch operations
+ * @param data output buffer to write to
+ * @param size number of elements to write
+ * @param nmemb size of each element to write
+ * @param p input object to read from
+ */
+size_t HTTPRequestResponse::send_write(char *data, size_t size, size_t nmemb, void *p) {
+  try {
+    if (p == nullptr) {
+      return CALLBACK_ABORT;
+    }
+    auto *callback = reinterpret_cast<HTTPUploadCallback *>(p);
+    return callback->getDataChunk(data, size * nmemb);
+  } catch (...) {
+    return CALLBACK_ABORT;
+  }
+}
+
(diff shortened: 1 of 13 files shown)


libminifi/src/utils/BaseHTTPClient.cpp
```diff
@@ -160,4 +160,118 @@ std::string URL::toString() const {
...
+  }
+  return 0;
+}
+
+size_t HTTPUploadByteArrayInputCallback::setPosition(int64_t offset) {
+  if (stop) {
+    return HTTPRequestResponse::SEEKFUNC_FAIL;
+  }
+  if (getBufferSize() <= static_cast<size_t>(offset)) {
+    return HTTPRequestResponse::SEEKFUNC_FAIL;
```
adamdebreceni, line 243: I see that this is the same as the original version (so we shouldn't tinker with it in this PR), but do you think it is ok to fail a seek to the end of the buffer?

libminifi/src/utils/BaseHTTPClient.cpp
```diff
@@ -160,4 +160,118 @@ std::string URL::toString() const {
...
+  pos = offset;
+  seek(pos);
+  return HTTPRequestResponse::SEEKFUNC_OK;
+}
+
+size_t HTTPUploadStreamContentsCallback::getDataChunk(char *data, size_t size) {
+  logger_->log_trace("HTTPUploadStreamContentsCallback is asked for up to %zu bytes", size);
+
+  std::vector<std::byte> buffer(size);
+  size_t num_read = input_stream_->read(buffer);
```
adamdebreceni, line 254: can't we read directly into `data`? might there be a requirement to not mess up `data` if we can't read the full size (or maybe on error)?

libminifi/src/utils/BaseHTTPClient.cpp
```diff
@@ -160,4 +160,118 @@ std::string URL::toString() const {
...
+}
+
+size_t HTTPUploadStreamContentsCallback::setPosition(int64_t offset) {
+  if (offset == 0) {
+    logger_->log_debug("HTTPUploadStreamContentsCallback is ignoring request to rewind to the beginning");
+  } else {
+    logger_->log_warn("HTTPUploadStreamContentsCallback is ignoring request to seek to position %" PRId64, offset);
+  }
+
+  return HTTPRequestResponse::SEEKFUNC_OK;
```
adamdebreceni, line 274: some streams (like `FileStream`) might properly support `seek`s, could not calling `seek` cause problems? (corrupted data, failed request etc.) should we return `SEEKFUNC_FAIL` here?

libminifi/src/utils/BaseHTTPClient.cpp
fgerlits: I don't know if there is such a requirement.  You're right, it's probably safe to read into the buffer directly.  I'll change it.

libminifi/src/utils/BaseHTTPClient.cpp
fgerlits: Yeah, it's not clear to me when and why curl would call seek during an upload.  One thing I've seen during testing was a call to `seek(0)` _after_ the upload; if this call returns `SEEKFUNC_FAIL` then the upload fails, but if I don't do anything just return `SEEKFUNC_OK` then it works fine.

What do you think we should do?

libminifi/src/utils/BaseHTTPClient.cpp
adamdebreceni: yes a Jira would be great

fgerlits: https://issues.apache.org/jira/browse/MINIFICPP-1971

libminifi/src/utils/BaseHTTPClient.cpp
adamdebreceni: I see you have added it to this jira (https://issues.apache.org/jira/browse/MINIFICPP-1971), we will investigate there

adamdebreceni (approved):

szaszm (approved):

libminifi/src/utils/BaseHTTPClient.cpp
```diff
@@ -160,4 +160,116 @@ std::string URL::toString() const {
...
+    size_t len = buffer_size - pos;
+    if (len <= 0) {
+      return 0;
+    }
+    auto *ptr = getBuffer(pos);
+
+    if (ptr == nullptr) {
+      return 0;
+    }
+    if (len > size)
```
adam-markovics, line 228: I would use `std::min` instead

libminifi/src/utils/BaseHTTPClient.cpp
```diff
@@ -160,4 +160,116 @@ std::string URL::toString() const {
...
+}
+
+size_t HTTPUploadByteArrayInputCallback::getDataChunk(char *data, size_t size) {
+  if (stop) {
+    return HTTPRequestResponse::CALLBACK_ABORT;
+  }
+  size_t buffer_size = getBufferSize();
+  if (pos <= buffer_size) {
+    size_t len = buffer_size - pos;
+    if (len <= 0) {
```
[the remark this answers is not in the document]
fgerlits: This is pre-existing code, I'd rather not change it in this pull request.  I have added your suggestions to https://issues.apache.org/jira/browse/MINIFICPP-1971.

libminifi/src/utils/BaseHTTPClient.cpp
fgerlits: Also added to https://issues.apache.org/jira/browse/MINIFICPP-1971.

adam-markovics (approved):

PR #1444, opened 2022-10-28
MINIFICPP-1952 Reset the callback when no longer needed

fgerlits:
Without this, the callback would remain in the `HTTPClient` stored in the `ResourceQueue` in the `InvokeHTTP` processor, but the stream inside the callback would be pointing to a flow file resource claim which no longer exists.

This went mostly unnoticed, except in two of the test cases (`HTTPTestsResponseBodyinAttribute` and `HTTPTestsResponseBody`) in the MacOS CI job, which got stuck in the `InvokeHTTP` destructor.  Thanks, AppleClang!

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/http-curl/processors/InvokeHTTP.cpp
+++ b/extensions/http-curl/processors/InvokeHTTP.cpp
@@ -334,6 +334,10 @@ void InvokeHTTP::onTriggerWithClient(const std::shared_ptr<core::ProcessContext>
 
   logger_->log_debug("onTrigger InvokeHTTP with %s to %s", client.getRequestMethod(), client.getURL());
 
+  const auto remove_callback_from_client_at_exit = gsl::finally([&client] {
+    client.setUploadCallback({});
+  });
+
   std::string transaction_id = utils::IdGenerator::getIdGenerator()->generate().to_string();
 
   if (shouldEmitFlowFile(client)) {


adam-markovics (approved):

martinzink (approved):

extensions/http-curl/processors/InvokeHTTP.cpp
```diff
@@ -334,6 +334,10 @@ void InvokeHTTP::onTriggerWithClient(const std::shared_ptr<core::ProcessContext>
 
   logger_->log_debug("onTrigger InvokeHTTP with %s to %s", client.getRequestMethod(), client.getURL());
 
+  const auto remove_callback_from_client_at_exit = gsl::finally([&client] {
+    client.setUploadCallback({});
```
adamdebreceni, line 338: I feel like this is the responsibility of `ResourceQueue` and should be handled in `ResourceQueue::returnResource`, what I am thinking is a `resetter: std::function<void(ResourceType&)>` (probably an extra constructor argument to `ResourceQueue`) which when set would be called in `returnResource`, what do you think?

lordgamez (approved):

extensions/http-curl/processors/InvokeHTTP.cpp
adamdebreceni: we can address this and any related issues with ResourceQueue in https://issues.apache.org/jira/browse/MINIFICPP-1973

adamdebreceni (approved):

arpadboda (approved):

PR #1449, opened 2022-11-09
MINIFICPP-1978 - Flush MergeContent bundles when its size would grow beyond max group size

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/libarchive/BinFiles.h
```diff
@@ -87,8 +87,10 @@ class Bin {
       }
     }
 
-    if ((queued_data_size_ + flow->getSize()) > maxSize_ || (queue_.size() + 1) > maxEntries_)
+    if ((queued_data_size_ + flow->getSize()) > maxSize_ || (queue_.size() + 1) > maxEntries_) {
+      closed_ = true;
```
szaszm, line 91: Do you think it would make sense to penalize flow files that are larger than the max group size? Currently if there is such a flow file, it will keep flushing out all bins regardless of how "full" they are.

adamdebreceni: incoming flow files that are larger than max group size are immediately assigned their own bin, and flushed by themselves, so they won't be a problem, flow files that are large but not max size large, could cause the flush of a single bin, as we only try to insert into a single bin (the last in the group's queue) before creating a new bin for it

szaszm (approved):

lordgamez (approved):

martinzink (approved):

adam-markovics:
Could you please update BinFiles documentation in `PROCESSORS.md`? To me it is not clear what this Processor is doing, and it could also be confusing for end users. Currently description is: "Bins flow files into buckets based on the number of entries or size of entries". I don't understand the following verbs and nouns in this context : "bins", "buckets", "entries". Please add some more explanation. Property description should also be updated, for example format is missing for Max Bin Age.

adamdebreceni:
> Could you please update BinFiles documentation in `PROCESSORS.md`? To me it is not clear what this Processor is doing, and it could also be confusing for end users. Currently description is: "Bins flow files into buckets based on the number of entries or size of entries". I don't understand the following verbs and nouns in this context : "bins", "buckets", "entries". Please add some more explanation. Property description should also be updated, for example format is missing for Max Bin Age.

definitely something we should address, created a ticket for it: https://issues.apache.org/jira/browse/MINIFICPP-1982

adam-markovics (approved):

PR #1459, opened 2022-11-24
MINIFICPP-1987 Configuring processor metrics with regular expressions

lordgamez:
Select processor metrics with regular expressions using `processorMetrics/` prefix.

Also fixes the issue that only processor metrics were allowed to be requested specifically in the DescribeMetrics operation. Now system metrics can also be specified in the `metricsClass` operation argument.

https://issues.apache.org/jira/browse/MINIFICPP-1987

---------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


C2.md
```diff
@@ -117,6 +117,10 @@ a configuration of an agent
     nifi.c2.root.class.definitions.metrics.metrics.processorMetrics.name=ProcessorMetric
     nifi.c2.root.class.definitions.metrics.metrics.processorMetrics.classes=GetFileMetrics
 
+Processor metrics can also be configured using regular expressions with the `processorMetrics/` prefix, so the following definition is also valid:
```
szaszm, line 120: I would go for one of `processor/` or `processors/` prefixes instead, because it's obvious from the context it's used in that this is filtering metrics classes.

lordgamez: The reason I went with this instead is because we are filtering metrics not processors, so one would think that the `processors/` prefix means we have to specify `processors/TailFile` or `processors/GetFile` for example. In fact those would not match any processor metrics, because we use regex match instead of regex search in the implementation.

The other reason is that in this example it is obvious that we are referring to metrics, but in this context `nifi.c2.root.class.definitions.metrics` is an arbitrary user defined class, so the class may not contain the `metrics` in the class definition. In case the metrics are defined in an another C2 node like `nifi.c2.root.class.definitions.mynodes` class for example or requested in the `nifi.c2.root.classes` it may not be so obvious we want to filter for metrics. C2 can support any response node that may or may not be metric nodes.

fgerlits (approved):

szaszm (approved):

PR #1460, opened 2022-11-28
MINIFICPP-1995 Add configuring path for flowfile_checkpoint directory

lordgamez:
Added `nifi.flowfile.checkpoint.directory.default` property with default value `${MINIFI_HOME}/flowfile_checkpoint`
If repositories are set to empty value the default value is used instead of failure.

https://issues.apache.org/jira/browse/MINIFICPP-1995

-------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/CONFIGURE.md
+++ b/CONFIGURE.md
@@ -119,7 +119,7 @@ for TCP and secure HTTPS communications.
             name: tominifi
             max concurrent tasks: 2
             Properties:
-        SSL Context Service: SSLServiceName
+                SSL Context Service: SSLServiceName
     Controller Services:
     - name: SSLServiceName
       id: 2438e3c8-015a-1000-79ca-83af40ec1974
--- a/CONFIGURE.md
+++ b/CONFIGURE.md
@@ -13,10 +13,10 @@
   limitations under the License.
 -->
 
-# Apache NiFi -  Configuring MiNiFi - C++ 
+# Apache NiFi -  Configuring MiNiFi - C++
 
 ## Configuring
-The 'conf' directory in the root contains a template config.yml document.  
+The 'conf' directory in the root contains a template config.yml document.
 
 This is partly compatible with the format used with the Java MiNiFi application. MiNiFi C++ is currently compatible with version 1 of the MiNiFi YAML schema.
 Additional information on the YAML format for the config.yml and schema versioning can be found in the [MiNiFi System Administrator Guide](https://nifi.apache.org/minifi/system-admin-guide.html).
@@ -73,7 +73,7 @@ It's recommended to create your configuration in YAML format or configure the ag
 ### Scheduling strategies
 Currently Apache NiFi MiNiFi C++ supports TIMER_DRIVEN, EVENT_DRIVEN, and CRON_DRIVEN. TIMER_DRIVEN uses periods to execute your processor(s) at given intervals.
 The EVENT_DRIVEN strategy awaits for data be available or some other notification mechanism to trigger execution. CRON_DRIVEN executes at the desired intervals
-based on the CRON periods. Apache NiFi MiNiFi C++ supports standard CRON expressions without intervals ( */5 * * * * ). 
+based on the CRON periods. Apache NiFi MiNiFi C++ supports standard CRON expressions without intervals ( */5 * * * * ).
 
 ### SiteToSite Security Configuration
 
@@ -94,14 +94,14 @@ based on the CRON periods. Apache NiFi MiNiFi C++ supports standard CRON express
 
     if you do not want to enable client certificate base authorization
     nifi.security.need.ClientAuth=false
-  
-You have the option of specifying an SSL Context Service definition for the RPGs instead of the properties above. 
-This will link to a corresponding SSL Context service defined in the flow. 
-    
-To do this specify the SSL Context Service Property in your RPGs and link it to 
+
+You have the option of specifying an SSL Context Service definition for the RPGs instead of the properties above.
+This will link to a corresponding SSL Context service defined in the flow.
+
+To do this specify the SSL Context Service Property in your RPGs and link it to
 a defined controller service. If you do not take this approach the options, above, will be used
 for TCP and secure HTTPS communications.
-    
+
     Remote Processing Groups:
     - name: NiFi Flow
       id: 2438e3c8-015a-1000-79ca-83af40ec1998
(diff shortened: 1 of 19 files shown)


encrypt-config/tests/resources/minifi.properties
```diff
@@ -24,6 +24,7 @@ nifi.provenance.repository.directory.default=${MINIFI_HOME}/provenance_repositor
 nifi.provenance.repository.max.storage.time=1 MIN
 nifi.provenance.repository.max.storage.size=1 MB
 nifi.flowfile.repository.directory.default=${MINIFI_HOME}/flowfile_repository
+nifi.flowfile.checkpoint.directory.default=${MINIFI_HOME}/flowfile_checkpoint
```
szaszm, line 27: I think we can omit this from test resources if it's the same as the default. It makes sense to add to the example config, but in test resources, it's just redundant.

lordgamez: Updated in 28dfa20abfc159c5da28c85026fae4f459b430bb

examples/kubernetes/daemon-set-log-collection/daemon-set-log-collection.yml
```diff
@@ -30,6 +30,7 @@ data:
     nifi.provenance.repository.max.storage.time=1 MIN
     nifi.provenance.repository.max.storage.size=1 MB
     nifi.flowfile.repository.directory.default=${MINIFI_HOME}/flowfile_repository
+    nifi.flowfile.checkpoint.directory.default=${MINIFI_HOME}/flowfile_checkpoint
```
szaszm, line 33: I think the property is not relevant here, so I would leave it out from the example. I would ideally also remove all other properties that are using their default values, to keep the example minimal, but that doesn't need to happen in this PR.

examples/kubernetes/sidecar-log-collection/sidecar-log-collection.yml
```diff
@@ -32,6 +32,7 @@ data:
     nifi.provenance.repository.max.storage.time=1 MIN
     nifi.provenance.repository.max.storage.size=1 MB
     nifi.flowfile.repository.directory.default=${MINIFI_HOME}/flowfile_repository
+    nifi.flowfile.checkpoint.directory.default=${MINIFI_HOME}/flowfile_checkpoint
```
szaszm, line 35: I think the property is not relevant here, so I would leave it out from the example. I would ideally also remove all other properties that are using their default values, to keep the example minimal, but that doesn't need to happen in this PR.

CONFIGURE.md
```diff
@@ -119,23 +119,23 @@ for TCP and secure HTTPS communications.
             name: tominifi
             max concurrent tasks: 2
             Properties:
-		SSL Context Service: SSLServiceName
```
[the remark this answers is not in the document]
lordgamez: Good catch, updated in cbac8731aa8f068600e1cb76a7fc0b4bdc6e72fc

examples/kubernetes/daemon-set-log-collection/daemon-set-log-collection.yml
lordgamez: Added jira ticket for this impovement: https://issues.apache.org/jira/browse/MINIFICPP-2004

examples/kubernetes/sidecar-log-collection/sidecar-log-collection.yml
lordgamez: Added jira ticket for this impovement: https://issues.apache.org/jira/browse/MINIFICPP-2004

szaszm (approved):

extensions/rocksdb-repos/FlowFileRepository.h
```diff
@@ -100,11 +100,17 @@ class FlowFileRepository : public ThreadedRepository, public SwapManager {
...
+    if (configure->get(Configure::nifi_flowfile_repository_directory_default, value) && !value.empty()) {
       directory_ = value;
     }
     logger_->log_debug("NiFi FlowFile Repository Directory %s", directory_);
 
+    value.clear();
+    if (configure->get(Configure::nifi_flowfile_checkpoint_directory_default, value) && !value.empty()) {
+      checkpoint_dir_ = value;
+    }
+    logger_->log_debug("NiFi FlowFile Checkpoint Directory %s", checkpoint_dir_);
```
fgerlits, line 112: I think you need to either add `.string()` here, or make the conversion templates in `Logger.h` smarter to accept `path`s.

lordgamez: Added fix in 9b7fd9553203ec6bfae427546e0fa214e695f17f

fgerlits (approved):

libminifi/include/core/logging/Logger.h
```diff
@@ -72,6 +76,8 @@ inline decltype(auto) conditional_convert(const Arg& val) {
     return val;
   } else if constexpr (std::is_same_v<decltype(std::declval<const Arg&>().c_str()), const char*>) {
     return val.c_str();
+  } else if constexpr (std::is_same_v<decltype(std::declval<const Arg&>().string()), std::string>) {
+    return val.string().c_str();
```
adamdebreceni, line 80: the `::string() const` method could return a temporary which is destroyed before using the result of `c_str()` leaving us with UB, I think for this to be safe, the `conditional_stringify`  functions should call the `::string() const` method, the returned temporary string is kept alive until the end of the full expression the `conditional_stringify` is part of, which is after `format_string` returns

lordgamez: Good catch, updated in fb658f6abd33164157570344436dddae13700823

adamdebreceni (approved):

fgerlits:
Unfortunately, the code did not compile on Windows after 373dbdd98b16d1a45788c5bb10c926368af0c891, so I have reverted both c07ee3e10c54a19820f7b5ac8d2ee63768ed5759 and 373dbdd98b16d1a45788c5bb10c926368af0c891, and added a `.string()` to the log line: 378501d61d0d0a98c1688f28c7a35161af5e0d3a

lordgamez:
> Unfortunately, the code did not compile on Windows after [373dbdd](https://github.com/apache/nifi-minifi-cpp/commit/373dbdd98b16d1a45788c5bb10c926368af0c891), so I have reverted both [c07ee3e](https://github.com/apache/nifi-minifi-cpp/commit/c07ee3e10c54a19820f7b5ac8d2ee63768ed5759) and [373dbdd](https://github.com/apache/nifi-minifi-cpp/commit/373dbdd98b16d1a45788c5bb10c926368af0c891), and added a `.string()` to the log line: [378501d](https://github.com/apache/nifi-minifi-cpp/commit/378501d61d0d0a98c1688f28c7a35161af5e0d3a)

Thanks for the update, you are right it's better to keep it simple at this point.

PR #1461, opened 2022-11-28
MINIFICPP-1983 Mount files in MiNiFi test containers and additional refactoring

lordgamez:
This PR contains multiple changes and refactoring in the docker test environment:

- Moved docker cluster specific classes to separate `cluster` directory under `test/integration` directory
- minifi.properties and all additional configuration files are now mounted instead of building a separate image for every config
- Added `MinifiOptions` to specify minifi deployment options like additional properties
- New minifi test image is only build if additonal packages are needed (in case of SQL tests)
- Removed inheritance hierarchy for `DockerTestCluster`
- `SingleNodeDockerCluster` renamed to `ContainerStore` and only contains container deployment specific functionality
- Extension specific checkers are extracted to separate checker classes
- Docker communicator extracted for communicated with docker containers
- Moved general container operations to `Container` base class
- Flushing container logs in case of all assertion failures

https://issues.apache.org/jira/browse/MINIFICPP-1983

-------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


docker/test/integration/cluster/checkers/PostgresChecker.py
```diff
@@ -0,0 +1,18 @@
...
+        (code, output) = self.container_communicator.execute_command(postgresql_container_name, ["psql", "-U", "postgres", "-c", query])
+        return code == 0 and str(number_of_rows) + " rows" in output
+
+    def check_query_results(self, postgresql_container_name, query, number_of_rows, timeout_seconds):
+        start_time = time.perf_counter()
+        while (time.perf_counter() - start_time) < timeout_seconds:
+            if self.__query_postgres_server(postgresql_container_name, query, number_of_rows):
+                return True
+            time.sleep(2)
+        return False
```
adamdebreceni, line 18: could we use the `retry_check` wrapper for this method?

docker/test/integration/cluster/containers/Container.py
```diff
@@ -63,13 +63,25 @@ def log_source(self):
         return LogSource.FROM_DOCKER_CONTAINER
 
     def stop(self):
-        raise NotImplementedError()
+        logging.info('Stopping minifi docker container "%s"...', self.name)
```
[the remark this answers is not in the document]
lordgamez: That should be not be minifi specific, fixed log messages in d5329427a1cb89b2cc9d99a85251b62aa0e8a5d4

docker/test/integration/cluster/checkers/PostgresChecker.py
lordgamez: Because the timeout is a function parameter that cannot be used as the parameter of the annotation, so I would stick with this solution.

adamdebreceni: makes sense, thanks!

adamdebreceni (approved):

docker/test/integration/cluster/checkers/AzureChecker.py
```diff
@@ -0,0 +1,70 @@
...
+            time.sleep(1)
+        return False
+
+    def check_azure_blob_storage_is_empty(self, timeout_seconds):
+        start_time = time.perf_counter()
+        while (time.perf_counter() - start_time) < timeout_seconds:
+            if self.__get_blob_and_snapshot_count() == 0:
+                return True
+            time.sleep(1)
+        return False
```
szaszm, line 70: It would probably make sense to check at least once even if timeout is 0.

martinzink (approved):

docker/test/integration/cluster/checkers/AzureChecker.py
lordgamez: Updated in d465264923264b0b3f4c9317ed0f5c50adc9729c

docker/test/integration/cluster/checkers/AzureChecker.py
```diff
@@ -0,0 +1,74 @@
...
+        return len(list(container_client.list_blobs(include=['deleted'])))
+
+    def check_azure_blob_and_snapshot_count(self, blob_and_snapshot_count, timeout_seconds):
+        start_time = time.perf_counter()
+        while True:
+            if self.__get_blob_and_snapshot_count() == blob_and_snapshot_count:
+                return True
+            time.sleep(1)
+            if timeout_seconds < (time.perf_counter() - start_time):
+                break
```
szaszm, line 63: Did you consider extracting the waiting for condition with timeout logic to a higher order function? These loops in d465264 are similar enough, and we could get rid of some code duplication, but I'm not insisting, +1 even if you don't want to change it.

szaszm (approved):

docker/test/integration/cluster/checkers/AzureChecker.py
szaszm: I was thinking about something like this. Untested example, and my python skills are not the greatest, so treat this as pseudocode. :) 
```python
def wait_for(action, timeout, predicate_transform = lambda x: x, check_period = 1):
    start_time = time.perf_counter()
    while True:
        result = action()
        if predicate_transform(result):
            return (True, result)
        time.sleep(check_period)
        if timeout_seconds < (time.perf_counter() - start_time):
            break
    return (False, None)


(happened, _) = wait_for(lambda: self.__get_blob_and_snapshot_count() == blob_and_snapshot_count, timeout_seconds)

# and the more complex example
def get_logs():
    logging.info('Waiting for app-logs `%s` in container `%s`', log_entry, container_name)
    return self.get_app_log(container_name)

def predicate(x):
    (status, logs) = x
    return logs is not None or status == 'exited'
    
(happened, (status, logs)) = wait_for(get_logs, timeout_seconds, predicate)
if not happened or status == 'exited':
    return False
# ...
```

lordgamez: Thanks, I think it's a good idea! Although I only prefer to use it in trivial cases, so I would not complicate the utility function with the result transformation. Updated in 5ca3ac0037c3df330cd8cfe2df58c2b0169483b7

PR #1483, opened 2022-12-23
MINIFICPP-2019 Fix auto-generation of PROCESSORS.md

fgerlits:
https://issues.apache.org/jira/browse/MINIFICPP-2019

The `PROCESSORS.md` file included in this PR is almost auto-generated (by merging the output of `minifi docs ...` on Linux and on Windows), except for a few sections mentioned in https://issues.apache.org/jira/browse/MINIFICPP-2021.

`CollectKubernetesPodMetrics`, `ExecuteJavaProcessor` and `SourceInitiatedSubscriptionListener` were missing from `PROCESSORS.md`; I have added the auto-generated docs of these.

Also, the ordering of the processors was inconsistent; now they are ordered in case-insensitive lexicographic order.

---

Thank you for submitting a contribution to Apache NiFi - MiNiFi C++.

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [x] Is there a JIRA ticket associated with this PR? Is it referenced
     in the commit message?

- [x] Does your PR title start with MINIFICPP-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.

- [x] Has your PR been rebased against the latest commit within the target branch (typically main)?

- [ ] Is your initial contribution a single, squashed commit?

### For code changes:
- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?
- [ ] If applicable, have you updated the LICENSE file?
- [ ] If applicable, have you updated the NOTICE file?

### For documentation related changes:
- [ ] Have you ensured that format looks appropriate for the output in which it is rendered?

### Note:
Please ensure that once the PR is submitted, you check GitHub Actions CI results for build issues and submit an update to your PR as soon as possible.

PR #1492, opened 2023-01-16
MINIFICPP-2021 Auto-generate extra doc sections in PROCESSORS.md

fgerlits:
https://issues.apache.org/jira/browse/MINIFICPP-2021

This is a follow-up on #1483, broken up into two parts to make reviewing somewhat more manageable.

After this PR, everything in PROCESSORS.md is auto-generated except for the `Predefined Groups` section in `PerformanceDataMonitor`, which is unlike any of the other sections, and I didn't want to invent a new category just for this.  Maybe it could be added to the `Description`, but it seemed too large for that.  Let me know what you think.

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/http-curl/processors/InvokeHTTP.cpp
+++ b/extensions/http-curl/processors/InvokeHTTP.cpp
@@ -150,10 +150,10 @@ const core::Relationship InvokeHTTP::RelFailure("failure",
     "timeout or general exception. It will have new attributes detailing the request.");
 
 
-const core::OutputAttribute InvokeHTTP::StatusCode{"invokehttp.status.code", { Success, RelResponse, RelRetry, RelNoRetry }, "The status code that is returned"};
-const core::OutputAttribute InvokeHTTP::StatusMessage{"invokehttp.status.message", { Success, RelResponse, RelRetry, RelNoRetry }, "The status message that is returned"};
-const core::OutputAttribute InvokeHTTP::RequestUrl{"invokehttp.request.url", { Success, RelResponse, RelRetry, RelNoRetry }, "The original request URL"};
-const core::OutputAttribute InvokeHTTP::TxId{"invokehttp.tx.id", { Success, RelResponse, RelRetry, RelNoRetry }, "The transaction ID that is returned after reading the response"};
+const core::OutputAttribute InvokeHTTP::StatusCode{STATUS_CODE, { Success, RelResponse, RelRetry, RelNoRetry }, "The status code that is returned"};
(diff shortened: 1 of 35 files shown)


minifi_main/AgentDocs.cpp
```diff
@@ -43,27 +81,17 @@ void AgentDocs::generate(const std::filesystem::path& docsdir, std::ostream &gen
...
   for (const auto &group : minifi::AgentBuild::getExtensions()) {
     struct Components descriptions = build_description_.getClassDescriptions(group);
-    for (const auto &processorName : descriptions.processors_) {
-      processorSet.insert(std::make_pair(extractClassName(processorName.full_name_), processorName));
+    for (const auto& processor_description : descriptions.processors_) {
+      processorSet.insert(std::make_pair(extractClassName(processor_description.full_name_), processor_description));
     }
   }
   for (const auto &processor : processorSet) {
     const auto& filename = docsdir / processor.first;
```
martinzink, line 89: I think we could also add the .md extension to these files

fgerlits: We could, but then we'd need to remove the extension later, at line 171.  I think it would be better to do this separately, together with either removing or fixing the logic of the `*.extra` files.

martinzink (approved):

extensions/http-curl/processors/InvokeHTTP.cpp
```diff
@@ -149,6 +149,13 @@ const core::Relationship InvokeHTTP::RelFailure("failure",
     "The original FlowFile will be routed on any type of connection failure, "
     "timeout or general exception. It will have new attributes detailing the request.");
 
+
+const core::OutputAttribute InvokeHTTP::StatusCode{"invokehttp.status.code", { Success, RelResponse, RelRetry, RelNoRetry }, "The status code that is returned"};
```
adamdebreceni, line 153: can we use the `STATUS_CODE` and other constants from `InvokeHTTP.h`?

adamdebreceni: we could also work in the other direction in a subsequent PR and use these instead of constant strings

adamdebreceni (approved):

extensions/http-curl/processors/InvokeHTTP.cpp
fgerlits: good idea, fixed in b520a96bc29abfb4019ae92cd167ce806b06a4c4

PR #1495, opened 2023-01-20
MINIFICPP-2036 Upgrade gcc to version 11

fgerlits:
- on centos: use devtoolset-11 instead of devtoolset-10
- on alpine: use version 3.16 instead of 3.13
- ~~on both: do not install the boost package, as it is no longer required~~

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -56,6 +56,7 @@ RUN apk --no-cache add gcc \
   libusb-dev \
   gpsd-dev \
   python3-dev \
+  boost-dev \
   doxygen \
   ccache \
   lua-dev
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -16,7 +16,7 @@
 # under the License.
 #
 
-ARG BASE_ALPINE_IMAGE="alpine:3.13"
+ARG BASE_ALPINE_IMAGE="alpine:3.16"
 
 # Build image
 FROM ${BASE_ALPINE_IMAGE} AS build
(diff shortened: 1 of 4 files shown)


docker/Dockerfile
```diff
@@ -16,7 +16,7 @@
 # under the License.
 #
 
-ARG BASE_ALPINE_IMAGE="alpine:3.13"
+ARG BASE_ALPINE_IMAGE="alpine:3.16"
```
szaszm, line 19: Why not 3.17.1, the current latest?

fgerlits: because alpine 3.17 uses gcc 12, which would not match the title of this PR

we could update this and other CI jobs to gcc 12 (except for centos 7, which doesn't have a `devtoolset-12` yet)

but maybe do that in a separate PR?

szaszm (approved):

docker/Dockerfile
```diff
@@ -55,7 +56,6 @@ RUN apk --no-cache add gcc \
   libusb-dev \
   gpsd-dev \
   python3-dev \
-  boost-dev \
```
lordgamez, line 58: I think the boost library is still needed for the Bustache extension if that is build with the image

fgerlits: you're right, I missed that -- fixed in 0732b67f91282dc048b9be357456f166c733e053

lordgamez (approved):

martinzink (approved):

adamdebreceni (approved):

PR #1497, opened 2023-01-26
MINIFICPP-1869 Metrics-related refactoring

lordgamez:
Notable changes:
- FlowController ise separated from C2Client
- C2Client is separated by functionality to C2Agent and C2MetricsPublisher
- MetricsPublisherStore is introduced to instantiate metrics publishers when configured through MetricsPublisherFactory
- Response nodes are ensured to be not null
- ConnectionStore is aggregated instead of inherited from
- Root process group functionalities and flow change operations are moved to RootProcessGroupWrapper
- ResponseNodeLoader is shared between metrics publishers
- FlowConfiguration is shared between FlowController and ResponseNodeLoader
- Unnecessary contructor parameters are removed from FlowController, metrics and C2 are made optional
- Controller service provider update is centralized to the extracted FlowController::updateFromPayload function
- ResponseNodeLoader's state now contains a pointer to the root process group
- RestSender initialization crash avoided when controller is nullptr

https://issues.apache.org/jira/browse/MINIFICPP-1869

-----------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/src/FlowController.cpp
```diff
@@ -205,13 +177,12 @@ int16_t FlowController::stop() {
...
      * -After the threadpool is stopped we can make sure that processors don't need repos and controllers anymore */
-    if (this->root_) {
-      this->root_->drainConnections();
-    }
+    root_wrapper_.drainConnections();
     this->flow_file_repo_->stop();
     this->provenance_repo_->stop();
     // stop the ControllerServices
     disableAllControllerServices();
+    initialized_ = false;
```
adamdebreceni, line 185: although `unload` is equivalent to calling `stop` and setting `initialized_`, conceptually the flow controller has the following state transitions: `empty` -(`load`)-> `initialized` -(`start`)-> `running` -(`stop`)-> `initialized` -(`unload`)-> `empty`

IMO we should either keep `unload` even if it can be inlined, or also remove `load`

lordgamez: I think the `unload` was just complicating things as even in tests either `unload` or `stop` (or both of them together) was used  for the same functionality. `load` is a bit different as it also calls `stop` for the flow controller and reloads the flow controller if a new config is applied so I would keep that. I see what the original intent was, but it does not reflect the current functionality, so I would remove `unload` only.

libminifi/include/core/state/nodes/MetricsBase.h
```diff
@@ -37,28 +37,28 @@ namespace org::apache::nifi::minifi::state::response {
 class ResponseNode : public core::Connectable, public PublishedMetricProvider {
  public:
   ResponseNode()
-      : core::Connectable("metric"),
-        is_array_(false) {
+    : core::Connectable("metric"),
```
martinzink, line 40: Why did you change these? Shouldnt it be 4 space indent from the level of the constructor (which starts at 2)?

libminifi/include/core/state/nodes/MetricsBase.h
```diff
@@ -37,28 +37,28 @@ namespace org::apache::nifi::minifi::state::response {
...
-      : core::Connectable(std::move(name), uuid),
-        is_array_(false) {
+    : core::Connectable(std::move(name), uuid),
+      is_array_(false) {
   }
 
   ~ResponseNode() override = default;
 
-  static std::vector<state::response::SerializedResponseNode> serializeAndMergeResponseNodes(const std::vector<std::shared_ptr<ResponseNode>>& nodes) {
+  static std::vector<SerializedResponseNode> serializeAndMergeResponseNodes(const std::vector<gsl::not_null<std::shared_ptr<ResponseNode>>>& nodes) {
```
adamdebreceni, line 56: should we use the `SharedResponseNode` alias here as well?

libminifi/include/core/state/nodes/MetricsBase.h
```diff
@@ -159,7 +161,7 @@ class ObjectNode : public ResponseNode {
 class ResponseNodeSource {
  public:
   virtual ~ResponseNodeSource() = default;
-  virtual std::shared_ptr<ResponseNode> getResponseNodes() = 0;
+  virtual SharedResponseNode getResponseNodes() = 0;
```
adamdebreceni, line 164: can this return more than one response nodes? should we drop the `s`?

adamdebreceni (approved):

libminifi/include/core/state/nodes/MetricsBase.h
lordgamez: You are right, I thought the double indentation is only needed if it conflicts with the indentation of the implementation code under it, but the [style guide](https://google.github.io/styleguide/cppguide.html#Constructor_Initializer_Lists) insists on 4 spaces all times, updated in c71bdaa3378b7087a33b7bd5f9a6bf4c3dd8cff8

libminifi/include/core/state/nodes/MetricsBase.h
lordgamez: Updated in c71bdaa3378b7087a33b7bd5f9a6bf4c3dd8cff8

libminifi/include/core/state/nodes/MetricsBase.h
lordgamez: Good point, updated in c71bdaa3378b7087a33b7bd5f9a6bf4c3dd8cff8

libminifi/src/core/state/nodes/ResponseNodeLoader.cpp
```diff
@@ -25,89 +25,129 @@
...
 
-  if (!root) {
+  std::lock_guard<std::mutex> guard(root_mutex_);
+  if (!root_) {
     return;
   }
 
   std::vector<core::Processor*> processors;
-  root->getAllProcessors(processors);
+  root_->getAllProcessors(processors);
```
fgerlits, line 78: could we reduce the scope of this lock?  e.g:
```c++
  std::vector<core::Processor*> processors;
  {
    std::lock_guard<std::mutex> guard(root_mutex_);
    if (!root_) {
      return;
    }
    root_->getAllProcessors(processors);
  }
```

libminifi/src/core/state/nodes/ResponseNodeLoader.cpp
```diff
@@ -25,89 +25,129 @@
...
   if (repository_metrics != nullptr) {
     repository_metrics->addRepository(provenance_repo_);
     repository_metrics->addRepository(flow_file_repo_);
   }
 }
 
-void ResponseNodeLoader::initializeQueueMetrics(const std::shared_ptr<ResponseNode>& response_node, core::ProcessGroup* root) {
-  if (!root) {
+void ResponseNodeLoader::initializeQueueMetrics(const SharedResponseNode& response_node) const {
+  std::lock_guard<std::mutex> guard(root_mutex_);
```
fgerlits, line 128: I would reduce the scope of this lock to just the `getConnections()` call, too.

libminifi/test/flow-tests/TestControllerWithFlow.h
```diff
@@ -64,14 +65,12 @@ class TestControllerWithFlow: public TestController {
...
+    auto flow = std::make_shared<core::YamlConfiguration>(core::ConfigurationContext{ff_repo, content_repo, stream_factory, configuration_, yaml_path_.string()});
     auto root = flow->getRoot();
     root_ = root.get();
-    controller_ = std::make_shared<minifi::FlowController>(
-        prov_repo, ff_repo, configuration_,
-        std::move(flow),
-        content_repo, DEFAULT_ROOT_GROUP_NAME,
-        std::make_shared<utils::file::FileSystem>(), []{});
+    auto metrics_publisher_store = std::make_unique<minifi::state::MetricsPublisherStore>(configuration_, prov_repo, ff_repo, flow);
+    metrics_publisher_store_ = metrics_publisher_store.get();
```
fgerlits, line 72: Why do we save `metrics_publisher_store_`?  I couldn't find any place where we use it.  Also, saving a bare pointer to an object owned by someone else looks dangerous.

libminifi/include/FlowController.h
```diff
@@ -216,27 +185,29 @@ class FlowController : public core::controller::ForwardingControllerServiceProvi
   std::unique_ptr<FlowControlProtocol> protocol_;
   std::chrono::steady_clock::time_point start_time_;
 
- private:
+ protected:
```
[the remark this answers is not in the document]
lordgamez: As I checked all of it the protected members can be private, updated in 3d374f175fb0263a5148db8e785bbf21954823fb

libminifi/include/c2/C2MetricsPublisher.h
```diff
@@ -24,69 +24,53 @@
...
   std::optional<state::response::NodeReporter::ReportedNode> getMetricsNode(const std::string& metrics_class) const override;
   std::vector<state::response::NodeReporter::ReportedNode> getHeartbeatNodes(bool include_manifest) const override;
 
-  void stopC2();
-  void initializeResponseNodes(core::ProcessGroup* root);
-  void clearResponseNodes();
+  /**
+   * Retrieves the agent manifest to be sent as a response to C2 DESCRIBE manifest
+   * @return the agent manifest response node
+   */
```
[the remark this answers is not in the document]
lordgamez: Updated in 3d374f175fb0263a5148db8e785bbf21954823fb

libminifi/include/c2/C2Utils.h
```diff
@@ -0,0 +1,34 @@
...
+#include <string>
+#include <memory>
+
+#include "properties/Configure.h"
+#include "utils/StringUtils.h"
+
+namespace org::apache::nifi::minifi::c2 {
+
+static constexpr const char* UPDATE_NAME = "C2UpdatePolicy";
+static constexpr const char* C2_METRICS_PUBLISHER = "C2MetricsPublisher";
```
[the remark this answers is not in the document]
lordgamez: Updated in 3d374f175fb0263a5148db8e785bbf21954823fb

libminifi/src/core/state/MetricsPublisherFactory.cpp
```diff
@@ -0,0 +1,45 @@
...
+std::unique_ptr<MetricsPublisher> createMetricsPublisher(const std::string& name, const std::shared_ptr<Configure>& configuration,
+    const std::shared_ptr<state::response::ResponseNodeLoader>& response_node_loader) {
+  auto ptr = core::ClassLoader::getDefaultClassLoader().instantiate(name, name);
+  if (!ptr) {
+    throw std::runtime_error("Configured metrics publisher class \"" + name + "\" could not be instantiated.");
+  }
+
+  auto metrics_publisher = utils::dynamic_unique_cast<MetricsPublisher>(std::move(ptr));
+  if (!metrics_publisher) {
+    throw std::runtime_error("Configured metrics publisher class \"" + name + "\" could not be instantiated.");
```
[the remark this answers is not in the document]
lordgamez: Updated in 3d374f175fb0263a5148db8e785bbf21954823fb

libminifi/src/core/state/MetricsPublisherFactory.cpp
```diff
@@ -0,0 +1,45 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "core/state/MetricsPublisherFactory.h"
+
+namespace org::apache::nifi::minifi::state {
+
+std::unique_ptr<MetricsPublisher> createMetricsPublisher(const std::string& name, const std::shared_ptr<Configure>& configuration,
```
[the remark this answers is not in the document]
lordgamez: Updated in 3d374f175fb0263a5148db8e785bbf21954823fb

libminifi/src/core/state/nodes/ResponseNodeLoader.cpp
lordgamez: We should not reduce the scope in this case, because the underlying resources of the returned raw pointers can be freed while we iterate over them, so we should keep the lock until they are processed.

libminifi/src/core/state/nodes/ResponseNodeLoader.cpp
lordgamez: This should be kept due to the same issue as above.

libminifi/test/flow-tests/TestControllerWithFlow.h
lordgamez: The raw pointer is used in the `PythonManifestTests` to retrieve the C2 metrics publisher through it. Not the cleanest solution, but I could not find a better one. The raw pointer should have the same lifetime as the `controller_`, its owner so it should be okay.

libminifi/src/core/state/nodes/ResponseNodeLoader.cpp
fgerlits: Yes, good point.  Maybe we could collect the metrics into a local map under the `root_mutex_` lock, and separately (after releasing the `root_mutex_` lock) insert them into `component_metrics_` under the `component_metrics_mutex_`?  Locking a mutex while already holding another mutex can easily cause problems later on.

lordgamez: Good point, it can be an improvement, updated in 1c8b48094649174a87b7dcf41d4ee6f1bf1e78bf

fgerlits (approved):

PR #1498, opened 2023-01-26
MINIFICPP-2020 Improve error messages when MiNiFi is not able to start

fgerlits:
This came out of investigating https://issues.apache.org/jira/browse/MINIFICPP-2020, and doesn't solve any of the problems (also, it's not clear whether there are any problems really), but at least it improves the error messages.

In particular, it was quite bad that in one case MiNiFi just quit without giving any error messages at all.  Also, I did not like the fact that `minifi docs` (without any arguments) started up MiNiFi in normal mode, ignoring the "docs" argument, instead of telling the user to give additional arguments.

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/minifi_main/MiNiFiMain.cpp
+++ b/minifi_main/MiNiFiMain.cpp
@@ -133,9 +133,9 @@ int main(int argc, char **argv) {
 #ifdef WIN32
   RunAsServiceIfNeeded();
 
-  bool isStartedByService = false;
-  HANDLE terminationEventHandler = GetTerminationEventHandle(&isStartedByService);
-  if (terminationEventHandler == nullptr) {
+  auto [isStartedByService, terminationEventHandler] = GetTerminationEventHandle();
+  if (isStartedByService && !terminationEventHandler) {
+    std::cerr << "Fatal error: started by service, but could not create the termination event handler\n";
     return -1;
   }
 
@@ -150,9 +150,10 @@ int main(int argc, char **argv) {
 #ifdef WIN32
   if (isStartedByService) {
     if (!CreateServiceTerminationThread(logger, terminationEventHandler)) {
+      logger->log_error("Fatal error: started by service, but could not create the service termination thread");
       return -1;
     }
-  } else {
+  } else if (terminationEventHandler) {
     CloseHandle(terminationEventHandler);
   }
 #endif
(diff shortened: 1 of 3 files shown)


minifi_main/MiNiFiWindowsService.h
```diff
@@ -20,10 +20,12 @@
 #ifdef WIN32
 
 #include <memory>
+#include <utility>
+
 #include "core/Core.h"
 
 void RunAsServiceIfNeeded();
-HANDLE GetTerminationEventHandle(bool* isStartedByService);
+std::tuple<bool, HANDLE> GetTerminationEventHandle();
```
szaszm, line 28: I think a trivial aggregate struct would be better, since it would allow naming the parts.

fgerlits: we only call this function from one place, and the parts are named where we call it, so I'm not sure it's worth it

did you mean something like this? https://github.com/fgerlits/nifi-minifi-cpp/commit/02a2b692c4d924dfbb522ac0eb71a007c84dbd69

fgerlits: or I guess we could do just this: https://github.com/fgerlits/nifi-minifi-cpp/commit/c81662e94abefbea9ac993a2cdb551d16ee55ecd -- still not sure it's worth it

szaszm: You can still use structured bindings at the call site.

minifi_main/MiNiFiMain.cpp
```diff
@@ -258,7 +259,11 @@ int main(int argc, char **argv) {
 
     minifi::core::extension::ExtensionManager::get().initialize(configure);
 
-    if (argc >= 3 && std::string("docs") == argv[1]) {
+    if (argc >= 2 && std::string("docs") == argv[1]) {
+      if (argc == 2) {
```
adamdebreceni, line 263: would it make sense to also check for too many arguments? e.g. if the user accidentally added a space in one of the paths (or if it already contains spaces but is not properly escaped)

adamdebreceni (approved):

minifi_main/MiNiFiWindowsService.h
fgerlits: OK, done in c81662e94abefbea9ac993a2cdb551d16ee55ecd

minifi_main/MiNiFiMain.cpp
fgerlits: yes, good idea, done in 4388f346770f907b725052893d8063d33ab9a6e3

martinzink (approved):

lordgamez (approved):

PR #1499, opened 2023-01-30
MINIFICPP-1372 - Allow async content deletion

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/CONFIGURE.md
+++ b/CONFIGURE.md
@@ -364,7 +364,7 @@ When multiple repositories use the same directory (as with `minifidb://` scheme)
 
 When a flow file content is no longer needed we can specify the deletion strategy.
 
-    # any value other than 0 enables async cleanup with the specified period
+    # any value other than 0 enables garbage collection with the specified frequency
     # while a value of 0 sec triggers an immediate deletion as soon as the resource
     # is not needed
     # (the default value is 1 sec)
--- a/CONFIGURE.md
+++ b/CONFIGURE.md
@@ -360,6 +360,17 @@ Finally, as the last line is commented out, it will make the state manager use p
 
 When multiple repositories use the same directory (as with `minifidb://` scheme) they should either be all plaintext or all encrypted with the same key.
 
+### Configuring Repository Cleanup
+
+When a flow file content is no longer needed we can specify the deletion strategy.
+
+    # any value other than 0 enables async cleanup with the specified period
+    # while a value of 0 sec triggers an immediate deletion as soon as the resource
+    # is not needed
+    # (the default value is 1 sec)
+    nifi.database.content.repository.purge.period = 1 sec
+
+
 ### Configuring Volatile and NO-OP Repositories
 Each of the repositories can be configured to be volatile ( state kept in memory and flushed
  upon restart ) or persistent. Currently, the flow file and provenance repositories can persist
--- a/conf/minifi.properties
+++ b/conf/minifi.properties
@@ -38,7 +38,7 @@ nifi.content.repository.class.name=DatabaseContentRepository
 # nifi.flowfile.repository.rocksdb.compaction.period=2 min
 # nifi.database.content.repository.rocksdb.compaction.period=2 min
 
-# setting this value to 0 enables synchronous deletion
+# setting this value to "0" enables synchronous deletion
 # nifi.database.content.repository.purge.period = 1 sec
 
 #nifi.remote.input.secure=true
--- a/conf/minifi.properties
+++ b/conf/minifi.properties
@@ -38,6 +38,9 @@ nifi.content.repository.class.name=DatabaseContentRepository
 # nifi.flowfile.repository.rocksdb.compaction.period=2 min
 # nifi.database.content.repository.rocksdb.compaction.period=2 min
 
+# setting this value to 0 enables synchronous deletion
+# nifi.database.content.repository.purge.period = 1 sec
+
 #nifi.remote.input.secure=true
 #nifi.security.need.ClientAuth=
 #nifi.security.client.certificate=
--- a/extensions/rocksdb-repos/DatabaseContentRepository.cpp
+++ b/extensions/rocksdb-repos/DatabaseContentRepository.cpp
@@ -251,6 +251,9 @@ void DatabaseContentRepository::runGc() {
     if (!opendb) {
       continue;
     }
+    // keys_to_delete_ is not persisted, in memory only, and is lost on restart
+    // the clearOrphans method is executed during agent startup making sure that this
+    // does not cause a content leak
     std::vector<std::string> keys;
     {
       std::lock_guard guard(keys_mtx_);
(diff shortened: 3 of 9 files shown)


libminifi/include/core/ContentRepository.h
```diff
@@ -28,20 +28,18 @@
...
 #include "utils/GeneralUtils.h"
+#include "core/Repository.h"
 
 namespace org::apache::nifi::minifi::core {
 
 /**
  * Content repository definition that extends StreamManager.
  */
-class ContentRepository : public StreamManager<minifi::ResourceClaim>, public utils::EnableSharedFromThis<ContentRepository> {
+class ContentRepository : public virtual core::Repository, public StreamManager<minifi::ResourceClaim>, public utils::EnableSharedFromThis<ContentRepository> {
```
lordgamez, line 38: Why is the inheritance from core::Repository needed for ContentRepository? There is a discussion about a common base class introduced in #1490 (https://github.com/apache/nifi-minifi-cpp/pull/1490#discussion_r1093458349). A suggested change is that core::Repository should be renamed as it represents the a key-value based repository while content repo has a stream based interface so it should not be a common base class. This will probably conflict with #1485 maybe we should try rebasing this change on that.

adamdebreceni: I am fine with waiting until the mentioned PRs are merged

adamdebreceni: in light of changes since and future plans, rewrote this PR to not use inheritance

extensions/rocksdb-repos/DatabaseContentRepository.cpp
```diff
@@ -198,27 +212,63 @@ bool DatabaseContentRepository::exists(const minifi::ResourceClaim &streamId) {
...
+      std::lock_guard guard(keys_mtx_);
+      keys = std::exchange(keys_to_delete_, std::vector<std::string>{});
+    }
+    auto batch = opendb->createWriteBatch();
+    for (auto& key : keys) {
+      batch.Delete(key);
+    }
+    rocksdb::Status status;
+    status = opendb->Write(rocksdb::WriteOptions(), &batch);
+    if (status.ok()) {
```
szaszm, line 259: We should handle the case when there is an interruption (kill, shutdown, etc) between adding the keys to the vector, and actually deleting them, including an interruption between these lines: emptying `keys_to_delete_` and actually deleting. For the latter part, doing the delete before emptying the list, but while holding the mutex may do the trick, but may not be the best idea from a performance standpoint. What do you think?

Do we store `keys_to_delete_` on disk?

lordgamez (approved):

extensions/rocksdb-repos/DatabaseContentRepository.cpp
```diff
@@ -198,27 +212,63 @@ bool DatabaseContentRepository::exists(const minifi::ResourceClaim &streamId) {
...
+      logger_->log_debug("Deleting resource %s", content_path);
+      return true;
+    } else if (status.IsNotFound()) {
+      logger_->log_debug("Resource %s was not found", content_path);
+      return true;
+    } else {
+      logger_->log_debug("Attempted, but could not delete %s", content_path);
+      return false;
+    }
   }
```
fgerlits, line 234: Nitpicking, but I would make the inside of this `if` into a separate function.  This way, we have to make sure to have a `return` at the end of every code path; that way, the compiler would check it for us.

fgerlits (commented):
LGTM, just some minor comments

conf/minifi.properties
```diff
@@ -38,6 +38,9 @@ nifi.content.repository.class.name=DatabaseContentRepository
 # nifi.flowfile.repository.rocksdb.compaction.period=2 min
 # nifi.database.content.repository.rocksdb.compaction.period=2 min
 
+# setting this value to 0 enables synchronous deletion
```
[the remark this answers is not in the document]
adamdebreceni: based on the time parsing logic, it is not allowed, should we explicitly check for "0" or rather some other value? (e.g. "sync")

extensions/rocksdb-repos/DatabaseContentRepository.cpp
adamdebreceni: on a restart we first comb through the content repository and delete all content without a flowfile referencing them (in `clearOrphans`) that should prevent leaking content

extensions/rocksdb-repos/DatabaseContentRepository.cpp
adamdebreceni: done

CONFIGURE.md
```diff
@@ -224,6 +224,17 @@ Finally, as the last line is commented out, it will make the state manager use p
 
 When multiple repositories use the same directory (as with `minifidb://` scheme) they should either be all plaintext or all encrypted with the same key.
 
+### Configuring Repository Cleanup
+
+When a flow file content is no longer needed we can specify the deletion strategy.
+
+    # any value other than 0 enables async cleanup with the specified period
```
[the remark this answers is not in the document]
adamdebreceni: done

libminifi/test/rocksdb-tests/DBContentRepositoryTests.cpp
```diff
@@ -103,16 +105,35 @@ TEST_CASE("Delete Claim", "[TestDBCR2]") {
...
+
+  SECTION("Async") {
+    configuration->set(minifi::Configure::nifi_dbcontent_repository_purge_period, "100 ms");
+    REQUIRE(content_repo->initialize(configuration));
+    content_repo->start();
+
+    content_repo->remove(*claim);
+
+    // an immediate read will still be able to access the content
+    REQUIRE_FALSE(minifi::io::isError(content_repo->read(*claim)->read(readstr)));
```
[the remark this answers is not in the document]
adamdebreceni: it wasn't safe as the gc runs first before waiting, flipped it so that the gc does not run at 0, so it should be safe now

conf/minifi.properties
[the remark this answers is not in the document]
fgerlits: I'm fine with any of these options, as long as the documentation is in line with the reality.  Maybe we could even update the time parsing logic to allow "0"?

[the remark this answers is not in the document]
adamdebreceni: added explicit check for "0" while we sort out if we want to allow parsing "0" as time in all places (in probably a separate PR)

extensions/rocksdb-repos/DatabaseContentRepository.cpp
```diff
@@ -107,10 +115,15 @@ void DatabaseContentRepository::start() {
...
   if (compaction_period_.count() != 0) {
-    compaction_thread_ = std::make_unique<utils::StoppableThread>([this] () {
+    compaction_thread_ = std::make_unique<utils::StoppableThread>([this] {
       runCompaction();
     });
   }
+  if (purge_period_.count() != 0) {
+    gc_thread_ = std::make_unique<utils::StoppableThread>([this] {
+      runGc();
+    });
```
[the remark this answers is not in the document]
adamdebreceni: as I understand, compaction and deletion could run concurrently without blocking each other, they also operate on different time scales, merging them could be troublesome

fgerlits (approved):

martinzink (approved):

extensions/rocksdb-repos/DatabaseContentRepository.cpp
adamdebreceni: added

szaszm (approved):

PR #1507, opened 2023-02-09
MINIFICPP-1997 Fix transiently failing FileSystemRepositoryTests

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-1997

------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/test/unit/FileSystemRepositoryTests.cpp
+++ b/libminifi/test/unit/FileSystemRepositoryTests.cpp
@@ -28,6 +28,9 @@
 #include "../Catch.h"
 #include "utils/Literals.h"
 #include "core/repository/FileSystemRepository.h"
+#include "utils/IntegrationTestUtils.h"
+
+using namespace std::literals::chrono_literals;
 
 TEST_CASE("Test Physical memory usage", "[testphysicalmemoryusage]") {
   TestController controller;
@@ -47,7 +50,6 @@ TEST_CASE("Test Physical memory usage", "[testphysicalmemoryusage]") {
     stream->write(fragment.as_span<const std::byte>());
   }
 
-  const auto end_memory = utils::OsUtils::getCurrentProcessPhysicalMemoryUsage();
-
-  REQUIRE(gsl::narrow<size_t>(end_memory - start_memory) < 1_MB);
+  using org::apache::nifi::minifi::utils::verifyEventHappenedInPollTime;
+  REQUIRE(verifyEventHappenedInPollTime(5s, [&] { return gsl::narrow<size_t>(utils::OsUtils::getCurrentProcessPhysicalMemoryUsage() - start_memory) < 5_MB; }, 100ms));
 }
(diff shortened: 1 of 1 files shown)


martinzink (approved):

szaszm:
Could you explain what's happening here? Why do we have to wait for memory usage to decrease after a small increase? Why do we have to wait more and allow a larger difference? Did minifi become heavier over time, or is there something else at play?

lordgamez:
> Could you explain what's happening here? Why do we have to wait for memory usage to decrease after a small increase? Why do we have to wait more and allow a larger difference? Did minifi become heavier over time, or is there something else at play?

According to @adebreceni the original intent for this test was to check if the allocated memory in the filesystem repository is getting freed after writing the content which is 20MB in this test, so any memory check below 20MB should be sufficient. In very rare cases the memory used by MiNiFi can go above 1MB during the test (which is the original check limit) so it may fail. Usually the memory usage was about 2MB in failure cases, so 5MB should be more than enough to check for. There is also a time limit for this check to allow some time for the agent to release the memory, not just the exact time point after calling the `write()` function.

adamdebreceni (approved):

libminifi/test/unit/FileSystemRepositoryTests.cpp
```diff
@@ -47,7 +50,6 @@ TEST_CASE("Test Physical memory usage", "[testphysicalmemoryusage]") {
     stream->write(fragment.as_span<const std::byte>());
   }
 
-  const auto end_memory = utils::OsUtils::getCurrentProcessPhysicalMemoryUsage();
-
-  REQUIRE(gsl::narrow<size_t>(end_memory - start_memory) < 1_MB);
+  using org::apache::nifi::minifi::utils::verifyEventHappenedInPollTime;
+  REQUIRE(verifyEventHappenedInPollTime(5s, [&] { return gsl::narrow<size_t>(utils::OsUtils::getCurrentProcessPhysicalMemoryUsage() - start_memory) < 5_MB; }, 100ms));
```
fgerlits, line 54: It's not likely to happen, but I don't think we want to terminate if the current memory usage is lower than `start_memory`.
```suggestion
  CHECK(verifyEventHappenedInPollTime(5s, [&] {
    const auto end_memory = utils::OsUtils::getCurrentProcessPhysicalMemoryUsage();
    REQUIRE(end_memory > 0);
    return end_memory < start_memory + 5_MB;
  }, 100ms));
```

I would also put a `REQUIRE(start_memory > 0)` after line 42.

lordgamez: Updated in 45c5e56b0d3804b6fa7f990b4bb3c15666d0de48

fgerlits: `gsl::narrow<size_t>(end_memory - start_memory)` still terminates if `end_memory` is less than `start_memory`.  I think `end_memory < start_memory + int64_t{5_MB}` would be better.

lordgamez: Oh I see, you are right, fixed in 26e45b1525e76a4ce6d4d2844226868b0c29f068

fgerlits (approved):

szaszm (approved):

PR #1518, opened 2023-02-28
MINIFICPP-2057 Improve the performance of ListFile

fgerlits:
On a sample test data of 20,000 files on Windows, `onTrigger` now takes ~2.5 sec instead of ~7.5 sec in the case when all 20,000 files have been processed already.  The main changes are:
* except for the file name (with path) and modification time, only query anything else (eg. file size) when and if we need it;
* check first if the file is new, and only later if it matches the selection criteria;
* instead of collecting the file names in a vector first and then processing them in a loop, we process the files directly in the `list_dir` callback (without this, the `onTrigger` duration was ~3.5 sec).

NOTE: there are three changes in behavior:
1. When storing the list of files with the latest modification time, we used to store the path only, which doesn't uniquely identify the file.  I'm pretty sure this was a bug, now we store path + filename, and there is a new unit test to verify that this works.
2. If we are not able to find the modification time of the file (apparently, this can happen on Windows), we used to skip the file.  Now we list the file, with the modification time set to the epoch of the `system_clock` .
3. If the `Path Filter` is set, we used to list all files in the `Input Directory` root, plus all files in subdirectories the names of which match the filter.  Now we only list files in matching subdirectories, only including the root if the filter regex includes it (eg., `"good_subdir|ugly_subdir|\."`).  Also, `Path Filter` is ignored if `Recurse Subdirectories` is false -- in this case, all files in the `Input Directory` are listed, but none in the subdirectories.  The new behavior matches the current behavior of NiFi; I think the old behavior matched the behavior of an earlier version of NiFi.

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/standard-processors/processors/ListFile.h
+++ b/extensions/standard-processors/processors/ListFile.h
@@ -81,19 +81,15 @@ class ListFile : public core::Processor {
  private:
   struct ListedFile : public utils::ListedObject {
     [[nodiscard]] std::chrono::time_point<std::chrono::system_clock> getLastModified() const override {
-      return std::chrono::time_point_cast<std::chrono::milliseconds>(utils::file::FileUtils::to_sys(last_modified_time));
+      return last_modified_time;
     }
 
     [[nodiscard]] std::string getKey() const override {
       return full_file_path.string();
     }
 
-    std::filesystem::path filename;
-    std::filesystem::path absolute_path;
-    std::filesystem::file_time_type last_modified_time;
-    std::filesystem::path relative_path;
+    std::chrono::time_point<std::chrono::system_clock> last_modified_time;
     std::filesystem::path full_file_path;
-    uint64_t file_size = 0;
   };
 
   bool fileMatchesFilters(const ListedFile& listed_file);
--- a/extensions/standard-processors/processors/ListFile.h
+++ b/extensions/standard-processors/processors/ListFile.h
@@ -85,7 +85,7 @@ class ListFile : public core::Processor {
     }
 
     [[nodiscard]] std::string getKey() const override {
-      return absolute_path.string();
+      return full_file_path.string();
     }
 
     std::filesystem::path filename;
--- a/libminifi/include/utils/file/FileUtils.h
+++ b/libminifi/include/utils/file/FileUtils.h
@@ -126,13 +126,22 @@ inline std::optional<std::filesystem::file_time_type> last_write_time(const std:
   return std::nullopt;
 }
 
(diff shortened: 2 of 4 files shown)


extensions/standard-processors/processors/ListFile.h
```diff
@@ -81,19 +81,15 @@ class ListFile : public core::Processor {
...
     [[nodiscard]] std::string getKey() const override {
-      return absolute_path.string();
+      return full_file_path.string();
     }
 
-    std::filesystem::path filename;
-    std::filesystem::path absolute_path;
-    std::filesystem::file_time_type last_modified_time;
-    std::filesystem::path relative_path;
+    std::chrono::time_point<std::chrono::system_clock> last_modified_time;
```
szaszm, line 91: Does this change have any implications for reusing existing state from an older minifi version before the change? In other words, does this change the contents or the interpretation of the stored state?

fgerlits: `ListedFile::getLastModified()` used to convert the `file_time_type` to `time_point<system_clock>` before storing it in the state manager.  Now we do this conversion earlier, but the stored data doesn't change.

libminifi/include/utils/file/FileUtils.h
```diff
@@ -126,15 +126,24 @@ inline std::optional<std::filesystem::file_time_type> last_write_time(const std:
...
+inline std::optional<std::string> format_time(const time_t& time, const std::string& format) {
   std::array<char, 128U> result{};
-  if (std::strftime(result.data(), result.size(), format.c_str(), gmtime(&last_write_time_t)) != 0) {
+  if (std::strftime(result.data(), result.size(), format.c_str(), gmtime(&time)) != 0) {
     return std::string(result.data());
   }
   return std::nullopt;
 }
 
+inline std::optional<std::string> format_time(const std::chrono::system_clock::time_point& time, const std::string& format) {
```
martinzink, line 137: We have some time_point to str utilities in TimeUtil.h, that deal with these problem. (maybe we should change the input parameter from sys_seconds to std::chrono::system_clock::timepoint)

fgerlits: Makes sense, I've changed it in 9dc26a64f32870b99c9e9a562b37500b4638c943.

lordgamez (approved):

szaszm (approved):

martinzink (approved):

adamdebreceni (approved):

PR #1519, opened 2023-03-01
MINIFICPP-2059 Handle content repository remove failures

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2059

NOTE: This PR focuses on the FileSystemRepository as the original issue comes from files not being deleted in the content repository on a Windows filesystem when using FileSystemRepository. Testing the same retry mechanism is missing from this PR and a separate [Jira ticket](https://issues.apache.org/jira/browse/MINIFICPP-2062) is opened for it which includes creating a mock for the RocksDB content repository for easier testability.

------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/core/ContentRepository.h
+++ b/libminifi/include/core/ContentRepository.h
@@ -21,6 +21,7 @@
 #include <memory>
 #include <string>
 #include <utility>
+#include <list>
 
 #include "properties/Configure.h"
 #include "ResourceClaim.h"
@@ -53,10 +54,18 @@ class ContentRepository : public StreamManager<minifi::ResourceClaim>, public ut
 
   virtual void clearOrphans() = 0;
 
+  bool remove(const minifi::ResourceClaim &streamId) override {
+    return removeKey(streamId.getContentFullPath());
+  }
+
  protected:
+  void removeFromPurgeList();
+  virtual bool removeKey(const std::string& content_path) = 0;
+
   std::string directory_;
   std::mutex count_map_mutex_;
   std::map<std::string, uint32_t> count_map_;
+  std::list<std::string> purge_list_;
 };
 
 }  // namespace org::apache::nifi::minifi::core
--- a/libminifi/include/core/ContentRepository.h
+++ b/libminifi/include/core/ContentRepository.h
@@ -54,9 +54,7 @@ class ContentRepository : public StreamManager<minifi::ResourceClaim>, public ut
 
   virtual void clearOrphans() = 0;
 
-  bool remove(const minifi::ResourceClaim &streamId) override {
-    return removeKey(streamId.getContentFullPath());
-  }
+  bool remove(const minifi::ResourceClaim &streamId) final override;
 
  protected:
   void removeFromPurgeList();
@@ -64,6 +62,7 @@ class ContentRepository : public StreamManager<minifi::ResourceClaim>, public ut
 
   std::string directory_;
   std::mutex count_map_mutex_;
+  std::mutex purge_list_mutex_;
   std::map<std::string, uint32_t> count_map_;
   std::list<std::string> purge_list_;
 };
--- a/libminifi/include/core/ContentRepository.h
+++ b/libminifi/include/core/ContentRepository.h
@@ -54,7 +54,7 @@ class ContentRepository : public StreamManager<minifi::ResourceClaim>, public ut
 
   virtual void clearOrphans() = 0;
 
-  bool remove(const minifi::ResourceClaim &streamId) final override;
+  bool remove(const minifi::ResourceClaim &streamId) final;
 
  protected:
   void removeFromPurgeList();
--- a/libminifi/src/core/ContentRepository.cpp
+++ b/libminifi/src/core/ContentRepository.cpp
@@ -61,6 +61,7 @@ void ContentRepository::incrementStreamCount(const minifi::ResourceClaim &stream
 }
 
 void ContentRepository::removeFromPurgeList() {
+  std::lock_guard<std::mutex> lock(purge_list_mutex_);
   for (auto it = purge_list_.begin(); it != purge_list_.end();) {
     if (removeKey(*it)) {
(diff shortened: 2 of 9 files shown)


libminifi/src/core/ContentRepository.cpp
```diff
@@ -60,16 +60,29 @@ void ContentRepository::incrementStreamCount(const minifi::ResourceClaim &stream
...
 
+void ContentRepository::removeFromPurgeList() {
+  for (auto it = purge_list_.begin(); it != purge_list_.end();) {
+    if (removeKey(*it)) {
+      purge_list_.erase(it++);
+    } else {
+      ++it;
+    }
+  }
+}
```
szaszm, line 71: This is a nice mechanism, but I have one concern: What if we just keep getting errors due to some persistent issue?
Can we limit the memory consumed by this list? E.g. stop operation if it grows above 1M or so entries.
Can we keep retrying deletion of the purge list after an agent restart or flow update?

lordgamez: It depends if we expect any persistent issues that have to be dealt with. My thinking was that we are creating this list for any temporary issues that should be resolved in a few tries, and it should only affect a small number of entries. In case of a restart I think the `clearOrphans` mechanism should remove the entries we did not manage to delete before, maybe if that fails those can be added to the purge list for handling them later. What do you think? Should we discuss this with the team?

szaszm: Let's stay with no limit for now, but implement retry in `clearOrphans` as well, so an agent restart doesn't cause content leaks.

libminifi/include/core/ContentRepository.h
```diff
@@ -53,10 +54,18 @@ class ContentRepository : public StreamManager<minifi::ResourceClaim>, public ut
...
+  bool remove(const minifi::ResourceClaim &streamId) override {
+    return removeKey(streamId.getContentFullPath());
+  }
+
  protected:
+  void removeFromPurgeList();
+  virtual bool removeKey(const std::string& content_path) = 0;
+
   std::string directory_;
   std::mutex count_map_mutex_;
```
adamdebreceni, line 66: rename this mutex as it no longer protects just the `count_map_` but the `purge_list_` as well

libminifi/include/core/ContentRepository.h
```diff
@@ -53,10 +54,18 @@ class ContentRepository : public StreamManager<minifi::ResourceClaim>, public ut
 
   virtual void clearOrphans() = 0;
 
+  bool remove(const minifi::ResourceClaim &streamId) override {
```
adamdebreceni, line 57: can this be `final`?

libminifi/src/core/ContentRepository.cpp
```diff
@@ -60,16 +60,29 @@ void ContentRepository::incrementStreamCount(const minifi::ResourceClaim &stream
...
   auto count = count_map_.find(str);
   if (count != count_map_.end() && count->second > 1) {
     count_map_[str] = count->second - 1;
     return StreamState::Alive;
   } else {
     count_map_.erase(str);
-    remove(streamId);
+    if (!remove(streamId)) {
+      purge_list_.push_back(streamId.getContentFullPath());
+    }
```
adamdebreceni, line 85: why isn't this in `remove`?

libminifi/src/core/ContentRepository.cpp
```diff
@@ -60,16 +60,29 @@ void ContentRepository::incrementStreamCount(const minifi::ResourceClaim &stream
...
+      purge_list_.erase(it++);
+    } else {
+      ++it;
+    }
+  }
+}
+
 ContentRepository::StreamState ContentRepository::decrementStreamCount(const minifi::ResourceClaim &streamId) {
   std::lock_guard<std::mutex> lock(count_map_mutex_);
+  removeFromPurgeList();
```
adamdebreceni, line 75: the cleanup logic is quite similar to `FlowFileRepository`'s background `flush` thread (and the planned `DatabaseContentRepository`'s background batch cleanup), I think we should move this cleanup task out of `decrementStreamCount`'s responsibilities

lordgamez: You are right it shouldn't be the responsibility of `decrementStreamCount`, I moved it to be part of the `remove` in bde565c405e3e9419020445a48d41da561b58729 I'm not sure if there should be a separate thread for this, what do you think?

libminifi/src/core/ContentRepository.cpp
lordgamez: Updated in bde565c405e3e9419020445a48d41da561b58729

libminifi/include/core/ContentRepository.h
lordgamez: Updated in bde565c405e3e9419020445a48d41da561b58729

libminifi/include/core/ContentRepository.h
lordgamez: I prefer having a separate mutex for the purge_list and separating the `count_map_` and  `purge_list_` processing blocks, updated in bde565c405e3e9419020445a48d41da561b58729

libminifi/src/core/ContentRepository.cpp
lordgamez: Added purge list update in `clearOprhans` in bde565c405e3e9419020445a48d41da561b58729

libminifi/src/core/ContentRepository.cpp
adamdebreceni: I'm not sure either, I think it is good as it is now
(later, if needed, we could move it to its own thread later or even integrate it into the planned async purge, we'll see)

adamdebreceni (approved):

szaszm:
Rebasing this on main would probably fix the centos CI job.

martinzink (approved):

PR #1533, opened 2023-03-14
MINIFICPP-2078 Remove cleanup from __del__ functions

lordgamez:
Cleaning up docker resources is done manually in the behave `environment.py` file manually by calling the `cleanup` functions. The was cleanup is also called redundantly from the resources' `__del__` function which is called by the GC at unpredictable times. This caused that the docker remove function could be called while a new container with the same name is instantiated in a next test case which killed that container.

https://issues.apache.org/jira/browse/MINIFICPP-2078

---------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


fgerlits:
Should we remove the `__del__` function from `DockerTestDirectoryBindings`, too, or is that one harmless?

lordgamez:
> Should we remove the `__del__` function from `DockerTestDirectoryBindings`, too, or is that one harmless?

I checked and that can be kept as it only handles temp directories and those can be kept until the end of the test runs, so they should be harmless.

fgerlits (approved):

adamdebreceni (approved):

PR #1538, opened 2023-03-21
MINIFICPP-2073 Separate docker build from docker tests in CI

martinzink:

### For all changes:
     in the commit message?

### For code changes:

### Note:


.github/workflows/ci.yml
```diff
@@ -223,13 +227,132 @@ jobs:
           if [ -d ~/.ccache ]; then mv ~/.ccache .; fi
           mkdir build
           cd build
-          cmake -DUSE_SHARED_LIBS= -DSTRICT_GSL_CHECKS=AUDIT -DCI_BUILD=ON -DDISABLE_JEMALLOC=ON -DENABLE_AWS=ON -DENABLE_LIBRDKAFKA=ON -DENABLE_MQTT=ON -DENABLE_AZURE=ON -DENABLE_SQL=ON \
-              -DENABLE_SPLUNK=ON -DENABLE_GCP=ON -DENABLE_OPC=ON -DENABLE_PYTHON_SCRIPTING=ON -DENABLE_LUA_SCRIPTING=ON -DENABLE_KUBERNETES=ON -DENABLE_TEST_PROCESSORS=ON -DENABLE_PROMETHEUS=ON \
-              -DDOCKER_BUILD_ONLY=ON -DDOCKER_CCACHE_DUMP_LOCATION=$HOME/.ccache ..
+          cmake ${DOCKER_CMAKE_FLAGS} ..
           make docker
+      - name: Save docker image
+        run: cd build && docker save -o minifi_docker.tar apacheminificpp:0.13.0
```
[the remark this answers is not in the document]
martinzink: It wasnt super simple to go with the suggested route, because this versioning-scheme  is used/anticipated in multiple places.
Found a simpler method though: [we can simply extract the version from CMake](https://github.com/apache/nifi-minifi-cpp/pull/1538/commits/6a7cd955d9c0e37815ee27ab2dfcf149a0b932cb).

.github/workflows/ci.yml
szaszm, line 1: I think it would be better to run the 4 parallel jobs on the same VM, to save cloud resources. Maybe we could play with buffering the output to avoid the interleaving of tests, or tag each line with a thread number.

martinzink: Thats a good idea, however that would require a substantial refactor. Currently the tests are referencing the images by hardcoded names(e.g. minifi-cpp-flow) so running multiple tests at the same time are currently unfeasable. I can create a jira ticket to investigate how much effort that refactoring would need, but it is certainly out of the scope of this PR.

This PR saves cloud resources when we inevitably have to rerun something. Instead of reruning the whole building/running process we can rerun only the affected quadrant of tests.

lordgamez: There is already a ticket for running tests in parallel, but as Martin mentioned it requires larger changes on the test framework: https://issues.apache.org/jira/browse/MINIFICPP-1641

lordgamez: I think it's good to separate the test jobs and the build jobs and it would be cool to have that for the other actions too. Although I'm not convinced that we need to rerun these jobs so often that we need to separate them to 4 quadrants. I think it's more common that some of the tests in ctest runs fail than the ones running in docker. I would rather encourage identifying and fixing the flaky tests cases instead of making it easier to rerun them (which would help people disregard the issues).

lordgamez: @martinzink all right, I'm okay with that
@szaszm unfortunately that will not work due to the previously mentioned container naming conflicts

szaszm (approved):

lordgamez (approved):

.github/workflows/ci.yml
```diff
@@ -302,13 +306,132 @@ jobs:
...
+      - id: checkout
+        uses: actions/checkout@v3
+      - id: run_cmake
+        name: Run CMake
+        run: |
+          mkdir build
+          cd build
+          cmake ${DOCKER_CMAKE_FLAGS} ..
+      - name: Download artifact
+        uses: actions/download-artifact@v3
```
adamdebreceni, line 333: is it guaranteed that this steps downloads the previous step's artifact? could another (PR) run override the artifact?

martinzink: yes, these simple download/upload artifacts actions (provided by github) are limited to the current workflow

PR #1554, opened 2023-04-12
MINIFICPP-2058 Fix AWS extension link error on ARM64

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2058

----------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/aws/CMakeLists.txt
+++ b/extensions/aws/CMakeLists.txt
@@ -33,7 +33,9 @@ add_library(minifi-aws SHARED ${SOURCES})
 
 target_link_libraries(minifi-aws PUBLIC ${LIBMINIFI} Threads::Threads)
 
+target_wholearchive_library_private(minifi-aws AWS::aws-checksums)
 target_wholearchive_library_private(minifi-aws AWS::aws-cpp-sdk-s3)
+
 get_target_property(AWS_SDK_INCLUDE_DIRS AWS::aws-cpp-sdk-s3 INTERFACE_INCLUDE_DIRECTORIES)
 target_include_directories(minifi-aws INTERFACE ${AWS_SDK_INCLUDE_DIRS})
 if(WIN32)
--- a/extensions/aws/CMakeLists.txt
+++ b/extensions/aws/CMakeLists.txt
@@ -34,7 +34,7 @@ add_library(minifi-aws SHARED ${SOURCES})
 target_link_libraries(minifi-aws PUBLIC ${LIBMINIFI} Threads::Threads)
 
 target_wholearchive_library_private(minifi-aws AWS::aws-cpp-sdk-s3)
-if(NOT CMAKE_SYSTEM_PROCESSOR MATCHES "(x86)|(X86)")
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "(arm64)|(ARM64)|(aarch64)|(armv8)")
     target_wholearchive_library_private(minifi-aws AWS::aws-checksums)
 endif()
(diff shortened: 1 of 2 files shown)


extensions/aws/CMakeLists.txt
szaszm, line 1: Is there a way to keep that symbol without keeping all of the unused symbols? Or is this the officially documented way?

lordgamez: There isn't much documentation on this, most examples I found in other repositories links all the resulting libraries to the client library. It also seemed that not only this single symbol was missing in the linking process, because when this was resolved there were other missing aws symbols from other libraries. Unfortunately this was the only way I could make it work on both ARM64 linux and on Windows as there were other alternative ways that did not work on Windows.

lordgamez: It seems that I was wrong as the linkage problem didn't originate from the aws crt library as I originally suspected. All the missing symbols can be resolved if we link the aws-checksums as a whole archive. I tested it on arm and it solves the issue, updated in 82e8074988ef5caf6b06e68140c3faae8ce0fd83

martinzink: Now with this fix, it seems we can remove the workarounds I've included in MINIFICPP-2048 that fixed the compile issues in the tests. https://github.com/apache/nifi-minifi-cpp/commit/06d5467c6ca2d9a20ca9a969fd768e03025c362d#diff-4a191f6c5036df0f852be61f427c6cc029220f0dd8dbc9600fcfa59e760b2edfR21 https://github.com/apache/nifi-minifi-cpp/commit/06d5467c6ca2d9a20ca9a969fd768e03025c362d#diff-4a191f6c5036df0f852be61f427c6cc029220f0dd8dbc9600fcfa59e760b2edfR38

szaszm: Can we restrict the "wholearchive" linkage to ARM64?

martinzink: I think something like 
```
if (CMAKE_SYSTEM_PROCESSOR MATCHES "(aarch64)|(arm64)")
    target_wholearchive_library_private(minifi-aws AWS::aws-checksums)
endif()
```
should work

lordgamez: Restricted linkage and removed test workaround in 3ad1d51a4b678eed4550251c33f5a0a62d19d524

martinzink (approved):

szaszm (approved):

extensions/aws/CMakeLists.txt
```diff
@@ -34,6 +34,9 @@ add_library(minifi-aws SHARED ${SOURCES})
 target_link_libraries(minifi-aws PUBLIC ${LIBMINIFI} Threads::Threads)
 
 target_wholearchive_library_private(minifi-aws AWS::aws-cpp-sdk-s3)
+if(NOT CMAKE_SYSTEM_PROCESSOR MATCHES "(x86)|(X86)")
```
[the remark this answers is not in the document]
lordgamez: I actually made the change before I read Martin's comment, I thought it would be better to link the library on any non-x86 architecture as it may be needed there as well. But as I checked https://stackoverflow.com/questions/70475665/what-are-the-possible-values-of-cmake-system-processor there are more possibilities than I thought, like you said AMD64 also should have been included on windows. So it may be better to only concentrate on arm64 possibilities and we'll see in the future if we want to support more. I changed it to match only on arm64, ARM64, aarch64 and armv8 in 4037275310298711c7f949011b3e09e32a10ccdb

adamdebreceni (approved):

fgerlits (approved):

PR #1560, opened 2023-04-18
MINIFICPP-2101 Compilation fix in PutSFTPTests (libc++)

martinzink:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/standard-processors/tests/unit/ListFileTests.cpp
+++ b/extensions/standard-processors/tests/unit/ListFileTests.cpp
@@ -48,7 +48,6 @@ using org::apache::nifi::minifi::utils::verifyLogLinePresenceInPollTime;
 
 class ListFileTestFixture {
  public:
-  static const std::string FORMAT_STRING;
   ListFileTestFixture();
 
  protected:
@@ -62,8 +61,6 @@ class ListFileTestFixture {
   std::filesystem::path first_sub_file_abs_path_;
   std::filesystem::path second_sub_file_abs_path_;
 };
-
-const std::string ListFileTestFixture::FORMAT_STRING = "%Y-%m-%dT%H:%M:%SZ";
 
 ListFileTestFixture::ListFileTestFixture()
   : plan_(test_controller_.createPlan()),
@@ -107,8 +104,12 @@ TEST_CASE_METHOD(ListFileTestFixture, "Input Directory is empty", "[testListFile
   REQUIRE_THROWS_AS(test_controller_.runSession(plan_, true), minifi::Exception);
 }
 
+std::string get_last_modified_time_formatted_string(const std::filesystem::path& path) {
+  return date::format("%Y-%m-%dT%H:%M:%SZ", std::chrono::time_point_cast<std::chrono::seconds>(utils::file::to_sys(*utils::file::last_write_time(path))));
+}
+
 TEST_CASE_METHOD(ListFileTestFixture, "Test listing files only once with default parameters", "[testListFile]") {
-  test_controller_.runSession(plan_);
+  TestController::runSession(plan_);
   REQUIRE(verifyLogLinePresenceInPollTime(3s, "key:filename value:standard_file.log"));
   REQUIRE(verifyLogLinePresenceInPollTime(3s, "key:filename value:empty_file.txt"));
   REQUIRE(verifyLogLinePresenceInPollTime(3s, "key:filename value:sub_file_one.txt"));
@@ -135,19 +136,19 @@ TEST_CASE_METHOD(ListFileTestFixture, "Test listing files only once with default
   if (auto owner = utils::file::FileUtils::get_file_owner(standard_file_abs_path_)) {
     REQUIRE(LogTestController::getInstance().countOccurrences("key:file.owner value:" + *owner) == 4);
   }
-  REQUIRE(verifyLogLinePresenceInPollTime(3s, "key:file.lastModifiedTime value:" + *utils::file::FileUtils::get_last_modified_time_formatted_string(empty_file_abs_path_, FORMAT_STRING)));
-  REQUIRE(verifyLogLinePresenceInPollTime(3s, "key:file.lastModifiedTime value:" + *utils::file::FileUtils::get_last_modified_time_formatted_string(standard_file_abs_path_, FORMAT_STRING)));
-  REQUIRE(verifyLogLinePresenceInPollTime(3s, "key:file.lastModifiedTime value:" + *utils::file::FileUtils::get_last_modified_time_formatted_string(first_sub_file_abs_path_, FORMAT_STRING)));
-  REQUIRE(verifyLogLinePresenceInPollTime(3s, "key:file.lastModifiedTime value:" + *utils::file::FileUtils::get_last_modified_time_formatted_string(second_sub_file_abs_path_, FORMAT_STRING)));
+  REQUIRE(verifyLogLinePresenceInPollTime(3s, "key:file.lastModifiedTime value:" + get_last_modified_time_formatted_string(empty_file_abs_path_)));
+  REQUIRE(verifyLogLinePresenceInPollTime(3s, "key:file.lastModifiedTime value:" + get_last_modified_time_formatted_string(standard_file_abs_path_)));
+  REQUIRE(verifyLogLinePresenceInPollTime(3s, "key:file.lastModifiedTime value:" + get_last_modified_time_formatted_string(first_sub_file_abs_path_)));
+  REQUIRE(verifyLogLinePresenceInPollTime(3s, "key:file.lastModifiedTime value:" + get_last_modified_time_formatted_string(second_sub_file_abs_path_)));
   plan_->reset();
   LogTestController::getInstance().clear();
-  test_controller_.runSession(plan_, true);
+  TestController::runSession(plan_, true);
   REQUIRE_FALSE(LogTestController::getInstance().contains("key:file.size", 0s, 0ms));
 }
 
 TEST_CASE_METHOD(ListFileTestFixture, "Test turning off recursive file listing", "[testListFile]") {
   plan_->setProperty(list_file_processor_, "Recurse Subdirectories", "false");
-  test_controller_.runSession(plan_);
+  TestController::runSession(plan_);
   REQUIRE(verifyLogLinePresenceInPollTime(3s, "key:filename value:standard_file.log"));
   REQUIRE(verifyLogLinePresenceInPollTime(3s, "key:filename value:empty_file.txt"));
   REQUIRE_FALSE(LogTestController::getInstance().contains("key:filename value:sub_file_one.txt", 0s, 0ms));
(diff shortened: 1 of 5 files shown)


libminifi/src/utils/file/FileUtils.cpp
```diff
@@ -74,13 +74,23 @@ time_t to_time_t(std::filesystem::file_time_type file_time) {
...
+#endif
+}
+
+std::filesystem::file_time_type from_sys(std::chrono::system_clock::time_point sys_time) {
+#if defined(WIN32)
+  return std::chrono::clock_cast<std::chrono::file_clock>(sys_time);
+#elif defined(_LIBCPP_VERSION) && (_LIBCPP_VERSION < 14000)
+  return std::chrono::file_clock::from_time_t(std::chrono::system_clock::to_time_t(sys_time));
+#else
+  return std::chrono::file_clock::from_sys(sys_time);
```
fgerlits, line 93: For which compiler is the duration cast needed in `to_sys`?  Does it need to be added for that compiler here in `from_sys`, as well?

libminifi/src/utils/file/FileUtils.cpp
fgerlits, line 1: some unit tests for these conversion functions would be useful

libminifi/src/utils/file/FileUtils.cpp
martinzink: On libc++ (if the _LIBCPP_HAS_NO_INT128 is not defined) file_clock uses a 128 bit resolution while the system_clock only uses 64 bit, due to this we only need the cast in from_sys.

fgerlits: I don't get it.  If `std::chrono::file_clock::to_sys(file_time)` returns 128 bits and needs to be duration-cast to 64 bits, then doesn't `std::chrono::file_clock::from_sys(sys_time)` return 64 bits and needs to be duration-cast to 128 bits?

fgerlits: Ah, got it.  The duration conversion happens automatically in the `return` statement, but we need the `time_point_cast` when converting from a 128-bit underlying type to a 64-bit one (but not the other way round).

That's a bit nasty -- I hope `clock_cast` will fix this problem when it's available on all compilers.

martinzink: You only need to cast it where it can lose precision otherwise the cast is implicit.
So from hours to minutes you dont need duration_cast while you would need in the other direction.

Of course we could add the explicit cast to other function aswell, if we expect that on some platfrom system_clock will be more precise than file_clock (or simply to improve readability) What do you think?

fgerlits: Yes, I think adding the `time_point_cast` in the unnecessary direction would help to make the code less confusing to humans.  Hopefully, the compiler will optimize it out.  Even more hopefullier, we can soon remove all this mess and simply use `clock_cast`.

martinzink:
The windows test failure seems related, I'll investigate

libminifi/src/utils/file/FileUtils.cpp
martinzink: I've added some unit tests in https://github.com/apache/nifi-minifi-cpp/pull/1560/commits/d8e7f065500dc282786918880057f7a23ca0b942
I've only added back and forth tests since we can't parse file_clock on every platform and the file_clock epoch is unspecified, so its seems impossible to specify the same time on system_clock and file_clock independently.

libminifi/src/utils/file/FileUtils.cpp
martinzink: It turns out we can't rely on clock_cast on windows either, because (for some reason it needs timezone information which should be available after msi install but not neccesarly there during tests of course we could load it like CronTests do, but I think we can skip the hassle since the other platforms dont have clock_cast)

Anyways, I've refactored these in https://github.com/apache/nifi-minifi-cpp/pull/1560/commits/d8e7f065500dc282786918880057f7a23ca0b942

I've also removed the to_time_t functions from FileUtils since it was only used in a single test (ListFileTests) https://github.com/apache/nifi-minifi-cpp/pull/1560/commits/256d695e88751dbf5519546f53eebb72e4f3776f

libminifi/test/unit/FileUtilsTests.cpp
```diff
@@ -477,3 +477,27 @@ TEST_CASE("FileUtils::get_relative_path", "[TestGetRelativePath]") {
   REQUIRE(*FileUtils::get_relative_path(path, base_path / "") == std::filesystem::path("subdir") / "file.log");
   REQUIRE(*FileUtils::get_relative_path(base_path, base_path) == ".");
 }
+
+TEST_CASE("file_clock to system_clock conversion tests") {
+  using namespace std::chrono;
+
+  static_assert(system_clock::duration::period::num == file_clock::duration::period::num);
+  constexpr auto lowest_den = system_clock::period::den < file_clock::period::den ? system_clock::period::den : file_clock::period::den;
+  using LeastPreciseDurationType = duration<std::common_type_t<system_clock::duration::rep, file_clock::duration::rep>, std::ratio<system_clock::duration::period::num, lowest_den>>;
```
fgerlits, line 486: minor, but it's slightly confusing that you sometimes write `x_clock::period` and sometimes `x_clock::duration::period`; I would use the shorter form everywhere

libminifi/test/unit/FileUtilsTests.cpp
```diff
@@ -477,3 +477,27 @@ TEST_CASE("FileUtils::get_relative_path", "[TestGetRelativePath]") {
   REQUIRE(*FileUtils::get_relative_path(path, base_path / "") == std::filesystem::path("subdir") / "file.log");
   REQUIRE(*FileUtils::get_relative_path(base_path, base_path) == ".");
 }
+
+TEST_CASE("file_clock to system_clock conversion tests") {
+  using namespace std::chrono;
+
+  static_assert(system_clock::duration::period::num == file_clock::duration::period::num);
+  constexpr auto lowest_den = system_clock::period::den < file_clock::period::den ? system_clock::period::den : file_clock::period::den;
```
fgerlits, line 485: `std::min` doesn't work here?

martinzink: sure thing, good idea, I've changed it in https://github.com/apache/nifi-minifi-cpp/pull/1560/commits/41f3dd4ef1416a1bf5ab38d7052e08678adc2d88

extensions/standard-processors/tests/unit/ListFileTests.cpp
```diff
@@ -107,8 +104,12 @@ TEST_CASE_METHOD(ListFileTestFixture, "Input Directory is empty", "[testListFile
   REQUIRE_THROWS_AS(test_controller_.runSession(plan_, true), minifi::Exception);
 }
 
+std::string get_last_modified_time_formatted_string(const std::filesystem::path& path) {
+  return date::format("%Y-%m-%dT%H:%M:%SZ", std::chrono::time_point_cast<std::chrono::seconds>(utils::file::to_sys(*utils::file::last_write_time(path))));
```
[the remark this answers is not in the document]
martinzink: you are right, I've changed it in https://github.com/apache/nifi-minifi-cpp/pull/1560/commits/41f3dd4ef1416a1bf5ab38d7052e08678adc2d88

fgerlits (approved):

lordgamez (approved):

adamdebreceni (approved):

libminifi/src/utils/file/FileUtils.cpp
```diff
@@ -66,21 +66,21 @@ bool contains(const std::filesystem::path& file_path, std::string_view text_to_s
   return std::search(view.begin(), view.end(), searcher) != view.end();
 }
 
-time_t to_time_t(std::filesystem::file_time_type file_time) {
-#if defined(_LIBCPP_VERSION) && _LIBCPP_VERSION < 14000
-  return std::chrono::file_clock::to_time_t(file_time);
+std::chrono::system_clock::time_point to_sys(std::filesystem::file_time_type file_time) {
+  using namespace std::chrono;  // NOLINT(build/namespaces)
+#if defined(WIN32) || (defined(_LIBCPP_VERSION) && (_LIBCPP_VERSION < 14000))
+  return system_clock::now() + duration_cast<system_clock::duration>(file_time - file_clock::now());
```
fgerlits, line 72: I think this causes occasional unit test failures because we read `now()` twice, and sometimes we get two different ticks back.

As an alternative, this worked for me on Windows:
```c++
  // to_sys
  return utc_clock::to_sys(file_clock::to_utc(file_time));

  // from_sys
  return file_clock::from_utc(utc_clock::from_sys(sys_time));
```

I'm not sure if it works on libc++ version < 14, too, but I'm also not sure if we really need to support libc++ version < 14.

fgerlits (changes requested):
"request changes" instead of "comment"

libminifi/src/utils/file/FileUtils.cpp
martinzink: It seems this method produces the same error which I encountered when I was using clock_cast (clock cast probably does this under the hood) (it loads timezonedb which based on the comments fails on older windowses https://github.com/microsoft/STL/issues/1911)
Also the macOS CI does seem to use libc++<14 (altough I might be able to update that)

martinzink: More info on the windows issue: https://github.com/microsoft/STL/issues/2446

martinzink: To support older windows it seems we would have to use the double now method or resort to hardcoded magic numbers like the file epoch is 1601.01.01. Whats your opinion?

fgerlits: I see, this is unfortunate.

I think hard-coding the epoch differences (`std::filesystem::__std_fs_file_time_epoch_adjustment` in the case of Windows, and `0` in the case of `libc++ < 14`) is the least bad option.

libminifi/test/unit/FileUtilsTests.cpp
```diff
@@ -516,3 +516,38 @@ TEST_CASE("FileUtils::path_size", "[TestPathSize]") {
...
+  }
+
+  {
+    system_clock::time_point system_now = system_clock::now();
+    file_clock::time_point file_now = file_clock ::now();
+
+    file_clock::time_point converted_system_now = FileUtils::from_sys(system_now);
+    system_clock::time_point converted_file_now = FileUtils::to_sys(file_now);
+
+    CHECK(system_now-converted_file_now < 10ms);
```
fgerlits, line 550: This doesn't really test what we want to test, because `system_now - converted_file_now` is either zero or negative.  The reverse should be OK:
```c++
    CHECK(converted_file_now - system_now < 10ms);
```
or, better:
```c++
    CHECK(0ms <= converted_file_now - system_now);
    CHECK(converted_file_now - system_now < 10ms);
```

martinzink: You are right, thanks. I've refactored it a bit so its more clear which one is earlier in  https://github.com/apache/nifi-minifi-cpp/pull/1560/commits/8139796ee615269c86d693942e6fc3b87d16a3bc

fgerlits (approved):

PR #1561, opened 2023-04-18
MINIFICPP-2103 JNI extension compile fix (libc++)

martinzink:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/jni/jvm/JniReferenceObjects.h
+++ b/extensions/jni/jvm/JniReferenceObjects.h
@@ -104,7 +104,7 @@ class JniByteInputStream {
     }
 
     // seek to offset
-    auto remaining = gsl::narrow<uint64_t>(size);
+    auto remaining = gsl::narrow<size_t>(size);
     int writtenOffset = 0;
     int read = 0;
     do {
@@ -128,8 +128,8 @@ class JniByteInputStream {
     return read;
   }
 
-  int64_t read(uint8_t &arr) {
-    return stream_->read(arr);
+  int64_t read(uint8_t &arr) const {
+    return gsl::narrow<int64_t>(stream_->read(arr));
   }
 
   std::shared_ptr<minifi::io::InputStream> stream_;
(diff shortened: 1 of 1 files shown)


extensions/jni/jvm/JniReferenceObjects.h
```diff
@@ -128,8 +128,8 @@ class JniByteInputStream {
     return read;
   }
 
-  int64_t read(uint8_t &arr) {
-    return stream_->read(arr);
+  int64_t read(uint8_t &arr) const {
+    return gsl::narrow<int64_t>(stream_->read(arr));
```
fgerlits, line 132: If `stream_` is a `minifi::io::InputStream`, then `read()` can return `static_cast<size_t>(-1)`, which will cause this to terminate (or return an incorrect value if `size_t` is smaller than 64 bits).  We need to check if the return value `io::isError()`, and return `-1` explicitly in that case.

extensions/jni/jvm/JniReferenceObjects.h
```diff
@@ -128,8 +128,8 @@ class JniByteInputStream {
     return read;
   }
 
-  int64_t read(uint8_t &arr) {
-    return stream_->read(arr);
+  int64_t read(uint8_t &arr) const {
```
fgerlits, line 131: also, nitpicking and old code, but it's weird to call a reference-to-byte parameter "arr"

extensions/jni/jvm/JniReferenceObjects.h
martinzink: Nice catch, I've added the error check in https://github.com/apache/nifi-minifi-cpp/pull/1561/commits/dd896082c79acc2b122a3e74ab36c527f1fee446

fgerlits (approved):

lordgamez (approved):

adamdebreceni (approved):

PR #1568, opened 2023-05-03
MINIFICPP-2112 Fix flow update and restart with minifi controller

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2112

---------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/src/FlowController.cpp
```diff
@@ -193,7 +192,6 @@ int16_t FlowController::stop() {
     this->content_repo_->stop();
     // stop the ControllerServices
     disableAllControllerServices();
-    initialized_ = false;
```
lordgamez, line 196: Initialized state should only be false when we do not have a valid flow loaded.

libminifi/src/core/ProcessGroup.cpp
```diff
@@ -133,6 +133,7 @@ void ProcessGroup::startProcessingProcessors(TimerDrivenSchedulingAgent& timeSch
   for (const auto processor : failed_processors_) {
     try {
       logger_->log_debug("Starting %s", processor->getName());
+      processor->setScheduledState(core::ScheduledState::RUNNING);
```
lordgamez, line 136: Scheduling state needs to be reset to RUNNING if FlowController is stopped through controller or C2 otherwise scheduler will not be able to restart the processor.

cmake/DockerConfig.cmake
```diff
@@ -47,8 +47,9 @@ add_custom_target(
              -DENABLE_ELASTICSEARCH=OFF
              -DENABLE_LUA_SCRIPTING=OFF
              -DENABLE_PYTHON_SCRIPTING=OFF
-	     -DENABLE_OPC=OFF
-             -DENABLE_ENCRYPT_CONFIG=OFF \"
+             -DENABLE_OPC=OFF
+             -DENABLE_ENCRYPT_CONFIG=OFF
+             -DDISABLE_CONTROLLER=ON \"
```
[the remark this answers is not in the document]
lordgamez: That makes sense, added in 765bea7fbb282048bc9c3a223f667b37620e6039

controller/tests/ControllerTests.cpp
```diff
@@ -314,14 +314,14 @@ TEST_CASE_METHOD(ControllerTestFixture, "Test listComponents", "[controllerTests
   }
 
   using org::apache::nifi::minifi::utils::verifyEventHappenedInPollTime;
-  REQUIRE(verifyEventHappenedInPollTime(500ms, [&] { return controller_->isRunning(); }, 20ms));
+  REQUIRE(verifyEventHappenedInPollTime(5s, [&] { return controller_->isRunning(); }, 20ms));
```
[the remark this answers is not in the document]
lordgamez: Sometimes these checks were flaky in the CI probably due to the introduced communication of the command handlings threads through the `ConcurrentQueue`. 10 times is probably an overkill as it should probably be under 2 seconds in the CI as well, but I wanted to make sure to remove the flakyness and only that the functionality.

libminifi/include/FlowController.h
```diff
@@ -149,6 +149,22 @@ class FlowController : public core::controller::ForwardingControllerServiceProvi
...
+      --update_block_count_;
+    }
+    bool isUpdating() const {
+      return update_block_count_ > 0;
+    }
+
+   private:
+    std::atomic<uint32_t> update_block_count_;
+  };
+
```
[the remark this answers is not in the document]
lordgamez: Good point, added a scoped UpdateLock in ffefd48806fffcc07dbbff20d1159e206e19824f

docker/test/integration/features/CMakeLists.txt
```diff
@@ -87,6 +87,10 @@ if (ENABLE_PROMETHEUS)
     set(ENABLED_BEHAVE_TESTS "${ENABLED_BEHAVE_TESTS};${CMAKE_CURRENT_SOURCE_DIR}/prometheus.feature")
 endif()
 
+if (NOT DISABLE_CURL AND NOT DISABLE_CONTROLLER)
+    set(ENABLED_BEHAVE_TESTS "${ENABLED_BEHAVE_TESTS};${CMAKE_CURRENT_SOURCE_DIR}/minifi_controller.feature")
```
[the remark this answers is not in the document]
lordgamez: Updated in ffefd48806fffcc07dbbff20d1159e206e19824f

libminifi/src/FlowController.cpp
```diff
@@ -127,35 +127,34 @@ bool FlowController::applyConfiguration(const std::string &source, const std::st
...
-    if (!started) {
-      logger_->log_error("Failed to start new flow, restarting previous flow");
-      root_wrapper_.restoreBackup();
-      load(true);
-      start();
-    } else {
-      root_wrapper_.clearBackup();
+    auto update_lock = updating_.getUpdateLock();
+    {
+      std::lock_guard<std::recursive_mutex> flow_lock(mutex_);
```
szaszm, line 134: The inner scope is not necessary here, the two locks could just follow each other.

libminifi/include/FlowController.h
```diff
@@ -149,6 +149,34 @@ class FlowController : public core::controller::ForwardingControllerServiceProvi
...
+        update_state_.beginUpdate();
+      }
+
+      ~UpdateLock() {
+        update_state_.endUpdate();
+      }
+     private:
+      UpdateState& update_state_;
+    };
+
```
szaszm, line 171: Please check the rule of 5 for this class. I think copying the object will result in double decrease and underflowing update count.

libminifi/src/FlowController.cpp
lordgamez: Updated in 1fd1685fbccb60386133baeee87ddcffa4747487

libminifi/include/FlowController.h
lordgamez: The latter sounds good, thanks, updated in 1fd1685fbccb60386133baeee87ddcffa4747487

libminifi/include/c2/ControllerSocketProtocol.h
```diff
@@ -61,6 +66,46 @@ class ControllerSocketProtocol {
   std::weak_ptr<ControllerSocketReporter> controller_socket_reporter_;
   std::shared_ptr<Configure> configuration_;
   std::shared_ptr<core::logging::Logger> logger_ = core::logging::LoggerFactory<ControllerSocketProtocol>::getLogger();
+  std::mutex initialization_mutex_;
+
+  // Some commands need to restart the controller socket to reinitialize the socket with new data for example new SSL data in case of a flow update
+  // These commands are handled on a separate thread, and while these commands are handled other incoming commands are dropped
+  class SocketRestartCommandProcessor {
```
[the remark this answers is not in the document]
lordgamez: The reason the controller socket is inside the FlowController is because it can be effected by the flow update. If the controller socket uses SSL the SSL context may be read from the flow configuration file. If a the flow config changes it has to be restarted in case there is a change in the used SSLContextService.

libminifi/include/c2/ControllerSocketProtocol.h
```diff
@@ -61,6 +66,46 @@ class ControllerSocketProtocol {
...
+    }
+
+   private:
+    mutable std::atomic_bool is_socket_restarting_ = false;
+    state::StateMonitor& update_sink_;
+    std::thread command_processor_thread_;
+    std::mutex cv_mutex_;
+    std::condition_variable command_queue_condition_variable_;
+    std::atomic_bool running_ = true;
+    moodycamel::ConcurrentQueue<CommandData> command_queue_;
```
szaszm, line 105: The minifi `utils::ConditionConcurrentQueue` could be a better alternative here. It's not as performant in high load scenarios, but performance doesn't really matter here, and it's less likely to change the order of commands, and it has builtin waiting dequeue functions.

`moodycamel::ConcurrentQueue` works with per-thread arrays, and the consumer always picks the first non-empty thread array. This means that it doesn't keep any ordering between different sources, and it may return elements that were inserted later if they happen to be in the first bucket/array. Strictly speaking, this still gives happens-before guarantees within a producer thread, but nothing between threads. If the second thread inserted something much earlier, it will still return the item from the first thread if there are any. We used to have strange priority and CPU starving issues when we used to use this queue for scheduling processors. It's unlikely to cause any issues here, other than occasional command reordering between multiple independent clients.

https://github.com/cameron314/concurrentqueue#high-level-design

docker/test/integration/features/minifi_controller.feature
```diff
@@ -0,0 +1,76 @@
...
+
+  Scenario: FlowController can be stopped and restarted
+    Given a GenerateFlowFile processor
+    And a file with the content "test" is present in "/tmp/input"
+    And controller socket properties are set up
+    When all instances start up
+    And the FlowController component is stopped through MiNiFi controller
+    And the FlowController component is started through MiNiFi controller
+    Then the GenerateFlowFile component is running
+    And the FlowController component is running
```
adamdebreceni, line 58: this seems to do the same setup as the previous one, yet it expects the opposite result

libminifi/include/c2/ControllerSocketProtocol.h
lordgamez: Thanks for the detailed description, good to know these differences. I replaced the in `moodycamel::ConcurrentQueue` with the `utils::ConditionConcurrentQueue` in 87bce05ddabb9a2ee89ecac9681c123c0adbb17a

cmake/MiNiFiOptions.cmake
```diff
@@ -120,6 +120,8 @@ add_minifi_option(ENABLE_KUBERNETES "Enables the Kubernetes extensions." ON)
 add_minifi_option(ENABLE_TEST_PROCESSORS "Enables test processors" OFF)
 add_minifi_option(ENABLE_PROMETHEUS "Enables Prometheus support." ON)
 add_minifi_option(DISABLE_JEMALLOC "Disables jemalloc." OFF)
+add_minifi_option(DISABLE_CURL "Disables curl." OFF)
+add_minifi_option(DISABLE_CONTROLLER "Disables build of MiNiFi controller binary." OFF)
```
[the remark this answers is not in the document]
lordgamez: These options were present already so I didn't want to change them as they may break our builds. The options are only added here to be available for the docker builds as well.

docker/test/integration/features/minifi_controller.feature
lordgamez: Good catch! As the start operation is not immediate the previous test could still pass with the negative checks before the start operation was finished. I fixed the test in 87bce05ddabb9a2ee89ecac9681c123c0adbb17a

cmake/MiNiFiOptions.cmake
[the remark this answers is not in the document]
adamdebreceni: I see, thanks for clarifying

PR #1576, opened 2023-05-18
MINIFICPP-2121 - Use std::atomic_flag instead of semaphore

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/minifi_main/MiNiFiMain.cpp
+++ b/minifi_main/MiNiFiMain.cpp
@@ -75,9 +75,8 @@ namespace minifi = org::apache::nifi::minifi;
 namespace core = minifi::core;
 namespace utils = minifi::utils;
 
- // Variables that allow us to avoid a timed wait.
-static sem_t *flow_controller_running;
-static sem_t *process_running;
+static std::atomic_flag flow_controller_running;
+static std::atomic_flag process_running;
 
 /**
  * Removed the stop command from the signal handler so that we could trigger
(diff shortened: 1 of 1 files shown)


minifi_main/MiNiFiMain.cpp
```diff
@@ -91,27 +90,26 @@ static sem_t *process_running;
...
+    process_running.wait(false);
     return TRUE;
   }
   return FALSE;
 }
 
 void SignalExitProcess() {
-  sem_post(flow_controller_running);
+  flow_controller_running.clear();
+  flow_controller_running.notify_all();
```
szaszm, line 105: Is this approach signal safe?

adamdebreceni: according to the standard calling a non-static member function of `std::atomic_flag` is a "plain lock-free atomic operation" which is allowed in signal handlers

minifi_main/MiNiFiMain.cpp
```diff
@@ -91,27 +90,26 @@ static sem_t *process_running;
...
   if (signal == CTRL_C_EVENT || signal == CTRL_BREAK_EVENT) {
-    int ret = ETIMEDOUT;
-    while (ret == ETIMEDOUT) {
-      if (flow_controller_running) { sem_post(flow_controller_running); }
-      const struct timespec timeout_100ms { .tv_sec = 0, .tv_nsec = 100000000};
-      ret = sem_timedwait(process_running, &timeout_100ms);
-    }
+    flow_controller_running.clear();
+    flow_controller_running.notify_all();
+    process_running.wait(false);
```
fgerlits, line 97: should this be `wait(true)`?  from the docs, it looks to me that `wait(x)` waits for the flag to be _not_ x, and we want to wait for `process_running` to become `false` (ie, not `true`)

adamdebreceni: good catch, indeed that seems to be the case

lordgamez (approved):

fgerlits (approved):

szaszm (approved):

PR #1577, opened 2023-05-19
MINIFICPP-2020 - Protect MINIFI_HOME from mutual access

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/utils/FileMutex.h
+++ b/libminifi/include/utils/FileMutex.h
@@ -21,11 +21,18 @@
 #include <array>
 #include <mutex>
 
+#ifdef WIN32
+#include <windows.h>
+#endif
+
 namespace org::apache::nifi::minifi::utils {
 
 class FileMutex {
  public:
   explicit FileMutex(std::filesystem::path path);
+  ~FileMutex() {
+    gsl_Expects(!file_handle_.has_value());
+  }
 
   FileMutex(const FileMutex&) = delete;
   FileMutex(FileMutex&&) = delete;
@@ -39,7 +46,11 @@ class FileMutex {
   std::filesystem::path path_;
 
   std::mutex mtx_;
-  std::array<std::byte, 24> file_handle_;
+#ifdef WIN32
+  std::optional<HANDLE> file_handle_;
+#else
+  std::optional<int> file_handle_;
+#endif
 };
 
 }  // namespace org::apache::nifi::minifi::utils
--- a/libminifi/include/utils/FileMutex.h
+++ b/libminifi/include/utils/FileMutex.h
@@ -1,0 +1,45 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <filesystem>
+#include <array>
+#include <mutex>
+
+namespace org::apache::nifi::minifi::utils {
+
+class FileMutex {
+ public:
+  explicit FileMutex(std::filesystem::path path);
+
+  FileMutex(const FileMutex&) = delete;
+  FileMutex(FileMutex&&) = delete;
+  FileMutex& operator=(const FileMutex&) = delete;
+  FileMutex& operator=(FileMutex&&) = delete;
+
+  void lock();
+  void unlock();
+
+ private:
+  std::filesystem::path path_;
+
+  std::mutex mtx_;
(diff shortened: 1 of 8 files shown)


libminifi/test/unit/FileMutexTests.cpp
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  TestController controller;
+  auto lock_file = controller.createTempDirectory() / "LOCK";
+
+  std::cout << "Locking file the first time '" << lock_file << "'" << std::endl;
+  minifi::utils::FileMutex mtx{lock_file};
+  std::unique_lock lock{mtx};
+
+  int second_lock = std::system((utils::file::get_executable_path().string() + " " + lock_file.string()).c_str());
```
martinzink, line 42: I don't like this convoluted way of testing. I get that this mimics the way how it is setup in minifiexe, but since that code is not shared I think it might be better if we just use catch2 and test the filelock by itself (create a lock and then expect throw when trying to relock it)

adamdebreceni: although that would certainly simplify things, that would not test the only purpose of FileMutex: inter-process locking

martinzink: Good point you are right, but I think we should include test cases for unlocking aswell (if the lock goes out of scope)

martinzink, line 42: Clangtidy doesnt like the system call, could you check that?

libminifi/test/unit/FileMutexTests.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+  assert(second_lock != 0);
```
martinzink, line 44: I think we should include a test case for unlocking aswell (if the lock goes out of scope).

martinzink (approved):

libminifi/src/utils/FileMutex.cpp
```diff
@@ -0,0 +1,107 @@
...
+#ifdef O_CLOEXEC
+  flags |= O_CLOEXEC;
+#endif
+  int fd = open(path_.string().c_str(), flags, 0644);
+  if (fd < 0) {
+    throw std::runtime_error("Failed to open file '" + path_.string() + "' to be locked: " + std::strerror(errno));
+  }
+
+  errno = 0;
+  struct flock f{};
```
lordgamez, line 79: Please rename to something more descriptive

libminifi/src/utils/FileMutex.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+void FileMutex::unlock() {
+  std::lock_guard guard(mtx_);
+  errno = 0;
+  struct flock f{};
+  f.l_type = F_UNLCK;
+  int value = fcntl(getOsHandle<int>(file_handle_), F_SETLK, &f);
+  if (value == -1) {
+    std::string err_str = "Failed to unlock file '" + path_.string() + "': " + std::strerror(errno);
+    close(getOsHandle<int>(file_handle_));
```
lordgamez, line 99: Maybe this could be set in `gsl::finally` instead

libminifi/src/utils/FileMutex.cpp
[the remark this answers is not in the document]
adamdebreceni: I've raised this question before, the feedback was that it was a pain and we have actively moved away from this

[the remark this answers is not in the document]
lordgamez: Okay, thanks!

libminifi/src/utils/FileMutex.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include "utils/gsl.h"
+
+template<typename T>
+static T& getOsHandle(std::array<std::byte, 24>& file_handle) {
+  void* ptr = reinterpret_cast<void*>(file_handle.data());
+  size_t size = file_handle.size();
+  void* result = std::align(alignof(T), sizeof(T), ptr, size);
+  gsl_Assert(result);
+  return *reinterpret_cast<T*>(result);
+}
```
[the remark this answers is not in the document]
adamdebreceni: done, made file_handle_ OS-specific

libminifi/src/utils/FileMutex.cpp
adamdebreceni: renamed

libminifi/src/utils/FileMutex.cpp
adamdebreceni: added

libminifi/include/utils/FileMutex.h
```diff
@@ -0,0 +1,58 @@
...
+
+ private:
+  std::filesystem::path path_;
+
+  std::mutex mtx_;
+#ifdef WIN32
+  std::optional<HANDLE> file_handle_;
+#else
+  std::optional<int> file_handle_;
+#endif
```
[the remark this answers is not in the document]
adamdebreceni: added pid to lock file

libminifi/src/utils/FileMutex.cpp
```diff
@@ -0,0 +1,171 @@
...
+#include "utils/OsUtils.h"
+#include "utils/Error.h"
+
+#ifdef WIN32
+
+namespace org::apache::nifi::minifi::utils {
+
+FileMutex::FileMutex(std::filesystem::path path): path_(std::move(path)) {}
+
+// we cannot assume the logging system to be initialized
```
szaszm, line 32: As long as it's only used in one place, it's fine like this. If it's ever needed somewhere, where logging is available, you can pass in a log callback to make logging polymorphic and independent of the logging facilities.

szaszm (approved):

PR #1581, opened 2023-05-31
MINIFICPP-2125 fix for waking up prematurely after processor yields

martinzink:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/utils/ThreadPool.h
+++ b/libminifi/include/utils/ThreadPool.h
@@ -106,10 +106,6 @@ class Worker {
     return next_exec_time_;
   }
 
-  virtual std::chrono::milliseconds getWaitTime() const {
-    return run_determinant_->wait_time();
-  }
-
 
   std::shared_ptr<std::promise<T>> getPromise() const;
 
--- a/libminifi/include/utils/ThreadPool.h
+++ b/libminifi/include/utils/ThreadPool.h
@@ -94,7 +94,7 @@ class Worker {
       promise->set_value(result);
       return false;
     }
-    next_exec_time_ = std::max(next_exec_time_ + run_determinant_->wait_time(), std::chrono::steady_clock::now());
+    next_exec_time_ = std::max(next_exec_time_, std::chrono::steady_clock::now() + run_determinant_->wait_time());
     return true;
   }
 
--- a/libminifi/include/utils/ThreadPool.h
+++ b/libminifi/include/utils/ThreadPool.h
@@ -106,7 +106,6 @@ class Worker {
     return next_exec_time_;
   }
 
-
   std::shared_ptr<std::promise<T>> getPromise() const;
 
   const TaskId &getIdentifier() const {
--- a/libminifi/src/TimerDrivenSchedulingAgent.cpp
+++ b/libminifi/src/TimerDrivenSchedulingAgent.cpp
@@ -33,8 +33,7 @@ utils::TaskRescheduleInfo TimerDrivenSchedulingAgent::run(core::Processor* proce
     if (processor->isYield())
       return utils::TaskRescheduleInfo::RetryIn(processor->getYieldTime());
 
-    auto next_execution_time_point = trigger_start_time + processor->getSchedulingPeriod();
-    return utils::TaskRescheduleInfo::RetryIn(next_execution_time_point - std::chrono::steady_clock::now());
+    return utils::TaskRescheduleInfo::RetryAfter(trigger_start_time + processor->getSchedulingPeriod());
   }
   return utils::TaskRescheduleInfo::Done();
 }
--- a/libminifi/src/TimerDrivenSchedulingAgent.cpp
+++ b/libminifi/src/TimerDrivenSchedulingAgent.cpp
@@ -28,11 +28,13 @@ namespace org::apache::nifi::minifi {
 utils::TaskRescheduleInfo TimerDrivenSchedulingAgent::run(core::Processor* processor, const std::shared_ptr<core::ProcessContext> &processContext,
                                          const std::shared_ptr<core::ProcessSessionFactory> &sessionFactory) {
   if (this->running_ && processor->isRunning()) {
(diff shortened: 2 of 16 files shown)


libminifi/include/utils/ThreadPool.h
```diff
@@ -94,7 +94,7 @@ class Worker {
       promise->set_value(result);
       return false;
     }
-    next_exec_time_ = std::max(next_exec_time_ + run_determinant_->wait_time(), std::chrono::steady_clock::now());
+    next_exec_time_ = std::max(next_exec_time_, std::chrono::steady_clock::now() + run_determinant_->wait_time());
```
szaszm, line 97: This seems to change behavior, especially for long running processors: If a processor is scheduled to run once every 5 seconds, and it takes 4 seconds to run, then I'd expect the next execution to happen 1 second after finishing execution, not 5 seconds later. The latter would mean effectively running every 9 seconds instead of the configured 5.

libminifi/test/unit/ThreadPoolTests.cpp
szaszm, line 1: What is the function called `function` used for at line 28? Can we remove it if it's unused?

libminifi/src/core/Processor.cpp
```diff
@@ -372,26 +372,26 @@ void Processor::setMaxConcurrentTasks(const uint8_t tasks) {
...
 
 void Processor::clearYield() {
-  yield_expiration_ = std::chrono::system_clock::time_point();
+  yield_expiration_ = std::chrono::steady_clock::time_point();
 }
 
 std::chrono::milliseconds Processor::getYieldTime() const {
   auto yield_expiration = yield_expiration_.load();
-  auto current_time = std::chrono::system_clock::now();
+  auto current_time = std::chrono::steady_clock::now();
```
szaszm, line 392: Now that timestamps are changed from system_clock to steady_clock, does this affect serialized timestamps? I don't think `steady_clock` is guaranteed to be steadily increasing in sync with real time between executions or reboots, so reading persisted timestamps anywhere may result in incorrect behavior.

libminifi/src/utils/ThreadPool.cpp
```diff
@@ -151,10 +153,10 @@ void ThreadPool<T>::manageWorkers() {
 
   if (nullptr != thread_manager_) {
     while (running_) {
-      auto waitperiod = std::chrono::milliseconds(500);
+      auto wait_period = 500ms;
       {
-        std::unique_lock<std::recursive_mutex> lock(manager_mutex_, std::try_to_lock);
-        if (!lock.owns_lock()) {
+        std::unique_lock<std::recursive_mutex> manager_lock(manager_mutex_, std::try_to_lock);
+        if (!manager_lock.owns_lock()) {
```
szaszm, line 159: It looks like this mutex lock should be retried after waiting. Currently it just proceeds to access the shared data without synchronization, after waiting a bit, if the first lock attempt fails.

libminifi/src/core/Processor.cpp
martinzink: afaik we dont serialize these anywhere

libminifi/test/unit/ThreadPoolTests.cpp
```diff
@@ -84,7 +81,33 @@ TEST_CASE("ThreadPoolTest2", "[TPT2]") {
...
   std::future<int> fut;
-  pool.execute(std::move(functor), fut);  // NOLINT(bugprone-use-after-move)
+  pool.execute(std::move(functor), fut);
   fut.wait();
   REQUIRE(20 == fut.get());
 }
+
+TEST_CASE("Worker wait time should be relative to the last run") {
+  std::mutex cv_m;
+  std::condition_variable cv;
```
szaszm, line 79: You could use `std::promise<void>` or `std::atomic_flag` (like #1576) for a simpler code.

martinzink: I've simplified this test in https://github.com/apache/nifi-minifi-cpp/pull/1581/commits/187f07eb04efef01712e81cd9f3fc1dfa93bc93d

libminifi/src/TimerDrivenSchedulingAgent.cpp
```diff
@@ -28,11 +28,13 @@ namespace org::apache::nifi::minifi {
...
                                          const std::shared_ptr<core::ProcessSessionFactory> &sessionFactory) {
   if (this->running_ && processor->isRunning()) {
+    auto trigger_start_time = std::chrono::steady_clock::now();
     this->onTrigger(processor, processContext, sessionFactory);
     if (processor->isYield())
       return utils::TaskRescheduleInfo::RetryIn(processor->getYieldTime());
 
-    return utils::TaskRescheduleInfo::RetryIn(std::chrono::duration_cast<std::chrono::milliseconds>(processor->getSchedulingPeriodNano()));
+    auto next_execution_time_point = trigger_start_time + processor->getSchedulingPeriod();
+    return utils::TaskRescheduleInfo::RetryIn(next_execution_time_point - std::chrono::steady_clock::now());
```
szaszm, line 37: If this is negative, we shouldn't schedule processors in the past.

szaszm: I think this is surprising behavior, and it would be better to have a narrow contract precondition in the constructor, and calculate it where it's called.

adamdebreceni (approved):

libminifi/src/TimerDrivenSchedulingAgent.cpp
martinzink: good point, I've changed the behaviour in https://github.com/apache/nifi-minifi-cpp/pull/1581/commits/adac20e8318a942ba0220a3a4b71b889aa6675ae, and created a helper function that does this calculation (its based on the time_point so I think thats expectable)

szaszm (approved):

PR #1592, opened 2023-06-20
MINIFICPP-2131 Refactored GetTCP

martinzink:
**This breaks compatibility with old configs containing GetTCP**

Reworked GetTCP because it was unsafe with multiple endpoints, didnt work like NiFi's implementation, used too much unneccesary resources, and lacked features that are present in similar processors (e.g. ListenTCP/ListenSyslog).

Previously it had a private threadpool with a configurable amount of private threads that would connect to the configured endpoints and read from them creating flowfiles from these background threads.

After this PR, it will work similarly to ListenTCP with the main difference that GetTCP will initiate the connetion. (client vs server role).

---

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/standard-processors/processors/GetTCP.cpp
```diff
@@ -17,275 +17,287 @@
...
 
-  stay_connected_ = true;
-  if (context->getProperty(StayConnected.getName(), value)) {
-    stay_connected_ = utils::StringUtils::toBool(value).value_or(true);
+  char delimiter = '\n';
+  if (auto delimiter_str = context->getProperty(MessageDelimiter)) {
+    auto parsed_delimiter = utils::StringUtils::parseCharacter(*delimiter_str);
+    if (!parsed_delimiter)
+      throw Exception(PROCESS_SCHEDULE_EXCEPTION, fmt::format("Invalid delimiter: {} (it must be a single (escaped or not) character", *delimiter_str));
+    delimiter = **parsed_delimiter;
```
adamdebreceni, line 118: won't this only check if the returned value is expected, i.e. an `std::optional<char>`, so dereferencing it is not yet safe?

extensions/standard-processors/tests/unit/GetTCPTests.cpp
```diff
@@ -15,391 +14,286 @@
...
-  logAttribute->onTrigger(context2, session2);
+ private:
+  asio::awaitable<void> sendMessages(auto& socket) {
+    while (true) {
+      std::string message_to_send;
+      if (!messages_to_send_.tryDequeue(message_to_send)) {
+        co_await minifi::utils::net::async_wait(10ms);
+        continue;
+      }
+      co_await asio::async_write(socket, asio::buffer(message_to_send), minifi::utils::net::use_nothrow_awaitable);
```
adamdebreceni, line 105: how is this loop terminated?

extensions/standard-processors/processors/GetTCP.cpp
martinzink: you are right, I've fixed this (unsuccesfully) in https://github.com/apache/nifi-minifi-cpp/pull/1592/commits/9dce92909600a74ba320bd3c56612acd0c8ab5fa (and fixed the fix in https://github.com/apache/nifi-minifi-cpp/pull/1592/commits/c824ab601e3bbdc01177a55496096ddc625356b1)

libminifi/test/resources/TestC2Metrics.yml
```diff
@@ -31,8 +31,8 @@ Processors:
       run duration nanos: 0
       auto-terminated relationships list:
       Properties:
-          endpoint-list: localhost:8776
-          end-of-message-byte: d
+          Endpoint List: localhost:8776
+          Message Delimiter: \r
           reconnect-interval: 100ms
           connection-attempt-timeout: 2000
```
adamdebreceni, line 37: should we update the properties to their new names?

martinzink: sure thing, best not the confuse anyone :+1: https://github.com/apache/nifi-minifi-cpp/pull/1592/commits/ed21b198d8fb54a4f1c483b63cf9f1013ecee5f9

adamdebreceni (approved):

extensions/standard-processors/processors/GetTCP.cpp
```diff
@@ -17,275 +17,287 @@
...
-    socket_ring_buffer_.try_dequeue(socket_ptr);
+  if (client_)
+    client_->stop();
+}
+
+void GetTCP::transferAsFlowFile(const utils::net::Message& message, core::ProcessSession& session) {
+  auto flow_file = session.create();
+  session.writeBuffer(flow_file, message.message_data);
+  flow_file->setAttribute("tcp.port", std::to_string(message.server_port));
+  flow_file->setAttribute("tcp.sender", message.sender_address.to_string());
```
fgerlits, line 165: I think a single `tcp.sender` output attribute containing `host:port` would be easier to understand.  This way, it's not clear if `tcp.port` is the server port or client port.  Alternatively, we could have `tcp.sender.host` and `tcp.sender.port`.

extensions/standard-processors/processors/GetTCP.cpp
```diff
@@ -17,275 +17,287 @@
...
-#endif
-    if (hostAndPort.size() != 2) {
+
+    if (bytes_read == 0)
       continue;
+
+    if (!max_queue_size_ || max_queue_size_ > concurrent_queue_.size()) {
+      utils::net::Message message{read_message.substr(0, bytes_read), utils::net::IpProtocol::TCP, socket.lowest_layer().remote_endpoint().address(), socket.lowest_layer().remote_endpoint().port()};
+      if (last_was_partial || current_is_partial)
+        message.is_partial = true;
```
fgerlits, line 253: why is this message partial if the previous message was partial?

martinzink: Partial messages happen when we would read more than the configured max_size and the delimiter has not ben found. In that case the message (larger than the maxsize) would have been split at max_size intervals which would make all the flowfiles which didnt end with the delimiter partial and the next one aswell (thats probably the last part of the message (ending with delimiter)).

extensions/standard-processors/processors/GetTCP.cpp
martinzink: You are right, originally I mimicked how ListenTCP works, but this might be better. Also Nifi works the same way in GetTCP, so I went with `source.endpoint` in https://github.com/apache/nifi-minifi-cpp/pull/1592/commits/54e037239cd49365a89976cc280c7bfcc0e198aa similar to Nifi https://github.com/apache/nifi/blob/main/nifi-nar-bundles/nifi-tcp-bundle/nifi-tcp-processors/src/main/java/org/apache/nifi/processors/gettcp/GetTCP.java#L61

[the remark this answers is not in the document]
martinzink: Good idea, changed it in https://github.com/apache/nifi-minifi-cpp/pull/1592/commits/54e037239cd49365a89976cc280c7bfcc0e198aa

extensions/standard-processors/processors/GetTCP.cpp
```diff
@@ -17,275 +17,287 @@
...
+       connections_to_make.emplace_back(hostname_service_pair[0], hostname_service_pair[1]);
+     }
   }
 
-  int handlers = 0;
-  if (context->getProperty(ConcurrentHandlers.getName(), handlers)) {
-    concurrent_handlers_ = handlers;
-  }
+  if (connections_to_make.empty())
+    throw Exception(PROCESS_SCHEDULE_EXCEPTION, "No valid endpoint in Endpoint List property");
```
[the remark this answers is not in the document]
martinzink: good idea, changed it in https://github.com/apache/nifi-minifi-cpp/pull/1592/commits/54e037239cd49365a89976cc280c7bfcc0e198aa#diff-8d484e99475c978743af82b7ee8ba45954457ecbac4c033744d1cc95ad9f8240R112

extensions/standard-processors/processors/GetTCP.cpp
```diff
@@ -17,275 +17,287 @@
...
+    auto parsed_delimiter = utils::StringUtils::parseCharacter(*delimiter_str);
+    if (!parsed_delimiter || !parsed_delimiter->has_value())
+      throw Exception(PROCESS_SCHEDULE_EXCEPTION, fmt::format("Invalid delimiter: {} (it must be a single (escaped or not) character", *delimiter_str));
+    delimiter = **parsed_delimiter;
   }
 
-  int connects = 0;
-  if (context->getProperty(ConnectionAttemptLimit.getName(), connects)) {
-    connection_attempt_limit_ = connects;
+  std::optional<asio::ssl::context> ssl_context_;
```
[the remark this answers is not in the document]
martinzink: good catch changed it in https://github.com/apache/nifi-minifi-cpp/pull/1592/commits/54e037239cd49365a89976cc280c7bfcc0e198aa#diff-8d484e99475c978743af82b7ee8ba45954457ecbac4c033744d1cc95ad9f8240R122

extensions/standard-processors/processors/GetTCP.cpp
```diff
@@ -17,275 +17,287 @@
...
+  flow_file->setAttribute("tcp.port", std::to_string(message.server_port));
+  flow_file->setAttribute("tcp.sender", message.sender_address.to_string());
+  if (message.is_partial)
+    session.transfer(flow_file, Partial);
+  else
+    session.transfer(flow_file, Success);
+}
+
+void GetTCP::onTrigger(const std::shared_ptr<core::ProcessContext>&, const std::shared_ptr<core::ProcessSession>& session) {
+  gsl_Expects(session && max_batch_size_ > 0);
```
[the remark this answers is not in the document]
martinzink: Good catch, I've added PROCESS_SCHEDULE_EXCEPTION in https://github.com/apache/nifi-minifi-cpp/pull/1592/commits/54e037239cd49365a89976cc280c7bfcc0e198aa#diff-8d484e99475c978743af82b7ee8ba45954457ecbac4c033744d1cc95ad9f8240R139-R141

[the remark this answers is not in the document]
martinzink: Good catch, I've added PROCESS_SCHEDULE_EXCEPTION in https://github.com/apache/nifi-minifi-cpp/pull/1592/commits/54e037239cd49365a89976cc280c7bfcc0e198aa#diff-8d484e99475c978743af82b7ee8ba45954457ecbac4c033744d1cc95ad9f8240R139-R141

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -343,10 +331,11 @@ void TailFile::onSchedule(const std::shared_ptr<core::ProcessContext> &context,
...
 
-  std::string value;
-
-  if (context->getProperty(Delimiter.getName(), value)) {
-    delimiter_ = parseDelimiter(value);
+  if (auto delimiter_str = context->getProperty(Delimiter)) {
+    auto parsed_delimiter = utils::StringUtils::parseCharacter(*delimiter_str);
+    if (!parsed_delimiter)
+      throw Exception(PROCESS_SCHEDULE_EXCEPTION, fmt::format("Invalid delimiter: {} (it must be a single character, whether escaped or not)", *delimiter_str));
+    delimiter_ = *parsed_delimiter;
```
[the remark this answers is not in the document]
martinzink: I've reverted to a less radical variant. We will log errors but continue like before. https://github.com/apache/nifi-minifi-cpp/pull/1592/commits/dcabfe37f396f64ad15d292442ca30a8865b830c

extensions/standard-processors/processors/GetTCP.cpp
fgerlits: Makes sense, thanks!

fgerlits (approved):

extensions/standard-processors/processors/GetTCP.cpp
martinzink: On second thought it might be easier to understand if I simply rename these bools https://github.com/apache/nifi-minifi-cpp/pull/1592/commits/91d8d09ddf13e2abc56314cb99dbeeb0019a4107

extensions/standard-processors/processors/GetTCP.cpp
```diff
@@ -17,275 +17,290 @@
...
   setSupportedProperties(properties());
   setSupportedRelationships(relationships());
 }
 
-void GetTCP::onSchedule(const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSessionFactory> &sessionFactory) {
-  std::string value;
-  if (context->getProperty(EndpointList.getName(), value)) {
-    endpoints = utils::StringUtils::split(value, ",");
+void GetTCP::onSchedule(const std::shared_ptr<core::ProcessContext>& context, const std::shared_ptr<core::ProcessSessionFactory>&) {
+  std::vector<utils::net::ConnectionId> connections_to_make;
```
lordgamez, line 99: I would extract parsing the endpoints, parsing the delimiter and getting the ssl_context into separate functions.

extensions/standard-processors/processors/GetTCP.cpp
```diff
@@ -17,275 +17,290 @@
...
-
-  for (auto &initEndpoint : endpoints) {
-    std::vector<std::string> hostAndPort = utils::StringUtils::split(initEndpoint, ":");
-    auto realizedHost = hostAndPort.at(0);
-#ifdef WIN32
-    if ("localhost" == realizedHost) {
-      realizedHost = org::apache::nifi::minifi::io::Socket::getMyHostName();
+
+GetTCP::TcpClient::TcpClient(char delimiter,
+    asio::steady_timer::duration timeout_duration,
```
lordgamez, line 188: I think the parameters should only use 1 indentation level to not be on the same level as the initializer list

lordgamez (approved):
LGTM, only minor requests

extensions/standard-processors/processors/GetTCP.cpp
martinzink: good idea, extracted these in https://github.com/apache/nifi-minifi-cpp/pull/1592/commits/1c380c824e18c8c0ec2ae6eada4ba4644e444feb

extensions/standard-processors/processors/GetTCP.cpp
martinzink: Yeah looks strange, I agree. I've checked the [styleguide](https://google.github.io/styleguide/cppguide.html#Constructor_Initializer_Lists) and it looks like we should align the initializer list with the first line (i.e 6 spaces). I've checked and we are using this style and your recommendation extensively in the codebase. I've went with the styleguide recommendation. https://github.com/apache/nifi-minifi-cpp/pull/1592/commits/1c380c824e18c8c0ec2ae6eada4ba4644e444feb#diff-8d484e99475c978743af82b7ee8ba45954457ecbac4c033744d1cc95ad9f8240R218-R224

PR #1595, opened 2023-06-26
MINIFICPP-2137 Rewrite MiNiFi Controller to use asio

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2137

Note: depends on https://github.com/apache/nifi-minifi-cpp/pull/1583 as there was a compile issue with LibreSSL on Windows that does not appear with OpenSSL. It seems to be due to some Windows specific macro definition. If we plan to move to OpenSSL I don't think it's worth debugging.

```
2023-06-23T15:41:12.8423242Z D:\a\nifi-minifi-cpp\b\thirdparty\libressl-install\include\openssl/x509v3.h(194): error C2143: syntax error: missing ')' before 'constant'
2023-06-23T15:41:12.8558554Z D:\a\nifi-minifi-cpp\b\thirdparty\libressl-install\include\openssl/x509v3.h(194): error C2143: syntax error: missing ';' before 'constant'
2023-06-23T15:41:12.8582850Z D:\a\nifi-minifi-cpp\b\thirdparty\libressl-install\include\openssl/x509v3.h(194): error C2059: syntax error: ')'
2023-06-23T15:41:12.8727617Z D:\a\nifi-minifi-cpp\b\thirdparty\libressl-install\include\openssl/x509v3.h(194): error C2238: unexpected token(s) preceding ';'
2023-06-23T15:41:12.8746940Z D:\a\nifi-minifi-cpp\b\thirdparty\libressl-install\include\openssl/x509v3.h(202): error C2143: syntax error: missing ')' before 'constant'
2023-06-23T15:41:12.8752091Z D:\a\nifi-minifi-cpp\b\thirdparty\libressl-install\include\openssl/x509v3.h(202): error C2143: syntax error: missing ';' before 'constant'
2023-06-23T15:41:12.8754426Z D:\a\nifi-minifi-cpp\b\thirdparty\libressl-install\include\openssl/x509v3.h(202): error C2059: syntax error: ')'
2023-06-23T15:41:12.8876950Z D:\a\nifi-minifi-cpp\b\thirdparty\libressl-install\include\openssl/x509v3.h(202): error C2238: unexpected token(s) preceding ';'
2023-06-23T15:41:12.9189256Z D:\a\nifi-minifi-cpp\b\thirdparty\libressl-install\include\openssl/x509v3.h(231): error C2143: syntax error: missing ')' before 'constant'
[...]

PR #1599, opened 2023-07-04
MINIFICPP-2152 Remove own server and client socket implementations

lordgamez:
- Rewrite SiteToSite and CSite2SiteTests to use Asio library instead of custom socket implementation
- Remove custom socket implementation classes and related utils and tests: ClientSocket.h, DescriptorStream.h, EndianCheck.h, ServerSocket.h, Sockets.h, StreamFactory.h, SecureDescriptorStream.h, TLSServerSocket.h, TLSSocket.h
- Remove ProvenanceReportingTest as it was not part of the test suite, it was failing and was unmaintained
- Remove SecureSocketGetTCPTests as it will be replaced in https://issues.apache.org/jira/browse/MINIFICPP-2154

Depends on https://github.com/apache/nifi-minifi-cpp/pull/1592 and https://github.com/apache/nifi-minifi-cpp/pull/1595

https://issues.apache.org/jira/browse/MINIFICPP-2152

-----------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/src/utils/net/DNS.cpp
```diff
@@ -127,4 +69,9 @@ nonstd::expected<std::string, std::error_code> reverseDnsLookup(const asio::ip::
   return results->host_name();
 }
 
+std::string getMyHostName() {
+  static const std::string HOSTNAME = asio::ip::host_name();
+  return HOSTNAME;
+}
```
szaszm, line 75: Is this static caching actually needed? Is querying the hostname too expensive?

libminifi/src/utils/net/Socket.cpp
```diff
@@ -14,19 +14,15 @@
...
-
 #include "utils/net/Socket.h"
-#include "Exception.h"
-#include <cstring>
-#include <system_error>
+
 #ifdef WIN32
-#ifndef WIN32_LEAN_AND_MEAN
-#define WIN32_LEAN_AND_MEAN
-#endif /* WIN32_LEAN_AND_MEAN */
```
szaszm, line 25: Why was this removed? This macro reduces the complexity of windows headers if defined before including, and we're still including a windows header below.

libminifi/src/utils/net/Socket.cpp
```diff
@@ -14,19 +14,15 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 #include "utils/net/Socket.h"
-#include "Exception.h"
-#include <cstring>
```
szaszm, line 20: `<cstring>` is still used for `memcpy`.

libminifi/include/utils/net/Socket.h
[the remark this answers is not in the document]
lordgamez: Updated in 1081ccfb9608f8f9dd7933cf48b66cb684276837

libminifi/include/utils/net/Socket.h
```diff
@@ -23,86 +23,16 @@
 #endif /* WIN32_LEAN_AND_MEAN */
 #include <WinSock2.h>
 #else
-#include <sys/types.h>
-#include <sys/socket.h>
```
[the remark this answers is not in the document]
lordgamez: Updated in 1081ccfb9608f8f9dd7933cf48b66cb684276837

libminifi/include/utils/net/DNS.h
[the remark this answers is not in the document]
lordgamez: Updated in 1081ccfb9608f8f9dd7933cf48b66cb684276837

libminifi/src/utils/net/DNS.cpp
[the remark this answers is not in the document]
lordgamez: Updated in 1081ccfb9608f8f9dd7933cf48b66cb684276837

libminifi/src/utils/net/DNS.cpp
lordgamez: I'm not sure how expensive is it, but as it should not change at all, I think caching it wouldn't hurt.

libminifi/src/utils/net/Socket.cpp
lordgamez: Updated in 1081ccfb9608f8f9dd7933cf48b66cb684276837

libminifi/src/utils/net/Socket.cpp
lordgamez: '<cstring>` is still there it was only moved a bit lower

libminifi/src/utils/net/DNS.cpp
szaszm: Hostname can change at runtime. On Linux, you can use the `hostname` command.

libminifi/include/core/state/nodes/DeviceInformation.h
szaszm, line 1: It would be nice to move the logic to the .cpp file, and avoid polluting other translation units with windows and other heavyweight headers.

libminifi/include/utils/net/AsioSocketUtils.h
```diff
@@ -61,6 +62,64 @@ asio::awaitable<std::tuple<std::error_code>> handshake(SslSocket& socket, asio::
...
+      return false;
+    }
+
+    return true;
+  }
+
+  bool connectTcpSocketOverSsl();
+  bool connectTcpSocket();
+
+  asio::io_context io_context_;
```
szaszm, line 116: The old stream factory could maybe be used to inject a common `io_context`, instead of spawning up unique threads for every single connection.

libminifi/src/utils/net/DNS.cpp
```diff
@@ -15,85 +15,20 @@
  * limitations under the License.
  */
 #include "utils/net/DNS.h"
+
+#ifndef WIN32
+#include <cstring>
+#endif /* WIN32 */
+
```
szaszm, line 22: What is this used for?

lordgamez: I don't think it's used anymore, removed in 9ddaa71b8aecb723fc4c8eaaf49873a5939179f4

libminifi/include/utils/net/AsioSocketUtils.h
lordgamez: The asio::io_context does not spawn new threads, it runs synchronously on the thread it was called, in this case through the socket we pass it to with the read/write operations, it should make no difference using separate io contexts for each connection.

libminifi/include/core/Processor.h
fgerlits, line 1: to fix the `clang-tidy` warning, I think it would be better to change the `name` parameter to a `string_view` rather than the `const string&` that `clang-tidy` suggests

libminifi/include/utils/net/AsioSocketUtils.h
```diff
@@ -63,6 +64,64 @@ asio::awaitable<std::tuple<std::error_code>> handshake(SslSocket& socket, asio::
...
+  explicit AsioSocketConnection(SocketData socket_data);
+  int initialize() override;
+  size_t read(std::span<std::byte> out_buffer) override {
+    gsl_Expects(stream_);
+    return stream_->read(out_buffer);
+  }
+  size_t write(const uint8_t *value, size_t len) override {
+    gsl_Expects(stream_);
+    return stream_->write(value, len);
+  }
```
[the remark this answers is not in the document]
lordgamez: Updated in 9c5203b1db3be04c310bc47be0e1199fb0c85102

libminifi/include/core/Processor.h
fgerlits: you're right, `HTTPClient` had the same problem, but `Processor` does, too

lordgamez: Updated in 9c5203b1db3be04c310bc47be0e1199fb0c85102

fgerlits (approved):

lordgamez:
> I think we could also remove the now unused close function from FileUtils.h https://github.com/apache/nifi-minifi-cpp/blob/main/libminifi/include/utils/file/FileUtils.h#L422-L428 but otherwise LGTM 👍

Good point, removed the close function in 8526c33ebf3ae5090b165bccf6b2586c0bd7db62

martinzink (approved):

libminifi/include/utils/net/AsioSocketUtils.h
```diff
@@ -63,6 +64,64 @@ asio::awaitable<std::tuple<std::error_code>> handshake(SslSocket& socket, asio::
...
+    }
+
+    asio::ip::tcp::endpoint local_endpoint(asio::ip::address::from_string(local_network_interface_), 0);
+    asio::error_code err;
+    socket.open(local_endpoint.protocol(), err);
+    if (err) {
+      logger_->log_error("Failed to open socket on network interface '%s' with the following message: '%s'", local_network_interface_, err.message());
+      return false;
+    }
+    socket.set_option(asio::ip::tcp::socket::reuse_address(true));
```
szaszm, line 105: What's the reason for this `reuse_address(true)` option? I think normally it's used to be able to quickly restart a server program that binds to the same address and port as the old process (of the same software) did. We should avoid `bind` and `reuse_address(true)` for client connections: let closed connections linger in the TIME_WAIT state, and let the dynamic binding of ports assign another random port to the client process.

libminifi/src/utils/net/AsioSocketUtils.cpp
```diff
@@ -44,4 +47,66 @@ asio::ssl::context getSslContext(const controllers::SSLContextService& ssl_conte
...
+    result = connectTcpSocket();
+  }
+  return result ? 0 : -1;
+}
+
+bool AsioSocketConnection::connectTcpSocketOverSsl() {
+  auto ssl_context = utils::net::getSslContext(*socket_data_.ssl_context_service);
+  asio::ssl::stream<asio::ip::tcp::socket> socket(io_context_, ssl_context);
+
+  bindToLocalInterface(socket.lowest_layer());
```
szaszm, line 67: As mentioned in the previous comment: client sockets (that initiate a connection) should probably not bind to anything.

libminifi/include/utils/net/AsioSocketUtils.h
lordgamez: This is only used if the `local_network_interface_` is set, used by only the `SiteToSitePeer` class. We have the option in the `SiteToSitePeer` constructor to specify a network interface to bind to and I could only port it this way to asio, although this was restricted to Windows systems on the main branch.

lordgamez: I'm not sure why `reuse_address` was used here, but I think it can be removed.

szaszm: I found the original code now, and it used `getifaddrs` to find the IP address of the specified interface, when needed. There doesn't seem to be a similar utility in asio.

How is the code here supposed to achieve this? According to the docs, `asio::ip::address::from_string` converts dotted decimal form IP addresses (or hex notation IPv6 address) to binary form, but it doesn't say anything about converting interface names to interface addresses.

lordgamez: Updated in a42e21d6c626774f5b9d38b021727a9b37c9eafe

libminifi/src/utils/net/AsioSocketUtils.cpp
lordgamez: Updated in a42e21d6c626774f5b9d38b021727a9b37c9eafe

libminifi/include/utils/net/AsioSocketUtils.h
```diff
@@ -89,28 +94,71 @@ class AsioSocketConnection : public io::BaseStream {
...
+      inet_ntop(AF_INET, &(ipv4_addr->sin_addr), ip_str, INET_ADDRSTRLEN);
+      address = ip_str;
+    } else if (item_found->ifa_addr->sa_family == AF_INET6) {
+      struct sockaddr_in6 *ipv6_addr = (struct sockaddr_in6 *)item_found->ifa_addr;
+      char ip_str[INET6_ADDRSTRLEN];
+      inet_ntop(AF_INET6, &(ipv6_addr->sin6_addr), ip_str, INET6_ADDRSTRLEN);
+      address = ip_str;
+    } else {
+      logger_->log_error("Invalid IP address family found");
+      return;
```
szaszm, line 145: I think we have a `sockaddr_ntop` helper function in Socket.h to replace this logic.

lordgamez: Good point, updated in fca226d812d928c168c05ede1a113fb87e77da25

szaszm (approved):

PR #1600, opened 2023-07-04
MINIFICPP-2133 Add TLS 1.3 support

lordgamez:
- Upgrade Asio to version 1.28.1
- Enable TLS 1.3 in version setting of `HTTPClient`
- Remove restricting TLS version to 1.2 in asio socket connections
- Enable TLS 1.3 in ListenTcpTests
- Move retrieving local and remote addresses after accepting connection to avoid reading remote address after the connection has already been closed by remote client
- Shutdown SSL properly to avoid `stream truncated` errors: https://stackoverflow.com/questions/25587403/boost-asio-ssl-async-shutdown-always-finishes-with-an-error
- Fix tests to not expect error failed handshake in case of missing certificate as of TLS 1.3 handshake failure is only detected by the client on the first read operation: https://stackoverflow.com/questions/62459802/tls-1-3-client-does-not-report-failed-handshake-when-client-certificate-verifica
```
By the time the client sends its "Finished" message, it has already received the "Finished" and so the handshake has completed and it can immediately start sending application data.

This of course means that the client won't know whether the server has accepted the certificate or not until it next reads data from the server.
```
- Only enable legacy SSL versions if legacy version is requested

Depends on https://github.com/apache/nifi-minifi-cpp/pull/1599

https://issues.apache.org/jira/browse/MINIFICPP-2133

-------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/http-curl/client/HTTPClient.cpp
+++ b/extensions/http-curl/client/HTTPClient.cpp
@@ -167,6 +167,11 @@ bool HTTPClient::setSpecificSSLVersion(utils::SSLVersion specific_version) {
         ssl_context_service_->setMaxTlsVersion(TLS1_2_VERSION);
         break;
       }
+      case utils::SSLVersion::TLSv1_3: {
+        ssl_context_service_->setMinTlsVersion(TLS1_3_VERSION);
+        ssl_context_service_->setMaxTlsVersion(TLS1_3_VERSION);
+        break;
+      }
       default: break;
     }
   }
@@ -181,6 +186,8 @@ bool HTTPClient::setSpecificSSLVersion(utils::SSLVersion specific_version) {
       return CURLE_OK == curl_easy_setopt(http_session_.get(), CURLOPT_SSLVERSION, static_cast<int>(CURL_SSLVERSION_TLSv1_1) | static_cast<int>(CURL_SSLVERSION_MAX_TLSv1_1));
     case utils::SSLVersion::TLSv1_2:
       return CURLE_OK == curl_easy_setopt(http_session_.get(), CURLOPT_SSLVERSION, static_cast<int>(CURL_SSLVERSION_TLSv1_2) | static_cast<int>(CURL_SSLVERSION_MAX_TLSv1_2));
+    case utils::SSLVersion::TLSv1_3:
+      return CURLE_OK == curl_easy_setopt(http_session_.get(), CURLOPT_SSLVERSION, static_cast<int>(CURL_SSLVERSION_TLSv1_3) | static_cast<int>(CURL_SSLVERSION_MAX_TLSv1_3));
     default: return false;
   }
 #else
@@ -205,6 +212,10 @@ bool HTTPClient::setMinimumSSLVersion(utils::SSLVersion minimum_version) {
         ssl_context_service_->setMinTlsVersion(TLS1_2_VERSION);
         break;
       }
+      case utils::SSLVersion::TLSv1_3: {
+        ssl_context_service_->setMinTlsVersion(TLS1_3_VERSION);
+        break;
+      }
       default: break;
     }
   }
@@ -220,6 +231,9 @@ bool HTTPClient::setMinimumSSLVersion(utils::SSLVersion minimum_version) {
       break;
     case utils::SSLVersion::TLSv1_2:
       ret = curl_easy_setopt(http_session_.get(), CURLOPT_SSLVERSION, CURL_SSLVERSION_TLSv1_2);
+      break;
+    case utils::SSLVersion::TLSv1_3:
+      ret = curl_easy_setopt(http_session_.get(), CURLOPT_SSLVERSION, CURL_SSLVERSION_TLSv1_3);
       break;
   }
 
--- a/extensions/standard-processors/tests/unit/PutTCPTests.cpp
+++ b/extensions/standard-processors/tests/unit/PutTCPTests.cpp
@@ -32,6 +32,7 @@
 #include "IntegrationTestUtils.h"
 
 using namespace std::literals::chrono_literals;
+using org::apache::nifi::minifi::utils::verifyLogLinePresenceInPollTime;
 
 namespace org::apache::nifi::minifi::processors {
 
@@ -64,12 +65,14 @@ class CancellableTcpServer : public utils::net::TcpServer {
         logger_->log_error("Error during accepting new connection: %s", accept_error.message());
         break;
       }
+      std::error_code error;
+      auto remote_address = socket.lowest_layer().remote_endpoint(error).address();
       auto cancellable_timer = std::make_shared<asio::steady_timer>(io_context_);
       cancellable_timers_.push_back(cancellable_timer);
       if (ssl_data_)
-        co_spawn(io_context_, secureSession(std::move(socket)) || wait_until_cancelled(cancellable_timer), asio::detached);
+        co_spawn(io_context_, secureSession(std::move(socket), std::move(remote_address), port_) || wait_until_cancelled(cancellable_timer), asio::detached);
       else
-        co_spawn(io_context_, insecureSession(std::move(socket)) || wait_until_cancelled(cancellable_timer), asio::detached);
+        co_spawn(io_context_, insecureSession(std::move(socket), std::move(remote_address), port_) || wait_until_cancelled(cancellable_timer), asio::detached);
     }
   }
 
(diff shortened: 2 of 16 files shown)


libminifi/src/controllers/SSLContextService.cpp
```diff
@@ -196,16 +196,16 @@ bool SSLContextService::configure_ssl_context(SSL_CTX *ctx) {
   }
 
   // Security level set to 0 for backwards compatibility to support TLS versions below v1.2
-  SSL_CTX_set_security_level(ctx, 0);
+  if (minimum_tls_version_ < TLS1_2_VERSION || maximum_tls_version_ < TLS1_2_VERSION) {
```
adamdebreceni, line 199: isn't it enough to check the `minimum_tls_version_`? I assume the value `-1` is "don't care", so if we do not care about the `maximum_tls_version_` event if the `minimum_tls_version_` disallows it we set the security level to 0

lordgamez: By default only TLS 1.2 or above is enabled in OpenSSL3. If we set the minimum version or the maximum below that, we should enable legacy mode to allow versions older than 1.2 to be used after version negotiation. But thinking about what you said, I think the `-1` value should also be checked, because if we "do not care" then we should better use the default values which is allow the TLS 1.2+ versions only. Legacy versions should only be used when specified. Updated in 234608e9d16a8e2c7d60b049fb2264bcaaf1c2fd

adamdebreceni: I'm still having wrapping my head around this, so for the negotiation to be able to chose < 1.2 we have to set the security level to 0, but we also have to either specify a minimum version < 1.2 or don't set the minimum version at all, so as I understand the < 1.2 can only NOT be negotiated if `minimum_tls_version_ >= 1.2` all other cases allow for < 1.2 but for that we need to also set the security level to 0, so the condition should only be `minimum_tls_version_ != -1 && minimum_tls_version_ < TLS1_2_VERSION`, so the user explicitly set the minimum version to allow pre-1.2, it seems to me that the maximum version does not really play a role

adamdebreceni: I see, so `-1` is not "don't care" but the default

adamdebreceni (approved):

szaszm:
Can we collect the affected parts, and which SSL/TLS versions are accepted in each case? I think we should aim to only accept TLS 1.2 and 1.3 in all cases, unless the user explicitly allows deprecated/insecure versions.

lordgamez:
> Can we collect the affected parts, and which SSL/TLS versions are accepted in each case? I think we should aim to only accept TLS 1.2 and 1.3 in all cases, unless the user explicitly allows deprecated/insecure versions.

This is the case currently. We only support TLS 1.2 and 1.3, only HTTPClient has the option to set a specific or a minimum/maxiumum version that would allow the use of a legacy TLS version.

libminifi/src/controllers/SSLContextService.cpp
```diff
@@ -196,16 +196,16 @@ bool SSLContextService::configure_ssl_context(SSL_CTX *ctx) {
   }
 
   // Security level set to 0 for backwards compatibility to support TLS versions below v1.2
-  SSL_CTX_set_security_level(ctx, 0);
+  if ((minimum_tls_version_ != 0 && minimum_tls_version_ < TLS1_2_VERSION) || (maximum_tls_version_ != -1 && maximum_tls_version_ < TLS1_2_VERSION)) {
```
fgerlits, line 199: should this be `if ((minimum_tls_version_ != -1 && ...`?

lordgamez: Yes, good catch, fixed in 8f8721b1712189699a0c4af09e0a06bd3fe763ee

fgerlits (approved):

libminifi/test/Utils.h
```diff
@@ -175,7 +175,7 @@ std::error_code sendMessagesViaSSL(const std::vector<std::string_view>& contents
     const asio::ip::tcp::endpoint& remote_endpoint,
     const std::filesystem::path& ca_cert_path,
     const std::optional<minifi::utils::net::SslData>& ssl_data = std::nullopt,
-    asio::ssl::context::method method = asio::ssl::context::tlsv12_client) {
+    asio::ssl::context::method method = asio::ssl::context::tls_client) {
```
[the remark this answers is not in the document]
lordgamez: As this is a test client it may or may not have the same SSL options as MiNiFi so I don't think we have to use those options.

extensions/http-curl/client/HTTPClient.cpp
```diff
@@ -163,6 +163,11 @@ bool HTTPClient::setSpecificSSLVersion(utils::SSLVersion specific_version) {
         ssl_context_service_->setMaxTlsVersion(TLS1_2_VERSION);
         break;
       }
+      case utils::SSLVersion::TLSv1_3: {
```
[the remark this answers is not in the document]
lordgamez: If no specific version is set then the SSL security level is set to 1 in the SSL context, which means that SSLv3, TLS 1.0, TLS 1.1 and DTLS 1.0 are all disabled in OpenSSL 3.1.

szaszm (approved):

extensions/standard-processors/tests/unit/PutTCPTests.cpp
```diff
@@ -64,12 +65,14 @@ class CancellableTcpServer : public utils::net::TcpServer {
         logger_->log_error("Error during accepting new connection: %s", accept_error.message());
         break;
       }
+      std::error_code error;
+      auto remote_address = socket.lowest_layer().remote_endpoint(error).address();
```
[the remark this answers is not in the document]
lordgamez: I think it's not necessary as the remote address is unused in the context of this test case.

extensions/standard-processors/tests/unit/PutTCPTests.cpp
```diff
@@ -400,9 +403,9 @@ TEST_CASE("PutTCP test missing client cert", "[PutTCP]") {
   auto port = test_fixture.addSSLServer();
   test_fixture.setPutTCPPort(port);
 
-  trigger_expect_failure(test_fixture, "message for invalid-cert server");
+  test_fixture.trigger("message for invalid-cert server");
```
martinzink, line 406: I know this is due to a change in TLS (you mentioned in the PR aswell)
`Fix tests to not expect error failed handshake in case of missing certificate as of TLS 1.3 handshake failure is only detected by the client on the first read operation: https://stackoverflow.com/questions/62459802/tls-1-3-client-does-not-report-failed-handshake-when-client-certificate-verifica`

But I think we should emphasize the meaning of this:
If someone has an invalid client certification on a minifi client which is not accepted by the server (the message will be disregarded), then using TLSV1.2 will route the messages to `failure` however using TLSv1.3 can (most often than not based on my testing) route these to `success` (sometimes `failure`)

This means feature loss on v1.3 compared to v1.2.
Do you see any way how we can resolve this?
If not then I think we should thoroughly document this phenomenon in PROCESSORS.md

martinzink (changes requested):

extensions/standard-processors/tests/unit/PutTCPTests.cpp
lordgamez: When I tested it, it seemed the routing depended on if the message could be sent before the server reset the connection. If it was reset before the message was sent it resulted in failure otherwise the result was success. Unfortunately in this processor we do not read through the connection, only write to the socket, so we cannot know if the handshake succeeded or not using TLS 1.3 other than the connection was reset or not while sending the message. We can document this, but I do not see any other option how this could be handled.

lordgamez: The failure relationship is documented like this at the moment: "FlowFiles that encountered IO errors are send out this relationship."

It does not say anything about the message arriving on the server or not, and I think no IO errors are reported if the handshake fails so it can match the current functionality, but I suppose it's better to add a note about this.

lordgamez: Added a note in 195fe306286c57fdb91e00c62c014172ccd6c8b4

martinzink: Yeah thats what I feared, without a more complex application layer we cant gurantee anything.

Still a note could be useful to the user, but I am not sure where can we put something like this.
The processor description seems a bit generic for a note like this, but anything other than that will be overwritten by the doc generation. 🤷 Maybe a code comment wil be sufficient?

martinzink (approved):

PR #1583, opened 2023-05-31
MINIFICPP-1719 Replace LibreSSL with OpenSSL 3.1

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-1719

Ticket to allow setup-nasm action: https://issues.apache.org/jira/browse/INFRA-24658

--------------------
Thank you for submitting a contribution to Apache NiFi - MiNiFi C++.

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [ ] Is there a JIRA ticket associated with this PR? Is it referenced
     in the commit message?

- [ ] Does your PR title start with MINIFICPP-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.

- [ ] Has your PR been rebased against the latest commit within the target branch (typically main)?

- [ ] Is your initial contribution a single, squashed commit?

### For code changes:
- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?
- [ ] If applicable, have you updated the LICENSE file?
- [ ] If applicable, have you updated the NOTICE file?

### For documentation related changes:
- [ ] Have you ensured that format looks appropriate for the output in which it is rendered?

### Note:
Please ensure that once the PR is submitted, you check GitHub Actions CI results for build issues and submit an update to your PR as soon as possible.

PR #1603, opened 2023-07-06
MINIFICPP-2098 Add ENABLE_ALL for all extensions

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2098

--------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/test-processors/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@
 # under the License.
 #
 
-if (SKIP_TESTS AND NOT ENABLE_TEST_PROCESSORS)
```
lordgamez, line 20: Test processors can be used in both unit and docker tests. As SKIP_TESTS is always disabled in docker builds, it should not be relevant in docker tests. In unit tests there should not be an "AND" but an "OR" logical connection between skipping the tests and enabling the test processors, but that would break the test processor build in the docker tests. Because of this we should just skip checking the SKIP_TESTS flag altogether.

extensions/test-processors/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@
...
+if (SKIP_TESTS AND NOT (ENABLE_ALL OR ENABLE_TEST_PROCESSORS))
```
adamdebreceni, line 20: do we want to add the test processors as well? if enabled, this extension is also added to the build artifact (the `register_extension` does not seem to discriminate), so either we should improve `register_extension` to not add this or hold off on enabling this with `ENABLE_ALL`

lordgamez: My thought was that as these processors are only built if the `SKIP_TESTS` is `OFF`, which means we are building the tests that need this extension anyway. If in this case we do not specifically set `ENABLE_TEST_PROCESSORS` because we have already set `ENABLE_ALL` to `ON` then we probably should build the test processors as they will be needed for building the tests.

adamdebreceni: we could tie it to just `SKIP_TESTS` as you mention this is needed anyway for the tests

lordgamez: The problem with that is that these processors are also used in the docker tests where the SKIP_TESTS is always ON.

adamdebreceni: building a package with `ENABLE_ALL` will now include the test processors which could be confusing

lordgamez: That could be confusing, removed in 24d0b4b9e848d2eb77ee7b4f3aa4e0b62130e45f

adamdebreceni (approved):

extensions/kubernetes/CMakeLists.txt
```diff
@@ -14,8 +14,7 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-
-if (NOT (CMAKE_SYSTEM_NAME STREQUAL "Linux" AND ENABLE_KUBERNETES))
+if (NOT (CMAKE_SYSTEM_NAME STREQUAL "Linux" AND (ENABLE_ALL OR ENABLE_KUBERNETES)))
```
fgerlits, line 17: You added a `NOT OPENSSL_OFF` condition to this in #1583; please rebase and resolve the conflict.

lordgamez: Thanks, updated in latest push.

martinzink (approved):

szaszm (approved):

PR #1612, opened 2023-07-20
MINIFICPP-2159 Add heterogeneous lookup to FlatMap

martinzink:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/libarchive/CompressContent.cpp
+++ b/extensions/libarchive/CompressContent.cpp
@@ -88,7 +88,7 @@ void CompressContent::processFlowFile(const std::shared_ptr<core::FlowFile>& flo
     std::string attr;
     flowFile->getAttribute(core::SpecialFlowAttribute::MIME_TYPE, attr);
     if (attr.empty()) {
-      logger_->log_error("No %s attribute existed for the flow, route to failure", core::SpecialFlowAttribute::MIME_TYPE);
+      logger_->log_error("No %s attribute existed for the flow, route to failure", core::SpecialFlowAttribute::MIME_TYPE.data());
       session->transfer(flowFile, Failure);
       return;
     }
--- a/extensions/libarchive/CompressContent.cpp
+++ b/extensions/libarchive/CompressContent.cpp
@@ -88,7 +88,7 @@ void CompressContent::processFlowFile(const std::shared_ptr<core::FlowFile>& flo
     std::string attr;
     flowFile->getAttribute(core::SpecialFlowAttribute::MIME_TYPE, attr);
     if (attr.empty()) {
-      logger_->log_error("No %s attribute existed for the flow, route to failure", core::SpecialFlowAttribute::MIME_TYPE.data());
+      logger_->log_error("No %s attribute existed for the flow, route to failure", std::string(core::SpecialFlowAttribute::MIME_TYPE));
       session->transfer(flowFile, Failure);
       return;
     }
--- a/libminifi/include/utils/FlatMap.h
+++ b/libminifi/include/utils/FlatMap.h
@@ -145,9 +145,7 @@ class FlatMap{
     return data_.rbegin()->second;
   }
 
-  template<typename T>
-  requires std::equality_comparable_with<K, T>
-  const V& at(const T& key) const {
+  const V& at(std::equality_comparable_with<K> auto const& key) const {
     auto it = find(key);
     if (it != end()) {
       return it->second;
@@ -155,9 +153,7 @@ class FlatMap{
     throw std::out_of_range("utils::FlatMap::at");
   }
 
-  template<typename T>
-  requires std::equality_comparable_with<K, T>
-  V& at(const T& key) {
+  V& at(std::equality_comparable_with<K> auto const& key) {
     auto it = find(key);
     if (it != end()) {
       return it->second;
@@ -172,9 +168,7 @@ class FlatMap{
     return iterator{data_.begin() + offset};
   }
 
-  template<typename T>
-  requires std::equality_comparable_with<K, T>
-  std::size_t erase(const T& key) {
+  std::size_t erase(std::equality_comparable_with<K> auto const& key) {
     for (auto it = data_.begin(); it != data_.end(); ++it) {
       if (it->first == key) {
         std::swap(*data_.rbegin(), *it);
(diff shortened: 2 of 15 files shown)


extensions/libarchive/CompressContent.cpp
```diff
@@ -88,7 +88,7 @@ void CompressContent::processFlowFile(const std::shared_ptr<core::FlowFile>& flo
     std::string attr;
     flowFile->getAttribute(core::SpecialFlowAttribute::MIME_TYPE, attr);
     if (attr.empty()) {
-      logger_->log_error("No %s attribute existed for the flow, route to failure", core::SpecialFlowAttribute::MIME_TYPE);
+      logger_->log_error("No %s attribute existed for the flow, route to failure", core::SpecialFlowAttribute::MIME_TYPE.data());
```
szaszm, line 91: This is unsafe, unless you can guarantee that after the end of the pointed-to string span, there is a null byte. I'm not sure what the standard requires for string_views initialized from plain old string literals. ("Works on my compiler" is not safe enough.)

libminifi/include/utils/FlatMap.h
```diff
@@ -138,7 +134,9 @@ class FlatMap{
     return data_.size();
   }
 
-  V& operator[](const K& key) {
+  template<typename T>
+  requires std::constructible_from<K, T> && std::equality_comparable_with<K, T>
```
szaszm, line 138: Why do we need `constuctible_from`? If we really need it, are we really interested in constructing K from an object of T?

libminifi/include/utils/FlatMap.h
szaszm, line 1: I would keep the overloads taking the key type, and only add the new functionality as new overloads.

libminifi/include/utils/FlatMap.h
```diff
@@ -282,20 +290,18 @@ class FlatMap{
...
 
-  bool empty() const noexcept {
+  [[nodiscard]] bool empty() const noexcept {
     return data_.empty();
   }
 
-  bool contains(const K& key) const {
+  template<typename T>
+  requires std::equality_comparable_with<K, T>
+  bool contains(const T& key) const {
```
szaszm, line 299:
```suggestion
  bool contains(std::equality_comparable_with<K> auto const& key) const {
```

libminifi/include/utils/FlatMap.h
martinzink: operator[] insert a default constructred value at the specified key if its not already there. for that we need to create the actual keytype thats why we need the constraint

libminifi/include/utils/FlatMap.h
martinzink: Wouldnt that just clutter the interface unnecessarily? the implementation hasnt changed we only relied on the operator== before this change aswell.

libminifi/include/utils/FlatMap.h
martinzink: good idea, changed it and the others (except for operator[]) in https://github.com/apache/nifi-minifi-cpp/pull/1612/commits/f2092883c378f21a8d7ec2ab53c3826bcbef07c2

extensions/libarchive/CompressContent.cpp
martinzink: you are right, good catch changed it in https://github.com/apache/nifi-minifi-cpp/pull/1612/commits/0b50c8b3c9a35f245d4a606d84f4fbd3634f54e5 
It would be good if our logger would accept string_view directly but thats for a new PR (maybe we could switch to fmt style aswell)

adamdebreceni: based on the standard it should be safe as `std::string_view::data` returns `s` if constructed from `std::string_view::string_view(const char* s)`, but agree that playing it safe could not hurt

libminifi/include/utils/FlatMap.h
szaszm: I don't know why I originally suggested this, I think I just prefer avoiding templates in the common case. One thing it could theoretically help with is allowing conversions to the key type, when the source type is not equality comparable to the key type, but convertible to it.

extensions/libarchive/CompressContent.cpp
szaszm: If it's safe, then I'm fine with the original as well, I just wasn't sure

adamdebreceni: I think your suggestion still makes sense, just wanted to check it if it is guaranteed or not

libminifi/include/utils/FlatMap.h
martinzink: it looks like MSVC doesnt like constraint auto, Im getting
```
C:\nifi-minifi-cpp\libminifi\test\unit\FlatMapTests.cpp(51): fatal error C1001: Internal compiler error.
(compiler file 'D:\a\_work\1\s\src\vctools\Compiler\CxxFE\sl\p1\c\trees.c', line 24533)
 To work around this problem, try simplifying or changing the program near the locations listed above.
If possible please provide a repro here: https://developercommunity.visualstudio.com/
```
I've found some reported issues with constraint auto and MSVC but those issues should be solved, ours is a bit more complex. I will try to create an MRE about this issue and file a bug with microsoft, in the meantime should I revert to explicit templates?

szaszm: I'd use the shortest for that's working with all targeted compilers. 
That translates to first fallback to this:
```c++
  template<std::equality_comparable_with<K> T>
  bool contains(T const& key) const {
```
and second fallback to the original version.

On an unrelated note, we should add `noexcept` wherever it makes sense. It may make our life easier down the line.

extensions/libarchive/CompressContent.cpp
martinzink: I'll go with the explicit std::string ctor version (thats safe for sure) and we can revisit this when we add support for string_view in our logger (or when we switch it to fmt).

libminifi/include/utils/FlatMap.h
martinzink: Good idea, the first fallback works on MSVC aswell. https://github.com/apache/nifi-minifi-cpp/pull/1612/commits/617fc38829b51a666b87ab11bd68ff9daea16523
Adding noexcept where applicable is a good idea, I didnt find any obvious target on a quick glance, and I didnt want to burden this PR with a more complex analysis. Should I file a new jira for this?

lordgamez (approved):

libminifi/include/utils/FlatMap.h
szaszm: For `noexcept`, I was thinking `operator[]`, `find`, `erase` and `contains`. It may need to be conditional, e.g. `erase` does `swap`, so it needs to be `std::is_nothrow_swappable_v`. I was hoping for `noexcept` `find`, but it would have to rely on a complex condition, so it may not be feasible after all.

If nothing obvious stands out, don't bother with the new Jira.

szaszm (approved):

adamdebreceni (approved):

PR #1615, opened 2023-07-26
Update Catch2 from 2.13.10 to 3.4.0

martinzink:

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/standard-processors/tests/unit/ExecuteProcessTests.cpp
```diff
@@ -170,6 +170,9 @@ TEST_CASE_METHOD(ExecuteProcessTestsFixture, "ExecuteProcess buffer long outputs
   CHECK(success_flow_files[0]->getAttribute("command") == command);
   CHECK(success_flow_files[0]->getAttribute("command.arguments") == arguments);
 }
-
+#else
+TEST_CASE("ExecuteProcessTests") {
+  SUCCEED("To avoid empty test executable");
+}
```
lordgamez, line 176: It may be better to remove the `#ifdef` and remove the test file in the `CMakeLists.txt` with `list(REMOVE_ITEM ...)` if we are building on Windows.

libminifi/test/unit/FileUtilsTests.cpp
```diff
@@ -186,9 +184,8 @@ TEST_CASE("TestFileUtils::addFilesMatchingExtension", "[TestAddFilesMatchingExte
 
 TEST_CASE("FileUtils::last_write_time and last_write_time_point work", "[last_write_time][last_write_time_point]") {
   using namespace std::chrono;
-  namespace fs = std::filesystem;
 
-  fs::file_time_type time_before_write = file_clock::now();
+  file_clock::time_point time_before_write = file_clock::now();
```
lordgamez, line 188: Just nitpicking, but it would be simpler to just use `auto` for these time points.

libminifi/test/unit/OsUtilTests.cpp
```diff
@@ -39,6 +39,10 @@ TEST_CASE("Test userIdToUsername for well-known SIDs", "[OsUtils]") {
   CHECK_FALSE(minifi::utils::OsUtils::userIdToUsername("S-1-3-4").empty());
   CHECK_FALSE(minifi::utils::OsUtils::userIdToUsername("S-1-5-80-0").empty());
 }
+#else
+TEST_CASE("OsUtilTests") {
```
lordgamez, line 43: We could also remove this from the file list in cmake instead.

libminifi/test/unit/WindowsCertStoreLocationTests.cpp
```diff
@@ -51,4 +51,10 @@ TEST_CASE("CurrentUser and LocalMachine are among the allowed values of WindowsC
   CHECK_FALSE(ranges::contains(allowed_locations, "SomebodyElsesComputer"));
 }
 
+#else
+#include "../Catch.h"
+TEST_CASE("WindowsCertStoreLocation") {
```
lordgamez, line 56: This could be removed as well.

libminifi/test/unit/OsUtilTests.cpp
martinzink: I didnt want to remove this from cmake because its reasonable that we would add non windows tests to this file (contrary to the others because those are inherently platform dependent). So instead I've added a test that checks if the machine arch is supported (this may become useful for more obscure platforms we might want to support in the future)

libminifi/test/unit/FileUtilsTests.cpp
martinzink: you are absolutely right, I've fixed this in https://github.com/apache/nifi-minifi-cpp/pull/1615/commits/44c7d79b3d130f483271f92ef7b8b224533e4137

extensions/standard-processors/tests/unit/ExecuteProcessTests.cpp
martinzink: Good idea, I've fixed it in https://github.com/apache/nifi-minifi-cpp/pull/1615/commits/44c7d79b3d130f483271f92ef7b8b224533e4137

libminifi/test/unit/WindowsCertStoreLocationTests.cpp
martinzink: I removed this in https://github.com/apache/nifi-minifi-cpp/pull/1615/commits/44c7d79b3d130f483271f92ef7b8b224533e4137

libminifi/test/unit/OsUtilTests.cpp
lordgamez: Makes sense, thanks.

lordgamez (approved):

szaszm (approved):

PR #1616, opened 2023-07-26
MINIFICPP-2173 Fix MergeContent error handling issues

lordgamez:
- Remove intermediate individual `ProcessSession` for every processed bin
- Rollback flow files of a merge if the merge fails with an exception and put them back to the ready bins
- Remove `merge_flow` from session in case the merge fails to avoid throwing an exception for not finding the transfer relationship
- Do not yield and continue processing of the bins in case putting one of the flow files in a bin fails
- Log stream size and offset values in case of a `StreamSlice` creation failure
- Remove unused function parameters and use references instead of raw pointers

https://issues.apache.org/jira/browse/MINIFICPP-2173

-------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/libarchive/BinFiles.cpp
+++ b/extensions/libarchive/BinFiles.cpp
@@ -80,7 +80,7 @@ void BinFiles::onSchedule(core::ProcessContext *context, core::ProcessSessionFac
   }
 }
 
-void BinFiles::preprocessFlowFile(core::ProcessContext* /*context*/, core::ProcessSession* /*session*/, const std::shared_ptr<core::FlowFile>& flow) {
+void BinFiles::preprocessFlowFile(const std::shared_ptr<core::FlowFile>& flow) {
   // handle backward compatibility with old segment attributes
   std::string value;
   if (!flow->getAttribute(BinFiles::FRAGMENT_COUNT_ATTRIBUTE, value) && flow->getAttribute(BinFiles::SEGMENT_COUNT_ATTRIBUTE, value)) {
@@ -208,14 +208,14 @@ bool BinManager::offer(const std::string &group, const std::shared_ptr<core::Flo
   return true;
 }
 
-bool BinFiles::resurrectFlowFiles(const std::shared_ptr<core::ProcessContext> &context, const std::shared_ptr<core::ProcessSession> &session) {
+bool BinFiles::resurrectFlowFiles(core::ProcessSession &session) {
   auto flow_files = file_store_.getNewFlowFiles();
   // these are already processed FlowFiles, that we own
   bool had_failure = false;
   for (auto &file : flow_files) {
-    std::string group_id = getGroupId(context.get(), file);
+    std::string group_id = getGroupId(file);
     if (!binManager_.offer(group_id, file)) {
-      session->transfer(file, Failure);
+      session.transfer(file, Failure);
       had_failure = true;
     }
     // no need to route successfully captured such files as we already own them in the Self relationship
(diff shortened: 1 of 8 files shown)


libminifi/src/io/StreamSlice.cpp
```diff
@@ -23,7 +23,8 @@ namespace org::apache::nifi::minifi::io {
 StreamSlice::StreamSlice(std::shared_ptr<io::InputStream> stream, size_t offset, size_t size) : stream_(std::move(stream)), slice_offset_(offset), slice_size_(size) {
   stream_->seek(slice_offset_);
   if (stream_->size() < slice_offset_ + slice_size_)
-    throw std::invalid_argument("StreamSlice is bigger than the Stream");
+    throw std::invalid_argument("StreamSlice is bigger than the Stream, Stream size: " + std::to_string(stream_->size()) +
+      ", StreamSlice size: " + std::to_string(slice_size_) + ", offset: " + std::to_string(slice_offset_));
```
martinzink, line 27: nitpicking but we could utilize fmt::format for a more compact and more readable version
```suggestion
    throw std::invalid_argument(fmt::format("StreamSlice is bigger than the Stream, Stream size: {}, StreamSlice size: {}, offset: {}", stream_->size(), slice_size_, slice_offset_));
```

extensions/libarchive/MergeContent.cpp
```diff
@@ -246,27 +247,30 @@ bool MergeContent::processBin(core::ProcessContext *context, core::ProcessSessio
     mimeType = "application/zip";
   } else {
     logger_->log_error("Merge format not supported %s", mergeFormat_);
+    session.remove(merge_flow);
     return false;
   }
 
   std::shared_ptr<core::FlowFile> mergeFlow;
```
martinzink, line 254: nitpicking and old code, but this seems unused
```suggestion
```

extensions/libarchive/MergeContent.cpp
```diff
@@ -213,18 +213,19 @@ bool MergeContent::processBin(core::ProcessContext *context, core::ProcessSessio
         });
   }
 
-  std::shared_ptr<core::FlowFile> merge_flow = std::static_pointer_cast<FlowFileRecord>(session->create());
+  std::shared_ptr<core::FlowFile> merge_flow = std::static_pointer_cast<FlowFileRecord>(session.create());
```
martinzink, line 216: I think it might be better to introduce a function to processSession that checks if the flowfile is in a valid transferred state, and then let a RAII take care of the removal instead. This way we dont have to keep in mind that we have to remove it during the various early return paths.
something like this should do it
```
  auto remove_merge_flow = gsl::finally([&](){
    if (!session.hasBeenTransferred(*merge_flow))
      session.remove(merge_flow);
  });
```
whre the hasbeentransferred should check if the updated_relationships_ or added_flowfiles_ maps contain a nonnull relationship* for the flowfile

libminifi/src/io/StreamSlice.cpp
lordgamez: Updated in 8a269048b9b6f451f0baedd584a4358286908a60

extensions/libarchive/MergeContent.cpp
lordgamez: Removed in 8a269048b9b6f451f0baedd584a4358286908a60

extensions/libarchive/MergeContent.cpp
lordgamez: Good idea, updated in 8a269048b9b6f451f0baedd584a4358286908a60

libminifi/src/core/ProcessSession.cpp
```diff
@@ -1147,4 +1147,8 @@ bool ProcessSession::existsFlowFileInRelationship(const Relationship &relationsh
   });
 }
 
+bool ProcessSession::hasBeenTransferred(const core::FlowFile &flow) const {
+  return updated_relationships_.contains(flow.getUUID()) || added_flowfiles_.contains(flow.getUUID());
```
martinzink, line 1151: Shouldnt we also check if the relationship at the flow uuid is not nullptr?

martinzink (approved):

extensions/libarchive/BinFiles.h
```diff
@@ -284,19 +285,24 @@ class BinFiles : public core::Processor {
...
-  static void transferFlowsToFail(core::ProcessContext *context, core::ProcessSession *session, std::unique_ptr<Bin> &bin);
+  static void transferFlowsToFail(core::ProcessSession &session, std::unique_ptr<Bin> &bin);
   // moves owned flows to session
-  static void addFlowsToSession(core::ProcessContext *context, core::ProcessSession *session, std::unique_ptr<Bin> &bin);
+  static void addFlowsToSession(core::ProcessSession &session, std::unique_ptr<Bin> &bin);
+
+  bool resurrectFlowFiles(core::ProcessSession &session);
+  void assumeOwnerShipOfNextBatch(core::ProcessSession &session);
+  std::deque<std::unique_ptr<Bin>> gatherReadyBins(core::ProcessContext &context);
+  void processReadyBins(std::deque<std::unique_ptr<Bin>> ready_bins, core::ProcessSession &session);
```
szaszm, line 305: Could you write a comment explaining the inner working of BinFiles and MergeContent, as a class or file level comment? I guess you had to understand it to some degree to make this change, and I would appreciate sharing this knowledge.

If that's too big of an ask, could you document just the concepts related to this patch?

Some words worth explaining IMO:
- resurrect (flow file)
- batch
- bin
- ready bin
- preprocess (flow file)
- group
- bundle
- entry

Or just the ones related to this PR:
- resurrect (flow file)
- batch (how is it different from bin, group, bundle, etc.)
- ready bins

This would also help with reviewing.

lordgamez:
Some overview of the inner workings of MergeContent:

- Resurrect flow file is the process of restoring the unprocessed flow files from the flow file repository after restart. This is done for each processor and in case of the MergeContent this also means restoring flow files previously stored in the internal Self relationship. After restoring these flow files they also have to be sorted into their respective bins to restore the same bin state as before the restart.
- Batches and batch size represent the maximum number of flow files to be processed from the incoming relationship.
- A bin represents a set of flow files that belong together defined by the processor properties. In MergeContent correlated flow files are defined by the `CorrelationAttributeName` property which defines the attribute that provides the `groupid` for the bin the flow file belongs to.
- When a bin reaches a limit defined by the maximum age or the maximum size, the bin becomes ready, and ready bins can be merged.
- Preprocess flow file only checks for deprecated attributes and sets the value of the deprecated attribute to the value of the new renamed attribute.
- Groups are set of bins with the same groupid. It can happen that a bin cannot accept a new flow files because it would go above its size limit for example, so a new bin is created with this new flow file and added to the same group of bins.
- I think bundle is just a synonym for bin in the user descriptions. It is more understandable from a user's perspective than bin.
- Entry is a member flow file of a bin

adamdebreceni (approved):

extensions/libarchive/BinFiles.h
szaszm: I could have been more specific, but I meant a code comment, for future reference. A `/**` or `///` style class comment above MergeContent could house this information.

lordgamez: Sorry, I misunderstood, I thought the comments were only needed for review purposes. I added and cleaned up some comments in fc0467b393bfec1f6a23dbf78aa1ced9ec96f684

extensions/libarchive/BinFiles.cpp
```diff
@@ -203,93 +208,97 @@ bool BinManager::offer(const std::string &group, const std::shared_ptr<core::Flo
...
+    if (!binManager_.offer(group_id, file)) {
+      session.transfer(file, Failure);
+      had_failure = true;
     }
+    // no need to route successfully captured such files as we already own them in the Self relationship
   }
+  return had_failure;
+}
 
+void BinFiles::assumeOwnerShipOfNextBatch(core::ProcessSession &session) {
```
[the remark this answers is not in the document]
lordgamez: Updated in 124c1034ff66b2b2a839165a272c760e89b9ab5f

szaszm (approved):

PR #1623, opened 2023-08-02
MINIFICPP-2116 - Support GCC13

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/sftp/processors/ListSFTP.cpp
```diff
@@ -535,9 +535,9 @@ void ListSFTP::listByTrackingTimestamps(
       /* Determine the minimum reliable timestamp based on precision */
       auto minimum_reliable_timestamp = now - listing_lag;
       if (remote_system_timestamp_precision == TARGET_SYSTEM_TIMESTAMP_PRECISION_SECONDS) {
-        std::chrono::floor<std::chrono::seconds>(minimum_reliable_timestamp);
+        minimum_reliable_timestamp = std::chrono::floor<std::chrono::seconds>(minimum_reliable_timestamp);
```
martinzink, line 538: This seems like an actual bugfix :+1: , should we file a Jira for this? (and maybe even some tests?)

adamdebreceni: create a jira for it: https://issues.apache.org/jira/browse/MINIFICPP-2176

.github/workflows/gcc13-compat.yml
```diff
@@ -0,0 +1,54 @@
...
+          echo -e "127.0.0.1\t$HOSTNAME" | sudo tee -a /etc/hosts > /dev/null
+      - name: build
+        run: |
+          eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"
+          brew install gcc@13
+          export CC=gcc-13
+          export CXX=g++-13
+          ./bootstrap.sh -e -t
+          cd build
+          cmake -DUSE_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release -DCI_BUILD=ON -DSTRICT_GSL_CHECKS=AUDIT -DFAIL_ON_WARNINGS=ON -DENABLE_AWS=ON -DENABLE_AZURE=ON -DENABLE_BUSTACHE=ON -DENABLE_COAP=ON \
```
martinzink, line 30: I reckon we wont run this too frequently and #1603 is merged, we could replace this long list with a simple ENABLE_ALL ?

.github/workflows/gcc13-compat.yml
```diff
@@ -0,0 +1,46 @@
...
+      - id: install_deps
+        run: |
+          sudo apt update
+          sudo apt install -y ccache libfl-dev libpcap-dev libboost-all-dev openjdk-8-jdk maven libusb-1.0-0-dev libpng-dev libgps-dev libsqliteodbc lua5.3 liblua5.3-dev flake8 parallel
+          echo "PATH=/usr/lib/ccache:$PATH" >> $GITHUB_ENV
+          echo -e "127.0.0.1\t$HOSTNAME" | sudo tee -a /etc/hosts > /dev/null
+      - name: build
+        run: |
+          eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"
+          brew install gcc@13
```
szaszm, line 20: You can use the Toolchain test PPA to install GCC 13: https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test

cmake/GoogleCloudCpp.cmake
```diff
@@ -66,3 +66,5 @@ if (WIN32)
 else()
     target_compile_options(google_cloud_cpp_storage PUBLIC -Wno-error=deprecated-declarations)
 endif()
+
+target_compile_options(google_cloud_cpp_rest_internal PUBLIC -Wno-error)
```
szaszm, line 70: I think this should go in a similar platform check as above, since `-Wno-error` is not a valid MSVC flag.

thirdparty/libwebsockets/enum-int.patch
szaszm, line 1: Can we patch libwebsockets to use our own OpenSSL build?

adamdebreceni: I think it is using our openssl https://github.com/apache/nifi-minifi-cpp/blob/a266a30475b76d3ceeecaa38f4711f331d0a3901/cmake/KubernetesClientC.cmake#L34

cmake/GoogleCloudCpp.cmake
adamdebreceni: moved

fgerlits (approved):

thirdparty/azure-sdk-cpp/gcc13.patch
lordgamez, line 1: Could we upstream this to the azure sdk as a community contribution?

thirdparty/libwebsockets/enum-int.patch
lordgamez, line 1: Is this issue gcc13 specific or some other warning due to the type difference? If it's a gcc13 specific error we could include it in the gcc13 patch if it's a warning can we turn off the werror in the 3rd party lib?

szaszm: Why did it need to be patched then? If it's unused, it probably shouldn't be built.

thirdparty/azure-sdk-cpp/gcc13.patch
adamdebreceni: opened a PR https://github.com/Azure/azure-sdk-for-cpp/pull/4857

thirdparty/libwebsockets/enum-int.patch
adamdebreceni: removed patch and disabled werror

adamdebreceni: I don't think this is an openssl implementation simply the usage side, but @lordgamez might be able to shed some more light on it (as the author of `libwebsockets/openssl3.patch`)

lordgamez: The libwebsockets is using our openssl build. The goal of the `libwebsockets/openssl3.patch` is to not use the `CHECK_FUNCTION_EXISTS` functions for checking the availability of specific openssl functions as they try to find them on the system. As we want to use our own lib instead of the system provided one, we simply set those openssl functions to be available because they are provided by our openssl build.

I thought the reason for the `thirdparty/libwebsockets/enum-int.patch` was to avoid a warning only occurring on GCC13 that's why I suggested disabling `Werror` on this third-party library.

szaszm (approved):

lordgamez (approved):

szaszm:
After trying to merge this, I'm still getting the error:
```
/home/szaszm/nifi-minifi-cpp-3/build-gcc-make/_deps/websockets-src/lib/tls/openssl/openssl-server.c:713:1: error: conflicting types for ‘lws_tls_server_abort_connection’ due to enum/integer mismatch; have ‘int(struct lws *)’ [-Werror=enum-int-mismatch]
  713 | lws_tls_server_abort_connection(struct lws *wsi)
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /home/szaszm/nifi-minifi-cpp-3/build-gcc-make/_deps/websockets-src/lib/tls/./private-lib-tls.h:143,
                 from /home/szaszm/nifi-minifi-cpp-3/build-gcc-make/_deps/websockets-src/lib/core/./private-lib-core.h:265,
                 from /home/szaszm/nifi-minifi-cpp-3/build-gcc-make/_deps/websockets-src/lib/tls/openssl/openssl-server.c:25:
/home/szaszm/nifi-minifi-cpp-3/build-gcc-make/_deps/websockets-src/lib/tls/./private-network.h:171:1: note: previous declaration of ‘lws_tls_server_abort_connection’ with type ‘enum lws_ssl_capable_status(struct lws *)’
  171 | lws_tls_server_abort_connection(struct lws *wsi);
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
```

szaszm (changes requested):
the above issue

adamdebreceni:
changed it to a cached variable (similar to the variables above this) which fixes the issue

cmake/BundledAzureSdkCpp.cmake
```diff
@@ -55,14 +59,17 @@ function(use_bundled_libazure SOURCE_DIR BINARY_DIR)
 
     set(AZURE_SDK_CMAKE_ARGS ${PASSTHROUGH_CMAKE_ARGS}
         -DWARNINGS_AS_ERRORS=OFF
-        -DCMAKE_INSTALL_PREFIX=${INSTALL_DIR})
+        -DCMAKE_INSTALL_PREFIX=${INSTALL_DIR}
+        -DDISABLE_AZURE_CORE_OPENTELEMETRY=ON
+        -DBUILD_TRANSPORT_CURL=ON
+        -DCMAKE_CXX_FLAGS=-DCURL_STATICLIB)
```
szaszm, line 65: Would it work if we used `target_compile_definitions` instead, or does that not get passed down? Also, do we not need this definition when including Azure SDK headers?

adamdebreceni: moved this directive to the curl target

szaszm (approved):

martinzink (approved):

PR #1624, opened 2023-08-02
MINIFICPP-2175 Increase CMAKE requirements to 3.24

martinzink:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -332,7 +332,7 @@ jobs:
           path: /tmp/bin
   docker_build:
     name: "Docker build for integration tests"
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     timeout-minutes: 180
     steps:
       - id: checkout
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -16,7 +16,7 @@
 # under the License.
 #
 
-ARG BASE_IMAGE="alpine:3.16"
+ARG BASE_IMAGE="alpine:3.17"
 
 # Build image
 FROM ${BASE_IMAGE} AS build
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -16,7 +16,7 @@
 # under the License.
 #
 
-ARG BASE_IMAGE="alpine:3.17"
+ARG BASE_IMAGE="alpine:3.18"
 
 # Build image
 FROM ${BASE_IMAGE} AS build
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -16,7 +16,7 @@
 # under the License.
 #
 
-ARG BASE_IMAGE="alpine:3.18"
+ARG BASE_IMAGE="alpine:3.17"
 
 # Build image
 FROM ${BASE_IMAGE} AS build
--- a/linux.sh
+++ b/linux.sh
@@ -19,3 +19,30 @@ verify_gcc_enable(){
   #feature="$1"
   [ "$COMPILER_MAJOR" -ge 11 ] && echo true || echo false
 }
+
+install_cmake_from_binary() {
+  CMAKE_VERSION="3.24.4"
+  CMAKE_URL="https://cmake.org/files/v3.24/cmake-3.24.4-linux-x86_64.tar.gz"
+  EXPECTED_SHA256="cac77d28fb8668c179ac02c283b058aeb846fe2133a57d40b503711281ed9f19"
+
+  TMP_DIR=$(mktemp -d)
+
+  install_pkgs wget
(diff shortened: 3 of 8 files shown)


docker/Dockerfile
```diff
@@ -16,7 +16,7 @@
 # under the License.
 #
 
-ARG BASE_IMAGE="alpine:3.16"
+ARG BASE_IMAGE="alpine:3.17"
```
lordgamez, line 19: I think we could use 3.18 if we are to upgrade

.github/workflows/ci.yml
```diff
@@ -334,7 +334,7 @@ jobs:
           path: /tmp/bin
   docker_build:
     name: "Docker build for integration tests"
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
lordgamez, line 337: Isn't it better to test on older platforms if it's still supported? As I see bootstrap will take care of the cmake requirements anyway.

martinzink: This is only the platform that calls the docker build (the build itself is in alpine), and due to being only docker build we skip the bootstrapping process so its easier to build it on a platform that supports cmake 24 out of the box.

adamdebreceni:
what requires 3.24?

linux.sh
```diff
@@ -19,3 +19,30 @@ verify_gcc_enable(){
...
+  CMAKE_VERSION="3.24.4"
+  CMAKE_URL="https://cmake.org/files/v3.24/cmake-3.24.4-linux-x86_64.tar.gz"
+  EXPECTED_SHA256="cac77d28fb8668c179ac02c283b058aeb846fe2133a57d40b503711281ed9f19"
+
+  TMP_DIR=$(mktemp -d)
+
+  install_pkgs wget
+  wget -P "$TMP_DIR" "$CMAKE_URL"
+
+  ACTUAL_SHA256=$(sha256sum "$TMP_DIR/cmake-$CMAKE_VERSION-linux-x86_64.tar.gz" | awk '{print $1}')
```
szaszm, line 33: A small detail, but `cut` is much simler than `awk`, so it may be better to default to that.
```suggestion
  ACTUAL_SHA256=$(sha256sum "$TMP_DIR/cmake-$CMAKE_VERSION-linux-x86_64.tar.gz" | cut -d " " -f 2)
```

linux.sh
```diff
@@ -19,3 +19,30 @@ verify_gcc_enable(){
...
+  if [ "$ACTUAL_SHA256" != "$EXPECTED_SHA256" ]; then
+    echo "ERROR: SHA-256 verification failed. Aborting."
+    rm -r "$TMP_DIR"
+    exit 1
+  fi
+
+  tar -C "$TMP_DIR" -zxvf "$TMP_DIR/cmake-$CMAKE_VERSION-linux-x86_64.tar.gz"
+  sudo mv "$TMP_DIR/cmake-$CMAKE_VERSION-linux-x86_64" /opt/cmake-$CMAKE_VERSION
+
+  sudo ln -s "/opt/cmake-$CMAKE_VERSION/bin/cmake" /usr/local/bin/cmake
```
szaszm, line 44: We should print the details of this installation procedure, so the user at least knows what we're doing to their system. The best would be a message and a `set -x` before these commands, so they know what they're giving sudo password for. I'd also remove the verbose flag from tar, to avoid spamming the console, and keep the important parts visible.
```suggestion
  echo "Installing CMake $CMAKE_VERSION to /opt/cmake-$CMAKE_VERSION..."
  set -x
  tar -C "$TMP_DIR" -zxf "$TMP_DIR/cmake-$CMAKE_VERSION-linux-x86_64.tar.gz"
  sudo mv "$TMP_DIR/cmake-$CMAKE_VERSION-linux-x86_64" /opt/cmake-$CMAKE_VERSION

  sudo ln -s "/opt/cmake-$CMAKE_VERSION/bin/cmake" /usr/local/bin/cmake
  set +x
```

docker/Dockerfile
martinzink: good idea, https://github.com/apache/nifi-minifi-cpp/pull/1624/commits/c32c9828925057c9967555612a4cac92b3b81e80

szaszm (approved):

linux.sh
szaszm: sorry for the bad snippet: I didn't check the output of `sha256sum`, just wrote it by memory, which was wrong

docker/Dockerfile
martinzink: This caused a lot of tests to fail, so I've reverted it. Will investigate later, but probably in a separate PR

lordgamez (approved):

PR #1631, opened 2023-08-17
MINIFICPP-2174 Send all cached compressed log files through C2

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2174

-------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/core/logging/LoggerConfiguration.h
+++ b/libminifi/include/core/logging/LoggerConfiguration.h
@@ -98,15 +98,15 @@ class LoggerConfiguration {
    */
   void initialize(const std::shared_ptr<LoggerProperties> &logger_properties);
 
-  static std::unique_ptr<io::InputStream> getCompressedLog(bool flush = false) {
-    return getCompressedLog(std::chrono::milliseconds{0}, flush);
+  static std::vector<std::unique_ptr<io::InputStream>> getCompressedLogs(bool flush = false) {
+    return getCompressedLogs(std::chrono::milliseconds{0}, flush);
   }
 
   void initializeAlertSinks(core::controller::ControllerServiceProvider* controller, const std::shared_ptr<AgentIdentificationProvider>& agent_id);
 
   template<class Rep, class Period>
-  static std::unique_ptr<io::InputStream> getCompressedLog(const std::chrono::duration<Rep, Period>& time, bool flush = false) {
-    return getConfiguration().compression_manager_.getCompressedLog(time, flush);
+  static std::vector<std::unique_ptr<io::InputStream>> getCompressedLogs(const std::chrono::duration<Rep, Period>& time, bool flush = false) {
+    return getConfiguration().compression_manager_.getCompressedLogs(time, flush);
   }
 
   /**
--- a/libminifi/include/core/logging/LoggerConfiguration.h
+++ b/libminifi/include/core/logging/LoggerConfiguration.h
@@ -98,15 +98,15 @@ class LoggerConfiguration {
    */
   void initialize(const std::shared_ptr<LoggerProperties> &logger_properties);
 
-  static std::vector<std::unique_ptr<io::InputStream>> getCompressedLogs(bool flush = false) {
-    return getCompressedLogs(std::chrono::milliseconds{0}, flush);
+  static std::vector<std::unique_ptr<io::InputStream>> getCompressedLogs() {
+    return getCompressedLogs(std::chrono::milliseconds{0});
   }
 
   void initializeAlertSinks(core::controller::ControllerServiceProvider* controller, const std::shared_ptr<AgentIdentificationProvider>& agent_id);
 
   template<class Rep, class Period>
-  static std::vector<std::unique_ptr<io::InputStream>> getCompressedLogs(const std::chrono::duration<Rep, Period>& time, bool flush = false) {
-    return getConfiguration().compression_manager_.getCompressedLogs(time, flush);
+  static std::vector<std::unique_ptr<io::InputStream>> getCompressedLogs(const std::chrono::duration<Rep, Period>& time) {
+    return getConfiguration().compression_manager_.getCompressedLogs(time);
   }
 
   /**
--- a/libminifi/include/core/logging/internal/LogCompressorSink.h
+++ b/libminifi/include/core/logging/internal/LogCompressorSink.h
@@ -21,6 +21,7 @@
 #include <memory>
 #include <atomic>
 #include <utility>
+#include <vector>
 
 #include "spdlog/common.h"
 #include "spdlog/details/log_msg.h"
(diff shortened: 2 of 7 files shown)


libminifi/include/core/logging/LoggerConfiguration.h
```diff
@@ -98,15 +98,15 @@ class LoggerConfiguration {
    */
   void initialize(const std::shared_ptr<LoggerProperties> &logger_properties);
 
-  static std::unique_ptr<io::InputStream> getCompressedLog(bool flush = false) {
-    return getCompressedLog(std::chrono::milliseconds{0}, flush);
+  static std::vector<std::unique_ptr<io::InputStream>> getCompressedLogs(bool flush = false) {
+    return getCompressedLogs(std::chrono::milliseconds{0}, flush);
```
szaszm, line 102: I think it would be better to create a new stream class that delegates to a list of inner streams. What do you think?

lordgamez: I think it's a good idea, unfortunately I tried doing it and it cannot be handled as a single stream because of the compression. As the segments are compressed individually when forwarding the input stream to the zlib decompressor it only returns the first segment decompressed, probably due to finding the end of the archive.

szaszm: This new stream class that handles delegating to an inner list of streams could handle reading from the next stream when the first one is over, and hide the complexity inside the new class. Zlib could be involved in a lower level, like this:
ConcatStream contains std::vector<InputStream>, where each stream is a ZlibDecompressStream
Each ZlibDecompressStream contains the original BufferStream (or whatever) with the compressed data.

libminifi/src/FlowController.cpp
```diff
@@ -473,8 +473,10 @@ std::vector<BackTrace> FlowController::getTraces() {
 
 std::map<std::string, std::unique_ptr<io::InputStream>> FlowController::getDebugInfo() {
   std::map<std::string, std::unique_ptr<io::InputStream>> debug_info;
-  if (auto logs = core::logging::LoggerConfiguration::getCompressedLog(true)) {
-    debug_info["minifi.log.gz"] = std::move(logs);
+  auto logs = core::logging::LoggerConfiguration::getCompressedLogs(true);
+  for (size_t i = 0; i < logs.size(); ++i) {
+    std::string index_str = i == logs.size() - 1 ? "" : "." + std::to_string(logs.size() - 1 - i);
+    debug_info["minifi.log" + index_str + ".gz"] = std::move(logs[i]);
```
szaszm, line 479: The logs can be decompressed inside the debug archive. The tarball itself is compressed anyway.

libminifi/include/core/logging/LoggerConfiguration.h
lordgamez: Yes, that was the approach I tried in a previous commit, but the logger test with multiple segments failed with this implementation. https://github.com/lordgamez/nifi-minifi-cpp/commit/2f3d2c0800f5b9fc29e3f3fde64dbcd8d9ede283

I tried modifying the `ZlibDecompressStream` class to be able to handle this kind of `InputStreamBundle` but evenn if I try to adjust the zlib libraries `z_stream` input and output pointers the decompression after the first segment still fails to read the proper data: https://github.com/lordgamez/nifi-minifi-cpp/commit/7ac83b33051fe672f984d0885c198b3baa07e1d6

libminifi/src/FlowController.cpp
lordgamez: I'm not sure if it's worth the performance impact the decompression may add to the agent. We introduced the compression because there may be a flood of repeated logs that may take up a lot of memory and this way we could highly reduce the stored content. If we decompress the logs here it can inflate exponentially which may cause a momentary resource usage surge, which can be unnecessary as we will compress this bundle again in the next step.

libminifi/include/core/logging/internal/LogCompressorSink.h
```diff
@@ -57,26 +58,44 @@ class LogCompressorSink : public spdlog::sinks::base_sink<std::mutex> {
...
     }
-    return std::move(compressed.buffer_);
+    return log_segments;
   }
 
   size_t getMaxCacheSize() const {
     return cached_logs_.getMaxSize();
   }
 
+  size_t getMaxCacheSegmenSize() const {
```
adamdebreceni, line 87: typo `getMaxCacheSegmentSize`

adamdebreceni (approved):

libminifi/include/core/logging/internal/LogCompressorSink.h
lordgamez: Fixed in ea9710346c044d7cd624e2d38bdda5f731e0057a

libminifi/include/core/logging/LoggerConfiguration.h
szaszm: I was hoping for some code simplification, but since it's not trivial, I don't want to block this PR with it.

szaszm (approved):

libminifi/include/core/logging/internal/LogCompressorSink.h
```diff
@@ -57,26 +58,44 @@ class LogCompressorSink : public spdlog::sinks::base_sink<std::mutex> {
...
-    LogBuffer compressed;
-    if (!compressed_logs_.tryDequeue(compressed, time) && flush) {
-      return createEmptyArchive();
+
+    std::vector<std::unique_ptr<io::InputStream>> log_segments;
+    const auto segment_count = compressed_logs_.itemCount();
+    for (size_t i = 0; i < segment_count; ++i) {
+      LogBuffer compressed;
+      if (!compressed_logs_.tryDequeue(compressed, time) && flush) {
+        break;
```
fgerlits, line 72: I don't understand the purpose of checking `flush` here (I didn't understand it before this change either, but it got more complex).  As far as I can see, we only call this with `flush = true`.

Can we remove the `flush` argument?  If not, then I think the logic needs to be made clearer, either by a comment or some other way.

libminifi/test/unit/LoggerTests.cpp
```diff
@@ -276,7 +280,7 @@ TEST_CASE("Setting either properties to 0 disables in-memory compressed logs", "
   log_config.initialize(properties);
   auto logger = log_config.getLogger("DisableCompressionTestLogger");
   logger->log_error("Hi there");
-  REQUIRE((logging::LoggerConfiguration::getCompressedLog(true) == nullptr) == is_nullptr);
+  REQUIRE(logging::LoggerConfiguration::getCompressedLogs(true).empty() == is_nullptr);
```
fgerlits, line 283: it would be better to rename `is_nullptr` to `is_empty`

libminifi/include/core/logging/internal/LogCompressorSink.h
lordgamez: You are right, there is no reason to keep that argument at the moment, removed in 84abc4df7100b28695cb75a753bb9df3f64a0620

libminifi/test/unit/LoggerTests.cpp
lordgamez: Updated in d32b92c94c133b8baf0d81be52bb780da799f05e

fgerlits (approved):

PR #1638, opened 2023-08-23
MINIFICPP-1774 Set properties from command line arguments

lordgamez:
- Remove ArgParser class and use thirdparty argparse library instead
- Add -p/--property option to override configuration properties

https://issues.apache.org/jira/browse/MINIFICPP-1774

------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


minifi_main/MiNiFiMain.cpp
szaszm, line 1: Could we add a `--help` option that just prints help and exists with 0?

lordgamez: The `-h/--help` option is provided by the `argparse` library by default.

szaszm: How does it exit? Does it call `exit(0)` internally? (That seems to be a bit of an overreach.)

lordgamez: It exits with exit code 0, seems to work the same way as python's argparse, here's the more information: https://github.com/p-ranav/argparse#default-arguments

lordgamez: It does call `exit(0)` with the default `--help` implementation here: https://github.com/p-ranav/argparse/blob/b0930ab0288185815d6dc67af59de7014a6272f7/include/argparse/argparse.hpp#L1094C13-L1094C13

LICENSE
```diff
@@ -3416,3 +3416,13 @@ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
 LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
+
+This product bundles 'argparse' which is available under The MIT License.
```
adamdebreceni, line 3420: we are already using `cxxopts` library for the minificontroller, does `argparse` provide some extra features?

lordgamez: One feature I cannot see in `cxxopts` that we used here with the `--docs` options is to provide variable number of parameters for an option. In `argparse` I could define `nargs` with 1 or 2 options depending on if the user wants to specify a second parameter for the PROCESSORS.md output. We also want to reuse the `--property` multiple times which looks to be a feature of `cxxopts` 3.0, but currently we are using 2.2. It seems that `cxxopts` is more lightweight, but with less features than `argparse` which looks to be in parity with the features of the python's argparse lib. I only saw the `ArgumentParser` class used in `EncryptConfig` so I replaced that, but I suppose it would be better to have a single argument parser in the overall codebase. What do you think, should we replace `cxxopts` in minificontroller too or can we use `cxxopts` with our options somehow?

minifi_main/MiNiFiMain.cpp
szaszm: I still find it strange, but it works fine in our case, so I'm closing this thread.

minifi_main/MiNiFiMain.cpp
```diff
@@ -134,7 +135,93 @@ void writeJsonSchema(const std::shared_ptr<minifi::Configure> &configuration, st
...
+}
+
+void dumpDocsIfRequested(const argparse::ArgumentParser& parser, const std::shared_ptr<minifi::Configure>& configure) {
+  if (!parser.is_used("--docs")) {
+    return;
+  }
+  const auto& docs_params = parser.get<std::vector<std::string>>("--docs");
+  if (utils::file::create_dir(docs_params[0]) != 0) {
+    std::cerr << "Working directory doesn't exist and cannot be created: " << docs_params[0] << std::endl;
+    exit(1);
```
szaszm, line 158: Please keep the `exit` vs `std::exit` calls consistent in the same file

minifi_main/MiNiFiMain.cpp
```diff
@@ -134,7 +135,93 @@ void writeJsonSchema(const std::shared_ptr<minifi::Configure> &configuration, st
...
+
+  auto parent_dir = std::filesystem::path(schema_path).parent_path();
+  if (utils::file::create_dir(parent_dir) != 0) {
+    std::cerr << "JSON schema parent directory doesn't exist and cannot be created: " << parent_dir << std::endl;
+    exit(1);
+  }
+  std::cout << "Writing json schema to " << schema_path << std::endl;
+  std::ofstream schema_file{schema_path};
+  writeJsonSchema(configure, schema_file);
+  std::exit(0);
```
szaszm, line 199: `exit` doesn't unwind the stack, so local variable destructors are not called. I'd prefer either normal return from main, or at least closing all files and other resources before calling exit.
```suggestion
  {
    std::ofstream schema_file{schema_path};
    writeJsonSchema(configure, schema_file);
  }
  std::exit(0);
```

minifi_main/MiNiFiMain.cpp
lordgamez: Updated in 8e64616285aa944ae5e35bece9c07b025f737577

minifi_main/MiNiFiMain.cpp
lordgamez: Updated in 8e64616285aa944ae5e35bece9c07b025f737577

adamdebreceni (approved):

LICENSE
```diff
@@ -1250,60 +1250,6 @@ Redistribution and use in source and binary forms, with or without modification,
...
-IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-THE SOFTWARE.
-
-This product bundles zlib which is available under the zlib license:
-
- (C) 1995-2017 Jean-loup Gailly and Mark Adler
```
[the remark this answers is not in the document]
lordgamez: Good catch, updated in fedcf987b23f949b053a4cced79ef7b9fd2478e6

controller/MiNiFiController.cpp
```diff
@@ -106,81 +109,108 @@ int main(int argc, char **argv) {
...
+  argument_parser.add_argument("--start")
+    .metavar("COMPONENT")
+    .nargs(argparse::nargs_pattern::at_least_one)
+    .help("Starts provided components");
+  argument_parser.add_argument("-l", "--list")
+    .action([](const std::string& value) {
+      static const std::vector<std::string> choices = {"components", "connections"};
+      if (ranges::contains(choices, value)) {
+        return value;
+      }
```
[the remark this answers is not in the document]
lordgamez: Updated in fedcf987b23f949b053a4cced79ef7b9fd2478e6

fgerlits (approved):

szaszm (approved):

PR #1642, opened 2023-08-30
MINIFICPP-2204 Fix build with clang16 and upgrade in CI

lordgamez:

### For all changes:
     in the commit message?

### For code changes:

### Note:


.github/workflows/ci.yml
```diff
@@ -181,8 +181,8 @@ jobs:
           name: ubuntu-binaries
           path: build/bin
   ubuntu_20_04_clang:
```
martinzink, line 183: The name of the job was not updated

extensions/procfs/CpuStat.h
```diff
@@ -34,7 +34,7 @@ class CpuStatData {
  public:
   static std::optional<CpuStatData> parseCpuStatLine(std::istream& iss);
 
-  auto operator<=>(const CpuStatData& rhs) const = default;
+  bool operator>(const CpuStatData& rhs) const;
```
martinzink, line 37: What was the problem with the default spaceship operator?

lordgamez: Clang complained that the compared types are not default comparable so the default cooperation operator was implicitly deleted.

.github/workflows/ci.yml
lordgamez: Good catch, updated in ccb13b396bc13ea6786ff612b5ca771357795ca6

extensions/procfs/CpuStat.h
lordgamez: I'm not sure if we can, the compiler complains about not having a viable three-way comparison function for the members, which are of `SystemClockDuration` aka `std::chrono::duration<uint64_t, std::ratio<1, HZ>>` type. If that's not comparable according to clang I don't think we can do much about it.

szaszm: There is an implementation in TimeUtil.h, try including that

lordgamez: Thanks for pointing this out, it seems that this implementation was a workaround for this exact issue, but it was restricted to versions under libc++ 16. Removing this restrictions fixes the issue in 359181bf99d0e3b8cac9e0db92edd2c10eef42f5

martinzink (approved):

libminifi/include/utils/TimeUtil.h
```diff
@@ -35,15 +35,15 @@
 #include "StringUtils.h"
 
 // libc++ doesn't define operator<=> on durations, and apparently the operator rewrite rules don't automagically make one
-#if defined(_LIBCPP_VERSION) && _LIBCPP_VERSION < 16000
+#if defined(_LIBCPP_VERSION)
```
fgerlits, line 38: According to cppreference, `operator<=>(duration, duration)` should exist in C++20, so this looks like a bug/not-yet-implemented feature.  Can we add a static assert here which will fail when `libc++` fixes/implements this?  Otherwise, we won't notice that this workaround can be removed.

fgerlits (approved):

libminifi/include/utils/requirements/HasSpaceshipOperator.h
```diff
@@ -0,0 +1,31 @@
...
+
+#include <type_traits>
+
+namespace org::apache::nifi::minifi::utils {
+
+template<typename T, typename = std::void_t<>>
+struct has_spaceship_operator: std::false_type {};
+
+template<typename T>
+struct has_spaceship_operator<T, std::void_t<decltype(std::declval<T>() <=> std::declval<T>())>> : std::true_type {};
```
szaszm, line 29: I managed to accidentally comment on the commit instead of the PR. So here it comes again:

1. I'd keep this in the same file as the usage, as an implementation detail. (`namespace detail`)
2. We should use the concepts language tool instead of legacy metaprogramming hacks.

```cpp
template<typename T>
concept has_spaceship_operator = requires(T t, T u) {
  { t <=> u };
};
```

szaszm:
Try rebasing to main to solve the windows CI problem (#1650)

libminifi/include/utils/requirements/HasSpaceshipOperator.h
lordgamez: Updated in 96933e8c5206c3dd96f0a794260f41b49cce4d19

szaszm (approved):

PR #1648, opened 2023-08-31
MINIFICPP-2209 Small HTTPClient refactor: preconditions and exception…

szaszm:
… safety

- Remove unused header includes
- Fix exception safety in curl_slist generation: no longer leaks partial slist
- Rename `getCurlSList` to `toCurlSlist`, which is a function that converts an `unordered_map<string,string>` of HTTP headers to a `curl_slist*` linked list
- Simplified `curl_slist` deleter to a decltype(lambda)
- Remove unused `result` from `replaceInvalidCharactersInHttpHeaderFieldName`. (`ranges::actions::transform` is in-place)
- Tightened callback preconditions to fail fast on programming errors

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/http-curl/client/HTTPClient.cpp
+++ b/extensions/http-curl/client/HTTPClient.cpp
@@ -16,18 +16,15 @@
  */
 #include "HTTPClient.h"
 
-#include <algorithm>
 #include <cinttypes>
 #include <memory>
 #include <string>
 #include <utility>
 #include <vector>
 
-#include "Exception.h"
 #include "utils/gsl.h"
 #include "utils/StringUtils.h"
 #include "core/Resource.h"
-#include "utils/RegexUtils.h"
 #include "range/v3/algorithm/all_of.hpp"
 #include "range/v3/action/transform.hpp"
 #include "utils/HTTPUtils.h"
@@ -328,18 +325,15 @@ void HTTPClient::setRequestHeader(std::string key, std::optional<std::string> va
 }
 
 namespace {
-struct CurlSListFreeAll {
-  void operator()(struct curl_slist* slist) const {
-    curl_slist_free_all(slist);
-  }
-};
+using CurlSlistDeleter = decltype([](struct curl_slist* slist) { curl_slist_free_all(slist); });
 
-std::unique_ptr<struct curl_slist, CurlSListFreeAll> getCurlSList(const std::unordered_map<std::string, std::string>& request_headers) {
+std::unique_ptr<struct curl_slist, CurlSlistDeleter> toCurlSlist(const std::unordered_map<std::string, std::string>& request_headers) {
   curl_slist* new_list = nullptr;
+  const auto guard = gsl::finally([&new_list]() { curl_slist_free_all(new_list); });
   for (const auto& [header_key, header_value] : request_headers)
     new_list = curl_slist_append(new_list, utils::StringUtils::join_pack(header_key, ": ", header_value).c_str());
 
-  return {new_list, {}};
+  return {std::exchange(new_list, nullptr), {}};
 }
 }  // namespace
 
@@ -366,7 +360,7 @@ bool HTTPClient::submit() {
     curl_easy_setopt(http_session_.get(), CURLOPT_NOPROGRESS, 1);
   }
 
-  auto headers = getCurlSList(request_headers_);
+  const auto headers = toCurlSlist(request_headers_);
   if (headers) {
     curl_slist_append(headers.get(), "Expect:");
     curl_easy_setopt(http_session_.get(), CURLOPT_HTTPHEADER, headers.get());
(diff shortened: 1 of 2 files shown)


extensions/http-curl/client/HTTPClient.cpp
```diff
@@ -328,18 +325,15 @@ void HTTPClient::setRequestHeader(std::string key, std::optional<std::string> va
...
-  void operator()(struct curl_slist* slist) const {
-    curl_slist_free_all(slist);
-  }
-};
+using CurlSlistDeleter = decltype([](struct curl_slist* slist) { curl_slist_free_all(slist); });
 
-std::unique_ptr<struct curl_slist, CurlSListFreeAll> getCurlSList(const std::unordered_map<std::string, std::string>& request_headers) {
+std::unique_ptr<struct curl_slist, CurlSlistDeleter> toCurlSlist(const std::unordered_map<std::string, std::string>& request_headers) {
   curl_slist* new_list = nullptr;
+  const auto guard = gsl::finally([&new_list]() { curl_slist_free_all(new_list); });
```
fgerlits, line 332: this could be
```suggestion
  const auto guard = gsl::finally([&new_list]() { CurlSListDeleter()(new_list); });
```
to avoid duplication of the logic

fgerlits: I think it would improve readability: "use `CurlSListDeleter` on `new_list` if there is an exception, otherwise transfer ownership to the return value, which also uses `CurlSListDeleter`" -- instead of "use `curl_slist_free_all` or `CurlSListDeleter` depending on the situation", and then one has to do an extra step to find out that those two are the same.

But I agree it doesn't matter much either way.

fgerlits (approved):

extensions/http-curl/client/HTTPClient.cpp
szaszm: `CurlSListDeleter` is a type whose whole purpose is to call `curl_slist_free_all` as a `unique_ptr` deleter. I can add a comment to make that clear if this helps the case, but I prefer not to use it for anything other than a `unique_ptr` deleter for `curl_slist`.

extensions/http-curl/client/HTTPClient.cpp
```diff
@@ -328,18 +325,15 @@ void HTTPClient::setRequestHeader(std::string key, std::optional<std::string> va
...
   for (const auto& [header_key, header_value] : request_headers)
     new_list = curl_slist_append(new_list, utils::StringUtils::join_pack(header_key, ": ", header_value).c_str());
```
adamdebreceni, line 334: we could eliminate the need for the guard and the raw pointer, we also did not handle the error case of `curl_slist_append` returning a nullptr

from [curl.se](https://curl.se/libcurl/c/curl_slist_append.html)
`A null pointer is returned if anything went wrong, otherwise the new list pointer is returned. To avoid overwriting an existing non-empty list on failure, the new list should be returned to a temporary variable which can be tested for NULL before updating the original list pointer. `

```suggestion
  for (const auto& [header_key, header_value] : request_headers)
    std::unique_ptr<struct curl_slist, CurlSlistDeleter> appended_list{curl_slist_append(new_list.get(), utils::StringUtils::join_pack(header_key, ": ", header_value).c_str())};
    if (!appended_list) {
      throw std::runtime_error("Failed to append to slist");
    }
    new_list.release();
    new_list = std::move(appended_list);
```

extensions/http-curl/client/HTTPClient.cpp
```diff
@@ -328,18 +325,15 @@ void HTTPClient::setRequestHeader(std::string key, std::optional<std::string> va
 }
 
 namespace {
-struct CurlSListFreeAll {
-  void operator()(struct curl_slist* slist) const {
-    curl_slist_free_all(slist);
-  }
-};
+using CurlSlistDeleter = decltype([](struct curl_slist* slist) { curl_slist_free_all(slist); });
```
adamdebreceni, line 328: if the only purpose of `CurlSlistDeleter` is to be used in a `std::unique_ptr` it might be better to create a `CurlSlistUniquePtr` typedef

szaszm: I don't really think saving a few characters of typing with the alias is worth the extra mental indirection, I prefer spelling out the actual type of `std::unique_ptr<struct curl_slist, CurlSlistDeleter>` in this case.

extensions/http-curl/client/HTTPClient.cpp
szaszm: Handled the possible failure of `curl_slist_append` in [ef82531](https://github.com/apache/nifi-minifi-cpp/pull/1648/commits/ef82531959f1ee395775a825db37ccb8d68daf47).

I used a different approach that avoids overlapping ownership and the calling of `release`, because both of these make me uncomfortable. The raw pointer remained, but annotated it with `gsl::owner`. I should have done that originally.

adamdebreceni (approved):

lordgamez (approved):

PR #1651, opened 2023-09-06
MINIFICPP-2193 - Add manifest to debug bundle

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/src/c2/C2Agent.cpp
+++ b/libminifi/src/c2/C2Agent.cpp
@@ -694,8 +694,6 @@ C2Payload C2Agent::bundleDebugInfo(std::map<std::string, std::unique_ptr<io::Inp
     manifest.children = std::move(reported_manifest.serialized_nodes);
     manifest.array = reported_manifest.is_array;
     std::string manifest_str = manifest.to_pretty_string();
-    std::ofstream{"/home/adam/Desktop/manifest.json", std::ios::binary} << manifest_str;
-    std::cout << "Manifest size: " << manifest_str.size() << std::endl;
     if (!archiver->newEntry({MANIFEST_FILE_NAME, manifest_str.size()})) {
       throw C2DebugBundleError(fmt::format("Couldn't initialize archive entry for '{}'", MANIFEST_FILE_NAME));
     }
--- a/libminifi/src/c2/C2Agent.cpp
+++ b/libminifi/src/c2/C2Agent.cpp
@@ -676,7 +676,7 @@ C2Payload C2Agent::bundleDebugInfo(std::map<std::string, std::unique_ptr<io::Inp
   if (!archiver) {
     throw C2DebugBundleError("Couldn't instantiate archiver");
   }
-  for (auto&[filename, stream] : files) {
+  for (auto& [filename, stream] : files) {
     size_t file_size = stream->size();
     if (!archiver->newEntry({filename, file_size})) {
       throw C2DebugBundleError("Couldn't initialize archive entry for '" + filename + "'");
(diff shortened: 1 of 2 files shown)


libminifi/src/c2/C2Agent.cpp
```diff
@@ -686,6 +686,25 @@ C2Payload C2Agent::bundleDebugInfo(std::map<std::string, std::unique_ptr<io::Inp
...
   }
+  if (auto node_reporter = node_reporter_.lock()) {
+    constexpr const char* MANIFEST_FILE_NAME = "manifest.json";
+    auto reported_manifest = node_reporter->getAgentManifest();
+    state::response::SerializedResponseNode manifest;
+    manifest.name = std::move(reported_manifest.name);
+    manifest.children = std::move(reported_manifest.serialized_nodes);
+    manifest.array = reported_manifest.is_array;
+    std::string manifest_str = manifest.to_pretty_string();
+    std::ofstream{"/home/adam/Desktop/manifest.json", std::ios::binary} << manifest_str;
```
szaszm, line 697: debug leftover?

szaszm (approved):

libminifi/src/c2/C2Agent.cpp
```diff
@@ -686,6 +686,23 @@ C2Payload C2Agent::bundleDebugInfo(std::map<std::string, std::unique_ptr<io::Inp
       throw C2DebugBundleError("Error while writing file '" + filename + "' into the debug bundle");
     }
   }
+  if (auto node_reporter = node_reporter_.lock()) {
+    constexpr const char* MANIFEST_FILE_NAME = "manifest.json";
```
fgerlits, line 690: minor, but this should be `static constexpr`

szaszm: I think it doesn't make a difference in the case of `constexpr` strings, but I'm fine with either.

https://godbolt.org/z/9vvGK8Wfd

libminifi/src/c2/C2Agent.cpp
```diff
@@ -686,6 +686,23 @@ C2Payload C2Agent::bundleDebugInfo(std::map<std::string, std::unique_ptr<io::Inp
...
+    auto reported_manifest = node_reporter->getAgentManifest();
+    state::response::SerializedResponseNode manifest;
+    manifest.name = std::move(reported_manifest.name);
+    manifest.children = std::move(reported_manifest.serialized_nodes);
+    manifest.array = reported_manifest.is_array;
```
lordgamez, line 695: This could be refactored to:
```suggestion
    state::response::SerializedResponseNode manifest{
      .name = std::move(reported_manifest.name),
      .children = std::move(reported_manifest.serialized_nodes),
      .array = reported_manifest.is_array
    };
```

libminifi/src/c2/C2Agent.cpp
fgerlits: The difference is that in the case of plain `constexpr`, the compiler is allowed to optimize out putting the pointer on the stack, and in the case of `static constexpr`, it is required to.  You can see the difference if you change the example to `-O0`.

In practice, of course we always do optimized builds, and putting a pointer on the stack is not costly, so maybe I should have written "very minor" instead of "minor".  But I think as a matter of coding style, we should always make function-local `constexpr` variables `static`.

adamdebreceni: added static

libminifi/src/c2/C2Agent.cpp
adamdebreceni: makes sense, we don't actually even need the variable, refactored

fgerlits (approved):

lordgamez (approved):

PR #1653, opened 2023-09-08
MINIFICPP-1076 Yielding processor should still respect scheduling period

martinzink:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/test/unit/SchedulingAgentTests.cpp
+++ b/libminifi/test/unit/SchedulingAgentTests.cpp
@@ -176,9 +176,15 @@ TEST_CASE_METHOD(SchedulingAgentTestFixture, "Cron Driven no future triggers") {
   CHECK(first_task_reschedule_info.isFinished());
 }
 
-TEST_CASE_METHOD(SchedulingAgentTestFixture, "Timer driven should wait for scheduled run even if it yielded") {
-  count_proc_->setSchedulingPeriod(1min);
-  count_proc_->setYieldPeriodMsec(10ms);
+TEST_CASE_METHOD(SchedulingAgentTestFixture, "Timer driven should respect both yield and run schedule") {
+  SECTION("Fast yield slow schedule") {
+    count_proc_->setSchedulingPeriod(1min);
+    count_proc_->setYieldPeriodMsec(10ms);
+  }
+  SECTION("Slow yield fast schedule") {
+    count_proc_->setSchedulingPeriod(10ms);
+    count_proc_->setYieldPeriodMsec(1min);
+  }
   count_proc_->setShouldYield(true);
   auto timer_driven_agent = std::make_shared<TimerDrivenSchedulingAgent>(gsl::make_not_null(controller_services_provider_.get()), test_repo_, test_repo_, content_repo_, configuration_, thread_pool_);
   timer_driven_agent->start();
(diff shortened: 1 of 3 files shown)


libminifi/test/unit/SchedulingAgentTests.cpp
```diff
@@ -36,133 +38,166 @@ class CountOnTriggersProcessor : public minifi::core::Processor {
...
-    auto cron_driven_agent = std::make_shared<CronDrivenSchedulingAgent>(gsl::make_not_null(controller_services_provider_.get()), test_repo, test_repo, content_repo, configuration, thread_pool);
-    cron_driven_agent->start();
-    auto first_task_reschedule_info = cron_driven_agent->run(count_proc.get(), context, factory);
-    CHECK(!first_task_reschedule_info.isFinished());
-    CHECK(first_task_reschedule_info.getNextExecutionTime() <= std::chrono::steady_clock::now() + 1s);
-    CHECK(count_proc->getNumberOfTriggers() == 0);
-
-    std::this_thread::sleep_until(first_task_reschedule_info.getNextExecutionTime());
-    auto second_task_reschedule_info = cron_driven_agent->run(count_proc.get(), context, factory);
+    CHECK(std::chrono::round<std::chrono::minutes>(time_until_new_years_day - wait_time_till_next_execution_time) == 0min);
```
szaszm, line 144: I'm afraid that this is going to rarely fail when we go over the minute mark between the two timestamps, so I'd do a check for absolute difference being smaller than epsilon. (e.g. 1 min)

szaszm (approved):

libminifi/test/unit/SchedulingAgentTests.cpp
```diff
@@ -36,133 +38,167 @@ class CountOnTriggersProcessor : public minifi::core::Processor {
...
-  count_proc->incrementActiveTasks();
-  count_proc->setScheduledState(core::RUNNING);
-  auto node = std::make_shared<core::ProcessorNode>(count_proc.get());
-  auto context = std::make_shared<core::ProcessContext>(node, nullptr, repo, repo, content_repo);
-  std::shared_ptr<core::ProcessSessionFactory> factory = std::make_shared<core::ProcessSessionFactory>(context);
-  count_proc->setSchedulingPeriod(125ms);
 #ifdef WIN32
-  utils::dateSetInstall(TZ_DATA_DIR);
+    utils::dateSetInstall(TZ_DATA_DIR);
+    date::set_install(TZ_DATA_DIR);
```
fgerlits, line 67: Why do we need both?  `utils::dateSetInstall()` seems to just call `date::set_install()`.  In fact, why do we have `utils::dateSetInstall()` at all?

libminifi/test/unit/SchedulingAgentTests.cpp
```diff
@@ -36,133 +38,167 @@ class CountOnTriggersProcessor : public minifi::core::Processor {
...
+  count_proc_->setSchedulingPeriod(10ms);
+  count_proc_->setYieldPeriodMsec(1min);
+  count_proc_->setShouldYield(true);
+  auto timer_driven_agent = std::make_shared<TimerDrivenSchedulingAgent>(gsl::make_not_null(controller_services_provider_.get()), test_repo_, test_repo_, content_repo_, configuration_, thread_pool_);
+  timer_driven_agent->start();
+  auto first_task_reschedule_info = timer_driven_agent->run(count_proc_.get(), context_, factory_);
+  CHECK(!first_task_reschedule_info.isFinished());
+  CHECK(first_task_reschedule_info.getNextExecutionTime() > std::chrono::steady_clock::now() + 100ms);
+  CHECK(count_proc_->getNumberOfTriggers() == 1);
+}
```
fgerlits, line 201: I don't insist if you think it's clearer this way, but these two tests are identical except for the first two lines, so they could be merged into a single test with two SECTIONs around those pairs of lines.  To me, that would be more readable.

libminifi/test/unit/SchedulingAgentTests.cpp
martinzink: the tz_data_dir we set here is kept inside a static variable, so unfortunetly on windows we need to set it per library (or binary) basis.
The utils one sets it so it can be used from the libminifi the other one sets it so it can be used from the test binary.

(with an installed minifi the problem doesnt occur because we the location is set via an environment variable)

libminifi/test/unit/SchedulingAgentTests.cpp
martinzink: It just didnt occur to me, but it makes perfect sense to merge them, good idea https://github.com/apache/nifi-minifi-cpp/pull/1653/commits/c2d2ea6dbecaec9db5750b3d097878735d90d1eb

libminifi/test/unit/SchedulingAgentTests.cpp
martinzink: but maybe this should warrant a comment here or in the testutils.cpp for future reference

fgerlits (approved):

PR #1659, opened 2023-09-16
MINIFICPP-2218 Refactor expected monadic functions

szaszm:
- rename flatMap -> andThen, map -> transform, to match the standardized names
- implement transformError to match the standardized expected monadic interface
- simplified expected monadic function implementations
  - changed `std::decay_t` to `std::remove_cvref_t` in some cases. Decay is older, and does the same for simple objects, but remove_cvref is simpler, so I expect a lower chance of surprises, and the meaning is more obvious from the name
  - changed `std::is_same_v<T, void>` to `std::is_void_v<T>`
  - a bit less `decltype` with complex expressions, thanks to `std::invoke_result_t`. It's not fully equivalent, but I don't expect us to ever hit the edge cases.
  - using proper concept checks to restrict templates instead of ugly and complex metaprogramming hacks

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/CPPLINT.cfg
+++ b/CPPLINT.cfg
@@ -1,2 +1,2 @@
 set noparent
-filter=-runtime/reference,-runtime/string,-build/c++11,-build/include_subdir,-whitespace/forcolon,-build/namespaces_literals,-readability/check
+filter=-runtime/reference,-runtime/string,-build/c++11,-build/include_subdir,-whitespace/forcolon,-build/namespaces_literals,-readability/check,-build/include_what_you_use
--- a/libminifi/include/utils/expected.h
+++ b/libminifi/include/utils/expected.h
@@ -32,11 +32,14 @@ inline constexpr bool is_expected_v<nonstd::expected<V, E>> = true;
 template<typename T>
 concept expected = is_expected_v<std::remove_cvref_t<T>>;
 
+// from cppreference.com: The type must not be an array type, a non-object type, a specialization of std::unexpected, or a cv-qualified type.
+// base template: not cv-qualified, must be an object type, not an array type
 template<typename T>
 inline constexpr bool is_valid_unexpected_type_v = std::is_same_v<T, std::remove_cvref_t<T>> && std::is_object_v<T> && !std::is_array_v<T>;
 
-template<typename T>
-inline constexpr bool is_valid_unexpected_type_v<nonstd::unexpected_type<T>> = false;
+// specialization: the type must not be a specialization of std::unexpected
+template<typename E>
+inline constexpr bool is_valid_unexpected_type_v<nonstd::unexpected_type<E>> = false;
 
 template<typename T>
 concept valid_unexpected_type = is_valid_unexpected_type_v<T>;
@@ -77,19 +80,19 @@ auto operator|(Expected&& object, transform_wrapper<F> f) {
 template<expected Expected, typename F>
 auto operator|(Expected&& object, and_then_wrapper<F> f) {
   using value_type = typename std::remove_cvref_t<Expected>::value_type;
-  if constexpr (!std::is_void_v<value_type>) {
-    using function_return_type = std::remove_cvref_t<std::invoke_result_t<F, decltype(*std::forward<Expected>(object))>>;
-    static_assert(is_expected_v<function_return_type>, "flatMap expects a function returning expected");
-    if (object.has_value()) {
-      return std::invoke(std::forward<F>(f.function), *std::forward<Expected>(object));
-    } else {
-      return function_return_type{nonstd::unexpect, std::forward<Expected>(object).error()};
-    }
-  } else {
+  if constexpr (std::is_void_v<value_type>) {
     using function_return_type = std::remove_cvref_t<std::invoke_result_t<F>>;
     static_assert(is_expected_v<function_return_type>, "flatMap expects a function returning expected");
     if (object.has_value()) {
       return std::invoke(std::forward<F>(f.function));
+    } else {
+      return function_return_type{nonstd::unexpect, std::forward<Expected>(object).error()};
+    }
+  } else {
+    using function_return_type = std::remove_cvref_t<std::invoke_result_t<F, decltype(*std::forward<Expected>(object))>>;
+    static_assert(is_expected_v<function_return_type>, "flatMap expects a function returning expected");
+    if (object.has_value()) {
+      return std::invoke(std::forward<F>(f.function), *std::forward<Expected>(object));
     } else {
       return function_return_type{nonstd::unexpect, std::forward<Expected>(object).error()};
     }
--- a/libminifi/include/utils/expected.h
+++ b/libminifi/include/utils/expected.h
@@ -19,8 +19,6 @@
 #include <utility>
 #include "nonstd/expected.hpp"
 #include "utils/detail/MonadicOperationWrappers.h"
-#include "utils/GeneralUtils.h"
-#include "utils/meta/detected.h"
 
 namespace org::apache::nifi::minifi::utils {
 namespace detail {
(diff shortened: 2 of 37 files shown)


CPPLINT.cfg
```diff
@@ -1,2 +1,2 @@
 set noparent
-filter=-runtime/reference,-runtime/string,-build/c++11,-build/include_subdir,-whitespace/forcolon,-build/namespaces_literals,-readability/check
+filter=-runtime/reference,-runtime/string,-build/c++11,-build/include_subdir,-whitespace/forcolon,-build/namespaces_literals,-readability/check,-build/include_what_you_use
```
szaszm, line 2: `build/include_what_you_use` produces many false positives: it doesn't check the namespace, just the name of the function. I ran into another example in #1648
```
szaszm@szaszm-pc:~/nifi-minifi-cpp-3/build-gcc-make$ make linter
Built target minifi-usb-camera-extensions-linter
/home/szaszm/nifi-minifi-cpp-3/libminifi/include/utils/detail/MonadicOperationWrappers.h:61:  Add #include <algorithm> for transform  [build/include_what_you_use] [4]
Done processing /home/szaszm/nifi-minifi-cpp-3/libminifi/include/utils/detail/MonadicOperationWrappers.h
/home/szaszm/nifi-minifi-cpp-3/libminifi/include/DiskSpaceWatchdog.h:33:  At least two spaces is best between code and comments  [whitespace/comments] [2]
/home/szaszm/nifi-minifi-cpp-3/libminifi/include/DiskSpaceWatchdog.h:52:  At least two spaces is best between code and comments  [whitespace/comments] [2]
Done processing /home/szaszm/nifi-minifi-cpp-3/libminifi/include/DiskSpaceWatchdog.h
Total errors found: 1
/home/szaszm/nifi-minifi-cpp-3/libminifi/include/properties/Decryptor.h:46:  Add #include <algorithm> for transform  [build/include_what_you_use] [4]
Done processing /home/szaszm/nifi-minifi-cpp-3/libminifi/include/properties/Decryptor.h
/home/szaszm/nifi-minifi-cpp-3/libminifi/include/core/logging/LoggerConfiguration.h:65:  Add #include <algorithm> for transform  [build/include_what_you_use] [4]
Done processing /home/szaszm/nifi-minifi-cpp-3/libminifi/include/core/logging/LoggerConfiguration.h
```

libminifi/test/unit/ExpectedTest.cpp
```diff
@@ -484,3 +484,63 @@ TEST_CASE("expected valueOrElse", "[expected][valueOrElse]") {
...
+    REQUIRE(!ret);
+    REQUIRE(ret.error() == 42);
+  }
+
+  {
+    const nonstd::expected<int, int> e = nonstd::make_unexpected(21);
+    auto ret = std::move(e) | utils::transformError(mul2);  // NOLINT(performance-move-const-arg)
+    REQUIRE(!ret);
+    REQUIRE(ret.error() == 42);
+  }
```
fgerlits, line 517: I think `clang-tidy` has a point here: this test doesn't make much sense, since we can't move from a `const` variable.  I would remove this test case, and the others marked with `NOLINT(performance-move-const-arg)`, too.  Maybe they could be replaced with more realistic tests where the `expected` comes from the return value of a function.

libminifi/include/utils/expected.h
```diff
@@ -31,28 +29,42 @@ inline constexpr bool is_expected_v = false;
...
-  if constexpr (std::is_same_v<value_type, void>) {
-    using function_return_type = std::decay_t<decltype(std::invoke(std::forward<F>(f.function)))>;
+template<typename T>
+concept expected = is_expected_v<std::remove_cvref_t<T>>;
+
+template<typename T>
+inline constexpr bool is_valid_unexpected_type_v = std::is_same_v<T, std::remove_cvref_t<T>> && std::is_object_v<T> && !std::is_array_v<T>;
+
+template<typename T>
+inline constexpr bool is_valid_unexpected_type_v<nonstd::unexpected_type<T>> = false;
```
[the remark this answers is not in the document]
szaszm: I can rename it, but I don't understand the problem: it's a generic name used for type parameters when communicating more context isn't necessary. Is it really confusing to have identically named `typename T` in a partial template specialization and a related concept definition?
Would it be better to use separate names for all of them (let's say `T`, `U` and `V`), or some other arrangement (like `T` for the first one, `E` for the partial specialization, `F` for the concept)?

libminifi/include/utils/expected.h
```diff
@@ -61,56 +73,54 @@ auto operator|(Expected&& object, map_wrapper<F> f) {
...
-template<typename Expected, typename F, typename = std::enable_if_t<is_expected_v<utils::remove_cvref_t<Expected>>>>
-auto operator|(Expected&& object, flat_map_wrapper<F> f) {
-  using value_type = typename utils::remove_cvref_t<Expected>::value_type;
-  if constexpr (std::is_same_v<value_type, void>) {
-    using function_return_type = std::decay_t<decltype(std::invoke(std::forward<F>(f.function)))>;
+// andThen
+template<expected Expected, typename F>
+auto operator|(Expected&& object, and_then_wrapper<F> f) {
+  using value_type = typename std::remove_cvref_t<Expected>::value_type;
+  if constexpr (!std::is_void_v<value_type>) {
```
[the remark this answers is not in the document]
szaszm: No specific reason, I just had to rewrite the void support after attempting to remove it, and realizing that it's still needed, and it just happened to be in this order. I can flip it if it makes things clearer.

libminifi/test/unit/ExpectedTest.cpp
szaszm: It doesn't make much sense to use const rvalue refs, but this still checks that the implementation matches the expectations in this unlikely case. I just copy pasted the transform tests from above, which were copy pasted from TartanLlama's test suite.

How would you structure the suggested new test case, that ends up testing the expected implementation, and not the C++ language itself?

libminifi/include/utils/expected.h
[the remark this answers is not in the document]
fgerlits: The way I see it, the `T`s in lines 35 and 41 refer to the same thing, and if we rewrite this template to use `E`, then `T = nonstd::unexpected_type<E>`.  Using `T` here makes it unclear that this is a template specialization.

If you think using three different template variables is better, that's fine, too.

libminifi/include/utils/expected.h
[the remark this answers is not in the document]
fgerlits: Yes, I think using the same order everywhere would make things clearer.

libminifi/test/unit/ExpectedTest.cpp
fgerlits: As it is, this is the same test as the one at lines 499-502, just with a confusing `move` added which doesn't do anything.

I think something like this would be slightly different from lines 506-509:
```c++
  {
    const auto expected_provider = [] { return = nonstd::expected<int, int>(nonstd::unexpected, 21); };
    auto ret = expected_provider() | utils::transformError(mul2);
    REQUIRE(!ret);
    REQUIRE(ret.error() == 42);
  }
```
but if you don't think it's different enough, I'm OK with removing this test case.

szaszm: I don't mind replacing the mutable rvalue case with this, since it's what is being tested in your example, but I'd still keep the const rvalue case, with the NOLINT exclusion, to keep a full coverage of possible input reference types.

libminifi/include/utils/expected.h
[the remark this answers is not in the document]
szaszm: Both specializations refer to a possible `E` argument for `unexpected<E>`. Shouldn't both of them be changed to `E` instead?

libminifi/test/unit/ExpectedTest.cpp
```diff
@@ -153,124 +153,124 @@ TEST_CASE("expected map", "[expected][map]") {
...
     REQUIRE(ret.error() == 21);
   }
 
   {
     const nonstd::expected<int, int> e(nonstd::unexpect, 21);
-    auto ret = std::move(e) | utils::flatMap(fail);  // NOLINT(performance-move-const-arg)
+    auto ret = std::move(e) | utils::andThen(fail);  // NOLINT(performance-move-const-arg)
     REQUIRE(!ret);
     REQUIRE(ret.error() == 21);
   }
```
martinzink, line 276: There is a TEST_CASE here with a comment 
`// commented out parts depend on https://github.com/martinmoene/expected-lite/pull/40 (waiting for release)`
I think we upgraded to the release which it dependend upon. Could you enable those test cases aswell?

fgerlits (approved):

libminifi/test/unit/ExpectedTest.cpp
szaszm: Thanks for pointing this out, reenabled them in [47aede1](https://github.com/apache/nifi-minifi-cpp/pull/1659/commits/47aede10d43a4b9e0c0b27b67dda824bd5748096)

martinzink (approved):

PR #1668, opened 2023-09-27
MINIFICPP-2226 Remove flakiness from LoggerTests

lordgamez:
Randomization is removed from the testcase to remove flakiness. We only check if multiple segments are present in the compressed logs.

https://issues.apache.org/jira/browse/MINIFICPP-2226

------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/test/unit/LoggerTests.cpp
+++ b/libminifi/test/unit/LoggerTests.cpp
@@ -20,7 +20,6 @@
 #include <memory>
 #include <vector>
 #include <ctime>
-#include <random>
 #include "../TestBase.h"
 #include "../Catch.h"
 #include "core/logging/LoggerConfiguration.h"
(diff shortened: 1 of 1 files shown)


martinzink (approved):

libminifi/test/unit/LoggerTests.cpp
```diff
@@ -364,24 +363,14 @@ TEST_CASE("Test sending multiple segments at once", "[ttl16]") {
...
+  for (size_t i = 0; i < 100; ++i) {
     logger->log_error(log_str.c_str());
   }
 
   LoggerTestAccessor::runCompression(log_config);
-
   auto compressed_logs = logging::LoggerConfiguration::getCompressedLogs();
-  REQUIRE(compressed_logs.size() == SEGMENT_COUNT);
-  auto logs = decompress(compressed_logs[SEGMENT_COUNT - 1]);
+  REQUIRE(compressed_logs.size() > 1);
```
fgerlits, line 373: Was randomization of the content of the log lines really a problem?  Do we understand why the number of segments is sometimes 4 instead of 5?

It seems to me that the change in this one line: "number of segments > 1" instead of "number of segments == 5" fixes the flakiness, but it makes the test much weaker.

I would prefer to track down the underlying reason why the segment count returned by `getCompressedLogs()` is not always 5, and fix that.  (E.g. is there any multi-threading going on? can we add a timeout, or (better) some way to query if all pending tasks have finished?)

lordgamez: The point of this test case was to see if multiple segments are sent if the compressed logs are segmented due to their size. If it can be achieved by using fix input log entries, it should be enough to test this use case, and the result should be more consistent. Why do you think the randomization in this test should be kept?

lordgamez: I think the inconsistency comes from the `LogCompressorSink` that creates a separate thread for processing and compressing the incoming logs in parallel. I will have to check it how the test can be changed to make the results more consistent.

lordgamez:
@fgerlits took over the investigation of this ticket as it seems to be a bug not a flaky test case issue

PR #1674, opened 2023-10-02
MINIFICPP-2239 Add bandwidth limit to InvokeHTTP processor

lordgamez:
- Additionally fix ListenHTTP response setting
- Add processor name and id if input connection is required or forbidden

https://issues.apache.org/jira/browse/MINIFICPP-2239

--------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/http-curl/client/HTTPClient.h
```diff
@@ -149,6 +149,8 @@ class HTTPClient : public utils::BaseHTTPClient, public core::Connectable {
 
   void setFollowRedirects(bool follow);
 
+  void setMaximumUploadBandwidth(uint64_t max_bandwidth);
+  void setMaximumDownloadBandwidth(uint64_t max_bandwidth);
```
szaszm, line 153: This needs a unit throughout the interface. A suffix in the identifiers is good enough, with a comment somewhere. Is it Bps?

libminifi/src/core/Processor.cpp
```diff
@@ -345,15 +345,15 @@ void Processor::validateAnnotations() const {
   switch (getInputRequirement()) {
     case annotation::Input::INPUT_REQUIRED: {
       if (!hasIncomingConnections()) {
-        throw Exception(PROCESS_SCHEDULE_EXCEPTION, "INPUT_REQUIRED was specified for the processor, but no incoming connections were found");
+        throw Exception(PROCESS_SCHEDULE_EXCEPTION, "INPUT_REQUIRED was specified for processor '" + getName()  + "' (uuid: '" + getUUIDStr() + "'), but no incoming connections were found");
```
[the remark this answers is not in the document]
lordgamez: Good idea, updated in fc9144144a07100efc52ef0c80a462db7995f38e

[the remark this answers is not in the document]
szaszm: I also suggest not removing the "the" from before the word "processor"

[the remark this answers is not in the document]
lordgamez: Updated in c8dd0743843bb2b2fd505113ee66693ecf1537df

extensions/http-curl/processors/InvokeHTTP.h
```diff
@@ -177,8 +177,20 @@ class InvokeHTTP : public core::Processor {
       .withDefaultValue(magic_enum::enum_name(invoke_http::InvalidHTTPHeaderFieldHandlingOption::transform))
       .withAllowedValues(magic_enum::enum_names<invoke_http::InvalidHTTPHeaderFieldHandlingOption>())
       .build();
+  EXTENSIONAPI static constexpr auto UploadBandwidthLimit = core::PropertyDefinitionBuilder<>::createProperty("Upload Bandwidth Limit")
+      .withDescription("Maximum data per second to send, use 0 for unlimited.")
+      .withPropertyType(core::StandardPropertyTypes::DATA_SIZE_TYPE)
+      .withDefaultValue("0 B")
+      .isRequired(true)
+      .build();
+  EXTENSIONAPI static constexpr auto DownloadBandwidthLimit = core::PropertyDefinitionBuilder<>::createProperty("Download Bandwidth Limit")
```
szaszm, line 186: I would use the world "speed" instead of "bandwidth".

If we want to be precise, the word "bandwidth" for general network transfer speeds is inaccurate: modern networks can transfer much more data even on narrower bands than older protocols. In everyday speech, in the context of internet speed, the two words are used as synonyms, but in this context, I think "speed" is clearer.

lordgamez: Updated in 4f1691020c4bb6b7c39c131d1aa15b2d0f2ecedd

szaszm (approved):

extensions/http-curl/processors/InvokeHTTP.h
martinzink: Im a bit torn on this because the current problem is that the dimension of the property doesnt match its name. 
Ideally to a "speed" we could accept something like 100 KBps, but that would require a much larger change, maybe we could  rename it so its name at least matches the expected unit e.g. "Upload data limit per seconds"

What do you guys think? @lordgamez  @szaszm

szaszm: It's not ideal, but I don't think renaming to "upload data limit per second" would make it much better. Technically it's more fitting, but it's still confusing, and the name is less like what a user would look for.

I'd prefer to go with the current version, and if we introduce the necessary units to the ecosystem later, then we can extend the range of accepted values to include the proper unit.

A compromise could be naming it "Upload Speed Limit (/sec)".

martinzink: I like the "Upload Speed Limit (/sec)" naming

lordgamez: I think the "per second" is a good idea, maybe we could name them "Data Upload Limit Per Second" and "Data Download Limit Per Second"?

szaszm: I'd prefer to keep referring to "download speed" and "upload speed" in the property name.

Another alternative approach would be to use a custom property type, with no corresponding C2/NiFi validator, and just parse the proper data rate units in onSchedule. I'd like that the most, but it requires a bit more additional code.

szaszm (changes requested):
let's wait for some more input and finish the discussion about the property naming, before merging

extensions/http-curl/processors/InvokeHTTP.h
szaszm: If there are no objections, I suggest we keep the name, and change the format to e.g. "MB/s" meaning mebibytes/sec (to be consistent with data size units), with agent-side validation, and no specific validator in the C2 manifest.

PROCESSORS.md
```diff
@@ -1320,6 +1320,8 @@ In the list below, the names of required properties appear in bold. Any other pr
...
+| **Upload Speed Limit**                          | 0 B/s                    |                             | Maximum data per second to send, use 0 B/s for unlimited.                                                                                                                                                                                                                                                       |
+| **Download Speed Limit**                        | 0 B/s                    |                             | Maximum data per second to receive, use 0 B/s for unlimited.                                                                                                                                                                                                                                                    |
```
szaszm, line 1324: It would be better to treat an empty value as no limit. We can also keep treating "0 B/s" as no limit, but I'd document the empty as unlimited instead.
```suggestion
| **Upload Speed Limit**                          | 0 B/s                    |                             | Maximum data per second to send. Leave this empty if you want no limit.                                                                                                                                                                                                                                                       |
| **Download Speed Limit**                        | 0 B/s                    |                             | Maximum data per second to receive. Leave this empty if you want no limit.                                                                                                                                                                                                                                                    |
```

libminifi/include/core/TypedValues.h
```diff
@@ -149,10 +144,39 @@ class DataSizeValue : public TransformableValue, public state::response::UInt64V
...
+
+  DataTransferSpeedValue()
+      : state::response::UInt64Value(0) {
+  }
+
+  static std::string removePerSecSuffix(const std::string &input) {
+    auto lower_case_input = utils::StringUtils::toLower(input);
+    if (lower_case_input.ends_with("/sec")) {
+      return input.substr(0, input.size() - 4);
+    }
```
szaszm, line 169: I'd prefer to keep the set of accepted formats smaller, and only accept "/s", not "/sec" or "ps". It's easy to extend later, but it harder to break what worked before.

lordgamez: Updated in f8833c3b01c5edf0ff60fff593b2f21ab91763cd

libminifi/include/core/PropertyType.h
```diff
@@ -319,6 +319,24 @@ class TimePeriodPropertyType : public PropertyType {
   }
 };
 
+class DataTransferSpeedPropertyType : public PropertyType {
+ public:
+  constexpr ~DataTransferSpeedPropertyType() override {}  // NOLINT see comment at parent
+
+  [[nodiscard]] std::string_view getValidatorName() const override { return "DATA_SIZE_VALIDATOR"; }
```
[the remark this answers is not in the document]
lordgamez: Changed to `DATA_TRANSFER_SPEED_VALIDATOR` in f8833c3b01c5edf0ff60fff593b2f21ab91763cd, it should be ignored on C2 server side if not implemented.

extensions/http-curl/processors/InvokeHTTP.cpp
```diff
@@ -115,6 +115,8 @@ void InvokeHTTP::setupMembersFromProperties(const core::ProcessContext& context)
 
   use_chunked_encoding_ = (context.getProperty(UseChunkedEncoding) | utils::andThen(&utils::StringUtils::toBool)).value_or(false);
   send_date_header_ = context.getProperty<bool>(DateHeader).value_or(true);
+  context.getProperty(UploadSpeedLimit, maximum_upload_speed_);
+  context.getProperty(DownloadSpeedLimit, maximum_download_speed_);
```
fgerlits, line 119: These will parse the the properties as numbers, ignoring the suffix, so `Download Speed Limit = 10 MB/s` will result in `maximum_download_speed_ = 10` instead of `maximum_download_speed_ = 10000000`.

If you changed the type of these two fields to `core::DataTransferSpeedValue`, then it would work.  In that case, we would need to call `getValue()` when passing the values to curl.

libminifi/include/core/TypedValues.h
```diff
@@ -149,10 +144,36 @@ class DataSizeValue : public TransformableValue, public state::response::UInt64V
...
+  DataTransferSpeedValue()
+      : state::response::UInt64Value(0) {
+  }
+
+  static std::string removePerSecSuffix(const std::string &input) {
+    auto lower_case_input = utils::StringUtils::trim(utils::StringUtils::toLower(input));
+    if (lower_case_input.ends_with("/s")) {
+      return lower_case_input.substr(0, lower_case_input.size() - 2);
+    }
+    throw utils::internal::ParseException("Couldn't parse DataTransferSpeedValue, no valid suffix (/s, /sec, ps) found!");
```
fgerlits, line 170: this error message is out of date

docker/test/integration/features/MiNiFi_integration_test_driver.py
```diff
@@ -244,11 +246,21 @@ def check_for_num_files_generated(self, num_flowfiles, timeout_seconds):
...
-    def check_for_num_file_range_generated(self, min_files, max_files, wait_time_in_seconds):
+    def check_for_num_file_range_generated_after_timeout(self, min_files: int, max_files: int, wait_time_in_seconds: int):
         output_validator = NumFileRangeValidator(min_files, max_files)
         output_validator.set_output_dir(self.file_system_observer.get_output_dir())
         self.__check_output_after_time_period(wait_time_in_seconds, output_validator)
 
+    def check_for_num_file_range_generated(self, min_files: int, max_files: int, wait_time_in_seconds: int):
+        output_validator = NumFileRangeValidator(min_files, max_files)
+        output_validator.set_output_dir(self.file_system_observer.get_output_dir())
+        self.__check_output_over_time_period(wait_time_in_seconds, output_validator)
```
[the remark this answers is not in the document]
lordgamez: Updated in 41364556e36680ca7f8d564ffa08fe110188a34c

extensions/http-curl/processors/InvokeHTTP.cpp
lordgamez: Good catch, updated in 41364556e36680ca7f8d564ffa08fe110188a34c

libminifi/include/core/TypedValues.h
lordgamez: Updated in 41364556e36680ca7f8d564ffa08fe110188a34c

fgerlits (approved):

martinzink (approved):

libminifi/include/core/PropertyType.h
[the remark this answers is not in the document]
lordgamez: It seems that some C2 servers do not ignore the new validator, but return on error on it, so I removed the validator for this type in 2ec3285238625ebd7348f517d14cf6840f456351

PROCESSORS.md
```diff
@@ -1320,6 +1320,8 @@ In the list below, the names of required properties appear in bold. Any other pr
...
+| Upload Speed Limit                              |                          |                             | Maximum data per second to send (e.g. '500 KB/s'). Leave this empty if you want no limit.                                                                                                                                                                                                                       |
+| Download Speed Limit                            |                          |                             | Maximum data per second to receive (e.g. '500 KB/s'). Leave this empty if you want no limit.                                                                                                                                                                                                                    |
```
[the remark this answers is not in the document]
lordgamez: Updated in d305cce439c0b60fac04b646b79f46f3a4e68d54

szaszm (approved):

PR #1698, opened 2023-11-28
MINIFICPP-2262 Add processor to push logs to Grafana Loki through gRPC

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2262

---------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


CMakeLists.txt
```diff
@@ -352,6 +356,10 @@ if (ENABLE_ALL OR ENABLE_PROMETHEUS OR NOT DISABLE_CIVET)
     list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/civetweb/dummy")
 endif()
 
+if (ENABLE_ALL OR ENABLE_GCP OR ENABLE_GRAFANA_LOKI)
+    include(Abseil)
+endif()
```
martinzink, line 361: Why do we need this here? Couldnt we include abseil twice? in gcp and grafana aswell (after the early returns ofcourse), since its a fetchcontent makeavailable we should be able to call it twice

extensions/grafana-loki/CMakeLists.txt
```diff
@@ -16,20 +16,46 @@
...
 target_include_directories(minifi-grafana-loki PRIVATE BEFORE "${CMAKE_SOURCE_DIR}/extensions/http-curl")
-target_link_libraries(minifi-grafana-loki ${LIBMINIFI})
-target_link_libraries(minifi-grafana-loki minifi-http-curl)
+target_link_libraries(minifi-grafana-loki ${LIBMINIFI} minifi-http-curl)
+add_dependencies(minifi-grafana-loki minifi-http-curl)
+
+if (ENABLE_GRPC)
+    target_include_directories(minifi-grafana-loki PRIVATE BEFORE "${GRPC_INCLUDE_DIR}" "${PROTOBUF_INCLUDE_DIR}")
+    target_link_libraries(minifi-grafana-loki grpc++ protobuf::libprotobuf)
+    add_dependencies(minifi-grafana-loki grpc grpc++ protobuf::libprotobuf grafana-loki-protos)
```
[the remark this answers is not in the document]
szaszm: `target_link_libraries`: they should be linked together
`add_dependencies`: these targets should be compiled before the first in the list, not allowing concurrent compilation.

We should avoid `add_dependencies`, unless we're running the binaries of the result of a previous build step, or similar. In this case, is that the generation of sources with protobuf?

extensions/grafana-loki/protos/grafana-loki-push.proto
szaszm, line 1: Is this file coming from somewhere else? We should include the source, and use the appropriate license.

lordgamez: This file defines the grpc service and the protobuf messages structures according to grafana's protocol definition in this file: https://github.com/grafana/loki/blob/main/pkg/push/push.proto

Only the used fields and attributes are kept, removing the go specific attributes and imports. In the Grafana repository the containing directory holds an Apache license (https://github.com/grafana/loki/blob/main/pkg/push/LICENSE) so I think we should be okay with the license header, I can add a reference comment to the original proto file it that's okay.

LICENSE
```diff
@@ -2442,52 +2443,6 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
-This product bundles 'protobuf' within 'OpenCV' under a 3-Clause BSD license:
```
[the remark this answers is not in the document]
lordgamez: Yes, I disabled the build of the dnn module of OpenCV in `cmake/BundledOpenCV.cmake` that used the protobuf library as we don't use it in our OpenCV processors.

cmake/MiNiFiOptions.cmake
```diff
@@ -122,6 +122,7 @@ add_minifi_option(ENABLE_KUBERNETES "Enables the Kubernetes extensions." ON)
 add_minifi_option(ENABLE_TEST_PROCESSORS "Enables test processors" OFF)
 add_minifi_option(ENABLE_PROMETHEUS "Enables Prometheus support." ON)
 add_minifi_option(ENABLE_GRAFANA_LOKI "Enable Grafana Loki support" OFF)
+add_minifi_option(ENABLE_GRPC "Enable gRPC for Grafana Loki extension" ON)
```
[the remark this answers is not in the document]
lordgamez: Good point, renamed to `ENABLE_GRPC_FOR_LOKI` in 407abbc65a8900dc9937dc0587c52d68693335ff

CMakeLists.txt
lordgamez: You are right, updated in 407abbc65a8900dc9937dc0587c52d68693335ff

extensions/grafana-loki/CMakeLists.txt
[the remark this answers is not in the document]
lordgamez: The grafana-loki-protos is the target to generate the protobuf cpp files, the minifi-grafana-loki target should depend on that, but you are right the other targets should not be included in the dependencies, updated in 407abbc65a8900dc9937dc0587c52d68693335ff

.dockerignore
```diff
@@ -57,6 +57,11 @@ extensions/expression-language/Scanner.cpp
 extensions/expression-language/location.hh
 extensions/expression-language/position.hh
 extensions/expression-language/stack.h
+extensions/grafana-loki/protobuf-generated/grafana-loki-push.grpc.pb.h
+extensions/grafana-loki/protobuf-generated/grafana-loki-push.grpc.pb.cc
+extensions/grafana-loki/protobuf-generated/grafana-loki-push.pb.h
+extensions/grafana-loki/protobuf-generated/grafana-loki-push.pb.cc
```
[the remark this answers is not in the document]
lordgamez: Good idea, moved them in 407abbc65a8900dc9937dc0587c52d68693335ff

extensions/grafana-loki/protos/grafana-loki-push.proto
lordgamez: Added reference in 407abbc65a8900dc9937dc0587c52d68693335ff

extensions/grafana-loki/PushGrafanaLoki.cpp
```diff
@@ -0,0 +1,249 @@
...
+  setUpStateManager(context);
+  setUpStreamLabels(context);
+
+  if (auto log_line_metadata_attributes = context.getProperty(LogLineMetadataAttributes)) {
+    log_line_metadata_attributes_ = utils::string::splitAndTrimRemovingEmpty(*log_line_metadata_attributes, ",");
+  }
+
+  auto log_line_batch_wait = context.getProperty<core::TimePeriodValue>(LogLineBatchWait);
+  auto log_line_batch_size = context.getProperty<uint64_t>(LogLineBatchSize);
+    if (log_line_batch_size && *log_line_batch_size < 1) {
```
szaszm, line 141: strange indentation here

lordgamez: Good catch, it seems that some review changes from MINIFICPP-2261 were lost when rebasing this PR. I restored those changes in 970b779746f31f57ea7e95e34008b5cea28ba843

szaszm (approved):

adamdebreceni (approved):

martinzink (approved):

PR #1706, opened 2023-12-12
MINIFICPP-2264 GenerateFlowFile: 'Custom Text' should be reevaluated …

martinzink:
…once per batch

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/standard-processors/processors/GenerateFlowFile.cpp
```diff
@@ -58,57 +58,69 @@ void generateData(std::vector<char>& data, bool textData = false) {
...
+// If the Data Format is text and if Unique FlowFiles is false, the custom text has to be evaluated once per batch
+void GenerateFlowFile::regenerateNonUniqueData(core::ProcessContext& context) {
+  std::string custom_text;
+  context.getProperty(CustomText, custom_text, nullptr);
+  if (!custom_text.empty()) {
+    if (textData_ && !unique_flow_file_) {
+      non_unique_data_.assign(custom_text.begin(), custom_text.end());
+      return;
+    }
+    logger_->log_warn("Custom Text property is set, but not used!");
```
fgerlits, line 101: I think this would be simpler, and would avoid doing some unnecessary work:
```c++
  if (!textData_ || unique_flow_file_) return;

  std::string custom_text;
  context.getProperty(CustomText, custom_text, nullptr);
  if (!custom_text.empty()) {
    non_unique_data_.assign(custom_text.begin(), custom_text.end());
  }
```
We would lose the warning, but we have warned in `onSchedule()` already.

Also, if the expanded custom text is empty now, but it wasn't empty before, I think the flow file should be empty, so
```c++
  if (!textData_ || unique_flow_file_) return;

  std::string custom_text;
  context.getProperty(CustomText, custom_text, nullptr);
  non_unique_data_.assign(custom_text.begin(), custom_text.end());
```
would be better.

martinzink: The last suggestion would remove the possibility to use GenerateFlowFiles with non unique randomly generated set length data. (UniqueFlowFiles -> false, FileSize -> 10B, without CustomData, would generate empty flowfiles)
How about we make a distinction between CustomText not set and CustomText evaluated to be empty?

```
void GenerateFlowFile::regenerateNonUniqueData(core::ProcessContext& context) {
  if (!textData_ || unique_flow_file_) return;

  if (auto custom_text = context.getProperty(CustomText, nullptr)) {
    non_unique_data_.assign(custom_text->begin(), custom_text->end());
  }
}
```

martinzink: Nifi seem to work similarly:
  - CustomText not set -> 10B random text
  - CustomText -> ${something_that_doesnt_exists}

One thing thats different that nifi doesnt allow empty string for customtext

martinzink: https://github.com/apache/nifi-minifi-cpp/pull/1706/commits/89c76dc4a13ab07d6a73385f5eda37aafc2dc31b#diff-45136a0df5089a3527479406b275001e601b5a9e02fa3a345be89e18e141b03bR96-R98

I've also fixed the cut off description of Unique FlowFiles https://github.com/apache/nifi-minifi-cpp/pull/1706/commits/89c76dc4a13ab07d6a73385f5eda37aafc2dc31b#diff-fd2410931e7fdc4bf8b3ce23f5f7a27c7aacdf9337320626d86d806448c90b9bR1138

And simplified the tests so they don't rely on PutFile (also speeds things up a bit because we dont have to wait for filesystem io)

fgerlits: Yes, good point: we used to treat a `Custom Text` set to empty the same as if it wasn't set at all, which is consistent with most other processors, and we probably don't want to change that.

I'm OK with either remembering in `onSchedule` if `Custom Text` was empty, and using that info here, or to just treat a `Custom Text` expanded to empty as if it was empty originally -- whichever you prefer.  A non-empty `Custom Text` expanding to empty is an edge case which may never happen in real life.

martinzink: I've refactored a little bit so Its easier to see what determines the output mode
Made sure that we dont break compatibility so CustomText "" is the same as no CustomText, but if CustomText were to evaluate to "" we would create an empty FF.
https://github.com/apache/nifi-minifi-cpp/pull/1706/commits/626d69c49f1c2ba04ccbd36d3ceba1fff6062e76

martinzink: Added an other test that covers this last use case https://github.com/apache/nifi-minifi-cpp/pull/1706/commits/13f0f74c7e46e4228c0b1cab7381a3b8416094ec

fgerlits (approved):

extensions/standard-processors/processors/GenerateFlowFile.h
```diff
@@ -70,14 +66,15 @@ class GenerateFlowFile : public core::Processor {
       .withDefaultValue("Binary")
       .build();
   EXTENSIONAPI static constexpr auto UniqueFlowFiles = core::PropertyDefinitionBuilder<>::createProperty("Unique FlowFiles")
-      .withDescription("If true, each FlowFile that is generated will be unique. If false, a random value will be generated and all FlowFiles")
+      .withDescription("If true, each FlowFile that is generated will be unique. "
+          "If false, a random value will be generated and all FlowFiles will get the same content but this offers much higher throughput")
```
adamdebreceni, line 70: all FlowFiles in a batch?

extensions/standard-processors/processors/GenerateFlowFile.cpp
```diff
@@ -57,58 +56,78 @@ void generateData(std::vector<char>& data, bool textData = false) {
...
+  mode_ = getMode(is_unique, is_text, has_custom_text, file_size_);
+
+  if (!isUnique(mode_)) {
+    non_unique_data_.resize(gsl::narrow<size_t>(file_size_));
+    generateData(non_unique_data_, isText(mode_));
   }
 
+  logger_->log_trace("GenerateFlowFile is configured in {} mode", magic_enum::enum_name(mode_));
+  if (mode_ != Mode::CustomText && has_custom_text)
+    logger_->log_warn("Custom Text property is set, but not used!");
```
szaszm, line 100: It may be worth clarifying why it's not used, or how to get it to be used.

extensions/standard-processors/processors/GenerateFlowFile.cpp
```diff
@@ -57,58 +56,78 @@ void generateData(std::vector<char>& data, bool textData = false) {
...
 
-void GenerateFlowFile::onTrigger(core::ProcessContext&, core::ProcessSession& session) {
-  for (uint64_t i = 0; i < batchSize_; i++) {
+void GenerateFlowFile::onTrigger(core::ProcessContext& context, core::ProcessSession& session) {
+  refreshNonUniqueData(context);
+  for (uint64_t i = 0; i < batch_size_; i++) {
     // For each batch
-    std::shared_ptr<core::FlowFile> flowFile = session.create();
-    if (!flowFile) {
+    std::shared_ptr<core::FlowFile> flow_file = session.create();
```
szaszm, line 116: The comment on line 115 is confusing. The for cycle is one batch, and each iteration is one flow file, but it suggests as if each iteration was one batch.

extensions/standard-processors/processors/GenerateFlowFile.cpp
```diff
@@ -57,58 +56,78 @@ void generateData(std::vector<char>& data, bool textData = false) {
...
       return;
     }
-    if (uniqueFlowFile_) {
-      std::vector<char> data(gsl::narrow<size_t>(fileSize_));
-      if (fileSize_ > 0) {
-        generateData(data, textData_);
-      }
-      session.writeBuffer(flowFile, data);
+    if (mode_ == Mode::Empty) {
+      // noop
```
szaszm, line 122: Shouldn't we emit an empty flow file here?

martinzink: we do, the creation is in line 116 and the transfer is after all these ifs in line 130, We only skip the modification of the created empty flowfile

extensions/standard-processors/processors/GenerateFlowFile.h
martinzink: All flowfiles period.
The description is directly from NiFi. And afaik nifi and minifi works in the same way so if this is false we only generate the date once. (except if CustomText is set then we should evaluate it once per batch)

extensions/standard-processors/processors/GenerateFlowFile.cpp
martinzink: Clarified this in https://github.com/apache/nifi-minifi-cpp/pull/1706/commits/59becf11fea3e4d986b11acf6e8f33765a8b3408

extensions/standard-processors/processors/GenerateFlowFile.cpp
martinzink: Removed the comment (wasnt doing anything useful in my opinion) in https://github.com/apache/nifi-minifi-cpp/pull/1706/commits/59becf11fea3e4d986b11acf6e8f33765a8b3408

lordgamez (approved):

extensions/standard-processors/processors/GenerateFlowFile.h
adamdebreceni: yes that is what I meant, so the description is not true

szaszm (approved):

extensions/standard-processors/processors/GenerateFlowFile.h
adamdebreceni: as discussed the description of CustomText is clear enough on how to use that and what to expect

adamdebreceni (approved):

PR #1708, opened 2023-12-13
MINIFICPP-2229 Encrypt sensitive properties in the flow configuration

fgerlits:
https://issues.apache.org/jira/browse/MINIFICPP-2229

---

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/azure/processors/AzureBlobStorageProcessorBase.h
```diff
@@ -51,10 +51,13 @@ class AzureBlobStorageProcessorBase : public AzureStorageProcessorBase {
       .withDescription("The storage account key. This is an admin-like password providing access to every container in this account. "
           "It is recommended one uses Shared Access Signature (SAS) token instead for fine-grained control with policies.")
       .supportsExpressionLanguage(true)
+      .isSensitive(true)
       .build();
   EXTENSIONAPI static constexpr auto SASToken = core::PropertyDefinitionBuilder<>::createProperty("SAS Token")
       .withDescription("Shared Access Signature token. Specify either SAS Token (recommended) or Storage Account Key together with Storage Account Name if Managed Identity is not used.")
-      .supportsExpressionLanguage(true).build();
+      .supportsExpressionLanguage(true)
```
szaszm, line 58: I don't think this supports EL. The EL extension only overrides the getProperty overloads that take a flow file for context, and the `onSchedule` of this class calls the overload without a flow file parameter.

libminifi/src/core/FlowConfiguration.cpp
```diff
@@ -174,4 +181,22 @@ std::shared_ptr<core::controller::ControllerServiceNode> FlowConfiguration::crea
   return controllerServicesNode;
 }
 
+std::string FlowConfiguration::decryptProperty(const std::string& encrypted_value) const {
+  static const utils::Regex is_encrypted{R"(enc\{.*\})"};
+  if (!utils::regexMatch(encrypted_value, is_encrypted)) {
```
szaszm, line 186: It may be premature optimization, but I'd avoid regex here, and use `starts_with("enc{")` and `ends_with('}')` instead.

extensions/azure/processors/AzureBlobStorageProcessorBase.h
fgerlits: `onSchedule` only does a sanity check that `SASToken` is non-empty; in `AzureBlobStorageProcessorBase::getAzureCredentialsFromProperties()` we use `getProperty()` with a flow file argument

libminifi/src/core/FlowConfiguration.cpp
fgerlits: good idea, fixed in e3c2ca4db93eed4e6ea18b49662bc16be1fb2477

libminifi/src/core/FlowConfiguration.cpp
```diff
@@ -174,4 +180,21 @@ std::shared_ptr<core::controller::ControllerServiceNode> FlowConfiguration::crea
...
+  static constexpr std::string_view WrapperBegin = "enc{";
+  static constexpr std::string_view WrapperEnd = "}";
+
+  if (!(encrypted_value.starts_with(WrapperBegin) && encrypted_value.ends_with(WrapperEnd))) {
+    // this is normal: sensitive properties come from the C2 server in cleartext over TLS
+    return encrypted_value;
+  }
+
+  const std::string unwrapped_value = encrypted_value.substr(WrapperBegin.size(), encrypted_value.length() - (WrapperBegin.size() + WrapperEnd.size()));
+  return sensitive_properties_encryptor_.decrypt(unwrapped_value);
```
szaszm, line 193: I think it would be better to change this snippet and the encryption providers to pass the data as `std::string_view` instead of `const std::string&`. If you think this change has its place in this PR, consider changing it. If not, maybe just this `decryptProperty` function can be changed to take `std::string_view`, it makes a copy of the value (substring) anyway.

fgerlits: I have changed this and a few of the related functions to take a `string_view` argument in 38b383ab2e4810505116a11c2a3c5e06b77f9243

martinzink:
I found a couple more properties we might wanna flag as sensitive

https://github.com/fgerlits/nifi-minifi-cpp/blob/MINIFICPP-2229_Encrypt-sensitive-properties-in-the-flow-definition/extensions/splunk/SplunkHECProcessor.h#L48-L51

https://github.com/fgerlits/nifi-minifi-cpp/blob/MINIFICPP-2229_Encrypt-sensitive-properties-in-the-flow-definition/extensions/elasticsearch/ElasticsearchCredentialsControllerService.h#L41-L44

https://github.com/fgerlits/nifi-minifi-cpp/blob/MINIFICPP-2229_Encrypt-sensitive-properties-in-the-flow-definition/extensions/sql/services/DatabaseService.h#L55-L58

fgerlits:
> I found a couple more properties we might wanna flag as sensitive

Thanks!  I have added them in d3228e93aa06a98189d08005ba4e4095ef007a04.

libminifi/src/core/flow/StructuredConfiguration.cpp
```diff
@@ -620,20 +628,26 @@ void StructuredConfiguration::parseRPGPort(const Node& port_node, core::ProcessG
...
     if (nodeVal) {
       Node propertiesNode = nodeVal["value"];
-      // must insert the sequence in differently.
-      const auto rawValueString = propertiesNode.getString().value();
-      logger_->log_debug("Found {}={}", property_name, rawValueString);
+      auto rawValueString = propertiesNode.getString().value();
+      if (myProp.isSensitive()) {
+        rawValueString = decryptProperty(rawValueString);
+      }
+      logger_->log_debug("Found property {}", property_name);
```
szaszm, line 641: Could this be changed in the future to allow decrypting properties that are _not_ marked as sensitive, without breaking backwards compatibility?
I'm thinking about a scenario where we accidentally marked a property as sensitive that's not really sensitive, so we make the property non-sensitive, but existing configs have the encrypted value after upgrade to this hypothetical future version.

I'm not requesting the change now, just want to make sure that should we ever make a property non-sensitive, we can also make this change then.

fgerlits: Yes, good point.  We could even make the change now; the one drawback is that no properties would be allowed to start with `enc{` and end with `}`.  I don't know if that is a big problem.  Alternatively, we could introduce a new (and possibly temporary) `wasSensitive()` flag if/when we change a property from sensitive to non-sensitive.

libminifi/include/core/flow/StructuredConfiguration.h
```diff
@@ -34,6 +34,8 @@
 #include "utils/file/FileSystem.h"
 #include "core/flow/Node.h"
 #include "FlowSchema.h"
+#include "rapidjson/document.h"
+#include "yaml-cpp/yaml.h"
```
adamdebreceni, line 38: although it would be nice not to have this this dependency here, I understand that it would require extra indirections/tests which would bloat this PR, could we create a ticket to tuck the clone/modification/save implementations away like we do with flow::Node?

adamdebreceni (approved):

martinzink (approved):

libminifi/src/core/flow/StructuredConfiguration.cpp
szaszm: I don't think it's a big problem to break properties that use the `enc{`/`}` wrappers on their own, but it's also not ideal. To limit the set of broken property values, we could check the format of the string inside the wrapper, and only act if it looks like what our encrypted values look.

Maybe it's best to leave it as is for now, and figure this out later if/when necessary.

libminifi/include/core/flow/StructuredConfiguration.h
fgerlits: I've created https://issues.apache.org/jira/browse/MINIFICPP-2281 for this

szaszm (approved):

libminifi/src/core/flow/StructuredConfiguration.cpp
```diff
@@ -900,4 +918,131 @@ void StructuredConfiguration::addNewId(const std::string& uuid) {
   }
 }
 
+void StructuredConfiguration::encryptSensitivePropertiesInYaml(YAML::Node property_yamls, const std::map<std::string, Property>& properties) const {
+  for (auto kv : property_yamls) {
```
lordgamez, line 922: Minor: this could be const ref

```suggestion
  for (const auto& kv : property_yamls) {
```

lordgamez (approved):

libminifi/src/core/flow/StructuredConfiguration.cpp
lordgamez: I see, thanks for the explanation

PR #1710, opened 2023-12-19
MINIFICPP-2275 Use utils::string instead of utils::StringUtils

fgerlits:
https://issues.apache.org/jira/browse/MINIFICPP-2275

It is probably easier to review the two original commits separately:
794916124ad5f3c69ab3451e4d96124b54d009e1 is, or should be, only `StringUtils` -> `string` replacements;
3afefe3438d1fbb2fc0f74e236fa22c7b4c48066 is clang-tidy fixes; this is where bugs (or solutions which could be improved) are more likely

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/test/unit/StagingQueueTests.cpp
+++ b/libminifi/test/unit/StagingQueueTests.cpp
@@ -92,7 +92,7 @@ TEST_CASE("Modify and overflow triggered automatic commit", "[TestStagingQueue4]
   }
   queue.modify([] (MockItem& item) {
     // a new item has been allocated
-    REQUIRE(item.data_ == "");
+    REQUIRE(item.data_.empty());
   });
   REQUIRE(queue.size() == 11);
   MockItem out;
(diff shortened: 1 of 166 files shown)


szaszm (approved):

martinzink (approved):

lordgamez (approved):

libminifi/test/unit/StagingQueueTests.cpp
```diff
@@ -92,7 +92,7 @@ TEST_CASE("Modify and overflow triggered automatic commit", "[TestStagingQueue4]
   }
   queue.modify([] (MockItem& item) {
     // a new item has been allocated
-    REQUIRE(item.data_ == "");
+    REQUIRE(item.data_.empty());
```
adamdebreceni, line 95: catch deconstructs the expression and logs the left-and right-hand side of an equality on failure, whereas with the change we only learn that it is not empty

fgerlits: True, this is one of the more questionable clang-tidy checks, but in this case, I think it makes sense:
* start out with an empty queue
* add something
* commit
* check that the (newly started) queue is empty.

So in this particular case, I would leave it as `.empty()`, but I'll change it if you think `== ""` would/could give us useful information.

fgerlits: True, this is one of the more questionable clang-tidy checks, but in this case, I think it makes sense:
* start out with an empty queue
* add something
* commit (explicitly or implicitly)
* check that the (newly started) queue is empty.

So in this particular case, I would leave it as `.empty()`, but I'll change it if you think `== ""` would/could give us useful information.

adamdebreceni (approved):

PR #1712, opened 2024-01-08
MINIFICPP-2276 Support FlowFileTransform NiFi Python processors

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2276

-----------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


docker/test/integration/features/environment.py
```diff
@@ -47,6 +47,13 @@ def before_scenario(context, scenario):
 
     logging.info("Integration test setup at {time:%H:%M:%S.%f}".format(time=datetime.datetime.now()))
     context.test = MiNiFi_integration_test(context=context, feature_id=context.feature_id)
+
+    if "USE_NIFI_PYTHON_PROCESSORS" in scenario.effective_tags:
+        if not context.image_store.is_conda_available_in_minifi_image() and context.image_store.get_minifi_image_python_version() < (3, 8, 1):
+            scenario.skip("NiFi Python processor tests use langchain library therefore Python 3.8.1 or later is needed in the MiNiFi container.")
```
szaszm, line 53: I'd rephrase this.
```suggestion
            scenario.skip("NiFi Python processor tests use langchain library which requires Python 3.8.1 or later.")
```

libminifi/include/core/ProcessContext.h
```diff
@@ -135,6 +135,10 @@ class ProcessContext : public controller::ControllerServiceLookup, public core::
     return getProperty(property.name, value);
   }
 
+  virtual bool getProperty(bool /*supports_expression_language*/, std::string_view property_name, std::string& value, const std::shared_ptr<FlowFile>& /*flow_file*/) {
+    return getProperty(property_name, value);
+  }
```
szaszm, line 140: Why do we need a new virtual overload? I'd prefer a single virtual function as an customization point, and possibly other non-static overloads to handle different use cases, to keep things simple.

libminifi/include/core/ConfigurableComponent.h
```diff
@@ -212,6 +212,8 @@ class ConfigurableComponent {
   // Dynamic properties
   std::map<std::string, Property> dynamic_properties_;
 
+  virtual Property* findProperty(const std::string& name) const;
+
```
szaszm, line 216: Doesn't getProperty already handle this use case?

extensions/python/tests/PythonManifestTests.cpp
```diff
@@ -101,7 +101,7 @@ TEST_CASE("Python processor's description is part of the manifest") {
     REQUIRE(getNode(MyPyProc->children, "type").value == "org.apache.nifi.minifi.processors.MyPyProc");
 
     auto& rels = getNode(MyPyProc->children, "supportedRelationships").children;
-    REQUIRE(rels.size() == 2);
+    REQUIRE(rels.size() == 3);
```
szaszm, line 104: What changed here?

extensions/python/PythonProcessor.h
```diff
@@ -37,7 +38,7 @@ class PythonProcessor {
 
   void setDescription(const std::string& desc);
 
-  void addProperty(const std::string& name, const std::string& description, const std::string& defaultvalue, bool required, bool el);
+  void addProperty(const std::string& name, const std::string& description, const std::optional<std::string>& defaultvalue, bool required, bool el, const std::optional<int64_t>& validator_value);
```
szaszm, line 41: What's a validator value?

libminifi/include/core/ProcessContext.h
lordgamez: We need an overload  with the `flow_file` parameter that only needs the `property_name` as that is the only one available and this version was already available as a private function. It is probably better to create a new `Property` constructor overload with these parameters, wrap these and pass it to the already available `getProperty` overload.

libminifi/include/core/ConfigurableComponent.h
lordgamez: In the python processors' properties are stored separately in the `python_properties_` vector. Because of this we need to override how to search for a property in getProperty and setProperty functions in the `ExecutePythonProcessor` class. Overriding this `findProperty` provides this, all other `getProperty` methods are either template functions or have an output reference parameter, which we cannot use for this.

extensions/python/tests/PythonManifestTests.cpp
lordgamez: I added the "original" default relationship to the python processors as the NiFi python processors have it.

extensions/python/PythonProcessor.h
lordgamez: I added a mapping of the NiFi validators to corresponding MiNiFi Property types and represented them with an int value, with the options listed in the PropertyTypeCode enum. On the python side it's called a validator, but maybe it would be better to change it property_type_code here and there as well.

docker/test/integration/features/environment.py
lordgamez: Updated in e8757f9c370b2ec92866a3731af1e6d9ae0edb74

libminifi/include/core/ProcessContext.h
lordgamez: Updated in e8757f9c370b2ec92866a3731af1e6d9ae0edb74

extensions/python/PythonProcessor.h
lordgamez: Updated in e8757f9c370b2ec92866a3731af1e6d9ae0edb74

libminifi/include/core/ConfigurableComponent.h
szaszm: I was under the impression that `getProperty` was already `virtual`, but apparently that's not the case, so this new customization point makes sense knowing this. Thanks!

extensions/python/ExecutePythonProcessor.h
```diff
@@ -118,7 +126,30 @@ class ExecutePythonProcessor : public core::Processor {
...
  private:
+  enum class PropertyTypeCode : int64_t {
+    INTEGER_TYPE = 0,
+    LONG_TYPE = 1,
+    BOOLEAN_TYPE = 2,
+    DATA_SIZE_TYPE = 3,
+    TIME_PERIOD_TYPE = 4,
+    NON_BLANK_TYPE = 5,
+    PORT_TYPE = 6
+  };
```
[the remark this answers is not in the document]
lordgamez: This is actually a reference to our `StandardPropertyTypes` found in MiNiFi defined in the `libminifi/include/core/PropertyType.h`, but we can remove the suffix and I suppose it's better to have it as part of the `core::PropertyType`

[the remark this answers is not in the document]
lordgamez: Updated in ee23065b91cd8ba88fb7db2dfeaa5c1cb31dae55

extensions/python/types/PyProcessContext.cpp
```diff
@@ -65,12 +66,31 @@ PyObject* PyProcessContext::getProperty(PyProcessContext* self, PyObject* args)
...
+    if (!context->getProperty(property_name, value)) {
+      Py_RETURN_NONE;
+    }
+  } else {
+    auto py_flow = reinterpret_cast<PyScriptFlowFile*>(script_flow_file);
+    const auto flow_file = py_flow->script_flow_file_.lock();
+    if (!flow_file) {
+      PyErr_SetString(PyExc_AttributeError, "tried reading FlowFile outside 'on_trigger'");
+      return nullptr;
+    }
```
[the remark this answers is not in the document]
lordgamez: Currently the held type of is `std::weak_ptr<core::FlowFile>` in both `PyScriptFlowFile` and `PyScriptFlowFile`, where this is the way the held objects are handled in both files. We can refactor this in the future, but for the time being I think we should stick with using the same handling here as well.

extensions/python/ExecutePythonProcessor.cpp
```diff
@@ -44,8 +45,7 @@ void ExecutePythonProcessor::initialize() {
 
   try {
     loadScript();
-  } catch(const std::runtime_error&) {
-    logger_->log_warn("Could not load python script while initializing. In case of non-native python processor this is normal and will be done in the schedule phase.");
```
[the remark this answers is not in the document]
lordgamez: Unfortunately the python processors are instantiated twice. We have to instantiate the processors defined in the flow, but also we instantiate all python processors when building the manifest to get the description and the processor properties. In the latter scenario the initialize is called twice, once before setting the script file in the `PythonObjectFactory.h` (to initialize the supported properties), and in that case the loadScript() will always fail, because the ScriptFile property value is not set yet. As this will always happen the log is useless here. This double initialization is a bit wonky and should be worked out sometime in the future.

extensions/python/PythonScriptEngine.cpp
```diff
@@ -180,4 +198,66 @@ void PythonScriptEngine::evaluateModuleImports() {
...
+      }
+
+      return PyList_Size(arg_list.get());
+    }();
+
+    if (num_args > 1) {
+      auto kwargs = OwnedDict::create();
+      auto value = OwnedObject(Py_None);
+      kwargs.put("jvm", value);
+      auto args = OwnedObject(PyTuple_New(0));
```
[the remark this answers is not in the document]
lordgamez: In NiFi there are scenarios when the `jvm` object is passed to the constructor, either as an arg or as a named kwarg and we only expect this one optional argument. We only want to pass this as part of the kwargs argument of the constructor call of the Python processor object, but before that we need to pass the positional args as well. As we do not want to pass any positional args, the python C API expects us to pass a zero length tuple which is a `PyTuple_New(0)` object to the `PyObject_Call` function in this case.

extensions/python/types/PyProcessContext.cpp
```diff
@@ -65,12 +66,31 @@ PyObject* PyProcessContext::getProperty(PyProcessContext* self, PyObject* args)
...
+  if (!script_flow_file) {
+    if (!context->getProperty(property_name, value)) {
+      Py_RETURN_NONE;
+    }
+  } else {
+    auto py_flow = reinterpret_cast<PyScriptFlowFile*>(script_flow_file);
+    const auto flow_file = py_flow->script_flow_file_.lock();
+    if (!flow_file) {
+      PyErr_SetString(PyExc_AttributeError, "tried reading FlowFile outside 'on_trigger'");
+      return nullptr;
```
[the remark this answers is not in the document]
lordgamez: Py_RETURN_NONE returns a Python object with the Python `None` value which can be a valid return value in some cases, while the returning `nullptr` indicates an error, and python throws an exception if nullptr is returned from the C API with the error set in the `PyErr_SetString`.

extensions/python/ExecutePythonProcessor.cpp
```diff
@@ -146,11 +151,38 @@ std::unique_ptr<PythonScriptEngine> ExecutePythonProcessor::createScriptEngine()
...
+    return const_cast<core::Property*>(&*it);
+  }
+
+  return nullptr;
+}
+
+std::map<std::string, core::Property> ExecutePythonProcessor::getProperties() const {
+  auto result = ConfigurableComponent::getProperties();
+
+  std::lock_guard<std::mutex> lock(configuration_mutex_);
```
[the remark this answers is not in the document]
lordgamez: You are right, it should be a separate mutex for the `python_properties_`, it didn't really make sense this way. Updated in 6593d9b63514dd2a3b2060b747680f37691abecd

extensions/python/PYTHON.md
```diff
@@ -106,20 +127,39 @@ class VaderSentiment(object):
...
+
+### Sentiment Analysis
 
 The SentimentAnalysis processor will perform a Vader Sentiment Analysis. This requires that you install nltk and VaderSentiment
 		pip install nltk
 		pip install VaderSentiment
+
+## Using NiFi Python Processors
+
+MiNiFi C++ supports the use of NiFi Python processors, that are inherited from the FlowFileTransform base class. To use these processors, you must copy the Python processor module under the nifi_python_processors directory located under the python directory (by default it can be found at ${minifi_root}/minifi-python/nifi_python_processors). To see how to write NiFi Python processors, please refer to the Python Developer Guide under the [Apache NiFi documentation](https://nifi.apache.org/documentation/v2/).
```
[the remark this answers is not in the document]
lordgamez: Updated in 6593d9b63514dd2a3b2060b747680f37691abecd

extensions/python/types/Types.h
```diff
@@ -160,7 +160,18 @@ class Long : public ReferenceHolder<reference_type> {
   }
 
   int64_t asInt64() {
-    return static_cast<int64_t>(PyLong_AsLongLong(this->ref_.get()));
+    auto long_value = PyLong_AsLongLong(this->ref_.get());
+    if (long_value == -1 && PyErr_Occurred()) {
+      throw PyException();
+    }
+    return static_cast<int64_t>(long_value);
```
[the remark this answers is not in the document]
lordgamez: Updated in 6593d9b63514dd2a3b2060b747680f37691abecd

libminifi/src/core/ConfigurableComponent.cpp
```diff
@@ -36,19 +36,28 @@ ConfigurableComponent::ConfigurableComponent()
 
 ConfigurableComponent::~ConfigurableComponent() = default;
 
+Property* ConfigurableComponent::findProperty(const std::string& name) const {
```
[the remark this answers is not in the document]
lordgamez: Updated in 6593d9b63514dd2a3b2060b747680f37691abecd

extensions/python/ExecutePythonProcessor.cpp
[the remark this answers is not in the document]
fgerlits: I see, thanks! Can you create a Jira for clearing this up later?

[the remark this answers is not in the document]
lordgamez: Sure, added jira ticket: https://issues.apache.org/jira/browse/MINIFICPP-2304

extensions/python/ExecutePythonProcessor.h
```diff
@@ -97,10 +99,8 @@ class ExecutePythonProcessor : public core::Processor {
...
 
-  void addProperty(const std::string &name, const std::string &description, const std::string &defaultvalue, bool required, bool el) {
-    python_properties_.emplace_back(
-        core::PropertyDefinitionBuilder<>::createProperty(name).withDefaultValue(defaultvalue).withDescription(description).isRequired(required).supportsExpressionLanguage(el).build());
-  }
+  void addProperty(const std::string &name, const std::string &description, const std::optional<std::string> &defaultvalue, bool required, bool el,
+      bool sensitive, const std::optional<int64_t>& property_type_code);
 
   const std::vector<core::Property> &getPythonProperties() const {
     return python_properties_;
```
fgerlits, line 106: I think we should acquire the mutex here, too, and return a copy.  The copy will happen anyway at the calling site in `PythonCreator::registerScriptDescription`; locking the mutex adds some overhead, but it is necessary.

lordgamez: Good point, updated in d3f47c57d21829d1715bd794f8c08a462f4fa361

fgerlits (approved):

adamdebreceni (approved):

extensions/python/PythonScriptEngine.cpp
```diff
@@ -180,4 +198,66 @@ void PythonScriptEngine::evaluateModuleImports() {
...
+      processor_instance_ = OwnedObject(PyObject_Call(python_class->get(), args.get(), kwargs.get()));
+    } else {
+      processor_instance_ = OwnedObject(PyObject_CallObject(python_class->get(), nullptr));
+    }
+
+    if (processor_instance_.get() == nullptr) {
+      throw PythonScriptException(PyException().what());
+    }
+
+    auto result = PyObject_SetAttrString(processor_instance_.get(), "logger", bindings_[std::string("log")]->get());
```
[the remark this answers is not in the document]
lordgamez: Good point about the interface, it should definitely be changed. Currently the issue is due to this ambiguity build error:
 
```
/home/ggyimesi/projects/nifi-minifi-cpp-fork/extensions/python/PythonScriptEngine.cpp:242:88: error: ambiguous overload for ‘operator[]’ (operand types are ‘org::apache::nifi::minifi::extensions::python::OwnedDict’ {aka ‘org::apache::nifi::minifi::extensions::python::Dict<org::apache::nifi::minifi::extensions::python::ReferenceType::OWNED>’} and ‘const char [4]’)
  242 |     auto result = PyObject_SetAttrString(processor_instance_.get(), "logger", bindings_["log"]->get());
      |                                                                                        ^
/home/ggyimesi/projects/nifi-minifi-cpp-fork/extensions/python/PythonScriptEngine.cpp:242:88: note: candidate: ‘operator[](long int, const char*)’ (built-in)
In file included from /home/ggyimesi/projects/nifi-minifi-cpp-fork/extensions/python/PythonBindings.h:24,
                 from /home/ggyimesi/projects/nifi-minifi-cpp-fork/extensions/python/PythonScriptEngine.h:20,
                 from /home/ggyimesi/projects/nifi-minifi-cpp-fork/extensions/python/PythonScriptEngine.cpp:21:
/home/ggyimesi/projects/nifi-minifi-cpp-fork/extensions/python/types/Types.h:290:36: note: candidate: ‘std::optional<org::apache::nifi::minifi::extensions::python::ObjectReference<org::apache::nifi::minifi::extensions::python::ReferenceType::BORROWED> > org::apache::nifi::minifi::extensions::python::Dict<reference_type>::operator[](std::string_view) [with org::apache::nifi::minifi::extensions::python::ReferenceType reference_type = org::apache::nifi::minifi::extensions::python::ReferenceType::OWNED; std::string_view = std::basic_string_view<char>]’
  290 |   std::optional<BorrowedReference> operator[](std::string_view key) {
      |                                    ^~~~~~~~
/home/ggyimesi/projects/nifi-minifi-cpp-fork/extensions/python/PythonScriptEngine.cpp:246:88: error: ambiguous overload for ‘operator[]’ (operand types are ‘org::apache::nifi::minifi::extensions::python::OwnedDict’ {aka ‘org::apache::nifi::minifi::extensions::python::Dict<org::apache::nifi::minifi::extensions::python::ReferenceType::OWNED>’} and ‘const char [12]’)
```

It doesn't need to be `std::string` but it still needs to explicitly define the parameter type due to this. I changed it to std::string_view which is the expected type currently in 6d382e80a43d2c4de48091dff2e91e7b951edfb1

szaszm (approved):

PR #1721, opened 2024-01-26
MINIFICPP-2277 Add virtualenv support for python processors

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2277

Depends on https://github.com/apache/nifi-minifi-cpp/pull/1712

---------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/docker/python-verify/conda.Dockerfile
+++ b/docker/python-verify/conda.Dockerfile
@@ -31,14 +31,12 @@ USER root
 RUN wget https://repo.anaconda.com/archive/Anaconda3-2023.09-0-Linux-x86_64.sh -P /tmp \
     && echo "6c8a4abb36fbb711dc055b7049a23bbfd61d356de9468b41c5140f8a11abd851 /tmp/Anaconda3-2023.09-0-Linux-x86_64.sh" | sha256sum -c \
     && bash /tmp/Anaconda3-2023.09-0-Linux-x86_64.sh -b -p /opt/conda  \
-    && chown -R ${USER}:${USER} /opt/conda \
-    && mkdir /home/${USER}  \
-    && chown -R ${USER}:${USER} /home/${USER}
+    && chown -R ${USER}:${USER} /opt/conda
 
 USER ${USER}
 
 RUN ${CONDA_HOME}/bin/conda init bash
-RUN ${CONDA_HOME}/bin/conda install langchain -c conda-forge
+RUN ${CONDA_HOME}/bin/conda install "langchain<=0.17.0" -c conda-forge
 
 WORKDIR ${MINIFI_HOME}
 
--- a/docker/test/integration/cluster/containers/MinifiContainer.py
+++ b/docker/test/integration/cluster/containers/MinifiContainer.py
@@ -36,8 +36,6 @@ class MinifiOptions:
         self.use_nifi_python_processors_with_system_python_packages_installed = False
         self.use_nifi_python_processors_with_virtualenv = False
         self.use_nifi_python_processors_with_virtualenv_packages_installed = False
-        self.create_python_virtualenv = False
-        self.install = False
         self.config_format = "json"
         self.use_flow_config_from_url = False
         self.set_ssl_context_properties = False
--- a/docker/test/integration/cluster/containers/MinifiContainer.py
+++ b/docker/test/integration/cluster/containers/MinifiContainer.py
@@ -33,7 +33,11 @@ class MinifiOptions:
         self.enable_prometheus = False
         self.enable_prometheus_with_ssl = False
         self.enable_sql = False
-        self.use_nifi_python_processors = False
+        self.use_nifi_python_processors_with_system_python_packages_installed = False
+        self.use_nifi_python_processors_with_virtualenv = False
+        self.use_nifi_python_processors_with_virtualenv_packages_installed = False
+        self.create_python_virtualenv = False
+        self.install = False
         self.config_format = "json"
         self.use_flow_config_from_url = False
         self.set_ssl_context_properties = False
@@ -148,6 +152,14 @@ class MinifiContainer(FlowContainer):
                 f.write("controller.socket.port=9998\n")
                 f.write("controller.socket.local.any.interface=false\n")
 
+            if self.options.use_nifi_python_processors_with_virtualenv:
+                f.write("nifi.python.virtualenv.directory=/opt/minifi/minifi-current/venv\n")
+            elif self.options.use_nifi_python_processors_with_virtualenv_packages_installed:
+                f.write("nifi.python.virtualenv.directory=/opt/minifi/minifi-current/venv-with-langchain\n")
+
+            if self.options.use_nifi_python_processors_with_virtualenv:
+                f.write("nifi.python.install.packages.automatically=true\n")
+
     def _setup_config(self):
         self._create_properties()
         if not self.options.use_flow_config_from_url:
@@ -166,7 +178,9 @@ class MinifiContainer(FlowContainer):
 
         if self.options.enable_sql:
             image = self.image_store.get_image('minifi-cpp-sql')
-        elif self.options.use_nifi_python_processors:
+        elif self.options.use_nifi_python_processors_with_system_python_packages_installed:
+            image = self.image_store.get_image('minifi-cpp-nifi-python-system-python-packages')
+        elif self.options.use_nifi_python_processors_with_virtualenv or self.options.use_nifi_python_processors_with_virtualenv_packages_installed:
             image = self.image_store.get_image('minifi-cpp-nifi-python')
         else:
             image = 'apacheminificpp:' + MinifiContainer.MINIFI_TAG_PREFIX + MinifiContainer.MINIFI_VERSION
--- a/encrypt-config/tests/ConfigFileEncryptorTests.cpp
+++ b/encrypt-config/tests/ConfigFileEncryptorTests.cpp
@@ -77,7 +77,7 @@ TEST_CASE("ConfigFileEncryptor can encrypt the sensitive properties", "[encrypt-
     uint32_t num_properties_encrypted = encryptSensitivePropertiesInFile(test_file, KEY);
 
     REQUIRE(num_properties_encrypted == 1);
-    REQUIRE(test_file.size() == 110);
+    REQUIRE(test_file.size() == 115);
     REQUIRE(check_encryption(test_file, Configuration::nifi_rest_api_password, original_password.length()));
 
     SECTION("calling encryptSensitiveValuesInMinifiProperties a second time does nothing") {
--- a/extensions/python/ExecutePythonProcessor.h
+++ b/extensions/python/ExecutePythonProcessor.h
@@ -35,10 +35,6 @@
 #include "PythonScriptEngine.h"
 #include "PythonScriptEngine.h"
 
-#if defined(__GNUC__) || defined(__GNUG__)
-#pragma GCC visibility push(hidden)
-#endif
-
 namespace org::apache::nifi::minifi::extensions::python::processors {
 
 class ExecutePythonProcessor : public core::Processor {
@@ -162,7 +158,3 @@ class ExecutePythonProcessor : public core::Processor {
 };
 
 }  // namespace org::apache::nifi::minifi::extensions::python::processors
-
-#if defined(__GNUC__) || defined(__GNUG__)
-#pragma GCC visibility pop
-#endif
--- a/extensions/python/PYTHON.md
+++ b/extensions/python/PYTHON.md
@@ -177,7 +177,7 @@ It is possible to set a virtualenv to be used by the Python processors in Apache
 
 ## Automatically install dependencies from requirements.txt files
 
-It is possible to automatically install the dependencies of the Python processors in the virtualenv defined in requirements.txt files. To enable this feature, the `nifi.python.install.packages.automatically` property must be set to true. If this property is set to true, then all requirements.txt files that appear under the MiNiFi Python directory and its subdirectories (defined by the `nifi.python.processor.dir` property) will be used to install the Python packages. If the `nifi.python.virtualenv.directory` property is set, the packages are installed in the virtualenv, otherwise this option is ignored. Due to install schema differences in different platforms, system level packages are expected to be installed manually by the user.
+It is possible to automatically install the dependencies of the Python processors defined in requirements.txt files into a virtualenv. To enable this feature, the `nifi.python.install.packages.automatically` property must be set to true, and the `nifi.python.virtualenv.directory` property must be set to a directory where a virtualenv either already exists, or it can be set up. In this case, all requirements.txt files that appear under the MiNiFi Python directory (defined by the `nifi.python.processor.dir` property) and its subdirectories will be used to install the Python packages into the given virtualenv. Due to install schema differences in different platforms, system level packages are expected to be installed manually by the user.
 
     # in minifi.properties
     nifi.python.install.packages.automatically=true
--- a/extensions/python/PYTHON.md
+++ b/extensions/python/PYTHON.md
@@ -25,6 +25,9 @@ This readme defines the configuration parameters to use ExecutePythonProcessor t
 - [Configuration](#configuration)
 - [Processors](#processors)
 - [Using NiFi Python Processors](#using-nifi-python-processors)
+- [Use Python processors from virtualenv](#use-python-processors-from-virtualenv)
+- [Automatically install dependencies from requirements.txt files](#automatically-install-dependencies-from-requirementstxt-files)
+- [Set python binary for virtualenv creation and package installation](#set-python-binary-for-virtualenv-creation-and-package-installation)
 
 
 ## Requirements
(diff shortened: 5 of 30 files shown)


docker/python-verify/conda.Dockerfile
```diff
@@ -31,14 +31,12 @@ USER root
...
-    && chown -R ${USER}:${USER} /opt/conda \
-    && mkdir /home/${USER}  \
-    && chown -R ${USER}:${USER} /home/${USER}
+    && chown -R ${USER}:${USER} /opt/conda
 
 USER ${USER}
 
 RUN ${CONDA_HOME}/bin/conda init bash
-RUN ${CONDA_HOME}/bin/conda install langchain -c conda-forge
+RUN ${CONDA_HOME}/bin/conda install "langchain<=0.17.0" -c conda-forge
```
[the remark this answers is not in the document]
lordgamez: Currently the latest version of langchain is 0.17.0 and as we run the python tests on multiple distributions and with virtualenvs or with conda in the CI, we run these tests with version 0.17.0 or bwloe. The tests with these versions are tested and passing, but the langchain library is rapidly changing so we don't want to use the latest version in the tests because they may break with a future version, so it is better to have the version fixed.

extensions/python/PythonScriptEngine.cpp
```diff
@@ -68,6 +67,73 @@ void initThreads() {
...
+  for (const auto& requirements_file_path : requirement_file_paths) {
+    logger->log_info("Installing python packages from the following requirements.txt file: {}", requirements_file_path.string());
+    std::string pip_command;
+#if WIN32
+    pip_command.append("\"").append((PythonConfigState::getInstance().virtualenv_path / "Scripts" / "activate.bat").string()).append("\" && ");
+#else
+    pip_command.append(". \"").append((PythonConfigState::getInstance().virtualenv_path / "bin" / "activate").string()).append("\" && ");
+#endif
+    pip_command.append("\"").append(PythonConfigState::getInstance().python_binary).append("\" -m pip install --no-cache-dir -r \"").append(requirements_file_path.string()).append("\"");
+    auto return_value = std::system(encapsulateCommandInQuotesIfNeeded(pip_command).c_str());
```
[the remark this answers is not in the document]
lordgamez: Yes it looks strange, but there is an issue with std::system on Windows that needs the additional quotes on the whole command to work correctly. If I remember well it passes the system command as a parameter to the `cmd.exe /C` command and the additional quotes are needed to parse the system command's arguments correctly.

docker/test/integration/cluster/containers/MinifiContainer.py
```diff
@@ -33,7 +33,11 @@ def __init__(self):
         self.enable_prometheus = False
         self.enable_prometheus_with_ssl = False
         self.enable_sql = False
-        self.use_nifi_python_processors = False
+        self.use_nifi_python_processors_with_system_python_packages_installed = False
+        self.use_nifi_python_processors_with_virtualenv = False
+        self.use_nifi_python_processors_with_virtualenv_packages_installed = False
+        self.create_python_virtualenv = False
+        self.install = False
```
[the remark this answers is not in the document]
lordgamez: They are unused, removed in 44654b11b315ed6010424cb72abb334b8b286377

extensions/python/PYTHON.md
```diff
@@ -155,10 +158,33 @@ In the flow configuration these Python processors can be referenced by their ful
...
+
+    # in minifi.properties
+    nifi.python.virtualenv.directory=${MINIFI_HOME}/minifi-python-env
+
+**NOTE:* Using different python versions for the system and the virtualenv is not supported. The virtualenv must be created using the same python version as the system python.
+
+## Automatically install dependencies from requirements.txt files
+
+It is possible to automatically install the dependencies of the Python processors in the virtualenv defined in requirements.txt files. To enable this feature, the `nifi.python.install.packages.automatically` property must be set to true. If this property is set to true, then all requirements.txt files that appear under the MiNiFi Python directory and its subdirectories (defined by the `nifi.python.processor.dir` property) will be used to install the Python packages. If the `nifi.python.virtualenv.directory` property is set, the packages are installed in the virtualenv, otherwise this option is ignored. Due to install schema differences in different platforms, system level packages are expected to be installed manually by the user.
```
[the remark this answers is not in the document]
lordgamez: You are right it's more clear this way, updated in 44654b11b315ed6010424cb72abb334b8b286377

extensions/python/PythonConfigState.h
```diff
@@ -0,0 +1,67 @@
...
+#include "ExecutePythonProcessor.h"
+#include "PythonObjectFactory.h"
+#include "agent/agent_version.h"
+#include "agent/build_description.h"
+#include "utils/file/FileUtils.h"
+#include "utils/StringUtils.h"
+#include "range/v3/algorithm.hpp"
+#include "properties/Configuration.h"
+#include "utils/file/FilePattern.h"
+#include "range/v3/view/filter.hpp"
```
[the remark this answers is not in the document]
lordgamez: Good point, left them in by mistake, removed the unneeded ones in 44654b11b315ed6010424cb72abb334b8b286377

extensions/python/PythonScriptEngine.cpp
```diff
@@ -68,6 +67,73 @@ void initThreads() {
...
+  if (auto binary = configuration->get(minifi::Configuration::nifi_python_env_setup_binary)) {
+    python_binary = *binary;
+  }
+  return python_binary;
+}
+
+void createVirtualEnvIfSpecified(const std::shared_ptr<Configure> &configuration) {
+  if (auto path = configuration->get(minifi::Configuration::nifi_python_virtualenv_directory)) {
+    PythonConfigState::getInstance().virtualenv_path = *path;
+    if (!std::filesystem::exists(PythonConfigState::getInstance().virtualenv_path) || !std::filesystem::is_empty(PythonConfigState::getInstance().virtualenv_path)) {
```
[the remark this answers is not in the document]
lordgamez: Good catch, we should only create a virtualenv if the directory is empty, probably didn't update the condition while refactoring and didn't retest this, updated in 44654b11b315ed6010424cb72abb334b8b286377

extensions/python/PythonConfigState.h
```diff
@@ -0,0 +1,50 @@
...
+  PythonConfigState& operator=(const PythonConfigState&) = delete;
+
+  bool isPackageInstallationNeeded() const {
+    return install_python_packages_automatically && !virtualenv_path.empty();
+  }
+
+  static PythonConfigState& getInstance() {
+    static PythonConfigState config;
+    return config;
+  }
```
szaszm, line 40: I'd prefer no new singletons. Would it be possible? Is there any reason for this being a singleton?

extensions/python/PythonScriptEngine.cpp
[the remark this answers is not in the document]
szaszm: Could you add this explanation into a code comment, maybe above the quoting function? I had the very same question when reading the code.

docker/python-verify/conda.Dockerfile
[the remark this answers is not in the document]
lordgamez: This is just for using fixed version of the library, to avoid breaking the tests with newer versions, see comment https://github.com/apache/nifi-minifi-cpp/pull/1721#discussion_r1513218719

encrypt-config/tests/ConfigFileEncryptorTests.cpp
```diff
@@ -77,7 +77,7 @@ TEST_CASE("ConfigFileEncryptor can encrypt the sensitive properties", "[encrypt-
     uint32_t num_properties_encrypted = encryptSensitivePropertiesInFile(test_file, KEY);
 
     REQUIRE(num_properties_encrypted == 1);
-    REQUIRE(test_file.size() == 110);
+    REQUIRE(test_file.size() == 115);
```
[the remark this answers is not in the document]
lordgamez: When the default minifi.properties file changes this tests always needs to be updated, as it checks the line count.

extensions/python/PythonConfigState.h
lordgamez: You are right, I also try to avoid them when possible. In this case the python configuration is read in the PythonCreator from the Configuration object, but has to be used in the PythonScriptEngine which is initialized in each python processor. Due to the interface we use to create the python processors (instantiating the ExecutePythonProcessor class from the PythonObjectFactory) there is no clean way currently to reach the `Configuration` object from the processors' scope. Even though it's better to avoid them I think this is still a cleaner solution than anything else I could come up with.

szaszm: Since PythonScriptEngine is created on a per-processor basis, we can't let it handle the creation of the shared execution environment. That should be extracted to a separate class, which could be owned by e.g. PythonCreator, setting up the virtualenv and packages, so that individual PythonScriptEngines can be created in that environment. PythonConfigState could be part of this new class, and PythonCreator could inject the Configuration object into it. What do you think about this?

lordgamez: It makes sense in the context of this PR, I'm not sure what was the original intention of the `PythonScriptEngine::initialize` static function, as it was previously called with and empty bode from the `PythonCreator`, so my thought was that it was supposed to implement the global python environment specific initialization, but it can be moved to the `PythonCreator` or a separate class.

The problem is that in the PR depending on this https://github.com/apache/nifi-minifi-cpp/pull/1727 we introduce the support for inline python dependency installation, which is separately done for every NiFi python processor initialization that is called from the `ExecutePythonProcessor`. We also only want to initialize the processors that are part of the minifi flow config in the future. The inline python dependency installation needs the information stored in the `PythonConfigState` that is why it needs to be reached from outside the `PythonCreator`

szaszm: I think it could be solved by returning or moving the list of dependencies from the python processor to the execution environment class, which can install them as needed. A callback could probably work.

extensions/python/ExecutePythonProcessor.h
```diff
@@ -35,10 +35,6 @@
 #include "PythonScriptEngine.h"
 #include "PythonScriptEngine.h"
 
-#if defined(__GNUC__) || defined(__GNUG__)
-#pragma GCC visibility push(hidden)
```
lordgamez, line 39: This was added due to pybind11 requiring -fvisibility=hidden, but as we do not use pybind11 anymore this can be removed

extensions/python/PythonConfigState.h
lordgamez: Extracted dependency installation to the `PythonDependencyInstaller` class and moved the dependency install to `PythonCreator`, to install dependencies while processing all the python processor script files in 5b20de2ffdc4d8f2bfd8907d1a47294440a11fc2

extensions/python/PythonDependencyInstaller.cpp
```diff
@@ -0,0 +1,167 @@
...
+    auto return_value = std::system(encapsulateCommandInQuotesIfNeeded(pip_command).c_str());
+    if (return_value != 0) {
+      throw PythonScriptException(fmt::format("The following command to install python packages failed: '{}'", pip_command));
+    }
+  }
+}
+
+void PythonDependencyInstaller::evalScript(std::string_view script) {
+  GlobalInterpreterLock gil;
+  const auto script_file = "# -*- coding: utf-8 -*-\n" + std::string(script);
```
szaszm, line 125: We could save a large string copy by using `utils::string::join_pack`. Even though the temporary string is an rvalue, the concatenated result can't reuse its buffer.

extensions/python/PythonScriptEngine.h
```diff
@@ -84,7 +48,7 @@ class PythonScriptEngine {
   PythonScriptEngine& operator=(const PythonScriptEngine& other) = delete;
   PythonScriptEngine& operator=(PythonScriptEngine&& other) = delete;
 
-  static void initialize() {}
+  static void initialize(const std::shared_ptr<Configure> &configuration, const std::shared_ptr<core::logging::Logger>& logger);
```
szaszm, line 51: Where is the implementation of this function?

extensions/python/PythonDependencyInstaller.h
```diff
@@ -0,0 +1,45 @@
...
+  explicit PythonDependencyInstaller(const std::shared_ptr<Configure> &configuration);
+  void installDependenciesFromRequirementsFiles() const;
+
+ private:
+  std::vector<std::filesystem::path> getRequirementsFilePaths() const;
+  void createVirtualEnvIfSpecified() const;
+  static void evalScript(std::string_view script);
+  void addVirtualenvToPath() const;
+
+  PythonConfigState config_state_;
```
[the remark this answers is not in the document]
lordgamez: Updated in 99359a0eda994fd5efa3095ebb76811ff1136f05

extensions/python/PythonDependencyInstaller.cpp
lordgamez: Good point, updated in 99359a0eda994fd5efa3095ebb76811ff1136f05

extensions/python/PythonScriptEngine.h
lordgamez: Good catch, it was removed, but the declaration remained, removed in 99359a0eda994fd5efa3095ebb76811ff1136f05

libminifi/src/Configuration.cpp
```diff
@@ -152,7 +152,11 @@ const std::unordered_map<std::string_view, gsl::not_null<const core::PropertyVal
   {Configuration::controller_socket_local_any_interface, gsl::make_not_null(&core::StandardPropertyTypes::BOOLEAN_TYPE)},
   {Configuration::controller_socket_host, gsl::make_not_null(&core::StandardPropertyTypes::VALID_TYPE)},
   {Configuration::controller_socket_port, gsl::make_not_null(&core::StandardPropertyTypes::PORT_TYPE)},
-  {Configuration::controller_ssl_context_service, gsl::make_not_null(&core::StandardPropertyTypes::VALID_TYPE)}
+  {Configuration::controller_ssl_context_service, gsl::make_not_null(&core::StandardPropertyTypes::VALID_TYPE)},
+  {Configuration::nifi_flow_file_repository_check_health, gsl::make_not_null(&core::StandardPropertyTypes::BOOLEAN_TYPE)},
```
szaszm, line 156: This line was missing?

szaszm (approved):

libminifi/src/Configuration.cpp
lordgamez: Yes, the `nifi_flow_file_repository_check_health` was missing from the property map.

extensions/python/PythonDependencyInstaller.cpp
```diff
@@ -0,0 +1,166 @@
...
+std::string getPythonBinary(const std::shared_ptr<Configure> &configuration) {
+#if WIN32
+  std::string python_binary_ = "python";
+#else
+  std::string python_binary_ = "python3";
+#endif
+  if (auto binary = configuration->get(minifi::Configuration::nifi_python_env_setup_binary)) {
+    python_binary_ = *binary;
+  }
+  return python_binary_;
```
adamdebreceni, line 38: accidental trailing underscore?

extensions/python/PythonDependencyInstaller.cpp
```diff
@@ -0,0 +1,166 @@
...
+#endif
+}
+
+}  // namespace
+
+PythonDependencyInstaller::PythonDependencyInstaller(const std::shared_ptr<Configure> &configuration) {
+  python_binary_ = getPythonBinary(configuration);
+  std::string automatic_install_str;
+  install_python_packages_automatically_ =
+    configuration->get(Configuration::nifi_python_install_packages_automatically, automatic_install_str) && utils::string::toBool(automatic_install_str).value_or(false);
```
adamdebreceni, line 57: we could use the monadic functions here

extensions/python/PythonDependencyInstaller.cpp
```diff
@@ -0,0 +1,166 @@
...
+  for (const auto& entry : std::filesystem::recursive_directory_iterator(std::filesystem::path{python_processor_dir_})) {
+    if (std::filesystem::is_regular_file(entry.path()) && entry.path().filename() == "requirements.txt") {
+      paths.push_back(entry.path());
+    }
+  }
+  return paths;
+}
+
+void PythonDependencyInstaller::createVirtualEnvIfSpecified() const {
+  if (virtualenv_path_.empty()) {
```
adamdebreceni, line 88: should we log a warning if the virtualenv_path is empty but the automatic install is requested?

extensions/python/PythonDependencyInstaller.cpp
lordgamez: Good idea, updated in 33e8770038e823678763c9bef60cc656c07ba16b

extensions/python/PythonDependencyInstaller.cpp
lordgamez: Sounds good, added warning in 33e8770038e823678763c9bef60cc656c07ba16b

adamdebreceni (approved):

PR #1722, opened 2024-02-06
MINIFICPP-2278 Add custom relationship support for python processors

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2278

Depends on https://github.com/apache/nifi-minifi-cpp/pull/1712

---------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/docker/test/integration/resources/python/RotatingForwarder.py
+++ b/docker/test/integration/resources/python/RotatingForwarder.py
@@ -1,0 +1,33 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from nifiapi.flowfiletransform import FlowFileTransform, FlowFileTransformResult
+
+
+class RotatingForwarder(FlowFileTransform):
+    """
+    Forwards flow files to a different relationship each time it is called in a round robin manner.
+    """
+    def __init__(self, **kwargs):
+        self.relationship_index = 0
+        self.relationships = ["first", "second", "third", "fourth"]
+
+    def transform(self, context, flowFile):
+        content = flowFile.getContentsAsBytes().decode()
+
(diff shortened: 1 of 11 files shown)


docker/test/integration/resources/python/RotatingForwarder.py
fgerlits, line 1: This is not the best example of dynamic relationships, since all the relationships are hard-coded into the processor.  I understand that even this was not supported until now, but in the rest of the code (eg. RouteText, RouteOnAttribute), "supports dynamic relationships" means the user can add (outgoing) relationships by configuring the processor's properties.

lordgamez: The goal of this PR is to support the same dynamic relationships that NiFi has in their python processors. Previously python processors only supported the predefined "success", "failure" and "original" relationships predefined in the ExecutePythonProcessor processor and bound to the REL_SUCCESS, REL_FAILURE and REL_ORIGINAL python objects. In NiFi the result of the `transform` method is a `FlowFileTransformResult` object where the relationship can be any custom string that represents a relationship. This is the dynamic relationship we would like to support, which may differ from the dynamic relationship concept previously used in MiNiFi C++.

fgerlits: Should we give it a different name?  It's a bit confusing that we use "dynamic relationships" in two different meanings.

lordgamez: I would rename it to "custom relationship" and set the `SupportsDynamicRelationships` property to false, what do you think?

fgerlits: yes, that sounds good

lordgamez: Updated in a70fe1b486262582113b52761724b197941e4ab9

fgerlits (approved):

adamdebreceni (approved):

extensions/python/types/PyProcessSession.cpp
```diff
@@ -308,6 +321,30 @@ PyObject* PyProcessSessionObject::transfer(PyProcessSessionObject* self, PyObjec
...
+  }
+
+  const auto flow_file = reinterpret_cast<PyScriptFlowFile*>(script_flow_file)->script_flow_file_.lock();
+  if (!flow_file) {
+    PyErr_SetString(PyExc_AttributeError, "tried reading FlowFile outside 'on_trigger'");
+    Py_RETURN_NONE;
+  }
+
+  BorrowedStr name = BorrowedStr::fromTuple(args, 0);
+  session->transferToCustomRelationship(flow_file, relationship_name);
```
szaszm, line 344: It may be worth null-checking and maybe empty-checking `relationship_name`, what do you think?

lordgamez: I agree, it's worth checking here just in case, updated in 227877f0eb542585476c73314bd232c91964aa50

szaszm (approved):

PR #1725, opened 2024-02-09
MINIFICPP-2283 Create tool to encrypt sensitive properties in config.yml

fgerlits:
https://issues.apache.org/jira/browse/MINIFICPP-2283

I have also upgraded argparse from 2.9 to 3.0 so I can use the new `.flag()` syntax.  (After the latest edit, we no longer use `.flag()`, but I kept the upgrade because why not.)

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/encrypt-config/EncryptConfig.h
+++ b/encrypt-config/EncryptConfig.h
@@ -37,7 +37,6 @@ class EncryptConfig {
   [[nodiscard]] std::filesystem::path bootstrapFilePath() const;
   [[nodiscard]] std::filesystem::path propertiesFilePath() const;
   [[nodiscard]] std::filesystem::path flowConfigPath() const;
-  static std::string flowConfigContent(const std::filesystem::path& config_path);
 
   [[nodiscard]] EncryptionKeys getEncryptionKeys(std::string_view property_name) const;
   [[nodiscard]] std::string hexDecodeAndValidateKey(const std::string& key, const std::string& key_name) const;
--- a/encrypt-config/EncryptConfig.h
+++ b/encrypt-config/EncryptConfig.h
@@ -21,11 +21,7 @@
 
 #include "Utils.h"
 
-namespace org {
-namespace apache {
-namespace nifi {
-namespace minifi {
-namespace encrypt_config {
+namespace org::apache::nifi::minifi::encrypt_config {
 
 class EncryptConfig {
  public:
@@ -35,26 +31,28 @@ class EncryptConfig {
   };
 
   explicit EncryptConfig(const std::string& minifi_home);
-  EncryptionType encryptSensitiveProperties() const;
+  [[nodiscard]] EncryptionType encryptionType() const;
 
-  void encryptFlowConfig() const;
+  void encryptSensitiveValuesInMinifiProperties() const;
+  void encryptSensitiveValuesInFlowConfig() const;
+  void encryptFlowConfigBlob() const;
 
  private:
-  std::filesystem::path bootstrapFilePath() const;
-  std::filesystem::path propertiesFilePath() const;
+  [[nodiscard]] std::filesystem::path bootstrapFilePath() const;
+  [[nodiscard]] std::filesystem::path propertiesFilePath() const;
+  [[nodiscard]] std::filesystem::path flowConfigPath() const;
+  static std::string flowConfigContent(const std::filesystem::path& config_path);
 
-  EncryptionKeys getEncryptionKeys() const;
-  std::string hexDecodeAndValidateKey(const std::string& key, const std::string& key_name) const;
-  void writeEncryptionKeyToBootstrapFile(const utils::crypto::Bytes& encryption_key) const;
+  [[nodiscard]] EncryptionKeys getEncryptionKeys(std::string_view property_name) const;
+  [[nodiscard]] std::string hexDecodeAndValidateKey(const std::string& key, const std::string& key_name) const;
+  void writeEncryptionKeyToBootstrapFile(const std::string& encryption_key_name, const utils::crypto::Bytes& encryption_key) const;
 
-  void encryptSensitiveProperties(const EncryptionKeys& keys) const;
+  void encryptSensitiveValuesInMinifiProperties(const EncryptionKeys& keys) const;
+  void encryptSensitiveValuesInFlowConfig(const EncryptionKeys& keys) const;
 
   const std::filesystem::path minifi_home_;
   EncryptionKeys keys_;
+  EncryptionKeys sensitive_properties_keys_;
 };
 
-}  // namespace encrypt_config
-}  // namespace minifi
-}  // namespace nifi
-}  // namespace apache
-}  // namespace org
+}  // namespace org::apache::nifi::minifi::encrypt_config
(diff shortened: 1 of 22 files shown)


encrypt-config/EncryptConfig.h
```diff
@@ -21,40 +21,29 @@
...
+  [[nodiscard]] bool isReencrypting() const;
 
-  EncryptionKeys getEncryptionKeys() const;
-  std::string hexDecodeAndValidateKey(const std::string& key, const std::string& key_name) const;
-  void writeEncryptionKeyToBootstrapFile(const utils::crypto::Bytes& encryption_key) const;
+ private:
+  [[nodiscard]] std::filesystem::path bootstrapFilePath() const;
+  [[nodiscard]] std::filesystem::path propertiesFilePath() const;
+  [[nodiscard]] std::filesystem::path flowConfigPath() const;
+  static std::string flowConfigContent(const std::filesystem::path& config_path);
```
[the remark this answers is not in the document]
fgerlits: yes, good idea -- done in 195fdc307610b4dcdf40d76ad83345f240382f2e

encrypt-config/FlowConfigEncryptor.cpp
```diff
@@ -0,0 +1,173 @@
...
+#include "Defaults.h"
+#include "Utils.h"
+#include "utils/file/FileSystem.h"
+#include "utils/Id.h"
+
+namespace minifi = org::apache::nifi::minifi;
+
+namespace {
+enum class Type {
+  Processor, ControllerService
```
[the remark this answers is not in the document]
fgerlits: agreed, it looks better: b62a3c5b77b4a4acefcd806a81fd7d2e8aa20be4

lordgamez (approved):

adamdebreceni (approved):

libminifi/include/core/ProcessGroup.h
```diff
@@ -205,6 +205,8 @@ class ProcessGroup : public CoreComponent {
    */
   std::shared_ptr<core::controller::ControllerServiceNode> findControllerService(const std::string &nodeId) const;
 
+  std::vector<std::shared_ptr<core::controller::ControllerServiceNode>> getAllControllerServices() const;
```
szaszm, line 208: Does this really need to give ownership to the callers?

libminifi/src/core/json/JsonFlowSerializer.cpp
```diff
@@ -47,14 +47,21 @@ void JsonFlowSerializer::encryptSensitiveProperties(rapidjson::Value &property_j
...
+      const std::string_view value_sv = component_overrides.contains(name) ? component_overrides.at(name) : std::string_view{value.GetString(), value.GetStringLength()};
       const std::string encrypted_value = utils::crypto::property_encryption::encrypt(value_sv, encryption_provider);
       value.SetString(encrypted_value.c_str(), encrypted_value.size(), alloc);
     }
+    component_overrides.erase(name);
+  }
+
+  for (const auto& [name, value] : component_overrides) {
+    const std::string encrypted_value = utils::crypto::property_encryption::encrypt(value, encryption_provider);
+    property_jsons.AddMember(rapidjson::Value(name, alloc), rapidjson::Value(encrypted_value, alloc), alloc);
```
szaszm, line 59: What's the idea behind overrides? I see that the override values are used in plaintext to override existing properties in the flow definition, but if they were not used, then they are used encrypted. I don't understand why it works that way.

libminifi/include/core/ProcessGroup.h
fgerlits: do you mean the vector should contain `const ControllerServiceNode*`s?

libminifi/src/core/json/JsonFlowSerializer.cpp
fgerlits: Encrypted or not is a red herring. The property value in the JSON (or YAML) may or may not be already encrypted; if it isn't, then it will be encrypted now.  Overrides are used by the `encrypt-config` tool to replace the existing property value with a new value given by the user.  "Override" may not be the best name for this; do you have a suggestion for a better name?

fgerlits: Also, this part of the code handles the case when the property given by the user to the `encrypt-config` tool does not exist in the JSON, yet.  In this case, we add the property to the JSON flow configuration.

libminifi/include/core/ProcessGroup.h
szaszm: That would be an improvement. Optionally, they could be wrapped in gsl::not_null. Whether a const refernce is enough for all consumers also needs to be considered, but if it is, then less unnecessary exposure is better. Currently it's pointers to mutable objects.

libminifi/src/core/json/JsonFlowSerializer.cpp
szaszm: Overrides as a name is fine for me, but their expected format needs to be clearly defined.
I think we should standardize whether the values in the overrides are already encrypted (if sensitive) or not encrypted. I'm not sure how we currently decide about encryption based on this control flow.

szaszm: My bad, I missed the encrypt call on line 51 at first.

libminifi/include/core/ProcessGroup.h
fgerlits: I have changed this in 929aafecc10ae3d41ce8623d062de58c4f785abe

Now the interface is a bit of a mess, with a mix of return types; I have created a Jira to fix this: https://issues.apache.org/jira/browse/MINIFICPP-2312

szaszm: I'm assuming they can't be null, so I think it would be nicer to wrap the pointers in not_null. Or have runtime checks more graceful than gsl_Expects if they can be null.

fgerlits: They can't be null, but I think `not_null` should be added to the container in `ControllerServiceMap`, so I would prefer to do this as part of MINIFICPP-2312.

martinzink (approved):

encrypt-config/FlowConfigEncryptor.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+  bool encrypt_whole_flow_config_file = (configure->get(minifi::Configure::nifi_flow_configuration_encrypt) | minifi::utils::andThen(minifi::utils::string::toBool)).value_or(false);
+  auto encryptor = encrypt_whole_flow_config_file ? minifi::utils::crypto::EncryptionProvider::create(minifi_home) : std::nullopt;
+  auto filesystem = std::make_shared<minifi::utils::file::FileSystem>(encrypt_whole_flow_config_file, encryptor);
+
+  minifi::core::extension::ExtensionManager::get().initialize(configure);
+
+  minifi::core::flow::AdaptiveConfiguration adaptive_configuration{minifi::core::ConfigurationContext{
+      .flow_file_repo = minifi::core::createRepository("flowfilerepository"),
+      .content_repo = std::make_shared<minifi::core::repository::VolatileContentRepository>(),
```
szaszm, line 112: Would this work with passing null or NoopRepository? I don't think we need repositories for modifying the config.

fgerlits: Good point; I've removed the repository creation in 0506380be319a185920c571ee0737556d86c89bd.

szaszm (approved):

PR #1734, opened 2024-02-28
MINIFICPP-2203 Add support for building Windows MSI without any redistributables included

lordgamez:
- Also remove unused FORCE_WINDOWS build option

------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


cmake/MiNiFiOptions.cmake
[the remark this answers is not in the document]
lordgamez: If everything is OFF it works the same way as currently, installing the VS redistributables, I didn't want to change the current behavior. I also think it's better to not include these by default, if we can all agree to this I can change the behavior.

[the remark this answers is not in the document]
lordgamez: I discussed with @fgerlits as well and we agreed that this default would be better and would probably have no impact on any user. I changed the default in 236dcf321054fd163ecbbb0ba3ffba9c31412705

cmake/MiNiFiOptions.cmake
```diff
@@ -82,6 +80,8 @@ list(APPEND STRICT_GSL_CHECKS_Values AUDIT ON DEBUG_ONLY OFF)
 set_property(CACHE STRICT_GSL_CHECKS PROPERTY STRINGS ${STRICT_GSL_CHECKS_Values})
 
 if (WIN32)
+    add_minifi_option(INSTALLER_MERGE_MODULES "Creates installer with merge modules" OFF)
+    add_minifi_option(INSTALLER_WITH_VC_REDISTRIBUTABLES "Creates installer with Visual C++ redistributables included" OFF)
```
szaszm, line 84: The naming is confusing, and not specific enough. Merge modules and bundled DLLs are both Visual C++ redistributable runtimes (runtime meaning library or DLL in this case), they're just different ways of installing the same DLLs of the C and C++ standard libraries. Under the old version, if the user didn't ask for merge modules, they got bundled DLLs.

There is also the Universal C Runtime (UCRT) DLLs that are not part of the same Visual C++ redistributable runtimes package, but they are also standard library DLLs, they also fall under the same license terms, and are not redistributable under the Apache License. If someone wants to NOT include the Visual C++ redistributable runtime, then they probably want a package free of Microsoft blobs, and they don't want to include the UCRT either.

The previous version was fine, as it treated both kinds of redistributables under the same banner, disabling them altogether. This new version seems to treat the Visual C++ redistributable runtime package separately from the UCRT, and I don't see the rationale behind this.

lordgamez: I agree the naming should be changed, to be more clear, and clarify which option bundles what artifacts exactly. It also needs to be specified if a bundle falls under a different license.

On the other hand I would keep all three bundle options separate. The rationale behind this version is it's easier and more direct to specify a single option for each installer bundle option instead of combining build options like `INCLUDE_MS_BLOBS` + `USE_UCRT_DLLS`. So I would keep them separately like `INCLUDE_UCRT_IN_INSTALLER`, `INCLUDE_VC_REDIST_DLLS`, and `INCLUDE_VC_REDIST_MERGE_MODULES`. This way you can directly specify which libraries you want to include in the installer, and by not specifying anything, no Microsoft blobs will be included. What are your thoughts on this?

martinzink, line 84: Could we add the MINIFI_ prefix to these new options?
(In my opinion we should add that to every option thats ours so it doesnt conflict with thirdparty options and it makes it easier to filter out thirdparty related options (e.g. python bootstrap).)

So adding it to new flags should be good way to start this transition.

szaszm: This sounds good to me

lordgamez: Updated the flag names and descriptions in b2da5ed65397525433093fa654edfe0d7836ca13 also making them mutually exclusive

lordgamez: Updated in b2da5ed65397525433093fa654edfe0d7836ca13

martinzink (approved):

CMakeLists.txt
```diff
@@ -493,20 +504,21 @@ if(WIN32)
...
-        else()
-            message(FATAL_ERROR "Could not determine architecture, CMAKE_SIZEOF_VOID_P is unexpected: ${CMAKE_SIZEOF_VOID_P}")
         endif()
-        set(CPACK_WIX_TEMPLATE "${CMAKE_CURRENT_SOURCE_DIR}/msi/WixWinMergeModules.wsi")
-    else()
-        message("Creating installer with redistributables")
+
+        file(READ "${CMAKE_CURRENT_SOURCE_DIR}/msi/MergeModulesFeature.xml" WIX_EXTRA_FEATURES)
+        configure_file("${CMAKE_CURRENT_SOURCE_DIR}/msi/WixWin.wsi.in" "${CMAKE_CURRENT_SOURCE_DIR}/msi/WixWin.wsi")
+        set(CPACK_WIX_TEMPLATE "${CMAKE_CURRENT_SOURCE_DIR}/msi/WixWin.wsi")
```
fgerlits, line 519: these two lines are the same in every branch, so they could be moved to after the `if` block

lordgamez: Good point, updated in b58c443735e7d065d37619945f6fe88e6fc534ef

fgerlits (approved):

szaszm (approved):

PR #1739, opened 2024-03-01
MINIFICPP-2282 Support re-encryption of sensitive properties

fgerlits:
Added a new `--re-encrypt` option to the `flow-config` action of the `encrypt-config` binary, which can decrypt the sensitive properties using the `.old` key and re-encrypt them with a new key (either supported by the user or auto-generated).

https://issues.apache.org/jira/browse/MINIFICPP-2282

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/encrypt-config/EncryptConfigMain.cpp
+++ b/encrypt-config/EncryptConfigMain.cpp
@@ -60,15 +60,21 @@ int main(int argc, char* argv[]) try {
   }
 
   EncryptConfig encrypt_config{argument_parser.get("-m")};
+
   std::string operation = argument_parser.get("operation");
+  const auto re_encrypt = argument_parser.get<bool>("--re-encrypt");
+  const auto component_id = argument_parser.present("--component-id");
+  const auto property_name = argument_parser.present("--property-name");
+  const auto property_value = argument_parser.present("--property-value");
+
+  if (operation != OPERATION_FLOW_CONFIG && (re_encrypt || component_id || property_name || property_value)) {
+    std::cerr << "Unsupported option for operation '" << operation << "'\n\n" << argument_parser;
+    return 5;
+  }
 
   if (operation == OPERATION_MINIFI_PROPERTIES) {
     encrypt_config.encryptSensitiveValuesInMinifiProperties();
   } else if (operation == OPERATION_FLOW_CONFIG) {
-    auto re_encrypt = argument_parser.get<bool>("--re-encrypt");
-    auto component_id = argument_parser.present("--component-id");
-    auto property_name = argument_parser.present("--property-name");
-    auto property_value = argument_parser.present("--property-value");
     encrypt_config.encryptSensitiveValuesInFlowConfig(re_encrypt, component_id, property_name, property_value);
   } else if (operation == OPERATION_WHOLE_FLOW_CONFIG_FILE) {
     encrypt_config.encryptWholeFlowConfigFile();
--- a/encrypt-config/EncryptConfigMain.cpp
+++ b/encrypt-config/EncryptConfigMain.cpp
@@ -48,6 +48,9 @@ int main(int argc, char* argv[]) try {
   argument_parser.add_argument("--property-value")
       .metavar("VALUE")
       .help(minifi::utils::string::join_pack("The new value of the sensitive property (", OPERATION_FLOW_CONFIG, " only)"));
+  argument_parser.add_argument("--re-encrypt")
+      .flag()
+      .help(minifi::utils::string::join_pack("Decrypt all properties with the old key and re-encrypt them with a new key (", OPERATION_FLOW_CONFIG, " only)"));
 
   try {
     argument_parser.parse_args(argc, argv);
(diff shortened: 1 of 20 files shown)


encrypt-config/FlowConfigEncryptor.cpp
```diff
@@ -61,114 +62,150 @@ std::vector<SensitiveProperty> listSensitiveProperties(const minifi::core::Proce
...
   configure->loadConfigureFile(DEFAULT_NIFI_PROPERTIES_FILE);
 
-  bool encrypt_whole_flow_config_file = (configure->get(minifi::Configure::nifi_flow_configuration_encrypt) | minifi::utils::andThen(minifi::utils::string::toBool)).value_or(false);
-  auto encryptor = encrypt_whole_flow_config_file ? minifi::utils::crypto::EncryptionProvider::create(minifi_home) : std::nullopt;
-  auto filesystem = std::make_shared<minifi::utils::file::FileSystem>(encrypt_whole_flow_config_file, encryptor);
+  bool encrypt_whole_flow_config_file = (configure->get(Configure::nifi_flow_configuration_encrypt) | utils::andThen(utils::string::toBool)).value_or(false);
+  auto whole_file_encryptor = encrypt_whole_flow_config_file ? utils::crypto::EncryptionProvider::create(minifi_home) : std::nullopt;
+  auto filesystem = std::make_shared<utils::file::FileSystem>(encrypt_whole_flow_config_file, whole_file_encryptor);
+
+  auto sensitive_properties_encryptor = is_re_encrypting ?
```
szaszm, line 167: I think calling it decryptor would be more descriptive, since the old key is used for decrypting, before reencrypting with the new key.

fgerlits: makes sense, changed in 762c4dfbaca0d4f7c9a40b3799017d8dfe48161b (although line 179 is going to look a bit weird now)

encrypt-config/EncryptConfigMain.cpp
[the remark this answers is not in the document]
fgerlits: it is set to an encrypted empty string, but I agree that we should not add new properties during re-encryption

encrypt-config/EncryptConfigMain.cpp
```diff
@@ -62,18 +65,19 @@ int main(int argc, char* argv[]) try {
   if (operation == OPERATION_MINIFI_PROPERTIES) {
     encrypt_config.encryptSensitiveValuesInMinifiProperties();
   } else if (operation == OPERATION_FLOW_CONFIG) {
+    auto re_encrypt = argument_parser.get<bool>("--re-encrypt");
     auto component_id = argument_parser.present("--component-id");
     auto property_name = argument_parser.present("--property-name");
     auto property_value = argument_parser.present("--property-value");
```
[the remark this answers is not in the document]
fgerlits: fixed in 9ab4f07115b554b646c601253ecd0e16bc8a5e71

encrypt-config/EncryptConfigMain.cpp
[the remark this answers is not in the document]
fgerlits: fixed in e8ca161921504becfb0044d1e86828bcb4a3ce98

libminifi/include/core/json/JsonFlowSerializer.h
```diff
@@ -26,12 +26,12 @@ class JsonFlowSerializer : public core::flow::FlowSerializer {
...
 
   [[nodiscard]] std::string serialize(const core::ProcessGroup& process_group, const core::flow::FlowSchema& schema, const utils::crypto::EncryptionProvider& encryption_provider,
-      const std::unordered_map<utils::Identifier, std::unordered_map<std::string, std::string>>& overrides) const override;
+      const core::flow::Overrides& overrides) const override;
 
  private:
   void encryptSensitiveProperties(rapidjson::Value& property_jsons, rapidjson::Document::AllocatorType& alloc,
       const std::map<std::string, Property>& properties, const utils::crypto::EncryptionProvider& encryption_provider,
-      std::unordered_map<std::string, std::string> component_overrides) const;
+      const core::flow::Overrides& overrides, const utils::Identifier& component_id) const;
```
szaszm, line 34: This used to only take the overrides for the specific component, but not it gets all of them. Is this changed intentionally? Why?

fgerlits:
There is a problem with this pull request: if you run `encrypt-config flow-config --re-encrypt` twice, then it encrypts the already-encrypted properties a second time, making them unusable. I'll put the PR in draft mode until I fix this.

libminifi/include/core/json/JsonFlowSerializer.h
fgerlits: Yes, it is intentional: it seemed simpler to just pass along a (const) reference to the `Overrides` object along with the processor/controller service ID, so the function can call the `get` functions on the object with the ID.

The other options I considered were to either pass in `required_component_overrides` and `optional_component_overrides` separately, or create a new type which contains these. I didn't like either of these options. Do you have a third idea which is better than the current code?

szaszm: I was thinking of the second one, creating a new type with the component overrides, and maybe change Overrides to map component id to this new type. (Maybe change back to raw unordered_map?)
Adding a "required" flag to the value could work, too, to keep the underlying type simpler. I'm not a fan of widening the dependency, but on the other hand, this version looks simpler than the alternatives.

I've shared my thoughts, but leave the final decision to you: go with whichever option you like the most, after considering the arguments for each option.

fgerlits: fixed in dd8b12e336e34acbd33de3b7bc41c66431bd50ae

adamdebreceni (approved):

szaszm (approved):

lordgamez (approved):

PR #1749, opened 2024-03-28
MINIFICPP-2251 Make OPENSSL and CURL mandatory

lordgamez:
MINIFICPP-1872 Unify test utilities

- Remove OpenSSL and Curl options and make them part of the libminifi core
- Remove http extension and make it part of libminifi core and InvokeHTTP processor part of standard-processors
- Move HTTP tests as part of libminifi integration tests
- Extract test library in libminifi, with separate integration and unit test library
- Make test utilities part of the test library
- Move all tests to their proper unittest or integrationtest folder in libminifi
- Convert integration tests to use the catch2 framework and remove executable parametarization

https://issues.apache.org/jira/browse/MINIFICPP-2251
https://issues.apache.org/jira/browse/MINIFICPP-1872

### For all changes:
     in the commit message?

### For code changes:

### Note:


README.md
```diff
@@ -185,7 +184,7 @@ On all distributions please use -DUSE_SHARED_LIBS=OFF to statically link zlib, l
 #### Libraries
 * libuuid
 * librocksdb (built and statically linked)
-* libcurl-openssl (If not available or desired, NSS will be used)
+* libcurl-openssl
```
szaszm, line 187:
```suggestion
* libcurl-openssl (built and statically linked)
```

arch.sh
```diff
@@ -31,10 +31,9 @@ bootstrap_compiler(){
     sudo pacman -S --noconfirm gcc
 }
 build_deps(){
-    COMMAND="sudo pacman -S --noconfirm --needed cmake zlib openssl util-linux make patch which pkgconf"
+    COMMAND="sudo pacman -S --noconfirm --needed cmake zlib openssl util-linux make patch which pkgconf perl bzip2 curl"
```
szaszm, line 34: Doesn't OpenSSL also need NASM?

libminifi/test/unit/SchedulingAgentTests.cpp
```diff
@@ -63,8 +64,8 @@ class SchedulingAgentTestFixture {
     count_proc_->setScheduledState(core::RUNNING);
 
 #ifdef WIN32
-    utils::dateSetInstall(TZ_DATA_DIR);
     date::set_install(TZ_DATA_DIR);
+    minifi::utils::timeutils::dateSetInstall(TZ_DATA_DIR);
```
szaszm, line 68: Why was this reordered?

libminifi/CMakeLists.txt
```diff
@@ -78,17 +76,17 @@ endif()
...
-if (MINIFI_OPENSSL)
-    list(APPEND LIBMINIFI_LIBRARIES OpenSSL::Crypto OpenSSL::SSL)
-endif()
 if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9)
     list(APPEND LIBMINIFI_LIBRARIES stdc++fs)
 endif()
 target_link_libraries(core-minifi ${CMAKE_DL_LIBS} ${LIBMINIFI_LIBRARIES})
+if (APPLE)
+    target_link_libraries(core-minifi "-framework CoreFoundation -framework SystemConfiguration")
+endif()
```
szaszm, line 89: Was this moved from somewhere else, or what made this necessary?

lordgamez: It was needed in `minifi-http-curl` which is now part of `core-minifi` so it was moved here

arch.sh
lordgamez: Yes, but only on Windows

README.md
lordgamez: Updated in 6c350a979f1fbdd0e2e020feeb0ad7087b64771c

libminifi/test/unit/SchedulingAgentTests.cpp
lordgamez: It was removed, and readded so that's probably why the order changed, updated in 6c350a979f1fbdd0e2e020feeb0ad7087b64771c

README.md
adamdebreceni: did @szaszm mean line 186 (as well)?

extensions/expression-language/tests/integration/UpdateAttributeIntegrationTest.cpp
```diff
@@ -38,11 +38,10 @@ class TestHarness : public IntegrationBase {
   }
 
   void runAssertions() override {
-    using org::apache::nifi::minifi::utils::verifyLogLinePresenceInPollTime;
-    assert(verifyLogLinePresenceInPollTime(std::chrono::milliseconds(wait_time_),
+    REQUIRE(minifi::test::utils::verifyLogLinePresenceInPollTime(std::chrono::milliseconds(wait_time_),
         "key:route_check_attr value:good",
         "key:variable_attribute value:replacement_value"));
-    assert(false == verifyLogLinePresenceInPollTime(std::chrono::milliseconds(200), "ProcessSession rollback"));  // No rollback happened
+    REQUIRE(false == minifi::test::utils::verifyLogLinePresenceInPollTime(std::chrono::milliseconds(200), "ProcessSession rollback"));  // No rollback happened
```
adamdebreceni, line 44: REQUIRE_FALSE?

README.md
lordgamez: Good point, actually it seems that a lot of build and runtime dependencies are not relevant and should be updated at a couple of places. I created a separate jira ticket for this: https://issues.apache.org/jira/browse/MINIFICPP-2331 and additionally removed the unneeded build and runtime system dependencies here in 96d3b6a6896c37bb5b493a035a6f258e1bc97ada

extensions/expression-language/tests/integration/UpdateAttributeIntegrationTest.cpp
lordgamez: Updated in 96d3b6a6896c37bb5b493a035a6f258e1bc97ada

README.md
szaszm: I didn't notice line 186, but the same applies there.
We shouldn't remove the listed dependencies, even if a user doesn't need to install them. Those that don't need to be installed used to be marked with `(built and statically linked)`, meaning no system version is necessary, but users know that code from those libraries is used. If you want to remove them from this list, then we need to add them somewhere else.
If you want to do the addition of a second list in MINIFICPP-2331, then you should also do the removal from the existing list there. I understand that the list is not perfect at the moment, but I believe this half-change in 52c35a6334740c874f3515e3e95448eaeca8f143 makes it worse by removing even more information.

I see a few valid options for us in the context of this PR:
1. Mark openssl and curl as `(built and statically linked)`, and not touch the rest of the entries
2. Same as 1, plus do the same for other entries that have been recently changed. If something is now built and statically linked by our build system, mark it as such. If something is no longer used, remove it.
3. Do all changes described in [MINIFICPP-2331](https://issues.apache.org/jira/browse/MINIFICPP-2331) that are relevant for these two lists.

szaszm: After a discussion, we agreed to keep all "built and statically linked" entries removed, but add a note that these are only the libraries used from the system, and to add a reference to the NOTICE file for the full list of dependencies for any interested users.

lordgamez: Added a separate section with link to the LICENSE and NOTICE files to find information on the bundled thirdparty components and clarified the list of required system packages in b1d87f4a5ce46fb23bd23ded6775b62279314d72

fgerlits (approved):
I have only checked 263 of the files (tried to make it a random sample), but I haven't found any problems.

james94:
@szaszm I agree **MINIFI_OPENSSL** and **ENABLE_CURL** should now no longer be minifi_options, they must be enabled in order to successfully build libminifi. However, for the case of building libarchive, we can build libarchive with openssl enabled set to true or false. Did you make it mandatory for minifi extensions that use openssl?

In my PR-1775 for MINIFICPP-2346: I do note about the critical minifi options that must be enabled to successfully build libminifi. I found that **ENABLE_OPENWSMAN, ENABLE_CIVET** must also be ON for libminifi to build succssfully. Those 4 minifi options are the minimum to build libminifi C++ core-minifi, mainexe and standard processors. 

- I also mention about the options that must be enabled to build GTESTs. Furthermore, I am wrapping up testing enabling all the remaining extension.

lordgamez:
> @szaszm I agree **MINIFI_OPENSSL** and **ENABLE_CURL** should now no longer be minifi_options, they must be enabled in order to successfully build libminifi. However, for the case of building libarchive, we can build libarchive with openssl enabled set to true or false. Did you make it mandatory for minifi extensions that use openssl?
> 
> In my PR-1775 for MINIFICPP-2346: I do note about the critical minifi options that must be enabled to successfully build libminifi. I found that **ENABLE_OPENWSMAN, ENABLE_CIVET** must also be ON for libminifi to build succssfully. Those 4 minifi options are the minimum to build libminifi C++ core-minifi, mainexe and standard processors.
> 
>     * I also mention about the options that must be enabled to build GTESTs. Furthermore, I am wrapping up testing enabling all the remaining extension.

@james94 how did you find the ENABLE_OPENWSMAN must be ON for libminifi build to succeed? I managed to build it successfully without it and it should be a separate extension that should not have any effect on the core libminifi.

We also had some discussions in the past if civetweb should be part of the core libminifi, but we decided not to due to the following reasons:

- It is not actually required by core libminifi, it is only needed for the tests and some of the extensions. If you build libminifi with SKIP_TESTS=ON and turn off civetweb extension the libminifi should build successfully
- We had some concerns about the civetweb codebase so we may replace it in the future with a different library, so we would rather not include it in the core libminifi

libminifi/include/utils/TimeUtil.h
```diff
@@ -42,6 +42,10 @@
 
 #include "date/date.h"
 
+#ifdef WIN32
+#include "date/tz.h"
+#endif
```
szaszm, line 47: Why is this windows-only? The `dateSetInstall` function declaration doesn't use anything from this header, so if that's the culprit, then this include should be moved to the cpp file.

lordgamez: It was moved from the TestUtils and it only contained `dateSetInstall` that was WIN32 specific, so it was probably due to this. I moved it to the cpp file in 82084145ce5f0dd020144aea53c0c791bb69d3aa

james94:
@lordgamez hi, 

when I was talking about ENABLE_OPENWSMAN, I think I was more referring to the combination of **libminifi "core-minifi"** and **MiNiFi Main's  minifiexe** being dependent on the following external libs being enabled: ENABLE_OPENWSMAN, ENABLE_OPENSSL, ENABLE_CIVET, ENABLE_CURL. Also SKIP_TESTS was set to ON. 

Yes I do agree with you some of these externa libs like **ENABLE_OPENWSMAN** shouldn't impact **libminifi "core-minifi"**. 

I will get back to you as I am adding conan v2 updates on top of your branch with my new branch MINIFICPP-2346-P1.

libminifi/test/libtest/unit/TestUtils.h
```diff
@@ -0,0 +1,223 @@
...
+  explicit ExceptionSubStringMatcher(std::vector<std::string> exception_substr) :
+      possible_exception_substrs_(std::move(exception_substr)) {}
+
+  bool match(T const& script_exception) const override {
+    for (auto& possible_exception_substr : possible_exception_substrs_) {
+      if (std::string(script_exception.what()).find(possible_exception_substr) != std::string::npos)
+        return true;
+    }
+    return false;
+  }
```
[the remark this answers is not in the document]
lordgamez: Updated in 888959e1a1b28f4a688daeda35849da49dc05166

libminifi/test/libtest/unit/TestUtils.cpp
```diff
@@ -14,59 +14,117 @@
...
-namespace internal {
-struct JsonContext {
-  const JsonContext *parent{nullptr};
-  std::string_view member;
+std::filesystem::path putFileToDir(const std::filesystem::path& dir_path, const std::filesystem::path& file_name, const std::string& content) {
+  auto file_path = dir_path/file_name;
+  std::ofstream out_file(file_path, std::ios::binary | std::ios::out);
+  if (out_file.is_open()) {
+    out_file << content;
+  }
```
szaszm, line 71: Let's not fail silently here.
```suggestion
  assert(out_file.is_open());
  out_file << content;
```

szaszm (approved):

PR #1755, opened 2024-04-09
MINIFICPP-2325 Add missing compiler flags

fgerlits:
These flags are on by default when we use msbuild, but they need to be added manually when we build using ninja. Without these flags, the resulting dlls were roughly twice as large when using ninja compared to using msbuild. After adding the flags, the sizes are back to normal.

I have also replaced prep_for_win_package.cmake with a copy command, so it works both with msbuild and ninja.

https://issues.apache.org/jira/browse/MINIFICPP-2325

---

### For all changes:
     in the commit message?

### For code changes:

### Note:


martinzink (approved):

szaszm:
CI failure:
```
CPack: Executing pre-build script: D:/a/nifi-minifi-cpp/nifi-minifi-cpp/cmake/prep_for_win_package.cmake
CMake Error at D:/a/nifi-minifi-cpp/nifi-minifi-cpp/cmake/prep_for_win_package.cmake:7 (file):
  file COPY cannot find "/bin//minifi.exe": File exists.
```

minifi_main/CMakeLists.txt
```diff
@@ -74,12 +74,10 @@ if (WIN32)
...
 
-if (NOT WIN32)
-add_custom_command(TARGET minifiexe POST_BUILD
-    COMMAND cat ${CMAKE_BINARY_DIR}/all.log)
+if (WIN32)
+    add_custom_command(TARGET minifiexe POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:minifiexe> ${CMAKE_BINARY_DIR}/minifi_main/)
 else()
-#add_custom_command(TARGET minifiexe POST_BUILD
-#           COMMAND type ${CMAKE_BINARY_DIR}/all.log)
+    add_custom_command(TARGET minifiexe POST_BUILD COMMAND cat ${CMAKE_BINARY_DIR}/all.log)
```
szaszm, line 80: I think this `cat` can go, the contents of that log file is bogus anyway.

fgerlits: Ah OK, I didn't realize this is the part where we print that standard processors have been disabled.  Removed in 98a74b7c1fbc044fcc899d63034cbe4354ab6265.

lordgamez (approved):

PR #1762, opened 2024-04-19
MINIFICPP-2295 Add SSLContextService support for NiFi Python processors

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2295

-------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/python/PYTHON.md
+++ b/extensions/python/PYTHON.md
@@ -91,7 +91,7 @@ def describe(processor):
 
 onInitialize is also passed the processor reference and can be where you set properties. The first argument is the property display name,
 followed by the description, and default value. The next three arguments are booleans describing if the property is required, support expression language, and if it is a sensitive property.
-The last argument is the property type code. The property type code is an integer that represents the type of the property. The supported property type codes and their corresponding types:
+The seventh argument is the property type code. The property type code is an integer that represents the type of the property. The supported property type codes and their corresponding types:
 ```
 INTEGER = 0
 LONG = 1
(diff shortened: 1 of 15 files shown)


extensions/python/PYTHON.md
```diff
@@ -102,10 +102,12 @@ NON_BLANK = 5
 PORT = 6
 ```
 
+The last parameter of addProperty is the controller service type. If the property is a controller service, the controller service type should be provided. It should be the non-qualified type name of the controller service. Currently SSLContextService is the only controller service type supported.
+
 ```python
 def onInitialize(processor):
   processor.setSupportsDynamicProperties()
-  processor.addProperty("property name","description","default value", True /*required*/, False /*expression language supported*/, False /*sensitive*/, 1 /*property type code*/)
+  processor.addProperty("property name", "description", "default value", True /*required*/, False /*expression language supported*/, False /*sensitive*/, 1 /*property type code*/, None /*controller service type name*/)
```
szaszm, line 110: These `/**/` style comments are not valid python. I'm assuming this comes from the NiFi Python API. Otherwise I'd consider changing to keyword arguments.

lordgamez: Good point, I added comment for the arguments in 0b35f1cf419e8b57955f65f9192cf39b8d8a5fde. Due to the implementation of `PyProcessor::addProperty(PyProcessor* self, PyObject* args)`  these are handled as positional arguments, so using keyword arguments would be misleading.

szaszm (approved):

extensions/python/types/PyProcessContext.cpp
```diff
@@ -104,6 +107,30 @@ PyObject* PyProcessContext::getStateManager(PyProcessContext* self, PyObject*) {
...
+    throw PyException();
+  }
+
+  if (auto controller_service = context->getControllerService(controller_service_name)) {
+    std::string controller_service_type_str = controller_service_type;
+    if (controller_service_type_str == "SSLContextService") {
+      auto ssl_ctx_service = std::dynamic_pointer_cast<controllers::SSLContextService>(controller_service);
+      return object::returnReference(std::weak_ptr(ssl_ctx_service));
+    }
+  }
```
fgerlits, line 129: Since we only support `SSLContextService` for now, I think we should change the titles (commit, Jira, PR) to reflect this, eg. "Add SSLContextService support for NiFi python processors".

martinzink (approved):

extensions/python/types/PyProcessContext.cpp
lordgamez: Updated the titles, unfortunately all python bindings use static function definitions so no logger object is present to log the errors, and we should not throw an exception here, only return a NONE object like NiFi does.

PR #1765, opened 2024-04-22
MINIFICPP-2333 Don't register example python processors

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2333

--------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


docker/test/integration/cluster/containers/MinifiContainer.py
```diff
@@ -177,6 +178,8 @@ def deploy(self):
 
         if self.options.enable_sql:
             image = self.image_store.get_image('minifi-cpp-sql')
+        elif self.options.enable_example_minifi_python_processors:
+            image = self.image_store.get_image('minifi-cpp-with-example-python-processors')
         elif self.options.use_nifi_python_processors_with_system_python_packages_installed:
             image = self.image_store.get_image('minifi-cpp-nifi-python-system-python-packages')
```
fgerlits, line 184: Why do we only allow one of these at a time?  I would expect (both `enable_sql` and) `enable_example_minifi_python_processors` to be add-ons to the other images rather than stand-alone images themselves.

lordgamez: I think we just didn't have the need for it yet. These added features to the base minifi image are usually used in separate tests or even separate features, so we didn't need to combine them. I think it can be refactored in the future to have a list of required add-ons instead of a single parameter, but that should be done when one of the tests require it.

fgerlits (approved):

PR #1766, opened 2024-04-22
MINIFICPP-2337 Fix OpenSSL toggle in bootstrap.sh

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2337

-------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -445,7 +445,8 @@ build_cmake_command(){
       set_value=OFF
       option_value="${!option}"
       if { [[ "$option_value" = "${FALSE}" ]] && { [[ "$FOUND_VALUE" == "DISABLE"* ]] || [[ "$FOUND_VALUE" == *"OFF" ]]; }; } || \
-         { [[ "$option_value" = "${TRUE}" ]] && [[ "$FOUND_VALUE" == "ENABLE"* ]]; }; then
+         { [[ "$option_value" = "${TRUE}" ]] && [[ "$FOUND_VALUE" == "ENABLE"* ]]; } ||
+         { [[ "$option_value" = "${TRUE}" ]] && [[ "$FOUND_VALUE" != "ENABLE"* ]] && [[ "$FOUND_VALUE" != "DISABLE"* ]]; }; then
         set_value=ON
       fi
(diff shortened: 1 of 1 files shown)


bootstrap.sh
```diff
@@ -445,7 +445,8 @@ build_cmake_command(){
       set_value=OFF
       option_value="${!option}"
       if { [[ "$option_value" = "${FALSE}" ]] && { [[ "$FOUND_VALUE" == "DISABLE"* ]] || [[ "$FOUND_VALUE" == *"OFF" ]]; }; } || \
-         { [[ "$option_value" = "${TRUE}" ]] && [[ "$FOUND_VALUE" == "ENABLE"* ]]; }; then
+         { [[ "$option_value" = "${TRUE}" ]] && [[ "$FOUND_VALUE" == "ENABLE"* ]]; } ||
+         { [[ "$option_value" = "${TRUE}" ]] && [[ "$FOUND_VALUE" != "ENABLE"* ]] && [[ "$FOUND_VALUE" != "DISABLE"* ]]; }; then
```
szaszm, line 449: What is the problem that's fixed here? How did it not work?

lordgamez: Enabling OpenSSL is done with the MINIFI_OPENSSL option, but only the ENABLE*/DISABLE* options were handled here, so the openssl was never set to ON regardless of the option being toggled or not.

szaszm (approved):

martinzink (approved):

bootstrap.sh
fgerlits, line 449: I am a bit bothered by the fact that both `FOOBAR_OFF=FALSE` and `FOOBAR_OFF=TRUE` will set the foobar option to `ON`, although I don't think we have any CMake variables ending with `OFF` at the moment.

Maybe remove the `|| [[ "$FOUND_VALUE" == *"OFF" ]]` logic to avoid the impression that we support variables like this?

lordgamez: I think you are right, we went great lengths last time to remove any "negative" options so we should not introduce options with _OFF anyways. I removed the logic to handle `*"OFF"` options.

fgerlits (approved):

PR #603, opened 2019-06-28
MINIFICPP-929 mmap

achristianson:
****--DRAFT PR please review... lots of code changes, so many eyes are welcome--****

This PR adds a mmap() interface to allow processors to map FlowFile payloads to a memory address. This increases efficiency and performance significantly for some use cases. The change does not negatively impact performance in almost all cases, as shown in benchmarks.

Original/full reason/justification:

"Currently, MiNiFi - C++ only support stream-oriented i/o to FlowFile payloads. This can limit performance in cases where in-place access to the payload is desirable. In cases where data can be accessed randomly and in-place, a significant speedup can be realized by mapping the payload into system memory address space. This is natively supported at the kernel level in Linux, MacOS, and Windows via the mmap() interface on files. Other repositories, such as the VolatileRepository, already store the entire payload in memory, so it is natural to pass through this memory block as if it were a memory-mapped file. While the DatabaseContentRepostory does not appear to natively support a memory map interface, accesses via an emulated memory-map interface should be possible with no performance degradation with respect to a full read via the streaming interface.

Cases where in-place, random access is beneficial include, but are not limited to:

    in-place parsing of JSON (e.g. RapidJSON supports parsing in-place, at least for strings).
    access of payload via protocol buffers
    random access of large files on disk, where it would otherwise require many seek() and read() syscalls

The interface should be accessible by processors via a mmap() call on ProcessSession (adjacent to read() and write()). A MemoryMapCallback should be provided, which is called back via a process() call where the argument is an instance of BaseMemoryMap. The BaseMemoryMap is extended for each type of repository that MiNiFi - C++ supports, including: FileSystemRepository, VolatileRepository, and DatabaseContentRepository.

As part of the change, in addition to extensive unit test coverage, benchmarks should be written such that the performance impact can be empirically measured and evaluated."

Here is the full benchmark suite:

```
-------------------------------------------------------------------------------------------------------------------
Benchmark                                                                         Time             CPU   Iterations
-------------------------------------------------------------------------------------------------------------------
FSMemoryMapBMFixture/MemoryMap_FileSystemRepository_Read_Tiny                  2956 ns         2923 ns       240558
[2019-06-debugl 14:10:44.663] [org::apache::nifi::minifi::core::logging::LoggerConfiguration] [debug] org::apache::nifi::minifi::io::FileStream logger got sink
s from namespace root and level error from namespace root
[...]

PR #1775, opened 2024-05-02
MINIFICPP-2346: Switch to C++ Conan Packager to Replace CMake FetchContent for Faster MiNiFi CPP Builds

james94:
Also kept backward compatibility for cmake FetchContent by adding two MiNiFi CPP options: one for using CMake FetchContent and the other for using the more faster approach conan packager to manage C++ external lib dependencies. With conan, usually conan will install prebuilt binary conan packages from conancenter or some other conan repo where the prebuilt binary conan packages reside. For now, we'll stick with using conancenter as it is open source.

Verified I can run conan install command from MiNiFi root folder where conanfile.py resides and successfully install abseil prebuilt binary conan package for MiNiFi C++, then ran cmake generate from MiNiFi's build folder and then started to run make. Stopped early on make since MiNiFi would try to download and build all the remaining external lib dependencies with CMake FetchContent since they arent managed by conan yet.

Also in Abseil.cmake's file, I added conditional statement to check if user set the USE_CONAN_PACKAGER option or the USE_CMAKE_FETCH_CONTENT option. Will do similar approach for the remaining external lib dependencies

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,6 +45,10 @@ include(WholeArchive)
 
 include(MiNiFiOptions)
 
+message("Checking MiNiFi Options Updated by conanfile.py")
+message("USE_CONAN_PACKAGER = ${USE_CONAN_PACKAGER}")
+message("USE_CMAKE_FETCH_CONTENT = ${USE_CMAKE_FETCH_CONTENT}")
+
 include(DockerConfig)
 if (DOCKER_BUILD_ONLY)
     return()
--- a/cmake/Abseil.cmake
+++ b/cmake/Abseil.cmake
@@ -16,20 +16,26 @@
 # specific language governing permissions and limitations
 # under the License.
 #
-include(FetchContent)
-set(ABSL_PROPAGATE_CXX_STD ON CACHE INTERNAL absl-propagate-cxx)
-set(ABSL_ENABLE_INSTALL ON CACHE INTERNAL "")
-set(BUILD_TESTING OFF CACHE STRING "" FORCE)
+if(USE_CONAN_PACKAGER)
+    message("FAST: Using Conan Packager to manage prebuilt Abseil install")
+    include(${CMAKE_BINARY_DIR}/absl-config.cmake)
+elseif(USE_CMAKE_FETCH_CONTENT)
+    message("SLOW: Using CMAKE's FetchContent to manage source build Abseil install")
+    include(FetchContent)
+    set(ABSL_PROPAGATE_CXX_STD ON CACHE INTERNAL absl-propagate-cxx)
+    set(ABSL_ENABLE_INSTALL ON CACHE INTERNAL "")
+    set(BUILD_TESTING OFF CACHE STRING "" FORCE)
 
-set(PATCH_FILE "${CMAKE_SOURCE_DIR}/thirdparty/abseil/rename-crc32.patch")
-set(PC ${Bash_EXECUTABLE}  -c "set -x &&\
-        (\\\"${Patch_EXECUTABLE}\\\" -p1 -R -s -f --dry-run -i \\\"${PATCH_FILE}\\\" || \\\"${Patch_EXECUTABLE}\\\" -p1 -N -i \\\"${PATCH_FILE}\\\")")
+    set(PATCH_FILE "${CMAKE_SOURCE_DIR}/thirdparty/abseil/rename-crc32.patch")
+    set(PC ${Bash_EXECUTABLE}  -c "set -x &&\
+            (\\\"${Patch_EXECUTABLE}\\\" -p1 -R -s -f --dry-run -i \\\"${PATCH_FILE}\\\" || \\\"${Patch_EXECUTABLE}\\\" -p1 -N -i \\\"${PATCH_FILE}\\\")")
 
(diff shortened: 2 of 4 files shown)


szaszm:
There are a few things we would need for this to be viable:
- The python bootstrap should be updated to support selecting each extension with its dependencies installed from Conan vs FetchContent
- I think CMake should be updated to allow selecting the dependency source on a per-extension basis. This is also needed to implement the first point.
- (optional) If we're doing the per-extension dependency source selection, a normal find_package could be added as well, to use the system version of libs. This may not be trivial to get working, but if it works at least at a few places, it can be an easy win.
- I wouldn't flood FetchContent users with "SLOW" messages. Until we can ensure that all dependencies have viable binary packages on conan center, there isn't necessarily a build speed difference. We need to discuss how to handle these cases with Conan. And even after we can ensure binary packages for all major targets, FetchContent can still be a viable option, even if it builds the packages.

Let's see how this progresses, these can be done separately as well.

james94:
@lordgamez thanks for the suggestion, I'll switch the default option to USE_FETCH_CONTENT for now. I'm excited to hear it was planned for 1.0.0 release to bring this feature to the build infrastructure. When is the planned release date, so I can track that target date?

james94:
@szaszm thanks for sharing the requirements and what you guys are looking for to make this PR viable. I will shift the updates to meet those requirements. I appreciate it.

szaszm:
1.0.0 will be released some time after NiFi 2.0, but there is no clear target date. We were just discussing the idea (so no commitment), as third party library builds are already taking up the majority of the build time. I think someone would have given it a second try in the following months.
I did a first try of moving everything to Conan 2-3 years ago, but it quickly got too complex. Me not being very familiar with Conan probably didn't help it.
Thanks for the contribution! :)

martinzink:
You will need to rebase your PR for the CI to work.

james94:
@martinzink thanks for the suggestion. I brought MINIFI-2346 to up to date with latest upstream MiNiFi C++ main branch. Ubuntu 20.04 and Docker build for integration tests passed. The other CI/CDs failed. Its probably related to some of the changes I am introducing to the build infrastructure. I have mainly been testing the updates in a Ubuntu 22.04 docker container. I haven't tested them yet on Windows or Mac OS X. Those 2 last OSs, I'd test later. I also would need to look into why CI/CD is failing too. After I finish refactoring build infrastructure to support Conan version 2 and keep backward compatibility with the standalone CMake approach, I will work toward updating the MiNiFi bootstrap py script with conan support too. I will get to the CI/CD issues as soon as I can, but those will probably be the last updates.

szaszm:
@james94 I'm busy with other commitments to keep up with this, but I'd like to prevent wasting too much of your time on something we can't use. We can't review a 17k diff, so it will never be merged as is stands now.

If you want to pursue this, please create a separate pull request, where you only convert a single dependency to build with conan, so we can provide feedback there, and only change what's absolutely necessary. A good first target could be OpenSSL, because it takes forever to build on Windows.

Please make an effort to keep the diff as small as possible, in the 100s if possible, but keep it under 1000. We can review many small pull requests, but not giants like this. Example: #603 didn't get any reviews, despite being a very valuable feature

As the conversion of the first dependency is reviewed, we both learn from it, and subsequent dependency source conversions will be easier. Please only pursue 1 or max 2 features at a time, so we can keep up and provide the necessary feedback. This will inevitably take a few months.

I'll also give some early feedback on things I noticed so far:
- Please use a proper multi-choice option for selecting the source of a dependency instead of USE_CONAN_PACKAGER and USE_CMAKE_FETCH_CONTENT binary options. For example: MINIFI_OPENSSL_SOURCE: [conan|find_package|bundled]. The second option could keep just find_package, and rely on sources from the system. If it doesn't work, just having [conan|bundled] is fine.
- Please don't include patches that are not strictly necessary to make conan builds work
- Please prefix minifi-related options with MINIFI_ in cmake. Not all options follow this now, but future new CMake options should.
- Please don't add new dependencies in the context of converting to Conan. For example, minifi doesn't depend on ffmpeg or libtiff at the moment. If they're needed for a new feature, the new feature needs to be a separate pull request.

james94:
Thanks for your suggestion @szaszm

I was considering to break this PR into multiple parts because like you said there is a lot to review. Actually, with each major update and commit hash, my goal was to make it easier to track each section of the new update that I made. For example with just updating **OpenSSL** to have USE_CONAN_PACKAGER support and backward compatibility for USE_CMAKE_FETCH_CONTENT, which impacts **libminifi**, we could reference my current commit hash from **UPDATE (May 4, 2024) - [2d1b884](https://github.com/apache/nifi-minifi-cpp/pull/1775/commits/2d1b8842c6809380f981f9857543805efbfd00cc)** and extract out just the part for OpenSSL, so its a smaller PR following your suggestions. Once its approved, we're on the same page for what you guys are looking for, we could look toward bringing up USE_CONAN_PACKAGER support for building the whole **libminifi** like I do in commit **2d1b884**.

I will follow up with a smaller PR for **OpenSSL** impacting part of **libminifi**.

martinzink:
Closing this because of inactivity.

PR #1792, opened 2024-05-21
MINIFICPP-2367 Add support for non-sensitive parameters

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2367

-----------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/CONFIGURE.md
+++ b/CONFIGURE.md
@@ -19,6 +19,7 @@
 
 - [Table of Contents](#table-of-contents)
 - [Configuring](#configuring)
+  - [Parameter Contexts](#parameter-contexts)
   - [Configuring flow configuration format](#configuring-flow-configuration-format)
   - [Scheduling strategies](#scheduling-strategies)
   - [Configuring encryption for flow configuration](#configuring-encryption-for-flow-configuration)
@@ -41,6 +42,7 @@
   - [Configuring Repository storage locations](#configuring-repository-storage-locations)
   - [Configuring compression for rocksdb database](#configuring-compression-for-rocksdb-database)
   - [Configuring compaction for rocksdb database](#configuring-compaction-for-rocksdb-database)
+  - [Global RocksDB options](#global-rocksdb-options)
     - [Shared database](#shared-database)
   - [Configuring Repository encryption](#configuring-repository-encryption)
     - [Mixing encryption with shared backend](#mixing-encryption-with-shared-backend)
@@ -57,6 +59,7 @@
   - [Network Prioritizer Controller Service](#network-prioritizer-controller-service)
   - [Disk space watchdog](#disk-space-watchdog)
   - [Extension configuration](#extension-configuration)
+  - [Python processors](#python-processors)
 - [Log configuration](#log-configuration)
   - [Log appenders](#log-appenders)
   - [Log levels](#log-levels)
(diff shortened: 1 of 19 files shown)


libminifi/src/core/ParameterTokenParser.cpp
```diff
@@ -0,0 +1,82 @@
...
+        tokens_.push_back(ParameterToken{token_name, token_start, i - token_start + 1});
+        token_start = 0;
+      } else {
+        state = ParseState::OutsideToken;
+      }
+    } else {
+      if (state != ParseState::InToken) {
+        state = ParseState::OutsideToken;
+      }
+    }
```
adamdebreceni, line 58: should we check for an unterminated token as well (assert that the final state is OutsideToken)? `test #{missing end`

libminifi/src/core/ParameterTokenParser.cpp
```diff
@@ -0,0 +1,82 @@
...
+  if (tokens_.empty()) {
+    return input_;
+  }
+  std::string result;
+  uint32_t last_end = 0;
+  for (const auto& token : tokens_) {
+    result.append(input_.substr(last_end, token.getStart() - last_end));
+    auto parameter = parameter_context.getParameter(token.getName());
+    if (!parameter.has_value()) {
+      throw ParameterException("Parameter '" + token.getName() + "' not found");
```
adamdebreceni, line 72: as I understand the parameter starter sequence is `#{` so if the property is `##other##{test}` it should be translated to `##other#{test}` so the first double `#` is preserved while the latter is treated as escaped, I don't think the escaped `#` is deduplicated in during replacement

libminifi/src/core/ParameterTokenParser.cpp
lordgamez: In that case we do not handle it as a token, and I don't think that needs to be an error (as I checked it is the same way in NiFi). There is a separate test for that "Unfinished token is not a token" in `ParameterTokenParserTest.cpp`

libminifi/src/core/ParameterTokenParser.cpp
lordgamez: Good catch! Adjusted the escape handling to handle this case in 3fb949e219fab94f470d9ac6f895dd8f961c50ad

adamdebreceni (approved):

martinzink (approved):

CONFIGURE.md
```diff
@@ -121,8 +124,101 @@ It's recommended to create your configuration in YAML format or configure the ag
...
+                "identifier": "8a772a10-7c34-48e7-b152-b1a32c5db83e",
+                "type": "org.apache.nifi.minifi.processors.TailFile",
+                "schedulingStrategy": "TIMER_DRIVEN",
+                "schedulingPeriod": "1000 ms",
+                "properties": {
+                    "File to Tail": "#{tail_base_dir}/test_file2.log"
+                }
+            }
+        ],
+        "parameterContextName": "root-process-group-context"
```
szaszm, line 185: Are the parameter contexts identified by name in nifi too?

extensions/standard-processors/tests/unit/YamlConfigurationTests.cpp
```diff
@@ -1072,3 +1075,581 @@ TEST_CASE("Test serialization", "[YamlConfiguration]") {
...
+  REQUIRE(proc);
+  CHECK(proc->getProperty("Batch Size") == "1");
+  CHECK(proc->getProperty("Lookup frequency") == "12 min");
+  auto* subproc = flow->findProcessorByName("SubTailFile");
+  REQUIRE(subproc);
+  CHECK(subproc->getProperty("Batch Size") == "12");
+  CHECK(subproc->getProperty("Lookup frequency") == "1 sec");
+}
+
+TEST_CASE("Subprocessgroups cannot inherit parameters from parent processgroup", "[YamlConfiguration]") {
```
szaszm, line 1439: Is this on purpose? I thought they'd be inherited.

libminifi/include/core/ParameterContext.h
```diff
@@ -0,0 +1,61 @@
...
+
+#include <string>
+#include <unordered_set>
+#include <optional>
+
+#include "Core.h"
+
+namespace org::apache::nifi::minifi::core {
+
+class ParameterException : public std::runtime_error {
```
szaszm, line 28: This could be derived from minifi::Exception. Maybe have its own type enum value

extensions/standard-processors/tests/unit/YamlConfigurationTests.cpp
lordgamez: Yes, this is on purpose. In NiFi parameter contexts are not inherited between process groups. It has to be explicitly selected which parameter context should be used for each process group. There is a feature in NiFi which will be implemented later to use parameter context inheritance. That allows parameter contexts to be derived from other parameter contexts and using the derived contexts allows the process group to use the parameters from the base parameter context.

libminifi/include/core/ConfigurableComponent.h
```diff
@@ -136,6 +136,8 @@ class ConfigurableComponent {
    */
   bool getDynamicProperty(const std::string& name, std::string &value) const;
 
+  bool getDynamicProperty(const std::string& name, core::Property &item) const;
```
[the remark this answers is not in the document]
lordgamez: It is currently used in the tests to check if dynamic property sequences are handled properly. In the other overload only 1 value can be queried, but if the property has multiple values as a value sequence we need this overload to check those values.

libminifi/include/core/ProcessGroup.h
```diff
@@ -217,6 +217,9 @@ class ProcessGroup : public CoreComponent {
 
   void verify() const;
 
+  void setParameterContextName(const std::string &name);
+  std::string getParameterContextName() const;
+
```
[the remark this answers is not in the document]
lordgamez: Good idea, updated in 7404acec9914456ec059be5b109e09d6c8dff9d2

libminifi/src/core/flow/StructuredConfiguration.cpp
```diff
@@ -799,6 +862,19 @@ void StructuredConfiguration::parsePorts(const flow::Node& node, core::ProcessGr
   }
 }
 
+void StructuredConfiguration::parseParameterContextName(const flow::Node& node, core::ProcessGroup* parent) {
+  if (!parent) {
+    logger_->log_error("parseParameterContextName: no parent group was provided");
+    return;
+  }
```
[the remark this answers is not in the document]
lordgamez: Updated in 7404acec9914456ec059be5b109e09d6c8dff9d2

libminifi/include/core/ParameterContext.h
lordgamez: Updated in 7404acec9914456ec059be5b109e09d6c8dff9d2

libminifi/src/core/ParameterTokenParser.cpp
```diff
@@ -0,0 +1,105 @@
...
+    result.append(input_.substr(last_end, token->getStart() - last_end));
+    if (token->getType() == ParameterToken::ParameterTokenType::Escaped) {
+      result.append(token->getValue().value());
+      last_end = token->getStart() + token->getSize();
+      continue;
+    }
+
+    if (!parameter_context) {
+      throw ParameterException("Property references a parameter in its value, but no parameter context was provided.");
+    }
```
szaszm, line 87: I'd consider moving the check outside and making the `parameter_context` parameter not nullable (reference).

libminifi/include/core/FlowConfiguration.h
```diff
@@ -134,6 +135,7 @@ class FlowConfiguration : public CoreComponent {
   utils::ChecksumCalculator& getChecksumCalculator() { return checksum_calculator_; }
 
  protected:
+  std::unordered_map<std::string, std::unique_ptr<ParameterContext>> parameter_contexts_;
```
szaszm, line 138: We could make this `gsl::not_null`, and not even store null contexts if they are even possible. And just give stronger guarantees to any code that uses this member.

libminifi/src/core/ParameterTokenParser.cpp
lordgamez: In some cases parameter context is not required here, when no parameter tokens are present only the escaped `#` characters are replaced.

libminifi/include/core/FlowConfiguration.h
lordgamez: Updated in b7ea62e8d312dcb41c2893263cdbfcc46ab22064

szaszm (approved):

PR #68, opened 2017-03-17
MINIFI-193: Ensure safe UTF encoding

phrocker:
Since the C++ library is agnostic of UTF-8 we can safely write the bytes.
Since we won't be interpreting the UTF-8 code in the core library
we do not need any additional dependencies. Nor do we need to worry about
encoding beyond proper serialization and deserialization of the byte array.

Thank you for submitting a contribution to Apache NiFi - MiNiFi C++.

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [x] Is there a JIRA ticket associated with this PR? Is it referenced
     in the commit message?

- [x] Does your PR title start with MINIFI-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.

- [ ] Has your PR been rebased against the latest commit within the target branch (typically master)?

- [x] Is your initial contribution a single, squashed commit?

### For code changes:
- [x] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?
- [x] If applicable, have you updated the LICENSE file?
- [x] If applicable, have you updated the NOTICE file?

### For documentation related changes:
- [x] Have you ensured that format looks appropriate for the output in which it is rendered?

### Note:
Please ensure that once the PR is submitted, you check travis-ci for build issues and submit an update to your PR as soon as possible.

PR #64, opened 2017-03-08
MINIFI-235: Use the appropriate connection when requested_hostname_ i…

phrocker:
…s not localhost

Thank you for submitting a contribution to Apache NiFi - MiNiFi C++.

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [x] Is there a JIRA ticket associated with this PR? Is it referenced
     in the commit message?

- [x] Does your PR title start with MINIFI-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.

- [x] Has your PR been rebased against the latest commit within the target branch (typically master)?

- [x] Is your initial contribution a single, squashed commit?

### For code changes:
- [x] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?
- [x] If applicable, have you updated the LICENSE file?
- [x] If applicable, have you updated the NOTICE file?

### For documentation related changes:
- [x] Have you ensured that format looks appropriate for the output in which it is rendered?

### Note:
Please ensure that once the PR is submitted, you check travis-ci for build issues and submit an update to your PR as soon as possible.

PR #1793, opened 2024-05-21
Integrated Conan2 for OpenSSL, CURL, ZLIB & Build MiNiFi

james94:
Verified I can Install & Build MiNiFi w OpenSSL, CURL, ZLIB using conan, then I set the MINIFI_HOME environment variable & ran minifi binary executable.

Will add the steps to reproduce (later we can integrate conan into the python bootstrap, the goal was to introduce you guys to the conan build approach with minimal changes): @lordgamez @szaszm

~~~bash
# make sure to install conan2 for your environment
sudo pip install --force-reinstall -v "conan==2.3.1"

# conanfile.py is in root dir of MiNiFi C++ project
cd $HOME/nifi-minifi-cpp

# create MINIFI_HOME env variable for binary executable minifi
export MINIFI_HOME=$(pwd)

# In case you need to delete all conan packages
# conan remove "*" -c

# install conan packages for MiNiFi C++ using conanfile.py invoking Conan
conan install . --build=missing --output-folder=build_conan -pr=$HOME/nifi-minifi-cpp/etc/build/conan/profiles/release-linux

# build MiNiFi C++ using conanfile.py invoking Conan & CMake
conan build . --output-folder=build_conan -pr=$HOME/nifi-minifi-cpp/etc/build/conan/profiles/release-linux

# verify we can run minifi binary executable
./build_conan/bin/minifi
~~~

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/cmake/BundledZLIB.cmake
+++ b/cmake/BundledZLIB.cmake
@@ -18,56 +18,65 @@
 function(use_bundled_zlib SOURCE_DIR BINARY_DIR)
     message("Using bundled zlib")
 
-    # Define byproducts
-    if (WIN32)
-        string(TOLOWER "${CMAKE_BUILD_TYPE}" build_type)
-        if (build_type MATCHES relwithdebinfo OR build_type MATCHES release)
-            set(BYPRODUCT "lib/zlibstatic.lib")
+    if(USE_CONAN_PACKAGER)
+        message("Using Conan Packager to manage installing prebuilt zlib external lib")
+        include(${CMAKE_BINARY_DIR}/ZLIBConfig.cmake)
+
+    elseif(USE_STANDALONE_CMAKE)
+        message("Using CMAKE's ExternalProject_Add to manage source building zlib external lib")
+    
+        # Define byproducts
+        if (WIN32)
+            string(TOLOWER "${CMAKE_BUILD_TYPE}" build_type)
+            if (build_type MATCHES relwithdebinfo OR build_type MATCHES release)
+                set(BYPRODUCT "lib/zlibstatic.lib")
+            else()
+                set(BYPRODUCT "lib/zlibstaticd.lib")
+            endif()
         else()
-            set(BYPRODUCT "lib/zlibstaticd.lib")
+            set(BYPRODUCT "lib/libz.a")
         endif()
-    else()
-        set(BYPRODUCT "lib/libz.a")
+
+        # Set build options
+        set(ZLIB_CMAKE_ARGS ${PASSTHROUGH_CMAKE_ARGS}
+                "-DCMAKE_INSTALL_PREFIX=${BINARY_DIR}/thirdparty/zlib-install"
+                )
+
+        # Build project
+        ExternalProject_Add(
+            zlib-external
+            URL "https://github.com/madler/zlib/archive/v1.2.11.tar.gz"
+            URL_HASH "SHA256=629380c90a77b964d896ed37163f5c3a34f6e6d897311f1df2a7016355c45eff"
+            SOURCE_DIR "${BINARY_DIR}/thirdparty/zlib-src"
+            CMAKE_ARGS ${ZLIB_CMAKE_ARGS}
+            BUILD_BYPRODUCTS "${BINARY_DIR}/thirdparty/zlib-install/${BYPRODUCT}"
+            EXCLUDE_FROM_ALL TRUE
+        )
+
+        # Set variables
+        set(ZLIB_FOUND "YES" CACHE STRING "" FORCE)
+        set(ZLIB_INCLUDE_DIRS "${BINARY_DIR}/thirdparty/zlib-install/include" CACHE STRING "" FORCE)
+        set(ZLIB_LIBRARIES "${BINARY_DIR}/thirdparty/zlib-install/${BYPRODUCT}" CACHE STRING "" FORCE)
+        set(ZLIB_VERSION_STRING "1.2.11" CACHE STRING "" FORCE)
+        set(ZLIB_VERSION_MAJOR 1 CACHE STRING "" FORCE)
+        set(ZLIB_VERSION_MINOR 2 CACHE STRING "" FORCE)
+        set(ZLIB_VERSION_PATCH 11 CACHE STRING "" FORCE)
+
+        # Set exported variables for FindPackage.cmake
+        set(PASSTHROUGH_VARIABLES ${PASSTHROUGH_VARIABLES} "-DEXPORTED_ZLIB_INCLUDE_DIRS=${ZLIB_INCLUDE_DIRS}" CACHE STRING "" FORCE)
+        set(PASSTHROUGH_VARIABLES ${PASSTHROUGH_VARIABLES} "-DEXPORTED_ZLIB_LIBRARIES=${ZLIB_LIBRARIES}" CACHE STRING "" FORCE)
+        set(PASSTHROUGH_VARIABLES ${PASSTHROUGH_VARIABLES} "-DEXPORTED_ZLIB_VERSION_STRING=${ZLIB_VERSION_STRING}" CACHE STRING "" FORCE)
+        set(PASSTHROUGH_VARIABLES ${PASSTHROUGH_VARIABLES} "-DEXPORTED_ZLIB_VERSION_MAJOR=${ZLIB_VERSION_MAJOR}" CACHE STRING "" FORCE)
+        set(PASSTHROUGH_VARIABLES ${PASSTHROUGH_VARIABLES} "-DEXPORTED_ZLIB_VERSION_MINOR=${ZLIB_VERSION_MINOR}" CACHE STRING "" FORCE)
+        set(PASSTHROUGH_VARIABLES ${PASSTHROUGH_VARIABLES} "-DEXPORTED_ZLIB_VERSION_PATCH=${ZLIB_VERSION_PATCH}" CACHE STRING "" FORCE)
+
+        # Create imported targets
+        file(MAKE_DIRECTORY ${ZLIB_INCLUDE_DIRS})
+
+        add_library(ZLIB::ZLIB STATIC IMPORTED)
+        set_target_properties(ZLIB::ZLIB PROPERTIES IMPORTED_LOCATION "${ZLIB_LIBRARIES}")
+        add_dependencies(ZLIB::ZLIB zlib-external)
+        set_property(TARGET ZLIB::ZLIB APPEND PROPERTY INTERFACE_INCLUDE_DIRECTORIES ${ZLIB_INCLUDE_DIRS})
+
     endif()
-
-    # Set build options
-    set(ZLIB_CMAKE_ARGS ${PASSTHROUGH_CMAKE_ARGS}
-            "-DCMAKE_INSTALL_PREFIX=${BINARY_DIR}/thirdparty/zlib-install"
-            )
-
-    # Build project
-    ExternalProject_Add(
-        zlib-external
-        URL "https://github.com/madler/zlib/archive/v1.2.11.tar.gz"
-        URL_HASH "SHA256=629380c90a77b964d896ed37163f5c3a34f6e6d897311f1df2a7016355c45eff"
-        SOURCE_DIR "${BINARY_DIR}/thirdparty/zlib-src"
-        CMAKE_ARGS ${ZLIB_CMAKE_ARGS}
-        BUILD_BYPRODUCTS "${BINARY_DIR}/thirdparty/zlib-install/${BYPRODUCT}"
-        EXCLUDE_FROM_ALL TRUE
-    )
-
-    # Set variables
-    set(ZLIB_FOUND "YES" CACHE STRING "" FORCE)
-    set(ZLIB_INCLUDE_DIRS "${BINARY_DIR}/thirdparty/zlib-install/include" CACHE STRING "" FORCE)
-    set(ZLIB_LIBRARIES "${BINARY_DIR}/thirdparty/zlib-install/${BYPRODUCT}" CACHE STRING "" FORCE)
-    set(ZLIB_VERSION_STRING "1.2.11" CACHE STRING "" FORCE)
-    set(ZLIB_VERSION_MAJOR 1 CACHE STRING "" FORCE)
-    set(ZLIB_VERSION_MINOR 2 CACHE STRING "" FORCE)
-    set(ZLIB_VERSION_PATCH 11 CACHE STRING "" FORCE)
-
-    # Set exported variables for FindPackage.cmake
-    set(PASSTHROUGH_VARIABLES ${PASSTHROUGH_VARIABLES} "-DEXPORTED_ZLIB_INCLUDE_DIRS=${ZLIB_INCLUDE_DIRS}" CACHE STRING "" FORCE)
-    set(PASSTHROUGH_VARIABLES ${PASSTHROUGH_VARIABLES} "-DEXPORTED_ZLIB_LIBRARIES=${ZLIB_LIBRARIES}" CACHE STRING "" FORCE)
-    set(PASSTHROUGH_VARIABLES ${PASSTHROUGH_VARIABLES} "-DEXPORTED_ZLIB_VERSION_STRING=${ZLIB_VERSION_STRING}" CACHE STRING "" FORCE)
-    set(PASSTHROUGH_VARIABLES ${PASSTHROUGH_VARIABLES} "-DEXPORTED_ZLIB_VERSION_MAJOR=${ZLIB_VERSION_MAJOR}" CACHE STRING "" FORCE)
-    set(PASSTHROUGH_VARIABLES ${PASSTHROUGH_VARIABLES} "-DEXPORTED_ZLIB_VERSION_MINOR=${ZLIB_VERSION_MINOR}" CACHE STRING "" FORCE)
-    set(PASSTHROUGH_VARIABLES ${PASSTHROUGH_VARIABLES} "-DEXPORTED_ZLIB_VERSION_PATCH=${ZLIB_VERSION_PATCH}" CACHE STRING "" FORCE)
-
-    # Create imported targets
-    file(MAKE_DIRECTORY ${ZLIB_INCLUDE_DIRS})
-
-    add_library(ZLIB::ZLIB STATIC IMPORTED)
-    set_target_properties(ZLIB::ZLIB PROPERTIES IMPORTED_LOCATION "${ZLIB_LIBRARIES}")
-    add_dependencies(ZLIB::ZLIB zlib-external)
-    set_property(TARGET ZLIB::ZLIB APPEND PROPERTY INTERFACE_INCLUDE_DIRECTORIES ${ZLIB_INCLUDE_DIRS})
 endfunction(use_bundled_zlib)
--- a/cmake/MiNiFiOptions.cmake
+++ b/cmake/MiNiFiOptions.cmake
@@ -42,6 +42,8 @@ add_minifi_option(DOCKER_SKIP_TESTS "Skip building tests in docker image targets
 add_minifi_option(DOCKER_PUSH "Push created images to the specified tags" OFF)
 
 add_minifi_option(PORTABLE "Instructs the compiler to remove architecture specific optimizations" ON)
+add_minifi_option(USE_CONAN_PACKAGER "Instructs cmake to use conan packager to install external libs & build MiNiFi C++" OFF)
+add_minifi_option(USE_STANDALONE_CMAKE "Instructs cmake to use only cmake to install external libs & build MiNiFi C++" ON)
 add_minifi_option(USE_SHARED_LIBS "Builds using shared libraries" ON)
 add_minifi_dependent_option(STATIC_BUILD "Attempts to statically link as many dependencies as possible." ON "NOT USE_SHARED_LIBS" OFF)
 add_minifi_option(LIBC_STATIC "Instructs the build system to statically link libstdc++ and glibc into minifiexe. Experiemental" OFF)
--- a/conanfile.py
+++ b/conanfile.py
@@ -1,0 +1,74 @@
+from conan import ConanFile
+from conan.tools.env import VirtualRunEnv
+from conan.tools.cmake import CMake, CMakeToolchain
+
+required_conan_version = ">=2.1.0"
+
+shared_requires = ("openssl/3.2.1", "libcurl/8.6.0")
+
+class MiNiFiCppMain(ConanFile):
+    name = "minifi-cpp-main"
+    version = "0.99.0"
+    license = "Apache-2.0"
+    requires = shared_requires
+    settings = "os", "compiler", "build_type", "arch"
+    generators = "CMakeDeps"
+    options = {"shared": [True, False], "fPIC": [True, False]}
+
+    default_options = {"shared": False, "fPIC": True,}
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["USE_CONAN_PACKAGER"] = "ON"
+        tc.variables["USE_STANDALONE_CMAKE"] = "OFF"
+
+        tc.variables["ENABLE_OPENWSMAN"] = "OFF"
+        tc.variables["ENABLE_CIVET"] = "OFF"
+        tc.variables["SKIP_TESTS"] = "ON"
+        tc.variables["ENABLE_EXPRESSION_LANGUAGE"] = "OFF"
+        tc.variables["ENABLE_BZIP2"] = "OFF"
+        tc.variables["ENABLE_ROCKSDB"] = "OFF"
+        tc.variables["BUILD_ROCKSDB"] = "OFF"
+
+        if self.settings.os == "Windows":
+            tc.variables["ENABLE_WEL"] = "OFF"
+            tc.variables["ENABLE_PDH"] = "OFF"
+            tc.variables["ENABLE_SMB"] = "OFF"
+        elif self.settings.os == "Linux":
+            tc.variables["ENABLE_SYSTEMD"] = "OFF"
+            tc.variables["ENABLE_PROCFS"] = "OFF"
+
+        tc.variables["ENABLE_LIBARCHIVE"] = "OFF"
+        tc.variables["ENABLE_LZMA"] = "OFF"
+        tc.variables["ENABLE_GPS"] = "OFF"
+        tc.variables["ENABLE_COAP"] = "OFF"
+        tc.variables["ENABLE_SQL"] = "OFF"
+        tc.variables["ENABLE_MQTT"] = "OFF"
+        tc.variables["ENABLE_PCAP"] = "OFF"
+        tc.variables["ENABLE_LIBRDKAFKA"] = "OFF"
+        tc.variables["ENABLE_LUA_SCRIPTING"] = "OFF"
+        tc.variables["ENABLE_PYTHON_SCRIPTING"] = "OFF"
+        tc.variables["ENABLE_SENSORS"] = "OFF"
+        tc.variables["ENABLE_USB_CAMERA"] = "OFF"
+        tc.variables["ENABLE_AWS"] = "OFF"
+        tc.variables["ENABLE_OPENCV"] = "OFF"
+        tc.variables["ENABLE_BUSTACHE"] = "OFF"
+        tc.variables["ENABLE_SFTP"] = "OFF"
+        tc.variables["ENABLE_AZURE"] = "OFF"
+        tc.variables["ENABLE_ENCRYPT_CONFIG"] = "OFF"
+        tc.variables["ENABLE_SPLUNK"] = "OFF"
+        tc.variables["ENABLE_ELASTICSEARCH"] = "OFF"
+        tc.variables["ENABLE_GCP"] = "OFF"
+        tc.variables["ENABLE_KUBERNETES"] = "OFF"
+        tc.variables["ENABLE_TEST_PROCESSORS"] = "OFF"
+        tc.variables["ENABLE_PROMETHEUS"] = "OFF"
+        tc.variables["ENABLE_GRAFANA_LOKI"] = "OFF"
+        tc.variables["ENABLE_GRPC_FOR_LOKI"] = "OFF"
+        tc.variables["ENABLE_CONTROLLER"] = "OFF"
+
+        tc.generate()
+
+    def build(self):
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.build()
--- a/cmake/BundledLibcURL.cmake
+++ b/cmake/BundledLibcURL.cmake
@@ -16,83 +16,93 @@
 # under the License.
 
 function(use_bundled_curl SOURCE_DIR BINARY_DIR)
-    # Define patch step
-    set(PATCH_FILE_1 "${SOURCE_DIR}/thirdparty/curl/module-path.patch")
-    set(PC ${Bash_EXECUTABLE} -c "set -x && \
-            (\"${Patch_EXECUTABLE}\" -p1 -R -s -f --dry-run -i \"${PATCH_FILE_1}\" || \"${Patch_EXECUTABLE}\" -p1 -N -i \"${PATCH_FILE_1}\")")
-    # Define byproducts
-    string(TOLOWER "${CMAKE_BUILD_TYPE}" build_type)
-    if (WIN32)
-        if (build_type MATCHES relwithdebinfo OR build_type MATCHES release)
-            set(BYPRODUCT "lib/libcurl.lib")
+    if(USE_CONAN_PACKAGER)
+        message("Using Conan Packager to manage installing prebuilt libcurl external lib")
+        include(${CMAKE_BINARY_DIR}/CURLConfig.cmake)
+        
+    elseif(USE_STANDALONE_CMAKE)
+        message("Using CMAKE's ExternalProject_Add to manage source building libcurl external lib")
+
+
+        # Define patch step
+        set(PATCH_FILE_1 "${SOURCE_DIR}/thirdparty/curl/module-path.patch")
+        set(PC ${Bash_EXECUTABLE} -c "set -x && \
+                (\"${Patch_EXECUTABLE}\" -p1 -R -s -f --dry-run -i \"${PATCH_FILE_1}\" || \"${Patch_EXECUTABLE}\" -p1 -N -i \"${PATCH_FILE_1}\")")
+        # Define byproducts
+        string(TOLOWER "${CMAKE_BUILD_TYPE}" build_type)
+        if (WIN32)
+            if (build_type MATCHES relwithdebinfo OR build_type MATCHES release)
+                set(BYPRODUCT "lib/libcurl.lib")
+            else()
+                set(BYPRODUCT "lib/libcurl-d.lib")
+            endif()
         else()
-            set(BYPRODUCT "lib/libcurl-d.lib")
+            include(GNUInstallDirs)
+            string(REPLACE "/" ";" LIBDIR_LIST ${CMAKE_INSTALL_LIBDIR})
+            list(GET LIBDIR_LIST 0 LIBDIR)
+            if (build_type MATCHES relwithdebinfo OR build_type MATCHES release)
+                set(BYPRODUCT "${LIBDIR}/libcurl.a")
+            else()
+                set(BYPRODUCT "${LIBDIR}/libcurl-d.a")
+            endif()
         endif()
-    else()
-        include(GNUInstallDirs)
-        string(REPLACE "/" ";" LIBDIR_LIST ${CMAKE_INSTALL_LIBDIR})
-        list(GET LIBDIR_LIST 0 LIBDIR)
-        if (build_type MATCHES relwithdebinfo OR build_type MATCHES release)
-            set(BYPRODUCT "${LIBDIR}/libcurl.a")
-        else()
-            set(BYPRODUCT "${LIBDIR}/libcurl-d.a")
+
+        # Set build options
+        set(CURL_CMAKE_ARGS ${PASSTHROUGH_CMAKE_ARGS}
+                "-DCMAKE_INSTALL_PREFIX=${BINARY_DIR}/thirdparty/curl-install"
+                -DBUILD_CURL_EXE=OFF
+                -DBUILD_TESTING=OFF
+                -DBUILD_SHARED_LIBS=OFF
+                -DHTTP_ONLY=ON
+                -DCURL_CA_PATH=none
+                -DCURL_USE_LIBSSH2=OFF
+                -DUSE_LIBIDN2=OFF
+                -DCURL_USE_LIBPSL=OFF
+                -DCURL_USE_OPENSSL=ON
+                )
+
+        append_third_party_passthrough_args(CURL_CMAKE_ARGS "${CURL_CMAKE_ARGS}")
+
+        # Build project
+        ExternalProject_Add(
+                curl-external
+                URL "https://github.com/curl/curl/releases/download/curl-8_4_0/curl-8.4.0.tar.gz"
+                URL_HASH "SHA256=816e41809c043ff285e8c0f06a75a1fa250211bbfb2dc0a037eeef39f1a9e427"
+                SOURCE_DIR "${BINARY_DIR}/thirdparty/curl-src"
+                LIST_SEPARATOR % # This is needed for passing semicolon-separated lists
+                CMAKE_ARGS ${CURL_CMAKE_ARGS}
+                PATCH_COMMAND ${PC}
+                BUILD_BYPRODUCTS "${BINARY_DIR}/thirdparty/curl-install/${BYPRODUCT}"
+                EXCLUDE_FROM_ALL TRUE
+        )
+
+        # Set dependencies
+        add_dependencies(curl-external ZLIB::ZLIB OpenSSL::SSL OpenSSL::Crypto)
+
+        # Set variables
+        set(CURL_FOUND "YES" CACHE STRING "" FORCE)
+        set(CURL_INCLUDE_DIR "${BINARY_DIR}/thirdparty/curl-install/include" CACHE STRING "" FORCE)
+        set(CURL_INCLUDE_DIRS "${CURL_INCLUDE_DIR}" CACHE STRING "" FORCE)
+        set(CURL_LIBRARY "${BINARY_DIR}/thirdparty/curl-install/${BYPRODUCT}" CACHE STRING "" FORCE)
+        set(CURL_LIBRARIES "${CURL_LIBRARY}" CACHE STRING "" FORCE)
+
+        # Set exported variables for FindPackage.cmake
+        set(PASSTHROUGH_VARIABLES ${PASSTHROUGH_VARIABLES} "-DEXPORTED_CURL_INCLUDE_DIR=${CURL_INCLUDE_DIR}" CACHE STRING "" FORCE)
+        set(PASSTHROUGH_VARIABLES ${PASSTHROUGH_VARIABLES} "-DEXPORTED_CURL_LIBRARY=${CURL_LIBRARY}" CACHE STRING "" FORCE)
+
+        # Create imported targets
+        file(MAKE_DIRECTORY ${CURL_INCLUDE_DIRS})
+
+        add_library(CURL::libcurl STATIC IMPORTED)
+        set_target_properties(CURL::libcurl PROPERTIES IMPORTED_LOCATION "${CURL_LIBRARY}")
+        add_dependencies(CURL::libcurl curl-external)
+        target_include_directories(CURL::libcurl INTERFACE ${CURL_INCLUDE_DIRS})
+        target_link_libraries(CURL::libcurl INTERFACE ZLIB::ZLIB Threads::Threads OpenSSL::SSL OpenSSL::Crypto)
+        if (APPLE)
+            target_link_libraries(CURL::libcurl INTERFACE "-framework CoreFoundation")
+            target_link_libraries(CURL::libcurl INTERFACE "-framework SystemConfiguration")
+            target_link_libraries(CURL::libcurl INTERFACE "-framework CoreServices")
         endif()
-    endif()
 
-    # Set build options
-    set(CURL_CMAKE_ARGS ${PASSTHROUGH_CMAKE_ARGS}
-            "-DCMAKE_INSTALL_PREFIX=${BINARY_DIR}/thirdparty/curl-install"
-            -DBUILD_CURL_EXE=OFF
-            -DBUILD_TESTING=OFF
-            -DBUILD_SHARED_LIBS=OFF
-            -DHTTP_ONLY=ON
-            -DCURL_CA_PATH=none
-            -DCURL_USE_LIBSSH2=OFF
-            -DUSE_LIBIDN2=OFF
-            -DCURL_USE_LIBPSL=OFF
-            -DCURL_USE_OPENSSL=ON
-            )
-
-    append_third_party_passthrough_args(CURL_CMAKE_ARGS "${CURL_CMAKE_ARGS}")
-
-    # Build project
-    ExternalProject_Add(
-            curl-external
-            URL "https://github.com/curl/curl/releases/download/curl-8_4_0/curl-8.4.0.tar.gz"
-            URL_HASH "SHA256=816e41809c043ff285e8c0f06a75a1fa250211bbfb2dc0a037eeef39f1a9e427"
-            SOURCE_DIR "${BINARY_DIR}/thirdparty/curl-src"
-            LIST_SEPARATOR % # This is needed for passing semicolon-separated lists
-            CMAKE_ARGS ${CURL_CMAKE_ARGS}
-            PATCH_COMMAND ${PC}
-            BUILD_BYPRODUCTS "${BINARY_DIR}/thirdparty/curl-install/${BYPRODUCT}"
-            EXCLUDE_FROM_ALL TRUE
-    )
-
-    # Set dependencies
-    add_dependencies(curl-external ZLIB::ZLIB OpenSSL::SSL OpenSSL::Crypto)
-
-    # Set variables
-    set(CURL_FOUND "YES" CACHE STRING "" FORCE)
-    set(CURL_INCLUDE_DIR "${BINARY_DIR}/thirdparty/curl-install/include" CACHE STRING "" FORCE)
-    set(CURL_INCLUDE_DIRS "${CURL_INCLUDE_DIR}" CACHE STRING "" FORCE)
-    set(CURL_LIBRARY "${BINARY_DIR}/thirdparty/curl-install/${BYPRODUCT}" CACHE STRING "" FORCE)
-    set(CURL_LIBRARIES "${CURL_LIBRARY}" CACHE STRING "" FORCE)
-
-    # Set exported variables for FindPackage.cmake
-    set(PASSTHROUGH_VARIABLES ${PASSTHROUGH_VARIABLES} "-DEXPORTED_CURL_INCLUDE_DIR=${CURL_INCLUDE_DIR}" CACHE STRING "" FORCE)
-    set(PASSTHROUGH_VARIABLES ${PASSTHROUGH_VARIABLES} "-DEXPORTED_CURL_LIBRARY=${CURL_LIBRARY}" CACHE STRING "" FORCE)
-
-    # Create imported targets
-    file(MAKE_DIRECTORY ${CURL_INCLUDE_DIRS})
-
-    add_library(CURL::libcurl STATIC IMPORTED)
-    set_target_properties(CURL::libcurl PROPERTIES IMPORTED_LOCATION "${CURL_LIBRARY}")
-    add_dependencies(CURL::libcurl curl-external)
-    target_include_directories(CURL::libcurl INTERFACE ${CURL_INCLUDE_DIRS})
-    target_link_libraries(CURL::libcurl INTERFACE ZLIB::ZLIB Threads::Threads OpenSSL::SSL OpenSSL::Crypto)
-    if (APPLE)
-        target_link_libraries(CURL::libcurl INTERFACE "-framework CoreFoundation")
-        target_link_libraries(CURL::libcurl INTERFACE "-framework SystemConfiguration")
-        target_link_libraries(CURL::libcurl INTERFACE "-framework CoreServices")
     endif()
 endfunction(use_bundled_curl SOURCE_DIR BINARY_DIR)
--- a/cmake/BundledOpenSSL.cmake
+++ b/cmake/BundledOpenSSL.cmake
@@ -18,120 +18,129 @@
 function(use_openssl SOURCE_DIR BINARY_DIR)
     message("Using bundled OpenSSL")
 
-    if(APPLE OR WIN32 OR CMAKE_SIZEOF_VOID_P EQUAL 4 OR CMAKE_SYSTEM_PROCESSOR MATCHES "(arm64)|(ARM64)|(aarch64)|(armv8)")
-        set(LIBDIR "lib")
-    else()
-        set(LIBDIR "lib64")
-    endif()
+    if(USE_CONAN_PACKAGER)
+        message("Using Conan Packager to manage installing prebuilt OpenSSL external lib")
+        include(${CMAKE_BINARY_DIR}/OpenSSLConfig.cmake)
 
-    # Define byproducts
-    set(BYPRODUCT_PREFIX "lib" CACHE STRING "" FORCE)
-    set(OPENSSL_BUILD_SHARED "NO" CACHE STRING "" FORCE)
+    elseif(USE_STANDALONE_CMAKE)
+        message("Using CMAKE's ExternalProject_Add to manage source building OpenSSL external lib")
 
-    if (WIN32)
-        set(BYPRODUCT_SUFFIX ".lib" CACHE STRING "" FORCE)
-    # Due to OpenSSL 3's static linking issue on x86 MacOS platform we make an exception to build a shared library instead
-    elseif (APPLE AND (CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64|amd64|AMD64"))
-        set(BYPRODUCT_SUFFIX ".dylib" CACHE STRING "" FORCE)
-        set(OPENSSL_BUILD_SHARED "YES" CACHE STRING "" FORCE)
-    else()
-        set(BYPRODUCT_SUFFIX ".a" CACHE STRING "" FORCE)
-    endif()
+        if(APPLE OR WIN32 OR CMAKE_SIZEOF_VOID_P EQUAL 4 OR CMAKE_SYSTEM_PROCESSOR MATCHES "(arm64)|(ARM64)|(aarch64)|(armv8)")
+            set(LIBDIR "lib")
+        else()
+            set(LIBDIR "lib64")
+        endif()
 
-    set(BYPRODUCTS
-            "${LIBDIR}/${BYPRODUCT_PREFIX}ssl${BYPRODUCT_SUFFIX}"
-            "${LIBDIR}/${BYPRODUCT_PREFIX}crypto${BYPRODUCT_SUFFIX}"
+        # Define byproducts
+        set(BYPRODUCT_PREFIX "lib" CACHE STRING "" FORCE)
+        set(OPENSSL_BUILD_SHARED "NO" CACHE STRING "" FORCE)
+
+        if (WIN32)
+            set(BYPRODUCT_SUFFIX ".lib" CACHE STRING "" FORCE)
+        # Due to OpenSSL 3's static linking issue on x86 MacOS platform we make an exception to build a shared library instead
+        elseif (APPLE AND (CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64|amd64|AMD64"))
+            set(BYPRODUCT_SUFFIX ".dylib" CACHE STRING "" FORCE)
+            set(OPENSSL_BUILD_SHARED "YES" CACHE STRING "" FORCE)
+        else()
+            set(BYPRODUCT_SUFFIX ".a" CACHE STRING "" FORCE)
+        endif()
+
+        set(BYPRODUCTS
+                "${LIBDIR}/${BYPRODUCT_PREFIX}ssl${BYPRODUCT_SUFFIX}"
+                "${LIBDIR}/${BYPRODUCT_PREFIX}crypto${BYPRODUCT_SUFFIX}"
+                )
+
+        if (OPENSSL_BUILD_SHARED)
+            set(OPENSSL_SHARED_FLAG "" CACHE STRING "" FORCE)
+        else()
+            set(OPENSSL_SHARED_FLAG "no-shared" CACHE STRING "" FORCE)
+        endif()
+
+        set(OPENSSL_BIN_DIR "${BINARY_DIR}/thirdparty/openssl-install" CACHE STRING "" FORCE)
+
+        FOREACH(BYPRODUCT ${BYPRODUCTS})
+            LIST(APPEND OPENSSL_LIBRARIES_LIST "${OPENSSL_BIN_DIR}/${BYPRODUCT}")
+        ENDFOREACH(BYPRODUCT)
+
+        if (OPENSSL_BUILD_SHARED)
+            install(FILES ${OPENSSL_LIBRARIES_LIST} DESTINATION bin COMPONENT bin)
+        endif()
+
+        # Set build options
+        set(OPENSSL_CMAKE_ARGS ${PASSTHROUGH_CMAKE_ARGS}
+                "-DCMAKE_INSTALL_PREFIX=${OPENSSL_BIN_DIR}"
+                "-DCMAKE_POLICY_DEFAULT_CMP0063=NEW"
+                # avoid polluting the global namespace, otherwise could interfere with the system openssl (e.g. python script extension using numpy)
+                "-DCMAKE_C_VISIBILITY_PRESET=hidden"
+                "-DCMAKE_CXX_VISIBILITY_PRESET=hidden"
+                "-DCMAKE_VISIBILITY_INLINES_HIDDEN=ON"
+                )
+
+        # Note: when upgrading to a later release than 3.1.1 the --no-apps could be used instead of --no-tests to minimize the build size
+        if (WIN32)
+            ExternalProject_Add(
+                    openssl-external
+                    URL https://github.com/openssl/openssl/releases/download/openssl-3.3.0/openssl-3.3.0.tar.gz
+                    URL_HASH "SHA256=53e66b043322a606abf0087e7699a0e033a37fa13feb9742df35c3a33b18fb02"
+                    SOURCE_DIR "${BINARY_DIR}/thirdparty/openssl-src"
+                    BUILD_IN_SOURCE true
+                    CONFIGURE_COMMAND perl Configure "CFLAGS=${PASSTHROUGH_CMAKE_C_FLAGS}" "CXXFLAGS=${PASSTHROUGH_CMAKE_CXX_FLAGS}" ${OPENSSL_SHARED_FLAG} no-tests "--prefix=${OPENSSL_BIN_DIR}" "--openssldir=${OPENSSL_BIN_DIR}"
+                    BUILD_BYPRODUCTS ${OPENSSL_LIBRARIES_LIST}
+                    EXCLUDE_FROM_ALL TRUE
+                    BUILD_COMMAND nmake
+                    INSTALL_COMMAND nmake install
+                )
+        else()
+            ExternalProject_Add(
+                    openssl-external
+                    URL https://github.com/openssl/openssl/releases/download/openssl-3.3.0/openssl-3.3.0.tar.gz
+                    URL_HASH "SHA256=53e66b043322a606abf0087e7699a0e033a37fa13feb9742df35c3a33b18fb02"
+                    SOURCE_DIR "${BINARY_DIR}/thirdparty/openssl-src"
+                    BUILD_IN_SOURCE true
+                    CONFIGURE_COMMAND ./Configure "CFLAGS=${PASSTHROUGH_CMAKE_C_FLAGS} -fPIC" "CXXFLAGS=${PASSTHROUGH_CMAKE_CXX_FLAGS} -fPIC" ${OPENSSL_SHARED_FLAG} no-tests "--prefix=${OPENSSL_BIN_DIR}" "--openssldir=${OPENSSL_BIN_DIR}"
+                    BUILD_BYPRODUCTS ${OPENSSL_LIBRARIES_LIST}
+                    EXCLUDE_FROM_ALL TRUE
             )
+        endif()
 
-    if (OPENSSL_BUILD_SHARED)
-        set(OPENSSL_SHARED_FLAG "" CACHE STRING "" FORCE)
-    else()
-        set(OPENSSL_SHARED_FLAG "no-shared" CACHE STRING "" FORCE)
-    endif()
+        # Set variables
+        set(OPENSSL_FOUND "YES" CACHE STRING "" FORCE)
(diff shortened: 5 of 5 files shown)


cmake/MiNiFiOptions.cmake
```diff
@@ -42,6 +42,8 @@ add_minifi_option(DOCKER_SKIP_TESTS "Skip building tests in docker image targets
 add_minifi_option(DOCKER_PUSH "Push created images to the specified tags" OFF)
 
 add_minifi_option(PORTABLE "Instructs the compiler to remove architecture specific optimizations" ON)
+add_minifi_option(USE_CONAN_PACKAGER "Instructs cmake to use conan packager to install external libs & build MiNiFi C++" OFF)
```
martinzink, line 45: Since there is only two exclusive options as of now, it doesnt make sense to create two separate options here.

I suggest (for now) to only have a single MINIFI_USE_CONAN_PACKAGER option, and replace the if elseif with a simple if else in the cmake files

And please prefix the new option with MINIFI_ (it makes easier to differentiate our options with thirdparty ones)

szaszm:
The macOS CI failure will be fixed in #1790, and the Docker integration test failure in #1795. The clang job fails because of a linter issue in the conanfile.

cmake/MiNiFiOptions.cmake
james94: I incorporated your feedback into commit [cf2face](https://github.com/apache/nifi-minifi-cpp/pull/1793/commits/cf2facee3a1ecc2909d439a4fe9cd1ca60830617)

cmake/BundledZLIB.cmake
```diff
@@ -18,56 +18,65 @@
 function(use_bundled_zlib SOURCE_DIR BINARY_DIR)
```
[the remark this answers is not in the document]
james94: I incorporated your feedback into commit [cf2face](https://github.com/apache/nifi-minifi-cpp/pull/1793/commits/cf2facee3a1ecc2909d439a4fe9cd1ca60830617)

I do like your suggestion, it makes the MiNiFi CMake code look cleaner with the conan v2.3.1 integration.

james94:
@szaszm If I get access to a macOS, I can looking into fixing the linter issue in the conanfile. Currently, I am testing MiNiFi from Ubuntu 22.04. I can later test it in Windows too if needed. For now, I addressed @martinzink suggested updates for integrating conan into MiNiFi build infrastructure.

szaszm:
@james94 You don't need a macOS to fix the python linter issue, it's not OS-specific. Just run this from a build directory:
```
cmake --build . --target flake8
```
And make sure `flake8` is installed.

szaszm:
@james94 what docker environment do you mean? I was trying on arch linux, no containers.

james94:
@szaszm I meant we could leverage a docker environment like the one that is mentioned in the README and extend it with conan2 support: https://github.com/apache/nifi-minifi-cpp?tab=readme-ov-file#building-a-docker-image

I usually just launch my own docker container with needed build dependencies for MiNiFi C++, so I can build it within my docker container.

james94:
@szaszm @martinzink @lordgamez just checking in, anything I can do to help speed up testing toward this PR?

martinzink (approved):
Sorry for the late reply, I was on holiday.
It would be nice if the minifi options were configurable through the conan options so we could  set the extensions without editing conanfile.py, but overall I like the direction (and that could be implemented in followup PRs) :+1:

james94:
Yes I agree with you @martinzink 
I do plan to add being able to configure minifi options mapping to conan options through the bootstrap py script, but that will be in a later PR, so we keep this initial PR light as @szaszm suggested.

cmake/MiNiFiOptions.cmake
```diff
@@ -42,6 +42,7 @@ add_minifi_option(DOCKER_SKIP_TESTS "Skip building tests in docker image targets
 add_minifi_option(DOCKER_PUSH "Push created images to the specified tags" OFF)
 
 add_minifi_option(PORTABLE "Instructs the compiler to remove architecture specific optimizations" ON)
+add_minifi_option(MINIFI_USE_CONAN_PACKAGER "Instructs cmake to use conan packager to install external libs & build MiNiFi C++" OFF)
```
szaszm, line 45: Instead of a single option, I think setting the package source per dependency would be better. I'm thinking about new CMake options like `MINIFI_LIBCURL_SOURCE` with possible values like `system` (find_package), `conan` or `build` (ExternalProject, FetchContent, etc). If you prefer not adding `system` now, that's fine.

You can see an example option with a predetermined set of possible values on line 76-78.

conanfile.py
szaszm, line 1: I think we should leave all feature selection options on the defaults, and use the normal FetchContent/ExternalProject build workflows for the rest of the dependencies. This would keep conan builds useful during the transition period, by offering a larger feature set, and it would allow us to track the reduction in compile-times as more dependencies are converted.

The above suggested package source options could be an exception, and they can be set here to Conan. I think it's a safe bet that if someone is calling CMake with Conan, they probably want the dependencies acquired through Conan.

What do you think?

szaszm:
I ran the following commands:
```
conan profile detect
export MINIFI_HOME=$(pwd)  # I was in the source tree root
conan install . --build=missing --output-folder=build_conan
conan build . --output-folder=build_conan
```

`conan install` ran suspiciously quickly, I don't think it had to build anything from source, but it did install zlib/1.3.1, openssl/3.2.1 and libcurl/8.6.0.

Unfortunately OpenSSL and Zlib were still built from source using ExternalProject during the CMake build. I think when calling CMake through Conan, they should be using the Conan packages.

CMakeLists.txt
szaszm, line 1: Another issue now, during `conan build`:
```
CMake Error at /home/szaszm/.local/lib/python3.10/site-packages/cmake/data/share/cmake-3.26/Modules/CMakeDetermineSystem.cmake:154 (message):
  Could not find toolchain file: conan_toolchain.cmake
Call Stack (most recent call first):
  CMakeLists.txt:5 (project)
```

I'm commenting this on a file so you can reply in a thread.

szaszm: I don't have that conan profile in the minifi source tree, so I just used the default from conan detect. It's the default of the environment, so I guess it shouldn't matter.

james94: Thanks for the heads up on the **conan profile** not being in the PR, I just realized **.gitignore** prevented it from showing up cause one of the folders got caught in the ignore. I updated the path to **release-linux** and now its in `nifi-minifi-cpp/etc/conan/profiles/release-linux`. You should be able to use it as you run **conan install** and **conan build** commands @szaszm 

Also I think the default conan profile should work out of box, but to be consistent, I would suggest you also give the **release-linux** conan profile a try and see if you get similar results that I got for **conan install and conan build**. I have also attached logs for your reference.

[minifi_conan_install_output_jguzman_060524.txt](https://github.com/user-attachments/files/15597404/minifi_conan_install_output_jguzman_060524.txt)

[minifi_conan_build_output_jguzman_060524.txt](https://github.com/user-attachments/files/15597409/minifi_conan_build_output_jguzman_060524.txt)

james94: @szaszm I just realized I had a forgot to update the **USE_CONAN_PACKAGER** in the conanfile.py to be **MINIFI_USE_CONAN_PACKAGER** variable. Thats why we were seeing the OpenSSL, ZLib, and LibCURL being built from source instead of using the conan prebuilt versions. I just updated the conanfile.py.

I have shared the output from when I ran `conan install . --build=missing --output-folder=build_conan` similar to how you did using the **default conan profile**. I also attached the output from running the conan build. You'll see this time, MiNiFi conan cmake uses conan's prebuilt packages for OpenSSL, ZLib and LibCURL instead of building them from source.

[conan_install_default_profile_jguzman_060824.log](https://github.com/user-attachments/files/15749397/conan_install_default_profile_jguzman_060824.log)

[conan_build_default_profile_jguzman_060824.log](https://github.com/user-attachments/files/15749398/conan_build_default_profile_jguzman_060824.log)

cmake/MiNiFiOptions.cmake
james94: @szaszm I think I see where you are talking about with a custom option that has multiple possible values. This is the example I found in MiNiFiOptions.cmake

~~~cmake
# Option: STRICT_GSL_CHECKS
# AUDIT: Enable all checks, including gsl_ExpectsAudit() and gsl_EnsuresAudit()
# ON: Enable all checks, excluding gsl_ExpectsAudit() and gsl_EnsuresAudit() (GSL default)
# DEBUG_ONLY: Like ON in the Debug configuration, OFF in others (MiNiFi C++ default)
# OFF: Throw on contract checking and assertion failures instead of calling std::terminate()
set(STRICT_GSL_CHECKS "DEBUG_ONLY" CACHE STRING "Contract checking and assertion failures call terminate")
list(APPEND STRICT_GSL_CHECKS_Values AUDIT ON DEBUG_ONLY OFF)
set_property(CACHE STRICT_GSL_CHECKS PROPERTY STRINGS ${STRICT_GSL_CHECKS_Values})
~~~

I agree I think it would be great to create new the CMake options for the source of how we retrieve external libs like the following:

~~~cmake
# Option: MINIFI_LIBCURL_SOURCE
# BUILD: MiNiFi LibCURL source retrieved from CMake's ExternalProject, FetchContent, etc
# CONAN: MiNiFi LibCURL source retrieved from conan packager
# SYSTEM: MiNiFi LibCURL source retrieved from native OS system
set(MINIFI_LIBCURL_SOURCE "BUILD" CACHE STRING "Builds LibCURL from source using standalone CMake approach"
list(APPEND MINIFI_LIBCURL_SOURCE_Values SYSTEM BUILD CONAN)
set_property(MINIFI_LIBCURL_SOURCE PROPERTY STRINGS ${MINIFI_LIBCURL_SOURCE_Values})
~~~

james94: I will make the update to this PR for **OpenSSL, LibCURL, ZLib** @szaszm

conanfile.py
james94: I agree with your suggestion for now that for all the feature selection options that are not **OpenSSL, CURL, ZLib**, we go with the default selected options currently in MiNiFiOptions.cmake and for now we use the normal **FetchContent/ExternalProject** for the rest of MiNiFi C++ external lib dependencies. 

Yes sounds good. For now, I will leave the default enabled MiNiFi Options set as they are in MiNiFiOptions.cmake and remove the **tc.variables[...]** ones from conanfile.py. We will only bring in **tc.variables[...]** variables when we are overriding one of the existing MiNiFi options from the CMake side with conan. This will only be relevant for the new MiNiFi options we discussed above for example **MINIFI_LIBCURL_SOURCE** and the other 2 libs sources.

james94:
@szaszm @martinzink  with respect to commit **"[49de56c](https://github.com/apache/nifi-minifi-cpp/pull/1793/commits/49de56cd6a559a567e4df7e6037473f8eebd74f6)"**: I verified I can build MiNiFi C++ using conan with the specific conan profile `etc/conan/profiles/release-linux` and with the sources for MINIFI_LIBCURL_SOURCE, MINIFI_OPENSSL_SOURCE
and MINIFI_ZLIB_SOURCE all set to CONAN successfully. There were 2 ENABLED options that I needed to set to OFF, which were ENABLE_LIBARCHIVE and ENABLE_AWS since the build would fail, but later I can get those working again since on my bigger PR-1775 "https://github.com/apache/nifi-minifi-cpp/pull/1775", I was able to keep those options ENABLED when building MiNiFi C++ with conan. I used the following command to build MiNiFi C++ with conan:

~~~bash
conan build . --build=missing --output-folder=build_conan -pr=etc/conan/profiles/release-linux
~~~

I did double check I can still build MiNiFi C++ using standalone CMake approach with all the default ENABLED options set to their original values as specified in MiNiFiOptions.cmake too. I used these commands to build MiNiFi C++ with cmake:

~~~bash
mkdir build_cmake
cd build_cmake

cmake ..
make -j $(nproc)
~~~

Please try the conan build MiNiFi C++ approach again when you get a chance

lordgamez:
@james94 I checked and it seems to build successfully for me. Could you please rebase the PR to the latest main brach to be able to run the CI on it?

james94:
@lordgamez thanks for verifying you can build MiNiFi C++ with conan on your side. I appreciate it. I will rebase on latest MiNiFi main branch.

james94:
@lordgamez @szaszm I merged latest upstream main branch into my forked main branch and then merged my forked main branch into my MINIFICPP-2346-P1 branch. Initially, I had probably close to **200** changes and after merging in upstream's main branch to my MINIFICPP-2346-P1 branch, the changes went up to **1,846** changes.  Is this okay? 

- Also I chose not to rebase and instead went for the merge approach since rebasing in the passed caused me to lose my updates on previous branches.

lordgamez:
Hi @james94, thanks for the update! Unfortunately the lua.org page where we retrieved the lua sources went down, that's the reason the CI jobs are failing. We merged a commit to move from lua.org to the github page to retrieve the source files with the following commit: https://github.com/apache/nifi-minifi-cpp/commit/b0b2cf63d0cebf32b71608d9337520b8c1f939cf Could you rebase it again to be able to run the CI correctly?

szaszm:
@james94 I could rebase your branch easily without data loss. The existing merge complicates things a bit, but not too much. Here's what I did:
```
git checkout main
git pull upstream main  # update local main branch from apache/nifi-minifi-cpp
git checkout MINIFICPP-2346-P1
git reset --hard HEAD~1  # remove merge commit from the top
git rebase -i main  # interactive rebase
# then edited the file, removing Gabor's commits, which were the first 6 IIRC, like here on the "commits" tab. "Making OpenSSL and curl mandatory", + 5x "review update" commits. At the end, exited the editor and git did the rebase.
# at the end, there is a successful rebase to the latest main, with no conflicts, and only your commits.
```

Thanks for sharing the conan profile at `etc/conan/profiles/release-linux`. I have concerns about the hardcoded compiler version and "system package manager" settings, but I don't know if this causes any real issues.

Tried compiling on a half-upgraded Ubuntu 22.04 to 24.04, the system packages should already be the 24.04 versions. I'll try again later with another system. I encountered these issues:
```
CMake Error at <line in CMakeLists.txt> (target_link_libraries):                                                                                                                                                                                                      
  Target "<target>" links to:                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                 
    zstd::zstd                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                 
  but the target was not found.  Possible reasons include:                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                 
    * There is a typo in the target name.                                                                                                                                                                                                                        
    * A find_package call is missing for an IMPORTED target.                                                                                                                                                                                                     
    * An ALIAS target is missing.                                                                                                                                                                                                                                
```
Affected targets are rocksdb, db_bench, db_sanity_test, write_stress, db_repl_stress, rocksdb_dump, rocksdb_undump, db_stress. It looks like all of them are RocksDB-related.

Also, I would appreciate if you could add the necessary changes to keep the AWS and libarchive extensions enabled and working, so that we can eventually merge this PR standalone without breaking important functionality.

james94:
Thanks @szaszm for the steps to run a git rebase without causing data loss. I will give it a try. I think the reason we are seeing the  `zstd::zstd` target not found issue is because we move from Ubuntu 22.04 to Ubuntu 24.04 and maybe **conancenter** doesn't provide the prebuilt conan package out of box. I think we just need to run these two conan commands:

~~~bash
conan install . --build=missing --output-folder=build_conan
conan build . --output-folder=build_conan
~~~

I will look into it today after work. Thanks again for sharing your steps.

james94:
@szaszm @lordgamez for this latest git push, I followed marton's steps to do an interactive rebase and also removed the top merge commit that I had to clean up the history and leave only my commits for the git history. Now PR-1793's branch MINIFICPP-2346-P1 should be on top of the latest MiNiFi main branch. I also verified I could build MiNiFi using conan with conan's cmake wrapper and then I followed up by verifying I could still build MiNiFi using standalone CMake. Also it looked like the TESTS were on there way to passing too. When you guys get a chance, please review.

conanfile.py
```diff
@@ -0,0 +1,37 @@
+from conan import ConanFile
+from conan.tools.env import VirtualRunEnv
+from conan.tools.cmake import CMake, CMakeToolchain
+
+required_conan_version = ">=2.1.0"
```
szaszm, line 5: Can we lower the Conan version requirement to 2.0? I'd like to use Nix to install Conan, but they only package 2.0.17, which doesn't satisfy the 2.1 requirement. :/
Last time I tried, pip messed up my path with broken CMake binaries.

conanfile.py
szaszm, line 1: the ubuntu clang CI failure:
image

Please check the flake8 linter before committing changes to python files.

szaszm, line 1: After a few tries, I got a working build, but there is no output package. Is there any way to change the default CMake target from `all` to `package`, so that a binary tarball is created by the build command? Or is there a better way of packaging minifi with Conan?

james94: Glad to hear you got a working build. As we provide more flexiblity for the users to configure MiNiFi to be built with more external libraries managed by conan, and integration into the bootstrap.py, the process to build a working MiNiFi instance will hopefully be a lot smoother.

For clarity on your question, I think your asking about during the MiNiFi build process in standalone CMake, we usually run `make package` to create an output package and then for conan approach, you're asking is there a way to change the default CMake target from `all` to `package`? 

- We can use conan's CMakeToolchain to pass it `tc.variables["<CMake_Definition>"]` that we can check in CMake, which could be used to potentially change the default CMake target from `all` to `package`.

Another place we can reference from Conan v2 is their documentation on the supported features they have for their cmake wrapper:

- https://docs.conan.io/2/reference/tools/cmake/cmake.html#cmake

conanfile.py
james94: Sorry to hear that pip messed up your path with broken CMake binaries. Yeah of course we can look into lowering the conan version to 2.0. I will also try out Nix too.

conanfile.py
james94: Thanks for the reminder. I will make sure to try out checking flake8 linter on the next time that I commit changes to the conanfile.py and related py files.

szaszm: I would change the default CMake target to `package` when building with Conan. When using CMake directly, the user can just call `make` or `make package` as they wish. But since it's not possible, or not immediately obvious how to do this with `conan build`, I'd do a build that's as complete as possible.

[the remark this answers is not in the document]
james94: I think it would be possible to have vcpkg and conan exist together within MiNiFi C++ infrastructure where the user can choose the packager they want to use to build MiNiFi. The great news since we're using conan v2 is that the CMake files use all the standalone CMake context, which should allow for integration of vcpkg and not cause a conflict with conan.

james94: I believe conan's equivalent to CMake's `make package` is leveraging conan's ConanFile class's **package(...)** method and making sure it comes with all the headers, binaries, etc that would be included when running `make package`. I will check MiNiFi's `make package` to verify that ConanFile class's **package(...)** creates a MiNiFi C++ conan package with the similar assets. For reference, here is my follow up PR that works toward this effort:

- "MINIFICPP-2346-P2" **PR-1813** Follow Up to "MINIFICPP-2346-P1": https://github.com/apache/nifi-minifi-cpp/pull/1813
- You'll notice a lot of changes in the diff, but that is because I haven't rebased on top of latest main branch, which I will do once we get this PR merged.

I can incorporate the ConanFile class's **package()** here to add the ability to run `conan create`, which is similar to `make package`. I will try to keep the changes to a minimum too.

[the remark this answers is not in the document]
james94: @szaszm I will come back to address the feedback this week. I will get back to you soon with the conan create package update, which is similar to cmake make package.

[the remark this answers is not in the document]
james94: I added ability to create **conan package** for **MiNiFi C++**, which is similar to CMake's **make package**.  I updated the steps at the top of my PR above to include conan create command. Please give it a try when you get a chance @szaszm

szaszm:
Please make sure the CI workflows are green. I approved their run now, feel free to ping me on slack if you need more approvals. Or let me know if you don't see their output, and I'll post it here then.

CMakeLists.txt
```diff
@@ -593,9 +596,11 @@ set(CPACK_PACKAGE_VENDOR "Apache NiFi")
...
 set(CPACK_PACKAGE_VERSION_MINOR ${PROJECT_VERSION_MINOR})
 set(CPACK_PACKAGE_VERSION_PATCH ${PROJECT_VERSION_PATCH})
-set(CPACK_PACKAGE_DESCRIPTION_FILE "${CMAKE_CURRENT_SOURCE_DIR}/README.md")
-configure_file("${CMAKE_CURRENT_SOURCE_DIR}/LICENSE" "${CMAKE_CURRENT_BINARY_DIR}/LICENSE.txt" COPYONLY)
-set(CPACK_RESOURCE_FILE_LICENSE "${CMAKE_CURRENT_BINARY_DIR}/LICENSE.txt")
+if(NOT MINIFI_BUILD_CONAN_PACKAGE)
+    set(CPACK_PACKAGE_DESCRIPTION_FILE "${CMAKE_CURRENT_SOURCE_DIR}/README.md")
+    configure_file("${CMAKE_CURRENT_SOURCE_DIR}/LICENSE" "${CMAKE_CURRENT_BINARY_DIR}/LICENSE.txt" COPYONLY)
+    set(CPACK_RESOURCE_FILE_LICENSE "${CMAKE_CURRENT_BINARY_DIR}/LICENSE.txt")
+endif()
```
szaszm, line 603: Why are these disabled in conan packages? We need to include the LICENSE and NOTICE files in every distribution, including binary ones.

szaszm (commented):
The code changes mostly look good, I'll do some more testing later to see how things work in practice

CMakeLists.txt
james94: When we create the **MiNiFi C++ conan package**, **LICENSE** and **NOTICE** are in the conan package. In the **conanfile.py**, we tell conan to include all the files assigned to **shared_sources** variable to be included in the conan package. In our conanfile.py's **package()** method, we also run **cmake.install()**, so in MiNiFi C++'s CMakeLists.txt files all CMake commands using **install(...)** will also copy over those sources (.h*, .i*, etc) and compiled binary assets (static libs, shared libs, minifi binary, etc) into our MiNiFi C++ conan package.

In the past when I tried maintaining that CPACK part way of doing it, when I ran **conan create**, it complained about those 3 lines. However, today with **conan 2.0.17** I removed that MINIFI_BUILD_CONAN_PACKAGE check and was able to successfully run **conan create** with those 3 lines enabled. I will push the update commit up to the repo now.

james94: Heres a view of my minifi cpp conan package p/ folder where you can see **LICENSE, NOTICE** and other assets copied over.

minifi_cpp_conan_package_folder_view_aug92024

conanfile.py
```diff
@@ -0,0 +1,115 @@
...
+
+        tc.variables["SKIP_TESTS"] = "OFF"
+        tc.variables["ENABLE_OPENWSMAN"] = "ON"
+        tc.variables["ENABLE_CIVET"] = "ON"
+        tc.variables["ENABLE_EXPRESSION_LANGUAGE"] = "ON"
+        tc.variables["ENABLE_BZIP2"] = "ON"
+        tc.variables["ENABLE_ROCKSDB"] = "ON"
+        tc.variables["BUILD_ROCKSDB"] = "ON"
+
+        tc.variables["ENABLE_LIBARCHIVE"] = "OFF"
```
james94, line 55: For ease of creating the MiNiFi C++ conan package, I disabled the following libraries LIBARCHIVE to CONTROLLER. For a follow up PR, I can work on making sure that each of these features can come ENABLED or DISABLED when creating the MiNiFi C++ conan package. For now, I think this is fine. What are your thoughts @szaszm ?

james94:
@szaszm I tried out running the flake8 linter from build_cmake/ directory:

~~~bash
cd $HOME/nifi-minifi-cpp

pip install flake8
export PATH="$PATH:~/.local/bin"
source ~/.bashrc

cmake --build . --target flake8
~~~

I get the following output (just showing the part related to conanfile.py:

~~~bash
/home/ubuntu/src/james/pipeline/nifi-minifi-cpp-develop/conanfile.py:2:1: F401 'conan.tools.env.VirtualRunEnv' imported but unused
/home/ubuntu/src/james/pipeline/nifi-minifi-cpp-develop/conanfile.py:14:99: W291 trailing whitespace
/home/ubuntu/src/james/pipeline/nifi-minifi-cpp-develop/conanfile.py:16:105: W291 trailing whitespace
/home/ubuntu/src/james/pipeline/nifi-minifi-cpp-develop/conanfile.py:22:1: E302 expected 2 blank lines, found 1
/home/ubuntu/src/james/pipeline/nifi-minifi-cpp-develop/conanfile.py:31:53: E231 missing whitespace after ','
/home/ubuntu/src/james/pipeline/nifi-minifi-cpp-develop/test_package/conanfile.py:7:1: E302 expected 2 blank lines, found 1
gmake[3]: *** [CMakeFiles/flake8.dir/build.make:70: CMakeFiles/flake8] Error 1
gmake[2]: *** [CMakeFiles/Makefile2:2645: CMakeFiles/flake8.dir/all] Error 2
gmake[1]: *** [CMakeFiles/Makefile2:2652: CMakeFiles/flake8.dir/rule] Error 2
gmake: *** [Makefile:667: flake8] Error 2
~~~

szaszm:
flake is a linter, showing you the parts of the code that don't follow the predefined rules. Please fix them in the files you add or change. Let me know if you have any questions.

conanfile.py
szaszm: As long as this only affects the Conan build, and they will be re-enabled in later iterations, then I'm fine with it. But if we end up long-term with a limited Conan build that lacks essential features, then that's a maintenance burden that adds little value.

test_package/src/AbstractProcessorTest.cpp
```diff
@@ -0,0 +1,114 @@
+/**
```
james94, line 1: I realized I forgot to add this conan test_package/ folder that comes with an example **AbstractProcessorTest.cpp**, which is the same one from the MiNiFi GTESTs. This helps us test if features are available from the MiNiFi conan package. We can simulate if we were to add MiNiFi C++ library to a C++ project, would the features we need be available to it. For now, I just copied this AbstractProcessorTest.cpp GTEST example. This test_package/ comes with CMakeLists.txt and its own conanfile.py too.

conanfile.py
james94: Yes it only affects the Conan build part of MiNiFi. As a reminder, in my bigger MiNiFi Conan PR for MiNiFi, I already verified I could ENABLE most of these MiNiFi features using Conan.  I just didn't want this PR to be super large. Yes we can ENABLE more of MiNiFi features using Conan as we do more follow up PRs.

test_package/src/AbstractProcessorTest.cpp
szaszm: I think we should run the same unit tests as we would run normally, even in a Conan build. I'm not a fan of duplicating a single test and introducing a new test package for conan builds, it adds too much maintenance burden.

james94: We can update the conanfile.py for the test_package to try to run existing MiNiFi C++ GTESTs, so we dont duplicate the already created tests. I was mainly copying over a MiNiFi test here for a quick check. I will try **cmake.ctest()**, so I can get rid of this copied test program.

conanfile.py
james94: I updated the conanfile.py, so for more features, we go with MiNiFi C++'s default settings and now only these particular features are DISABLED:

~~~python
        tc.variables["ENABLE_LIBARCHIVE"] = "OFF"
        tc.variables["ENABLE_AWS"] = "OFF"
        tc.variables["ENABLE_SQL"] = "OFF"
        tc.variables["ENABLE_GCP"] = "OFF"
        tc.variables["ENABLE_LUA_SCRIPTING"] = "OFF"
~~~

In a follow up PR, I will add support for ENABLING them using conan while still providing support for standalone CMake.

test_package/src/AbstractProcessorTest.cpp
james94: I decided to leave out the **cmake.ctest()** from being run automatically after conan finishes building MiNiFi C++, so the user will still need to go into **build_conan** folder and then run **ctest**. For the conan build and conan create approaches for MiNiFi C++, **217/219** TESTs PASS. For standalone CMake approach, **275/275** TESTs PASS.

cmake/MiNiFiOptions.cmake
james94: Currently in our conanfile.py, we have the following MINIFI external libraries sources set to CONAN:

~~~python
        tc.variables["MINIFI_LIBCURL_SOURCE"] = "CONAN"
        tc.variables["MINIFI_OPENSSL_SOURCE"] = "CONAN"
        tc.variables["MINIFI_ZLIB_SOURCE"] = "CONAN"
        tc.variables["MINIFI_CIVETWEB_SOURCE"] = "CONAN"
        tc.variables["MINIFI_LIBXML2_SOURCE"] = "CONAN"
        tc.variables["MINIFI_FMT_SOURCE"] = "CONAN"
        tc.variables["MINIFI_SPDLOG_SOURCE"] = "CONAN"
        tc.variables["MINIFI_CATCH2_SOURCE"] = "CONAN"
~~~

james94:
When you guys get a chance, can one of you approve the CI/CD workflow, so I can check out the result? @lordgamez @martinzink @szaszm

lordgamez:
Hi @james94, I checked the build with Conan locally and it works fine for me. The only tests that are failing are the `ControllerTests` and the `LoggerTests`. The `ControllerTests` are failing because of the disabled LibArchive extension so that is understandable. The `LoggerTests` are failing with an `fmt` formatter test where the expected log formatting is `1m equals to 60s`, but we get `1min equals to 60s`. Could it be due to different versions used in our built `fmt` lib version and the `fmt` version pulled by Conan? Could you please check that out?

I really appreciate the step by step instructions for building the project using Conan, it would be great to have it as part of the documentation as well. Could you please add it to the documentation in README.md or maybe in a separate markdown file and a link to that in the README.md?

Thanks for all your effort!

james94:
Hi @lordgamez. I will look into `LoggerTests` and see if it is a difference in their expected log formatting and get back to you.

Also thats a great idea to add a section on building MiNiFi C++ using conan. I will consider creating a new markdown file and linking it to the README.md to save space.

Also I notice that MiNiFi is under NiFi projects. When we click on **Quick Start Guide** or **Administrator's Guide**, we are sent to MiNiFi github. Will we eventually just have this part of MiNiFi directly linked to the website's documentation? Is there interest in this effort?

james94:
Hi @lordgamez @szaszm , I added a section to the README.md on building MiNiFi and creating a MiNiFi conan package using conan. In the README.md in that conan section, it points to a CONAN.md that goes step by step on how to use conan with MiNiFi too.

I will check the `LoggerTests` part this week to see if the format is showing differently in conan fmt compared to CMake installed fmt because of the fmt version difference and get back to you.

james94:
Hi @lordgamez @szaszm 

I fixed the LoggerTests fmt formatting issue:
> 
> For conan, we install fmt 10.2.1 prebuilt conan package to not conflict with spdlog 1.14.0 prebuilt conan package.
> However, when we switched to a newer version of fmt, that caused the LoggerTest to fail, which didn't happen
> when we were using fmt 10.1.0 that was version installed when we used the standalone CMake approach. So, we updated
> our TEST_CASE for 'fmt formatting works with the logger' to explicity create the chrono minutes duration,
> followed by fmt format and then we pass our formatted minutes duration to our logger->log_critical, so we get
> the expected fmt formatting to be '1m'. We didn't have to do this explicitly when we used fmt 10.1.0 since even
> if we passed our logger->log_critical '1min', it would auto format to '1m'. For fmt 10.2.1 to have backward compatibility
> with fmt 10.1.0 and the result we expected from this TEST_CASE in LoggerTests, we made this update.
> 
> I checked that when I build MiNiFi using conan and then build MiNiFi using standalone CMake, their LoggerTest PASSES.

Please let me know if these updates to LoggerTests.cpp are sufficient. Then when you get another chance, can you review my PR and approve if you think its ready?

libminifi/test/unit/LoggerTests.cpp
```diff
@@ -380,7 +380,9 @@ TEST_CASE("Setting max log entry length to unlimited results in unlimited log en
 TEST_CASE("fmt formatting works with the logger") {
   LogTestController::getInstance().setTrace<logging::Logger>();
   std::shared_ptr<logging::Logger> logger = logging::LoggerFactory<logging::Logger>::getLogger();
-  logger->log_critical("{} equals to {}", 1min, 60s);
+  std::chrono::minutes duration{1};
```
james94, line 383: Made this update for compatibility between fmt 10.1.0 that is installed by standalone CMake approach and fmt 10.2.1 that is installed by conan approach. I verified after building MiNiFi using both approaches that the LoggerTest passes.

lordgamez:
> Hi @lordgamez @szaszm
> 
> I fixed the LoggerTests fmt formatting issue "[7def7a0](https://github.com/apache/nifi-minifi-cpp/pull/1793/commits/7def7a081390fa991302e5d4882a0eff494dbfa3)":
> 
> > For conan, we install fmt 10.2.1 prebuilt conan package to not conflict with spdlog 1.14.0 prebuilt conan package.
> > However, when we switched to a newer version of fmt, that caused the LoggerTest to fail, which didn't happen
> > when we were using fmt 10.1.0 that was version installed when we used the standalone CMake approach. So, we updated
> > our TEST_CASE for 'fmt formatting works with the logger' to explicity create the chrono minutes duration,
> > followed by fmt format and then we pass our formatted minutes duration to our logger->log_critical, so we get
> > the expected fmt formatting to be '1m'. We didn't have to do this explicitly when we used fmt 10.1.0 since even
> > if we passed our logger->log_critical '1min', it would auto format to '1m'. For fmt 10.2.1 to have backward compatibility
> > with fmt 10.1.0 and the result we expected from this TEST_CASE in LoggerTests, we made this update.
> > I checked that when I build MiNiFi using conan and then build MiNiFi using standalone CMake, their LoggerTest PASSES.
> 
> Please let me know if these updates to LoggerTests.cpp are sufficient. Then when you get another chance, can you review my PR and approve if you think its ready?

Thanks for the fix @james94 ! It works now and all the tests pass (besides ControllerTests which is expected at this point). The only problem I found was that some of the HTTP tests seem to work differently and are taking way too much time to finish, especially ListenHTTPTests. Could this be due to different curl versions maybe?

Test run when building from source:
201/203 Test  #68: VerifyInvokeHTTPPostTest .....................   Passed   18.38 sec
199/203 Test  #64: InvokeHTTPTests ..............................   Passed   15.90 sec
198/203 Test   #1: ListenHTTPTests ..............................   Passed   13.93 sec

Test run using conan:
217/219 Test  #85: VerifyInvokeHTTPPostTest .....................   Passed   33.06 sec
218/219 Test  #81: InvokeHTTPTests ..............................   Passed   38.32 sec
219/219 Test   #9: ListenHTTPTests ..............................   Passed  102.88 sec

cmake/GetZLIB.cmake
szaszm, line 1: There is some issue with the zlib conan build, I'm getting missing headers. Did you test compilation with the zlib headers not installed on the system?
```
In file included from /home/szaszm/nifi-minifi-cpp/build_conan/extensions/librdkafka/kafka-external-prefix/src/kafka-external/src/rdkafka_buf.h:32,
                 from /home/szaszm/nifi-minifi-cpp/build_conan/extensions/librdkafka/kafka-external-prefix/src/kafka-external/src/rdkafka_int.h:88,
                 from /home/szaszm/nifi-minifi-cpp/build_conan/extensions/librdkafka/kafka-external-prefix/src/kafka-external/src/rdavl.c:29:
/home/szaszm/nifi-minifi-cpp/build_conan/extensions/librdkafka/kafka-external-prefix/src/kafka-external/src/rdcrc32.h:57:10: fatal error: zlib.h: No such file or directory
   57 | #include <zlib.h>
      |          ^~~~~~~~
compilation terminated.
```

szaszm: It works when I install the headers, but we shouldn't depend on system headers for a conan package.

szaszm:
Regarding the tests Gábor mentioned:
- ControllerTests: apparently that fails because it depends on libarchive
- slower HTTP tests: We should fix them, but I'm OK with merging this PR and fixing it in followups, since it doesn't regress the existing CMake build.

james94:
@lordgamez thanks for looking into the HTTP related CURL tests. Yeah for the **conanfile.py** file, we use **libcurl** version **8.6.0** and for **standalone CMake**, we use **8.4.0**. It could be a **libcurl** version difference, but we can also consider in **standalone CMake**, we configure **libcurl** in **cmake/BundledLibcURL.cmake** before being built:

~~~bash
            "-DCMAKE_INSTALL_PREFIX=${BINARY_DIR}/thirdparty/curl-install"
            -DBUILD_CURL_EXE=OFF
            -DBUILD_TESTING=OFF
            -DBUILD_SHARED_LIBS=OFF
            -DHTTP_ONLY=ON
            -DCURL_CA_PATH=none
            -DCURL_USE_LIBSSH2=OFF
            -DUSE_LIBIDN2=OFF
            -DCURL_USE_LIBPSL=OFF
            -DCURL_USE_OPENSSL=ON
~~~

Whereas for **conan**, we install that **libcurl** prebuilt library using the default configs. If we do want to get closer to configuring conan's **libcurl** like we do for the standalone CMake, we can either make some updates to our **conanfile.py** or make updates to **libcurl's** conan recipe. Heres **libcurl's** conan recipe from conan-center-index repo and line 566 that is an example of how they set "**BUILD_CURL_EXE**" to False aka OFF:

- https://github.com/conan-io/conan-center-index/blob/master/recipes/libcurl/all/conanfile.py#L566

Do you think we see a difference in HTTP performance because of these configurations to **libcurl** build?

cmake/GetZLIB.cmake
james94: I just double checked the **conanfile.py**, it appears zlibs is not in conan package **shared_requires** tuple. Maybe it was removed by accident. I will add it back. Here will be the updated **shared_requires** tuple

~~~bash
shared_requires = ("openssl/3.2.1", "libcurl/8.6.0", "civetweb/1.16", "libxml2/2.12.6", "fmt/10.2.1", "spdlog/1.14.0", "catch2/3.5.4", "zlib/1.2.11")
~~~

We will try **zlib/1.2.11** for conan package to match the **zlib** system library version. @szaszm

james94: @szaszm When building MiNiFi with conan build, **217/219 ctests PASS**.

	 34 - ProvenanceTests (Subprocess aborted)
	219 - ControllerTests (Failed)

I think **ProvenanceTests** from RocksDB fails because of 

        double free or corruption (!prev)
        Test #34: ProvenanceTests ........ Subprocess aborted***Exception: 0.19 sec

In regards to this **ProvenanceTests**, the main difference between MiNiFi conan build and standalone CMake build is that in conan build we use **zstd** and **zlib** conan packages to try to link into CMake's system library installed **RocksDB** while with standalone CMake approach, we use CMake to **build and install zlib & zstd external system libs** while installing system library **RocksDB**

@szaszm any thoughts on what could cause **ProvenanceTest** to have this **double free or corruption (!prev)** error when building MiNiFi using conan? This doesn't happen when I build MiNiFi with standalone CMake. Also in my much larger PR where I added support to install **patched RocksDB conan package** based on **standalone CMake's patched RocksDB**, I dont think I saw this ctest error: https://github.com/apache/nifi-minifi-cpp/pull/1775

When building MiNiFi with standalone CMake, **all 275 ctests PASS.**

lordgamez:
> @lordgamez thanks for looking into the HTTP related CURL tests. Yeah for the **conanfile.py** file, we use **libcurl** version **8.6.0** and for **standalone CMake**, we use **8.4.0**. It could be a **libcurl** version difference, but we can also consider in **standalone CMake**, we configure **libcurl** in **cmake/BundledLibcURL.cmake** before being built:
> 
> ```shell
>             "-DCMAKE_INSTALL_PREFIX=${BINARY_DIR}/thirdparty/curl-install"
>             -DBUILD_CURL_EXE=OFF
>             -DBUILD_TESTING=OFF
>             -DBUILD_SHARED_LIBS=OFF
>             -DHTTP_ONLY=ON
>             -DCURL_CA_PATH=none
>             -DCURL_USE_LIBSSH2=OFF
>             -DUSE_LIBIDN2=OFF
>             -DCURL_USE_LIBPSL=OFF
>             -DCURL_USE_OPENSSL=ON
> ```
> 
> Whereas for **conan**, we install that **libcurl** prebuilt library using the default configs. If we do want to get closer to configuring conan's **libcurl** like we do for the standalone CMake, we can either make some updates to our **conanfile.py** or make updates to **libcurl's** conan recipe. Heres **libcurl's** conan recipe from conan-center-index repo and line 566 that is an example of how they set "**BUILD_CURL_EXE**" to False aka OFF:
> 
> * https://github.com/conan-io/conan-center-index/blob/master/recipes/libcurl/all/conanfile.py#L566
> 
> Do you think we see a difference in HTTP performance because of these configurations to **libcurl** build?

Hi @james94, I think both options are plausible to cause the difference in the test run, we should try to build from source with the newer version and with the default build options too and see if one of those builds behaves the same way as the conan build, and maybe check out the logs what causes this behavior. Maybe some timeouts are different, or some additional checks are added in the newer version, not sure. Unfortunately I didn't have time to try it out this week, maybe I'll have some free time to check these builds next week, or if you're up for it and have some time for it feel free to check it out.

lordgamez:
Hi @james94, I tried building the curl library from source using version 8.6.0 and the default build options and it seems to be working the same way as before, having the runtime around 11-12 seconds, so I'm not sure at the moment what's causing the difference with the conan build.

james94:
@lordgamez thanks for looking into the CURL external lib for MiNiFi C++ and comparing the difference between standalone CMake CURL built from source and the conan CURL prebuilt conan package. It is surprising that the conan CURL has a slower runtime than the standalone CMake CURL. I can look into it further this week and get back to you.

cmake/GetZLIB.cmake
james94: @lordgamez @szaszm for the conan build MiNiFi side, do we need to get ProvenanceTests working for this PR or could we get that Test passing for a follow up PR?

cmake/BundledRocksDB.cmake
```diff
@@ -63,10 +64,17 @@ function(use_bundled_rocksdb SOURCE_DIR BINARY_DIR)
                 -DROCKSDB_INSTALL_ON_WINDOWS=ON
                 -DWITH_XPRESS=ON)
     else()
+    if(MINIFI_ZSTD_SOURCE STREQUAL "CONAN" AND MINIFI_ZLIB_SOURCE STREQUAL "CONAN")
+        list(APPEND ROCKSDB_CMAKE_ARGS
+            -DWITH_ZLIB=OFF
+            -DWITH_ZSTD=OFF)
+    else()
```
szaszm, line 71: I think we should fix RocksDB with Conan zlib and zstd instead, if that's possible. Or is that not possible or feasible?

cmake/GetZstd.cmake
```diff
@@ -0,0 +1,27 @@
...
+# specific language governing permissions and limitations
+# under the License.
+
+function(get_zstd)
+    if(MINIFI_ZSTD_SOURCE STREQUAL "CONAN")
+        message("Using Conan Packager to manage installing prebuilt zstd external lib")
+        find_package(zstd REQUIRED)
+        add_library(zstd::zstd ALIAS zstd::libzstd_static)
+    elseif(MINIFI_ZSTD_SOURCE STREQUAL "BUILD")
+        message("Using CMAKE's ExternalProject_Add to manage source building zstd external lib")
```
szaszm, line 24:
```suggestion
        message("Using CMake to build zstd from source")
```

cmake/GetZstd.cmake
```diff
@@ -0,0 +1,27 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+function(get_zstd)
+    if(MINIFI_ZSTD_SOURCE STREQUAL "CONAN")
+        message("Using Conan Packager to manage installing prebuilt zstd external lib")
```
szaszm, line 20:
```suggestion
        message("Using Conan Packager to install zstd")
```

cmake/GetZLIB.cmake
```diff
@@ -0,0 +1,27 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+function(get_zlib SOURCE_DIR BINARY_DIR)
+    if(MINIFI_ZLIB_SOURCE STREQUAL "CONAN")
+        message("Using Conan Packager to manage installing prebuilt zlib external lib")
```
szaszm, line 20:
```suggestion
        message("Using Conan to install zlib")
```

cmake/GetZLIB.cmake
```diff
@@ -0,0 +1,27 @@
...
+        find_package(ZLIB REQUIRED)
+    elseif(MINIFI_ZLIB_SOURCE STREQUAL "BUILD")
+        message("Using CMAKE's ExternalProject_Add to manage source building zlib external lib")
```
szaszm, line 23:
```suggestion
        message("Using CMake to build zlib from source")
```

etc/conan/profiles/release-linux
```diff
@@ -0,0 +1,12 @@
+[settings]
+os=Linux
+arch=x86_64
+build_type=Release
+compiler=gcc
+compiler.cppstd=gnu20
```
szaszm, line 6: We generally prefer ISO standard over GNU.
```suggestion
compiler.cppstd=20
```

szaszm (commented):
Hopefully the last round of comments. I'm not insisting on fixing the curl slowdown right now in this PR, but I consider that a prerequisite before we start treating the Conan build as stable and production-ready.

The Windows CI failure was aws-sdk-cpp not building with the newest MSVC, but it was fixed in #1862. It will work after a rebase.

cmake/BundledRocksDB.cmake
james94: It is possible to fix RocksDB with conan zlib and zstd. I will try to add support for RocksDB conan as alternative to RocksDB cmake, so we can undo the updates that I made to BundledRocksDB.cmake, so that BundledRocksDB.cmake just stays standalone CMake specific and separate from conan. I will follow up on this.

james94: @szaszm @lordgamez 
I just added a **RocksDB** conan recipe, so we can create that conan package and build it with the same patches that we use to build RocksDB using the BundledRocksDB approach. However, now **RocksDB** conan package is separate from RocksDB **BundledRocksDB**, which uses standalone CMake. 

- I will make sure to update the conan steps in the PR's description from earlier. 
- I already added this update into the **CONAN.md** by adding an appendix section that shows how to create a custom conan package, which in our case is **RocksDB** due to needing to apply patches during the conan package creation.

With **RocksDB** as a conan package, it uses **zlib** and **zstd** conan packages when creating the **RocksDB** conan package.

Thus, I switched **BundledRocksDB** back to its original self based on **main** branch before I added those conditional statements that accounted for **Conan**.

Now when we use **RocksDB**, we run **get_rocksdb()** to retrieve **RocksDB** using **Conan** or **standalone CMake** depending on how the build is configured.

After this update, I rebuilt **MiNiFi** using conan, reran the CTESTs **218** out of **219** passed.
I rebuilt **MiNiFi** using standalone CMake, reran the CTESTs **218** out of **219** passed.

cmake/GetZLIB.cmake
james94: I rebuilt MiNiFi after rebasing it on the latest **main** branch, I dont see the **ProvenanceTests** failing that I saw last time. My docker environment has the MiNiFi apt package dependencies installed and here was the Ubuntu `cat /etc/os-release` info:

~~~bash
PRETTY_NAME="Ubuntu 22.04.3 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.3 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
~~~

cmake/BundledRocksDB.cmake
james94: Here are the conan steps for quick reference @szaszm @lordgamez :

~~~bash
sudo pip install --force-reinstall -v "conan==2.0.17"

# conanfile.py is in root dir of MiNiFi C++ project
cd $HOME/nifi-minifi-cpp

# create MINIFI_HOME env variable for binary executable minifi
export MINIFI_HOME=$(pwd)

# In case you need to delete all conan packages
# conan remove "*" -c

# Create custom RocksDB conan package with MiNiFi's patches applied
pushd thirdparty/rocksdb/all

conan create . --user=minifi --channel=develop --version=8.10.2 -pr=../../../etc/conan/profiles/release-linux

popd # Exit thirdparty/rocksdb/all

# install conan packages for MiNiFi C++ using conanfile.py invoking Conan
# since we created default profile earlier, we can override it with our own minifi profile
# make sure path is correct; Our conanfile.py also includes our custom RocksDB conan package
conan install . --build=missing --output-folder=build_conan -pr=$HOME/nifi-minifi-cpp/etc/conan/profiles/release-linux

# build MiNiFi C++ using conanfile.py invoking Conan & CMake
conan build . --output-folder=build_conan -pr=$HOME/nifi-minifi-cpp/etc/conan/profiles/release-linux
~~~

cmake/GetZstd.cmake
james94: I updated each Get<lib>.cmake based on your suggestion in my latest commit. Thanks.

CMakeLists.txt
```diff
@@ -221,9 +221,17 @@ else()
...
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/ssl")
+
+
 if (ENABLE_BZIP2 AND (ENABLE_LIBARCHIVE OR (ENABLE_ROCKSDB AND NOT WIN32)))
-    include(BundledBZip2)
-    use_bundled_bzip2(${CMAKE_SOURCE_DIR} ${CMAKE_BINARY_DIR})
+    include(GetBZip2)
+    get_bzip2(${CMAKE_SOURCE_DIR} ${CMAKE_BINARY_DIR})
+    # include(BundledBZip2)
+    # use_bundled_bzip2(${CMAKE_SOURCE_DIR} ${CMAKE_BINARY_DIR})
```
szaszm, line 234: these lines can be deleted
```suggestion
```

cmake/GetCatch2.cmake
```diff
@@ -0,0 +1,27 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+function(get_catch2)
+    if(MINIFI_CATCH2_SOURCE STREQUAL "CONAN")
+        message("Using Conan Packager to manage installing prebuilt Catch2 external lib")
```
szaszm, line 20:
```suggestion
        message("Using Conan to install Catch2")
```

cmake/GetCivetWeb.cmake
```diff
@@ -0,0 +1,26 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+function(get_civetweb)
+    if(MINIFI_CIVETWEB_SOURCE STREQUAL "CONAN")
+        message("Using Conan Packager to manage installing prebuilt CivetWeb external lib")
```
szaszm, line 20:
```suggestion
        message("Using Conan to install CivetWeb")
```

cmake/GetFmt.cmake
```diff
@@ -0,0 +1,26 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+function(get_fmt)
+    if(MINIFI_FMT_SOURCE STREQUAL "CONAN")
+        message("Using Conan Packager to manage installing prebuilt Fmt external lib")
```
szaszm, line 20:
```suggestion
        message("Using Conan to install Fmt")
```

cmake/GetLibCURL.cmake
```diff
@@ -0,0 +1,27 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+function(get_curl SOURCE_DIR BINARY_DIR)
+    if(MINIFI_LIBCURL_SOURCE STREQUAL "CONAN")
+        message("Using Conan Packager to manage installing prebuilt libcurl external lib")
```
szaszm, line 20:
```suggestion
        message("Using Conan to install libcurl")
```

cmake/GetOpenSSL.cmake
```diff
@@ -0,0 +1,27 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+function(get_openssl SOURCE_DIR BINARY_DIR)
+    if(MINIFI_OPENSSL_SOURCE STREQUAL "CONAN")
+        message("Using Conan Packager to manage installing prebuilt OpenSSL external lib")
```
szaszm, line 20:
```suggestion
        message("Using Conan to install OpenSSL")
```

cmake/GetRocksDB.cmake
```diff
@@ -0,0 +1,34 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+function(get_rocksdb SOURCE_DIR BINARY_DIR)
+    if(MINIFI_ROCKSDB_SOURCE STREQUAL "CONAN")
+        message("Using Conan Packager to manage installing prebuilt RocksDB external lib")
```
szaszm, line 20:
```suggestion
        message("Using Conan to install RocksDB")
```

cmake/GetSpdlog.cmake
```diff
@@ -0,0 +1,31 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+function(get_spdlog)
+    include(GetFmt)
+    get_fmt()
+
+    if(MINIFI_SPDLOG_SOURCE STREQUAL "CONAN")
+        message("Using Conan Packager to manage installing prebuilt Spdlog external lib")
```
szaszm, line 23:
```suggestion
        message("Using Conan to install spdlog")
```

thirdparty/rocksdb/all/patches/dboptions_equality_operator.patch
szaszm, line 1: this patch should be reverted too

thirdparty/rocksdb/config.yml
szaszm, line 1: what is this file used for?

cmake/GetZstd.cmake
james94: Updated this one based on your feedback.

cmake/GetZLIB.cmake
james94: Updated this one based on your feedback.

cmake/GetZLIB.cmake
james94: Updated this one based on your feedback.

etc/conan/profiles/release-linux
james94: Updated this one based on your feedback.

CMakeLists.txt
james94: Thanks. Yes you're right we can delete these commented lines. I will do it in the follow up commit.

CONAN.md
```diff
@@ -0,0 +1,135 @@
...
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+
+# Apache NiFi - MiNiFi - C++ Conan Build Guide
+
+We will walk through the steps to build MiNiFi using conan version 2 that comes with CMake integration. We will also go through the process of creating a MiNiFi conan package for easier integration into other C++ projects. We will start with a discussion on building MiNiFi using the standalone CMake approach and the conan approach. After we have built MiNiFi and created a MiNiFi conan package, we will conclude by elaborating on some of the benefits that can come with integrating a MiNiFi conan package into new or existing C++ infrastructure with respect to real-time robotic systems.
```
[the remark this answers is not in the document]
james94: Thanks for suggesting we add a warning for the user to be aware about Conan support being experimental. I will add it into the follow up PR.

cmake/GetBZip2.cmake
```diff
@@ -0,0 +1,27 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+function(get_bzip2 SOURCE_DIR BINARY_DIR)
+    if(MINIFI_BZIP2_SOURCE STREQUAL "CONAN")
+        message("Using Conan Packager to manage installing prebuilt bzip2 external lib")
```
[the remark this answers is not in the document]
james94: Thanks for the suggestion on making the conan message more concise. I added in the commit "[94fe23e](https://github.com/apache/nifi-minifi-cpp/pull/1793/commits/94fe23e2d01d5e70a708e82bee04ce549a659e2a)"

cmake/GetCatch2.cmake
james94: Thanks for the suggestion on making the conan message more concise. I added in the commit "[94fe23e](https://github.com/apache/nifi-minifi-cpp/pull/1793/commits/94fe23e2d01d5e70a708e82bee04ce549a659e2a)"

cmake/GetCivetWeb.cmake
james94: Thanks for the suggestion on making the conan message more concise. I added in the commit "[94fe23e](https://github.com/apache/nifi-minifi-cpp/pull/1793/commits/94fe23e2d01d5e70a708e82bee04ce549a659e2a)"

cmake/GetFmt.cmake
james94: Thanks for the suggestion on making the conan message more concise. I added in the commit "[94fe23e](https://github.com/apache/nifi-minifi-cpp/pull/1793/commits/94fe23e2d01d5e70a708e82bee04ce549a659e2a)"

cmake/GetLibCURL.cmake
james94: Thanks for the suggestion on making the conan message more concise. I added in the commit "[94fe23e](https://github.com/apache/nifi-minifi-cpp/pull/1793/commits/94fe23e2d01d5e70a708e82bee04ce549a659e2a)"

cmake/GetLibXml2.cmake
```diff
@@ -0,0 +1,27 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+function(get_libxml2 SOURCE_DIR BINARY_DIR)
+    if(MINIFI_LIBXML2_SOURCE STREQUAL "CONAN")
+        message("Using Conan Packager to manage installing prebuilt LibXml2 external lib")
```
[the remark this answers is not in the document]
james94: Thanks for the suggestion on making the conan message more concise. I added in the commit "[94fe23e](https://github.com/apache/nifi-minifi-cpp/pull/1793/commits/94fe23e2d01d5e70a708e82bee04ce549a659e2a)"

cmake/GetOpenSSL.cmake
james94: Thanks for the suggestion on making the conan message more concise. I added in the commit "[94fe23e](https://github.com/apache/nifi-minifi-cpp/pull/1793/commits/94fe23e2d01d5e70a708e82bee04ce549a659e2a)"

cmake/GetRocksDB.cmake
james94: Thanks for the suggestion on making the conan message more concise. I added in the commit "[94fe23e](https://github.com/apache/nifi-minifi-cpp/pull/1793/commits/94fe23e2d01d5e70a708e82bee04ce549a659e2a)"

cmake/GetSpdlog.cmake
james94: Thanks for the suggestion on making the conan message more concise. I added in the commit "[94fe23e](https://github.com/apache/nifi-minifi-cpp/pull/1793/commits/94fe23e2d01d5e70a708e82bee04ce549a659e2a)"

thirdparty/rocksdb/config.yml
james94: Typically "config.yml" is used to define and manage different of a conan package. It lists all available versions of the library that can be built using the conan recipe. For example, **Conan Center Index's** automated build system uses the `config.yml` to:

- 1\. Identify available versions for a conan package
- 2\. Determine which conan recipe to use for each version
- 3\. Apply any necessary patches or configs specific to each version

The `config.yml` also helps for version control, consistency, automation and flexibility.

If we have multiple custom conan recipes in the **thirdparty/** folder, which we may do custom conan recipes over prebuilt packages because we need to apply patches while creating the conan package, we could have our own "local conan repository" similar to the "remote conan center". Thus, we can leverage automated building with config.yml's to create those custom conan recipes.

References:

- Local Recipes Index Repository (Conan Devops Guide): https://docs.conan.io/2/devops/devops_local_recipes_index.html
- Working with Conan repositories - Local Recipes Index Repository (Tutorial): https://docs.conan.io/2/tutorial/conan_repositories/setup_local_recipes_index.html#setup-local-recipes-index

thirdparty/rocksdb/all/patches/arm7.patch
```diff
@@ -4,7 +4,7 @@ index 225e3fa72..cd5f935f1 100644
 +++ b/utilities/transactions/lock/range/range_tree/lib/portability/toku_time.h
 @@ -131,6 +131,10 @@ static inline tokutime_t toku_time_now(void) {
    uint64_t result;
-   __asm __volatile__("mrs %[rt], cntvct_el0" : [ rt ] "=r"(result));
+   __asm __volatile__("mrs %[rt], cntvct_el0" : [rt] "=r"(result));
```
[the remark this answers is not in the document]
james94: The main reason I updated the patch here was because when I tried to create the RocksDB conan package, it complained about "expected" vs "actual" with line 7. Here was the error:

~~~bash
rocksdb/8.10.2@minifi/develop: WARN: /home/ubuntu/.conan2/p/rocksb6dac60094a95/s/patches/arm7.patch: file 1/1:	 b'utilities/transactions/lock/range/range_tree/lib/portability/toku_time.h'
rocksdb/8.10.2@minifi/develop: WARN: /home/ubuntu/.conan2/p/rocksb6dac60094a95/s/patches/arm7.patch:  hunk no.1 doesn't match source file at line 132
rocksdb/8.10.2@minifi/develop: WARN: /home/ubuntu/.conan2/p/rocksb6dac60094a95/s/patches/arm7.patch:   expected: b'  __asm __volatile__("mrs %[rt], cntvct_el0" : [ rt ] "=r"(result));'
rocksdb/8.10.2@minifi/develop: WARN: /home/ubuntu/.conan2/p/rocksb6dac60094a95/s/patches/arm7.patch:   actual  : b'  __asm __volatile__("mrs %[rt], cntvct_el0" : [rt] "=r"(result));'
rocksdb/8.10.2@minifi/develop: /home/ubuntu/.conan2/p/rocksb6dac60094a95/s/patches/arm7.patch: source file is different - b'utilities/transactions/lock/range/range_tree/lib/portability/toku_time.h'
rocksdb/8.10.2@minifi/develop: ERROR: 
Package 'e9ee71a0ada9817d6bde6ac5f5434fa0e06b4166' build failed
rocksdb/8.10.2@minifi/develop: WARN: Build folder /home/ubuntu/.conan2/p/b/rocks7495fe807ea4d/b/build/Release
*********************************************************
Recipe 'rocksdb/8.10.2@minifi/develop' cannot build its binary
It is possible that this recipe is not Conan 2.0 ready
If the recipe comes from ConanCenter, report it at https://github.com/conan-io/conan-center-index/issues
If it is your recipe, check if it is updated to 2.0
*********************************************************

ERROR: rocksdb/8.10.2@minifi/develop: Error in build() method, line 177
	apply_conandata_patches(self)
	ConanException: Failed to apply patch: /home/ubuntu/.conan2/p/rocksb6dac60094a95/s/patches/arm7.patch
~~~

By making that slight adjustment to the patch, when I created the custom RocksDB conan package, it was able to apply this patch successfully while creating that RocksDB conan package.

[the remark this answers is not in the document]
james94: I also do want to point out that when we build MiNiFi from source using the standalone CMake approach and we build that RocksDB applying its patches with this arm7.patch slightly updated, the BundledRocksDB external lib is built successfully too.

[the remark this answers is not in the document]
james94: Heres the result from creating the custom RocksDB conan package after applying this patch and the other relevant patches:

~~~bash
~/src/open_source/nifi-minifi-cpp/thirdparty/rocksdb/all (MINIFICPP-2346-P1)$ conan create . --user=minifi --channel=develop --version=8.10.2 -pr=../../../etc/conan/profiles/release-linux

# Output (Showing RocksDB conan package created successfully)
rocksdb/8.10.2@minifi/develop: package(): Packaged 104 '.h' files
rocksdb/8.10.2@minifi/develop: package(): Packaged 1 '.a' file: librocksdb.a
rocksdb/8.10.2@minifi/develop: package(): Packaged 1 file: COPYING
rocksdb/8.10.2@minifi/develop: package(): Packaged 1 '.leveldb' file: LICENSE.leveldb
rocksdb/8.10.2@minifi/develop: package(): Packaged 1 '.Apache' file: LICENSE.Apache
rocksdb/8.10.2@minifi/develop: Created package revision 7b0eeda125026e144c7d154e3bffbaa9
rocksdb/8.10.2@minifi/develop: Package 'e9ee71a0ada9817d6bde6ac5f5434fa0e06b4166' created
rocksdb/8.10.2@minifi/develop: Full package reference: rocksdb/8.10.2@minifi/develop#d31518436210c3f016be820242c23550:e9ee71a0ada9817d6bde6ac5f5434fa0e06b4166#7b0eeda125026e144c7d154e3bffbaa9
rocksdb/8.10.2@minifi/develop: Package folder /home/ubuntu/.conan2/p/b/rocksaebf4a8af630b/p
WARN: deprecated: Usage of deprecated Conan 1.X features that will be removed in Conan 2.X:
WARN: deprecated:     'cpp_info.names' used in: bzip2/1.0.8, zlib/1.3.1, lz4/1.9.4, rocksdb/8.10.2@minifi/develop, zstd/1.5.2
WARN: deprecated:     'cpp_info.build_modules' used in: lz4/1.9.4, bzip2/1.0.8
WARN: deprecated:     'env_info' used in: zstd/1.5.2, bzip2/1.0.8
~~~

CMakeLists.txt
james94: Heres the commit where I removed these comments: [816a882](https://github.com/apache/nifi-minifi-cpp/pull/1793/commits/816a882660102e6540eff98f4aaf203cd7ef217f)

CONAN.md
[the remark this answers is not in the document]
james94: Heres the commit where I added the warning message in CONAN.md: [816a882](https://github.com/apache/nifi-minifi-cpp/pull/1793/commits/816a882660102e6540eff98f4aaf203cd7ef217f)

thirdparty/rocksdb/all/patches/arm7.patch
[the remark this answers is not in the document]
james94: @szaszm I just learned that Conan supports fuzzy patching, so even if there are minor differences between patch code and actual source code, we can still apply the patch. Therefore, we can keep the original **arm7.patch** based on MiNiFi **main** branch and apply it to our rocksdb conan package. I reverted the **arm7.patch** back to how it is in **main** branch. I was able to create a custom **rocksdb** conan package and not run into the error I mentioned earlier.

I will double check that I can build MiNiFi using conan build and then conan create. I will also check I can build MiNiFi using standalone CMake. I will run the CTESTs too. Then I will push the latest code updates.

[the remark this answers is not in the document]
james94: @szaszm in my latest commit "[374fff2](https://github.com/apache/nifi-minifi-cpp/pull/1793/commits/374fff2415d5a1099efb95a0ec198d35472259ea)", I reverted **arm7.patch** back to how it was in **main** branch.

thirdparty/rocksdb/all/conanfile.py
```diff
@@ -0,0 +1,243 @@
...
+        if self.options.with_jemalloc:
+            deps.set_property("jemalloc", "cmake_file_name", "JeMalloc")
+            deps.set_property("jemalloc", "cmake_target_name", "JeMalloc::JeMalloc")
+        if self.options.with_zstd:
+            deps.set_property("zstd", "cmake_target_name", "zstd::zstd")
+        deps.generate()
+
+    def build(self):
+        for patch_data in self.conan_data.get("patches", {}).get(self.version, []):
+            patch(self, patch_file=patch_data["patch_file"], base_path=self.source_folder, strip=1, fuzz=True)
```
james94, line 177: Here is where we use **fuzzy patching** when applying the patches for creating the custom **RocksDB** conan package. Previously, we did **apply_conandata_patches()**, but that was too strict on the patch code vs actual source code having minor differences and caused creating the **RocksDB** conan package to fail. With **fuzzy patching**, we were able to successfully create the **RocksDB** conan package even with the patch code vs actual source code having minor differences. @szaszm

szaszm (approved):
thanks for your hard work and patience

lordgamez (approved):

fgerlits (approved):

james94:
Thanks for the support everyone on this Conan integration effort with MiNiFi.

PR #1805, opened 2024-05-29
MINIFICPP-2388 Fix warnings on Windows build

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2388

Depends on https://github.com/apache/nifi-minifi-cpp/pull/1802

---------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


.gitignore
```diff
@@ -65,6 +65,12 @@ __pycache__/
 /provenance_repository
 /logs
 msi/WixWin.wsi
+extensions/expression-language/el-scanner/Scanner.cpp
+extensions/expression-language/el-parser/Parser.cpp
+extensions/expression-language/el-parser/Parser.hpp
+extensions/expression-language/el-parser/location.hh
+extensions/expression-language/el-parser/position.hh
+extensions/expression-language/el-parser/stack.hh
```
martinzink, line 73: Why did you extract the local gitignore to this root one?
I prefered the previous approach, because it was more modular.

libminifi/include/io/CRCStream.h
```diff
@@ -27,6 +27,7 @@
 #include <cassert>
 
 #ifdef WIN32
+#pragma warning(disable : 4250)
```
szaszm, line 30: Could you please add a comment about what this does or prevents?

extensions/sftp/processors/PutSFTP.cpp
szaszm, line 1: I'd rather avoid the extra allocations from the `path::string` calls and keep `c_str`.

extensions/expression-language/CMakeLists.txt
```diff
@@ -96,20 +96,22 @@ find_package(FLEX REQUIRED)
...
+    ${CMAKE_CURRENT_SOURCE_DIR}/el-scanner/Scanner.cpp
     COMPILE_FLAGS --c++
 )
 
 add_flex_bison_dependency(el-scanner el-parser)
 
 include_directories(./ ../../libminifi/include  ../../libminifi/include/core)
 include_directories(SYSTEM ../../thirdparty/)
+include_directories(SYSTEM el-parser)
+include_directories(SYSTEM el-scanner)
```
szaszm, line 114: What warnings are fixed here? I'd rather not add any more directory-based include paths unless absolutely necessary.

extensions/opc/src/fetchopc.cpp
```diff
@@ -62,7 +62,7 @@ namespace org::apache::nifi::minifi::processors {
 
     if (idType_ == opc::OPCNodeIDType::Int) {
       try {
-        std::stoi(nodeID_);
+        static_cast<void>(std::stoi(nodeID_));
```
szaszm, line 65:
```suggestion
        // ensure that nodeID_ can be parsed as an int
        static_cast<void>(std::stoi(nodeID_));
```

extensions/opc/src/putopc.cpp
```diff
@@ -56,7 +56,7 @@ namespace org::apache::nifi::minifi::processors {
 
     if (idType_ == opc::OPCNodeIDType::Int) {
       try {
-        std::stoi(nodeID_);
+        static_cast<void>(std::stoi(nodeID_));
```
szaszm, line 59:
```suggestion
        // ensure that nodeID_ can be parsed as an int
        static_cast<void>(std::stoi(nodeID_));
```

extensions/sftp/processors/PutSFTP.cpp
lordgamez: On Windows the previous version didn't compile as std::filesystem::path cannot be converted implicitly to string, that is why we had to use the `.string()`. Even if it is changed to `.string().c_str()` wouldn't it still result in extra allocations?

extensions/expression-language/CMakeLists.txt
lordgamez: Previously we generated the parser and scanner header and source files into the `extensions/expression-language/` directory and the files in that directory were compiled with the flags handling warnings as errors. Because of this when some warnings were found in the generated header files included by our own source files that resulted in a compile error. Moving the generated files to separate include directories and marking them as system includes ignores the warnings in those header files.

.gitignore
lordgamez: In the first draft el-scanner and el-parser directories were not part of the repository as they were only created at build time with the generated files. As CPPLINT.cfg is now also added in those directories I think you are right, I'll add the .gitignore files there as well.

extensions/expression-language/CMakeLists.txt
szaszm: Ok, got it. But I still thing that we should not add directory-based include paths, and add them to each target as necessary. It would also be nicer to generate these files under the build directory, instead of the source tree.

.gitignore
lordgamez: Updated in be9cace2da1aa161129ac9c22986033dafc199db

libminifi/include/io/CRCStream.h
lordgamez: Updated in be9cace2da1aa161129ac9c22986033dafc199db

extensions/expression-language/CMakeLists.txt
lordgamez: Updated in be9cace2da1aa161129ac9c22986033dafc199db

extensions/opc/src/fetchopc.cpp
lordgamez: Updated in be9cace2da1aa161129ac9c22986033dafc199db

extensions/opc/src/putopc.cpp
lordgamez: Updated in be9cace2da1aa161129ac9c22986033dafc199db

extensions/sftp/processors/PutSFTP.cpp
lordgamez: It seems to me according to https://en.cppreference.com/w/cpp/filesystem/path/string that the required conversions are automatically performed as needed and the example shows non-ASCII filenames as well.

fgerlits (approved):

szaszm (approved):

martinzink (approved):

adamdebreceni (approved):

PR #1810, opened 2024-06-06
MINIFICPP-2394 fix LogAttribute AttributesToLog and AttributesToIgnor…

martinzink:
…e properties

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/standard-processors/processors/LogAttribute.cpp
```diff
@@ -48,98 +49,93 @@ void LogAttribute::onSchedule(core::ProcessContext& context, core::ProcessSessio
...
-    std::string value;
-    if (context.getProperty(LogLevel, value)) {
-      logLevelStringToEnum(value, level);
-    }
-    if (context.getProperty(LogPrefix, value)) {
-      dashLine = "-----" + value + "-----";
+  if (auto log_level_str = context.getProperty(LogLevel)) {
+    if (auto result = magic_enum::enum_cast<core::logging::LOG_LEVEL>(*log_level_str)) {
+      log_level_ = *result;
+    } else if (*log_level_str == "error") {  // TODO(MINIFICPP-2294) this could be avoided if config files were properly migrated
```
martinzink, line 66: "error" string is not parsed into core::logging:LOG_LEVEL::err implicitly

extensions/standard-processors/processors/LogAttribute.h
```diff
@@ -33,24 +33,23 @@
...
   }
   ~LogAttribute() override = default;
 
   EXTENSIONAPI static constexpr const char* Description = "Logs attributes of flow files in the MiNiFi application log.";
 
-  EXTENSIONAPI static constexpr auto LogLevel = core::PropertyDefinitionBuilder<5>::createProperty("Log Level")
+  EXTENSIONAPI static constexpr auto LogLevel = core::PropertyDefinitionBuilder<6>::createProperty("Log Level")
       .withDescription("The Log Level to use when logging the Attributes")
-      .withAllowedValues({"info", "trace", "error", "warn", "debug"})
+      .withAllowedValues({"trace", "debug", "info", "warn", "error", "critical"})
```
szaszm, line 52: Do we really need critical here? I'm not even sure about its usefulness elsewhere, but a LogAttribute message will certainly not be a critical error, which I consider those that kill or crash the minifi agent.

Do you prefer a consistent set of log levels, even if one of the options don't make sense to be user-exposed?

extensions/standard-processors/processors/LogAttribute.cpp
szaszm: Can we change the log level name to "error" instead, without breaking stuff? full words > half word here IMO

extensions/standard-processors/processors/LogAttribute.h
martinzink: I prefer consistency, it doesnt cost us anything. I can image a scenario where these logs are critical not from minifis point of view but rather from the flow's perspective. (e.g. you log out critically a failure relationship because its business critical to have the previous operation to be successfull)

extensions/standard-processors/processors/LogAttribute.cpp
martinzink: Good question but if i recall correctly its already used elsewhere as is (from user input), so the breaking change would apply there as well.
We could revisit this idea in a new PR, but I think you are right for these changes we would need the flow migration feature.

lordgamez (approved):

extensions/standard-processors/tests/unit/LogAttributeTests.cpp
```diff
@@ -0,0 +1,145 @@
...
+  const auto log_attribute = std::make_shared<LogAttribute>("log_attribute");
+  SingleProcessorTestController controller{log_attribute};
+  LogTestController::getInstance().setTrace<LogAttribute>();
+
+  const auto [log_level, log_prefix, expected_dash] = GENERATE(
+    std::make_tuple("info", "", "-------------------------------------------------"),
+    std::make_tuple("critical", "foo", "-----------------------foo------------------------"),
+    std::make_tuple("debug", "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Morbi mollis neque sit amet dui pretium sodales.",
+        "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Morbi mollis neque sit amet dui pretium sodales."),
+    std::make_tuple("error", "", "-------------------------------------------------"));
```
adamdebreceni, line 66: the expected default dashes in this file are 49 characters long

martinzink: good catch :+1: fixed it in https://github.com/apache/nifi-minifi-cpp/pull/1810/commits/a2dd56c5681a182325eeb18f1199ce99063fe633

szaszm (approved):

PR #1812, opened 2024-06-07
MINIFICPP-2399 Evaluate expr language in GetFile's Directory property

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2399

--------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/standard-processors/tests/unit/GetFileTests.cpp
```diff
@@ -292,3 +292,26 @@ TEST_CASE("GetFile sets attributes correctly") {
...
+
+TEST_CASE("GetFile can use expression language in Directory property") {
+  using minifi::processors::GetFile;
+  LogTestController::getInstance().setTrace<GetFile>();
+
+  const auto get_file = std::make_shared<GetFile>("GetFile");
+  minifi::test::SingleProcessorTestController test_controller(get_file);
+
+  std::filesystem::path base_dir = test_controller.createTempDirectory();
+  auto date_str = date::format("%Y-%m-%d", std::chrono::time_point_cast<std::chrono::seconds>(std::chrono::system_clock::now()));
```
fgerlits, line 304: nitpicking, but why do we cast (from microseconds?) to seconds if we only use the date part?

martinzink (approved):

extensions/standard-processors/tests/unit/GetFileTests.cpp
lordgamez: Good point, it's not needed here, updated in a9ebc13605cf3ac4ba35fd6816ee224f3395a9a9. The reason for this was that I started working with getDateTimeStr which takes `std::chrono::sys_seconds`, but after removing it's not needed anymore.

szaszm (approved):

fgerlits (approved):

extensions/standard-processors/tests/unit/GetFileTests.cpp
```diff
@@ -292,3 +292,26 @@ TEST_CASE("GetFile sets attributes correctly") {
     CHECK(flow_file->getAttribute(minifi::core::SpecialFlowAttribute::FILENAME) == "beta.txt");
   }
 }
+
+TEST_CASE("GetFile can use expression language in Directory property") {
```
martinzink, line 296: Based on the latest windows test failure you need something like here, and you would also need to define TZ_DATA_DIR in extensions/standard-processors/tests/CMakeLists.txt for unit tests aswell (currently its only defined for integration tests)
```suggestion
TEST_CASE("GetFile can use expression language in Directory property") {
#ifdef WIN32
  expression::dateSetInstall(TZ_DATA_DIR);
#endif
```

martinzink (changes requested):

extensions/standard-processors/tests/unit/GetFileTests.cpp
lordgamez: Thanks! Updated in 70c7209706b5acda7b95051f57fad8168bdd7712

martinzink (approved):

adamdebreceni (approved):

PR #1815, opened 2024-06-10
MINIFICPP-2398 Only commit once per time_slice in EventDrivenSchedulingAgent

martinzink:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/src/EventDrivenSchedulingAgent.cpp
+++ b/libminifi/src/EventDrivenSchedulingAgent.cpp
@@ -38,38 +38,48 @@ void EventDrivenSchedulingAgent::schedule(core::Processor* processor) {
 utils::TaskRescheduleInfo EventDrivenSchedulingAgent::run(core::Processor* processor,
     const std::shared_ptr<core::ProcessContext>& process_context,
     const std::shared_ptr<core::ProcessSessionFactory>& session_factory) {
-  if (this->running_) {
-    const auto start_time = std::chrono::steady_clock::now();
-    // trigger processor until it has work to do, but no more than the configured nifi.flow.engine.event.driven.time.slice
-
-    const auto process_session = session_factory->createSession();
-    process_session->setMetrics(processor->getMetrics());
-
-    try {
-      while (processor->isRunning() && (std::chrono::steady_clock::now() - start_time < time_slice_)) {
-        this->trigger(processor, process_context, process_session);
-        if (processor->isYield()) {
-          process_session->commit();
-          return utils::TaskRescheduleInfo::RetryAfter(processor->getYieldExpirationTime());
-        }
-      }
-      process_session->commit();
-    } catch (const std::exception& exception) {
-      logger_->log_warn("Caught \"{}\" ({}) during Processor::onTrigger of processor: {} ({})",
-          exception.what(), typeid(exception).name(), processor->getUUIDStr(), processor->getName());
-      processor->yield(admin_yield_duration_);
-      process_session->rollback();
-      throw;
-    } catch (...) {
-      logger_->log_warn("Caught unknown exception during Processor::onTrigger of processor: {} ({})", processor->getUUIDStr(), processor->getName());
-      processor->yield(admin_yield_duration_);
-      process_session->rollback();
-      throw;
-    }
-
-    return utils::TaskRescheduleInfo::RetryImmediately();  // Let's continue work as soon as a thread is available
+  if (!this->running_) {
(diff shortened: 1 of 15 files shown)


martinzink:
Thanks @lordgamez for noticing some issues with the PR, I converted into draft while I iron out those things (it still got some problems as the failing docker tests seems relevant to the changes)

lordgamez (approved):

martinzink:
Since we are in the discussion about deprecating the whole JNI extension, I think we safely ignore the clang tidy issues coming from that

libminifi/src/core/Processor.cpp
```diff
@@ -167,44 +167,43 @@ bool Processor::addConnection(Connectable* conn) {
...
+  const auto process_session = session_factory->createSession();
+  process_session->setMetrics(metrics_);
   try {
-    // Call the virtual trigger function
-    auto start = std::chrono::steady_clock::now();
-    onTriggerSharedPtr(context, session);
-    metrics_->addLastOnTriggerRuntime(std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - start));
-    start = std::chrono::steady_clock::now();
-    session->commit();
-    metrics_->addLastSessionCommitRuntime(std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - start));
```
szaszm, line 195: The session runtime metric seems to be missing after the changes

libminifi/src/EventDrivenSchedulingAgent.cpp
```diff
@@ -35,20 +35,51 @@ void EventDrivenSchedulingAgent::schedule(core::Processor* processor) {
...
+  }
+
+  const auto start_time = std::chrono::steady_clock::now();
+  // trigger processor until it has work to do, but no more than the configured nifi.flow.engine.event.driven.time.slice
+
+  const auto process_session = session_factory->createSession();
+  process_session->setMetrics(processor->getMetrics());
+
+
+  try {
```
adamdebreceni, line 55: as I understand SchedulingAgent::trigger catches any exception from Processor::onTrigger and wraps it in an expected, but we do not rethrow that, so the error-case for this try-catch is when process_session->commit() throws, previously SchedulingAgent::trigger wrapped both the Processor::onTrigger and commit/rollback exceptions and EventDrivenSchedulingAgent simply discarded it

gsl::finally should happen in case of an exception, but in this case I don't think we want to call commit when we get an exception here, so I would just move process_session->commit() after the while-loop, also we should make sure to log any exceptions from both commit and rollback

libminifi/src/core/Processor.cpp
martinzink: I've moved that logic into the (imho more fitting place) ProcessSession::commit
https://github.com/apache/nifi-minifi-cpp/pull/1815/files#diff-d55088efe0f1ad2a600dba414bfee749c5fb634ab44418146623f892bd6756a4R807-R922

libminifi/src/EventDrivenSchedulingAgent.cpp
martinzink: good point, how about this?
https://github.com/apache/nifi-minifi-cpp/pull/1815/commits/c6a9d06109f0af494b21982b7df51cfd97af5ee3

adamdebreceni: I don't know if EventDrivenSchedulingAgent::run can throw an exception and the threadpool can handle it, but it previously did not throw

adamdebreceni (approved):

libminifi/src/EventDrivenSchedulingAgent.cpp
```diff
@@ -35,20 +35,64 @@ void EventDrivenSchedulingAgent::schedule(core::Processor* processor) {
...
+    }
   }
-  return utils::TaskRescheduleInfo::Done();
+  if (needs_commit) {
+    try {
+      process_session->commit();
+    } catch (const std::exception& exception) {
+      logger_->log_warn("Caught \"{}\" ({}) during ProcessSession::commit after triggering processor: {} ({})",
+      exception.what(), typeid(exception).name(), processor->getUUIDStr(), processor->getName());
+      process_session->rollbackNoThrow();
```
szaszm, line 82: We shouldn't just throw away the exception here, at least logging something would be nice. If a rollback fails, it's probably a critical error, so `terminate` might be an appropriate handling.

szaszm (approved):

PR #1817, opened 2024-06-14
MINIFICPP-2343 Fix quoting when CMake calls generateVersion.bat

fgerlits:
If the source or build directories contained whitespace, CMake could not generate the `agent_version.cpp` file because the call to `generateVersion.bat` failed.

The two path parameters simply had to be quoted, but CMake calls the COMMAND directly without stripping `""`s, so I had to use PowerShell to convert the path to a short path (which looks like `C:\Users\JaneDo~1\src\minifi`).

I have also added more verbose logging to make it easier to diagnose issues which may arise in the future.

https://issues.apache.org/jira/browse/MINIFICPP-2343

---

### For all changes:
     in the commit message?

### For code changes:

### Note:


szaszm:
I'd prefer not relying on short paths that are just the legacy of the DOS era, and may disappear over time. Would it work if we added an extra set of quotes around the generateVersion.bat script path?

fgerlits:
> I'd prefer not relying on short paths that are just the legacy of the DOS era, and may disappear over time. Would it work if we added an extra set of quotes around the generateVersion.bat script path?

I have tried a zillion ways to quote the command.  In every case, either CMake stripped the quotes and gave me a `'C:/Users/Ferenc' is not recognized as an internal or external command, operable program or batch file` error, or it kept the quotes and then went and tried to find a command with a quote in its name, which of course failed.

If you have a specific suggestion for how to quote the command, I'll try it (again, probably), but most of the advice I have found online was "don't quote the command, quote everything else".  I'm not a fan of DOS file names, either, but this was the only thing that worked.

lordgamez (approved):

szaszm (approved):

PR #1822, opened 2024-06-18
MINIFICPP-2405 Fix test failure due to too low memory usage

fgerlits:
`MemoryUsageTest` has started failing on MacOS CI (AppleClang 15.0.0.15000040) because a vector containing 30,000,000 zero chars occupies less than 30,000,000 bytes in memory. Filling the vector with random numbers seems to solve the "issue".

https://issues.apache.org/jira/browse/MINIFICPP-2405

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/test/unit/MemoryUsageTest.cpp
+++ b/libminifi/test/unit/MemoryUsageTest.cpp
@@ -36,7 +36,9 @@ TEST_CASE("Test Physical memory usage", "[testphysicalmemoryusage]") {
   std::mt19937 gen(std::random_device{}());
   std::uniform_int_distribution dist(0, 255);
   std::vector<uint8_t> large_random_vector(30'000'000);
-  std::generate(begin(large_random_vector), end(large_random_vector), [&]() { return dist(gen); });
+  for (size_t i = 0; i < 10; ++i) {
+    large_random_vector[dist(gen) * 100'000 + dist(gen) * 1000 + dist(gen)] = dist(gen);
+  }
 
   const auto ram_usage_by_process = utils::OsUtils::getCurrentProcessPhysicalMemoryUsage();
   const auto ram_usage_by_system = utils::OsUtils::getSystemPhysicalMemoryUsage();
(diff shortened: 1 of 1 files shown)


martinzink (approved):

szaszm (approved):

libminifi/test/unit/MemoryUsageTest.cpp
```diff
@@ -22,28 +22,28 @@
...
 #include "utils/OsUtils.h"
 #include "unit/TestBase.h"
 #include "unit/Catch.h"
 
 TEST_CASE("Test Physical memory usage", "[testphysicalmemoryusage]") {
   constexpr bool cout_enabled = true;
-  std::vector<char> v(30000000);
-  // use v to prevent recent compilers from optimizing out the allocation
-  strcpy(v.data(), " bytes\n");  // NOLINT: "almost always, snprintf is better than strcpy", but formatting is unnecessary here
+  const auto large_random_vector = minifi::utils::crypto::randomBytes(30'000'000);
```
szaszm, line 33: There's no need for crypto random here, we could use normal PRNG, so it works better on systems without a fast crypto random generator.

And it would be nice to add a comment about why it's random, like in the PR description.

szaszm (changes requested):
:arrow_up:

libminifi/test/unit/MemoryUsageTest.cpp
fgerlits: I have changed it to use `<random>` in fa35ebc33e8198fa11450183c9f54705c7b2073f.  I don't know if it works "better" (it is about 3 times slower on my computer, 0.3s instead of 0.1s), but I hope we'll have a few years until a compiler can optimize this out, too.

szaszm: Is the extra time spent in `std::accumulate`? We might get away with only accessing a couple of elements to make it faster.
Also, with this many elements, we're likely running into signed integer overflow, which is undefined behavior.

fgerlits: `accumulate` is fast; calling `dist(gen)` is slow.  I have reduced the number of random numbers used from 30,000,000 to 40 in 4dddfd9b7e6563f0fe9c9f3ae2e2d5a073dd4161; now it runs in 0.02 seconds.

adamdebreceni (approved):

szaszm (approved):

lordgamez (approved):

PR #1826, opened 2024-06-21
MINFICPP-2243 - In ListenHTTP process incoming request only in onTrigger

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/civetweb/processors/ListenHTTP.cpp
+++ b/extensions/civetweb/processors/ListenHTTP.cpp
@@ -219,7 +219,7 @@ bool ListenHTTP::processIncomingFlowFile(core::ProcessSession &session) {
   return true;
 }
 
-bool ListenHTTP::processFlowFile(std::shared_ptr<core::FlowFile> flow_file) {
+bool ListenHTTP::processFlowFile(const std::shared_ptr<core::FlowFile>& flow_file) {
   ResponseBody response;
   flow_file->getAttribute("filename", response.uri);
   flow_file->getAttribute("mime.type", response.mime_type);
--- a/extensions/civetweb/processors/ListenHTTP.cpp
+++ b/extensions/civetweb/processors/ListenHTTP.cpp
@@ -31,8 +31,6 @@
 
 namespace org::apache::nifi::minifi::processors {
 
-const core::Relationship ListenHTTP::Self("__self__", "Marks the FlowFile to be owned by this processor");
-
 void ListenHTTP::initialize() {
   logger_->log_trace("Initializing ListenHTTP");
 
@@ -539,7 +537,7 @@ void ListenHTTP::notifyStop() {
 
 std::set<core::Connectable*> ListenHTTP::getOutGoingConnections(const std::string &relationship) {
   auto result = core::Processor::getOutGoingConnections(relationship);
-  if (relationship == Self.getName()) {
+  if (relationship == Self.name) {
     result.insert(this);
   }
   return result;
--- a/extensions/civetweb/processors/ListenHTTP.cpp
+++ b/extensions/civetweb/processors/ListenHTTP.cpp
@@ -277,7 +277,7 @@ class MgConnectionInputStream : public io::InputStream {
 
 class MgConnectionOutputStream : public io::OutputStream {
  public:
-  MgConnectionOutputStream(struct mg_connection* conn): conn_(conn) {}
+  explicit MgConnectionOutputStream(struct mg_connection* conn): conn_(conn) {}
 
   size_t write(const uint8_t *value, size_t len) override {
     const auto mg_write_return = mg_write(conn_, reinterpret_cast<const void*>(value), len);
--- a/extensions/civetweb/processors/ListenHTTP.cpp
+++ b/extensions/civetweb/processors/ListenHTTP.cpp
@@ -31,6 +31,8 @@
 
 namespace org::apache::nifi::minifi::processors {
 
+const core::Relationship ListenHTTP::Self("__self__", "Marks the FlowFile to be owned by this processor");
+
 void ListenHTTP::initialize() {
   logger_->log_trace("Initializing ListenHTTP");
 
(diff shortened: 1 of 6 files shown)


martinzink (approved):

extensions/civetweb/processors/ListenHTTP.h
```diff
@@ -141,18 +149,33 @@ class ListenHTTP : public core::Processor {
...
   };
 
   // HTTP request handler
   class Handler : public CivetHandler {
    public:
+    enum class FailureReason {
+      PROCESSOR_SHUTDOWN
+    };
+    using RequestValue = std::pair<std::reference_wrapper<core::ProcessSession>, std::promise<void>>;
+    using FailureValue = std::pair<FailureReason, std::promise<void>>;
```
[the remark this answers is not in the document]
adamdebreceni: changed

extensions/civetweb/processors/ListenHTTP.h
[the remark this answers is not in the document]
adamdebreceni: removed

extensions/civetweb/processors/ListenHTTP.cpp
```diff
@@ -191,52 +210,95 @@ bool ListenHTTP::processIncomingFlowFile(core::ProcessSession &session) {
...
+
+class MgConnectionInputStream : public io::InputStream {
+ public:
+  MgConnectionInputStream(struct mg_connection* conn, std::optional<size_t> size): conn_(conn), size_(size) {}
+
+  size_t read(std::span<std::byte> out_buffer) override {
+    const auto mg_read_return = mg_read(conn_, out_buffer.data(), std::min(out_buffer.size(), size_.value_or(std::numeric_limits<size_t>::max()) - offset_));
+    if (mg_read_return <= 0) {
+      return 0;
+    }
```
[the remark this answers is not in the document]
adamdebreceni: that makes sense, changed

extensions/civetweb/processors/ListenHTTP.h
```diff
@@ -39,15 +39,21 @@
...
+struct ListenHTTPTestAccessor;
+}  // namespace org::apache::nifi::minifi::test
 
 namespace org::apache::nifi::minifi::processors {
 
 class ListenHTTP : public core::Processor {
  private:
-  static constexpr std::string_view DEFAULT_BUFFER_SIZE_STR = "20000";
+  static constexpr std::string_view DEFAULT_BUFFER_SIZE_STR = "5";
+  static const core::Relationship Self;
```
szaszm, line 53:
```suggestion
  static constexpr core::RelationshipDefinition Self{"__self__", "Marks the FlowFile to be owned by this processor"};
```

szaszm: I committed this in [4ab6514](https://github.com/apache/nifi-minifi-cpp/pull/1826/commits/4ab6514baf370199eb515c3816218d46b002261f). Feel free to remove it if you don't like the change.

extensions/civetweb/processors/ListenHTTP.h
```diff
@@ -141,18 +147,39 @@ class ListenHTTP : public core::Processor {
...
+    };
+    struct RequestValue {
+      std::reference_wrapper<core::ProcessSession> session;
+      std::promise<void> ret;
+    };
+    struct FailureValue {
+      FailureReason reason;
+      std::promise<void> ret;
+    };
+    using Request = std::promise<nonstd::expected<RequestValue, FailureValue>>;
```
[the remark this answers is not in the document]
adamdebreceni: the processor should release all references to the ProcessSession before returning from onTrigger, and since writing the data into the flowfile happens on the Handler's thread, it is not enough for the Handler to wait for the onTrigger's resources (promise1) onTrigger should also wait for the Handler's thread to release these resources (promise2)

szaszm:
There are compilation errors in the CI

extensions/civetweb/processors/ListenHTTP.cpp
```diff
@@ -191,52 +208,97 @@ bool ListenHTTP::processIncomingFlowFile(core::ProcessSession &session) {
...
+    netstream_offset_ += gsl::narrow<size_t>(mg_read_return);
+    return gsl::narrow<size_t>(mg_read_return);
+  }
+
+ private:
+  struct mg_connection* conn_;
+  size_t netstream_offset_{0};  // how much has been read from conn_
+  std::optional<size_t> netstream_size_limit_;  // how much can we read from conn_
```
lordgamez, line 275: Should this be optional? It seems that we always have a specific message size when using this input stream for message processing.

extensions/civetweb/processors/ListenHTTP.h
```diff
@@ -165,28 +192,66 @@ class ListenHTTP : public core::Processor {
...
     static void sendHttp503(struct mg_connection *conn);
     bool authRequest(mg_connection *conn, const mg_request_info *req_info) const;
     void setHeaderAttributes(const mg_request_info *req_info, core::FlowFile& flow_file) const;
-    void writeBody(mg_connection *conn, const mg_request_info *req_info, bool include_payload = true);
-    static std::unique_ptr<io::BufferStream> createContentBuffer(struct mg_connection *conn, const struct mg_request_info *req_info);
-    void enqueueRequest(mg_connection *conn, const mg_request_info *req_info, std::unique_ptr<io::BufferStream>);
+    void writeBody(core::ProcessSession* payload_reader, mg_connection *conn, const mg_request_info *req_info);
+    void enqueueRequest(mg_connection *conn, const mg_request_info *req_info, bool write_body);
+
+    class RequestBuffer : public utils::ConcurrentQueue<Request> {
```
lordgamez, line 226: I think we could avoid the inheritance and use a wrapper around the concurrent queue instead. This way we don't need to modify the ConcurrentQueue and make it's enqueue method virtual, just have the wrapper as a proxy for the few methods that are used by the Handler class. We can also replace the `running_` with and atomic bool variable so no additional locks are needed in the wrapper.

extensions/civetweb/processors/ListenHTTP.cpp
szaszm: It can be missing if the incoming request doesn't have a specified length, according to the civetweb headers at the call site.

But civetweb gives us a `-1` in that case, and it's not checked before passing it into the contructor, on line 364, so that looks like a bug @adamdebreceni

extensions/civetweb/processors/ListenHTTP.h
adamdebreceni: removed the inheritance, unfortunately we have to employ a lock, as it can happen that we check `running_` on new incoming connection, which is true, then we call `notifyStop`, empty the queue, and only then enqueue the request, leaving it hanging, the ConditionalConcurrentQueue variant sadly does not solve this problem

extensions/civetweb/processors/ListenHTTP.cpp
lordgamez: Looks good to me now, thanks for fixing it!

extensions/civetweb/processors/ListenHTTP.h
lordgamez: Thanks!

lordgamez (approved):

PR #1832, opened 2024-06-26
MINIFICPP-2420 Fix GCC-13 warnings

fgerlits:
Fix maybe-uninitialized warnings in JoltUtils.cpp and Expression.cpp. In the latter, these were caused by an uninitialized variable in the date library (inlined here), which is fixed in the latest, unreleased, version.

Also:
- add missing `Py_INCREF(Py_None)` calls in PythonScriptEngine.cpp, which caused a none_dealloc exception at exit;
- use Ubuntu 24.04 CI runner instead of 20.04 (24.04 uses gcc 13).

https://issues.apache.org/jira/browse/MINIFICPP-2420

---

### For all changes:

### For code changes:

### Note:


martinzink (commented):
Now with gcc13 in the mainline ci we could remove the seldom used .github/workflows/gcc13-compat.yml

extensions/standard-processors/utils/JoltUtils.cpp
szaszm, line 1: I generally prefer `push_back` with a potential move over `emplace_back`, because type checking and errors work better if we use the overload that takes a concrete type.

cmake/Date.cmake
```diff
@@ -45,15 +45,9 @@ if (WIN32)
...
-set(PC ${Bash_EXECUTABLE}  -c "set -x &&\
-        (\\\"${Patch_EXECUTABLE}\\\" -p1 -R -s -f --dry-run -i \\\"${PATCH_FILE}\\\" || \\\"${Patch_EXECUTABLE}\\\" -p1 -N -i \\\"${PATCH_FILE}\\\")")
-
-
 FetchContent_Declare(date_src
-    URL             https://github.com/HowardHinnant/date/archive/refs/tags/v3.0.1.tar.gz
-    URL_HASH        SHA256=7a390f200f0ccd207e8cff6757e04817c1a0aec3e327b006b7eb451c57ee3538
-    PATCH_COMMAND "${PC}"
+    GIT_REPOSITORY  https://github.com/HowardHinnant/date.git
+    GIT_TAG         1ead6715dec030d340a316c927c877a3c4e5a00c  # master as of 2024-07-25
```
szaszm, line 50: I'd prefer to stay on the release tarball version, because fetching tarballs is faster during the build, and github is less likely to rate-limit our builds. Is this feasible?

fgerlits: sure; changed to fetching a tarball in 8352c1bc7091a218fb1eaab413ea1a06bc70e35d

extensions/standard-processors/utils/JoltUtils.cpp
fgerlits: I thought we preferred `emplace_back` over `push_back` in general, and the fact that the compiler can understand it better was a bonus.

In 32103fe66ac607b92e787591ff236bb861b600c8, I have changed it back to `push_back`s and only changed the one line where the compiler got confused by all the vectors of pairs of variants of pairs of vectors of variants (etc).

martinzink (approved):

lordgamez (approved):

extensions/standard-processors/utils/JoltUtils.cpp
szaszm: This is a good explanation of why I generally prefer `push_back`: https://abseil.io/tips/112

There is no wide consensus about this question in the C++ community. In this case, I think it doesn't matter that much, since we're actually placing newly constructed elements, but I think the slight type safety benefit of push_back is worth paying the possible cost of an extra move.

szaszm (approved):

PR #1834, opened 2024-06-28
MINIFICPP-2423 Fix RHEL8 bootstrap issues

szaszm:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -246,6 +246,16 @@ if [ -z "${CMAKE_COMMAND}" ] ||
   get_cmake_version
 fi
 
+# RHEL8: If cmake3 is too old, try cmake
+if [ "$CMAKE_COMMAND" = "cmake3" ] &&
+   [ "$CMAKE_MAJOR" -lt "$CMAKE_GLOBAL_MIN_VERSION_MAJOR" ] ||
+   [ "$CMAKE_MINOR" -lt "$CMAKE_GLOBAL_MIN_VERSION_MINOR" ] ||
+   [ "$CMAKE_REVISION" -lt "$CMAKE_GLOBAL_MIN_VERSION_REVISION" ] &&
+   [ -x "$(command -v cmake)" ]; then
+  CMAKE_COMMAND="cmake"
+  get_cmake_version
+fi
+
 if [ "$CMAKE_MAJOR" -lt "$CMAKE_GLOBAL_MIN_VERSION_MAJOR" ] ||
    [ "$CMAKE_MINOR" -lt "$CMAKE_GLOBAL_MIN_VERSION_MINOR" ] ||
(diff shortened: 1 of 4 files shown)


bootstrap.sh
```diff
@@ -246,6 +246,16 @@ if [ -z "${CMAKE_COMMAND}" ] ||
   get_cmake_version
 fi
 
+# RHEL8: If cmake3 is too old, try cmake
+if [ "$CMAKE_COMMAND" = "cmake3" ] &&
+   [ "$CMAKE_MAJOR" -lt "$CMAKE_GLOBAL_MIN_VERSION_MAJOR" ] ||
+   [ "$CMAKE_MINOR" -lt "$CMAKE_GLOBAL_MIN_VERSION_MINOR" ] ||
+   [ "$CMAKE_REVISION" -lt "$CMAKE_GLOBAL_MIN_VERSION_REVISION" ] &&
```
lordgamez, line 253: I think there are 2 problems with this:
- Due to the operator precedence the operands are evaluated in this order `([ "$CMAKE_COMMAND" = "cmake3" ] && [ "$CMAKE_MAJOR" -lt "$CMAKE_GLOBAL_MIN_VERSION_MAJOR" ]) || [ "$CMAKE_MINOR" -lt "$CMAKE_GLOBAL_MIN_VERSION_MINOR" ] || ([ "$CMAKE_REVISION" -lt "$CMAKE_GLOBAL_MIN_VERSION_REVISION" ] && [ -x "$(command -v cmake)" ])` while all the parts of the cmake version should be checked together
- I think the cmake version check doesn't work as it supposed to not just here but on line 235 and line 259 as well. For example with this check if the required version is 3.24.1 and our version is 3.25.0 it considers it lower than the required.

szaszm: both should be fixed now

bstrp_functions.sh
```diff
@@ -26,6 +26,24 @@ get_cmake_version(){
   CMAKE_REVISION=$(echo "$CMAKE_VERSION" | cut -d. -f3)
 }
 
+# Checks if a version is smaller than another
+compare_version() {
```
fgerlits, line 30: I think something like `version_is_less_than` would be better than `compare_version`

bstrp_functions.sh
```diff
@@ -457,3 +475,5 @@ read_feature_options(){
     *) echo -e "${RED}Please enter an option A-Z or AA-AF or 1-7...${NO_COLOR}" && sleep 2
   esac
 }
+
+# vim: shiftwidth=2 tabstop=2 expandtab
```
fgerlits, line 479: did you add these vim-specific comments intentionally?

szaszm: yes

bstrp_functions.sh
szaszm: renamed

fgerlits (approved):

lordgamez (approved):

PR #1849, opened 2024-08-01
MINIFICPP-2435 Add attributes to original flow file on failed transform in python

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2435

------------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/docker/test/integration/features/python.feature
+++ b/docker/test/integration/features/python.feature
@@ -156,3 +156,17 @@ Feature: MiNiFi can use python processors in its flows
     When all instances start up
 
     Then one flowfile with the contents "Check successful!" is placed in the monitored directory in less than 30 seconds
+
+  @USE_NIFI_PYTHON_PROCESSORS
+  Scenario: A MiNiFi instance can update attributes through native python processor
+    Given a GenerateFlowFile processor with the "File Size" property set to "0B"
+    And a FailureWithAttributes processor
+    And a LogAttribute processor
+    And python is installed on the MiNiFi agent with a pre-created virtualenv
+
+    And the "success" relationship of the GenerateFlowFile processor is connected to the FailureWithAttributes
+    And the "failure" relationship of the FailureWithAttributes processor is connected to the LogAttribute
+
+    When all instances start up
+
+    Then the Minifi logs contain the following message: "key:error.message value:Error" in less than 60 seconds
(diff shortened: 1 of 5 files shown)


docker/test/integration/features/python.feature
```diff
@@ -143,3 +143,17 @@ Feature: MiNiFi can use python processors in its flows
     When all instances start up
 
     Then one flowfile with the contents "Check successful!" is placed in the monitored directory in less than 30 seconds
+
+  @USE_NIFI_PYTHON_PROCESSORS
+  Scenario: A MiNiFi instance can update attributes through native python processor
+    Given a GenerateFlowFile processor with the "File Size" property set to "0B"
+    And a FailureWithAttributes processor
```
fgerlits, line 150: We could put an UpdateAttribute before the FailureWithAttributes, to verify that the attributes are merged, not replaced. Maybe also set an old value for `error.message`, to verify/document that the existing attribute gets overwritten.

extensions/python/pythonprocessors/nifiapi/flowfiletransform.py
```diff
@@ -98,12 +98,15 @@ def onTrigger(self, context: ProcessContext, session: ProcessSession):
             session.transfer(original_flow_file, self.REL_FAILURE)
             return
 
+        result_attributes = result.getAttributes()
         if result.getRelationship() == "failure":
             session.remove(flow_file)
+            if result_attributes is not None:
+                for attribute in result_attributes:
+                    original_flow_file.addAttribute(attribute, result_attributes[attribute])
```
fgerlits, line 106: I think
```suggestion
                for name, value in result_attributes.items():
                    original_flow_file.addAttribute(name, value)
```
would be nicer

docker/test/integration/features/python.feature
lordgamez: Good point, actually there was a bug in there because in NiFi the python API uses `addAttribute` and we used the same, but in NiFi it updates the attribute if it exists, but MiNiFi C++ doesn't, we have to use `setAttribute` instead. Updated in 30268333d5fff6c48822cb54925b8189ff6d0e49

extensions/python/pythonprocessors/nifiapi/flowfiletransform.py
lordgamez: Updated in 30268333d5fff6c48822cb54925b8189ff6d0e49

fgerlits (approved):

martinzink (approved):

szaszm (approved):

PR #1853, opened 2024-08-06
MINIFICPP-2440 Install all Python dependencies at once

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2440

------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/python/PYTHON.md
+++ b/extensions/python/PYTHON.md
@@ -172,10 +172,11 @@ Due to some differences between the NiFi and MiNiFi C++ processors and implement
 - In MiNiFi C++ when the processor is stopped the stop event handling is done in the `notifyStop()` method which does not have the context available. Due to this the `def onStopped(self, context)` cannot be called in NiFi Python processors, so the `onStopped` method is not supported in MiNiFi C++.
 - The interface of the `ProcessContext` class is a bit more limited in MiNiFi C++ compared to NiFi. The available methods in `ProcessContext` are `getProperty`, `getStateManager`, `getName` and `getProperties`.
 - Success relationship is always present in all Python processors even if custom relationships are defined in the Python processor with the `getRelationships` method.
+- MiNiFi C++ uses a single embedded Python interpreter for all Python processors, so the Python processors share the same Python interpreter. This means that the Python processors cannot have different Python versions or use different Python packages. The Python packages are installed on the system or in a single virtualenv that is shared by all Python processors.
 
 ## Use Python processors from virtualenv
 
-It is possible to set a virtualenv to be used by the Python processors in Apache MiNiFi C++. If the virtualenv directory is set, the Python processors will be executed using the packages installed in the virtualenv. If the virtualenv directory is not set, the Python processors will be executed using the packages installed on the system.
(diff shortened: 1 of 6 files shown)


fgerlits (approved):

szaszm (approved):

extensions/python/PYTHON.md
```diff
@@ -172,10 +172,11 @@ Due to some differences between the NiFi and MiNiFi C++ processors and implement
 - In MiNiFi C++ when the processor is stopped the stop event handling is done in the `notifyStop()` method which does not have the context available. Due to this the `def onStopped(self, context)` cannot be called in NiFi Python processors, so the `onStopped` method is not supported in MiNiFi C++.
 - The interface of the `ProcessContext` class is a bit more limited in MiNiFi C++ compared to NiFi. The available methods in `ProcessContext` are `getProperty`, `getStateManager`, `getName` and `getProperties`.
 - Success relationship is always present in all Python processors even if custom relationships are defined in the Python processor with the `getRelationships` method.
+- MiNiFi C++ uses a single embedded Python interpreter for all Python processors, so the Python processors share the same Python interpreter. This means that the Python processors cannot have different Python versions or use different Python packages. The Python packages are installed on the system or in a single virtualenv that is shared by all Python processors.
```
james94, line 175: In the case of NiFi 2.0, I believe that each new python processor uses its own virtual environment and probably its own python interpreter per processor. Will this be eventually the case for MiNiFi C++'s python processors? If we use NiFi Native Python Processors in MiNiFi C++, will this behavior with python interpreter happen here in MiNiFi C++? @lordgamez @szaszm

extensions/python/PYTHON.md
```diff
@@ -172,10 +172,11 @@ Due to some differences between the NiFi and MiNiFi C++ processors and implement
...
 
 ## Use Python processors from virtualenv
 
-It is possible to set a virtualenv to be used by the Python processors in Apache MiNiFi C++. If the virtualenv directory is set, the Python processors will be executed using the packages installed in the virtualenv. If the virtualenv directory is not set, the Python processors will be executed using the packages installed on the system.
+It is possible to set a virtualenv to be used by the Python processors in Apache MiNiFi C++. This is the default behavior. If the virtualenv directory is set, the Python processors will be executed using the packages installed in the virtualenv or if dependencies are not found there MiNiFi C++ will try to find the required packages installed on the system. If the virtualenv directory is not set, the Python processors will be executed using only the packages installed on the system.
```
james94, line 179: NiFi 2.0 seems to automate the process of using python virtualenvs, will this also be a feature for MiNiFi C++ in the future? @lordgamez @szaszm

lordgamez: Currently there are options to use either a virtualenv or the system packages for the python processors, and if the user wants to manually create the environment with the required packages or should they be installed automatically. The default configuration uses and creates the virtualenv automatically and installs the required packages for the processors similarly to NiFi.

extensions/python/PYTHON.md
lordgamez: I don't think it will be supported in the future due to the nature of the implementation differences between MiNiFi's and NiFi's python support. MiNiFi C++ uses python C API bindings to run python code with an embedded python interpreter in the same MiNiFi process. To create separate interpreters with separate environments for each processor that would probably need separate processes for each processor which is not feasible. As far as I know NiFi uses separate processes for each python processor and the binding with the java code is implemented using a TCP server for communicating between the java and the python implementation. The point of using MiNiFi C++ with python processors instead of NiFi would probably be the lightweight and faster nature of the C++ implementation, which would be lost if we would rewrite the bindings to a similar TCP server communication.

martinzink (approved):

PR #1863, opened 2024-09-05
MINIFICPP-2279 Add support for RecordTransform NiFi python processors

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2279

-------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/docker/test/integration/features/steps/steps.py
+++ b/docker/test/integration/features/steps/steps.py
@@ -25,6 +25,7 @@ from minifi.controllers.ElasticsearchCredentialsService import ElasticsearchCred
 from minifi.controllers.ODBCService import ODBCService
 from minifi.controllers.KubernetesControllerService import KubernetesControllerService
 from minifi.controllers.JsonRecordSetWriter import JsonRecordSetWriter
+from minifi.controllers.JsonRecordSetReader import JsonRecordSetReader
 
 from behave import given, then, when
 from behave.model_describe import ModelDescriptor
@@ -271,6 +272,7 @@ def step_impl(context):
 
 
 @given("a file with the content \"{content}\" is present in \"{path}\"")
+@given("a file with the content '{content}' is present in '{path}'")
 @then("a file with the content \"{content}\" is placed in \"{path}\"")
 def step_impl(context, content, path):
     context.test.add_test_data(path, content)
@@ -401,7 +403,7 @@ def step_impl(context, processor_name):
     processor.set_property('SSL Context Service', ssl_context_service.name)
 
 
-# RecordSetWriters
+# Record set reader and writer
 @given("a JsonRecordSetWriter controller service is set up for {processor_name}")
 def step_impl(context, processor_name):
     json_record_set_writer = JsonRecordSetWriter()
@@ -409,6 +411,21 @@ def step_impl(context, processor_name):
     processor = context.test.get_node_by_name(processor_name)
     processor.controller_services.append(json_record_set_writer)
     processor.set_property('Record Set Writer', json_record_set_writer.name)
+    processor.set_property('Record Writer', json_record_set_writer.name)
+
+
+@given("a JsonRecordSetWriter controller service is set up")
+def step_impl(context):
+    json_record_set_writer = JsonRecordSetWriter(name="JsonRecordSetWriter", output_grouping="Array")
+    container = context.test.acquire_container(context=context, name="minifi-cpp-flow")
+    container.add_controller(json_record_set_writer)
+
+
+@given("a JsonRecordSetReader controller service is set up")
+def step_impl(context):
+    json_record_set_reader = JsonRecordSetReader("JsonRecordSetReader")
+    container = context.test.acquire_container(context=context, name="minifi-cpp-flow")
+    container.add_controller(json_record_set_reader)
 
 
 # Kubernetes
--- a/docker/test/integration/features/steps/steps.py
+++ b/docker/test/integration/features/steps/steps.py
@@ -404,19 +404,9 @@ def step_impl(context, processor_name):
 
 
 # Record set reader and writer
-@given("a JsonRecordSetWriter controller service is set up for {processor_name}")
-def step_impl(context, processor_name):
-    json_record_set_writer = JsonRecordSetWriter()
-
-    processor = context.test.get_node_by_name(processor_name)
-    processor.controller_services.append(json_record_set_writer)
-    processor.set_property('Record Set Writer', json_record_set_writer.name)
-    processor.set_property('Record Writer', json_record_set_writer.name)
-
-
-@given("a JsonRecordSetWriter controller service is set up")
-def step_impl(context):
-    json_record_set_writer = JsonRecordSetWriter(name="JsonRecordSetWriter", output_grouping="Array")
+@given("a JsonRecordSetWriter controller service is set up with \"{}\" output grouping")
+def step_impl(context, output_grouping: str):
+    json_record_set_writer = JsonRecordSetWriter(name="JsonRecordSetWriter", output_grouping=output_grouping)
     container = context.test.acquire_container(context=context, name="minifi-cpp-flow")
     container.add_controller(json_record_set_writer)
 
--- a/extensions/python/pythonprocessors/nifiapi/recordtransform.py
+++ b/extensions/python/pythonprocessors/nifiapi/recordtransform.py
@@ -130,10 +130,10 @@ class RecordTransform(ProcessorBase):
             if result.getRecordJson() is None:
                 continue
             record_partition = result.getPartition()
-            try:
+            if record_partition in partitions:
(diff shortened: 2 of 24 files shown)


extensions/python/types/PyProcessSession.cpp
```diff
@@ -178,14 +178,24 @@ PyObject* PyProcessSessionObject::get(PyProcessSessionObject* self, PyObject*) {
...
   }
 
-  if (auto flow_file = session->create(nullptr))
+  std::shared_ptr<core::FlowFile> parent_flow_file;
+  auto arg_size = PyTuple_Size(args);
+  if (arg_size > 0) {
+    PyObject* script_flow_file = nullptr;
+    if (!PyArg_ParseTuple(args, "O!", PyScriptFlowFile::typeObject(), &script_flow_file)) {
+      return nullptr;
+    }
```
szaszm, line 194: I'm not sure returning null when the parent flow file is missing is the best approach. What about throwing an error when it's present, but invalid?

extensions/python/types/PyProcessSession.h
```diff
@@ -39,6 +39,7 @@ class PyProcessSession {
   void remove(const std::shared_ptr<core::FlowFile>& flow_file);
   std::string getContentsAsString(const std::shared_ptr<core::FlowFile>& flow_file);
   void putAttribute(const std::shared_ptr<core::FlowFile>& flow_file, std::string_view key, const std::string& value);
+  gsl::not_null<core::ProcessSession*> getSession() const { return session_; }
```
szaszm, line 42: This should return a reference instead.

extensions/python/types/PyRecordSetReader.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+static PyType_Spec PyRecordSetReaderTypeSpec{
+    .name = "minifi_native.RecordSetReader",
+    .basicsize = sizeof(PyRecordSetReader),
+    .itemsize = 0,
+    .flags = Py_TPFLAGS_DEFAULT,
+    .slots = PyRecordSetReaderTypeSpecSlots
+};
+
+int PyRecordSetReader::init(PyRecordSetReader* self, PyObject* args, PyObject*) {
```
szaszm, line 49: If the parameters are not nullable, they should be changed to references, or if the python API doesn't allow that, we could add a `gsl_Expects` check. If they CAN be null, and it's a user error, then we should create a python error or exception.
I'm pretty sure at least self is not nullable, because it's dereferenced without a check on line 58.

extensions/python/types/PyProcessSession.cpp
lordgamez: When we define a method using the python C API, if the return value is nullptr it signals the python C API that an error has occurred and it actually results in an exception on the python side, so it should already result in an exception.

extensions/python/types/PyRecordSetReader.cpp
lordgamez: It doesn't seem to be possible to use references due to the python C API only provides pointers for these methods. As I checked if the methods are bound to an object the self should never be null and the args parameter should always contain a value as it is passed as a tuple even if the tuple is empty. But in edge cases like memory corruption it could happen that these result in null values so it doesn't hurt to check them anyway, so I added the checks in 1366b3b397a13faa54769543083ff9b46290ebf7

extensions/python/types/PyProcessSession.h
lordgamez: Good point, updated in 1366b3b397a13faa54769543083ff9b46290ebf7

extensions/python/types/PyRecordSetReader.cpp
szaszm: If you think the checks are not worth the cost, we can change them to `gsl_ExpectsAudit`. It's like `gsl_Expects` in its documentation role, but by default the checks are not actually compiled in. But I don't think runtime cost is an issue here.

lordgamez: Sorry I didn't want to suggest the checks are not worth it, I think it's a good idea to keep the `gsl_Expects` check for any edge case that can occur.

extensions/python/types/PyProcessSession.h
```diff
@@ -39,10 +39,11 @@ class PyProcessSession {
   void remove(const std::shared_ptr<core::FlowFile>& flow_file);
   std::string getContentsAsString(const std::shared_ptr<core::FlowFile>& flow_file);
   void putAttribute(const std::shared_ptr<core::FlowFile>& flow_file, std::string_view key, const std::string& value);
+  core::ProcessSession& getSession() const { return session_; }
 
  private:
   std::vector<std::shared_ptr<core::FlowFile>> flow_files_;
-  gsl::not_null<core::ProcessSession*> session_;
+  core::ProcessSession& session_;
```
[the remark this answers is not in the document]
lordgamez: I think in this case our goal is not to allow move or reassignment as the session member should be the same for the lifetime of the `PyProcessSession` object. This is consistent with the `LuaProcessSession` where the session member is also a reference. I would keep this change due to this.

szaszm (approved):

libminifi/src/core/RecordField.cpp
```diff
@@ -0,0 +1,99 @@
+/**
+* Licensed to the Apache Software Foundation (ASF) under one or more
```
fgerlits, line 2: very minor, but a space is missing from the start of this line

docker/test/integration/features/steps/steps.py
```diff
@@ -401,14 +403,29 @@ def step_impl(context, processor_name):
...
     processor.controller_services.append(json_record_set_writer)
     processor.set_property('Record Set Writer', json_record_set_writer.name)
+    processor.set_property('Record Writer', json_record_set_writer.name)
+
+
+@given("a JsonRecordSetWriter controller service is set up")
+def step_impl(context):
+    json_record_set_writer = JsonRecordSetWriter(name="JsonRecordSetWriter", output_grouping="Array")
+    container = context.test.acquire_container(context=context, name="minifi-cpp-flow")
+    container.add_controller(json_record_set_writer)
```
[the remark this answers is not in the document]
lordgamez: Good idea, updated in ddaff1ec9ab40fb6070d0274bba7bb4b9e226e13

extensions/python/pythonprocessors/nifiapi/recordtransform.py
```diff
@@ -0,0 +1,152 @@
...
+        for result in results:
+            if result.getRecordJson() is None:
+                continue
+            record_partition = result.getPartition()
+            try:
+                partition_index = partitions.index(record_partition)
+                partitioned_results_list[partition_index].append(result)
+            except ValueError:
+                partitions.append(record_partition)
+                partitioned_results_list.append([result])
```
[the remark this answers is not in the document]
lordgamez: Makes sense, updated in ddaff1ec9ab40fb6070d0274bba7bb4b9e226e13

extensions/python/pythonprocessors/nifiapi/recordtransform.py
```diff
@@ -0,0 +1,152 @@
...
+            if result.getRelationship() == "success":
+                session.transfer(partitioned_flow_file, self.REL_SUCCESS)
+            else:
+                session.transferToCustomRelationship(partitioned_flow_file, result.getRelationship())
+
+        session.transfer(flow_file, self.REL_ORIGINAL)
+
+    @abstractmethod
+    def transform(self, context: ProcessContextProxy, flowFile: FlowFileProxy) -> RecordTransformResult:
+        pass
```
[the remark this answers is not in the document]
lordgamez: Good catch, fixed in ddaff1ec9ab40fb6070d0274bba7bb4b9e226e13

extensions/python/types/PyRecordSetReader.cpp
```diff
@@ -0,0 +1,112 @@
...
+    return -1;
+  self->record_set_reader_ = *static_cast<HeldType*>(record_set_reader);
+  return 0;
+}
+
+PyObject* PyRecordSetReader::read(PyRecordSetReader* self, PyObject* args) {
+  gsl_Expects(self && args);
+  auto record_set_reader = self->record_set_reader_.lock();
+  if (!record_set_reader) {
+    PyErr_SetString(PyExc_AttributeError, "tried reading ssl context service outside 'on_trigger'");
```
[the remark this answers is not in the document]
lordgamez: Updated in ddaff1ec9ab40fb6070d0274bba7bb4b9e226e13

libminifi/src/core/RecordField.cpp
lordgamez: Updated in ddaff1ec9ab40fb6070d0274bba7bb4b9e226e13

fgerlits (approved):

PR #1877, opened 2024-10-10
MINIFICPP-2463 TailFile doesnt pickup the data after the last delimit…

martinzink:
…er in rotated files

The actual change amidst the many clangtidy fixes and TailFile test optimizations (runtime went from ~6s to ~1.6s):

I've added a new is_rotated_: bool field to TailState, thats only set to true during the findRotatedFilesAfterLastReadTime fn, and we not only commit the split flowfile if its ended with delimiter but also if its rotated and not empty

---

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -724,24 +723,24 @@ void TailFile::processSingleFile(core::ProcessSession& session,
 void TailFile::updateFlowFileAttributes(const std::filesystem::path& full_file_name, const TailState& state,
                                         const std::filesystem::path& fileName, const std::string& baseName,
                                         const std::string& extension,
-                                        std::shared_ptr<core::FlowFile> &flow_file) const {
-  logger_->log_info("TailFile {} for {} bytes", fileName, flow_file->getSize());
-  std::string logName = textfragmentutils::createFileName(baseName, extension, state.position_, flow_file->getSize());
-  flow_file->setAttribute(core::SpecialFlowAttribute::PATH, state.path_.string());
-  flow_file->addAttribute(core::SpecialFlowAttribute::ABSOLUTE_PATH, full_file_name.string());
-  flow_file->setAttribute(core::SpecialFlowAttribute::FILENAME, logName);
+                                        core::FlowFile& flow_file) const {
```
james94, line 726: Interesting way to update the Flow File attributes switching from shared pointer to pass by reference. I see you also dereferenced the pointer as you passed *flow_file into updateFlowFileAttributes. Were you noticing problems with the shared pointer pass by reference previous approach?

martinzink: No particular problems but this makes the signature more clear and removes unnecessary restrictions, so this could help future refactors.
  this function could easily work on flowfiles owned by unique_ptrs or flowfiles on the stack
  it also ensures that the flowfile is not supposed to be a nullptr and moves the responsibility to ensure it is not null to the caller side (when it deferences the shared_ptr for example)

szaszm: @james94 passing everything by shared_ptr is a half decade old legacy of the codebase that we're trying to reduce. In most cases, we can assign each object to a clear owner, with unique_ptr, which makes it easier to reason about lifetimes, and ensures that we don't accidentally create shared_ptr ownership cycles like [MINIFICPP-839](https://issues.apache.org/jira/browse/MINIFICPP-839)

extensions/standard-processors/processors/TailFile.cpp
```diff
@@ -695,8 +694,8 @@ void TailFile::processSingleFile(core::ProcessSession& session,
       auto flow_file = session.create();
       session.write(flow_file, std::ref(file_reader));
 
-      if (file_reader.useLatestFlowFile()) {
-        updateFlowFileAttributes(full_file_name, state_copy, fileName, baseName, extension, flow_file);
+      if (file_reader.endedWithDelimiter() || (state.is_rotated_ && flow_file->getSize() > 0)) {
```
szaszm, line 697: is this the fix?

martinzink: yeah

> The actual change amidst the many clangtidy fixes and TailFile test optimizations (runtime went from ~6s to ~1.6s):
> 
> I've added a new is_rotated_: bool field to TailState, thats only set to true during the findRotatedFilesAfterLastReadTime fn, and we not only commit the split flowfile if its ended with delimiter but also if its rotated and not empty

lordgamez (approved):

szaszm (approved):

extensions/standard-processors/processors/TailFile.h
```diff
@@ -253,21 +253,21 @@ class TailFile : public core::Processor {
...
-                                std::shared_ptr<core::FlowFile> &flow_file) const;
+                                core::FlowFile& flow_file) const;
   static void updateStateAttributes(TailState &state, uint64_t size, uint64_t checksum);
-  bool isOldFileInitiallyRead(TailState &state) const;
+  bool isOldFileInitiallyRead(const TailState &state) const;
 
   static const char *CURRENT_STR;
   static const char *POSITION_STR;
-  static const int BUFFER_SIZE = 512;
+  static constexpr int BUFFER_SIZE = 512;
```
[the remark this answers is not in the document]
martinzink: Good idea but I rather do that in a separate PR, I've created a jira so we can address all of these inner buffer sizes together (possibly making them configurable) https://issues.apache.org/jira/browse/MINIFICPP-2483

extensions/standard-processors/tests/unit/TailFileTests.cpp
fgerlits, line 1: The refactoring changes all look like they are improvements, but they make it difficult to see if there is a specific new test which validates the bugfix. Is there one?

martinzink: There is no new test however I modified the `TEST_CASE("TailFile finds and finishes the renamed file and continues with the new log file", "[rotation]")` removed the deliberatly added delimiter after the rotation, so now it would fail without the fix
previously line 820
``` 
in_file_stream << DELIM;
in_file_stream.close();
```

fgerlits: OK, thanks!

PR #1883, opened 2024-10-21
MINIFICPP-2480 workaround python version incompatibilities around 3.6 vs 3.11

szaszm:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/python/PythonInterpreter.cpp
+++ b/extensions/python/PythonInterpreter.cpp
@@ -54,7 +54,7 @@ std::optional<version> getPythonVersion() {
   //                  "3.12.6 (main, Sep  8 2024, 13:18:56) [GCC 14.2.1 20240805]"
   std::string ver_str = Py_GetVersion();
   std::smatch match;
-  if (std::regex_search(ver_str, match, std::regex{"^(\\d+)\\.(\\d+)"})) {
+  if (std::regex_search(ver_str, match, std::regex{R"(^(\d+)\.(\d+))"})) {
     return version{std::stoi(match[1]), std::stoi(match[2])};
   } else {
     return std::nullopt;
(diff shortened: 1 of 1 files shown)


lordgamez (approved):

martinzink (approved):

extensions/python/PythonInterpreter.cpp
```diff
@@ -34,30 +44,53 @@ GlobalInterpreterLock::~GlobalInterpreterLock() {
...
 // PyEval_InitThreads might be marked deprecated (depending on the version of Python.h)
-// Python <= 3.6: This needs to be called manually after Py_Initialize to initialize threads
+// Python <= 3.6: This needs to be called manually after Py_Initialize to initialize threads (python < 3.6 is unsupported by us)
 // Python >= 3.7: Noop function since its functionality is included in Py_Initialize
 // Python >= 3.9: Marked as deprecated (still noop)
+// Python >= 3.11: removed
 // This can be removed if we drop the support for Python 3.6
 void initThreads() {
+  using namespace std::literals;
+  if (const auto version = getPythonVersion(); version->major == 3 && version->minor <= 6) { return; }
```
fgerlits, line 72:
```suggestion
  if (const auto version = getPythonVersion(); version && version->major == 3 && version->minor <= 6) { return; }
```
or we can log something and terminate if `version` is `nullopt`, but let's not just crash

fgerlits, line 72: I think the logic backwards: we want to call `PyEval_InitThreads()` if the version is <= 3.6, not when it isn't.

szaszm: you're right, this was an oversight. The version check was also wrong, because the early return is supposed to happen on versions newer than 3.6

szaszm: yes, my bad

szaszm: I would do the early return on nullopt instead, so `!version || (version->major == 3 && version->minor > 6)`. Not that it matters much, since it should never be nullopt, but in case some in unknown future version, parsing of the version string fails, I guess we should just not call these legacy functions there.

fgerlits: or it could be `!(version && version->major == 3 && version->minor <= 6)`, so we also return early on versions >= 4.0

fgerlits (approved):

adamdebreceni (approved):

PR #1888, opened 2024-10-30
MINIFICPP-2376 Add support for parameter context inheritance

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2376

------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/CONFIGURE.md
+++ b/CONFIGURE.md
@@ -139,12 +139,25 @@ Processor properties in flow configurations can be parameterized using parameter
  - `#` character can be used to escape the parameter syntax. E.g. if the `parameterName` parameter's value is `xxx` then `#{parameterName}` will be replaced with `xxx`, `##{parameterName}` will be replaced with `#{parameterName}`, and `#####{parameterName}` will be replaced with `##xxx`.
  - Sensitive parameters can only be assigned to sensitive properties and non-sensitive parameters can only be assigned to non-sensitive properties.
 
+Parameter contexts can be inherited from multiple other parameter contexts. The order of the parameter inheritance matters in the way that if a parameter is present in multiple parameter contexts, the parameter value assigned to the first parameter in the inheritance order will be used. No circular inheritance is allowed.
+
 An example for using parameters in a JSON configuration file:
 
 ```json
 {
     "parameterContexts": [
         {
+            "identifier": "235e6b47-ea22-45cd-a472-545801db98e6",
+            "name": "common-parameter-context",
+            "description": "Common parameter context",
+            "parameters": [
+                {
+                    "name": "common_timeout",
+                    "description": "Common timeout seconds",
+                    "sensitive": false,
+                    "value": "30"
+                }
+            ],
             "identifier": "804e6b47-ea22-45cd-a472-545801db98e6",
             "name": "root-process-group-context",
             "description": "Root process group parameter context",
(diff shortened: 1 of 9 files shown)


libminifi/src/core/flow/StructuredConfiguration.cpp
```diff
@@ -145,6 +145,41 @@ std::unique_ptr<core::ProcessGroup> StructuredConfiguration::getRootFrom(const N
...
+void StructuredConfiguration::verifyNoInheritanceCycles() const {
+  std::unordered_set<std::string> visited_parameter_contexts;
+  std::unordered_set<std::string> current_stack;
+  for (const auto& [parameter_context_name, parameter_context] : parameter_contexts_) {
+    if (hasInheritanceCycle(*parameter_context, visited_parameter_contexts, current_stack)) {
+      throw std::invalid_argument("Circular references in Parameter Context inheritance are not allowed. Inheritance cycle was detected in parameter context '" + parameter_context->getName() + "'");
+    }
+  }
+}
+
```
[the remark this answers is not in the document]
lordgamez: I'm not sure this requires a timeout or any restriction. The cycle detection run a depth-first search algorithm which has linear complexity based on the number of parameter contexts, so it should not be running that long even for large number of parameter contexts. But I suppose even in flows where parameter contexts could be generated, like using a parameter provider that would generate parameters based on file hierarchies the number of contexts should only be in the 1000s at most which would still not cause a problem with algorithm.

martinzink (approved):

szaszm (approved):

CONFIGURE.md
```diff
@@ -139,12 +139,25 @@ Processor properties in flow configurations can be parameterized using parameter
...
+            "description": "Common parameter context",
+            "parameters": [
+                {
+                    "name": "common_timeout",
+                    "description": "Common timeout seconds",
+                    "sensitive": false,
+                    "value": "30"
+                }
+            ],
             "identifier": "804e6b47-ea22-45cd-a472-545801db98e6",
```
fgerlits, line 161: I think a `}, {` is missing between these two parameter context definitions

libminifi/src/core/flow/StructuredConfiguration.cpp
```diff
@@ -139,6 +139,41 @@ std::unique_ptr<core::ProcessGroup> StructuredConfiguration::getRootFrom(const N
...
+  return false;
+}
+}  // namespace
+
+void StructuredConfiguration::verifyNoInheritanceCycles() const {
+  std::unordered_set<std::string> visited_parameter_contexts;
+  std::unordered_set<std::string> current_stack;
+  for (const auto& [parameter_context_name, parameter_context] : parameter_contexts_) {
+    if (hasInheritanceCycle(*parameter_context, visited_parameter_contexts, current_stack)) {
+      throw std::invalid_argument("Circular references in Parameter Context inheritance are not allowed. Inheritance cycle was detected in parameter context '" + parameter_context->getName() + "'");
```
fgerlits, line 172: very minor, but we could use `parameter_context_name` instead of `parameter_context->getName()`

CONFIGURE.md
lordgamez: Good catch, updated in https://github.com/apache/nifi-minifi-cpp/pull/1888/commits/1703d452516773353a1aa7587f17cb448235bb24

libminifi/src/core/flow/StructuredConfiguration.cpp
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1888/commits/1703d452516773353a1aa7587f17cb448235bb24

PR #1897, opened 2024-11-18
MINIFICPP-2491 Upgrade GRPC and Abseil libs

lordgamez:
Fixes Windows build CI failure

-----------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/thirdparty/grpc/remove-custom-commands-for-unused-dependencies.patch
+++ b/thirdparty/grpc/remove-custom-commands-for-unused-dependencies.patch
@@ -1,0 +1,425 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index fc4b6a40a5..ae494f1464 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -489,7 +489,7 @@ if(WIN32)
+   set(_gRPC_ALLTARGETS_LIBRARIES ${_gRPC_ALLTARGETS_LIBRARIES} ws2_32 crypt32)
+   set(_gRPC_STATIC_WIN32 STATIC)
+ endif()
+-  
++
+ if(BUILD_SHARED_LIBS AND WIN32)
+ 
+ # Currently for shared lib on Windows (i.e. a DLL) certain bits of source code
+@@ -890,411 +890,6 @@ protobuf_generate_grpc_cpp_with_import_path_correction(
+ protobuf_generate_grpc_cpp_with_import_path_correction(
+   test/core/tsi/alts/fake_handshaker/transport_security_common.proto test/core/tsi/alts/fake_handshaker/transport_security_common.proto
+ )
+-protobuf_generate_grpc_cpp_with_import_path_correction(
+-  third_party/envoy-api/envoy/annotations/deprecation.proto envoy/annotations/deprecation.proto
+-)
+-protobuf_generate_grpc_cpp_with_import_path_correction(
+-  third_party/envoy-api/envoy/annotations/resource.proto envoy/annotations/resource.proto
+-)
+-protobuf_generate_grpc_cpp_with_import_path_correction(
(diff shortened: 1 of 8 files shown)


thirdparty/grpc/remove-custom-commands-for-unused-dependencies.patch
```diff
@@ -0,0 +1,425 @@
...
++++ b/CMakeLists.txt
+@@ -489,7 +489,7 @@ if(WIN32)
+   set(_gRPC_ALLTARGETS_LIBRARIES ${_gRPC_ALLTARGETS_LIBRARIES} ws2_32 crypt32)
+   set(_gRPC_STATIC_WIN32 STATIC)
+ endif()
+-  
++
+ if(BUILD_SHARED_LIBS AND WIN32)
+ 
+ # Currently for shared lib on Windows (i.e. a DLL) certain bits of source code
```
fgerlits, line 13: This section of the patch looks like it's just a whitespace difference. Is it needed?

fgerlits (approved):

thirdparty/grpc/remove-custom-commands-for-unused-dependencies.patch
lordgamez: No it's not needed, my editor automatically removes line ending whitespaces so it probably added this when I created the diff.

thirdparty/grpc/remove-custom-commands-for-unused-dependencies.patch
szaszm, line 1: why are these dependencies unused, but still present?

lordgamez: Some libraries like the googleapis and evoy-api are built with the Grpc library as a thirdparty for cloud API support, but we don't use them for our use case, so we don't load those submodules and don't build them. These custom commands are used to generate protocol buffer interface for these APIs, but it would fail as we do not load those thirdparties.

thirdparty/grpc/remove-custom-commands-for-unused-dependencies.patch
lordgamez: Removed in https://github.com/apache/nifi-minifi-cpp/pull/1897/commits/18fc86102c9cce88d4189e2187a7653d629c5cf7

martinzink (approved):

adamdebreceni (approved):

szaszm (changes requested):
CI failure

thirdparty/grpc/remove-custom-commands-for-unused-dependencies.patch
lordgamez: Readded the whitespace change and added a note to be aware of the reverse patch dry run issue in https://github.com/apache/nifi-minifi-cpp/pull/1897/commits/116696f1b51ee689bb2e36aa6187489216d39d39

thirdparty/grpc/fix-constants-on-windows.patch
```diff
@@ -0,0 +1,17 @@
+diff --git a/third_party/upb/upb/message/internal/message.c b/third_party/upb/upb/message/internal/message.c
+index 57ab2d7371..5f321c7bb2 100644
+--- a/third_party/upb/upb/message/internal/message.c
++++ b/third_party/upb/upb/message/internal/message.c
```
szaszm, line 4:
```
D:\a\nifi-minifi-cpp\nifi-minifi-cpp\build\_deps\grpc-src\third_party\protobuf\upb\message\internal\message.c(22,14): error C2099: initializer is not a constant
const double kUpb_NaN = NAN;
```
The path seems different in the CI failure that happened after this patch was added.

szaszm (approved):

PR #928, opened 2020-10-20
MINIFICPP-1394 - Speed up ConsumeWindowsEventLog processor

adamdebreceni:
Thank you for submitting a contribution to Apache NiFi - MiNiFi C++.

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [ ] Is there a JIRA ticket associated with this PR? Is it referenced
     in the commit message?

- [ ] Does your PR title start with MINIFICPP-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.

- [ ] Has your PR been rebased against the latest commit within the target branch (typically main)?

- [ ] Is your initial contribution a single, squashed commit?

### For code changes:
- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?
- [ ] If applicable, have you updated the LICENSE file?
- [ ] If applicable, have you updated the NOTICE file?

### For documentation related changes:
- [ ] Have you ensured that format looks appropriate for the output in which it is rendered?

### Note:
Please ensure that once the PR is submitted, you check GitHub Actions CI results for build issues and submit an update to your PR as soon as possible.

PR #1900, opened 2024-11-26
MINIFICPP-2494 windows event log: fix event message formatting, refactor

szaszm:
Changes:
- change `enum METADATA` to `enum class Metadata`
- rewrite `getEventData` and `getEventMessage`. They no longer try with a stack buffer first. That looks like a premature optimization, since there are allocations in the function either way, and a user reported issues with the call that may be resolved by this. The rewrite aims to follow modern C++ best practices, instead of the unsafe C-like code it used to be.
- remove a few unused includes
- make `getComputerName` thread-safe and simpler. The max computer name is reduced to 256, based on the MSDN example. I think the overwhelming majority of computer FQDNs will fit in that just fine. Also changed it to truncate when it's longer, instead of bailing out and returning "N/A"
- other various drive-by refactors

----------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/windows-event-log/tests/MetadataWalkerTests.cpp
+++ b/extensions/windows-event-log/tests/MetadataWalkerTests.cpp
@@ -27,7 +27,7 @@
 #include "wel/XMLString.h"
 #include "pugixml.hpp"
 
-using METADATA = org::apache::nifi::minifi::wel::METADATA;
+using Metadata = org::apache::nifi::minifi::wel::Metadata;
 using MetadataWalker = org::apache::nifi::minifi::wel::MetadataWalker;
 using WindowsEventLogMetadata = org::apache::nifi::minifi::wel::WindowsEventLogMetadata;
 using WindowsEventLogMetadataImpl = org::apache::nifi::minifi::wel::WindowsEventLogMetadataImpl;
@@ -74,7 +74,7 @@ const short event_type_index = 178;  // NOLINT short comes from WINDOWS API
 
 class FakeWindowsEventLogMetadata : public WindowsEventLogMetadata {
  public:
-  [[nodiscard]] std::string getEventData(EVT_FORMAT_MESSAGE_FLAGS flags) const override { return "event_data_for_flag_" + std::to_string(flags); }
+  [[nodiscard]] nonstd::expected<std::string, std::error_code> getEventData(EVT_FORMAT_MESSAGE_FLAGS flags) const noexcept override { return "event_data_for_flag_" + std::to_string(flags); }
   [[nodiscard]] std::string getEventTimestamp() const override { return "event_timestamp"; }
   short getEventTypeIndex() const override { return event_type_index; }  // NOLINT short comes from WINDOWS API
 };
@@ -149,7 +149,7 @@ void extractMappingsTestHelper(const std::string &file_name,
                                bool update_xml,
                                bool resolve,
                                const std::map<std::string, std::string>& expected_identifiers,
-                               const std::map<METADATA, std::string>& expected_metadata,
+                               const std::map<Metadata, std::string>& expected_metadata,
                                const std::map<std::string, std::string>& expected_field_values) {
   std::string input_xml = readFile(file_name);
   REQUIRE(!input_xml.empty());
@@ -175,12 +175,12 @@ TEST_CASE("MetadataWalker extracts mappings correctly when there is a single Sid
 
   const std::map<std::string, std::string> expected_identifiers{{"S-1-0-0", "S-1-0-0"}};
 
-  using org::apache::nifi::minifi::wel::METADATA;
-  const std::map<METADATA, std::string> expected_metadata{
-      {METADATA::SOURCE, "Microsoft-Windows-Security-Auditing"},
-      {METADATA::TIME_CREATED, "event_timestamp"},
-      {METADATA::EVENTID, "4672"},
-      {METADATA::EVENT_RECORDID, "2575952"}};
+  using org::apache::nifi::minifi::wel::Metadata;
+  const std::map<Metadata, std::string> expected_metadata{
+      {Metadata::SOURCE, "Microsoft-Windows-Security-Auditing"},
+      {Metadata::TIME_CREATED, "event_timestamp"},
+      {Metadata::EVENTID, "4672"},
+      {Metadata::EVENT_RECORDID, "2575952"}};
 
   const std::map<std::string, std::string> expected_field_values{};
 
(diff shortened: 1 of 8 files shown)


extensions/windows-event-log/tests/MetadataWalkerTests.cpp
```diff
@@ -74,7 +74,7 @@ const short event_type_index = 178;  // NOLINT short comes from WINDOWS API
 
 class FakeWindowsEventLogMetadata : public WindowsEventLogMetadata {
  public:
-  [[nodiscard]] std::string getEventData(EVT_FORMAT_MESSAGE_FLAGS flags) const override { return "event_data_for_flag_" + std::to_string(flags); }
+  [[nodiscard]] nonstd::expected<std::string, std::error_code> getEventData(EVT_FORMAT_MESSAGE_FLAGS flags) const noexcept override { return "event_data_for_flag_" + std::to_string(flags); }
```
szaszm, line 77: Changed this to `noexcept`. There are allocations here. If they fail, the test will terminate and report failure, which is an acceptable behavior IMO.

extensions/windows-event-log/wel/MetadataWalker.cpp
```diff
@@ -90,10 +86,11 @@ bool MetadataWalker::for_each(pugi::xml_node &node) {
...
       std::function<std::string(const std::string &)> updateFunc = [&](const std::string &input) -> std::string {
         if (resolve_) {
-          auto resolved = windows_event_log_metadata_.getEventData(it->second);
-          if (!resolved.empty()) {
-            return resolved;
+          const auto resolved = windows_event_log_metadata_.getEventData(it->second);
+          if (resolved && !resolved->empty()) {
+            return *resolved;
           }
+          // TODO(szaszm): add error logging
```
szaszm, line 93: One behavior change in `getEventData`: allocation failure now returns an `error_code`, instead of an empty string. Since both empty strings and errors are ignored here, there is no change in behavior. Added a comment to do proper error handling later, but I didn't wanna do it in the context of this change. (`MetadataWalker` has no logger.)

extensions/windows-event-log/wel/MetadataWalker.cpp
```diff
@@ -123,40 +120,25 @@ std::vector<std::string> MetadataWalker::getIdentifiers(const std::string &text)
...
-    }
-    return "N/A";
-}
-
-std::map<std::string, std::string> MetadataWalker::getFieldValues() const {
-  return fields_values_;
-}
-
-std::map<std::string, std::string> MetadataWalker::getIdentifiers() const {
-  return replaced_identifiers_;
```
szaszm, line 159: moved inline to the header

extensions/windows-event-log/wel/WindowsEventLog.cpp
```diff
@@ -203,8 +191,4 @@ std::string WindowsEventLogHeader::createDefaultDelimiter(size_t length) const {
   }
 }
 
-EVT_HANDLE WindowsEventLogHandler::getMetadata() const {
-  return metadata_provider_.get();
-}
```
szaszm, line 208: moved inline to the header

extensions/windows-event-log/wel/WindowsEventLog.h
```diff
@@ -59,22 +59,19 @@ enum METADATA {
   UNKNOWN
 };
 
+using METADATA_NAMES = std::vector<std::pair<Metadata, std::string>>;
 
-// this is a continuous enum, so we can rely on the array
-
-using METADATA_NAMES = std::vector<std::pair<METADATA, std::string>>;
+nonstd::expected<std::string, std::error_code> formatEvent(gsl::not_null<EVT_HANDLE> metadata, gsl::not_null<EVT_HANDLE> event, EVT_FORMAT_MESSAGE_FLAGS flags) noexcept;
```
szaszm, line 64: `EVT_HANDLE` is an alias to `HANDLE` which is an alias to `void*`

extensions/windows-event-log/wel/WindowsEventLog.h
```diff
@@ -83,76 +80,50 @@ class WindowsEventLogHandler {
 class WindowsEventLogMetadata {
  public:
   virtual ~WindowsEventLogMetadata() = default;
-  [[nodiscard]] virtual std::string getEventData(EVT_FORMAT_MESSAGE_FLAGS flags) const = 0;
+  [[nodiscard]] virtual nonstd::expected<std::string, std::error_code> getEventData(EVT_FORMAT_MESSAGE_FLAGS flags) const noexcept = 0;
   [[nodiscard]] virtual std::string getEventTimestamp() const = 0;
   virtual short getEventTypeIndex() const = 0;  // NOLINT short comes from WINDOWS API
 
-  static std::string getMetadataString(METADATA val) {
-    static std::map<METADATA, std::string> map = {
```
szaszm, line 91: this was unused

extensions/windows-event-log/wel/WindowsEventLog.h
```diff
@@ -83,76 +80,50 @@ class WindowsEventLogHandler {
...
+    else if (val == "USER") { return USER; }
+    else if (val == "COMPUTER") { return COMPUTER; }
+    else { return UNKNOWN; }
   }
 
   static std::string getComputerName() {
-    static std::string computer_name;
-    if (computer_name.empty()) {
-      char buff[10248];
+    static std::string computer_name = []() -> std::string {
```
szaszm, line 105: Moved this to an immediately invoked lambda: this way the thread-safe static initialization takes care of only running the initialization once, and subsequent calls will just copy the static string. The old version was not thread-safe.

extensions/windows-event-log/wel/WindowsEventLog.h
```diff
@@ -186,17 +157,16 @@ class WindowsEventLogHeader {
 template<typename MetadataCollection>
 std::string WindowsEventLogHeader::getEventHeader(const MetadataCollection& metadata_collection) const {
   std::stringstream eventHeader;
-
   for (const auto& option : header_names_) {
-    auto name = option.second;
+    const auto& name = option.second;
```
szaszm, line 161: `option.second` is a `std::string`

libminifi/src/utils/OsUtils.cpp
```diff
@@ -296,7 +296,7 @@ int64_t OsUtils::getTotalPagingFileSize() {
   return total_paging_file_size;
 }
 
-std::error_code OsUtils::windowsErrorToErrorCode(DWORD error_code) {
+std::error_code OsUtils::windowsErrorToErrorCode(DWORD error_code) noexcept {
```
szaszm, line 299: No allocations happen here, so it might as well be `noexcept`

extensions/windows-event-log/wel/WindowsEventLog.h
```diff
@@ -59,22 +59,19 @@ enum METADATA {
   UNKNOWN
 };
 
+using METADATA_NAMES = std::vector<std::pair<Metadata, std::string>>;
 
-// this is a continuous enum, so we can rely on the array
-
-using METADATA_NAMES = std::vector<std::pair<METADATA, std::string>>;
+nonstd::expected<std::string, std::error_code> formatEvent(EVT_HANDLE metadata, EVT_HANDLE event, EVT_FORMAT_MESSAGE_FLAGS flags) noexcept;
```
szaszm, line 64: `EVT_HANDLE` is `void*`, but I couldn't wrap it in `gsl::not_null`, because it's incompatible with `void*`. (`operator*` cannot return a reference to `void`)

adamdebreceni:
I wouldn't call this premature optimization, this rewrite had a measurable 20% throughput increase: https://github.com/apache/nifi-minifi-cpp/pull/928#issuecomment-712732711

szaszm:
my bad then. Back to draft, and I'll wait for confirmation to see if this fixes the crash.

PR #1904, opened 2024-12-05
MINIFICPP-2485 fix: Support Expression Language in InvokeHTTP "Remote URL" property

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2485

----------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/standard-processors/processors/InvokeHTTP.cpp
+++ b/extensions/standard-processors/processors/InvokeHTTP.cpp
@@ -38,12 +38,13 @@ namespace org::apache::nifi::minifi::processors {
 namespace invoke_http {
 
 HttpClientStore::HttpClientWrapper HttpClientStore::getClient(const std::string& url) {
-  std::unique_lock<std::mutex> lock(clients_mutex_);
-  for (auto it = unused_clients_.begin(); it != unused_clients_.end(); ++it) {
-    if ((*it)->getURL() == url) {
-      used_clients_.splice(used_clients_.end(), unused_clients_, it);
-      return {*this, **it};
-    }
+  std::unique_lock lock(clients_mutex_);
+  const auto it = std::find_if(std::begin(unused_clients_), std::end(unused_clients_), [&url](const auto& client) {
+    return client->getURL() == url;
+  });
+  if (it != std::end(unused_clients_)) {
+    used_clients_.splice(used_clients_.end(), unused_clients_, it);
+    return {*this, **it};
   }
 
   if (used_clients_.size() + unused_clients_.size() < max_size_) {
@@ -59,18 +60,16 @@ HttpClientStore::HttpClientWrapper HttpClientStore::getClient(const std::string&
   }
 }
 
-void HttpClientStore::returnClient(minifi::http::HTTPClient& client) {
-  std::unique_lock<std::mutex> lock(clients_mutex_);
-  for (auto it = used_clients_.begin(); it != used_clients_.end(); ++it) {
-    if (it->get() != &client) {
-      continue;
-    }
-    unused_clients_.splice(unused_clients_.end(), used_clients_, it);
-    lock.unlock();
-    cv_.notify_one();
-    return;
+void HttpClientStore::returnClient(http::HTTPClient& client) {
+  std::unique_lock lock(clients_mutex_);
+  const auto it = std::find_if(std::begin(used_clients_), std::end(used_clients_),
+    [&client](const auto& elem) { return &client == elem.get(); });
+  if (it == std::end(used_clients_)) {
+    logger_->log_error("Couldn't find HTTP client in client store to be returned");
   }
-  logger_->log_error("Couldn't find HTTP client in client store to be returned");
+  unused_clients_.splice(unused_clients_.end(), used_clients_, it);
+  lock.unlock();
+  cv_.notify_one();
 }
 
 }  // namespace invoke_http
@@ -161,8 +160,8 @@ void InvokeHTTP::setupMembersFromProperties(const core::ProcessContext& context)
   }
 }
 
-gsl::not_null<std::unique_ptr<minifi::http::HTTPClient>> InvokeHTTP::createHTTPClientFromMembers(const std::string& url) const {
-  auto client = std::make_unique<minifi::http::HTTPClient>();
+gsl::not_null<std::unique_ptr<http::HTTPClient>> InvokeHTTP::createHTTPClientFromMembers(const std::string& url) const {
+  auto client = std::make_unique<http::HTTPClient>();
   client->initialize(method_, url, ssl_context_service_);
   setupClientTimeouts(*client, connect_timeout_, read_timeout_);
   client->setHTTPProxy(proxy_);
--- a/extensions/standard-processors/processors/InvokeHTTP.cpp
+++ b/extensions/standard-processors/processors/InvokeHTTP.cpp
@@ -39,43 +39,33 @@ namespace invoke_http {
 
 HttpClientStore::HttpClientWrapper HttpClientStore::getClient(const std::string& url) {
   std::unique_lock<std::mutex> lock(clients_mutex_);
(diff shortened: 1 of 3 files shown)


extensions/standard-processors/processors/InvokeHTTP.h
```diff
@@ -41,9 +42,60 @@
...
+
+ private:
+  friend class ::org::apache::nifi::minifi::test::HttpClientStoreTestAccessor;
+  void returnClient(minifi::http::HTTPClient& client);
+
+  std::mutex clients_mutex_;
+  std::condition_variable cv_;
+  const size_t max_size_;
+  size_t clients_created_{0};
+  std::vector<std::pair<std::unique_ptr<minifi::http::HTTPClient>, bool>> clients_;
```
adamdebreceni, line 94: since we are rotating a single element each time I think an `std::list` with `splice` would be clearer, also I think we should replace the `clients_created_` with `std::list::size` to act as a single source of truth

lordgamez: Could you elaborate a bit on this, please? How would you use the `splice` with `std::list` in a way that would replace the rotate function?

We were also discussing using an `std::list` instead of the `std::vector` previously with @szaszm and I don't remember why we decided against it. @szaszm can you maybe remember?

adamdebreceni:
```
#include <list>
#include <algorithm>
#include <vector>
#include <iostream>

int main() {
  // std::rotate(
  //   clients_.begin() + gsl::narrow<int64_t>(i),
  //   clients_.begin() + gsl::narrow<int64_t>(i + 1),
  //   clients_.begin() + gsl::narrow<int64_t>(clients_created_)
  // )
  std::list<int> clients{1, 2, 3, 4};
  for (auto it = clients.begin(); it != clients.end(); ++it) {
    if (*it == 2) {
      clients.splice(clients.end(), clients, it);
      break;
    }
  }
  std::cout << "after rotateLeft\n";  // 1, 3, 4, 2
  for (auto& val : clients) {
    std::cout << val << std::endl;
  }

  clients = std::list<int>{1, 2, 3, 4};
  // rotateRight(
  //   clients_.begin() + last_unused + 1,
  //   clients_.begin() + gsl::narrow<int64_t>(i) + 1,
  //   1
  // );
  clients.splice(std::next(clients.begin(), 1), clients, std::next(clients.begin(), 2));

  std::cout << "after rotateRight\n";  // 1, 3, 2, 4
  for (auto& val : clients) {
    std::cout << val << std::endl;
  }
}
```

adamdebreceni: I don't think it brings any performance benefits or things like that (I don't imagine we will have more than a handful of cached clients so the linear scanning works just fine), just that IMO it is a better fit to how we are actually using this container

lordgamez: Thanks for the example I see your point now, I think it's a bit simpler and removing the need to maintain the `clients_created_ variable` is worth this change, I updated it in https://github.com/apache/nifi-minifi-cpp/pull/1904/commits/028d6b5e699dca283b7b0d706bacbc79887dcb8a

adamdebreceni (approved):

extensions/standard-processors/processors/InvokeHTTP.h
```diff
@@ -41,9 +43,59 @@
...
+ private:
+  friend class ::org::apache::nifi::minifi::test::HttpClientStoreTestAccessor;
+  void returnClient(minifi::http::HTTPClient& client);
+
+  std::mutex clients_mutex_;
+  std::condition_variable cv_;
+  const size_t max_size_;
+  std::list<gsl::not_null<std::unique_ptr<minifi::http::HTTPClient>>> used_clients_;
+  std::list<gsl::not_null<std::unique_ptr<minifi::http::HTTPClient>>> unused_clients_;
+  std::function<gsl::not_null<std::unique_ptr<minifi::http::HTTPClient>>(const std::string&)> create_client_function_;
```
szaszm, line 95: I think we could do without the extra indirection of unique_ptr. List splicing doesn't move the data on the heap.

extensions/standard-processors/processors/InvokeHTTP.cpp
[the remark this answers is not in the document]
lordgamez: How would we make sure that the client returned to the HttpClientStore was a client managed by this client store in this case? Or should we just accept any clients? I'm not sure if we should manage clients not created by the same client store. Also if the used clients list is removed it should probably still needs to be replaced by at least a counter to see how many clients are created and in use to be compared with the defined max size.

extensions/standard-processors/processors/InvokeHTTP.h
lordgamez: That's true, unfortunately we would still need to copy or move the HTTPClient on line 56 and line 176 and those constructors are delete from the HTTPClient class.

extensions/standard-processors/processors/InvokeHTTP.cpp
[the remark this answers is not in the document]
szaszm: Since this is just an internal utility, I don't think it needs enforcing that clients are not mixed up. You wrote the code using the store, and it's not mixing items from two different stores. The store is not used anywhere else, so I don't think it's a problem.

A counter would be a good approach to track the max size IMO.

extensions/standard-processors/processors/InvokeHTTP.h
szaszm: line 56: It could be an in-place constructed to the list. Emplace and returning it by value from the create function should be able to do that.
line 176: return by value

extensions/standard-processors/processors/InvokeHTTP.cpp
```diff
@@ -35,6 +35,45 @@
...
+    return {*this, *used_clients_.back()};
+  }
+}
+
+void HttpClientStore::returnClient(minifi::http::HTTPClient& client) {
+  std::unique_lock<std::mutex> lock(clients_mutex_);
+  for (auto it = used_clients_.begin(); it != used_clients_.end(); ++it) {
+    if (it->get() != &client) {
+      continue;
+    }
```
szaszm, line 67: This should be a `find_if`.

extensions/standard-processors/processors/InvokeHTTP.cpp
```diff
@@ -35,6 +35,45 @@
...
+namespace invoke_http {
+
+HttpClientStore::HttpClientWrapper HttpClientStore::getClient(const std::string& url) {
+  std::unique_lock<std::mutex> lock(clients_mutex_);
+  for (auto it = unused_clients_.begin(); it != unused_clients_.end(); ++it) {
+    if ((*it)->getURL() == url) {
+      used_clients_.splice(used_clients_.end(), unused_clients_, it);
+      return {*this, **it};
+    }
+  }
```
szaszm, line 47: Another find_if

extensions/standard-processors/processors/InvokeHTTP.h
szaszm: I checked and returning by value doesn't seem to work without copy/move. I thought it could be elided. So nevermind, leave them as pointers.

extensions/standard-processors/processors/InvokeHTTP.cpp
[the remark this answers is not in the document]
szaszm: Thinking about it again, I don't see it as an issue. I still would've done it differently, but it's fine as is.

fgerlits (approved):

extensions/standard-processors/processors/InvokeHTTP.cpp
lordgamez: Thanks, I'm okay with these changes.

szaszm (approved):

martinzink (approved):

PR #1925, opened 2025-02-06
MINIFICPP-2524 Add FIPS compliancy support

lordgamez:
This change adds the option to enable FIPS mode in the OpenSSL library for cryptographic functions used by MiNiFi C++. This includes the following:
- Introduces the `nifi.openssl.fips.support.enable` option in `minifi.properties` to enable this option, if the option is enabled MiNiFi C++ will try to load the `openssl.cnf` that loads the FIPS provider and then enables FIPS mode in OpenSSL.
- Adds compilation of the FIPS provider library from OpenSSL 3.0.9 which is the latest FIPS validated OpenSSL version as listed here: https://openssl-library.org/source/
- Adds the FIPS provider library and the required `openssl.cnf` files to the install package under the $MINIFI_HOME/fips directory
- Adds the `openssl` binary to the install package under the $MINIFI_HOME/fips directory for running the module tests and generating the fipsmodule.cnf file on the target platform as required for the FIPS compliancy, referenced in these sources: https://github.com/openssl/openssl/discussions/25036, https://openssl-library.org/source/fips-doc/openssl-3.0.9-security-policy-2024-01-12.pdf Appendix A
- Updated paho-mqtt, librdkafka libraries and added `no-engine` flag for OpenSSL compilation to remove legacy API usage required for FIPS compliancy as referenced here: https://docs.openssl.org/master/man7/fips_module/#description

OpenSSL discussion about this issue: https://github.com/openssl/openssl/discussions/26378

https://issues.apache.org/jira/browse/MINIFICPP-2524

---------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


minifi_main/MiNiFiMain.cpp
```diff
@@ -183,6 +187,100 @@ void writeSchemaIfRequested(const argparse::ArgumentParser& parser, const std::s
...
+
+  if (!EVP_default_properties_enable_fips(nullptr, 1)) {
+    logger->log_error("Failed to enable FIPS mode");
+    std::exit(1);
+  }
+
+  if (!EVP_default_properties_is_fips_enabled(nullptr)) {
+    logger->log_error("FIPS mode is not enabled");
+    std::exit(1);
+  }
```
fgerlits, line 279: We should use `ERR_print_errors_fp()` to print the OpenSSL error(s) in these error cases.

lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1925/commits/a893f09f0598bcc9d38bcf2a23bee3f06868c985

minifi_main/MiNiFiMain.cpp
```diff
@@ -183,6 +187,100 @@ void writeSchemaIfRequested(const argparse::ArgumentParser& parser, const std::s
...
+  std::string content = buffer.str();
+  input_file.close();
+
+  const std::string placeholder = "${MINIFI_HOME}";
+  size_t pos = content.find(placeholder, 0);
+  if (pos == std::string::npos) {
+    return true;
+  }
+
+  std::replace(minifi_home_path.begin(), minifi_home_path.end(), '\\', '/');
```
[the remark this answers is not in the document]
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1925/commits/a893f09f0598bcc9d38bcf2a23bee3f06868c985

minifi_main/MiNiFiMain.cpp
```diff
@@ -333,6 +431,8 @@ int main(int argc, char **argv) {
     configure->loadConfigureFile(DEFAULT_NIFI_PROPERTIES_FILE);
     overridePropertiesFromCommandLine(argument_parser, configure);
 
+    initializeFipsMode(configure, minifiHome, logger);
```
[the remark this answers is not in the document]
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1925/commits/a893f09f0598bcc9d38bcf2a23bee3f06868c985

cmake/BundledOpenSSL.cmake
```diff
@@ -59,7 +65,8 @@ function(use_openssl SOURCE_DIR BINARY_DIR)
             no-module           # disable dynamically loadable engines
             no-pinshared        # don't pin shared libraries in the process memory
             enable-tfo          # Enable TCP Fast Open
-            no-ssl)             # disable SSLv3
+            no-ssl              # disable SSLv3
+            no-engine)
```
szaszm, line 69: What does this exactly do? INSTALL.md says:
> Don't build support for loading engines.

But I'm not sure what loadable engines are. Is it related to FIPS in any way? Also, the FIPS target omits most of these, probably leading to some unnecessary extra build steps, consider adding some more options there, like no-tests, no-docs.

minifi_main/MiNiFiMain.cpp
szaszm, line 1: I think the new functions should be extracted to a separate file.

cmake/BundledOpenSSL.cmake
lordgamez: I'm not that familiar with it, but there is an Engine API in OpenSSL that is controlled by this option. An engine that is loaded in OpenSSL is a hardware or software implementation used for performing cryptographic operations. Engines are not FIPS compatible and corresponding API is deprecated since OpenSSL 3.0, so for FIPS compliancy we should remove the engine support as well. (According to the documention it is also a good guidline for FIPS compliancy to remove anything that is deprecated in OpenSSL 3.0 as that is probably not FIPS compatible.)

minifi_main/MiNiFiMain.cpp
szaszm: I don't think `utils` is the place for that. It's only relevant for the agent process.

cmake/BundledOpenSSL.cmake
```diff
@@ -51,15 +57,12 @@ function(use_openssl SOURCE_DIR BINARY_DIR)
 
     set(OPENSSL_EXTRA_FLAGS
             no-tests            # Disable tests
-            no-apps             # disable executables
             no-capieng          # disable CAPI engine (legacy)
-            no-dso              # disable dynamic libraries
             no-docs             # disable docs and manpages
             no-legacy           # disable legacy modules
-            no-module           # disable dynamically loadable engines
-            no-pinshared        # don't pin shared libraries in the process memory
```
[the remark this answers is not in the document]
lordgamez: - no-apps is removed because we need the openssl binary to be packaged with fips
- no-dso and no-module need to be enabled for shared libraries to be loaded in our case the fips provider which cannot be statically linked
- IMO no-pinshared should not be enabled to have the loaded fips provider available through the lifetime of the application and avoid the accidental unloading of the fips shared library

minifi_main/MiNiFiMain.cpp
lordgamez: Moved to `minifi_main` in https://github.com/apache/nifi-minifi-cpp/pull/1925/commits/cbf74e02f1f7e12855229bdfbdde1ede810a2aa6

cmake/BundledOpenSSL.cmake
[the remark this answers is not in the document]
szaszm: It's fine with fips, but these are the library options, not the fips target options.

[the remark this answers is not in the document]
lordgamez: But these options are meant for the library options for loading the FIPS module not for the FIPS target.
- we need the binary of the latest OpenSSL target, not the FIPS target
- Without the options to be able to load shared libraries loading of the FIPS module fails with the following error:
```
Failed to load FIPS module
INSTALL FAILED
4077F3DFF87D0000:error:07880025:common libcrypto routines:provider_init:reason(37):crypto/provider_core.c:949:name=fips
```

minifi_main/MiNiFiMain.cpp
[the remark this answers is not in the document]
lordgamez: Had to revert this change in https://github.com/apache/nifi-minifi-cpp/pull/1925/commits/84a48c6a513e94f143d0fb15b4008efaeb8e1284 as loading FIPS needs to happen before the extension manager is initialized to work with the extensions

cmake/BundledOpenSSL.cmake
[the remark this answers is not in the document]
lordgamez: From what I read in the description I think it's better to keep shared libraries pinned, in our case the fips provider, as without this it could happen in runtime that the library is unloaded when unused, I think it's safer to have it loaded through the agent's lifetime. If FIPS is unused then it should not have any affect as no other shared libraries are used.

cmake/BundledOpenSSL.cmake
```diff
@@ -165,4 +169,77 @@ function(use_openssl SOURCE_DIR BINARY_DIR)
...
+                SOURCE_DIR "${BINARY_DIR}/thirdparty/openssl-fips-src"
+                BUILD_IN_SOURCE true
+                CONFIGURE_COMMAND ./Configure "CFLAGS=${PASSTHROUGH_CMAKE_C_FLAGS} -fPIC" "CXXFLAGS=${PASSTHROUGH_CMAKE_CXX_FLAGS} -fPIC" ${OPENSSL_SHARED_FLAG} ${OPENSSL_FIPS_EXTRA_FLAGS}  "--prefix=${OPENSSL_FIPS_BIN_DIR}" "--openssldir=${OPENSSL_FIPS_BIN_DIR}"
+                BUILD_BYPRODUCTS ${OPENSSL_FIPS_FILE_LIST}
+                EXCLUDE_FROM_ALL TRUE
+                INSTALL_COMMAND make install_fips
+        )
+    endif()
+
+    add_dependencies(OpenSSL::Crypto openssl-fips-external)
```
szaszm, line 243: is this necessary? I thought these two could be built in parallel.

lordgamez: This is needed to actually build the fips target. Without setting it as a dependency of any target in the build (as it is used only in runtime), it will not be built because no target has this library linked in build time and cmake will not make it part of the build hierarchy. The two openssl builds can actually run in parallel as it does not depend on the openssl-external target. OpenSSL::Crypto depends on both openssl-external and openssl-fips-external targets so they can be built in parallel.

martinzink (approved):

PR #1930, opened 2025-02-18
Working on api reduction

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extension-utils/include/utils/ProcessorConfigUtils.h
+++ b/extension-utils/include/utils/ProcessorConfigUtils.h
@@ -25,6 +25,7 @@
 #include "core/ProcessContext.h"
 #include "core/PropertyType.h"
 #include "utils/Enum.h"
+#include "minifi-cpp/Exception.h"
 
 namespace org::apache::nifi::minifi::utils {
 
--- a/extensions/expression-language/ProcessContextExpr.h
+++ b/extensions/expression-language/ProcessContextExpr.h
@@ -34,35 +34,19 @@ namespace org::apache::nifi::minifi::core {
  */
 class ProcessContextExpr final : public core::ProcessContextImpl {
  public:
-  ProcessContextExpr(Processor& processor, controller::ControllerServiceProvider* controller_service_provider,
-                     const std::shared_ptr<core::Repository> &repo, const std::shared_ptr<core::Repository> &flow_repo,
-                     const std::shared_ptr<core::ContentRepository> &content_repo = core::repository::createFileSystemRepository())
-      : core::ProcessContextImpl(processor, controller_service_provider, repo, flow_repo, content_repo),
-        logger_(logging::LoggerFactory<ProcessContextExpr>::getLogger()) {
-  }
-
-  ProcessContextExpr(Processor& processor, controller::ControllerServiceProvider* controller_service_provider,
-                     const std::shared_ptr<core::Repository> &repo, const std::shared_ptr<core::Repository> &flow_repo, const std::shared_ptr<minifi::Configure> &configuration,
-                     const std::shared_ptr<core::ContentRepository> &content_repo = core::repository::createFileSystemRepository())
-      : core::ProcessContextImpl(processor, controller_service_provider, repo, flow_repo, configuration, content_repo),
-        logger_(logging::LoggerFactory<ProcessContextExpr>::getLogger()) {
-  }
-
   ~ProcessContextExpr() override = default;
 
-  nonstd::expected<std::string, std::error_code> getProperty(std::string_view name, const FlowFile*) const override;
-  nonstd::expected<std::string, std::error_code> getDynamicProperty(std::string_view name, const FlowFile*) const override;
+  nonstd::expected<std::string, std::error_code> getProperty(ProcessContext& context, std::string_view name, const FlowFile*) const override;
+  nonstd::expected<std::string, std::error_code> getDynamicProperty(ProcessContext& context, std::string_view name, const FlowFile*) const override;
+//
+//  nonstd::expected<void, std::error_code> setProperty(std::string_view name, std::string value);
+//  nonstd::expected<void, std::error_code> setDynamicProperty(std::string name, std::string value);
 
-  nonstd::expected<void, std::error_code> setProperty(std::string_view name, std::string value) override;
-  nonstd::expected<void, std::error_code> setDynamicProperty(std::string name, std::string value) override;
-
-  std::map<std::string, std::string> getDynamicProperties(const FlowFile*) const override;
+  std::map<std::string, std::string> getDynamicProperties(ProcessContext& context, const FlowFile*) const override;
 
  private:
   mutable std::unordered_map<std::string, expression::Expression, utils::string::transparent_string_hash, std::equal_to<>> cached_expressions_;
   mutable std::unordered_map<std::string, expression::Expression, utils::string::transparent_string_hash, std::equal_to<>> cached_dynamic_expressions_;
-
-  std::shared_ptr<logging::Logger> logger_;
 };
 
 }  // namespace org::apache::nifi::minifi::core
--- a/extensions/python/ExecutePythonProcessor.cpp
+++ b/extensions/python/ExecutePythonProcessor.cpp
@@ -31,16 +31,12 @@
 #include "utils/StringUtils.h"
 #include "utils/file/FileUtils.h"
 #include "utils/ProcessorConfigUtils.h"
-#include "utils/PropertyErrors.h"
+#include "minifi-cpp/utils/PropertyErrors.h"
+#include "minifi-cpp/core/ProcessorDescriptor.h"
 
 namespace org::apache::nifi::minifi::extensions::python::processors {
 
-void ExecutePythonProcessor::initialize() {
-  if (getSupportedProperties().empty()) {
-    setSupportedProperties(Properties);
-    setSupportedRelationships(Relationships);
-  }
-
+void ExecutePythonProcessor::initialize(core::ProcessorDescriptor& self) {
   if (processor_initialized_) {
     logger_->log_debug("Processor has already been initialized, returning...");
     return;
@@ -56,6 +52,13 @@ void ExecutePythonProcessor::initialize() {
   // so that we can provide manifest of processor identity on C2
   python_script_engine_ = createScriptEngine();
   initalizeThroughScriptEngine();
+
+  std::vector<core::PropertyReference> all_properties{Properties.begin(), Properties.end()};
+  for (auto& python_prop : python_properties_) {
+    all_properties.push_back(python_prop.getReference());
+  }
+  self.setSupportedProperties(all_properties);
+  self.setSupportedRelationships(Relationships);
 }
 
 void ExecutePythonProcessor::initalizeThroughScriptEngine() {
@@ -80,7 +83,7 @@ void ExecutePythonProcessor::initalizeThroughScriptEngine() {
 }
 
 void ExecutePythonProcessor::onSchedule(core::ProcessContext& context, core::ProcessSessionFactory& /*sessionFactory*/) {
-  addAutoTerminatedRelationship(Original);
+  context.addAutoTerminatedRelationship(Original);
   if (!processor_initialized_) {
     loadScript();
     python_script_engine_ = createScriptEngine();
@@ -220,63 +223,8 @@ void ExecutePythonProcessor::addProperty(const std::string &name, const std::str
   python_properties_.emplace_back(property);
 }
 
-nonstd::expected<std::string, std::error_code> ExecutePythonProcessor::getProperty(std::string_view name) const {
-  if (auto non_python_property = ConfigurableComponentImpl::getProperty(name)) {
-    return *non_python_property;
-  }
-  std::lock_guard<std::mutex> lock(python_properties_mutex_);
-  auto it = ranges::find_if(python_properties_, [&name](const auto& item){
-    return item.getName() == name;
-  });
-  if (it != python_properties_.end()) {
-    return it->getValue() | utils::transform([](const std::string_view value_view) -> std::string { return std::string{value_view}; });
-  }
-  return nonstd::make_unexpected(core::PropertyErrorCode::NotSupportedProperty);
-}
-
-nonstd::expected<void, std::error_code> ExecutePythonProcessor::setProperty(std::string_view name, std::string value) {
-  auto set_non_python_property = ConfigurableComponentImpl::setProperty(name, value);
-  if (set_non_python_property || set_non_python_property.error() != core::PropertyErrorCode::NotSupportedProperty) {
-    return set_non_python_property;
-  }
-  std::lock_guard<std::mutex> lock(python_properties_mutex_);
-  auto it = ranges::find_if(python_properties_, [&name](const auto& item){
-    return item.getName() == name;
-  });
-  if (it != python_properties_.end()) {
-    return it->setValue(std::move(value));
-  }
-  return nonstd::make_unexpected(core::PropertyErrorCode::NotSupportedProperty);
-}
-
-nonstd::expected<core::PropertyReference, std::error_code> ExecutePythonProcessor::getPropertyReference(std::string_view name) const {
-  if (const auto non_python_property = ConfigurableComponentImpl::getPropertyReference(name)) {
-    return *non_python_property;
-  }
-  std::lock_guard<std::mutex> lock(python_properties_mutex_);
-  auto it = ranges::find_if(python_properties_, [&name](const auto& item){
-    return item.getName() == name;
-  });
-  if (it != python_properties_.end()) {
-    return it->getReference();
-  }
-  return nonstd::make_unexpected(core::PropertyErrorCode::NotSupportedProperty);
-}
-
-std::map<std::string, core::Property> ExecutePythonProcessor::getSupportedProperties() const {
-  auto result = ConfigurableComponentImpl::getSupportedProperties();
-
-  std::lock_guard<std::mutex> lock(python_properties_mutex_);
-
-  for (const auto &property : python_properties_) {
-    result.insert({ property.getName(), property });
-  }
-
-  return result;
-}
-
 std::vector<core::Relationship> ExecutePythonProcessor::getPythonRelationships() const {
-  auto relationships = getSupportedRelationships();
+  std::vector<core::Relationship> relationships{Relationships.begin(), Relationships.end()};
   auto custom_relationships = python_script_engine_->getCustomPythonRelationships();
   relationships.reserve(relationships.size() + std::distance(custom_relationships.begin(), custom_relationships.end()));
   relationships.insert(relationships.end(), custom_relationships.begin(), custom_relationships.end());
--- a/extensions/standard-processors/processors/DefragmentText.h
+++ b/extensions/standard-processors/processors/DefragmentText.h
@@ -115,7 +115,6 @@ class DefragmentText : public core::ProcessorImpl {
   void onSchedule(core::ProcessContext& context, core::ProcessSessionFactory& session_factory) override;
   void onTrigger(core::ProcessContext& context, core::ProcessSession& session) override;
   void restore(const std::shared_ptr<core::FlowFile>& flowFile) override;
-  std::set<core::Connectable*> getOutGoingConnections(const std::string &relationship) override;
 
  protected:
   class Buffer {
--- a/libminifi/src/core/FlowConfiguration.cpp
+++ b/libminifi/src/core/FlowConfiguration.cpp
@@ -26,6 +26,7 @@
 #include "processors/ProcessorUtils.h"
 #include "utils/StringUtils.h"
 #include "utils/file/FileUtils.h"
+#include "core/ProcessorProxy.h"
 
 namespace org::apache::nifi::minifi::core {
 
--- a/libminifi/src/core/Processor.cpp
+++ b/libminifi/src/core/Processor.cpp
@@ -1,0 +1,379 @@
+/**
+ * @file Processor.cpp
+ * Processor class implementation
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "core/Processor.h"
+
+#include <ctime>
+#include <cctype>
+
+#include <memory>
+#include <set>
+#include <string>
+#include <vector>
+
+#include "minifi-cpp/Connection.h"
+#include "core/Connectable.h"
+#include "core/logging/LoggerFactory.h"
+#include "minifi-cpp/core/ProcessorConfig.h"
+#include "minifi-cpp/core/ProcessContext.h"
+#include "minifi-cpp/core/ProcessorDescriptor.h"
+#include "minifi-cpp/core/ProcessSessionFactory.h"
+#include "utils/gsl.h"
+#include "range/v3/algorithm/any_of.hpp"
+#include "fmt/format.h"
+#include "Exception.h"
+#include "core/ProcessorProxy.h"
+#include "core/ProcessorMetrics.h"
+
+using namespace std::literals::chrono_literals;
+
+namespace org::apache::nifi::minifi::core {
+
+ProcessorProxy::ProcessorProxy(std::string_view name, std::unique_ptr<ProcessorApi> impl)
+    : ConnectableImpl(name),
+      state_(DISABLED),
+      scheduling_period_(MINIMUM_SCHEDULING_PERIOD),
+      run_duration_(DEFAULT_RUN_DURATION),
+      yield_period_(DEFAULT_YIELD_PERIOD_SECONDS),
+      active_tasks_(0),
+      logger_(logging::LoggerFactory<Processor>::getLogger(uuid_)) {
+  has_work_.store(false);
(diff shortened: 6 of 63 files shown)


extensions/python/ExecutePythonProcessor.cpp
```diff
@@ -58,6 +54,17 @@ void ExecutePythonProcessor::initialize() {
   initalizeThroughScriptEngine();
 }
 
+void ExecutePythonProcessor::initialize() {
+  initializeScript();
+  std::vector<core::PropertyReference> all_properties{Properties.begin(), Properties.end()};
+  for (auto& python_prop : python_properties_) {
+    all_properties.push_back(python_prop.getReference());
+  }
```
szaszm:
```suggestion
  ranges::transform(python_properties_, std::back_inserter(all_properties), &core::Property::getReference);
```

extensions/python/ExecutePythonProcessor.cpp
```diff
@@ -125,22 +134,18 @@ void ExecutePythonProcessor::loadScriptFromFile() {
...
-  std::string script_body = getProperty(ScriptBody.name).value_or("");
-  if (script_file.empty() && script_body.empty()) {
+  if (script_file_path_.empty() && script_to_exec_.empty()) {
     throw std::runtime_error("Neither Script Body nor Script File is available to execute");
   }
 
-  if (!script_file.empty()) {
-    if (!script_body.empty()) {
+  if (!script_file_path_.empty()) {
+    if (!script_to_exec_.empty()) {
```
szaszm, line 140: line 147 (return statement) can be removed now

extensions/expression-language/ProcessContextExpr.h
```diff
@@ -34,25 +34,12 @@ namespace org::apache::nifi::minifi::core {
...
-        logger_(logging::LoggerFactory<ProcessContextExpr>::getLogger()) {
-  }
-
+  using ProcessContextImpl::ProcessContextImpl;
   ~ProcessContextExpr() override = default;
 
   nonstd::expected<std::string, std::error_code> getProperty(std::string_view name, const FlowFile*) const override;
   nonstd::expected<std::string, std::error_code> getDynamicProperty(std::string_view name, const FlowFile*) const override;
-
+//
```
lordgamez: This can be removed

utils/src/core/ProcessorImpl.cpp
lordgamez, line 1: Could we also rename the other *Impl class files to *Impl.h and *Impl.cpp for consistency? It's also easier to find the correct file when searching for it in the editor. If that's a larger change we can do it under a separate jira ticket in a follow up PR.

extension-utils/include/utils/ProcessorConfigUtils.h
lordgamez, line 1: Having `utils` and `extension-utils` still sounds like placeholder names and definitely should be renamed, especially as both of them have another `utils` subdirectory. If it's a more complicated change we can have it in a separate PR. As most of the files under `utils` and `extension-utils` not really general utilities but rather the core framework and extension framework implementation it could be something like `core-framework`-`extension-framework`, `core-impl`-`extension-impl`(as the counterpart of the minifi-api), or `core-defaults`-`extension-defaults` (for the default implementation), but we can workshop these a bit more.

libminifi/test/libtest/unit/TestBase.h
```diff
@@ -201,13 +201,56 @@ class TempDirectory {
   bool is_owner_;
 };
 
+template<typename T>
+class TypedProcessorWrapper {
```
lordgamez, line 205: Just curious, what was the reason for adding a wrapper for the tested processors?

libminifi/include/Port.h
```diff
@@ -29,10 +30,13 @@ enum class PortType {
   OUTPUT
 };
 
-class Port final : public ForwardingNode {
+class Port;
```
lordgamez, line 33: Do we need a forward declaration here?

libminifi/include/Port.h
```diff
@@ -44,4 +48,15 @@ class Port final : public ForwardingNode {
   PortType port_type_;
 };
 
+class Port : public core::Processor {
```
lordgamez, line 53: Why do we need a separate Port and PortImpl here?

adamdebreceni: Port is handled as a processor in the system, since a lot of Processor forwards the call to a ProcessorApi (which has its default methods in ProcessorImpl), we either a) make Port not-a-processor, b) we virtualize all forwarding calls in Processor just to override in Port to not to forward, reimplementing some default behavior already in ProcessorImpl, the best course of action is to keep the semantics port-is-a-processor and mirror the same changes are for processor implementations

libminifi/include/Port.h
adamdebreceni: yes for the logger

libminifi/test/libtest/unit/TestBase.h
adamdebreceni: in a lot of tests we set processor-type specific fields, previously we dynamic_cast all over the place (unchecked), this one substitutes that

utils/src/core/ProcessorImpl.cpp
adamdebreceni: yes, there are quite a few such cases, I created a jira for it: https://issues.apache.org/jira/browse/MINIFICPP-2582

extensions/expression-language/ProcessContextExpr.h
adamdebreceni: removed

extension-utils/include/utils/ProcessorConfigUtils.h
adamdebreceni: renamed

lordgamez (approved):

extensions/python/ExecutePythonProcessor.cpp
adamdebreceni: removed

core-framework/include/core/ProcessorImpl.h
```diff
@@ -0,0 +1,160 @@
...
+
+class ProcessContext;
+class ProcessSession;
+class ProcessSessionFactory;
+
+#define BUILDING_DLL 1
+
+class ProcessorImpl : public virtual ProcessorApi {
+ public:
+  explicit ProcessorImpl(ProcessorMetadata info);
```
fgerlits: Nitpicking, but I find it confusing that these `ProcessorMetadata` parameters are named `info`, when there is a different `ProcessorInfo` class.  Could you rename them to `metadata` or `construction_params` or something similar?

extensions/python/ExecutePythonProcessor.cpp
```diff
@@ -58,6 +54,15 @@ void ExecutePythonProcessor::initialize() {
   initalizeThroughScriptEngine();
 }
 
+void ExecutePythonProcessor::initialize() {
+  initializeScript();
+  std::vector<core::PropertyReference> all_properties{Properties.begin(), Properties.end()};
+  ranges::transform(python_properties_, std::back_inserter(all_properties), &core::Property::getReference);
+  setSupportedProperties(all_properties);
+  setSupportedRelationships(Relationships);
+  logger_->log_debug("Processor has already been initialized, returning...");
```
fgerlits: It could be useful to make this log different from the log at line 41:
```suggestion
  logger_->log_debug("Processor has been initialized.");
```

extensions/standard-processors/processors/RouteText.cpp
```diff
@@ -61,14 +58,12 @@ void RouteText::onSchedule(core::ProcessContext& context, core::ProcessSessionFa
...
 
-    for (const auto& property_name : getDynamicPropertyKeys()) {
+    for (const auto& property_name : context.getDynamicPropertyKeys()) {
       core::RelationshipDefinition rel{property_name, "Dynamic Route"};
       dynamic_relationships_[property_name] = rel;
       relationships.push_back(rel);
       logger_->log_info("RouteText registered dynamic route '{}'", property_name);
     }
-
-    setSupportedRelationships(relationships);
```
fgerlits, line 71: why was this line removed? now the block before it has no effect

minifi-api/include/minifi-cpp/core/ProcessContext.h
```diff
@@ -33,14 +32,37 @@
 
 namespace org::apache::nifi::minifi::core {
 
+namespace detail {
+template<typename T>
+concept NotAFlowFile = !std::convertible_to<T &, const FlowFile &> && !std::convertible_to<T &, const std::shared_ptr<FlowFile> &>;
+}  // namespace detail
```
fgerlits: this concept is not used anywhere

extensions/smb/tests/ListAndFetchSmbTests.cpp
fgerlits, line 1: a couple of `.get().`s are missing in lines 68 and 69

core-framework/include/core/ProcessorImpl.h
adamdebreceni: renamed to metadata

extensions/python/ExecutePythonProcessor.cpp
adamdebreceni: changed

minifi-api/include/minifi-cpp/core/ProcessContext.h
adamdebreceni: removed

extensions/smb/tests/ListAndFetchSmbTests.cpp
adamdebreceni: added

extensions/standard-processors/processors/RouteText.cpp
adamdebreceni: it did not really make sense to call setSupportedRelationships here since the manifest is already populated with Class::Relationships and get/is SupportedRelationships is not really used anywhere, but since it could be used flow parse-time to verify a connection, I would not yet remove these methods

core-framework/include/core/ProcessorFactory.h
```diff
@@ -0,0 +1,57 @@
...
+    return std::make_unique<T>(metadata);
+  }
+
+  std::string getClassName() const override {
+    return class_name_;
+  }
+
+ protected:
+  std::string group_name_;
+  std::string class_name_;
```
szaszm: At least class_name_, but maybe both could become a string_view:
- class_name_ comes from core::className which returns a string_view to static strings
- group_name_ seems to come from different places, but they seem to be static strings too.

minifi-api/include/minifi-cpp/core/ClassLoader.h
```diff
@@ -62,15 +65,19 @@ class ClassLoader {
...
-  virtual std::optional<std::string> getGroupForClass(const std::string &class_name) const = 0;
+  [[nodiscard]] virtual std::optional<std::string> getGroupForClass(const std::string &class_name) const = 0;
+
+  [[nodiscard]] virtual std::unique_ptr<CoreComponent> instantiate(const std::string &class_name, const std::string &name, std::function<bool(CoreComponent*)> filter) = 0;
 
-  virtual std::unique_ptr<CoreComponent> instantiate(const std::string &class_name, const std::string &name, std::function<bool(CoreComponent*)> filter) = 0;
+  [[nodiscard]] virtual std::unique_ptr<CoreComponent> instantiate(const std::string &class_name, const utils::Identifier &uuid, std::function<bool(CoreComponent*)> filter) = 0;
 
-  virtual std::unique_ptr<CoreComponent> instantiate(const std::string &class_name, const utils::Identifier &uuid, std::function<bool(CoreComponent*)> filter) = 0;
+  [[nodiscard]] virtual std::unique_ptr<CoreComponent> instantiate(const std::string &class_name, const std::string &name, const utils::Identifier &uuid, std::function<bool(CoreComponent*)> filter) = 0;
```
szaszm: do we need so many overloads? Since most of them just delegate to other overloads, maybe it could be done with one virtual implementation and maybe a few nonvirtual convenience wrappers

extensions/standard-processors/processors/DefragmentText.h
```diff
@@ -116,7 +113,6 @@ class DefragmentText : public core::ProcessorImpl {
   void onSchedule(core::ProcessContext& context, core::ProcessSessionFactory& session_factory) override;
   void onTrigger(core::ProcessContext& context, core::ProcessSession& session) override;
   void restore(const std::shared_ptr<core::FlowFile>& flowFile) override;
-  std::set<core::Connectable*> getOutGoingConnections(const std::string &relationship) override;
```
szaszm, line 119: why is this gone?

extensions/standard-processors/processors/RouteText.cpp
```diff
@@ -57,18 +54,10 @@ void RouteText::onSchedule(core::ProcessContext& context, core::ProcessSessionFa
...
-    std::vector<core::RelationshipDefinition> relationships(static_relationships.begin(), static_relationships.end());
-
-    for (const auto& property_name : getDynamicPropertyKeys()) {
-      core::RelationshipDefinition rel{property_name, "Dynamic Route"};
-      dynamic_relationships_[property_name] = rel;
-      relationships.push_back(rel);
-      logger_->log_info("RouteText registered dynamic route '{}'", property_name);
-    }
-
-    setSupportedRelationships(relationships);
```
szaszm, line 71: this is no longer called a second time with the dynamic relationships added. Is this on purpose? Why?

libminifi/include/core/FlowConfiguration.h
```diff
@@ -84,8 +84,7 @@ class FlowConfiguration : public CoreComponentImpl {
   ~FlowConfiguration() override;
 
   // Create Processor (Node/Input/Output Port) based on the name
-  std::unique_ptr<core::Processor> createProcessor(const std::string &name, const utils::Identifier &uuid);
-  std::unique_ptr<core::Processor> createProcessor(const std::string &name, const std::string &fullname, const utils::Identifier &uuid);
+  std::unique_ptr<core::Processor> createProcessor(const std::string &class_short, const std::string &fullclass, const std::string &name, const utils::Identifier &uuid);
```
szaszm: why do we need two class names?

libminifi/include/core/reporting/SiteToSiteProvenanceReportingTask.h
```diff
@@ -35,8 +35,8 @@ namespace org::apache::nifi::minifi::core::reporting {
 class SiteToSiteProvenanceReportingTask : public minifi::RemoteProcessorGroupPort {
  public:
   explicit SiteToSiteProvenanceReportingTask(std::shared_ptr<Configure> configure)
-      : minifi::RemoteProcessorGroupPort(ReportTaskName, "", std::move(configure)),
-        logger_(logging::LoggerFactory<SiteToSiteProvenanceReportingTask>::getLogger()) {
+      : minifi::RemoteProcessorGroupPort(ReportTaskName, "", std::move(configure),
+                                         utils::IdGenerator::getIdGenerator()->generate(), logging::LoggerFactory<SiteToSiteProvenanceReportingTask>::getLogger()) {
```
szaszm:
```suggestion
      : minifi::RemoteProcessorGroupPort(ReportTaskName, "", std::move(configure),
            utils::IdGenerator::getIdGenerator()->generate(), logging::LoggerFactory<SiteToSiteProvenanceReportingTask>::getLogger()) {
```

libminifi/src/core/FlowConfiguration.cpp
```diff
@@ -71,28 +71,17 @@ FlowConfiguration::~FlowConfiguration() {
...
     return nullptr;
   }
   return processor;
 }
 
 std::unique_ptr<core::reporting::SiteToSiteProvenanceReportingTask> FlowConfiguration::createProvenanceReportTask() {
-  auto processor = std::make_unique<org::apache::nifi::minifi::core::reporting::SiteToSiteProvenanceReportingTask>(this->configuration_);
-  processor->initialize();
-  return processor;
+  return std::make_unique<org::apache::nifi::minifi::core::reporting::SiteToSiteProvenanceReportingTask>(this->configuration_);
```
szaszm: is this initialize call safe to remove?

libminifi/src/core/flow/StructuredConnectionParser.cpp
```diff
@@ -41,8 +41,7 @@ void StructuredConnectionParser::addFunnelRelationshipToConnection(minifi::Conne
     return;
   }
 
-  auto& processor_ref = *processor;
-  if (typeid(minifi::Funnel) == typeid(processor_ref)) {
+  if (dynamic_cast<minifi::Funnel*>(&processor->getImpl())) {
```
szaszm: I'd prefer typeid over dynamic_cast if inheritance is not a concern.

libminifi/test/libtest/unit/TestBase.h
```diff
@@ -201,13 +201,56 @@ class TempDirectory {
...
+ public:
+  TypedProcessorWrapper() = default;
+  TypedProcessorWrapper(core::Processor* proc): proc_(proc) {  // NOLINT(runtime/explicit)
```
szaszm, line 209: Please add a comment about why this is okay to be an implicit conversion. Something like "it's a wrapper over the same object, with no value change".

libminifi/test/libtest/unit/TestBase.h
```diff
@@ -201,13 +201,56 @@ class TempDirectory {
...
+    if (proc_) {
+      proc_->getImpl<T>();
+    }
+  }
+
+  explicit operator bool() const {
+    return proc_ != nullptr;
+  }
+
+  operator core::Processor*() const {
```
szaszm, line 219: I'm not sure about this implicit conversion, we could achieve the same with a dereference + an arrow operator, and that would be explicit.

libminifi/test/libtest/unit/TestBase.h
```diff
@@ -201,13 +201,56 @@ class TempDirectory {
...
+
+  T& get() const {
+    gsl_Assert(proc_);
+    return proc_->getImpl<T>();
+  }
+
+  core::Processor* operator->() const {
+    gsl_Assert(proc_);
+    return proc_;
+  }
```
szaszm, line 232: Consider adding a matching dereference operator

libminifi/test/libtest/unit/TestUtils.h
```diff
@@ -210,18 +213,17 @@ inline std::error_code hide_file(const std::filesystem::path& file_name) {
...
 template<typename T>
-concept NetworkingProcessor = std::derived_from<T, minifi::core::Processor>
-    && requires(T x) {
-      {T::Port} -> std::convertible_to<core::PropertyReference>;
-      {x.getPort()} -> std::convertible_to<uint16_t>;
-    };  // NOLINT(readability/braces)
+concept NetworkingProcessor =  requires(T x) {
+  {T::Port} -> std::convertible_to<core::PropertyReference>;
+  {x.getPort()} -> std::convertible_to<uint16_t>;
+};  // NOLINT(readability/braces)
```
szaszm: Why is the subclass status no longer required? All processors need to implement the ProcessorApi at the moment

libminifi/test/unit/ProcessorConfigUtilsTests.cpp
```diff
@@ -46,8 +53,11 @@ TEST_CASE("Parse enum property") {
   static constexpr auto prop = PropertyDefinitionBuilder<magic_enum::enum_count<TestEnum>()>::createProperty("prop")
       .withAllowedValues(magic_enum::enum_names<TestEnum>())
       .build();
-  auto proc = std::make_shared<TestProcessor>("test-proc");
-  proc->setSupportedProperties(std::to_array<core::PropertyReference>({prop}));
+  auto proc = minifi::test::utils::make_processor<TestProcessor>("test-proc");
+  proc->getImpl<TestProcessor>().on_initialize_ = [&] (auto& self) {
+    self.setSupportedProperties(std::to_array<core::PropertyReference>({prop}));
+  };
+  proc->initialize();
```
szaszm, line 60: why was this changed?

core-framework/include/core/ProcessorImpl.h
```diff
@@ -0,0 +1,160 @@
...
+#include <condition_variable>
+#include <functional>
+#include <memory>
+#include <mutex>
+#include <string>
+#include <string_view>
+#include <unordered_set>
+#include <unordered_map>
+#include <utility>
+#include <vector>
```
szaszm: My IDE shows that half of these are unnecessary: algorithm, chrono, string_view, unordered_set, unordered_map, utility, vector. I also see missing includes, like array, span

extensions/standard-processors/processors/DefragmentText.h
adamdebreceni: previously it was used to inject the `Self` relationship, that is no longer needed, any processor can transfer to `__self__` and all the relationship/connection handling has been moved to `Processor`, the individual implementations no longer has to/can override the behavior

extensions/standard-processors/processors/RouteText.cpp
adamdebreceni: calling `setSupportedRelationships` is actually useless (other than for python processors) because neither `getSupportedRelationships` nor `isSupportedRelationship` were used for verification purposes, although they could be, but in `onSchedule` the verification has already completed, so there is nobody to signal this new updates relationship set

libminifi/include/core/FlowConfiguration.h
adamdebreceni: even though the processor's name was available we did not forward it to the constructor, instead the constructor was executed using the class's name as the processor instance's name, so the parameters called `name` and `fullname` were actually the class's short name and fully qualified name. As for why it needs both, there is a fallback mechanism if the short name fails, and I did not want to poke that part.

libminifi/src/core/FlowConfiguration.cpp
adamdebreceni: it was safe, there is a call to it on [StructuredConfiguration.cpp:637](https://github.com/adamdebreceni/nifi-minifi-cpp/blob/b3fc3279b8f527f7456bc193434688f39b0cea76/libminifi/src/core/flow/StructuredConfiguration.cpp#L637)
but I thought about it, and instead restored this one, and removed the linked one, and also did some wrapping into the `Processor` wrapper instead of the responsibility being scattered

libminifi/test/unit/ProcessorConfigUtilsTests.cpp
adamdebreceni: `Processor::setSupportedProperties` will become is kind of unnecessary going forward, and should only be called by the implementation during initialize (indirectly on self), a number of member should be made protected/private/removed on both Processor and ProcessorImpl in upcoming PRs

core-framework/include/core/ProcessorFactory.h
adamdebreceni: changed class_name_

libminifi/include/core/reporting/SiteToSiteProvenanceReportingTask.h
adamdebreceni: done

libminifi/src/core/flow/StructuredConnectionParser.cpp
adamdebreceni: done

libminifi/test/libtest/unit/TestBase.h
adamdebreceni: done

libminifi/test/libtest/unit/TestBase.h
adamdebreceni: the purpose of this class is to behave as a core::Processor* while also storing the underlying type so we don't have to specify it when accessing it, having explicit conversions or otherwise adding frictions defeats the purpose

libminifi/test/libtest/unit/TestUtils.h
adamdebreceni: readded it

core-framework/include/core/ProcessorImpl.h
adamdebreceni: done

minifi-api/include/minifi-cpp/core/ClassLoader.h
adamdebreceni: agreed, but we would need to touch the ObjectFactory interface as well, I would rather change this separately, (also remove the instantiateRaw)

libminifi/include/core/FlowConfiguration.h
szaszm: maybe we should rename name to object_name to differentiate it from the other two names

libminifi/test/libtest/unit/TestBase.h
adamdebreceni: added

libminifi/include/core/FlowConfiguration.h
adamdebreceni: other two names? I see `class_short`, `fullclass` and `name`

martinzink (approved):

libminifi/include/core/FlowConfiguration.h
szaszm: I don't quite understand the question, but I'll try to clarify my point.
I've assumed that:
- class_short: short type name of the class, e.g. LogAttribute
- fullclass: long type name (including namespaces) of the class, e.g. org::apache::nifi::minifi::processors::LogAttribute
- name: object name, i.e. if I create a LogAttribute in the flow with the name "Log contents", then this will be "Log contents". I'd rename this. to object_name

adamdebreceni: oh, I see, renamed

libminifi/src/core/Processor.cpp
```diff
@@ -43,16 +44,22 @@ using namespace std::literals::chrono_literals;
 
 namespace org::apache::nifi::minifi::core {
 
-ProcessorImpl::ProcessorImpl(std::string_view name, std::shared_ptr<ProcessorMetrics> metrics)
+constexpr std::chrono::microseconds MINIMUM_SCHEDULING_PERIOD{30};
```
szaszm, line 47: is this a new constant?

extensions/standard-processors/processors/PutTCP.h
```diff
@@ -25,11 +25,13 @@
 #include <utility>
 #include <vector>
 
+#include "io/InputStream.h"
+#include "core/ProcessorImpl.h"
+#include "utils/Export.h"
```
szaszm, line 30:
```suggestion
#include "core/ProcessorImpl.h"
```
the linter is complaining about duplicate includes

PR #1941, opened 2025-03-12
MINIFICPP-2448 Add minifi.sh flowStatus command

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2448

-----------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/src/core/BulletinStore.cpp
```diff
@@ -68,6 +68,17 @@ std::deque<Bulletin> BulletinStore::getBulletins(std::optional<std::chrono::syst
...
 
+std::vector<Bulletin> BulletinStore::getBulletinsForProcessor(const std::string& processor_uuid) const {
+  std::lock_guard<std::mutex> lock(mutex_);
+  std::vector<Bulletin> bulletins;
+  for (const auto& bulletin : bulletins_) {
+    if (bulletin.source_id == processor_uuid) {
+      bulletins.push_back(bulletin);
+    }
+  }
+  return bulletins;
```
szaszm:
```suggestion
  return bulletins_
      | std::ranges::filter([&](const auto& elem) { return elem.source_id == processor_uuid; })
      | ranges::to<std::vector>();
```

controller/Controller.cpp
```diff
@@ -267,4 +267,24 @@ nonstd::expected<void, std::string> getDebugBundle(const utils::net::SocketData&
   return {};
 }
 
+bool getFlowStatus(const utils::net::SocketData& socket_data, const std::string& status_query, std::ostream &out) {
+  std::unique_ptr<io::BaseStream> connection_stream = std::make_unique<utils::net::AsioSocketConnection>(socket_data);
```
szaszm: Why does this need to be cast to a base class pointer?
```suggestion
  const auto connection_stream = std::make_unique<utils::net::AsioSocketConnection>(socket_data);
```

bin/minifi.sh
```diff
@@ -183,6 +183,25 @@ status_service() {
...
+    fi
+    if [ "$#" -lt 2 ]; then
+        echo "MiNiFi flowStatus operation requires a flow status query parameter like \"processor:TailFile:health,stats,bulletins\""
+        return
+    fi
+
+    if [ "$#" -lt 3 ]; then
+        exec "${bin_dir}/minificontroller" --flowstatus "$2"
+    elif [ "$#" -lt 3 ]; then
+        exec "${bin_dir}/minificontroller" --port "$2" --flowstatus "$3"
```
fgerlits: typo:
```suggestion
    elif [ "$#" -lt 4 ]; then
        exec "${bin_dir}/minificontroller" --port "$2" --flowstatus "$3"
```

libminifi/include/c2/FlowStatusRequest.h
```diff
@@ -0,0 +1,60 @@
...
+  instance,
+  systemdiagnostics
+};
+
+struct FlowStatusRequest {
+  FlowStatusQueryType query_type;
+  std::string identifier;
+  std::unordered_set<std::string> options;
+
+  explicit FlowStatusRequest(const std::string& query_string) {
```
fgerlits: I think this function is too long to be in the header; I would move it to a .cpp file.

libminifi/include/c2/FlowStatusRequest.h
```diff
@@ -0,0 +1,60 @@
...
+    }
+    auto query_type_result = magic_enum::enum_cast<FlowStatusQueryType>(query_parameters[0]);
+    if (!query_type_result) {
+      throw std::invalid_argument("Invalid query type: " + query_parameters[0]);
+    }
+    query_type = *query_type_result;
+    if (query_parameters.size() > 2) {
+      identifier = query_parameters[1];
+      auto option_vector = minifi::utils::string::splitAndTrimRemovingEmpty(query_parameters[2], ",");
+      options = std::unordered_set<std::string>(option_vector.begin(), option_vector.end());
```
fgerlits: Not super-important, but it would seem more natural to parse the options into enum values, as well. That would help with syntax checking, too. Currently, if I mistype "bulletins" to "bulletin" (for example), it will be silently accepted and ignored.

lordgamez: Good point, although in this case I would keep it this way just to be in line with the java implementation of the flowstatus. The other thing I noticed is that if there is a component name that is missing an option it would get an "invalid option name" message instead of an "incomplete query" message with implementing this change, which may be a bit misleading.

bin/minifi.sh
lordgamez: Good catch, updated in https://github.com/apache/nifi-minifi-cpp/pull/1941/commits/92d011455e789d896dd0005b533d57d50f2223d1

libminifi/include/c2/FlowStatusRequest.h
lordgamez: Good point, updated in https://github.com/apache/nifi-minifi-cpp/pull/1941/commits/92d011455e789d896dd0005b533d57d50f2223d1

libminifi/include/c2/FlowStatusRequest.h
lordgamez: After discussing this I updated the PR with validating the query options in https://github.com/apache/nifi-minifi-cpp/pull/1941/commits/5a69ef1b468bb5dec349bfef57d979cb7f7820c0

martinzink (approved):

libminifi/include/c2/FlowStatusRequest.h
```diff
@@ -0,0 +1,48 @@
...
+  flowfilerepositoryusage,
+  contentrepositoryusage
+};
+
+struct FlowStatusRequest {
+  FlowStatusQueryType query_type;
+  std::string identifier;
+  std::unordered_set<FlowStatusQueryOption> options;
+
+  explicit FlowStatusRequest(const std::string& query_string);
```
szaszm: A long comment explaining the format would be greatly appreciated. Also, consider taking a string_view.

libminifi/src/c2/ControllerSocketProtocol.cpp
```diff
@@ -187,6 +187,12 @@ void ControllerSocketProtocol::initialize() {
   }
 }
 
+void ControllerSocketProtocol::setRoot(core::ProcessGroup* root) {
+  if (auto controller_socket_reporter = controller_socket_reporter_.lock()) {
+    controller_socket_reporter->setRoot(root);
+  }
+}
+
```
szaszm, line 196: What happens to the pointed-to root process group on a flow update? Does the FlowController call setRoot with the new root process group before freeing the old root?

libminifi/test/unit/FlowStatusBuilderTests.cpp
szaszm, line 1: Is flowStatus thread-safe? Can we spam hundreds of requests in parallel and check that each one gets the proper result?
Is there no data race between writers of metrics / bulletins / etc and flowStatus readers? Are the reads / writes synchronized?

libminifi/include/c2/FlowStatusBuilder.h
szaszm, line 1: I think this whole class should be a function called `buildFlowStatus`, with parameters instead of data members.

libminifi/src/c2/ControllerSocketProtocol.cpp
lordgamez: In the RootProcessGroupWrapper::setNewRoot we create a backup of the old root before calling the setRoot with the new root process group. If the flow update succeeds we clear the old root, otherwise we restore it.

libminifi/test/unit/FlowStatusBuilderTests.cpp
lordgamez: I checked running hundreds of processes in parallel and it worked successfully. There should be no data race as the data is collected from the connection_map_, processors_ and connections_ that are synchronized under the root_mutex_ when they are read or updated.

libminifi/include/c2/FlowStatusBuilder.h
lordgamez: I'm not sure it's a good idea, it's easier to encapsulate these members in a class and cache them and handle the flow update with the changed root if we have a single object that handles all of this.

libminifi/include/c2/FlowStatusRequest.h
lordgamez: Added in https://github.com/apache/nifi-minifi-cpp/pull/1941/commits/f50560c454cd00d67e4fd5df37780ae2e824a62a

controller/Controller.cpp
lordgamez: Due to the virtual inheritance hierarchy the compiler was not able the find this specific overload from the AsioSockerConnection class, I added a using directive there to be able to find the read and write overloads with the std::span parameter and updated the return value to auto in https://github.com/apache/nifi-minifi-cpp/pull/1941/commits/f50560c454cd00d67e4fd5df37780ae2e824a62a

libminifi/src/core/BulletinStore.cpp
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1941/commits/f50560c454cd00d67e4fd5df37780ae2e824a62a

libminifi/include/c2/FlowStatusRequest.h
lordgamez: Changed parameter to string_view in https://github.com/apache/nifi-minifi-cpp/pull/1941/commits/d99583f402ee621145e4fd3518c2b2a0cc6a9779

libminifi/test/unit/FlowStatusBuilderTests.cpp
szaszm: would it be feasible to have something like that in a unit test?

lordgamez: I'm not sure if it's feasible on a unit test level, or if it would be worth checking this with random requests without any suspicion of a concurrency issue. As this is designed to be used through the CLI I think manual testing with hundreds of CLI requests should be enough.

szaszm (approved):

PR #1945, opened 2025-03-19
MINIFICPP-2541 - Simplify VolatileContentRepository

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/include/core/repository/VolatileContentRepository.h
```diff
@@ -17,113 +17,35 @@
...
-#include "AtomicRepoEntries.h"
-#include "io/AtomicEntryStream.h"
 #include "core/ContentRepository.h"
-#include "properties/Configure.h"
-#include "core/Connectable.h"
-#include "core/logging/LoggerFactory.h"
-#include "utils/GeneralUtils.h"
-#include "VolatileRepositoryData.h"
-#include "utils/Literals.h"
+#include <map>
```
martinzink, line 21: This caused the linter error
```suggestion
#include <unordered_map>
#include <string>
#include <memory>
#include <mutex>

#include "core/ContentRepository.h"
```

libminifi/src/core/repository/VolatileContentRepository.cpp
```diff
@@ -17,142 +16,133 @@
...
+      : data_(std::move(data)), data_store_mtx_(data_store_mtx), data_store_(data_store), total_size_(total_size) {}
 
-namespace org::apache::nifi::minifi::core::repository {
+  [[nodiscard]] size_t size() const override {
+    return data_->size();
+  }
+
+  size_t read(std::span<std::byte> out_buffer) override {
+    auto read_size = std::min(data_->size() - read_offset_, out_buffer.size());
+    std::copy_n(reinterpret_cast<const std::byte*>(data_->data()) + read_offset_, read_size, out_buffer.data());
```
szaszm, line 36: I think it would be better to use span operations instead of pointer arithmetic.
```suggestion
    const auto source_span = as_bytes(std::span{*data_}.subspan(read_offset_, read_size));
    std::copy_n(source_span.data(), source_span.size(), out_buffer.data());
```

libminifi/test/unit/performance/VolatileRepositoryPerfTests.cpp
szaszm, line 1: Could you add a section to the README or somewhere else about how to run these tests and get some numbers?

libminifi/include/core/repository/VolatileContentRepository.h
```diff
@@ -17,113 +17,38 @@
...
-  VolatileRepositoryData repo_data_;
-  bool minimize_locking_;
-
-  // mutex and master list that represent a cache of Atomic entries. this exists so that we don't have to walk the atomic entry list.
-  // The idea is to reduce the computational complexity while keeping access as maximally lock free as we can.
-  std::mutex map_mutex_;
-  std::map<ResourceClaim::Path, AtomicEntry<ResourceClaim::Path>*> master_list_;
+  mutable std::mutex data_mtx_;
+  std::unordered_map<std::string, std::shared_ptr<std::string>> data_;
+  std::atomic<size_t> total_size_;
```
szaszm, line 51: The copilot low confidence suggestion looks valid IMO. Before C++20, the underlying value used to be uninitialized. After that, it's default-initialized, which in the case of integer types means uninitialized. It doesn't seem to be initialized later either. Let's add a zero init.

https://en.cppreference.com/w/cpp/atomic/atomic/atomic

```suggestion
  std::atomic<size_t> total_size_{0};
```

libminifi/src/core/repository/VolatileContentRepository.cpp
```diff
@@ -17,142 +16,133 @@
...
-      delete claim_item->second;  // NOLINT(cppcoreguidelines-owning-memory)
-      master_list_.erase(content_path);
-      repo_data_.current_size -= size;
-    }
+  std::lock_guard lock(data_mtx_);
+  if (auto it = data_.find(content_path); it != data_.end()) {
+    total_size_ -= it->second->size();
+    data_.erase(it);
+    logger_->log_info("Deleting resource {}", content_path);
   }
```
lordgamez, line 144: We could add some log if the key is not found

libminifi/test/unit/performance/VolatileRepositoryPerfTests.cpp
adamdebreceni: added

cmake/FetchBenchmark.cmake
```diff
@@ -0,0 +1,27 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+include(FetchContent)
+
+FetchContent_Declare(
+        benchmark
```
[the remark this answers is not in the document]
adamdebreceni: added

libminifi/include/core/repository/LegacyVolatileContentRepository.h
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  bool isFull() const override {
+    return repo_data_.isFull();
+  }
+
+  /**
+   * Initialize the volatile content repo
+   * @param configure configuration
+   */
```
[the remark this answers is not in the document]
adamdebreceni: removed

libminifi/src/core/repository/VolatileContentRepository.cpp
adamdebreceni: added

.github/workflows/ci.yml
szaszm, line 1: I think we should run the perf tests separately with `--verbose`, so we can see the numbers in the build log. They can go before or after the normal unit tests. Since they take about no time, we don't even need to disable them in the normal test run, just also run them separately with output enabled.

cmake/MiNiFiOptions.cmake
```diff
@@ -44,6 +44,7 @@ endfunction()
 
 add_minifi_option(CI_BUILD "Build is used for CI." OFF)
 add_minifi_option(SKIP_TESTS "Skips building all tests." OFF)
+add_minifi_option(MINIFI_PERFORMANCE_TESTS "Build performance tests" OFF)
```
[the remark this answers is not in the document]
lordgamez: I think we can keep them optional as they are better to be on-demand when we want to measure any performance. Even if it would take no time to run the average user would not have any benefit from running them together with the other tests as they cannot really fail, but later when additional performance tests are added it might take a long time to run them.

Although on my machine it took a lot of time when running them together with the other tests:
```
200/204 Test #185: HTTPSiteToSiteTests ........................   Passed   10.63 sec
201/204 Test   #3: ListenHTTPTests ............................   Passed   11.73 sec
202/204 Test #200: PersistenceTests ...........................   Passed   16.62 sec
203/204 Test  #70: VerifyInvokeHTTPPostTest ...................   Passed   18.22 sec
204/204 Test #156: VolatileRepositoryPerfTests ................   Passed   95.64 sec
```

cmake/FetchBenchmark.cmake
lordgamez, line 1: When I run the configuration with the performance tests enabled locally it fails with the following error:
```
CMake Error at CMakeLists.txt:34 (message):
  Did not find Google Test sources! Either pass correct path in
  GOOGLETEST_PATH, or enable BENCHMARK_DOWNLOAD_DEPENDENCIES, or disable
  BENCHMARK_USE_BUNDLED_GTEST, or disable BENCHMARK_ENABLE_GTEST_TESTS /
  BENCHMARK_ENABLE_TESTING.


-- Configuring incomplete, errors occurred!
CMake Error at build/_deps/benchmark-src/cmake/GoogleTest.cmake:13 (message):
  CMake step for googletest failed: 1
Call Stack (most recent call first):
  build/_deps/benchmark-src/CMakeLists.txt:345 (include)


-- Configuring incomplete, errors occurred!
```

I suppose this is caused if the googletest target is created after the benchmark is configured. This can be fixed by adding the following, and this should be okay as we do not need to build the tests of the benchmark library anyway:

```
set(BENCHMARK_ENABLE_TESTING "OFF" CACHE STRING "" FORCE)
```

adamdebreceni: added

.github/workflows/ci.yml
adamdebreceni: added the perf tests to one of the builds, as we only care about relative performance for now

lordgamez (approved):

PR #1956, opened 2025-03-28
MINIFICPP-2552 schema generation: simplify usage, drive-by changes

szaszm:
- add line ending after the last JSON line
- exit through return from main instead of `std::exit`, so destructors are called normally
- removed checks for parent directory and regular file: writing to special files or nonexistent files is totally fine, and creating the parent directory is not the responsibility of minifi.
- error handling of std::ofstream
- `--schema` now has a default argument of `-`, meaning output to stdout
- no longer chdir to MINIFI_HOME on startup (I'm not sure about this one, it may cause issues elsewhere)

----

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/minifi_main/MiNiFiMain.cpp
+++ b/minifi_main/MiNiFiMain.cpp
@@ -130,7 +130,7 @@ void writeJsonSchema(const std::shared_ptr<minifi::Configure>& configuration, st
     pythoncreator->configure(configuration);
   }
 
-  out << minifi::docs::generateJsonSchema();
+  out << minifi::docs::generateJsonSchema() << '\n';
 }
 
 void overridePropertiesFromCommandLine(const argparse::ArgumentParser& parser, const std::shared_ptr<minifi::Configure>& configure) {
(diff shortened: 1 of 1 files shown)


minifi_main/MiNiFiMain.cpp
```diff
@@ -196,6 +196,7 @@ int main(int argc, char **argv) {
     .help("Generate documentation in the specified directory");
   argument_parser.add_argument("-s", "--schema")
     .metavar("PATH")
+    .default_value("-")
```
lordgamez, line 199: Maybe it would be better to change this argument to allow 0 or 1 values and if no value is provided then we write the schema to stdout.

adamdebreceni (approved):

minifi_main/MiNiFiMain.cpp
szaszm: That was my first idea too. I don't know if it's better or not, but I couldn't figure out how to do it, that's why I went with the default value approach instead. The `-` value for filename arguments is commonly used to mean stdout anyway.

lordgamez: This patch seems to be working for me:
```
diff --git a/minifi_main/MiNiFiMain.cpp b/minifi_main/MiNiFiMain.cpp
index 61d49a7bc..733b1194d 100644
--- a/minifi_main/MiNiFiMain.cpp
+++ b/minifi_main/MiNiFiMain.cpp
@@ -165,12 +165,14 @@ void overridePropertiesFromCommandLine(const argparse::ArgumentParser& parser, c
   if (!parser.is_used("--schema")) {
     return std::nullopt;
   }
-  const auto& schema_path = parser.get("--schema");
-  if (schema_path == "-") {
+  const auto& schema_path_args = parser.get<std::vector<std::string>>("--schema");
+  if (schema_path_args.empty()) {
     writeJsonSchema(configure, std::cout);
     return 0;
   }
 
+  const auto& schema_path = schema_path_args[0];
+
   std::cout << "Writing json schema to " << schema_path << std::endl;
   {
     std::ofstream schema_file{schema_path};
@@ -196,8 +198,8 @@ int main(int argc, char **argv) {
     .help("Generate documentation in the specified directory");
   argument_parser.add_argument("-s", "--schema")
     .metavar("PATH")
-    .default_value("-")
-    .help("Generate JSON schema to the specified path");
+    .nargs(0, 1)
+    .help("Generate JSON schema to the specified path. If no path is specified, the schema will be printed to stdout");
 
   try {
     argument_parser.parse_args(argc, argv);
```

szaszm: I want to keep the `-` meaning `stdout`, like most command line tools. Would you still prefer a separate empty check in that case too?

lordgamez: We can keep both approaches, but if you would only stick with the `-` approach I'm okay with that too. It was just more intuitive to me not give any parameters if `stdout` is used. You can decide I'm okay with either option.

lordgamez (approved):

minifi_main/MiNiFiMain.cpp
szaszm: with the current version, since the default value is `-`, `stdout` is used when no parameters are given, as well as when `-` is given as parameter.
So both of these print to the stdout:
- `./minifi --schema`
- `./minifi --schema -`

szaszm:
Response to Copilot low confidence comments:
1. I ensured that the callers only exit when the optional exit code is set (not nullopt).
2. I didn't read the whole codebase looking for edge cases, but I believe that this _probably_ won't cause an issue, also relying on our automated test coverage.

PR #1966, opened 2025-05-19
MINIFICPP-2558 Revamp site to site communication

lordgamez:
- Fix communicating with remote output port
- Add handling of remote port batch size properties
- Extend port management in config generation in docker integration tests
- Make input port in remote process groups optional (as using only output ports can also be a valid configuration)
- Add additional testing
- Refactor RPG and site to site clients
- Additional documentation

https://issues.apache.org/jira/browse/MINIFICPP-2565

-------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


libminifi/include/RemoteProcessGroupPort.h
```diff
@@ -139,14 +134,21 @@ class RemoteProcessorGroupPort : public core::ProcessorImpl {
...
-      this->setTriggerWhenEmpty(true);
+    if (direction_ == sitetosite::TransferDirection::RECEIVE) {
+      setTriggerWhenEmpty(true);
+    } else {
+      setTriggerWhenEmpty(false);
+    }
   }
 
-  void setTimeout(uint64_t timeout) {
+  void setTimeout(std::chrono::milliseconds timeout) {
```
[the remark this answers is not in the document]
szaszm: it's part of the type, and larger units are automatically converted, so I don't think it's necessary to repeat this info in a comment

SITE_TO_SITE.md
```diff
@@ -0,0 +1,262 @@
...
+        batch size:
+          size: 10 MB
+          count: 10
+          duration: 30 sec
+    Output Ports: []
+```
+
+Here is another example in NiFi style json format how to configure site-to-site protocol in MiNiFi C++ where the MiNiFi C++ instance is receiving data from NiFi using the HTTP protocol:
+
+```json
```
fgerlits: Here in the readme, I think it would be better to have YAML configurations for both cases, because YAML is more readable, and it would be easier to compare the two cases (input vs output port).

I think it's enough to include JSON versions of these configurations in the examples directory, and mention them in this readme (as you already do).

SITE_TO_SITE.md
```diff
@@ -0,0 +1,262 @@
...
+    }
+}
+```
+
+Notes on the configuration:
+
+- In the MiNiFi C++ configuration, in yaml configuration the remote input and output ports' `id` field, and in json configuration the ports' `identifier`, `instanceIdentifier`, and `targetId` fields should be set to the instance id of the input and output ports created in NiFi (`de7cc09a-0196-1000-2c63-ee6b4319ffb6` in the examples).
+- Connections from the remote output port to the processor should use the `undefined` relationship
+- `useCompression` can be set, but it is currently not supported in MiNiFi C++ so it will be set to false in the site-to-site messages
+- the `targetUri` and `targetUris` field in the remote process group should be set to the NiFi instance's URL, this can also use comma separated list of URLs if the remote process group is configured to use multiple NiFi nodes
```
fgerlits:
```suggestion
- the `url` field (`targetUri` or `targetUris` in JSON) field in the remote process group should be set to the NiFi instance's URL, this can also use comma separated list of URLs if the remote process group is configured to use multiple NiFi nodes
```

SITE_TO_SITE.md
```diff
@@ -0,0 +1,262 @@
...
+            {
+                "identifier": "327b446a-0043-48d1-8bb4-df65ba1afa60",
+                "instanceIdentifier": "2ed47dca-38f5-476d-9c37-5ea0a5072f1e",
+                "name": "https://localhost:8443/nifi",
+                "position": {
+                    "x": 235.0,
+                    "y": 71.00000762939453
+                },
+                "targetUri": "https://localhost:8443/nifi",
+                "targetUris": "https://localhost:8443/nifi",
```
fgerlits: We shouldn't have both `targetUri` and `targetUris` in the example, since minifi will only use the one it finds first in the config.

docker/test/integration/features/s2s.feature
```diff
@@ -25,73 +25,88 @@ Feature: Sending data from MiNiFi-C++ to NiFi using S2S protocol
...
     Given a GetFile processor with the "Input Directory" property set to "/tmp/input"
     And a file with the content "test" is present in "/tmp/input"
-    And a RemoteProcessGroup node opened on "http://nifi-${feature_id}:8080/nifi"
-    And the "success" relationship of the GetFile processor is connected to the input port on the RemoteProcessGroup
+    And a RemoteProcessGroup node with name "RemoteProcessGroup" is opened on "http://nifi-${feature_id}:8080/nifi"
+    And an input port with name "to_nifi" is created on the RemoteProcessGroup named "RemoteProcessGroup"
+    And the "success" relationship of the GetFile processor is connected to the to_nifi
 
-    And a NiFi flow receiving data from a RemoteProcessGroup "from-minifi"
+    And a NiFi flow is receiving data in an input port named "from-minifi" with the id of the port named "to_nifi" from the RemoteProcessGroup named "RemoteProcessGroup"
```
fgerlits: I understand this is already quite long, but I think
```suggestion
    And a NiFi flow is receiving data from the RemoteProcessGroup named "RemoteProcessGroup" in an input port named "from-minifi" which has the same id as the port named "to_nifi"
```
would be easier to understand, even though it is even longer.

docker/test/integration/minifi/flow_serialization/Minifi_flow_json_serializer.py
```diff
@@ -134,8 +179,8 @@ def serialize_node(self, connectable, root, visited):
             for proc in conn_procs:
                 root['connections'].append({
                     'name': str(uuid.uuid4()),
-                    'source': {'id': str(connectable.uuid)},
-                    'destination': {'id': str(proc.uuid) if not isinstance(proc, InputPort) else str(proc.instance_id)}
+                    'source': {'id': str(connectable.uuid) if not isinstance(connectable, InputPort) and not isinstance(connectable, OutputPort) else str(connectable.instance_id)},
+                    'destination': {'id': str(proc.uuid) if not isinstance(proc, InputPort) and not isinstance(proc, OutputPort) else str(proc.instance_id)}
```
fgerlits: this is getting too complex; it could be simplified e.g. by adding an `id_for_connection` method in `Connectable`, and overriding it in `InputPort` and `OutputPort`

extensions/standard-processors/tests/unit/FlowJsonTests.cpp
```diff
@@ -1786,4 +1808,69 @@ TEST_CASE("Parameter context names cannot conflict with parameter provider gener
...
+        "id": "00000000-0000-0000-0000-000000000001",
+        "name": "PutFile"
+      },
+      "selectedRelationships": [""]
+    }],
+    "remoteProcessGroups": [{
+      "name": "NiFi Flow",
+      "targetUri": "https://localhost:8090/nifi",
+      "communicationsTimeout": "19 sec",
+      "inputPorts": [{
```
fgerlits: this should be `outputPorts`

libminifi/include/sitetosite/Peer.h
```diff
@@ -17,88 +17,29 @@
...
-  uint16_t port_;
-
-  utils::Identifier port_id_;
-
-  // secore comms
+namespace org::apache::nifi::minifi::sitetosite {
 
-  bool secure_;
-};
+static constexpr std::array<char, 4> MAGIC_BYTES = { 'N', 'i', 'F', 'i' };
```
fgerlits: in a header, on namespace level, this should be
```suggestion
inline constexpr std::array<char, 4> MAGIC_BYTES = { 'N', 'i', 'F', 'i' };
```

libminifi/include/sitetosite/Peer.h
```diff
@@ -109,175 +50,109 @@ class PeerStatus {
...
         port_(port),
-        timeout_(std::chrono::seconds(30)),
-        logger_(core::logging::LoggerFactory<SiteToSitePeer>::getLogger()) {
-    url_ = "nifi://" + host_ + ":" + std::to_string(port_);
-    yield_expiration_ = std::chrono::system_clock::time_point();
-    timeout_ = std::chrono::seconds(30);
-    local_network_interface_ = io::NetworkInterface(ifc, nullptr);
+        url_("nifi://" + host_ + ":" + std::to_string(port_)),
+        local_network_interface_(io::NetworkInterface(ifc, nullptr)) {
+    timeout_.store(30s);
```
fgerlits: `timeout_` is already initialized below, by the default member initializer

libminifi/src/sitetosite/RawSiteToSiteClient.cpp
```diff
@@ -0,0 +1,431 @@
+/**
+ * Site2SiteProtocol class implementation
+ *
```
fgerlits: this should be updated or deleted

libminifi/include/sitetosite/SiteToSiteClient.h
```diff
@@ -19,246 +19,182 @@
...
-  /***
-   * versioning
-   */
-  uint32_t _supportedVersion[5] = {5, 4, 3, 2, 1};
-  int _currentVersionIndex{0};
-  uint32_t _currentVersion{_supportedVersion[_currentVersionIndex]};
-  uint32_t _supportedCodecVersion[1] = {1};
-  int _currentCodecVersionIndex{0};
-  uint32_t _currentCodecVersion{_supportedCodecVersion[_currentCodecVersionIndex]};
+  const std::vector<uint32_t> supported_version_ = {5, 4, 3, 2, 1};
```
fgerlits: nitpicking, but `supported_versions_` and `supported_codec_versions_` would be better names

libminifi/include/sitetosite/SiteToSite.h
```diff
@@ -23,35 +23,28 @@
...
+  RESOURCE_OK = 20,
+  DIFFERENT_RESOURCE_VERSION = 21,
+  NEGOTIATED_ABORT = 255
+};
+
+static constexpr uint32_t MAX_NUM_ATTRIBUTES = 25000;
+
+// Response Code Sequence Pattern
+static constexpr uint8_t CODE_SEQUENCE_VALUE_1 = static_cast<uint8_t>('R');
+static constexpr uint8_t CODE_SEQUENCE_VALUE_2 = static_cast<uint8_t>('C');
```
fgerlits: these should be `inline constexpr`, too

libminifi/include/sitetosite/SiteToSiteClient.h
```diff
@@ -19,246 +19,182 @@
...
+        transaction{std::move(transaction)},
+        payload{payload} {
   }
-  std::map<std::string, std::string> _attributes;
-  uint64_t _size{0};
-  std::shared_ptr<Transaction> transaction_;
-  const std::string & payload_;
-  std::shared_ptr<core::logging::Logger> logger_reference_;
+  std::map<std::string, std::string> attributes;
+  uint64_t size{0};
```
fgerlits: is this `size` field used anywhere?

libminifi/include/sitetosite/SiteToSiteClient.h
```diff
@@ -19,246 +19,182 @@
...
 };
 
-class SiteToSiteClient : public core::ConnectableImpl {
+class SiteToSiteClient {
  public:
-  SiteToSiteClient()
-      : core::ConnectableImpl("SitetoSiteClient") {
+  explicit SiteToSiteClient(std::unique_ptr<SiteToSitePeer> peer)
+      : peer_(std::move(peer)) {
+    gsl_Assert(peer_);
```
fgerlits: Can we change the type to `gsl::not_null` (and get rid of the assert)?

libminifi/include/sitetosite/SiteToSiteClient.h
```diff
@@ -19,246 +19,182 @@
...
+  bool confirm(const utils::Identifier &transaction_id);
 
-  // idleTimeout
-  std::chrono::milliseconds idle_timeout_{15000};
+  void handleTransactionError(const std::shared_ptr<Transaction>& transaction, core::ProcessContext& context, const std::exception& exception);
 
-  // Peer Connection
+  PeerState peer_state_{PeerState::IDLE};
+  utils::Identifier port_id_;
+  std::chrono::milliseconds idle_timeout_{15000};
```
fgerlits: we could write
```suggestion
  std::chrono::milliseconds idle_timeout_{15s};
```
since we are already including chrono literals

libminifi/include/sitetosite/SiteToSite.h
```diff
@@ -205,140 +147,184 @@ typedef enum {
...
- public:
-  MINIFIAPI static const char *RequestTypeStr[MAX_REQUEST_TYPE];
-  MINIFIAPI static RespondCodeContext respondCodeContext[21];
+struct ResponseCodeContext {
+  ResponseCode code = ResponseCode::UNRECOGNIZED_RESPONSE_CODE;
+  const std::string_view description;
+  bool has_description = false;
 };
 
+static constexpr std::array<ResponseCodeContext, 21> respond_code_contexts = {{
```
fgerlits: `response_code_contexts` would be a better name

(also this should be `inline constexpr`)

libminifi/src/sitetosite/SiteToSiteFactory.cpp
```diff
@@ -0,0 +1,83 @@
...
+}  // namespace
+
+std::unique_ptr<SiteToSiteClient> createClient(const SiteToSiteClientConfiguration &client_configuration) {
+  switch (client_configuration.getClientType()) {
+    case ClientType::RAW:
+      return createRawSocketSiteToSiteClient(client_configuration);
+    case ClientType::HTTP:
+      return createHttpSiteToSiteClient(client_configuration);
+  }
+  return nullptr;
```
fgerlits: It would be better to terminate here; could this be a `gsl_Assert(false)` or something similar?

libminifi/test/libtest/integration/HTTPHandlers.cpp
```diff
@@ -93,6 +95,25 @@ TransactionResponder::TransactionResponder(std::string base_url, std::string por
 }
 
 bool TransactionResponder::handlePost(CivetServer* /*server*/, struct mg_connection *conn) {
+  auto req_info = mg_get_request_info(conn);
+  std::unordered_map<std::string, std::string> expected_headers {
```
fgerlits: This could be massaged into a `static constexpr`, but I'm not sure we want to bother in a test class. But at least it should be `static const`.

libminifi/test/libtest/unit/DummyProcessor.h
```diff
@@ -41,7 +41,8 @@ class DummyProcessor : public minifi::core::ProcessorImpl {
       .isSensitive(true)
       .build();
   static constexpr auto Properties = std::array<core::PropertyReference, 2>{SimpleProperty, SensitiveProperty};
-  static constexpr auto Relationships = std::array<core::RelationshipDefinition, 0>{};
+  static constexpr core::RelationshipDefinition Success{"success", "Success relationship"};
+  static constexpr auto Relationships = std::array<core::RelationshipDefinition, 1>{Success};
```
fgerlits: no type conversion is needed here, so
```suggestion
  static constexpr auto Relationships = std::array{Success};
```
would work

libminifi/include/RemoteProcessGroupPort.h
```diff
@@ -155,102 +157,90 @@ class RemoteProcessorGroupPort : public core::ProcessorImpl {
...
-
-  sitetosite::CLIENT_TYPE client_type_;
-
-  // Remote Site2Site Info
-  bool site2site_secure_;
+  sitetosite::ClientType client_type_;
   std::vector<sitetosite::PeerStatus> peers_;
-  std::atomic<int> peer_index_;
+  std::atomic<int64_t> peer_index_;
   std::mutex peer_mutex_;
```
fgerlits: Do we need both `atomic` and `peer_mutex`?

libminifi/src/RemoteProcessGroupPort.cpp
```diff
@@ -0,0 +1,379 @@
...
+    logger_->log_error("No peers selected during scheduling");
+  }
+}
+
+void RemoteProcessGroupPort::notifyStop() {
+  transmitting_ = false;
+  RPGLatch count(false);  // we're just a monitor
+  // we use the latch
+  while (count.getCount() > 0) {
+  }
```
fgerlits: This doesn't seem to do anything. First, we create the `RPGLatch` object with a `count` member which is a pointer to a static atomic integer which is initialized to 0 and its value is never changed. Next, we check if `*count > 0` (it isn't), and finally we continue on our way.

It used to do something, because there was an `RPGLatch count;` line in `RemoteProcessorGroupPort::onTrigger()`. This was supposed to increment the atomic integer, but it was broken because the constructor and destructor incremented and decremented the pointer instead of the atomic integer it points to.

I think we should fix the `RPGLatch` class (or replace it with something from the standard library if there is a replacement) and put the counter increment back into `onTrigger()`.  Or if it is no longer needed, then we should remove the `RPGLatch` class.

libminifi/src/sitetosite/HttpSiteToSiteClient.cpp
```diff
@@ -0,0 +1,415 @@
...
+  return utils::Identifier::parse(utils::string::partAfterLastOccurrenceOf(uri, '/'));
+}
+
+std::optional<std::vector<PeerStatus>> parsePeerStatuses(const std::shared_ptr<core::logging::Logger> &logger, const std::string &entity, const utils::Identifier &id) {
+  try {
+    rapidjson::Document root;
+    rapidjson::ParseResult ok = root.Parse(entity.c_str());
+    if (!ok) {
+      std::stringstream ss;
+      ss << "Failed to parse archive lens stack from JSON string with reason: "
```
fgerlits: "archive lens stack"? -- seems to be a (very old) copy-paste error

libminifi/src/sitetosite/HttpSiteToSiteClient.cpp
```diff
@@ -0,0 +1,415 @@
...
+          << rapidjson::GetParseError_En(ok.Code())
+          << " at offset " << ok.Offset();
+
+      throw Exception(ExceptionType::GENERAL_EXCEPTION, ss.str());
+    }
+
+    std::vector<PeerStatus> peer_statuses;
+    if (!root.HasMember("peers") || !root["peers"].IsArray() || root["peers"].Size() <= 0) {
+      logger->log_debug("Peers is either not a member or is empty. String to analyze: {}", entity);
+      return peer_statuses;
```
fgerlits, line 58: Is it intentional that we return an empty vector here? In other error cases, we return a `nullopt`.

libminifi/src/sitetosite/HttpSiteToSiteClient.cpp
```diff
@@ -0,0 +1,415 @@
...
+  }
+
+  if (transaction->getState() == TransactionState::TRANSACTION_STARTED || transaction->getState() == TransactionState::DATA_EXCHANGED) {
+    if (current_code_ == ResponseCode::CONFIRM_TRANSACTION && transaction->getState() == TransactionState::DATA_EXCHANGED) {
+      auto stream = dynamic_cast<http::HttpStream*>(peer_->getStream());
+      if (!stream->isFinished()) {
+        logger_->log_debug("confirm read for {}, but not finished ", transaction->getUUIDStr());
+        if (stream->waitForDataAvailable()) {
+          response.code = ResponseCode::CONTINUE_TRANSACTION;
+          return response;
```
fgerlits: This style:
```suggestion
          return SiteToSiteResponse{.code = ResponseCode::CONTINUE_TRANSACTION};
```
would be more readable for me. As it is, with a `response` object created at the start and returned at various points, it's not clear whether the object has been modified earlier in the function.

libminifi/src/sitetosite/RawSiteToSiteClient.cpp
```diff
@@ -0,0 +1,431 @@
...
+    }
+
+    uint32_t count = 0;
+    if (const auto ret = peer_->read(count); ret == 0 || io::isError(ret)) {
+      tearDown();
+      return std::nullopt;
+    }
+
+    peers.push_back(PeerStatus(port_id_, host, gsl::narrow<uint16_t>(port), count, true));
+    logger_->log_trace("Site2Site Peer host {} port {} Secure {}", host, port, std::to_string(secure));
```
fgerlits, line 279: `secure` is no longer used in `PeerStatus` (only in the log); I can't see it added back in #1974, either -- is this OK?

libminifi/test/unit/SiteToSiteTests.cpp
```diff
@@ -0,0 +1,343 @@
...
+
+  std::string uuid_str = "C56A4180-65AA-42EC-A945-5FD21DEC0538";
+  minifi::utils::Identifier fake_uuid;
+  fake_uuid = uuid_str;
+  protocol.setPortId(fake_uuid);
+
+  REQUIRE_FALSE(SiteToSiteClientTestAccessor::bootstrap(protocol));
+}
+
+void iniitalizeMockRemoteClientReceiveDataResponses(SiteToSiteResponder& collector) {
```
fgerlits: typo:
```suggestion
void initializeMockRemoteClientReceiveDataResponses(SiteToSiteResponder& collector) {
```

libminifi/test/unit/SiteToSiteTests.cpp
```diff
@@ -0,0 +1,343 @@
...
+  collector.push_response("R");
+  collector.push_response("C");
+  auto addResponseCode = [&collector](sitetosite::ResponseCode code) {
+    std::string resp_code;
+    resp_code.insert(resp_code.begin(), magic_enum::enum_underlying(code));
+    collector.push_response(resp_code);
+  };
+  addResponseCode(sitetosite::ResponseCode::MORE_DATA);
```
fgerlits: I would rewrite this to
```suggestion
    collector.push_response(std::string{static_cast<char>(magic_enum::enum_underlying(sitetosite::ResponseCode::MORE_DATA))});
```
because why use 6 lines of code where 1 will do

libminifi/test/unit/SiteToSiteTests.cpp
```diff
@@ -0,0 +1,343 @@
...
+  static bool sendPacket(sitetosite::RawSiteToSiteClient& client, const sitetosite::DataPacket& packet) {
+    return client.sendPacket(packet);
+  }
+
+  static std::pair<uint64_t, uint64_t> readFlowFiles(sitetosite::RawSiteToSiteClient& client, const std::shared_ptr<sitetosite::Transaction>& transaction, core::ProcessSession& session) {
+    return client.readFlowFiles(transaction, session);
+  }
+};
+
+void initializeMockBootstrapResponses(const std::unique_ptr<SiteToSiteResponder>& collector) {
```
fgerlits: the parameter type of this could be `SiteToSiteResponder&`, too

libminifi/src/sitetosite/HttpSiteToSiteClient.cpp
lordgamez: I think in this case this is not an error because we received a valid json response, but it may be an issue on the peer's side, or it may not add the peers if there are none at the moment. I can change the debug log to a warning as it may be suspicious.

libminifi/src/sitetosite/RawSiteToSiteClient.cpp
lordgamez: It's part of the S2S protocol but we do not do anything with it on our side, so it must still be read here, but I removed it from our code as it was unused.

SITE_TO_SITE.md
lordgamez: Replaced with yaml in https://github.com/apache/nifi-minifi-cpp/pull/1966/commits/76e948709325fdbef2805bd8d563094696d394fb

SITE_TO_SITE.md
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1966/commits/76e948709325fdbef2805bd8d563094696d394fb

SITE_TO_SITE.md
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1966/commits/76e948709325fdbef2805bd8d563094696d394fb

docker/test/integration/features/s2s.feature
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1966/commits/76e948709325fdbef2805bd8d563094696d394fb

docker/test/integration/minifi/flow_serialization/Minifi_flow_json_serializer.py
lordgamez: Good idea, updated in https://github.com/apache/nifi-minifi-cpp/pull/1966/commits/76e948709325fdbef2805bd8d563094696d394fb

extensions/standard-processors/tests/unit/FlowJsonTests.cpp
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1966/commits/76e948709325fdbef2805bd8d563094696d394fb

libminifi/include/sitetosite/Peer.h
lordgamez: Removed in https://github.com/apache/nifi-minifi-cpp/pull/1966/commits/76e948709325fdbef2805bd8d563094696d394fb

libminifi/src/sitetosite/RawSiteToSiteClient.cpp
lordgamez: Deleted in https://github.com/apache/nifi-minifi-cpp/pull/1966/commits/76e948709325fdbef2805bd8d563094696d394fb

libminifi/include/sitetosite/SiteToSiteClient.h
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1966/commits/76e948709325fdbef2805bd8d563094696d394fb

libminifi/include/sitetosite/SiteToSite.h
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1966/commits/76e948709325fdbef2805bd8d563094696d394fb

libminifi/include/sitetosite/SiteToSiteClient.h
lordgamez: I think it is not used anymore, removed in https://github.com/apache/nifi-minifi-cpp/pull/1966/commits/76e948709325fdbef2805bd8d563094696d394fb

libminifi/include/sitetosite/SiteToSiteClient.h
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1966/commits/96757448fff14b52dd92982bdbcd0f0a9baed4f6

libminifi/include/sitetosite/SiteToSiteClient.h
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1966/commits/76e948709325fdbef2805bd8d563094696d394fb

libminifi/include/sitetosite/SiteToSite.h
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1966/commits/76e948709325fdbef2805bd8d563094696d394fb

libminifi/src/sitetosite/SiteToSiteFactory.cpp
lordgamez: Added std::terminate in https://github.com/apache/nifi-minifi-cpp/pull/1966/commits/96757448fff14b52dd92982bdbcd0f0a9baed4f6

libminifi/test/libtest/integration/HTTPHandlers.cpp
lordgamez: Updated to static const for now in https://github.com/apache/nifi-minifi-cpp/pull/1966/commits/76e948709325fdbef2805bd8d563094696d394fb

libminifi/test/libtest/unit/DummyProcessor.h
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1966/commits/76e948709325fdbef2805bd8d563094696d394fb

libminifi/include/RemoteProcessGroupPort.h
lordgamez: I don't think we need the atomic, but we need an additional block guarded by a mutex, updated in https://github.com/apache/nifi-minifi-cpp/pull/1966/commits/807bd4b12eccb288119d2cab9f78287f78d5ba2d

libminifi/src/RemoteProcessGroupPort.cpp
lordgamez: You are right, at the moment I don't see the use for the RPGLatch so I think it can be removed, I removed it in https://github.com/apache/nifi-minifi-cpp/pull/1966/commits/0582224996e0e59fc2f95892ee12d96a76c24535

libminifi/src/sitetosite/HttpSiteToSiteClient.cpp
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1966/commits/76e948709325fdbef2805bd8d563094696d394fb

libminifi/src/sitetosite/HttpSiteToSiteClient.cpp
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1966/commits/ad00aeac318bd66a106110a27f2b50acbd27c1e9

libminifi/test/unit/SiteToSiteTests.cpp
lordgamez: Fixed in https://github.com/apache/nifi-minifi-cpp/pull/1966/commits/76e948709325fdbef2805bd8d563094696d394fb

libminifi/test/unit/SiteToSiteTests.cpp
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1966/commits/76e948709325fdbef2805bd8d563094696d394fb

libminifi/test/unit/SiteToSiteTests.cpp
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1966/commits/76e948709325fdbef2805bd8d563094696d394fb

fgerlits (approved):

libminifi/test/libtest/integration/HTTPHandlers.cpp
```diff
@@ -280,7 +301,7 @@ void HeartbeatHandler::verifyJsonHasAgentManifest(const rapidjson::Document& roo
   REQUIRE(manifestHash.length() == 128);
 
   // throws if not a valid hexadecimal hash
-  const auto hashVec = utils::string::from_hex(manifestHash);
+  const auto hashVec = minifi::utils::string::from_hex(manifestHash);
```
szaszm, line 304: why? When we're in the minifi namespace, we don't need to mention it when using things from within the namespace.

SITE_TO_SITE.md
```diff
@@ -0,0 +1,133 @@
...
+- [Site-to-Site Configuration](#site-to-site-configuration)
+  - [Site-to-Site Configuration on NiFi side](#site-to-site-configuration-on-nifi-side)
+  - [Site-to-Site Configuration on MiNiFi C++ side](#site-to-site-configuration-on-minifi-c-side)
+- [Additional examples](#additional-examples)
+
+## Site-to-Site Overview
+
+Site-to-Site protocol allows data to be transferred between MiNiFi C++ and NiFi instances. MiNiFi C++ can send or receive data from NiFi using remote process groups. This is useful for scenarios where you want to send data from MiNiFi C++ to NiFi or vice versa. Site-to-Site protocol support raw TCP and HTTP protocols.
+
+At the moment site-to-site protocol is only supported between MiNiFi C++ and NiFi instances, it cannot be used to transfer data between multiple MiNiFi C++ instances. It is recommended to use processors like InvokeHTTP and ListenHTTP to transfer data between MiNiFi C++ instances.
```
szaszm, line 27: Is there a technical reason for this limitation?

SITE_TO_SITE.md
```diff
@@ -0,0 +1,133 @@
...
+    transport protocol: RAW
+    Input Ports:
+      - id: de7cc09a-0196-1000-2c63-ee6b4319ffb6  # this is the instance id of the input port created in NiFi
+        name: nifi-inputport
+        max concurrent tasks: 1
+        use compression: false  # currently not supported and ignored in MiNiFi C++
+        batch size:
+          size: 10 MB
+          count: 10
+          duration: 30 sec
```
szaszm, line 79: What does this do? I see it showed up in many places, but doesn't seem to be documented.

lordgamez: This implements the data limit configuration options for a remote port in a single transaction. These are the same remote port configuration options as defined in the NiFi's flow configuration files, they are documented for NiFi here: https://nifi.apache.org/docs/nifi-docs/html/user-guide.html#Remote_Port_Configuration

"For an Input Port, the batch settings control how NiFi sends data to the remote input port in a transaction. NiFi will transfer FlowFiles, as they are queued in incoming relationships, until any of the limits (Count, Size, Duration) is met. If none of the settings are configured, a 500 milliseconds batch duration is used by default.

For an Output Port, the batch settings tells the remote NiFi how NiFi prefers to receive data from the remote output port in a transaction. The remote NiFi will use the specified settings (Count, Size, Duration) to control the transfer of FlowFiles. If none of the settings are configured, a 5 seconds batch duration is used by default."

SITE_TO_SITE.md
lordgamez: Unfortunately we only implement remote ports in MiNiFi C++ at the moment that can send and receive data through ports created in NiFi. What we have as ports today in MiNiFi are only capable to transfer data between local process groups, so no networking is implemented there at the moment.

libminifi/test/libtest/integration/HTTPHandlers.cpp
lordgamez: I think there were some namespace changes and in the `minifi::test` namespace where this file we also have `test::utils`, so is if we use `utils` namespace the compiler looks for the function in the `minifi::test::utils` where it does not find this function, so we have to use `minifi::utils` here specifically to avoid the compilation error.

SITE_TO_SITE.md
szaszm: Why are they suddenly configured in a lot of places? Is it mandatory?

lordgamez: It is not mandatory, but it was unsupported before, so after implementing it I just added it to tests and documentation to show it is supported in MiNiFi as well.

libminifi/include/sitetosite/SiteToSiteClient.h
```diff
@@ -19,246 +19,180 @@
...
-  uint32_t _currentVersion{_supportedVersion[_currentVersionIndex]};
-  uint32_t _supportedCodecVersion[1] = {1};
-  int _currentCodecVersionIndex{0};
-  uint32_t _currentCodecVersion{_supportedCodecVersion[_currentCodecVersionIndex]};
+  const std::vector<uint32_t> supported_versions_ = {5, 4, 3, 2, 1};
+  uint32_t current_version_index_{0};
+  uint32_t current_version_{supported_versions_[current_version_index_]};
+  const std::vector<uint32_t> supported_codec_versions_ = {1};
+  uint32_t current_codec_version_index_{0};
+  uint32_t current_codec_version_{supported_codec_versions_[current_codec_version_index_]};
```
szaszm, line 164: Why is it ok to have these variables without synchronizatioen, while the rest of them are atomic below?

libminifi/src/sitetosite/SiteToSiteClient.cpp
```diff
@@ -20,735 +20,740 @@
...
-    throw;
-  } catch (...) {
-    if (transaction)
-      deleteTransaction(transactionID);
-    context.yield();
-    tearDown();
-    logger_->log_debug("Caught Exception during SiteToSiteClient::transferFlowFiles, type: {}", getCurrentExceptionTypeName());
+    logger_->log_debug("Site2Site transaction {} successfully sent flow record {}, content bytes {}", transaction_id.to_string(), transaction->getCurrentTransfers(), transaction->getBytes());
+  } catch (const std::exception& exception) {
+    handleTransactionError(transaction, context, exception);
```
szaszm, line 172: I'd prefer if this was done with RAII or gsl::finally

adamdebreceni (approved):

libminifi/src/RemoteProcessGroupPort.cpp
```diff
@@ -0,0 +1,369 @@
...
+      return result;
+    }
+  }
+  return std::nullopt;
+}
+
+void RemoteProcessGroupPort::refreshPeerList() {
+  auto connection = refreshRemoteSiteToSiteInfo();
+  if (!connection) {
+    logger_->log_debug("No port configured");
```
szaszm: Shouldn't this be a warning or error?

libminifi/src/RemoteProcessGroupPort.cpp
```diff
@@ -0,0 +1,369 @@
...
+  }
+  auto client = std::unique_ptr<http::BaseHTTPClient>(dynamic_cast<http::BaseHTTPClient*>(client_ptr));
+  client->initialize(http::HttpRequestMethod::GET, login_url.str(), ssl_service_);
+  // use a connection timeout. if this times out we will simply attempt re-connection
+  // so no need for configuration parameter that isn't already defined in Processor
+  client->setConnectionTimeout(10s);
+  client->setReadTimeout(idle_timeout_);
+
+  auto token = http::get_token(client.get(), rest_user_name, rest_password);
+  logger_->log_debug("Token from NiFi REST Api endpoint {},  {}", login_url.str(), token);
```
szaszm: We should probably not log the token because it's sensitive info

libminifi/include/sitetosite/SiteToSiteClient.h
lordgamez: I'm not sure what the original intent was. I think these do not need synchronization because these variables are only used in the class internally in the initialization of the client, during version negotiation, which is only done once and never in parallel. Meanwhile the others below can be changed from the outside using the setter functions where concurrency is possible.

libminifi/src/RemoteProcessGroupPort.cpp
lordgamez: You're right, it should be a warning at least, updated in https://github.com/apache/nifi-minifi-cpp/pull/1966/commits/6a61e1c192b1dfcb6c5f0977432f73ee12557367

libminifi/src/RemoteProcessGroupPort.cpp
lordgamez: Good point, updated in https://github.com/apache/nifi-minifi-cpp/pull/1966/commits/6a61e1c192b1dfcb6c5f0977432f73ee12557367

libminifi/src/sitetosite/SiteToSiteClient.cpp
lordgamez: I think for the `deleteTransaction` it would be good to use the `gsl::finally`, but in case of failure that should be run before the `tearDown` so it cannot be used in all cases. Everything else in `handleTransactionError` should only be used in error cases so I would keep it this way.

PR #1968, opened 2025-05-27
MINIFICPP-2567 C++23 Support

martinzink:

### For all changes:
     in the commit message?

### For code changes:

### Note:


thirdparty/couchbase/c++23_fixes.patch
fgerlits, line 1: Can you put a comment somewhere about what this patch is about, please?

minifi-api/include/minifi-cpp/core/RecordField.h
```diff
@@ -30,26 +30,8 @@ namespace org::apache::nifi::minifi::core {
...
-  explicit BoxedRecordField(std::unique_ptr<RecordField>&& _field) : field(std::move(_field)) {}
-  bool operator==(const BoxedRecordField&) const;
-  std::unique_ptr<RecordField> field = nullptr;
-};
-
-
 using RecordArray = std::vector<RecordField>;
-
-using RecordObject = std::unordered_map<std::string, BoxedRecordField>;
+using RecordObject = std::unordered_map<std::string, RecordField>;
```
fgerlits, line 34: Just out of curiosity, why is the boxing no longer necessary? (Or why was it previously necessary?)

martinzink: It was neccessary on centos 7, because it didnt compile without it. And now it doesnt compile with c++23 due to more restrictions on incomplete types with unique_ptr

szaszm:
The readme needs to be updated to reflect the supported compiler versions, and changes in the build procedure if any

lordgamez (approved):

thirdparty/couchbase/c++23_fixes.patch
szaszm, line 1: did you author these patches? Normally when there is text above the diff, I assume it might come from a mailing list or somewhere from another source, in which case we should give credit to the original author of the patch.

martinzink: yeah I did author them, the boilerplate is due using my IDE for the patch creation

martinzink:
I've included @lordgamez 's [clang20 fixes](https://github.com/apache/nifi-minifi-cpp/pull/1968/commits/e22e8555fac819fc660a4d75e9327d6fa8d0d2ad) and @fgerlits's [windows fixes](https://github.com/apache/nifi-minifi-cpp/pull/1968/commits/94324e0c51e30050322e909b2850f7181b1e0ac0). Thank you guys.

I've also made a bunch of gcc15 modifications so please rereview it

martinzink:
I've also added a new ci that checks the min and max supported gcc and clang versions. It successfully [ran](https://github.com/martinzink/nifi-minifi-cpp/actions/runs/16620893385) with ENABLE_ALL and MINIFI_FAIL_ON_WARNINGS

.github/workflows/compiler-support.yml
```diff
@@ -0,0 +1,52 @@
+# .github/workflows/compiler-support.yml
+
+name: 'Check supported Compilers'
+
+on:
+  workflow_dispatch:
```
fgerlits: Is the colon at the end of line 6 correct? I would expect
```suggestion
on: [workflow_dispatch]
```

CMakeLists.txt
```diff
@@ -316,8 +316,12 @@ else()
     if(MINIFI_FAIL_ON_WARNINGS)
         list(APPEND MINIFI_CPP_COMPILE_OPTIONS -Werror)
         # -Wrestrict may cause build failure in GCC 12  https://gcc.gnu.org/bugzilla/show_bug.cgi?id=104336
-        if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
-            list(APPEND MINIFI_CPP_COMPILE_OPTIONS -Wno-error=restrict)
+    endif()
+    if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+        list(APPEND MINIFI_CPP_COMPILE_OPTIONS -Wno-error=restrict)
```
fgerlits: Can we remove `-Wno-error=restrict`? The bug linked to in line 318 was fixed in GCC 12.4; I don't think we need to support GCC 12.1 to 12.3.

cmake/ExpectedLite.cmake
```diff
@@ -18,7 +18,10 @@
...
 
 FetchContent_Declare(expected-lite
-    URL      https://github.com/martinmoene/expected-lite/archive/refs/tags/v0.8.0.tar.gz
-    URL_HASH SHA256=27649f30bd9d4fe7b193ab3eb6f78c64d0f585c24c085f340b4722b3d0b5e701
+    URL      https://github.com/martinmoene/expected-lite/archive/refs/tags/v0.9.0.tar.gz
+    URL_HASH SHA256=e1b3ac812295ef8512c015d8271204105a71957323f8ab4e75f6856d71b8868d
+    SYSTEM
 )
+
+add_compile_definitions(nsel_CONFIG_SELECT_EXPECTED=1) # Due to https://gcc.gnu.org/bugzilla/show_bug.cgi?id=119714
```
fgerlits: Is this the same as
```suggestion
# Due to https://gcc.gnu.org/bugzilla/show_bug.cgi?id=119714
add_compile_definitions(nsel_CONFIG_SELECT_EXPECTED=nsel_EXPECTED_NONSTD)
```
? I think that would be more readable than "1".

extensions/sql/processors/ExecuteSQL.cpp
```diff
@@ -56,7 +56,8 @@ void ExecuteSQL::processOnTrigger(core::ProcessContext& context, core::ProcessSe
                       "No incoming FlowFile and the \"" + std::string{SQLSelectQuery.name} + "\" processor property is not specified");  // NOLINT(whitespace/braces)
     }
     logger_->log_debug("Using the contents of the flow file as the SQL statement");
-    query = to_string(session.readBuffer(input_flow_file));
+    std::string buffer_str = to_string(session.readBuffer(input_flow_file));
+    query = buffer_str;
```
fgerlits, line 56: what was the problem here?

libminifi/test/unit/ExpectedTest.cpp
```diff
@@ -558,6 +558,15 @@ TEST_CASE("expected orThrow") {
   nonstd::expected<int, std::string> unexpected{nonstd::unexpect, "hello"};
   nonstd::expected<int, std::string> expected{5};
 
+
   REQUIRE_THROWS_WITH(std::move(unexpected) | utils::orThrow("should throw"), "should throw, but got hello");
   CHECK((expected | utils::orThrow("should be 5")) == 5);
 }
+
+// This fails to compile with std::expected on GCC 15.1 due to https://gcc.gnu.org/bugzilla/show_bug.cgi?id=119714
+TEST_CASE("") {
```
fgerlits: Let's give this test a name; for example, the comment would work:
```suggestion
TEST_CASE("This fails to compile with std::expected on GCC 15.1 due to https://gcc.gnu.org/bugzilla/show_bug.cgi?id=119714") {
```

extensions/sql/processors/ExecuteSQL.cpp
martinzink: This was a false positive maybe uninitiziled warning on gcc

CMakeLists.txt
martinzink: https://github.com/apache/nifi-minifi-cpp/pull/1968/commits/05314ac3faa57fe01aad73fe526312ae04e06e18

martinzink:
https://github.com/martinzink/nifi-minifi-cpp/actions/runs/16651051825/job/47123587158 🟢

fgerlits (approved):

.dockerignore
```diff
@@ -64,3 +64,5 @@ libminifi/src/agent/agent_version.cpp
 **/cmake_install.cmake
 **/install_manifest.txt
 **/CTestTestfile.cmake
+bootstrap/venv
+bootstrap/option_state.json
```
szaszm:
```suggestion
bootstrap/option_state.json

```

cmake/ExpectedLite.cmake
```diff
@@ -18,7 +18,11 @@
...
 FetchContent_Declare(expected-lite
-    URL      https://github.com/martinmoene/expected-lite/archive/refs/tags/v0.8.0.tar.gz
-    URL_HASH SHA256=27649f30bd9d4fe7b193ab3eb6f78c64d0f585c24c085f340b4722b3d0b5e701
+    URL      https://github.com/martinmoene/expected-lite/archive/refs/tags/v0.9.0.tar.gz
+    URL_HASH SHA256=e1b3ac812295ef8512c015d8271204105a71957323f8ab4e75f6856d71b8868d
+    SYSTEM
 )
+
+# Due to https://gcc.gnu.org/bugzilla/show_bug.cgi?id=119714
+add_compile_definitions(nsel_CONFIG_SELECT_EXPECTED=nsel_EXPECTED_NONSTD)
```
szaszm: Please change this to use target-based commands, i.e. target_compile_definitions. We should never use legacy global / directory-based cmake commands!

libminifi/test/unit/ExpectedTest.cpp
```diff
@@ -473,7 +473,7 @@ TEST_CASE("expected valueOrElse", "[expected][valueOrElse]") {
   REQUIRE(gsl::narrow<int>("hello"sv.size()) == (ex | utils::valueOrElse([](const std::string& err) { return gsl::narrow<int>(err.size()); })));
   REQUIRE_THROWS_AS(ex | utils::valueOrElse([](std::string){ throw std::exception(); }), std::exception);  // NOLINT(performance-unnecessary-value-param)
   REQUIRE_THROWS_AS(ex | utils::valueOrElse([](const std::string&) -> int { throw std::exception(); }), std::exception);
-  REQUIRE_THROWS_WITH(std::move(ex) | utils::valueOrElse([](std::string&& error) -> int { throw std::runtime_error(error); }), "hello");
+  REQUIRE_THROWS_WITH([&] { std::move(ex) | utils::valueOrElse([](const std::string& error) -> int { throw std::runtime_error(error); }); }(), "hello");
```
szaszm: Why is the string taken by const ref now? The point of the test is to ensure that when an expected is moved, the callback gets an rvalue.

thirdparty/couchbase/remove-thirdparty.patch
```diff
@@ -76,7 +60,89 @@ index f02af02..f83c181 100644
      "CMAKE_C_VISIBILITY_PRESET hidden"
      "CMAKE_CXX_VISIBILITY_PRESET hidden"
      "CMAKE_POSITION_INDEPENDENT_CODE ON")
-@@ -159,93 +116,24 @@ if(NOT TARGET taocpp::json)
+ endif()
```
szaszm, line 63: this extra context seems unnecessary for the patch

thirdparty/iODBC/GCC-15-needs-typedef-SQLRETURN-HPROC.patch
szaszm, line 1: did you submit this upstream?

thirdparty/iODBC/GCC-15-needs-typedef-SQLRETURN-HPROC.patch
```diff
@@ -0,0 +1,25 @@
...
+ #include <dlf.h>
+ 
+-#if defined(_MAC) || defined (__cplusplus)
++#if defined(_MAC) || defined (__cplusplus) || (defined(__GNUC__) && __GNUC__ >= 15)
+ typedef SQLRETURN (* HPROC) (...);
+ #else
+ typedef SQLRETURN (* HPROC) ();
+-- 
+2.50.1
+
```
szaszm, line 25: Is this valid?

martinzink: What do you mean?

thirdparty/iODBC/GCC-15-needs-typedef-SQLRETURN-HPROC.patch
martinzink: Not yet, but good idea, will do so 👍

thirdparty/couchbase/remove-thirdparty.patch
martinzink: This part was done by @lordgamez 
Could you address this question pls

.dockerignore
martinzink: https://github.com/apache/nifi-minifi-cpp/pull/1968/commits/159b86bb17e244d670b0c4682131963e7b816127#diff-2f754321d62f08ba8392b9b168b83e24ea2852bb5d815d63e767f6c3d23c6ac5L68

cmake/ExpectedLite.cmake
martinzink: https://github.com/apache/nifi-minifi-cpp/pull/1968/commits/159b86bb17e244d670b0c4682131963e7b816127#diff-e87aa3282cfbdf1445ef0ab42c4cd85b6063ca633c8cc620097ac5dc311f3ce2R28

libminifi/test/unit/ExpectedTest.cpp
martinzink: reverted back in https://github.com/apache/nifi-minifi-cpp/pull/1968/commits/bc27bc6cd63309f6796bf56e57fe6baf6cce1114#diff-8238844cc1d0c5afd7689a05f1540c3926adbde6a54c3b7934f6f071254f674fR479

thirdparty/couchbase/remove-thirdparty.patch
lordgamez: As far as I remember I think the extra context was needed because it failed with the original context size when running cmake twice with the original patch.

libminifi/test/unit/ExpectedTest.cpp
martinzink: now i remember

`/home/runner/work/nifi-minifi-cpp/nifi-minifi-cpp/libminifi/test/unit/ExpectedTest.cpp:479:78: error: rvalue reference parameter 'error' is never moved from inside the function body [cppcoreguidelines-rvalue-reference-param-not-moved,-warnings-as-errors]
  479 |   REQUIRE_THROWS_MATCHES(std::move(ex) | utils::valueOrElse([](std::string&& error) -> int { throw std::runtime_error(error); }),`

thirdparty/couchbase/remove-thirdparty.patch
szaszm: then the line numbers likely need to be updated. Try remaking the patch with diff.

thirdparty/iODBC/GCC-15-needs-typedef-SQLRETURN-HPROC.patch
szaszm: I find the double dash weird, dash usually means removed line, github even highlights it in red. But if it works :shrug:

libminifi/test/unit/ExpectedTest.cpp
szaszm: We should turn off the linter for that line then. It needs to be rvalue reference for the test to make any sense.

CMakeLists.txt
```diff
@@ -315,9 +315,11 @@ else()
     list(APPEND MINIFI_CPP_COMPILE_OPTIONS -Wall -Wextra)
     if(MINIFI_FAIL_ON_WARNINGS)
         list(APPEND MINIFI_CPP_COMPILE_OPTIONS -Werror)
-        # -Wrestrict may cause build failure in GCC 12  https://gcc.gnu.org/bugzilla/show_bug.cgi?id=104336
-        if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
-            list(APPEND MINIFI_CPP_COMPILE_OPTIONS -Wno-error=restrict)
```
szaszm: GCC 12 is still supported right? In that case we should probably keep this.

fgerlits: This bug was fixed in GCC 12.4. We could update the README to say that we support GCC 12.4 or higher.

thirdparty/couchbase/remove-thirdparty.patch
lordgamez: The problem is with the reverse patch dry-run that it needs more context if the patch only contains deletion. In this case the reverse patch dry-run succeeds and the patch application does not run if we have a smaller context.

CMakeLists.txt
szaszm: I'd keep the workaround for gcc < 12.4

PR #1974, opened 2025-06-10
MINIFICPP-2565 Add compression support for site to site communication

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2565

Depends on #1966 

-------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


core-framework/include/io/CRCStream.h
```diff
@@ -66,6 +66,10 @@ class CRCStreamBase : public virtual StreamImpl {
     crc_ = crc32(0L, Z_NULL, 0);
   }
 
+  void setCrc(uint64_t crc) {
```
fgerlits, line 69: it's a bit strange that we use `uint64_t` to represent a crc**32**; do you know why we do this?

I don't really mind, it's just strange

libminifi/include/sitetosite/CompressionConsts.h
```diff
@@ -0,0 +1,28 @@
...
+ */
+#pragma once
+
+#include "io/InputStream.h"
+#include "io/BufferStream.h"
+
+namespace org::apache::nifi::minifi::sitetosite {
+
+static constexpr size_t COMPRESSION_BUFFER_SIZE = 65536;
+static constexpr std::array<char, 4> SYNC_BYTES = { 'S', 'Y', 'N', 'C' };
```
fgerlits: these should be `inline constexpr`

libminifi/include/sitetosite/CompressionInputStream.h
```diff
@@ -0,0 +1,54 @@
...
+#include "CompressionConsts.h"
+#include "core/logging/LoggerFactory.h"
+
+namespace org::apache::nifi::minifi::sitetosite {
+
+class CompressionInputStream : public io::InputStreamImpl {
+ public:
+  explicit CompressionInputStream(gsl::not_null<io::InputStream*> internal_stream)
+      : internal_stream_(internal_stream) {
+    buffer_.resize(COMPRESSION_BUFFER_SIZE);
```
fgerlits: instead of `resize`, we could write
```
std::vector<std::byte> buffer_(COMPRESSION_BUFFER_SIZE);
```
in line 48

libminifi/src/sitetosite/CompressionInputStream.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+  if (compressed_size == 0 && original_size != 0) {
+    logger_->log_error("Compressed size is 0 but original size is not");
+    return io::STREAM_ERROR;
+  }
+
+  if (compressed_size > COMPRESSION_BUFFER_SIZE) {
+    logger_->log_error("Compressed size exceeds buffer size");
+    return io::STREAM_ERROR;
+  }
```
fgerlits, line 60: we should check this earlier, because if `compressed_size` is too large, then even calling `subspan` with it is UB

libminifi/src/sitetosite/CompressionOutputStream.cpp
```diff
@@ -0,0 +1,128 @@
...
+      return ret;
+    }
+  }
+  auto ret = internal_stream_->write(reinterpret_cast<const uint8_t *>(SYNC_BYTES.data()), SYNC_BYTES.size());
+  if (io::isError(ret)) {
+    logger_->log_error("Failed to write sync bytes before compression: {}", ret);
+    return ret;
+  }
+
+  was_data_written_ = true;
```
fgerlits: It probably doesn't matter much, but I would move this line later, probably to line 100, so `was_data_written_` is not set unless the write was successful.

core-framework/include/io/CRCStream.h
lordgamez: I think it's due to the zlib's `crc32` function we use that returns a `uLong`.

libminifi/include/sitetosite/CompressionConsts.h
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1974/commits/a623c4ad576d2e4b3c934b216b9b158157ebfa36

libminifi/include/sitetosite/CompressionInputStream.h
lordgamez: Good idea, updated in https://github.com/apache/nifi-minifi-cpp/pull/1974/commits/a623c4ad576d2e4b3c934b216b9b158157ebfa36

libminifi/src/sitetosite/CompressionInputStream.cpp
lordgamez: Good catch, updated in https://github.com/apache/nifi-minifi-cpp/pull/1974/commits/a623c4ad576d2e4b3c934b216b9b158157ebfa36

libminifi/src/sitetosite/CompressionOutputStream.cpp
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1974/commits/a623c4ad576d2e4b3c934b216b9b158157ebfa36

SITE_TO_SITE.md
fgerlits, line 1: this file needs to be added to `packaging/rpm/expected-rpm-contents.in` to fix the Rocky (docker) build

lordgamez: Updated in MINIFICPP-2558 and rebased to it

libminifi/include/sitetosite/CompressionInputStream.h
```diff
@@ -0,0 +1,53 @@
...
+#include "io/InputStream.h"
+#include "io/BufferStream.h"
+#include "CompressionConsts.h"
+#include "core/logging/LoggerFactory.h"
+
+namespace org::apache::nifi::minifi::sitetosite {
+
+class CompressionInputStream : public io::InputStreamImpl {
+ public:
+  explicit CompressionInputStream(gsl::not_null<io::InputStream*> internal_stream)
```
szaszm: I'd make that a reference and convert to pointer on the init list

libminifi/include/sitetosite/CompressionOutputStream.h
```diff
@@ -0,0 +1,49 @@
...
+#include "io/BaseStream.h"
+#include "utils/gsl.h"
+#include "CompressionConsts.h"
+#include "core/logging/LoggerFactory.h"
+
+namespace org::apache::nifi::minifi::sitetosite {
+
+class CompressionOutputStream : public io::StreamImpl, public virtual io::OutputStreamImpl {
+ public:
+  explicit CompressionOutputStream(gsl::not_null<io::OutputStream*> internal_stream)
```
szaszm: same. In parameters, we don't need mutability, so references are a simpler way to represent not nullable pointers.

libminifi/src/sitetosite/CompressionOutputStream.cpp
```diff
@@ -0,0 +1,127 @@
...
+  if (value == nullptr || len == 0) {
+    return 0;
+  }
+
+  size_t bytes_left_to_write = len;
+  size_t bytes_written = 0;
+  while (bytes_left_to_write > 0) {
+    size_t free_spaces_left_in_buffer = buffer_.size() - buffer_offset_;
+    size_t bytes_to_write = std::min(bytes_left_to_write, free_spaces_left_in_buffer);
+    std::memcpy(reinterpret_cast<uint8_t*>(buffer_.data()) + buffer_offset_, value + bytes_written, bytes_to_write);
```
szaszm: Please use `std::span` in new code instead of pointer arithmetic and `reinterpret_cast`s

libminifi/include/sitetosite/CompressionInputStream.h
lordgamez: Good point, I'm not sure why gsl::not_null was used instead of a reference, but I think we can use a reference here and for the member variable as well, updated in https://github.com/apache/nifi-minifi-cpp/pull/1974/commits/bab2c49ea9f8f9489d7682f4c0a6071a9c7883b7

libminifi/include/sitetosite/CompressionOutputStream.h
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1974/commits/bab2c49ea9f8f9489d7682f4c0a6071a9c7883b7

libminifi/src/sitetosite/CompressionOutputStream.cpp
lordgamez: Good idea, updated for `CompressionOutputStream::write` and `CompressionInputStream::read` too in https://github.com/apache/nifi-minifi-cpp/pull/1974/commits/bab2c49ea9f8f9489d7682f4c0a6071a9c7883b7

libminifi/include/sitetosite/CompressionInputStream.h
szaszm: usually for non-static data members, references and (top level) const are not recommended: they prevent move and reassignment

lordgamez: I think in this case we never want to move the CompressionInputStream nor want to change the internal stream reference as it is a wrapper of a single existing stream, so I think it's better to keep it as a reference.

PR #1984, opened 2025-06-24
MINIFICPP-2583 Fix warnings of unhandled std::expected return values

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2583

------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/lua/tests/TestExecuteScriptProcessorWithLuaScript.cpp
+++ b/extensions/lua/tests/TestExecuteScriptProcessorWithLuaScript.cpp
@@ -34,8 +34,8 @@ TEST_CASE("Lua: hello world") {
   minifi::test::SingleProcessorTestController controller{std::make_unique<ExecuteScript>("ExecuteScript")};
   const auto execute_script = controller.getProcessor();
 
-  execute_script->setProperty(ExecuteScript::ScriptEngine.name, "lua");
-  execute_script->setProperty(ExecuteScript::ScriptBody.name, R"(print("Hello world!"))");
+  REQUIRE(execute_script->setProperty(ExecuteScript::ScriptEngine.name, "lua"));
+  REQUIRE(execute_script->setProperty(ExecuteScript::ScriptBody.name, R"(print("Hello world!"))"));
 
   CHECK_NOTHROW(controller.trigger());
 }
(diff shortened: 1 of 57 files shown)


extensions/standard-processors/processors/PutUDP.cpp
```diff
@@ -120,10 +120,10 @@ void PutUDP::onTrigger(core::ProcessContext& context, core::ProcessSession& sess
     session.transfer(flow_file, Failure);
   };
 
-  resolve_hostname()
+  (void)(resolve_hostname()
```
szaszm: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es48-avoid-casts

The C++ Core Guidelines recommends assignment to `std::ignore` to ignore `[[nodiscard]]` values. I think it may be slightly better, but it's fine if you like this more.

extensions/lua/tests/TestExecuteScriptProcessorWithLuaScript.cpp
```diff
@@ -46,8 +46,8 @@ TEST_CASE("Script engine is not set", "[executescriptMisconfiguration]") {
 
   auto execute_script = plan->addProcessor("ExecuteScript", "executeScript");
 
-  plan->setProperty(execute_script, ExecuteScript::ScriptEngine, "");
-  plan->setProperty(execute_script, ExecuteScript::ScriptFile, "/path/to/script.lua");
+  REQUIRE_FALSE(plan->setProperty(execute_script, ExecuteScript::ScriptEngine, ""));
```
szaszm, line 49: why is this supposed to fail?

lordgamez: The ScriptEngine property is configured to only allow "lua" and "python" values so it fails with setting empty string in it.

extensions/standard-processors/processors/PutUDP.cpp
lordgamez: Good point, updated in https://github.com/apache/nifi-minifi-cpp/pull/1984/commits/e178f089193cb421dcc0f9c7aa4ff10344daa097

szaszm (approved):

PR #1985, opened 2025-06-27
MINIFICPP-2576 Improve the performance of ConsumeWindowsEventLog

fgerlits:
https://issues.apache.org/jira/browse/MINIFICPP-2576

This PR contains two changes, both affecting the same files, so it is probably easiest to review the two commits separately.

change 1: Cache the results of `getEventData()` in `MetadataWalker::for_each()`. Previously, this typically did five Windows API calls to `EvtFormatMessage` per event.

change 2: If the value of the `Identifier Match Regex` property is of the form `.*something` (in particular, if the default value of `.*Sid` is used), then the calls to `regexMatch()` are replaced with calls to `endsWith()`.

Before these changes, processing took around 1.1 ms per event.  After change one, this got reduced to 0.5 ms per event; after both changes, the processing time is around 0.4 ms per event (as measured by the `"processed {} Events in {}"` log line in `ConsumeWindowsEventLog.cpp`).

--- 

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/extensions/windows-event-log/ConsumeWindowsEventLog.h
+++ b/extensions/windows-event-log/ConsumeWindowsEventLog.h
@@ -74,6 +74,8 @@ enum class JsonFormat {
   Simple,
   Flattened,
 };
+
+std::function<bool(std::string_view)> parseSidMatcher(const std::optional<std::string>& sid_matcher);
 }  // namespace cwel
 
 class Bookmark;
@@ -242,7 +244,7 @@ class ConsumeWindowsEventLog : public core::ProcessorImpl {
   std::optional<std::string> header_delimiter_;
   wel::EventPath path_;
   std::wstring wstr_query_;
-  std::optional<utils::Regex> regex_;
+  std::function<bool(std::string_view)> sid_matcher_;
   bool resolve_as_attributes_{false};
   bool apply_identifier_function_{false};
   std::string provenanceUri_;
--- a/extensions/windows-event-log/wel/MetadataWalker.h
+++ b/extensions/windows-event-log/wel/MetadataWalker.h
@@ -49,11 +49,12 @@ namespace org::apache::nifi::minifi::wel {
  */
 class MetadataWalker : public pugi::xml_tree_walker {
  public:
-  MetadataWalker(const WindowsEventLogMetadata& windows_event_log_metadata, std::string log_name, bool update_xml, bool resolve, utils::Regex const* regex,
+  MetadataWalker(const WindowsEventLogMetadata& windows_event_log_metadata, std::string log_name, bool update_xml, bool resolve,
+      std::function<bool(std::string_view)> sid_matcher,
       std::function<std::string(std::string)> user_id_to_username_fn)
       : windows_event_log_metadata_(windows_event_log_metadata),
         log_name_(std::move(log_name)),
-        regex_(regex),
+        sid_matcher_(std::move(sid_matcher)),
         update_xml_(update_xml),
         resolve_(resolve),
         user_id_to_username_fn_(std::move(user_id_to_username_fn)) {
@@ -82,8 +83,6 @@ class MetadataWalker : public pugi::xml_tree_walker {
     return "N/A";
   }
 
-  static std::string to_string(const wchar_t* pChar);
-
   /**
    * Updates text within the XML representation
    */
@@ -96,7 +95,7 @@ class MetadataWalker : public pugi::xml_tree_walker {
 
   const WindowsEventLogMetadata& windows_event_log_metadata_;
   const std::string log_name_;
-  utils::Regex const * const regex_;
+  std::function<bool(std::string_view)> sid_matcher_;
   const bool update_xml_;
   const bool resolve_;
   std::function<std::string(const std::string&)> user_id_to_username_fn_;
--- a/extensions/windows-event-log/wel/WindowsEventLog.cpp
+++ b/extensions/windows-event-log/wel/WindowsEventLog.cpp
@@ -42,6 +42,18 @@ std::string getEventTimestampStr(uint64_t event_timestamp) {
 }
 }  // namespace
 
+std::optional<std::string> EventDataCache::get(EVT_FORMAT_MESSAGE_FLAGS field, const std::string& key) const {
+    std::lock_guard<std::mutex> lock{mutex_};
+    const auto it = cache_.find(CacheKey{field, key});
+    if (it != cache_.end() && it->second.expiry > std::chrono::system_clock::now()) { return it->second.value; }
+    return std::nullopt;
+}
+
+void EventDataCache::set(EVT_FORMAT_MESSAGE_FLAGS field, const std::string& key, std::string value) {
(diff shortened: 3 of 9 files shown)


extensions/windows-event-log/ConsumeWindowsEventLog.h
```diff
@@ -117,6 +119,7 @@ class ConsumeWindowsEventLog : public core::ProcessorImpl {
       .build();
   EXTENSIONAPI static constexpr auto IdentifierMatcher = core::PropertyDefinitionBuilder<>::createProperty("Identifier Match Regex")
       .isRequired(false)
+      .withValidator(core::StandardPropertyValidators::NON_BLANK_VALIDATOR)
```
szaszm: why non blank validator if it's optional?

extensions/windows-event-log/wel/MetadataWalker.h
```diff
@@ -49,11 +49,12 @@ namespace org::apache::nifi::minifi::wel {
...
 class MetadataWalker : public pugi::xml_tree_walker {
  public:
-  MetadataWalker(const WindowsEventLogMetadata& windows_event_log_metadata, std::string log_name, bool update_xml, bool resolve, utils::Regex const* regex,
+  MetadataWalker(const WindowsEventLogMetadata& windows_event_log_metadata, std::string log_name, bool update_xml, bool resolve,
+      const std::function<bool(std::string_view)>& sid_matcher,
       std::function<std::string(std::string)> user_id_to_username_fn)
       : windows_event_log_metadata_(windows_event_log_metadata),
         log_name_(std::move(log_name)),
-        regex_(regex),
+        sid_matcher_(sid_matcher),
```
szaszm: I'd do pass by value + move

extensions/windows-event-log/wel/WindowsEventLog.h
```diff
@@ -58,12 +67,27 @@ enum METADATA {
...
+
+ private:
+  struct CacheItem {
+    std::string value;
+    std::chrono::system_clock::time_point expiry;
+  };
+
+  std::mutex mutex_;
+  std::chrono::milliseconds lifetime_;
+  std::unordered_map<std::tuple<EVT_FORMAT_MESSAGE_FLAGS, std::string>, CacheItem> cache_;
```
szaszm: I think a simple aggregate struct would be better than a tuple

extensions/windows-event-log/wel/WindowsEventLog.cpp
```diff
@@ -42,6 +42,18 @@ std::string getEventTimestampStr(uint64_t event_timestamp) {
 }
 }  // namespace
 
+std::optional<std::string> EventDataCache::get(EVT_FORMAT_MESSAGE_FLAGS field, const std::string& key) {
+    std::lock_guard<std::mutex> lock{mutex_};
+    const auto it = cache_.find(std::make_tuple(field, key));
+    if (it != cache_.end() && it->second.expiry > std::chrono::system_clock::now()) { return it->second.value; }
```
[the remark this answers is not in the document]
fgerlits: There is a fairly small set of possible cache keys; also, if the lookup fails because of cache expiry, we will replace the expired value with a new cache item in `WindowsEventLogHandler::getEventData()`.

extensions/windows-event-log/ConsumeWindowsEventLog.h
fgerlits: A blank value for `Identifier Match Regex` doesn't really make sense: what it does is disable Sid replacements, but the correct way to do that is to set the `Apply Identifier Function` property to `false`.

However, I reverted this change in 4bcc689a55689ce36262d7b591dcb8fa53d8ecfe, as it could break some flows, if someone has `Identifier Match Regex` set to a blank value.  I would say these flows are incorrect, but they used to work, so they should probably continue to work.

extensions/windows-event-log/wel/MetadataWalker.h
fgerlits: done: 6b5801d34539ec0b12a965e64a38ae5647401c4e

extensions/windows-event-log/wel/WindowsEventLog.h
fgerlits: done: b7d5e1253542941262e4a90342f6c1bbab3c86ca + e26dfd9afed55e9a9abe96f753c50309fc91fc74 (the latter was the robot's suggestion)

extensions/windows-event-log/wel/WindowsEventLog.h
```diff
@@ -41,6 +41,15 @@
 #include "pugixml.hpp"
 #include "utils/expected.h"
 
+template<>
+struct std::hash<std::tuple<EVT_FORMAT_MESSAGE_FLAGS, std::string>> {
```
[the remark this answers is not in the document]
fgerlits: done: e26dfd9afed55e9a9abe96f753c50309fc91fc74

lordgamez (approved):

extensions/windows-event-log/wel/WindowsEventLog.h
```diff
@@ -58,11 +58,39 @@ enum METADATA {
   COMPUTER,
   UNKNOWN
 };
+// this is a continuous enum, so we can rely on the array
```
szaszm: what array? This comment seems to apply to the enum, so I'd put it above that.

fgerlits: My theory is that it refers to the vector below it, which at some earlier point may have been an array indexed by enum values treated as integers (although I haven't found evidence for this in git history).

We should get rid of the `METADATA_NAMES` type together with the `getMetadataString()` and `getMetadataFromString()` functions, and replace them by magic_enum calls.  But I would prefer to do that separately in a different pull request, as it is not related to performance improvement.

For now, I have removed this log line: e359090170445e8f2769b13fa451cfc476af9918

martinzink (approved):

extensions/windows-event-log/wel/WindowsEventLog.h
```diff
@@ -58,11 +58,38 @@ enum METADATA {
...
-// this is a continuous enum, so we can rely on the array
+ private:
+  struct CacheKey {
+    EVT_FORMAT_MESSAGE_FLAGS field;
+    std::string key;
 
-using METADATA_NAMES = std::vector<std::pair<METADATA, std::string>>;
+    [[nodiscard]] bool operator==(const CacheKey& other) const noexcept {
+      return field == other.field && key == other.key;
+    }
```
szaszm: could this be implemented with an explicitly defaulted operator<=>?

fgerlits: I like the idea of `= default`, but I don't think we need ordering. How about this: 7e3400a9c02c0dd333ff447e1ef75b4ff27d96f1?

szaszm: looks good to me

PR #1986, opened 2025-07-01
MINIFICPP-2232 Remove option to load SSL context from flow config

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2232

--------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


martinzink (approved):

libminifi/test/unit/GeneralUtilsTest.cpp
```diff
@@ -27,55 +27,55 @@
...
+static_assert(-3 == minifi::utils::intdiv_ceil(-6, 2));
+static_assert(-3 == minifi::utils::intdiv_ceil(6, -2));
+static_assert(3 == minifi::utils::intdiv_ceil(-6, -2));
+static_assert(0 == minifi::utils::intdiv_ceil(0, -10));
 
 template<int N, int D, typename = void>
 struct does_compile : std::false_type {};
 
 template<int N, int D>
 struct does_compile<N, D,
```
fgerlits: this is a bit misleading: `intdiv_ceil(123, 0)` does compile, it just doesn't compile as a constant expression (because it throws)
```suggestion
struct compiles_as_a_constant_expression<N, D,
```

szaszm: I wrote this a few years ago, and that was intentional. It's only used for the test below, in forced constexpr context, as mentioned in the assert message too. I don't mind changing it if you think that would make it better, but I think in the context it's clear: it's only used in constexpr context.

libminifi/test/unit/GeneralUtilsTest.cpp
szaszm, line 1: What changed that made the `minifi::` prefix necessary? If it's not necessary, I prefer not adding the extra namespace prefix.

libminifi/test/unit/GeneralUtilsTest.cpp
fgerlits: I think the name I suggested would be better, but obviously this has nothing to do with the current change, so I don't mind if it stays as it is.

lordgamez: I think the suggested naming is more explicit, updated in https://github.com/apache/nifi-minifi-cpp/pull/1986/commits/68cd493925f83232bd843496eaecf9b472db3bb4

libminifi/test/unit/GeneralUtilsTest.cpp
lordgamez: I think some transitive header includes changed and it does not compile now without it. It's only needed for the `utils` namespace, otherwise if `minifi::` is not stated explicitly, the compiler looks for the function in the `minifi::test::utils` namespace instead.

libminifi/test/unit/GeneralUtilsTest.cpp
```diff
@@ -27,55 +27,55 @@
...
+  REQUIRE(42 == minifi::utils::dereference(pa));
+  REQUIRE(&a == &minifi::utils::dereference(pa));
 
   const auto uniq_a = gsl::make_not_null(std::make_unique<int>(99));
-  REQUIRE(99 == utils::dereference(uniq_a));
+  REQUIRE(99 == minifi::utils::dereference(uniq_a));
 }
 
-TEST_CASE("GeneralUtils::unsafe_dereference", "[unsafe_dereference]") {
+TEST_CASE("Generalminifi::utils::unsafe_dereference", "[unsafe_dereference]") {
```
adamdebreceni: these test names got accidentally got

lordgamez: Good catch, fixed in https://github.com/apache/nifi-minifi-cpp/pull/1986/commits/638d8001a9db6642b9f80d0f78d77931d41c4a39

PR #1987, opened 2025-07-08
WIP processor c api

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/python/ExecutePythonProcessor.cpp
+++ b/extensions/python/ExecutePythonProcessor.cpp
@@ -227,10 +227,10 @@ std::vector<core::Relationship> ExecutePythonProcessor::getPythonRelationships()
   return relationships;
 }
 
-void ExecutePythonProcessor::setLoggerCallback(const std::function<void(core::logging::LOG_LEVEL level, const std::string& message)>& callback) {
+void ExecutePythonProcessor::forEachLogger(const std::function<void(std::shared_ptr<core::logging::Logger>)>& callback) {
   gsl_Expects(logger_ && python_logger_);
-  logger_->setLogCallback(callback);
-  python_logger_->setLogCallback(callback);
+  callback(logger_);
+  callback(python_logger_);
 }
 
 REGISTER_RESOURCE(ExecutePythonProcessor, Processor);
--- a/libminifi/src/minifi-c.cpp
+++ b/libminifi/src/minifi-c.cpp
@@ -1,0 +1,561 @@
+/**
+* Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "minifi-c/minifi-c.h"
+
+#include <memory>
+#include <vector>
+
+#include "agent/agent_docs.h"
+#include "core/ProcessorMetrics.h"
+#include "minifi-cpp/core/Annotation.h"
+#include "minifi-cpp/core/ClassLoader.h"
+#include "minifi-cpp/core/ProcessContext.h"
+#include "minifi-cpp/core/ProcessSession.h"
+#include "minifi-cpp/core/ProcessorApi.h"
+#include "minifi-cpp/core/ProcessorDescriptor.h"
+#include "minifi-cpp/core/ProcessorFactory.h"
+#include "minifi-cpp/core/ProcessorMetadata.h"
+#include "minifi-cpp/core/PropertyValidator.h"
+#include "minifi-cpp/core/logging/Logger.h"
+#include "minifi-cpp/core/state/PublishedMetricProvider.h"
+#include "minifi-cpp/core/state/Value.h"
+
+namespace minifi = org::apache::nifi::minifi;
+
+namespace {
+
+std::string toString(MinifiStringView sv) {
+  return std::string(sv.data, sv.length);
+}
+
+std::string_view toStringView(MinifiStringView sv) {
+  return std::string_view(sv.data, sv.length);
+}
+
+minifi::core::annotation::Input toInputRequirement(MinifiInputRequirement req) {
+  switch (req) {
+    case MINIFI_INPUT_REQUIRED: return minifi::core::annotation::Input::INPUT_REQUIRED;
+    case MINIFI_INPUT_ALLOWED: return minifi::core::annotation::Input::INPUT_ALLOWED;
+    case MINIFI_INPUT_FORBIDDEN: return minifi::core::annotation::Input::INPUT_FORBIDDEN;
+  }
+  gsl_FailFast();
+}
+
+minifi::core::logging::LOG_LEVEL toLogLevel(MinifiLogLevel lvl) {
+  switch (lvl) {
+    case MINIFI_TRACE: return minifi::core::logging::trace;
+    case MINIFI_DEBUG: return minifi::core::logging::debug;
+    case MINIFI_INFO: return minifi::core::logging::info;
+    case MINIFI_WARNING: return minifi::core::logging::warn;
+    case MINIFI_ERROR: return minifi::core::logging::err;
+    case MINIFI_CRITICAL: return minifi::core::logging::critical;
+    case MINIFI_OFF: return minifi::core::logging::off;
+  }
+  gsl_FailFast();
+}
+
+minifi::state::response::SerializedResponseNode createSerializedResponseNode(const MinifiSerializedResponseNode* node) {
+  gsl_Expects(node);
+  std::vector<minifi::state::response::SerializedResponseNode> children;
+  for (int i = 0; i < node->children_count; ++i) {
+    children.push_back(createSerializedResponseNode(&node->children_ptr[i]));
+  }
+  return minifi::state::response::SerializedResponseNode{
+    .name = toString(node->name),
+    .value = {},
+    .array = static_cast<bool>(node->array),
+    .collapsible = static_cast<bool>(node->collapsible),
+    .keep_empty = static_cast<bool>(node->keep_empty),
+    .children = std::move(children)
+  };
+}
+
+minifi::core::Property createProperty(const MinifiProperty* property_description) {
+  gsl_Expects(property_description);
+  std::vector<std::string_view> allowed_values;
+  for (int i = 0; i < property_description->allowed_values_count; ++i) {
+    allowed_values.push_back(toStringView(property_description->allowed_values_ptr[i]));
+  }
+  std::vector<std::string_view> allowed_types;
+  for (int i = 0; i < property_description->types_count; ++i) {
+    allowed_types.push_back(toStringView(property_description->types_ptr[i]));
+  }
+  std::vector<std::string_view> dependent_properties;
+  for (int i = 0; i < property_description->dependent_properties_count; ++i) {
+    dependent_properties.push_back(toStringView(property_description->dependent_properties_ptr[i]));
+  }
+  std::vector<std::pair<std::string_view, std::string_view>> exclusive_of_properties;
+  for (int i = 0; i < property_description->exclusive_of_properties_count; ++i) {
+    exclusive_of_properties.push_back({toStringView(property_description->exclusive_of_property_names_ptr[i]), toStringView(property_description->exclusive_of_property_values_ptr[i])});
+  }
+  std::optional<std::string_view> default_value;
+  if (property_description->default_value) {
+    default_value = toStringView(*property_description->default_value);
+  }
+  return minifi::core::Property{minifi::core::PropertyReference{
+    toStringView(property_description->name),
+    toStringView(property_description->display_name),
+    toStringView(property_description->description),
+    static_cast<bool>(property_description->is_required),
+    static_cast<bool>(property_description->is_sensitive),
+    std::span(allowed_values),
+    std::span(allowed_types),
+    std::span(dependent_properties),
+    std::span(exclusive_of_properties),
+    default_value,
+    gsl::make_not_null((const minifi::core::PropertyValidator*)(void*)property_description->validator),
+    static_cast<bool>(property_description->supports_expression_language)
+  }};
+}
+
+class CProcessor;
+
+class CProcessorMetricsWrapper : public minifi::core::ProcessorMetricsImpl {
+ public:
+  class CProcessorInfoProvider : public ProcessorMetricsImpl::ProcessorInfoProvider {
+   public:
+    CProcessorInfoProvider(const CProcessor& source_processor): source_processor_(source_processor) {}
+
+    std::string getProcessorType() const override;
+    std::string getName() const override;
+    minifi::utils::SmallString<36> getUUIDStr() const override;
+
+    ~CProcessorInfoProvider() override = default;
+
+   private:
+    const CProcessor& source_processor_;
+  };
+
+  explicit CProcessorMetricsWrapper(const CProcessor& source_processor)
+      : minifi::core::ProcessorMetricsImpl(std::make_unique<CProcessorInfoProvider>(source_processor)),
+        source_processor_(source_processor) {
+  }
+
+  std::vector<minifi::state::response::SerializedResponseNode> serialize() override;
+
+  std::vector<minifi::state::PublishedMetric> calculateMetrics() override;
+
+ private:
+  const CProcessor& source_processor_;
+};
+
+class CProcessor : public minifi::core::ProcessorApi {
+ public:
+  CProcessor(MinifiProcessorCallbacks callbacks, minifi::core::ProcessorMetadata metadata)
+      : callbacks_(callbacks),
+        metrics_(std::make_shared<CProcessorMetricsWrapper>(*this)) {
+    metadata_ = metadata;
+    MinifiProcessorMetadata c_metadata;
+    auto uuid_str = metadata.uuid.to_string();
+    c_metadata.uuid = MinifiStringView{.data = uuid_str.data(), .length = gsl::narrow<uint32_t>(uuid_str.length())};
+    c_metadata.name = MinifiStringView{.data = metadata.name.data(), .length = gsl::narrow<uint32_t>(metadata.name.length())};
+    c_metadata.logger = (MinifiLogger)(void*)&metadata_.logger;
+    impl_ = callbacks_.create(c_metadata);
+  }
+  ~CProcessor() override {
+    callbacks_.destroy(impl_);
+  }
+
+  bool isWorkAvailable() override {
+    return static_cast<bool>(callbacks_.isWorkAvailable(impl_));
+  }
+
+  void restore(const std::shared_ptr<minifi::core::FlowFile>& file) override {
+    callbacks_.restore(impl_, (MinifiFlowFile)(void*)&file);
+  }
+
+  bool supportsDynamicProperties() const override {
+    return static_cast<bool>(callbacks_.supportsDynamicProperties(impl_));
+  }
+
+  bool supportsDynamicRelationships() const override {
+    return static_cast<bool>(callbacks_.supportsDynamicRelationships(impl_));
+  }
+
+  void initialize(minifi::core::ProcessorDescriptor& descriptor) override {
+    callbacks_.initialize(impl_, (MinifiProcessorDescriptor)(void*)(&descriptor));
+  }
+
+  bool isSingleThreaded() const override {
+    return static_cast<bool>(callbacks_.isSingleThreaded(impl_));
+  }
+
+  std::string getProcessorType() const override {
+    std::string result;
+    callbacks_.getProcessorType(impl_, [] (void* data, MinifiStringView sv) {
+      *reinterpret_cast<std::string*>(data) = std::string{sv.data, sv.length};
+    }, (void*)&result);
+    return result;
+  }
+
+  bool getTriggerWhenEmpty() const override {
+    return static_cast<bool>(callbacks_.getTriggerWhenEmpty(impl_));
+  }
+
+  void onTrigger(minifi::core::ProcessContext& process_context, minifi::core::ProcessSession& process_session) override {
+    callbacks_.onTrigger(impl_, (MinifiProcessContext)(void*)(&process_context), (MinifiProcessSession)(void*)(&process_session));
+  }
+
+  void onSchedule(minifi::core::ProcessContext& process_context, minifi::core::ProcessSessionFactory& process_session_factory) override {
+    callbacks_.onSchedule(impl_, (MinifiProcessContext)(void*)(&process_context), (MinifiProcessSessionFactory)(void*)(&process_session_factory));
+  }
+
+  void onUnSchedule() override {
+    callbacks_.onUnSchedule(impl_);
+  }
+
+  void notifyStop() override {
+    callbacks_.notifyStop(impl_);
+  }
+
+  minifi::core::annotation::Input getInputRequirement() const override {
+    return toInputRequirement(callbacks_.getInputRequirement(impl_));
+  }
+
+  gsl::not_null<std::shared_ptr<minifi::core::ProcessorMetrics>> getMetrics() const override {
+    return metrics_;
+  }
+
+  void forEachLogger(const std::function<void(std::shared_ptr<minifi::core::logging::Logger>)>& callback) override {
+    callbacks_.forEachLogger(impl_, (MinifiLoggerCallback)(void*)(&callback));
+  }
+
+  std::string getName() const {
+    return metadata_.name;
+  }
+
+  minifi::utils::Identifier getUUID() const {
+    return metadata_.uuid;
+  }
+
+  void serializeMetrics(std::vector<minifi::state::response::SerializedResponseNode>& resp) const {
+    callbacks_.serializeMetrics(impl_, (MinifiSerializedResponseNodeVec)(void*)&resp);
+  }
+
+  void calculateMetrics(std::vector<minifi::state::PublishedMetric>& resp) const {
+    callbacks_.calculateMetrics(impl_, (MinifiPublishedMetricVec)(void*)&resp);
+  }
+
+ private:
+  MinifiProcessorCallbacks callbacks_;
+  void* impl_;
+  minifi::core::ProcessorMetadata metadata_;
+  gsl::not_null<std::shared_ptr<minifi::core::ProcessorMetrics>> metrics_;
+};
+
+class CProcessorFactory : public minifi::core::ProcessorFactory {
+ public:
+  CProcessorFactory(std::string group_name, std::string class_name, MinifiProcessorCallbacks callbacks)
+    : group_name_(std::move(group_name)),
+      class_name_(std::move(class_name)),
+      callbacks_(callbacks) {}
+  std::unique_ptr<minifi::core::ProcessorApi> create(minifi::core::ProcessorMetadata metadata) override {
+    return std::make_unique<CProcessor>(callbacks_, metadata);
+  }
+
+  std::string getGroupName() const override {
+    return group_name_;
+  }
+
+  std::string getClassName() const override {
+    return class_name_;
+  }
+
+  ~CProcessorFactory() override = default;
+
+ private:
+  MinifiProcessorCallbacks callbacks_;
+  std::string group_name_;
+  std::string class_name_;
+};
+
+std::string CProcessorMetricsWrapper::CProcessorInfoProvider::getProcessorType() const {
+  return source_processor_.getProcessorType();
+}
+std::string CProcessorMetricsWrapper::CProcessorInfoProvider::getName() const {
+  return source_processor_.getName();
+}
+minifi::utils::SmallString<36> CProcessorMetricsWrapper::CProcessorInfoProvider::getUUIDStr() const {
+  return source_processor_.getUUID().to_string();
+}
+
+std::vector<minifi::state::response::SerializedResponseNode> CProcessorMetricsWrapper::serialize() override {
+  auto nodes = ProcessorMetricsImpl::serialize();
+  source_processor_.serializeMetrics(nodes);
+  return nodes;
+}
+
+std::vector<minifi::state::PublishedMetric> CProcessorMetricsWrapper::calculateMetrics() override {
+  auto nodes = ProcessorMetricsImpl::calculateMetrics();
+  source_processor_.calculateMetrics(nodes);
+  return nodes;
+}
+
+}  // namespace
+
+extern "C" {
+
+MinifiPropertyValidator MinifiGetStandardValidator(MinifiStandardPropertyValidator validator) {
+  switch (validator) {
+    case MINIFI_ALWAYS_VALID_VALIDATOR: return (MinifiPropertyValidator)(void*)(const minifi::core::PropertyValidator*)&minifi::core::StandardPropertyValidators::ALWAYS_VALID_VALIDATOR;
+    case MINIFI_NON_BLANK_VALIDATOR: return (MinifiPropertyValidator)(void*)(const minifi::core::PropertyValidator*)&minifi::core::StandardPropertyValidators::NON_BLANK_VALIDATOR;
+    case MINIFI_TIME_PERIOD_VALIDATOR: return (MinifiPropertyValidator)(void*)(const minifi::core::PropertyValidator*)&minifi::core::StandardPropertyValidators::TIME_PERIOD_VALIDATOR;
+    case MINIFI_BOOLEAN_VALIDATOR: return (MinifiPropertyValidator)(void*)(const minifi::core::PropertyValidator*)&minifi::core::StandardPropertyValidators::BOOLEAN_VALIDATOR;
+    case MINIFI_INTEGER_VALIDATOR: return (MinifiPropertyValidator)(void*)(const minifi::core::PropertyValidator*)&minifi::core::StandardPropertyValidators::INTEGER_VALIDATOR;
+    case MINIFI_UNSIGNED_INTEGER_VALIDATOR: return (MinifiPropertyValidator)(void*)(const minifi::core::PropertyValidator*)&minifi::core::StandardPropertyValidators::UNSIGNED_INTEGER_VALIDATOR;
+    case MINIFI_DATA_SIZE_VALIDATOR: return (MinifiPropertyValidator)(void*)(const minifi::core::PropertyValidator*)&minifi::core::StandardPropertyValidators::DATA_SIZE_VALIDATOR;
+    case MINIFI_PORT_VALIDATOR: return (MinifiPropertyValidator)(void*)(const minifi::core::PropertyValidator*)&minifi::core::StandardPropertyValidators::PORT_VALIDATOR;
+    default: gsl_FailFast();
+  }
+}
+
+MinifiPropertyValidator MinifiCreatePropertyValidator(const MinifiPropertyValidatorCreateInfo*) {
+  gsl_FailFast();
+}
+
+void MinifiSerializedResponseNodeVecPush(MinifiSerializedResponseNodeVec nodes, const MinifiSerializedResponseNode* node) {
+  ((std::vector<minifi::state::response::SerializedResponseNode>*)(void*)(nodes))->push_back(createSerializedResponseNode(node));
+}
+
+void MinifiPublishedMetricVecPush(MinifiPublishedMetricVec nodes, MinifiPublishedMetric node) {
+  gsl_Assert(nodes != MINIFI_NULL);
+  minifi::state::PublishedMetric metric{
+    .name = toString(node.name),
+    .value = node.value
+  };
+  for (int i = 0; i < node.labels_count; ++i) {
+    metric.labels[toString(node.label_keys[i])] = toString(node.label_values[i]);
+  }
+  ((std::vector<minifi::state::PublishedMetric>*)(void*)(nodes))->push_back(std::move(metric));
+}
+
+void MinifiRegisterProcessorClass(const MinifiProcessorClassDescription* class_description) {
+  gsl_Expects(class_description);
+  std::vector<minifi::core::Property> properties;
+  for (int i = 0; i < class_description->class_properties_count; ++i) {
+    properties.push_back(createProperty(&class_description->class_properties_ptr[i]));
+  }
+  std::vector<const minifi::core::DynamicProperty> dynamic_properties;
+  for (int i = 0; i < class_description->dynamic_properties_count; ++i) {
+    dynamic_properties.push_back(minifi::core::DynamicProperty{
+      .name = toStringView(class_description->dynamic_properties_ptr[i].name),
+      .value = toStringView(class_description->dynamic_properties_ptr[i].value),
+      .description = toStringView(class_description->dynamic_properties_ptr[i].description),
+      .supports_expression_language = static_cast<bool>(class_description->dynamic_properties_ptr[i].supports_expression_language)
+    });
+  }
+  std::vector<minifi::core::Relationship> relationships;
+  for (int i = 0; i < class_description->class_relationships_count; ++i) {
+    relationships.push_back(minifi::core::Relationship{
+      toString(class_description->class_relationships_ptr[i].name),
+      toString(class_description->class_relationships_ptr[i].description)
+    });
+  }
+  std::vector<std::vector<const minifi::core::RelationshipDefinition>> output_attribute_relationships;
+  std::vector<const minifi::core::OutputAttributeReference> output_attributes;
+  for (int i = 0; i < class_description->output_attributes_count; ++i) {
+    minifi::core::OutputAttributeReference ref{minifi::core::OutputAttributeDefinition{"", {}, ""}};
+    ref.name = toString(class_description->output_attributes_ptr[i].name);
+    ref.description = toStringView(class_description->output_attributes_ptr[i].description);
+    output_attribute_relationships.push_back({});
+    for (int j = 0; j < class_description->output_attributes_ptr[i].relationships_count; ++j) {
+      output_attribute_relationships.back().push_back(minifi::core::RelationshipDefinition{
+        .name = toStringView(class_description->output_attributes_ptr[i].relationships_ptr[j].name),
+        .description = toStringView(class_description->output_attributes_ptr[i].relationships_ptr[j].description)
+      });
+    }
+    ref.relationships = std::span(output_attribute_relationships.back());
+    output_attributes.push_back(ref);
+  }
+  auto module_name = toString(class_description->module_name);
+  minifi::AgentDocs::getMutableClassDescriptions()[module_name].processors_.push_back(minifi::ClassDescription{
+    .type_ = minifi::ResourceType::Processor,
+    .short_name_ = toString(class_description->short_name),
+    .full_name_ = toString(class_description->full_name),
+    .description_ = toString(class_description->description),
+    .class_properties_ = properties,
+    .dynamic_properties_ = dynamic_properties,
+    .class_relationships_ = relationships,
+    .output_attributes_ = output_attributes,
+    .supports_dynamic_properties_ = static_cast<bool>(class_description->supports_dynamic_properties),
+    .supports_dynamic_relationships_ = static_cast<bool>(class_description->supports_dynamic_relationships),
+    .inputRequirement_ = minifi::core::annotation::toString(toInputRequirement(class_description->input_requirement)),
+    .isSingleThreaded_ = static_cast<bool>(class_description->is_single_threaded),
+  });
+
+  minifi::core::ClassLoader::getDefaultClassLoader().getClassLoader(module_name).registerClass(
+    toString(class_description->short_name),
+    std::make_unique<CProcessorFactory>(module_name, toString(class_description->internal_name), class_description->callbacks)
+  );
+}
+
+void MinifiMinifiLoggerCallbackCall(MinifiLoggerCallback cb, MinifiLogger logger) {
+  gsl_Assert(logger != MINIFI_NULL);
+  (*((std::function<void(std::shared_ptr<minifi::core::logging::Logger>)>*)(void*)cb))(*(std::shared_ptr<minifi::core::logging::Logger>*)(void*)logger);
+}
+
+void MinifiProcessorDescriptorSetSupportedRelationships(MinifiProcessorDescriptor descriptor, uint32_t relationships_count, const MinifiRelationship* relationships_ptr) {
+  gsl_Assert(descriptor != MINIFI_NULL);
+  std::vector<minifi::core::RelationshipDefinition> relationships;
+  for (int i = 0; i < relationships_count; ++i) {
+    relationships.push_back(minifi::core::RelationshipDefinition{
+      .name = toString(relationships_ptr[i].name),
+      .description = toString(relationships_ptr[i].description),
+    });
+  }
+  ((minifi::core::ProcessorDescriptor*)(void*)descriptor)->setSupportedRelationships(relationships);
+}
+
+void MinifiProcessorDescriptorSetSupportedProperties(MinifiProcessorDescriptor descriptor, uint32_t properties_count, const MinifiProperty* properties_ptr) {
+  gsl_Assert(descriptor != MINIFI_NULL);
+  std::vector<minifi::core::Property> properties;
+  std::vector<minifi::core::PropertyReference> property_references;
+  for (int i = 0; i < properties_count; ++i) {
+    properties.push_back(createProperty(&properties_ptr[i]));
+    property_references.push_back(properties.back().getReference());
+  }
+  ((minifi::core::ProcessorDescriptor*)(void*)descriptor)->setSupportedProperties(property_references);
+}
+
+void MinifiProcessContextGetProperty(MinifiProcessContext context, MinifiStringView property_name, MinifiFlowFile flow_file, void(*result_cb)(void* data, MinifiStringView result), void(*error_cb)(void* data, MinifiStringView error), void* data) {
+  gsl_Assert(context != MINIFI_NULL);
+  auto result = ((minifi::core::ProcessContext*)(void*)context)->getProperty(toStringView(property_name), flow_file != MINIFI_NULL ? ((std::shared_ptr<minifi::core::FlowFile>*)(void*)flow_file)->get() : nullptr);
+  if (result) {
+    result_cb(data, MinifiStringView{
+      .data = result.value().data(),
+      .length = gsl::narrow<uint32_t>(result.value().length())
+    });
+  } else {
+    auto error_msg = result.error().message();
+    error_cb(data, MinifiStringView{
+      .data = error_msg.data(),
+      .length = gsl::narrow<uint32_t>(error_msg.length())
+    });
+  }
+}
+
+void MinifiLoggerSetMaxLogSize(MinifiLogger logger, int32_t max_size) {
+  gsl_Assert(logger != MINIFI_NULL);
+  (*((std::shared_ptr<minifi::core::logging::Logger>*)(void*)logger))->set_max_log_size(max_size);
+}
+
+void MinifiLoggerGetId(MinifiLogger logger, void(*cb)(void* data, MinifiStringView), void* data) {
+  gsl_Assert(logger != MINIFI_NULL);
+  auto id = (*((std::shared_ptr<minifi::core::logging::Logger>*)(void*)logger))->get_id();
+  if (id) {
+    cb(data, MinifiStringView{
+      .data = id.value().data(),
+      .length = gsl::narrow<uint32_t>(id.value().length())
+    });
+  }
+}
+
+void MinifiLoggerLogString(MinifiLogger logger, MinifiLogLevel level, MinifiStringView msg) {
+  gsl_Assert(logger != MINIFI_NULL);
+  (*((std::shared_ptr<minifi::core::logging::Logger>*)(void*)logger))->log_string(toLogLevel(level), toString(msg));
+}
+
+MinifiBool MinifiLoggerShouldLog(MinifiLogger logger, MinifiLogLevel level) {
+  gsl_Assert(logger != MINIFI_NULL);
+  return (*((std::shared_ptr<minifi::core::logging::Logger>*)(void*)logger))->should_log(toLogLevel(level));
+}
+
+MinifiLogLevel MinifiLoggerLevel(MinifiLogger logger) {
+  gsl_Assert(logger != MINIFI_NULL);
+  switch ((*((std::shared_ptr<minifi::core::logging::Logger>*)(void*)logger))->level()) {
+    case minifi::core::logging::LOG_LEVEL::trace: return MINIFI_TRACE;
+    case minifi::core::logging::LOG_LEVEL::debug: return MINIFI_DEBUG;
+    case minifi::core::logging::LOG_LEVEL::info: return MINIFI_INFO;
+    case minifi::core::logging::LOG_LEVEL::warn: return MINIFI_WARNING;
+    case minifi::core::logging::LOG_LEVEL::err: return MINIFI_ERROR;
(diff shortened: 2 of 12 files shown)


core-framework/c-api-framework/include/api/core/ProcessContext.h
```diff
@@ -0,0 +1,47 @@
...
+ public:
+  explicit ProcessContext(MinifiProcessContext impl): impl_(impl) {}
+
+  nonstd::expected<std::string, std::error_code> getProperty(std::string_view name, const FlowFile* flow_file = nullptr) const;
+  nonstd::expected<std::string, std::error_code> getProperty(const minifi::core::PropertyReference& property_reference, const FlowFile* flow_file = nullptr) const {
+    return getProperty(property_reference.name, flow_file);
+  }
+
+  bool hasNonEmptyProperty(std::string_view name) const;
+  void yield();
```
martinzink: I mentioned this already (and i cant recall what your answer was), but what if we were to remove yield from ProcessContext and moved yield functionality into onTrigger and onSchedule instead?

This way ProcessContext would only have const functions and yield feels like a result of ontrigger in my eyes.

core-framework/c-api-framework/include/api/core/ProcessorImpl.h
```diff
@@ -0,0 +1,110 @@
...
+#include <string>
+#include <string_view>
+#include <unordered_set>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+#include "minifi-cpp/core/Annotation.h"
+#include "minifi-cpp/core/DynamicProperty.h"
+// #include "minifi-cpp/core/ProcessorMetrics.h"
```
martinzink:
```suggestion
```

core-framework/c-api-framework/src/core/ProcessSession.cpp
```diff
@@ -0,0 +1,149 @@
...
+void ProcessSession::read(FlowFile& flow_file, const io::InputStreamCallback& callback) {
+  const auto status = MinifiProcessSessionRead(impl_, flow_file.getImpl(), [] (void* data, MinifiInputStream input) {
+    return (*static_cast<const io::InputStreamCallback*>(data))(std::make_shared<MinifiInputStreamWrapper>(input));
+  }, const_cast<io::InputStreamCallback*>(&callback));
+  if (status != MINIFI_SUCCESS) {
+    throw minifi::Exception(minifi::FILE_OPERATION_EXCEPTION, "Failed to process flowfile content");
+  }
+}
+
+void ProcessSession::setAttribute(FlowFile& ff, std::string_view key, std::string value) {
```
martinzink:
```suggestion
void ProcessSession::setAttribute(FlowFile& ff, std::string_view key, std::string value) {  // NOLINT(performance-unnecessary-value-param)
```

This is the last clang tidy issue, I think we can ignore this since the signature comes from the parent not much we can do about it here

martinzink (commented):
Since #2029 is already merged could you please rebase and retarget this PR to main?

core-framework/c-api-framework/include/api/core/ProcessContext.h
adamdebreceni: yes, I remember reaching the conclusion that it is a good idea having the "yield" be part of the return value of onTrigger, I'll implement the change

core-framework/common/src/utils/Id.cpp
```diff
@@ -0,0 +1,133 @@
...
+}
+
+bool Identifier::parseByte(Data &data, const uint8_t *input, int &charIdx, int &byteIdx) {
+  uint8_t upper = 0;
+  uint8_t lower = 0;
+  if (!string::from_hex(input[charIdx], upper)
+      || !string::from_hex(input[charIdx + 1], lower)) {
+    return false;
+  }
+  charIdx += 2;
```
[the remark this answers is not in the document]
martinzink: It was changed due to clang-tidy's nagging, and the functionality didnt changed instead of charIdx++, charIdx++, its now charIdx, charIdx+1 and charIdx += 2 (functionally the same)

minifi-api/include/minifi-c/minifi-c.h
```diff
@@ -0,0 +1,228 @@
...
+
+#define MINIFI_API_MAJOR_VERSION 0
+#define MINIFI_API_MINOR_VERSION 1
+#define MINIFI_API_PATCH_VERSION 0
+#define MINIFI_API_VERSION STRINGIFY(MINIFI_API_MAJOR_VERSION) "." STRINGIFY(MINIFI_API_MINOR_VERSION) "." STRINGIFY(MINIFI_API_PATCH_VERSION)
+#define MINIFI_API_VERSION_TAG "MINIFI_API_VERSION=[" MINIFI_API_VERSION "]"
+#define MINIFI_NULL 0
+#define OWNED
+
+typedef uint32_t MinifiBool;
```
szaszm: We could typedef `_Bool` or `bool` (if C23) as MinifiBool instead.

minifi-api/include/minifi-c/minifi-c.h
```diff
@@ -0,0 +1,228 @@
...
+MINIFI_DECLARE_CONST_HANDLE(MinifiPropertyValidator);
+MINIFI_DECLARE_HANDLE(MinifiFlowFile);
+MINIFI_DECLARE_HANDLE(MinifiLogger);
+MINIFI_DECLARE_HANDLE(MinifiProcessContext);
+MINIFI_DECLARE_HANDLE(MinifiProcessSession);
+MINIFI_DECLARE_HANDLE(MinifiInputStream);
+MINIFI_DECLARE_HANDLE(MinifiOutputStream);
+MINIFI_DECLARE_HANDLE(MinifiConfigure);
+MINIFI_DECLARE_HANDLE(MinifiExtension);
+MINIFI_DECLARE_HANDLE(MinifiPublishedMetrics);
```
szaszm: I would expose these as pointers to a forward-declared/incomplete struct, and omit the semi-transparent "handle" layer.

minifi-api/include/minifi-c/minifi-c.h
```diff
@@ -0,0 +1,228 @@
...
+  MinifiStringView name;
+  MinifiStringView display_name;
+  MinifiStringView description;
+  MinifiBool is_required;
+  MinifiBool is_sensitive;
+  uint32_t dependent_properties_count;
+  const MinifiStringView* dependent_properties_ptr;
+  uint32_t exclusive_of_properties_count;
+  const MinifiStringView* exclusive_of_property_names_ptr;
+  const MinifiStringView* exclusive_of_property_values_ptr;
```
szaszm: I think we should remove these from the API and all of their usages. They're never really enforced, and if we later want to have this functionality, it could be added back in a different form later, but I would not complicate the interface of this type with it.

minifi-api/include/minifi-c/minifi-c.h
```diff
@@ -0,0 +1,228 @@
...
+
+  const MinifiStringView* default_value;
+  uint32_t allowed_values_count;
+  const MinifiStringView* allowed_values_ptr;
+  MinifiPropertyValidator validator;
+
+  uint32_t types_count;
+  const MinifiStringView* types_ptr;
```
szaszm: I would replace this with a single optional type member, so 0 or 1 controller service types.

extensions/python/types/PyScriptFlowFile.cpp
```diff
@@ -135,7 +135,8 @@ PyObject* PyScriptFlowFile::setAttribute(PyScriptFlowFile* self, PyObject* args)
     return nullptr;
   }
 
-  return object::returnReference(flow_file->setAttribute(key, value));
+  flow_file->setAttribute(key, value);
```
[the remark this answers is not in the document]
adamdebreceni: the return value wasn't indicative of any error and was unused

minifi-api/include/minifi-c/minifi-c.h
```diff
@@ -0,0 +1,228 @@
...
+} MinifiProcessorCallbacks;
+
+typedef struct MinifiProcessorClassDescription {
+  MinifiStringView module_name;
+  MinifiStringView full_name;  // '::'-delimited fully qualified name e.g. 'org::apache::nifi::minifi::GenerateFlowFile'
+  MinifiStringView description;
+  uint32_t class_properties_count;
+  const MinifiProperty* class_properties_ptr;
+  uint32_t dynamic_properties_count;
+  const MinifiDynamicProperty* dynamic_properties_ptr;
```
szaszm: I think we're mixing description and runtime values here. In a processor description/manifest, dynamic properties are only represented as supported/unsupported, while their values are a runtime information. I would remove these members.

minifi-api/include/minifi-c/minifi-c.h
```diff
@@ -0,0 +1,228 @@
...
+
+typedef enum MinifiStatus {
+  MINIFI_SUCCESS = 0,
+  MINIFI_UNKNOWN_ERROR = 1,
+  MINIFI_NOT_SUPPORTED_PROPERTY = 2,
+  MINIFI_DYNAMIC_PROPERTIES_NOT_SUPPORTED = 3,
+  MINIFI_PROPERTY_NOT_SET = 4,
+  MINIFI_VALIDATION_FAILED = 5,
+  MINIFI_PROCESSOR_YIELD = 6
+} MinifiStatus;
```
szaszm, line 99: I'd add the full name of the enum as a prefix to the values, i.e. MINIFI_STATUS_SUCCESS instead of MINIFI_SUCCESS

```suggestion
typedef enum MinifiStatus {
  MINIFI_STATUS_SUCCESS = 0,
  MINIFI_STATUS_UNKNOWN_ERROR = 1,
  MINIFI_STATUS_NOT_SUPPORTED_PROPERTY = 2,
  MINIFI_STATUS_DYNAMIC_PROPERTIES_NOT_SUPPORTED = 3,
  MINIFI_STATUS_PROPERTY_NOT_SET = 4,
  MINIFI_STATUS_VALIDATION_FAILED = 5,
  MINIFI_STATUS_PROCESSOR_YIELD = 6
} MinifiStatus;
```

minifi-api/include/minifi-c/minifi-c.h
```diff
@@ -0,0 +1,228 @@
...
+  const MinifiOutputAttribute* output_attributes_ptr;
+  MinifiBool supports_dynamic_properties;
+  MinifiBool supports_dynamic_relationships;
+  MinifiInputRequirement input_requirement;
+  MinifiBool is_single_threaded;
+
+  MinifiProcessorCallbacks callbacks;
+} MinifiProcessorClassDescription;
+
+typedef enum MinifiStandardPropertyValidator {
```
szaszm: I'd remove "standard" and maybe "property" from the name for simplicity: there are only "standard" validators, and they always work on properties. I'd also call the values MINIFI_VALIDATOR_FOO.

minifi-api/include/minifi-c/minifi-c.h
```diff
@@ -0,0 +1,228 @@
...
+void MinifiProcessSessionRemove(MinifiProcessSession, MinifiFlowFile);
+MinifiStatus MinifiProcessSessionRead(MinifiProcessSession, MinifiFlowFile, int64_t(*cb)(void* user_ctx, MinifiInputStream), void* user_ctx);
+MinifiStatus MinifiProcessSessionWrite(MinifiProcessSession, MinifiFlowFile, int64_t(*cb)(void* user_ctx, MinifiOutputStream), void* user_ctx);
+
+void MinifiConfigureGet(MinifiConfigure, MinifiStringView, void(*cb)(void*, MinifiStringView), void*);
+
+uint64_t MinifiInputStreamSize(MinifiInputStream);
+
+int64_t MinifiInputStreamRead(MinifiInputStream, char*, uint64_t);
+int64_t MinifiOutputStreamWrite(MinifiOutputStream, const char*, uint64_t);
```
szaszm: I would name the second and third parameters, and in general those with no obvious meaning.

szaszm (commented):
early API comments

core-framework/c-api-framework/include/api/core/PublishedMetrics.h
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+#pragma once
+
+#include <vector>
+#include <string>
+
+namespace org::apache::nifi::minifi::api::core {
+
+using PublishedMetrics = std::vector<std::pair<std::string, double>>;
```
szaszm: YES thank you, the overcomplicated structure of metrics types has been bothering me ever since they were introduced.
suggestion: use a simple aggregate struct instead of a std::pair.

minifi-api/include/minifi-c/minifi-c.h
```diff
@@ -0,0 +1,228 @@
...
+
+typedef enum MinifiInputRequirement {
+  MINIFI_INPUT_REQUIRED = 0,
+  MINIFI_INPUT_ALLOWED = 1,
+  MINIFI_INPUT_FORBIDDEN = 2
+} MinifiInputRequirement;
+
+typedef struct MinifiStringView {
+  const char* data;
+  uint32_t length;
```
szaszm: this should be size_t IMO
```suggestion
  size_t length;
```

core-framework/c-api-framework/include/api/core/logging/Logger.h
```diff
@@ -0,0 +1,57 @@
...
+class Logger : public minifi::core::logging::Logger {
+ public:
+  explicit Logger(MinifiLogger impl): impl_(impl) {}
+
+  void set_max_log_size(int size) override;
+  std::optional<std::string> get_id() override;
+  void log_string(minifi::core::logging::LOG_LEVEL level, std::string str) override;
+  bool should_log(minifi::core::logging::LOG_LEVEL level) override;
+  [[nodiscard]] minifi::core::logging::LOG_LEVEL level() const override;
+  int getMaxLogSize() override;
```
szaszm: does all this need to be on the api?

core-framework/c-api-framework/include/api/core/FlowFile.h
```diff
@@ -0,0 +1,47 @@
...
+#include <array>
+#include <string_view>
+#include "minifi-c.h"
+#include "minifi-cpp/core/SpecialFlowAttribute.h"
+
+namespace org::apache::nifi::minifi::api::core {
+
+class FlowFile {
+ public:
+  explicit FlowFile(OWNED MinifiFlowFile impl): impl_(impl) {}
```
szaszm: since this is C++, we could replace OWNED with gsl::owner<T*> and use unique_ptr internally.

libminifi/src/minifi-c.cpp
```diff
@@ -0,0 +1,502 @@
...
+    return static_cast<bool>(initializer_(user_data_, reinterpret_cast<MinifiConfigure>(config.get())));
+  }
+
+  [[nodiscard]] const std::string& getName() const override {
+    return name_;
+  }
+
+ private:
+  std::string name_;
+  MinifiBool(*initializer_)(void*, MinifiConfigure);
```
szaszm: I would introduce type aliases for most function pointer types on the API, especially if they're used multiple times.

core-framework/include/core/Relationship.h
```diff
@@ -1,5 +1,5 @@
 /**
- * Licensed to the Apache Software Foundation (ASF) under one or more
+* Licensed to the Apache Software Foundation (ASF) under one or more
```
lordgamez: Nitpick, identation change

core-framework/common/include/utils/gsl.h
lordgamez, line 1: Is this file needed? Can't we just include "minifi-cpp/utils/gsl.h" where it is needed?

core-framework/c-api-framework/include/api/utils/ProcessorConfigUtils.h
```diff
@@ -0,0 +1,196 @@
...
+//     return std::nullopt;
+//   }
+//
+//   return typed_controller_service;
+// }
+//
+// template<typename ControllerServiceType>
+// std::shared_ptr<ControllerServiceType> parseControllerService(const core::ProcessContext& context, const core::PropertyReference& prop, const utils::Identifier& processor_uuid) {
+//   return parseOptionalControllerService<ControllerServiceType>(context, prop, processor_uuid) | utils::orThrow("Required Controller Service");
+// }
```
lordgamez: Were these commented block left in there by mistake?

core-framework/c-api-framework/src/core/ProcessorImpl.cpp
```diff
@@ -0,0 +1,87 @@
+/**
+ * @file Processor.cpp
+ * Processor class implementation
```
lordgamez: This part can be removed, it does not match with the current filename anyway.

core-framework/c-api-framework/src/core/ProcessorImpl.cpp
```diff
@@ -0,0 +1,87 @@
...
+#include "fmt/format.h"
+
+using namespace std::literals::chrono_literals;
+
+namespace org::apache::nifi::minifi::api::core {
+
+ProcessorImpl::ProcessorImpl(minifi::core::ProcessorMetadata metadata)
+    : metadata_(std::move(metadata)),
+      trigger_when_empty_(false),
+      // metrics_(std::make_shared<ProcessorMetricsImpl>(*this)),
```
lordgamez: Comment can be removed

extensions/llamacpp/processors/DefaultLlamaContext.cpp
```diff
@@ -88,7 +87,7 @@ std::optional<std::string> DefaultLlamaContext::applyTemplate(const std::vector<
   std::transform(messages.begin(), messages.end(), std::back_inserter(llama_messages),
                  [](const LlamaChatMessage& msg) { return llama_chat_message{.role = msg.role.c_str(), .content = msg.content.c_str()}; });
   std::string text;
-  text.resize(utils::configuration::DEFAULT_BUFFER_SIZE);
+  text.resize(4096);
```
lordgamez: Why was the default buffer size constant removed here?

extensions/llamacpp/tests/RunLlamaCppInferenceTests.cpp
```diff
@@ -15,12 +15,29 @@
...
+#include "unit/Catch.h"
+#include "unit/SingleProcessorTestController.h"
+#include "unit/TestBase.h"
+#include "utils/CProcessor.h"
+#include "unit/TestUtils.h"
+
+namespace org::apache::nifi::minifi::api::test::utils {
+
+template<typename T, typename ...Args>
+std::unique_ptr<minifi::core::Processor> make_custom_processor(minifi::core::ProcessorMetadata metadata, Args&&... args) {  // NOLINT(cppcoreguidelines-missing-std-forward)
```
lordgamez: Should this be moved to a common utils file? I suppose this will be reused in the future for other processor tests too.

extensions/python/ExecutePythonProcessor.cpp
```diff
@@ -38,9 +38,15 @@ namespace org::apache::nifi::minifi::extensions::python::processors {
...
-  setSupportedProperties(all_properties);
+  std::vector<core::Property> all_properties;
+  all_properties.reserve(Properties.size() + python_properties_.size());
+  for (auto& property : Properties) {
+    all_properties.emplace_back(property);
+  }
+  for (auto& python_property : python_properties_) {
+    all_properties.emplace_back(python_property);
+  }
+  setSupportedProperties(gsl::make_span(all_properties));
```
lordgamez: Properties is currently empty and we only want to include supported properties that are defined in the python processors, why was this changed?

libminifi/include/utils/CProcessor.h
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+  gsl::not_null<std::shared_ptr<minifi::core::ProcessorMetrics>> getMetrics() const override {
+    return metrics_;
+  }
+
+  std::vector<minifi::state::PublishedMetric> getCustomMetrics() const;
+
+  void forEachLogger(const std::function<void(std::shared_ptr<minifi::core::logging::Logger>)>&) override {
+    // pass
```
lordgamez: I suppose this comment is also unneeded

libminifi/src/core/ProcessSession.cpp
```diff
@@ -223,14 +223,15 @@ void ProcessSessionImpl::penalize(const std::shared_ptr<core::FlowFile> &flow) {
 }
 
 void ProcessSessionImpl::transfer(const std::shared_ptr<core::FlowFile>& flow, const Relationship& relationship) {
-  logger_->log_debug("Transferring {} from {} to relationship {}", flow->getUUIDStr(), process_context_->getProcessor().getName(), relationship.getName());
-  utils::Identifier uuid = flow->getUUID();
+  auto flow_file = std::dynamic_pointer_cast<FlowFile>(flow);
```
lordgamez: We should check at least with an assert that this is not null

libminifi/src/minifi-c.cpp
```diff
@@ -0,0 +1,502 @@
...
+}
+
+void MinifiRegisterProcessorClass(const MinifiProcessorClassDescription* class_description) {
+  gsl_Expects(class_description);
+
+  auto module_name = toString(class_description->module_name);
+  minifi::BundleDetails bundle{
+    .artifact = module_name,
+    .group = module_name,
+    .version = "1.0.0"
```
lordgamez: We shouldn't keep this 1.0.0 hardcoded value, could the version be part of the class description API?

extensions/python/ExecutePythonProcessor.cpp
martinzink: I think this was my change, due to clang tidy asking for reserving the vector beforehand. I wasnt aware that we are delibertly skipping the Properties it felt like a oversight given the **all_properties** name. If we only want to use python_properties we should keep the prereservation and add an explanatory comment and maybe rename all_properties?

core-framework/c-api-framework/include/api/core/ProcessorImpl.h
```diff
@@ -0,0 +1,109 @@
...
+
+class Connection;
+
+namespace core {
+
+class ProcessContext;
+class ProcessSession;
+class ProcessSessionFactory;
+
+class ProcessorImpl {
```
szaszm: Why do we need another ProcessorImpl, this time not derived from anything? What does this achieve? A classdoc explaining this would help me and other readers understand the code.

extension-framework/cpp-extension-lib/include/api/core/ProcessorImpl.h
szaszm, line 1: I would rename this component from "c-api-framework" to "cpp-extension-lib", since it seems to be a C++ library for extension developers targeting the MiNiFi C API.

core-framework/c-api-framework/include/api/core/ProcessorImpl.h
```diff
@@ -0,0 +1,109 @@
...
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+#include "minifi-cpp/core/Annotation.h"
+#include "minifi-cpp/core/DynamicProperty.h"
+#include "utils/gsl.h"
+#include "minifi-cpp/utils/Id.h"
+#include "minifi-cpp/core/OutputAttributeDefinition.h"
+#include "minifi-cpp/core/ProcessorMetadata.h"
```
szaszm: dependencies on MiNiFi C++ internals seems to defeat the purpose of this lib, but maybe I'm misunderstanding things.

minifi-api/include/minifi-c/minifi-c.h
adamdebreceni: we do validate these in `validateComponentProperties`

minifi-api/include/minifi-c/minifi-c.h
adamdebreceni: I plan to support custom property validators, they are executed flow parse-time and can short circuit a malformed configuration

adamdebreceni: I agree that there are only property validators currently, but from a `MinifiValidator` name I would have no idea what it validates

minifi-api/include/minifi-c/minifi-c.h
adamdebreceni: done

minifi-api/include/minifi-c/minifi-c.h
adamdebreceni: done

minifi-api/include/minifi-c/minifi-c.h
adamdebreceni: done

core-framework/c-api-framework/include/api/core/PublishedMetrics.h
adamdebreceni: done

minifi-api/include/minifi-c/minifi-c.h
adamdebreceni: done

core-framework/c-api-framework/include/api/core/FlowFile.h
adamdebreceni: done

core-framework/include/core/Relationship.h
adamdebreceni: done

core-framework/common/include/utils/gsl.h
adamdebreceni: done

core-framework/c-api-framework/include/api/utils/ProcessorConfigUtils.h
adamdebreceni: removed

core-framework/c-api-framework/src/core/ProcessorImpl.cpp
adamdebreceni: removed

core-framework/c-api-framework/src/core/ProcessorImpl.cpp
adamdebreceni: removed

libminifi/include/utils/CProcessor.h
adamdebreceni: removed

libminifi/src/core/ProcessSession.cpp
adamdebreceni: done

minifi-api/include/minifi-c/minifi-c.h
```diff
@@ -0,0 +1,224 @@
...
+} MinifiStringView;
+
+typedef struct MinifiRelationship {
+  MinifiStringView name;
+  MinifiStringView description;
+} MinifiRelationship;
+
+typedef struct MinifiOutputAttribute {
+  MinifiStringView name;
+  uint32_t relationships_count;
```
martinzink: Since we changed MinifiStringView::length to be size_t instead of uint32_t I think we should change every length to be size_t to be consistent, and also these are kind of vec-s and vec also return size_t from size()

adamdebreceni: done

libminifi/src/minifi-c.cpp
adamdebreceni: after the rebase the version is now correctly set from the extension

core-framework/c-api-framework/include/api/core/logging/Logger.h
adamdebreceni: removed, only kept log_string, should_log and set_max_log_size, the latter is only used in LogAttribute, we might be able to move this info into MinifiLoggerLogString as an extra parameter?

minifi-api/include/minifi-c/minifi-c.h
adamdebreceni: added parameter names

core-framework/c-api-framework/include/api/core/logging/Logger.h
adamdebreceni: I had to put level() back as AwsSdkLogger uses it, but since spdlog::logger exposes such a method as well I don't think it is a problem

core-framework/c-api-framework/include/api/core/logging/Logger.h
```diff
@@ -0,0 +1,55 @@
...
+
+#include "fmt/chrono.h"
+#include "fmt/ostream.h"
+#include "fmt/std.h"
+#include "minifi-c.h"
+#include "utils/Enum.h"
+#include "utils/GeneralUtils.h"
+#include "minifi-cpp/utils/gsl.h"
+#include "utils/SmallString.h"
+#include "minifi-cpp/core/logging/Logger.h"
```
fgerlits: This looks like a lot of includes for the code below; please remove those which are not needed.

minifi-api/include/minifi-c/minifi-c.h
```diff
@@ -24,24 +24,193 @@ extern "C" {
 
 #include <stddef.h>
 #include <stdint.h>
+#include <stdbool.h>
```
szaszm: Since C23, stdbool.h is deprecated, and `bool` is a built-in type.
```suggestion
#if __STDC_VERSION__ < 202311l
#  include <stdbool.h>
#endif  // < C23
```

minifi-api/include/minifi-c/minifi-c.h
```diff
@@ -24,24 +24,193 @@ extern "C" {
...
+#define MINIFI_API_MINOR_VERSION 1
+#define MINIFI_API_PATCH_VERSION 0
+#define MINIFI_API_VERSION STRINGIFY(MINIFI_API_MAJOR_VERSION) "." STRINGIFY(MINIFI_API_MINOR_VERSION) "." STRINGIFY(MINIFI_API_PATCH_VERSION)
+#define MINIFI_API_VERSION_TAG "MINIFI_API_VERSION=[" MINIFI_API_VERSION "]"
+#define MINIFI_NULL nullptr
+#define MINIFI_OWNED
+
+typedef bool MinifiBool;
+#define MINIFI_TRUE true
+#define MINIFI_FALSE false
```
szaszm: I don't think we need these, we can just use `true`/`false` directly: from `stdbool.h` or from language keywords since C23.

minifi-api/include/minifi-c/minifi-c.h
```diff
@@ -24,24 +24,193 @@ extern "C" {
...
+  size_t relationships_count;
+  const MinifiRelationship* relationships_ptr;
+  MinifiStringView description;
+} MinifiOutputAttribute;
+
+typedef struct MinifiDynamicProperty {
+  MinifiStringView name;
+  MinifiStringView value;
+  MinifiStringView description;
+  MinifiBool supports_expression_language;
```
fgerlits, line 78: Is `supports_expression_language` needed here?  I think all dynamic properties support expression language, and there is no way to disable expression languague evaluation on dynamic properties.

minifi-api/include/minifi-c/minifi-c.h
```diff
@@ -24,24 +24,193 @@ extern "C" {
...
+typedef enum MinifiInputRequirement {
+  MINIFI_INPUT_REQUIRED = 0,
+  MINIFI_INPUT_ALLOWED = 1,
+  MINIFI_INPUT_FORBIDDEN = 2
+} MinifiInputRequirement;
 
 typedef struct MinifiStringView {
   const char* data;
   size_t length;
 } MinifiStringView;
```
szaszm, line 57: We should have docs on all non-obvious parts. In this case, it's not obvious whether data is nullable and whether it's null-terminated. I'd document, that
- data is nullable (I'm assuming)
- length represents the number of bytes pointed-to by data
- invariant: data is null if and only if length == 0. 
- the bytes pointed-to by data may not be null-terminated, it is not a C string

```suggestion
/// Represents a non-owning read-only view to a sized, not null-terminated string.
/// invariant: `(data == NULL) == (length == 0)`, data is null if and only if length is zero.
typedef struct MinifiStringView {
  /// nullable, non-owning pointer to a not null-terminated string
  const char* data;
  
  /// the length of the buffer pointed-to by data
  size_t length;
} MinifiStringView;
```

minifi-api/include/minifi-c/minifi-c.h
```diff
@@ -24,24 +24,193 @@ extern "C" {
...
+  MinifiStringView name;
+  MinifiStringView description;
+} MinifiRelationship;
+
+typedef struct MinifiOutputAttribute {
+  MinifiStringView name;
+  size_t relationships_count;
+  const MinifiRelationship* relationships_ptr;
+  MinifiStringView description;
+} MinifiOutputAttribute;
```
szaszm: I would add -Definition to the name to match the internal C++ API and to make it clear that this is static metadata about the processor. Same with MinifiRelationship above.

minifi-api/include/minifi-c/minifi-c.h
```diff
@@ -24,24 +24,193 @@ extern "C" {
...
 
 typedef struct MinifiStringView {
   const char* data;
   size_t length;
 } MinifiStringView;
 
+typedef struct MinifiRelationship {
+  MinifiStringView name;
+  MinifiStringView description;
+} MinifiRelationship;
```
szaszm: We should separate static metadata from runtime data. The runtime data version of Relationship doesn't need description.

```suggestion
/// Represents an output relationship of a processor, part of the processor metadata
typedef struct MinifiRelationshipDefinition {
  /// Name, processors use this to transfer flow files to the connections connected to this relationship using ProcessSession transfer.
  MinifiStringView name;
  
  /// Human-readable description of what flow files get routed to this relationship. Included in C2 manifest and generated processor docs.
  MinifiStringView description;
} MinifiRelationshipDefinition;

typedef struct MinifiRelationshipName {
  MinifiStringView name;
} MinifiRelationshipName;
```

minifi-api/include/minifi-c/minifi-c.h
```diff
@@ -24,24 +24,193 @@ extern "C" {
 
 #include <stddef.h>
 #include <stdint.h>
+#include <stdbool.h>
 
-typedef struct MinifiConfig MinifiConfig;
-typedef struct MinifiExtension MinifiExtension;
+#define STRINGIFY_HELPER(X) #X
+#define STRINGIFY(X) STRINGIFY_HELPER(X)
```
szaszm: Since macros are unscoped and these are not part of the API, I'd add hard to type prefixes to discourage use outside of this header, or undef them after use.
```suggestion
// private macros, do not use outside of this header
#define MINIFI_PRIVATE_STRINGIFY_HELPER(X) #X
#define MINIFI_PRIVATE_STRINGIFY(X) MINIFI_PRIVATE_STRINGIFY_HELPER(X)
```

minifi-api/include/minifi-c/minifi-c.h
```diff
@@ -24,24 +24,193 @@ extern "C" {
...
 
+typedef struct MinifiRelationship {
+  MinifiStringView name;
+  MinifiStringView description;
+} MinifiRelationship;
+
+typedef struct MinifiOutputAttribute {
+  MinifiStringView name;
+  size_t relationships_count;
+  const MinifiRelationship* relationships_ptr;
```
szaszm: This doesn't need a full copy of all the relationship metadata: a relationship name as a unique identified would be enough here. I realize that this contradicts my previous comment about separating static metadata and runtime data, but a string view to a name seems to work both as a static reference and a runtime one.
```suggestion
  size_t relationships_count;
  const MinifiRelationshipName* relationships_ptr;
```

core-framework/c-api-framework/include/api/core/ProcessorImpl.h
adamdebreceni: it is just unfortunate naming, the minifi-api has been split into minifi-api/common and the "rest", they together constitute the minifi-api, while minifi-api/common is only used by the c-api-framework for cpp extension development using the c api

minifi-api/include/minifi-c/minifi-c.h
adamdebreceni: changed

minifi-api/include/minifi-c/minifi-c.h
adamdebreceni: removed

minifi-api/include/minifi-c/minifi-c.h
adamdebreceni: added comments

minifi-api/include/minifi-c/minifi-c.h
adamdebreceni: added comments, where description not needed, (MinifiOutputAttributeDefinition::relationships and MinifiProcessSessionTransfer) I just changed that to MinifiStringView

minifi-api/include/minifi-c/minifi-c.h
adamdebreceni: changed

minifi-api/include/minifi-c/minifi-c.h
adamdebreceni: changed it to use MinifiStringView

minifi-api/include/minifi-c/minifi-c.h
adamdebreceni: added

libminifi/src/minifi-c.cpp
adamdebreceni: since the extension loading revamp, there aren't really such function pointers anymore

minifi-api/include/minifi-c/minifi-c.h
adamdebreceni: removed

extensions/python/ExecutePythonProcessor.cpp
adamdebreceni: changed this to only use the python_properties_

extensions/llamacpp/tests/RunLlamaCppInferenceTests.cpp
adamdebreceni: moved this to unit/ProcessorUtils.h

extensions/llamacpp/processors/DefaultLlamaContext.cpp
adamdebreceni: for now I have added a constant with the same name in this file to be searchable, I think in the future we would like to use the `size_t getBufferSize(const Configure& configuration);` instead in `ConfigurationUtils.h` but it is not yet clear to me how we are going to do that, (maybe store it extension-locally on extension init, or get from context instead), the Configure interface is quite heavy, I don't think we want that to appear in its current form in the "c" extensions

libminifi/src/minifi-c.cpp
```diff
@@ -27,12 +47,205 @@ std::string toString(MinifiStringView sv) {
...
+      toString(class_description.class_relationships_ptr[i].name),
+      toString(class_description.class_relationships_ptr[i].description)
+    });
+  }
+  std::vector<std::vector<minifi::core::RelationshipDefinition>> output_attribute_relationships;
+  std::vector<minifi::core::OutputAttributeReference> output_attributes;
+  for (size_t attribute_idx = 0; attribute_idx < class_description.output_attributes_count; ++attribute_idx) {
+    minifi::core::OutputAttributeReference ref{minifi::core::OutputAttributeDefinition{"", {}, ""}};
+    ref.name = toStringView(class_description.output_attributes_ptr[attribute_idx].name);
+    ref.description = toStringView(class_description.output_attributes_ptr[attribute_idx].description);
```
martinzink: We dont copy the strings here, so this will cause problems later on (and since OutputAttributes only operate with string_view-s there is no good place to store these strings...)

Currently its not a huge issue because we dont really use these anywhere except the documentation generating (and we are yet to do that for C processors, I was working on a PR to enable that when I encountered these)

minifi-api/include/minifi-c/minifi-c.h
szaszm: The current invariant implies that data can't be null, but the comment over data implies that it can be null. If data is null, then the range `[data, data+length)` is not valid. Could you clarify this? I'm assuming it's either data is null and length is 0 or the range is a valid string.

minifi-api/include/minifi-c/minifi-c.h
szaszm: I think it's optional. It's similar on the NiFi API: https://github.com/apache/nifi-api/blob/main/src/main/java/org/apache/nifi/annotation/behavior/DynamicProperty.java
The ExpressionLanguageScope enum supports 3 values: none, environment, and flow file attributes. We only have on/off currently.

extension-framework/cpp-extension-lib/include/api/utils/minifi-c-utils.h
fgerlits, line 1: Do we need both `minifi-c-utils.h` files?  If we need both, can you rename one of them, please?

minifi-api/include/minifi-c/minifi-c.h
szaszm: I agree that with no extra context, `MinifiValidator` is not very descriptive, but you would only encounter this when defining property metadata. 

On custom validators: I think they would have to be independent of these: these serve as a basic type for the properties, and minifi has to agree with C2 servers and other C2 clients on the set of possible "types". If we add custom validators, they should go in a separate member of the property metadata.

minifi-api/include/minifi-c/minifi-c.h
```diff
@@ -24,24 +24,200 @@ extern "C" {
...
+  MINIFI_UNSIGNED_INTEGER_VALIDATOR = 5,
+  MINIFI_DATA_SIZE_VALIDATOR = 6,
+  MINIFI_PORT_VALIDATOR = 7
+} MinifiStandardPropertyValidator;
+
 typedef struct MinifiExtensionCreateInfo {
   MinifiStringView name;
   MinifiStringView version;
-  void(*deinit)(void* user_data);
+  void(*deinit)(void*);
```
szaszm: I'd keep the name

core-framework/c-api-framework/include/api/utils/minifi-c-utils.h
```diff
@@ -0,0 +1,95 @@
...
+      }();
+      properties.push_back(MinifiProperty{
+        .name = toStringView(prop.name),
+        .display_name = toStringView(prop.display_name),
+        .description = toStringView(prop.description),
+        .is_required = prop.is_required,
+        .is_sensitive = prop.is_sensitive,
+
+        .default_value = default_value_begin ? sv_cache.data() + default_value_begin.value() : nullptr,
+        .allowed_values_count = gsl::narrow<uint32_t>(prop.allowed_values.size()),
```
fgerlits: both sides are `size_t`, so the `narrow()` can be removed

minifi-api/include/minifi-cpp/core/ProcessorDescriptor.h
```diff
@@ -28,6 +28,7 @@ class ProcessorDescriptor {
 
   virtual void setSupportedRelationships(std::span<const RelationshipDefinition> relationships) = 0;
   virtual void setSupportedProperties(std::span<const PropertyReference> properties) = 0;
+  virtual void setSupportedProperties(std::span<const Property> properties) = 0;
```
szaszm, line 31: why is this necessary?

core-framework/c-api-framework/src/core/logging/Logger.cpp
```diff
@@ -0,0 +1,69 @@
...
+}
+
+}  // namespace
+
+void Logger::set_max_log_size(int size) {
+  MinifiLoggerSetMaxLogSize(impl_, size);
+}
+
+void Logger::log_string(minifi::core::logging::LOG_LEVEL level, std::string str) {
+  MinifiLoggerLogString(impl_, toCLogLevel(level), MinifiStringView{.data = str.data(), .length = gsl::narrow_cast<uint32_t>(str.length())});
```
fgerlits: `narrow_cast` can be removed

core-framework/c-api-framework/src/core/ProcessorImpl.cpp
```diff
@@ -0,0 +1,83 @@
...
+ProcessorImpl::~ProcessorImpl() {
+  logger_->log_debug("Destroying processor {} with uuid {}", getName(), getUUIDStr());
+}
+
+bool ProcessorImpl::isWorkAvailable() {
+  return false;
+}
+
+void ProcessorImpl::restore(const std::shared_ptr<FlowFile>& /*file*/) {
+  gsl_Assert("Not implemented");
```
fgerlits: this should be
```suggestion
  gsl_Assert(false && "Not implemented");
```
(also in `minifi::core::ProcessorImpl::restore`)

minifi-api/include/minifi-c/minifi-c.h
adamdebreceni: either length is 0, or [data, data + length) is a valid range, if length is 0 we don't care what data is, which invariant implies that data can't be null?

minifi-api/common/include/minifi-cpp/core/logging/Logger.h
szaszm: do we no longer log IDs?

minifi-api/include/minifi-c/minifi-c.h
adamdebreceni: we need this for docs generation, as they are in a separate section

core-framework/c-api-framework/include/api/core/logging/Logger.h
adamdebreceni: removed

minifi-api/include/minifi-c/minifi-c.h
adamdebreceni: readded

extension-framework/cpp-extension-lib/include/api/utils/minifi-c-utils.h
adamdebreceni: moved one of them as ExtensionInitUtils.h to extension-framework

core-framework/c-api-framework/include/api/utils/minifi-c-utils.h
adamdebreceni: removed

core-framework/c-api-framework/src/core/logging/Logger.cpp
adamdebreceni: removed

core-framework/c-api-framework/src/core/ProcessorImpl.cpp
adamdebreceni: fixed

minifi-api/include/minifi-c/minifi-c.h
adamdebreceni: removed the field and then immediately added it back since we are using getRawDynamicProperty to circumvent this, so there is a use-case for not supporting expression language

core-framework/src/utils/file/FileUtils.cpp
```diff
@@ -71,6 +71,33 @@ bool contains(const std::filesystem::path& file_path, std::string_view text_to_s
   return std::search(view.begin(), view.end(), searcher) != view.end();
 }
 
+// finds a string of length at most @max_length with a given @prefix
+std::optional<std::string> find(const std::filesystem::path& file_path, std::string_view prefix, size_t max_length) {
```
fgerlits: I would prefer this function to take an `istream` argument instead of a `path`. Or it could be broken up into two functions, where the one with a `path` argument calls the one with an `istream` argument.

Also, it (at least the version with an `istream` argument, if broken up) should be unit tested, as the logic is quite complex.

extensions/llamacpp/tests/RunLlamaCppInferenceTests.cpp
```diff
@@ -224,13 +227,14 @@ TEST_CASE("Invalid values for optional double type properties throw exception")
...
   }
 
-  REQUIRE_THROWS_WITH(controller.trigger(minifi::test::InputFlowFileData{.content = "42", .attributes = {}}),
-      fmt::format("Expected parsable float from RunLlamaCppInference::{}, but got GeneralParsingError (Parsing Error:0)", property_name));
+  REQUIRE_THROWS(controller.trigger(minifi::test::InputFlowFileData{.content = "42", .attributes = {}}));
+  CHECK(minifi::test::utils::verifyLogLinePresenceInPollTime(1s,
+      fmt::format("Expected parsable float from RunLlamaCppInference::{}, but got GeneralParsingError (Parsing Error:0)", property_name)));
 }
-
+//
```
fgerlits: nitpicking, but this `//` shouldn't be here

extensions/standard-processors/tests/integration/InvokeHTTPTests.cpp
```diff
@@ -388,8 +388,8 @@ TEST_CASE("Data transfer speed parsing") {
   CHECK(processors::invoke_http::parseDataTransferSpeed("10 kB/s") == 10_KiB);
   CHECK(processors::invoke_http::parseDataTransferSpeed("20 MB/s") == 20_MiB);
   CHECK(processors::invoke_http::parseDataTransferSpeed("19 TB/s") == 19_TiB);
-  CHECK(processors::invoke_http::parseDataTransferSpeed("1KBinvalidsuffix") == nonstd::make_unexpected(make_error_code(core::ParsingErrorCode::GeneralParsingError)));
-  CHECK(processors::invoke_http::parseDataTransferSpeed("1KB") == nonstd::make_unexpected(make_error_code(core::ParsingErrorCode::GeneralParsingError)));
+  CHECK(processors::invoke_http::parseDataTransferSpeed("1KBinvalidsuffix").error().message() == "GeneralParsingError");
+  CHECK(processors::invoke_http::parseDataTransferSpeed("1KB").error().message() == "GeneralParsingError");
```
fgerlits: Can we keep the old version? Now, if the return value of `parseDataTransferSpeed` is not an error, we get undefined behavior from `.error()` instead of (as before) a well-defined test failure.

libminifi/include/utils/CProcessor.h
```diff
@@ -0,0 +1,176 @@
...
+class CProcessor : public minifi::core::ProcessorApi {
+ public:
+  CProcessor(CProcessorClassDescription class_description, minifi::core::ProcessorMetadata metadata)
+      : class_description_(std::move(class_description)),
+        metrics_extension_(std::make_shared<CProcessorMetricsWrapper>(*this)) {
+    metadata_ = metadata;
+    MinifiProcessorMetadata c_metadata;
+    auto uuid_str = metadata.uuid.to_string();
+    c_metadata.uuid = MinifiStringView{.data = uuid_str.data(), .length = gsl::narrow<uint32_t>(uuid_str.length())};
+    c_metadata.name = MinifiStringView{.data = metadata.name.data(), .length = gsl::narrow<uint32_t>(metadata.name.length())};
```
fgerlits: two more `narrow()`s which can be removed

libminifi/src/utils/CProcessor.cpp
```diff
@@ -0,0 +1,55 @@
...
+  std::vector<minifi::state::response::SerializedResponseNode> nodes;
+  for (auto& custom_node : source_processor_.getCustomMetrics()) {
+    size_t transformed_name_length = 0;
+    bool should_uppercase_next_letter = true;
+    for (size_t i = 0; i < custom_node.name.size(); i++) {
+      if (should_uppercase_next_letter) {
+        custom_node.name[transformed_name_length++] = static_cast<char>(std::toupper(static_cast<unsigned char>(custom_node.name[i])));
+        should_uppercase_next_letter = false;
+      } else if (custom_node.name[i] == '_') {
+        should_uppercase_next_letter = true;
```
fgerlits: I would switch these two conditions, so that underscore characters get removed consistently:
```suggestion
      if (custom_node.name[i] == '_') {
        should_uppercase_next_letter = true;
      } else if (should_uppercase_next_letter) {
        custom_node.name[transformed_name_length++] = static_cast<char>(std::toupper(static_cast<unsigned char>(custom_node.name[i])));
        should_uppercase_next_letter = false;
```

We could also convert names like `HTTP_GET_count` to `HttpGetCount` (i.e., change non-initial letters to lowercase).

Also, this name conversion code should be in its own function, and unit tested with examples including multiple underscores, leading/trailing underscores, numbers etc.

cmake/Extensions.cmake
```diff
@@ -100,6 +100,64 @@ macro(register_extension extension-name extension-display-name extension-guard d
...
+    target_compile_definitions(${extension-name}
+            PRIVATE "EXTENSION_NAME=${extension-name}" "EXTENSION_VERSION=${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH}")
+    set_target_properties(${extension-name} PROPERTIES
+            ENABLE_EXPORTS True
+            POSITION_INDEPENDENT_CODE ON)
+    if(WIN32)
+        set_target_properties(${extension-name} PROPERTIES
+                RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin"
+                ARCHIVE_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin"
+                WINDOWS_EXPORT_ALL_SYMBOLS TRUE)
```
szaszm: I would add the export macros in the C API extensions and not export all symbols. CMake's GenerateExportHeader can make it simpler.

core-framework/src/utils/file/FileUtils.cpp
adamdebreceni: split it, and added tests

extensions/llamacpp/tests/RunLlamaCppInferenceTests.cpp
adamdebreceni: removed

extensions/standard-processors/tests/integration/InvokeHTTPTests.cpp
adamdebreceni: managed to revert back to the original

libminifi/include/utils/CProcessor.h
adamdebreceni: removed

libminifi/src/utils/CProcessor.cpp
adamdebreceni: moved this to its own function and added tests

extension-framework/cpp-extension-lib/include/api/core/ProcessorImpl.h
adamdebreceni: renamed and moved it under extension-framework

cmake/Extensions.cmake
szaszm, line 116: I would add the export macros in the C API extensions and not export all symbols. CMake's GenerateExportHeader can make it simpler. At least that's the vision, but maybe there are blockers preventing this.

extension-framework/cpp-extension-lib/include/api/core/logging/Logger.h
szaszm, line 1: do we no longer log IDs? because I don't see them on the interface

minifi-api/include/minifi-c/minifi-c.h
adamdebreceni: renamed

minifi-api/include/minifi-c/minifi-c.h
```diff
@@ -24,24 +24,203 @@ extern "C" {
...
+} MinifiOutputAttributeDefinition;
+
+typedef struct MinifiDynamicPropertyDefinition {
+  MinifiStringView name;
+  MinifiStringView value;
+  MinifiStringView description;
+  MinifiBool supports_expression_language;
+} MinifiDynamicPropertyDefinition;
+
+typedef struct MinifiPropertyValidator MinifiPropertyValidator;
```
martinzink:
```suggestion
```

seems like leftover before latest changes and not neccessary anymore

extension-framework/cpp-extension-lib/include/api/core/logging/Logger.h
adamdebreceni: the trimming + id append has been moved upstream into LoggerBase

minifi-api/include/minifi-cpp/core/ProcessorDescriptor.h
adamdebreceni: we previously had a `Property::getReference` that worked incorrectly discarding `types` and `allowed_values`, since despite its name a `Property` is not easy to convert to `PropertyReference`, the former stores `std::vector<std::string>` while the latter uses `std::span<const std::string_view>` (this caused an issue in python processors as well)

minifi-api/include/minifi-c/minifi-c.h
adamdebreceni: removed

cmake/Extensions.cmake
adamdebreceni: this only has utility in tests where we directly link the test to (in this case) `libminifi-llamacpp.dll`,  since we are directly driving the processor in the test (not waiting for the extension to register the callbacks) we need access to all member function symbols

martinzink (approved):

lordgamez (approved):

PR #1990, opened 2025-07-11
MINIFICPP-2587 Add Sparkplug support for ConsumeMQTT processor

lordgamez:
- Extract protobuf library to be built separately
- Upgrade grpc library to v1.72.2
- Change RecordReader interface to expect InputStream instead of FlowFile and ProcessSession
- Add SparkplugReader record reader controller
- Add Record Reader and Record Writer properties to ConsumeMQTT processor
- Add "Add Attributes As Fields" to add flow file attributes as record fields if records are used in ConsumeMQTT
- Make ConsumeMQTT trigger unit testable
- Add new attributes to ConsumeMQTT created flow files to match NiFi
- Implement record reader and writer usage in ConsumeMQTT

https://issues.apache.org/jira/browse/MINIFICPP-2587

Depends on https://github.com/apache/nifi-minifi-cpp/pull/1968

--------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


CONTROLLERS.md
```diff
@@ -277,6 +278,20 @@ In the list below, the names of required properties appear in bold. Any other pr
 | Password     |               |                  | The password used for authentication. Required if Username is set.<br/>**Sensitive Property: true**                             |
 
 
+## SparkplugReader
+
+### Description
+
+Reads Sparkplug B messages and turns them into individual Record objects. The reader expects a single Sparkplug B payload in a read operation, which is a protobuf-encoded binary message. This reader is typically used with MQTT processors like ConsumeMQTT.
```
szaszm:
```suggestion
Reads Sparkplug B messages and turns them into individual Record objects. The reader expects a single Sparkplug B payload in a read operation, which is a protobuf-encoded binary message. This reader is designed to be used with ConsumeMQTT, since Sparkplug B is an MQTT-based protocol.
```

docker/test/integration/cluster/checkers/sparkplug_b_pb2.py
fgerlits, line 1: Could we generate this when building (or before running) the tests instead of having the generated code in the repo?

docker/test/integration/features/steps/steps.py
```diff
@@ -435,6 +436,18 @@ def step_impl(context):
     context.test.start('mqtt-broker')
 
 
+@given("a SparkplugBReader controller service is set up")
+def step_impl(context):
+    json_record_set_reader = SparkplugBReader("SparkplugBReader")
+    container = context.test.acquire_container(context=context, name="minifi-cpp-flow")
+    container.add_controller(json_record_set_reader)
```
fgerlits: nitpicking, but this is not JSON:
```suggestion
    record_set_reader = SparkplugBReader("SparkplugBReader")
    container = context.test.acquire_container(context=context, name="minifi-cpp-flow")
    container.add_controller(record_set_reader)
```

PROCESSORS.md
```diff
@@ -433,10 +436,15 @@ In the list below, the names of required properties appear in bold. Any other pr
...
+| mqtt.broker          |              | URI of the sending broker                                                                                                             |
+| mqtt.topic           |              | Topic of the message                                                                                                                  |
+| mqtt.topic.segment.n |              | The nth topic segment of the message                                                                                                  |
+| mqtt.qos             |              | The quality of service for this message.                                                                                              |
+| mqtt.isDuplicate     |              | Whether or not this message might be a duplicate of one which has already been received.                                              |
+| mqtt.isRetained      |              | Whether or not this message was from a current publisher, or was "retained" by the server as the last message published on the topic. |
```
szaszm, line 448: I'm only seeing mqtt.broker in the flow file attributes with the record reader/writer set, but I'm seeing similar record attributes with an underscore prefix. We should make it clear which attribute ends up where.

extensions/mqtt/processors/ConsumeMQTT.cpp
```diff
@@ -26,17 +25,44 @@
...
+  std::string service_name = context.getProperty(property).value_or("");
+  if (!IsNullOrEmpty(service_name)) {
+    auto record_set_io = std::dynamic_pointer_cast<RecordSetIO>(context.getControllerService(service_name, processor_uuid));
+    if (!record_set_io) {
+      throw Exception(ExceptionType::PROCESS_SCHEDULE_EXCEPTION, fmt::format("'{}' property is set to invalid controller service '{}'", property.name, service_name));
+    }
+    return record_set_io;
+  }
+  return nullptr;
+}
```
fgerlits: we could use `utils::parseOptionalControllerService()` instead of this

extensions/mqtt/processors/ConsumeMQTT.cpp
```diff
@@ -58,8 +84,55 @@ void ConsumeMQTT::readProperties(core::ProcessContext& context) {
...
+  }
+  if (record_set.empty()) {
+    logger_->log_debug("No records to write, skipping FlowFile creation");
+    return;
+  }
+  std::shared_ptr<core::FlowFile> flow_file = session.create();
+  record_set_writer_->write(record_set, flow_file, session);
+  session.putAttribute(*flow_file, RecordCountOutputAttribute.name, std::to_string(record_set.size()));
+  session.putAttribute(*flow_file, BrokerOutputAttribute.name, uri_);
+  session.transfer(flow_file, Success);
```
fgerlits, line 113: In `tranferMessagesAsFlowFiles`, we add a bunch of flow file attributes using `putUserPropertiesAsAttributes` and `fillAttributeFromContentType`. Are those not needed / not useful here?

extensions/mqtt/processors/ConsumeMQTT.h
```diff
@@ -92,15 +107,26 @@ class ConsumeMQTT : public processors::AbstractMQTTProcessor {
...
+  EXTENSIONAPI static constexpr auto QosOutputAttribute = core::OutputAttributeDefinition<0>{"mqtt.qos", {}, "The quality of service for this message."};
+  EXTENSIONAPI static constexpr auto IsDuplicateOutputAttribute = core::OutputAttributeDefinition<0>{"mqtt.isDuplicate", {},
+      "Whether or not this message might be a duplicate of one which has already been received."};
+  EXTENSIONAPI static constexpr auto IsRetainedOutputAttribute = core::OutputAttributeDefinition<0>{"mqtt.isRetained", {},
+      "Whether or not this message was from a current publisher, or was \"retained\" by the server as the last message published on the topic."};
+  EXTENSIONAPI static constexpr auto RecordCountOutputAttribute = core::OutputAttributeDefinition<0>{"record.count", {}, "The number of records received"};
+  EXTENSIONAPI static constexpr auto OutputAttributes = std::array<core::OutputAttributeReference, 7>{BrokerOutputAttribute, TopicOutputAttribute, TopicSegmentOutputAttribute,
+      QosOutputAttribute, IsDuplicateOutputAttribute, IsRetainedOutputAttribute, RecordCountOutputAttribute};
```
fgerlits:
```suggestion
  EXTENSIONAPI static constexpr auto OutputAttributes = std::to_array<core::OutputAttributeReference>({BrokerOutputAttribute, TopicOutputAttribute, TopicSegmentOutputAttribute,
      QosOutputAttribute, IsDuplicateOutputAttribute, IsRetainedOutputAttribute, RecordCountOutputAttribute});
```
would be marginally nicer

docker/test/integration/cluster/checkers/sparkplug_b_pb2.py
lordgamez: I'm not sure if it's worth adding protoc as a required dependency for the test environment just for generating this file, so I think in this case it's easier just to add this file to the repo

extensions/mqtt/processors/ConsumeMQTT.cpp
lordgamez: My problem with that currently that we do not differentiate if the property was set, but it was invalid or if the property was not set at all. Maybe we should change the current utility to only return std::nullopt if the input is invalid, otherwise return nullptr, what do you think?

extensions/mqtt/processors/ConsumeMQTT.cpp
lordgamez: In this case multiple records can be written to the flow file with each record being a separate MQTT message while in the `tranferMessagesAsFlowFiles` every flow file is a single MQTT message. Those attributes are MQTT message specific, so they cannot be added to the flow file if it contains multiple MQTT messages. For this scenario we can set the `Add Attributes As Fields` property to set these attributes as record fields for each record in the flow file.

extensions/mqtt/processors/ConsumeMQTT.cpp
fgerlits: Good point. Another option would be to throw on invalid and return either nullopt or nullptr if the property is not set.

EDIT: we agreed that I will change `utils::parseOptionalControllerService()` to throw (and also to not use `optional`) in a separate PR.  For now, this can stay as it is, and we will rebase one or the other PR depending on which gets merged first.

CONTROLLERS.md
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1990/commits/7e2cad557da32447caada0536e44ada143104195

PROCESSORS.md
lordgamez: Added info on this in the description in https://github.com/apache/nifi-minifi-cpp/pull/1990/commits/7e2cad557da32447caada0536e44ada143104195

docker/test/integration/features/steps/steps.py
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1990/commits/7e2cad557da32447caada0536e44ada143104195

extensions/mqtt/processors/ConsumeMQTT.h
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1990/commits/7e2cad557da32447caada0536e44ada143104195

docker/test/integration/cluster/checkers/sparkplug_b_pb2.py
fgerlits: OK, makes sense

PR #1994, opened 2025-07-22
MINIFICPP-2591 Add new value option to FetchOPCProcessor lazy mode

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2591

-----------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/opc/src/fetchopc.cpp
```diff
@@ -107,30 +116,86 @@ void FetchOPCProcessor::onTrigger(core::ProcessContext& context, core::ProcessSe
     logger_->log_warn("Found no variables when traversing the specified node. No flowfiles are generated. Yielding...");
     yield();
   }
+
+  state_manager_->set(state_map);
```
fgerlits, line 120: Should this be in a `gsl::finally`? I.e., do we want to update the state if the processing failed? I would guess yes, but I'm not sure.

extensions/opc/src/fetchopc.cpp
```diff
@@ -140,7 +205,7 @@ bool FetchOPCProcessor::nodeFoundCallBack(const UA_ReferenceDescription *ref, co
   return true;
 }
 
-void FetchOPCProcessor::OPCData2FlowFile(const opc::NodeData& opc_node, core::ProcessContext&, core::ProcessSession& session) {
+void FetchOPCProcessor::OPCData2FlowFile(const opc::NodeData& opc_node, core::ProcessContext&, core::ProcessSession& session, const std::string& node_value) const {
```
fgerlits: Is there a branch where we write anything other than `opc::nodeValue2String(opc_node)`?  I can't find one, so I don't understand the reason for adding a new argument to the function.

extensions/opc/src/fetchopc.cpp
lordgamez: In case there is an error we do not call the `traverse` in the onTrigger on the OPC nodes, so we do not have any new when we return early, so I think there is no need to update the state in this case. We also do not want to store values that we did not write into a flowfile before.

extensions/opc/src/fetchopc.cpp
lordgamez: It is just a minimal optimization in the `writeFlowFileUsingLazyModeWithNewValue` function where we call the `nodeValue2String` beforehand to check if the value is new opposed to the one in the stored state, and if so we pass that to the `OPCData2FlowFile` instead of calling the `opc::nodeValue2String` again.

extensions/opc/src/fetchopc.cpp
fgerlits: makes sense, thanks for the clarification!

extensions/opc/src/fetchopc.cpp
fgerlits: In my opinion all this extra code is not worth it for the very small performance improvement (if any). We can keep it if you like it, but I would remove it.

lordgamez: I suppose you are right, it should not make that much difference, updated in https://github.com/apache/nifi-minifi-cpp/pull/1994/commits/d1fc7f4e49df800d8e08812c135644c2767aff52

adamdebreceni (approved):

PR #1995, opened 2025-07-29
MINIFICPP-2594 Add XMLReader controller service

lordgamez:
- Upgrade pugixml library to v1.15

https://issues.apache.org/jira/browse/MINIFICPP-2594

Depends on https://github.com/apache/nifi-minifi-cpp/pull/1999

------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/standard-processors/controllers/XMLReader.h
+++ b/extensions/standard-processors/controllers/XMLReader.h
@@ -66,6 +66,7 @@ class XMLReader final : public core::RecordSetReaderImpl {
   EXTENSIONAPI static constexpr auto Properties = std::array<core::PropertyReference, 4>{FieldNameForContent, ParseXMLAttributes, AttributePrefix, ExpectRecordsAsArray};
 
   EXTENSIONAPI static constexpr bool SupportsDynamicProperties = false;
+  EXTENSIONAPI static constexpr auto ImplementsApis = std::array{ RecordSetReader::ProvidesApi };
   ADD_COMMON_VIRTUAL_FUNCTIONS_FOR_CONTROLLER_SERVICES
 
   nonstd::expected<core::RecordSet, std::error_code> read(io::InputStream& input_stream) override;


CONTROLLERS.md
```diff
@@ -332,3 +333,21 @@ In the list below, the names of required properties appear in bold. Any other pr
...
+| **Parse XML Attributes**    | false         | true<br/>false   | When 'Schema Access Strategy' is 'Infer Schema' and this property is 'true' then XML attributes are parsed and added to the record as new fields. When the schema is inferred but this property is 'false', XML attributes and their values are ignored.                                                                                                                                                                                                                                                                                                                                      |
```
fgerlits: Unlike in NiFi, here we don't have a `Schema access strategy` property, so this description needs to be updated.

extensions/standard-processors/controllers/XMLReader.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+#include "core/Resource.h"
+#include "utils/TimeUtil.h"
+#include "utils/gsl.h"
+
+namespace org::apache::nifi::minifi::standard {
+
+namespace {
+bool hasChildNodes(const pugi::xml_node& node) {
+  return std::any_of(node.begin(), node.end(), [] (const pugi::xml_node& child) {
```
fgerlits: this (and most, maybe all, of the `any_of` and `all_of` clauses in this file) could be rewritten to use ranges:
```suggestion
  return std::ranges::any_of(node, [](const pugi::xml_node& child) {
```

extensions/standard-processors/controllers/XMLReader.cpp
```diff
@@ -0,0 +1,200 @@
...
+    const auto read_ret = gsl::narrow<int64_t>(input_stream.read(as_writable_bytes(std::span(content))));
+    if (io::isError(read_ret)) {
+      logger_->log_error("Failed to read XML data from input stream");
+      return -1;
+    }
+    if (!parseRecordsFromXml(record_set, content)) {
+      return -1;
+    }
+    return read_ret;
+  }(input_stream);
```
fgerlits: here too, as in #1999, I would keep the result as a `size_t`, and return `io::STREAM_ERROR` instead of -1

extensions/standard-processors/controllers/XMLReader.cpp
```diff
@@ -0,0 +1,200 @@
...
+    return;
+  } catch (const std::exception&) {
+  }
+
+  addRecordFieldToObject(record_object, name, core::RecordField(value));
+}
+
+void XMLReader::writeRecordFieldFromXmlNode(core::RecordObject& record_object, const pugi::xml_node& node) const {
+  writeRecordField(record_object, node.name(), node.child_value());
+}
```
fgerlits: since this function is 1 line and is only used in one place, I would get rid of it

extensions/standard-processors/controllers/XMLReader.cpp
```diff
@@ -0,0 +1,200 @@
...
+    return;
+  } else if (auto date = utils::timeutils::parseDateTimeStr(value)) {
+    addRecordFieldToObject(record_object, name, core::RecordField(*date));
+    return;
+  } else if (auto date = utils::timeutils::parseRfc3339(value)) {
+    addRecordFieldToObject(record_object, name, core::RecordField(*date));
+    return;
+  }
+
+  if (std::all_of(value.begin(), value.end(), ::isdigit)) {
```
fgerlits: should we allow whitespace around numbers?

extensions/standard-processors/tests/unit/XMLReaderTests.cpp
```diff
@@ -0,0 +1,250 @@
...
+#include "unit/Catch.h"
+#include "unit/TestBase.h"
+
+namespace org::apache::nifi::minifi::standard::test {
+
+class XMLReaderTestFixture {
+ public:
+  XMLReaderTestFixture() : xml_reader_("XMLReader") {
+    LogTestController::getInstance().setTrace<XMLReader>();
+  }
```
fgerlits: we should also call `LogTestController::getInstance().clear()` here, to make sure a test won't find the output of a previous test

fgerlits (commented):
I think we need to add pugixml to the NOTICE file, as well.

Also, TinyXml2 is listed in both the LICENCE and NOTICE files, but we don't use it (any more?) -- it can be removed.

extensions/standard-processors/controllers/XMLReader.cpp
lordgamez: I checked in NiFi and it is not allowed there either, it throws an exception when NiFi tries to parse a number with whitespaces around it: `java.lang.NumberFormatException: For input string: "  1"`
I think we should not allow it either in this case.

CONTROLLERS.md
lordgamez: Good catch, updated in https://github.com/apache/nifi-minifi-cpp/pull/1995/commits/044f0f5b530db43bb0d0771816b407c31894e0b9

extensions/standard-processors/controllers/XMLReader.cpp
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1995/commits/044f0f5b530db43bb0d0771816b407c31894e0b9

extensions/standard-processors/controllers/XMLReader.cpp
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1995/commits/044f0f5b530db43bb0d0771816b407c31894e0b9

extensions/standard-processors/controllers/XMLReader.cpp
lordgamez: Good point, removed in https://github.com/apache/nifi-minifi-cpp/pull/1995/commits/044f0f5b530db43bb0d0771816b407c31894e0b9

extensions/standard-processors/tests/unit/XMLReaderTests.cpp
lordgamez: Added in https://github.com/apache/nifi-minifi-cpp/pull/1995/commits/044f0f5b530db43bb0d0771816b407c31894e0b9

fgerlits (approved):

extensions/standard-processors/controllers/XMLReader.h
```diff
@@ -0,0 +1,95 @@
...
+          "with a \"wrapper element\". This property dictates whether the reader expects a FlowFile to consist of a single Record or a series of Records with a \"wrapper element\" "
+          "that will be ignored.")
+      .isRequired(true)
+      .withValidator(core::StandardPropertyValidators::BOOLEAN_VALIDATOR)
+      .withDefaultValue("false")
+      .build();
+
+  EXTENSIONAPI static constexpr auto Properties = std::array<core::PropertyReference, 4>{FieldNameForContent, ParseXMLAttributes, AttributePrefix, ExpectRecordsAsArray};
+
+  EXTENSIONAPI static constexpr bool SupportsDynamicProperties = false;
```
martinzink, line 68:
```suggestion
  EXTENSIONAPI static constexpr bool SupportsDynamicProperties = false;
  EXTENSIONAPI static constexpr auto ImplementsApis = std::array{ RecordSetReader::ProvidesApi };
```

Similarly to the writer PR youll need this for allowed types to work on c2

lordgamez: Good catch, updated in https://github.com/apache/nifi-minifi-cpp/pull/1995/commits/bd54d1c9ae78fb36da54e9b80d8047a8f0504522

PR #1997, opened 2025-08-04
MINIFICPP-2596 Add XMLRecordSetWriter controller service

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2596

Depends on https://github.com/apache/nifi-minifi-cpp/pull/1995

---------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


CONTROLLERS.md
```diff
@@ -351,3 +352,23 @@ In the list below, the names of required properties appear in bold. Any other pr
...
+| **Omit XML Declaration**    | false         | true<br/>false                                                        | Specifies whether or not to include XML declaration                                                                                                                       |
+| **Pretty Print XML**        | false         | true<br/>false                                                        | Specifies whether or not the XML should be pretty printed                                                                                                                 |
+| **Name of Record Tag**      |               |                                                                       | Specifies the name of the XML record tag wrapping the record fields.                                                                                                      |
+| **Name of Root Tag**        |               |                                                                       | Specifies the name of the XML root tag wrapping the record set. This property has to be defined if the writer is supposed to write multiple records in a single FlowFile. |
```
fgerlits: I don't understand the second sentence: the property is required, so it always has to be defined.

CONTROLLERS.md
```diff
@@ -351,3 +352,23 @@ In the list below, the names of required properties appear in bold. Any other pr
...
+
+| Name                        | Default Value | Allowable Values                                                      | Description                                                                                                                                                               |
+|-----------------------------|---------------|-----------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Array Tag Name              |               |                                                                       | Name of the tag used by property "Wrap Elements of Arrays" to write arrays                                                                                                |
+| **Wrap Elements of Arrays** | No Wrapping   | Use Property as Wrapper<br/>Use Property for Elements<br/>No Wrapping | Specifies how the writer wraps elements of fields of type array                                                                                                           |
```
fgerlits: NiFi has an explanation of each of the options; I think that would be useful here, too, because the difference between "Use Property as Wrapper" and "Use Property for Elements" is not obvious.

In the long term, it would be nice to have a way to attach descriptions to each allowable value (like NiFi does), but for now we can put this in the Description.

extensions/standard-processors/controllers/XMLRecordSetWriter.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+  array_tag_name_ = getProperty(ArrayTagName.name).value_or("");
+  if (array_tag_name_.empty() &&
+      (wrap_elements_of_arrays_ == WrapElementsOfArraysOptions::UsePropertyAsWrapper ||
+       wrap_elements_of_arrays_ == WrapElementsOfArraysOptions::UsePropertyForElements)) {
+    throw Exception(PROCESS_SCHEDULE_EXCEPTION, "Array Tag Name property must be set when Wrap Elements of Arrays is set to Use Property as Wrapper or Use Property for Elements");
+  }
+
+  omit_xml_declaration_ = getProperty(OmitXMLDeclaration.name).value_or("false") == "true";
+  pretty_print_xml_ = getProperty(PrettyPrintXML.name).value_or("false") == "true";
```
fgerlits: I don't like the fact that if the user manually sets these properties to "True" then we'll silently treat it as "false"; can we do some error checking, eg. with `parsing::parseBool()`?

extensions/standard-processors/controllers/XMLRecordSetWriter.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+  name_of_root_tag_ = getProperty(NameOfRootTag.name).value_or("");
+  if (name_of_root_tag_.empty()) {
+    throw Exception(PROCESS_SCHEDULE_EXCEPTION, "Name of Root Tag property must be set");
+  }
+}
+
+std::string XMLRecordSetWriter::formatXmlOutput(pugi::xml_document& xml_doc) const {
+  std::ostringstream xml_string_stream;
+  uint64_t xml_formatting_flags = 0;
```
fgerlits: we could make this `unsigned int` to start with, so we don't need to narrow it in line 64

extensions/standard-processors/controllers/XMLRecordSetWriter.cpp
```diff
@@ -0,0 +1,145 @@
...
+  } else if (std::holds_alternative<std::chrono::system_clock::time_point>(field.value_)) {
+    auto time_point = std::get<std::chrono::system_clock::time_point>(field.value_);
+    auto time_str = utils::timeutils::getDateTimeStr(std::chrono::time_point_cast<std::chrono::seconds>(time_point));
+    field_node.text().set(time_str.c_str());
+  } else if (std::holds_alternative<core::RecordObject>(field.value_)) {
+    const auto& record_object = std::get<core::RecordObject>(field.value_);
+    for (const auto& [obj_key, obj_field] : record_object) {
+      convertRecordField(obj_key, obj_field, field_node);
+    }
+  }
```
fgerlits: I think a visitor pattern like in `RecordField::toJson()` would be nicer.

CONTROLLERS.md
lordgamez: You are right, this was copied from NiFi's documentation, but it is a bit confusing. Updated in https://github.com/apache/nifi-minifi-cpp/pull/1997/commits/c211b83424ca4133fcb8e673a3e15d0e3a3341d3

CONTROLLERS.md
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1997/commits/c211b83424ca4133fcb8e673a3e15d0e3a3341d3

extensions/standard-processors/controllers/XMLRecordSetWriter.cpp
lordgamez: Good point, added in https://github.com/apache/nifi-minifi-cpp/pull/1997/commits/c211b83424ca4133fcb8e673a3e15d0e3a3341d3

extensions/standard-processors/controllers/XMLRecordSetWriter.cpp
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/1997/commits/c211b83424ca4133fcb8e673a3e15d0e3a3341d3

extensions/standard-processors/controllers/XMLRecordSetWriter.cpp
lordgamez: Good idea, updated in https://github.com/apache/nifi-minifi-cpp/pull/1997/commits/c211b83424ca4133fcb8e673a3e15d0e3a3341d3

CONTROLLERS.md
fgerlits: I would remove the second sentence completely; since the property is required (and has no default), "This property has to be set if..." is not applicable: the user always has to set this property.

lordgamez: Good point, updated in https://github.com/apache/nifi-minifi-cpp/pull/1997/commits/ce50c324ae43c61b0e589a8434db7c3211d2c1e6

extensions/standard-processors/controllers/XMLRecordSetWriter.h
```diff
@@ -0,0 +1,132 @@
...
+      .withDescription("Specifies the name of the XML root tag wrapping the record set.")
+      .withValidator(core::StandardPropertyValidators::NON_BLANK_VALIDATOR)
+      .isRequired(true)
+      .build();
+
+  EXTENSIONAPI static constexpr auto Properties = std::array<core::PropertyReference, 6>{
+    ArrayTagName, WrapElementsOfArrays, OmitXMLDeclaration, PrettyPrintXML, NameOfRecordTag, NameOfRootTag
+  };
+
+  EXTENSIONAPI static constexpr bool SupportsDynamicProperties = false;
```
martinzink, line 104: you will need the ImplementsApis for allowed types to work
```suggestion
  EXTENSIONAPI static constexpr bool SupportsDynamicProperties = false;
  EXTENSIONAPI static constexpr auto ImplementsApis = std::array{ RecordSetWriter::ProvidesApi };

```

lordgamez: Good catch, updated in https://github.com/apache/nifi-minifi-cpp/pull/1997/commits/5a0c4ee8d362ce664bb049bfdbae20640d72ae38

PR #2001, opened 2025-08-06
MINIFICPP-2598 Fixing inherited ControllerService manifest issues

martinzink:
Renamed SSLContextServiceImpl -> SSLContextService
 Added ControllerServiceApiDefinition
 Manifest now includes providedApiImplementations
 JsonRecordSetWriter now advertises that it implements RecordSetWriter
 JsonTreeReader now advertises that it implements RecordSetReader
 SSLContextService now advertises that it implements SSLContextServiceInterface

### For all changes:
     in the commit message?

### For code changes:

### Note:


minifi-api/include/minifi-cpp/controllers/SSLContextServiceInterface.h
```diff
@@ -33,8 +34,14 @@ namespace org::apache::nifi::minifi::controllers {
  * Justification: Abstracts SSL support out of processors into a
  * configurable controller service.
  */
-class SSLContextService : public virtual core::controller::ControllerService {
+class SSLContextServiceInterface : public virtual core::controller::ControllerService {
  public:
+  static constexpr auto ControllerServiceApiDefinition = core::ControllerServiceApiDefinition{
+    .artifact = "minifi-system",
+    .group = "org.apache.nifi.minifi",
+    .type = "org.apache.nifi.minifi.controllers.SSLContextServiceInterface",
```
szaszm, line 42: Maybe we should call this SSLContextService on the interface, and hide the interface-implementation relationship?

utils/include/core/controller/ControllerService.h
```diff
@@ -105,6 +106,9 @@ class ControllerServiceImpl : public ConfigurableComponentImpl, public Connectab
     linked_services_ = services;
   }
 
+
+  static constexpr auto ApiImplementations = std::array<ControllerServiceApiDefinition, 0>{};
+
```
szaszm: Is this needed in the empty base class?

minifi-api/include/minifi-cpp/controllers/SSLContextServiceInterface.h
lordgamez: I think it's cleaner this way that we have a the same interface type in the manifest as in the code for all controller service interface types, but if we want to remove this interface later I think we can rename it that time.

lordgamez (approved):

minifi-api/include/minifi-cpp/controllers/SSLContextServiceInterface.h
szaszm: At first I thought that would break API when changing it, but maybe not if we don't rename the actual SSLContextService

martinzink: I think that could cause a whole lot of issues down the line for not much of an upside.
And NiFi also has multiple SSLContextService implementations so I am not sure if we wanna remove this separation any time soon.

utils/include/core/controller/ControllerService.h
martinzink: Yeah, during compilation we reference it for every ControllerService typed resource https://github.com/martinzink/nifi-minifi-cpp/blob/69111f4d239c37f4f1046604de7e4d03e968846f/utils/include/agent/agent_docs.h#L73

extensions/standard-processors/controllers/JsonRecordSetWriter.h
```diff
@@ -83,6 +83,9 @@ class JsonRecordSetWriter final : public core::RecordSetWriterImpl {
   });
 
   EXTENSIONAPI static constexpr bool SupportsDynamicProperties = false;
+  EXTENSIONAPI static constexpr auto ApiImplementations = std::to_array<core::ControllerServiceApiDefinition>({
+    RecordSetWriter::ControllerServiceApiDefinition
+  });
```
fgerlits: since there is no type conversion, these could be shortened to
```suggestion
  EXTENSIONAPI static constexpr auto ApiImplementations = std::array{RecordSetWriter::ControllerServiceApiDefinition};
```

libminifi/include/controllers/SSLContextService.h
```diff
@@ -72,20 +72,20 @@ class SSLContext {
...
  */
-class SSLContextServiceImpl : public core::controller::ControllerServiceImpl, public SSLContextService {
+class SSLContextService : public core::controller::ControllerServiceImpl, public SSLContextServiceInterface {
  public:
-  explicit SSLContextServiceImpl(std::string_view name, const utils::Identifier &uuid = {})
+  explicit SSLContextService(std::string_view name, const utils::Identifier &uuid = {})
       : ControllerServiceImpl(name, uuid),
         initialized_(false),
-        logger_(core::logging::LoggerFactory<SSLContextService>::getLogger(uuid_)) {
+        logger_(core::logging::LoggerFactory<SSLContextServiceInterface>::getLogger(uuid_)) {
```
fgerlits: we should keep the class name in the log as "SSLContextService" (also in line 86)

minifi-api/include/minifi-cpp/controllers/RecordSetReader.h
```diff
@@ -30,6 +30,12 @@ namespace org::apache::nifi::minifi::core {
 
 class RecordSetReader : public virtual controller::ControllerService {
  public:
+  static constexpr auto ControllerServiceApiDefinition = core::ControllerServiceApiDefinition{
```
fgerlits: something like "ProvidesApi" would be a better name

come to think about it, I would also rename "ApiImplementations" to "ImplementsApis", but I feel less strongly about that

fgerlits (commented):
I would like to include the list of implemented APIs in `CONTROLLERS.md`, auto-generated.

But this can be done separately, it isn't needed for the release.

extensions/standard-processors/controllers/JsonRecordSetWriter.h
martinzink: 👍 good idea https://github.com/apache/nifi-minifi-cpp/pull/2001/commits/b714eb903d50d1359832004904777dd6d296a912

libminifi/include/controllers/SSLContextService.h
martinzink: good catch and I agree, but to be fair it was like this previously aswell (Impl used Interface logger)
changed it https://github.com/apache/nifi-minifi-cpp/pull/2001/commits/b714eb903d50d1359832004904777dd6d296a912

minifi-api/include/minifi-cpp/controllers/RecordSetReader.h
martinzink: 👍 you wanna rename only the member or the struct aswell?
renamed the member 
https://github.com/apache/nifi-minifi-cpp/pull/2001/commits/b714eb903d50d1359832004904777dd6d296a912

fgerlits: just the member, but some `ApiImplementations` got left unchanged

also: if b714eb9 comes back all green, that means we need to add some tests...

szaszm (approved):

minifi-api/include/minifi-cpp/controllers/RecordSetReader.h
martinzink: 👍 good catch, I've added a test that would have failed https://github.com/apache/nifi-minifi-cpp/pull/2001/commits/ded65512ca5985a1d92ff26b4b995333f97a2550 and fixed it in https://github.com/apache/nifi-minifi-cpp/pull/2001/commits/2eacb4478f9306af31c0ce3a0a537384914d94ce

PR #2005, opened 2025-08-12
MINIFICPP-2604 Upgrade Azure SDK to the latest release

lordgamez:
- Upgrade Azure Identity to version 1.13.0
- Upgrade Azure Storage Blobs to version 12.14.0
- Upgrade Azure Storage Files Datalake to version 12.13.0-beta.1

https://issues.apache.org/jira/browse/MINIFICPP-2604

-----------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/cmake/AzureSdkCpp.cmake
+++ b/cmake/AzureSdkCpp.cmake
@@ -38,10 +38,12 @@ set(DISABLE_AMQP ON CACHE INTERNAL "")
 
 set(PATCH_FILE_1 "${CMAKE_SOURCE_DIR}/thirdparty/azure-sdk-cpp/wil.patch")
 set(PATCH_FILE_2 "${CMAKE_SOURCE_DIR}/thirdparty/azure-sdk-cpp/fix-openssl-helper.patch")
+set(PATCH_FILE_3 "${CMAKE_SOURCE_DIR}/thirdparty/azure-sdk-cpp/fix-managed-identity.patch")
 
 set(PC ${Bash_EXECUTABLE}  -c "set -x &&\
         (\\\"${Patch_EXECUTABLE}\\\" -p1 -R -s -f --dry-run -i \\\"${PATCH_FILE_1}\\\" || \\\"${Patch_EXECUTABLE}\\\" -p1 -N -i \\\"${PATCH_FILE_1}\\\") &&\
-        (\\\"${Patch_EXECUTABLE}\\\" -p1 -R -s -f --dry-run -i \\\"${PATCH_FILE_2}\\\" || \\\"${Patch_EXECUTABLE}\\\" -p1 -N -i \\\"${PATCH_FILE_2}\\\")")
+        (\\\"${Patch_EXECUTABLE}\\\" -p1 -R -s -f --dry-run -i \\\"${PATCH_FILE_2}\\\" || \\\"${Patch_EXECUTABLE}\\\" -p1 -N -i \\\"${PATCH_FILE_2}\\\") &&\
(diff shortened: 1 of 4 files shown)


cmake/AzureSdkCpp.cmake
```diff
@@ -0,0 +1,57 @@
...
+
+FetchContent_Declare(asdkext
+    URL https://github.com/Azure/azure-sdk-for-cpp/archive/refs/tags/azure-identity_1.13.0.tar.gz
+    URL_HASH "SHA256=a18ce70f8da11266b6a1aec3c02233ca889ea6137471a9f633131f00dfd386c0"
+    PATCH_COMMAND "${PC}"
+    SYSTEM
+)
+
+if (WIN32)
+    add_compile_definitions(CURL_STATICLIB)
```
szaszm: We should use target-specific cmake commands here, not directory / global ones. Maybe we should add this definition to the interface definitions of the curl target, so that all of its consumers get the flag?

lordgamez: Good idea, updated in https://github.com/apache/nifi-minifi-cpp/pull/2005/commits/5614ecf515b185103362ccc06e9cf244187b310e

thirdparty/azure-sdk-cpp/fix-openssl-helper.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/sdk/attestation/azure-security-attestation/src/private/crypto/openssl/openssl_helpers.hpp b/sdk/attestation/azure-security-attestation/src/private/crypto/openssl/openssl_helpers.hpp
+index a52791cea..b5e76a0f9 100644
+--- a/sdk/attestation/azure-security-attestation/src/private/crypto/openssl/openssl_helpers.hpp
++++ b/sdk/attestation/azure-security-attestation/src/private/crypto/openssl/openssl_helpers.hpp
+@@ -39,7 +39,7 @@ namespace Azure { namespace Security { namespace Attestation { namespace _detail
+   using openssl_evp_pkey = openssl_unique_ptr<EVP_PKEY>;
+   using openssl_bio = openssl_unique_ptr<BIO>;
+ 
+-#ifdef __cpp_nontype_template_parameter_auto
++#if defined(__cpp_nontype_template_parameter_auto) && !defined(_WIN32)
```
fgerlits: can you add a comment about why we need to exclude msvc here, please?

lordgamez: Added comment in https://github.com/apache/nifi-minifi-cpp/pull/2005/commits/83360da0319638c7cf4b1198742f5f5ad163a1f5

fgerlits: Is this a known bug in MSVC? Or a C++20 feature not implemented, yet? It would be good to have some way of knowing when we can remove this workaround.

lordgamez: After discussing and checking this issue further in private, this seems to be a bug in the C++17 implementation/usage of the `make_openssl_unique` helper function in the Azure SDK. On Windows we override the `/std:c++14` flag set in the build files with `/std:c++latest` so we use the faulty code there. I opened an issue for this on the Azure SDK project page https://github.com/Azure/azure-sdk-for-cpp/issues/6697 and patched this issue in the latest commit: https://github.com/apache/nifi-minifi-cpp/pull/2005/commits/380d04246a5f496bcfb67ce3ccad65af6f1cbec5

martinzink (approved):

PR #2007, opened 2025-08-12
MINIFICPP-2589 Refactor ConsumeWindowsEventLog

fgerlits:
https://issues.apache.org/jira/browse/MINIFICPP-2589

* lots of renaming stuff
* use magic_enum instead of hand-rolled enum parsing
* some inlining / pulling out of methods
* unify the printing of Windows error messages using fmt::format

etc

---

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/extensions/windows-event-log/ConsumeWindowsEventLog.cpp
+++ b/extensions/windows-event-log/ConsumeWindowsEventLog.cpp
@@ -58,6 +58,22 @@ auto createTimer() {
 }
 }  // namespace
 
+namespace org::apache::nifi::minifi::wel {
+std::function<bool(std::string_view)> parseSidMatcher(const std::optional<std::string>& sid_matcher) {
+  if (!sid_matcher || sid_matcher->empty()) {
+    return [](std::string_view){ return false; };
+  }
+
+  if (std::smatch match; utils::regexMatch(*sid_matcher, match, utils::Regex{R"_(\.\*(\w+))_"})) {
+    std::string suffix = match[1];
+    return [suffix](std::string_view field_name) { return utils::string::endsWith(field_name, suffix); };
+  }
+
+  utils::Regex sid_matcher_regex{*sid_matcher};
+  return [sid_matcher_regex](std::string_view field_name) { return utils::regexMatch(field_name, sid_matcher_regex); };
+}
+}  // namespace org::apache::nifi::minifi::wel
+
 namespace org::apache::nifi::minifi::processors {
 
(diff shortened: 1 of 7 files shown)


extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -54,9 +50,16 @@
...
 using namespace std::literals::chrono_literals;
 
-namespace org::apache::nifi::minifi::processors {
+namespace {
+struct TimeDiff {
+  auto operator()() const {
+    return std::chrono::steady_clock::now() - time_;
+  }
+  const decltype(std::chrono::steady_clock::now()) time_ = std::chrono::steady_clock::now();
+};
```
szaszm: I'd do it like this, but you don't need to change it if you don't like it. I started thinking because I found the decltype thing weird.
```suggestion
const auto start_timer = [] {
  return [start_time = std::chrono::steady_clock::now()] {
    return std::chrono::steady_clock::now() - start_time;
  };
};
```

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
```diff
@@ -95,14 +97,36 @@ void ConsumeWindowsEventLog::initialize() {
...
 
-  if (name != wel::METADATA::UNKNOWN) {
-    header.emplace_back(std::make_pair(name, value));
-    return true;
+  auto key_value_pairs = utils::string::split(*event_header_property, ",");
+  for (const auto& key_value_pair : key_value_pairs) {
+    auto key_value = utils::string::split(key_value_pair, "=");
+    if (key_value.size() == 2) {
+      auto key = utils::string::trim(key_value.at(0));
+      auto value = utils::string::trim(key_value.at(1));
```
szaszm: I see lots of temporary string allocations and deallocations, I think we should minimize those. std::string_view and working with views could be one way.

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
fgerlits: done, with a small modification (I changed the outer lambda to a function): a10684bfcdaf1a55a37cf83d480203ef7b70f264

extensions/windows-event-log/ConsumeWindowsEventLog.cpp
fgerlits: done: 7c023eaea16bfa3356b61c5e897a7d0fdf19b585

PR #2010, opened 2025-08-14
MINIFICPP-2608 Refresh outdated linter cmake target

martinzink:

### For all changes:
     in the commit message?

### For code changes:

### Note:


.github/workflows/ci.yml
```diff
@@ -535,32 +535,18 @@ jobs:
         with:
           name: behavex_output
           path: build/behavex_output
-  cpplint:
-    name: "cpplint"
+  linters:
```
lordgamez, line 538: I don't think it's a good idea to merge the linters in a single job, it's better to see the results of them separately. If multiple of them fails only 1 CI run is needed to see all of them, otherwise you will only discover the second failure in the second CI run after fixing the first issue.

martinzink: Didnt think of that, makes sense. How about this? https://github.com/apache/nifi-minifi-cpp/pull/2010/commits/16b91ee67184d2dab5f228aa618e991deb2d6398
This should run all 3 before failing so we can see the outputs but doesnt require multiple jobs

lordgamez: Looks good, thanks!

lordgamez (approved):

PR #2019, opened 2025-09-02
MINIFICPP-2619 Fix deadlock of ThreadPool shutdown

lordgamez:
The test case "Timer driven should respect both yield and run schedule" of SchedulingAgentTests hung sometimes due to a race condition in the ThreadPool class. It could occur that while shutting down, the notify_all call would go out and after that the delayed scheduler thread would still try to process the worker queue one more time then wait for the notify forever. Moving the running_ variable check under the mutual mutex solves the issue.

https://issues.apache.org/jira/browse/MINIFICPP-2619

---------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/core-framework/src/utils/ThreadPool.cpp
+++ b/core-framework/src/utils/ThreadPool.cpp
@@ -98,8 +98,11 @@ void ThreadPool::run_tasks(const std::shared_ptr<WorkerThread>& thread) {
 }
 
 void ThreadPool::manage_delayed_queue() {
-  while (running_) {
+  while (true) {
     std::unique_lock<std::mutex> lock(worker_queue_mutex_);
+    if (!running_) {
+      return;
+    }
 
     // Put the tasks ready to run in the worker queue
     while (!delayed_worker_queue_.empty() && delayed_worker_queue_.top().getNextExecutionTime() <= std::chrono::steady_clock::now()) {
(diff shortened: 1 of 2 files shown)


core-framework/src/utils/ThreadPool.cpp
fgerlits, line 1: Do we need to make the same change in `manageWorkers()`?

Also: if we only read/write `running_` under the mutex, do we still need it to be `atomic`?

Also: `adjust_threads_` seems to be unused.

fgerlits: It's OK if you want to keep this change minimal, and someone will clean up this class later...

lordgamez: I think we don't need to do this under the `manageWorkers()` because there is no conditional variable waiting to be notified, so this issue won't happen there, the worst that could happen that the loop will run one more time after the thread pool is stopped, but the shutdown will wait for it to join.

The `running_` variable seems to be still used in other occurrences without a mutex, and as there are 2 mutexes here under which we also use the `running_` variable it could still happen that it is used by multiple threads so I would keep it as an atomic.

It's a good pont that `adjust_threads_` is unused, I removed it in https://github.com/apache/nifi-minifi-cpp/pull/2019/commits/6931a54bf6e7e118f95fb162a4a569590ebcabc7

fgerlits (approved):

szaszm (approved):

PR #2020, opened 2025-09-03
MINIFICPP-2613 - Move base metrics to libminifi

adamdebreceni:

### For all changes:
     in the commit message?

### For code changes:

### Note:


core-framework/include/core/ProcessorImpl.h
```diff
@@ -143,7 +142,7 @@ class ProcessorImpl : public virtual ProcessorApi {
 
   std::atomic<bool> trigger_when_empty_;
 
-  gsl::not_null<std::shared_ptr<ProcessorMetrics>> metrics_;
+  std::shared_ptr<CustomProcessorMetrics> metrics_;
```
fgerlits: I would rename this to `custom_metrics_`, and also rename `getMetrics()` at line 121 to `getCustomMetrics()`, to distinguish it better from the `metrics_` field in the `Processor` class.

libminifi/src/core/Processor.cpp
```diff
@@ -531,10 +535,14 @@ annotation::Input Processor::getInputRequirement() const {
 }
 
 state::response::SharedResponseNode Processor::getResponseNode() {
-  return getMetrics();
+  return metrics_;
 }
```
fgerlits, line 538: It looks to me like this function used to return all metrics (standard + custom), but it only returns the standard metrics now.

minifi-api/include/minifi-cpp/core/CustomProcessorMetrics.h
```diff
@@ -0,0 +1,31 @@
...
+#pragma once
+
+#include <vector>
+
+#include "minifi-cpp/core/state/Value.h"
+#include "minifi-cpp/core/state/PublishedMetricProvider.h"
+
+namespace org::apache::nifi::minifi::core {
+
+class CustomProcessorMetrics : public virtual state::PublishedMetricProvider {
```
fgerlits: nitpicking, but can this `virtual` be removed?

core-framework/src/core/ProcessorImpl.cpp
```diff
@@ -47,7 +46,7 @@ namespace org::apache::nifi::minifi::core {
 ProcessorImpl::ProcessorImpl(ProcessorMetadata metadata)
     : metadata_(std::move(metadata)),
       trigger_when_empty_(false),
-      metrics_(std::make_shared<ProcessorMetricsImpl>(*this)),
+      metrics_(nullptr),
```
lordgamez: this can be removed, as the shared_ptr will always be initialized with nullptr

libminifi/include/core/ProcessorMetrics.h
```diff
@@ -35,42 +34,42 @@ concept Summable = requires(T x) { x + x; };  // NOLINT(readability/braces)
...
-  std::atomic<uint64_t>& transferredBytes() override {return transferred_bytes_;}
-  const std::atomic<uint64_t>& transferredBytes() const override {return transferred_bytes_;}
-  std::atomic<uint64_t>& bytesRead() override {return bytes_read_;}
-  const std::atomic<uint64_t>& bytesRead() const override {return bytes_read_;}
-  std::atomic<uint64_t>& bytesWritten() override {return bytes_written_;}
-  const std::atomic<uint64_t>& bytesWritten() const override {return bytes_written_;}
-  std::atomic<uint64_t>& processingNanos() override {return processing_nanos_;}
-  const std::atomic<uint64_t>& processingNanos() const override {return processing_nanos_;}
+  std::vector<state::response::SerializedResponseNode> serialize() final;
+  std::vector<state::PublishedMetric> calculateMetrics() final;
```
lordgamez, line 46: This class still implements the interface of `CustomProcessorMetrics`, so I think we should keep the inheritance. I think the problem is actually with the name of the `CustomProcessorMetrics` as it indicates to be a specialization of a processor metrics class while this is the other way around. I think it should be renamed to something like `ProcessorMetricsCollector`, but I'm open to other suggestions. It seems logical that the `ProcessorMetrics` and the custom processor metrics also implement the same interface.

core-framework/include/core/ProcessorImpl.h
adamdebreceni: renamed these to `metrics_extension_` and `getMetricsExtension()` in line with the renaming of the interface to `ProcessorMetricsExtension`

libminifi/src/core/Processor.cpp
adamdebreceni: the `metrics_` object not only manages the standard metrics but also injects the metrics extensions during `serialize()` and `calculateMetrics()`

minifi-api/include/minifi-cpp/core/CustomProcessorMetrics.h
adamdebreceni: removed

core-framework/src/core/ProcessorImpl.cpp
adamdebreceni: removed

libminifi/include/core/ProcessorMetrics.h
adamdebreceni: as we discussed, I have renamed `CustomProcessorMetrics` to `ProcessorMetricsExtension`

PR #2022, opened 2025-09-03
MINIFICPP-2305 Fix various C2 integration test issues

lordgamez:
- Use separate temporary files for each integration test for flow config files, because they are being written after parsed and can cause flaky test failures due to concurrent writes
- Only copy test resources once
- Delete all remaining temp dirs after test runs

https://issues.apache.org/jira/browse/MINIFICPP-2305

------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/test/libtest/integration/IntegrationBase.cpp
+++ b/libminifi/test/libtest/integration/IntegrationBase.cpp
@@ -35,6 +35,8 @@ IntegrationBase::IntegrationBase(const std::optional<std::filesystem::path>& tes
       home_path_(home_path) {
   flow_config_path_.config_path = test_file_location;
   if (test_file_location && std::filesystem::exists(*test_file_location) && std::filesystem::is_regular_file(*test_file_location)) {
+    // The original configuration file is changed, because after the flow configuration is parsed, the sensitive values are encrypted, and the new configuration is written back to the file
+    // We create a temporary copy of the flow configuration file for the test harness to use, so that multiple tests using the same flow configuration can run in parallel
     flow_config_path_.temp_dir = std::make_unique<TempDirectory>();
     flow_config_path_.config_path = flow_config_path_.temp_dir->getPath() / "config.yml";
     std::filesystem::copy_file(*test_file_location, *flow_config_path_.config_path);


libminifi/test/libtest/integration/IntegrationBase.cpp
szaszm, line 1: I don't follow what's happening here. Why do we reuse the last flow config path? Isn't the purpose of the change to _not_ use the same path between tests?

lordgamez: The problem is that in some scenarios we use the same flow configuration file from different test processes and when the `FlowController` loads the configuration it encrypts the sensitive values and writes the encrypted configuration back to the file. Because of this when multiple test processes write the same file there can be a concurrency issues, usually the issue is one test reads the configuration file while the other test writes it and it sees the file being empty which causes the test failure. This fix copies the input flow configuration for each test process to a temporary directory where it is independently read and written for the test. 

We do not reuse the last flow config path, just store it for checking it later on in some assertions like in the C2ConfigEncryption test. Usually in these C2 integration tests the `run()` is only called once for the harness, but if a new `run()` would be called on the harness, then the last flow config path would be replaced with the input `test_file_path`, the old temporary directory would be deleted and a new one created.

fgerlits, line 1: Do we ever use the same harness object from more than one thread? The name `last_flow_config_path_` is confusing if it is only ever set once.

Could the harness object create the temp directory and copy the config file (and resources?) there during construction, and just use it without copying in `run()`? That would seem simpler.

lordgamez: I think we don't use it more than once, but I have to check all cases first. The reason it was done this way is that the `test_file_path` is the parameter of the `run()`, but if it is only used once in all use cases then it can probably be refactored that the parameter is moved to the constructor.

lordgamez: I refactored the harness classes to move the config path to the constructor in https://github.com/apache/nifi-minifi-cpp/pull/2022/commits/8ed160179e988c5144990861d4b1d8ba5fa3d2e0

extensions/standard-processors/tests/integration/TailFileIntegrationTest.cpp
```diff
@@ -36,7 +36,8 @@ namespace org::apache::nifi::minifi::test {
 
 class TailFileTestHarness : public IntegrationBase {
  public:
-  TailFileTestHarness() {
+  explicit TailFileTestHarness(const std::filesystem::path& test_file_location)
+      : IntegrationBase(test_file_location, {}, 2s) {
```
fgerlits: is the addition of the `2s` argument here intentional?

libminifi/test/libtest/integration/IntegrationBase.cpp
```diff
@@ -28,9 +29,16 @@
...
 namespace org::apache::nifi::minifi::test {
 
-IntegrationBase::IntegrationBase(std::chrono::milliseconds waitTime)
+IntegrationBase::IntegrationBase(const std::optional<std::filesystem::path>& test_file_location, const std::optional<std::filesystem::path>& home_path, std::chrono::milliseconds waitTime)
     : configuration(std::make_shared<minifi::ConfigureImpl>()),
-      wait_time_(waitTime) {
+      wait_time_(waitTime),
+      home_path_(home_path) {
+  flow_config_path_.config_path = test_file_location;
+  if (test_file_location && std::filesystem::exists(*test_file_location) && std::filesystem::is_regular_file(*test_file_location)) {
```
fgerlits, line 37: I think we can throw here if this condition is false; this should never happen.

libminifi/test/libtest/integration/IntegrationBase.h
```diff
@@ -31,26 +31,35 @@
...
 
+struct FlowConfigPath {
+  std::unique_ptr<TempDirectory> temp_dir;
+  std::optional<std::filesystem::path> config_path;
+};
+
 class IntegrationBase {
  public:
-  explicit IntegrationBase(std::chrono::milliseconds waitTime = std::chrono::milliseconds(DEFAULT_WAITTIME_MSECS));
+  explicit IntegrationBase(const std::optional<std::filesystem::path>& test_file_location = {}, const std::optional<std::filesystem::path>& home_path = {},
```
fgerlits, line 49: Does `test_file_location` (and `flow_config_path_.config_path`) still need to be an optional? We seem to set it everywhere. Could we do this?
```suggestion
  explicit IntegrationBase(const std::filesystem::path& test_file_location, const std::optional<std::filesystem::path>& home_path = {},
```

extensions/standard-processors/tests/integration/TailFileIntegrationTest.cpp
lordgamez: Good catch, removed in https://github.com/apache/nifi-minifi-cpp/pull/2022/commits/4d33ad4abe767c3f8b256b4989fe6ee439fd17d2

libminifi/test/libtest/integration/IntegrationBase.cpp
lordgamez: There are still some tests where no flow file path is used, also there are some tests where the test file is created later, so it may not be present yet when the harness is created.

libminifi/test/libtest/integration/IntegrationBase.h
lordgamez: There are still some tests like C2UpdateAssetTest, C2LogHeartbeatTest, C2CompressTest, C2AssetSyncTest where no flow config file is used so it can remain optional

fgerlits: OK, got it. I didn't see any in the diff, but I guess those tests did not require a change, so they're not in the diff.

fgerlits (approved):

libminifi/test/libtest/integration/IntegrationBase.cpp
szaszm: I think a comment explaining this logic next to the relevant code parts would be nice. Currently it looks like random file checks and copies to someone with no mental context.

lordgamez: Good point, added a comment in https://github.com/apache/nifi-minifi-cpp/pull/2022/commits/4f25f001458ecb328fc760558ba4fe9382371b8a

PR #2029, opened 2025-09-15
MINIFICPP-2629 Clean up minifi-api includes

lordgamez:
- Remove duplicate header files with the same content
- Remove proxy header files in core/extension-framework that only include minifi-api headers
- Include only the minifi-api headers when implementation is not used
- Additionally fix clang-tidy issues in these files

https://issues.apache.org/jira/browse/MINIFICPP-2629

--------

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/standard-processors/processors/PutUDP.cpp
```diff
@@ -90,7 +90,7 @@ void PutUDP::onTrigger(core::ProcessContext& context, core::ProcessSession& sess
     for (const auto& resolver_entry : resolved_query) {
       error.clear();
       udp::socket socket(io_context);
-      socket.open(resolver_entry.endpoint().protocol(), error);
+      std::ignore = socket.open(resolver_entry.endpoint().protocol(), error);
```
adamdebreceni, line 93: not related to this PR but we could check if our asio usage could use ASIO_NO_DEPRECATED=1, we seem to set it for couchbase

adamdebreceni: my bad, we remove it with the patch

martinzink (approved):

core-framework/src/utils/SystemCpuUsageTracker.cpp
```diff
@@ -76,7 +76,7 @@ double SystemCpuUsageTracker::getCpuUsageBetweenLastTwoQueries() const {
   if (total_diff + total_idle_diff == 0) {
     return -1.0;
   }
-  double percent = static_cast<double>(total_diff) / static_cast<double>(total_diff + total_idle_diff);
+  double percent = static_cast<double>(total_diff) / static_cast<double>(total_diff + total_idle_diff);  // NOLINT(clang-analyzer-optin.taint.TaintedDiv)
```
szaszm, line 80: I'd rather disable this check globally. If it can't prove from the previous check that the denominator is not zero, then it's mostly useless and not worth the overhead of maintaining NOLINT comments.

core-framework/src/utils/SystemCpuUsageTracker.cpp
```diff
@@ -44,7 +44,7 @@ void SystemCpuUsageTracker::queryCpuTimes() {
   previous_total_sys_ = total_sys_;
   previous_total_idle_ = total_idle_;
   gsl::owner<FILE*> file = fopen("/proc/stat", "r");
-  if (fscanf(file, "cpu %lu %lu %lu %lu", &total_user_, &total_user_low_, &total_sys_, &total_idle_) != 4) {  // NOLINT(cert-err34-c)
+  if (fscanf(file, "cpu %lu %lu %lu %lu", &total_user_, &total_user_low_, &total_sys_, &total_idle_) != 4) {  // NOLINT(cert-err34-c,clang-analyzer-unix.Stream)
```
szaszm: This warning looks valid to me. We should handle the NULL case.
```suggestion
  if (!file) { return; }
  if (fscanf(file, "cpu %lu %lu %lu %lu", &total_user_, &total_user_low_, &total_sys_, &total_idle_) != 4) {  // NOLINT(cert-err34-c)
```

extensions/kafka/PublishKafka.cpp
```diff
@@ -112,7 +112,7 @@ class PublishKafka::Messages {
   }
 
   template<typename Func>
-  auto modifyResult(size_t index, Func fun) -> decltype(fun(flow_files_.at(index))) {
+  auto modifyResult(size_t index, const Func& fun) -> decltype(fun(flow_files_.at(index))) {
```
szaszm, line 115: why was this changed? if there is no good justification, I'd change it back.

.clang-tidy
```diff
@@ -1,6 +1,7 @@
 Checks: >
   -clang-analyzer-optin.cplusplus.VirtualCall,
   -clang-analyzer-optin.core.EnumCastOutOfRange,
+  -clang-analyzer-unix.BlockInCriticalSection,
```
szaszm, line 4: why was this disabled?

extensions/sftp/processors/ListSFTP.cpp
```diff
@@ -690,7 +690,7 @@ void ListSFTP::listByTrackingEntities(
 
   time_t now = time(nullptr);
   uint64_t min_timestamp_to_list = (!initial_listing_complete_ && entity_tracking_initial_listing_target_ == ENTITY_TRACKING_INITIAL_LISTING_TARGET_ALL_AVAILABLE)
-      ? 0U : (now * 1000 - entity_tracking_time_window.count());
+      ? 0U : ((now * 1000) - entity_tracking_time_window.count());
```
szaszm: disagree with the extra parens: multiplication has higher precedence, it's not confusing, and it's on the left side anyway, so it's the same even in left to right order.

extensions/standard-processors/processors/AttributeRollingWindow.cpp
```diff
@@ -113,7 +113,7 @@ void AttributeRollingWindow::calculateAndSetAttributes(core::FlowFile &flow_file
   }());
   // https://math.stackexchange.com/questions/1720876/sums-of-squares-minus-square-of-sums
   const auto avg_of_squares = std::accumulate(std::begin(sorted_values), std::end(sorted_values), 0.0, [&](double acc, double value) {
-    return acc + std::pow(value, 2) / gsl::narrow_cast<double>(sorted_values.size());
+    return acc + (std::pow(value, 2) / gsl::narrow_cast<double>(sorted_values.size()));
```
szaszm: once again I disagree with the parens: division has higher precedence than addition, and I don't think anyone finds that confusing. same as in math

extensions/standard-processors/processors/SplitText.cpp
```diff
@@ -53,7 +53,7 @@ uint8_t LineReader::getEndLineSize(size_t newline_position) {
 }
 
 void LineReader::setLastLineInfoAttributes(uint8_t endline_size, const std::optional<std::string>& starts_with) {
-  const uint64_t size_from_beginning_of_stream = (current_buffer_count_ - 1) * buffer_size_ + buffer_offset_;
+  const uint64_t size_from_beginning_of_stream = ((current_buffer_count_ - 1) * buffer_size_) + buffer_offset_;
```
szaszm: same thing about parens here

extensions/standard-processors/tests/unit/SplitTextTests.cpp
```diff
@@ -81,7 +81,7 @@ TEST_CASE("Test LineReader with input larger than buffer length") {
 
 TEST_CASE("Test LineReader with input of same size as buffer length") {
   auto stream = std::make_shared<io::BufferStream>();
-  std::string input = std::string(BUFFER_SIZE - 1, 'a') + "\n" + std::string(BUFFER_SIZE * 2 - 1, 'b') + "\n";
+  std::string input = std::string(BUFFER_SIZE - 1, 'a') + "\n" + std::string((BUFFER_SIZE * 2) - 1, 'b') + "\n";
```
szaszm: same thing about parens here

extensions/standard-processors/utils/JoltUtils.h
```diff
@@ -58,7 +58,7 @@ class Spec {
       return nullptr;
     }
 
-    template<std::invocable<std::shared_ptr<core::logging::Logger>> OnEnterFn, std::invocable<std::shared_ptr<core::logging::Logger>> OnExitFn>
+    template<std::invocable<const std::shared_ptr<core::logging::Logger>&> OnEnterFn, std::invocable<const std::shared_ptr<core::logging::Logger>&> OnExitFn>
```
szaszm: disagree with this change: Whether the argument is const-ref or not doesn't make a difference from the standpoint of being able to call a function with it or not.

extensions/systemd/ConsumeJournald.cpp
```diff
@@ -78,7 +78,7 @@ void ConsumeJournald::onSchedule(core::ProcessContext& context, core::ProcessSes
     return process_old_messages ? journal.seekHead() : journal.seekTail();
   };
   worker_->enqueue([this, &seek_default] {
-    const auto cursor = state_manager_->get() | utils::transform([](std::unordered_map<std::string, std::string>&& m) { return m.at(CURSOR_KEY); });
+    const auto cursor = state_manager_->get() | utils::transform([](const std::unordered_map<std::string, std::string>& m) { return m.at(CURSOR_KEY); });
```
szaszm, line 81: why was this changed? we can capture a temporary by either reference type, but it's an rvalue, so an rvalue reference seems better fitting. It would allow moving out the cursor value if unordered_map had an rvalue qualified overload of at.

libminifi/test/unit/ResponseNodeLoaderTests.cpp
```diff
@@ -49,7 +50,7 @@ class ResponseNodeLoaderTestFixture {
   }
 
  protected:
-  template<typename T, typename = typename std::enable_if_t<std::is_base_of_v<minifi::core::ProcessorImpl, T>>>
+  template<typename T> requires(std::is_base_of_v<minifi::core::ProcessorImpl, T>)
```
szaszm: unnecessary parens

core-framework/src/utils/SystemCpuUsageTracker.cpp
lordgamez: When clang tidy was introduced I remember that it found cases where 0 division was a possible issue, so I wouldn't disable it for 1 false positive if it can help in other cases.

extensions/kafka/PublishKafka.cpp
lordgamez: Clang tidy warns for every unnecessary copy and suggests const ref instead.

.clang-tidy
lordgamez: As far as I remember there was an issue with a system call in a loop which did make sense in the context, and I think we should be able to have more judgement around these performance issues than the checker, so I didn't find this useful.

extensions/sftp/processors/ListSFTP.cpp
lordgamez: We should decide if we want to follow this check at all and explicitly put all parentheses in expressions or just disable this check altogether. I suppose it is better to judge ourselves which expression is complex enough to require explicit parentheses so we can disable this check.

extensions/standard-processors/utils/JoltUtils.h
lordgamez: This was again the check that suggests using cont refs instead of copying for the lambda parameters. I think it's better in this case to avoid increasing the shared pointer ref counter when it's unnecessary.

extensions/systemd/ConsumeJournald.cpp
lordgamez: clang tidy warns for using rvalue references for parameters that are not moved in the function body

extensions/sftp/processors/ListSFTP.cpp
szaszm: I didn't realize this was another clang-tidy check. Yes, please disable it. Putting parens around everything makes no sense, precedence is a well known and useful thing.

extensions/standard-processors/utils/JoltUtils.h
szaszm: This change doesn't touch the shared pointers, it just checks the interface of the callable. Whether it gets increased or not depends on the function signature, not the concept checks.

extensions/systemd/ConsumeJournald.cpp
szaszm: I forgot about the fact that the effective type of rvalue ref parameter variables within the function body is mutable lvalue refs. I guess it's fine either way.

core-framework/src/utils/SystemCpuUsageTracker.cpp
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/2029/commits/cf510446553c2545f87c74e4f3f28b46deaa3df4

extensions/sftp/processors/ListSFTP.cpp
lordgamez: Removed check and updated in https://github.com/apache/nifi-minifi-cpp/pull/2029/commits/cf510446553c2545f87c74e4f3f28b46deaa3df4

extensions/standard-processors/processors/AttributeRollingWindow.cpp
lordgamez: Removed check and updated in https://github.com/apache/nifi-minifi-cpp/pull/2029/commits/cf510446553c2545f87c74e4f3f28b46deaa3df4

extensions/standard-processors/processors/SplitText.cpp
lordgamez: Removed check and updated in https://github.com/apache/nifi-minifi-cpp/pull/2029/commits/cf510446553c2545f87c74e4f3f28b46deaa3df4

extensions/standard-processors/tests/unit/SplitTextTests.cpp
lordgamez: Removed check and updated in https://github.com/apache/nifi-minifi-cpp/pull/2029/commits/cf510446553c2545f87c74e4f3f28b46deaa3df4

extensions/standard-processors/utils/JoltUtils.h
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/2029/commits/cf510446553c2545f87c74e4f3f28b46deaa3df4

libminifi/test/unit/ResponseNodeLoaderTests.cpp
lordgamez: Removed check and updated in https://github.com/apache/nifi-minifi-cpp/pull/2029/commits/cf510446553c2545f87c74e4f3f28b46deaa3df4

.clang-tidy
szaszm: I've checked this: it warns for the read in FileMutex, and asio internals whenever networking is used. None of the warnings seemed useful: FileMutex is inherently meant to interact with files atomically (achieved with the mutex), and asio internal mutexes are not our responsibility.

PR #2035, opened 2025-09-22
MINIFICPP-2636 update ExecutePythonProcessor description

szaszm:
- onTrigger accepts session, not property
- justifying basic behavior with efficiency seems weird, removed
- remove the sentence about concurrency, since the processor is single threaded

On a somewhat unrelated note, should we drop this API in favor of the newer NiFi-compatible python scripting API, or is it still useful to maintain? I guess it's easier to write processors for the legacy API, but the new API seems better designed, plus it handles dependencies. If we drop it, it should be done before 1.0.

-----

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/PROCESSORS.md
+++ b/PROCESSORS.md
@@ -722,7 +722,7 @@ In the list below, the names of required properties appear in bold. Any other pr
 
 ### Description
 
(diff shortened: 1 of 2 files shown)


lordgamez:
On the question regarding the API, I think we should keep it for now. We still had users recently using the old API and I would not want to break those processors as it is not trivial to move the old processors to the new API. Keeping it does not require much more maintenance as the new API is mostly a wrapper around the old one, so most of the codebase would still remain even if the support for the old API is removed. But I agree that we should encourage people to use the common NiFi style python API as it is better designed and also portable between NiFi projects.

PR #2036, opened 2025-09-24
MINIFICPP-2637 Upgrade libxml2 to v2.15.0

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2637

---------------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


cmake/AzureSdkCpp.cmake
```diff
@@ -55,3 +55,8 @@ FetchContent_Declare(asdkext
 )
 
 FetchContent_MakeAvailable(asdkext)
+
+if (NOT WIN32)
+    add_dependencies(azure-storage-common LibXml2::LibXml2)
+    add_dependencies(azure-data-tables LibXml2::LibXml2)
+endif()
```
fgerlits, line 68: why is this not needed on Windows?

CMakeLists.txt
```diff
@@ -370,7 +370,7 @@ add_subdirectory(core-framework)
 add_subdirectory(extension-framework)
 add_subdirectory(libminifi)
 
-if (ENABLE_ALL OR ENABLE_AZURE)
+if (NOT WIN32 AND (ENABLE_ALL OR ENABLE_AZURE))
```
fgerlits: also here; why is Windows excluded?

cmake/AzureSdkCpp.cmake
lordgamez: In both azure storage and azure data tables target libxml2 is only used in non-windows builds:
```
if(WIN32)
    target_link_libraries(azure-storage-common PRIVATE bcrypt webservices)
else()
    find_package(LibXml2 REQUIRED)
    target_include_directories(azure-storage-common SYSTEM PRIVATE ${LIBXML2_INCLUDE_DIRS})
    target_link_libraries(azure-storage-common PRIVATE ${LIBXML2_LIBRARIES})
    find_package(OpenSSL REQUIRED)
    target_link_libraries(azure-storage-common PRIVATE OpenSSL::SSL OpenSSL::Crypto)
endif()
```

CMakeLists.txt
lordgamez: Only azure uses libxml2 as a thirdparty dependency and only in non-windows builds

cmake/LibXml2.cmake
szaszm, line 1: Now that it's changed to FetchContent, we should stop referencing libxml2 from the main CMakeLists.txt, and only include it from the modules that depend on it, i.e. the azure sdk. This is the usage pattern everywhere else.

cmake/LibXml2.cmake
```diff
@@ -0,0 +1,50 @@
...
+    "${libxml2_SOURCE_DIR}/include"
+    "${libxml2_BINARY_DIR}"
+    CACHE STRING "" FORCE)
+if (WIN32)
+    set(LIBXML2_LIBRARIES "${libxml2_BINARY_DIR}/libxml2s.lib" CACHE STRING "" FORCE)
+    set(LIBXML2_LIBRARY "${libxml2_BINARY_DIR}/libxml2s.lib" CACHE STRING "" FORCE)
+else()
+    set(LIBXML2_LIBRARIES "${libxml2_BINARY_DIR}/libxml2.a" CACHE STRING "" FORCE)
+    set(LIBXML2_LIBRARY "${libxml2_BINARY_DIR}/libxml2.a" CACHE STRING "" FORCE)
+endif()
```
szaszm, line 51: Did you test these with msbuild?

cmake/LibXml2.cmake
lordgamez: Good idea, updated in https://github.com/apache/nifi-minifi-cpp/pull/2036/commits/735bc93601f895b68a0c377adb20033e21784c07

cmake/LibXml2.cmake
lordgamez: Good point, I checked and there's a slight difference in the path using msbuild, updated in https://github.com/apache/nifi-minifi-cpp/pull/2036/commits/735bc93601f895b68a0c377adb20033e21784c07

PR #2040, opened 2025-10-01
MINIFICPP-2643 Fix clang tidy issues

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2643

Depends on https://github.com/apache/nifi-minifi-cpp/pull/2029

--------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/core-framework/include/core/ClassLoader.h
+++ b/core-framework/include/core/ClassLoader.h
@@ -37,7 +37,7 @@ std::unique_ptr<T> ClassLoader::instantiate(const std::string &class_name, const
 }
 
 template<class T>
-T *ClassLoader::instantiateRaw(const std::string &class_name, const std::string &name) {
+gsl::owner<T*> ClassLoader::instantiateRaw(const std::string &class_name, const std::string &name) {
   return dynamic_cast<T*>(instantiateRaw(class_name, name, [] (CoreComponent* obj) -> bool {
     return dynamic_cast<T*>(obj) != nullptr;
   }));
--- a/core-framework/include/core/ObjectFactory.h
+++ b/core-framework/include/core/ObjectFactory.h
@@ -72,17 +72,25 @@ class DefaultObjectFactory : public ObjectFactoryImpl {
   /**
    * Create a raw pointer to a new processor.
    */
-  CoreComponent* createRaw(const std::string &name) override {
-    T *ptr = new T(name);
-    return dynamic_cast<CoreComponent*>(ptr);
+  gsl::owner<CoreComponent*> createRaw(const std::string &name) override {
+    const gsl::owner<T*> ptr = new T(name);
+    if (const auto converted = dynamic_cast<CoreComponent*>(ptr)) {
+      return converted;
+    }
+    delete ptr;
+    return nullptr;
   }
 
   /**
    * Create a raw pointer to a new processor.
    */
-  CoreComponent* createRaw(const std::string &name, const utils::Identifier &uuid) override {
-    T *ptr = new T(name, uuid);
-    return dynamic_cast<CoreComponent*>(ptr);
+  gsl::owner<CoreComponent*> createRaw(const std::string &name, const utils::Identifier &uuid) override {
+    const gsl::owner<T*> ptr = new T(name, uuid);
+    if (const auto converted = dynamic_cast<CoreComponent*>(ptr)) {
+      return converted;
+    }
+    delete ptr;
+    return nullptr;
   }
 
   /**
--- a/libminifi/src/core/ClassLoader.cpp
+++ b/libminifi/src/core/ClassLoader.cpp
@@ -53,7 +53,7 @@ class ClassLoaderImpl : public ClassLoader {
 
   [[nodiscard]] std::unique_ptr<CoreComponent> instantiate(const std::string &class_name, const std::string &name, const utils::Identifier &uuid, std::function<bool(CoreComponent*)> filter) override;
 
-  [[nodiscard]] CoreComponent* instantiateRaw(const std::string &class_name, const std::string &name, std::function<bool(CoreComponent*)> filter) override;
+  [[nodiscard]] gsl::owner<CoreComponent*> instantiateRaw(const std::string &class_name, const std::string &name, std::function<bool(CoreComponent*)> filter) override;
 
   ~ClassLoaderImpl() override = default;
 
@@ -116,13 +116,13 @@ class ProcessorFactoryWrapper : public ObjectFactoryImpl {
     return std::unique_ptr<CoreComponent>{createRaw(name, uuid)};
   }
 
-  [[nodiscard]] CoreComponent* createRaw(const std::string &name) override {
+  [[nodiscard]] gsl::owner<CoreComponent*> createRaw(const std::string &name) override {
     return createRaw(name, utils::IdGenerator::getIdGenerator()->generate());
   }
 
-  [[nodiscard]] CoreComponent* createRaw(const std::string &name, const utils::Identifier &uuid) override {
+  [[nodiscard]] gsl::owner<CoreComponent*> createRaw(const std::string &name, const utils::Identifier &uuid) override {
     auto logger = logging::LoggerFactoryBase::getAliasedLogger(getClassName(), uuid);
-    return new Processor(name, uuid, factory_->create({.uuid = uuid, .name = name, .logger = logger}));  // NOLINT(cppcoreguidelines-owning-memory)
+    return new Processor(name, uuid, factory_->create({.uuid = uuid, .name = name, .logger = std::move(logger)}));
   }
 
   [[nodiscard]] std::string getGroupName() const override {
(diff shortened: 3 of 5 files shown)


libminifi/test/libtest/integration/ConnectionCountingServer.cpp
```diff
@@ -85,14 +85,13 @@ void ReverseBodyPostHandler::saveConnectionId(struct mg_connection* conn) {
...
-    minifi::utils::SmallString<36>* id = new minifi::utils::SmallString<36>(minifi::utils::IdGenerator::getIdGenerator()->generate().to_string());
+    auto id = new minifi::utils::SmallString<36>(minifi::utils::IdGenerator::getIdGenerator()->generate().to_string());
     *user_connection_data = reinterpret_cast<void*>(id);
     return 0;
   };
 
   connection_close = [](const struct mg_connection* conn) -> void {
-    auto user_connection_data = reinterpret_cast<minifi::utils::SmallString<36>*>(mg_get_user_connection_data(conn));
-    delete user_connection_data;
+    std::ignore = std::unique_ptr<minifi::utils::SmallString<36>>(reinterpret_cast<minifi::utils::SmallString<36>*>(mg_get_user_connection_data(conn)));
```
fgerlits: it's a bit weird to create a `unique_ptr` only to destruct it immediately; I think keeping the `delete` with a NOLINT comment would be better

lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/2040/commits/69c49f015f6818bc26b2af897f5ccf8a9bed0475

fgerlits (approved):

extensions/python/PythonBindings.cpp
```diff
@@ -84,7 +84,7 @@ PyInit_minifi_native(void) {
     Py_INCREF(type.first);
   }
   const auto result = std::all_of(std::begin(types), std::end(types), [&](std::pair<PyTypeObject*, std::string_view> type) {
-    return PyModule_AddObject(minifi_module_instance, type.second.data(), reinterpret_cast<PyObject*>(type.first)) == 0;
+    return PyModule_AddObject(minifi_module_instance, type.second.data(), reinterpret_cast<PyObject*>(type.first)) == 0;  // NOLINT(bugprone-suspicious-stringview-data-usage)
```
szaszm: This warning seems valid: the usage seems to assume a null-terminated string, which a string_view is not guaranteed to be. We should add a comment explaining how it's guaranteed that the string_view is null-terminated, or fix the bug if it's not.

libminifi/include/core/state/UpdatePolicy.h
```diff
@@ -94,7 +89,7 @@ class UpdatePolicy {
 class UpdatePolicyBuilder {
  public:
   static std::unique_ptr<UpdatePolicyBuilder> newBuilder(bool enable_all = false) {
-    std::unique_ptr<UpdatePolicyBuilder> policy = std::unique_ptr<UpdatePolicyBuilder>( new UpdatePolicyBuilder(enable_all));
+    std::unique_ptr<UpdatePolicyBuilder> policy = std::unique_ptr<UpdatePolicyBuilder>( new UpdatePolicyBuilder(enable_all));  // NOLINT(clang-analyzer-cplusplus.NewDeleteLeaks)
```
szaszm, line 92: why not change to make_unique?

libminifi/src/c2/HeartbeatJsonSerializer.cpp
```diff
@@ -28,7 +28,7 @@ namespace org::apache::nifi::minifi::c2 {
 static void serializeOperationInfo(rapidjson::Value& target, const C2Payload& payload, rapidjson::Document::AllocatorType& alloc) {
   gsl_Expects(target.IsObject());
 
-  target.AddMember("operation", rapidjson::Value(magic_enum::enum_name<Operation>(payload.getOperation()).data(), alloc), alloc);
+  target.AddMember("operation", rapidjson::Value(magic_enum::enum_name<Operation>(payload.getOperation()).data(), alloc), alloc);  // NOLINT(bugprone-suspicious-stringview-data-usage)
```
szaszm, line 32: I'd add a comment explaining that the enum names are static null-terminated strings behind the string_view, and add a static_assert checking that for a few (or all) enum values, just as code documentation.

libminifi/test/unit/SiteToSiteTests.cpp
```diff
@@ -236,9 +236,8 @@ TEST_CASE("TestSiteToSiteVerifySend using flowfile data", "[S2S]") {
 TEST_CASE("TestSiteToSiteVerifyNegotiationFail", "[S2S]") {
   auto collector = std::make_unique<SiteToSiteResponder>();
 
-  const char negotiated_abort_code = magic_enum::enum_underlying(sitetosite::ResourceNegotiationStatusCode::NEGOTIATED_ABORT);
-  std::string resp_code;
-  resp_code.insert(resp_code.begin(), negotiated_abort_code);
+  const auto negotiated_abort_code = magic_enum::enum_underlying(sitetosite::ResourceNegotiationStatusCode::NEGOTIATED_ABORT);
+  const std::string resp_code{reinterpret_cast<const char*>(&negotiated_abort_code), 1};
```
szaszm: I'd prefer to do this with no reinterpret_cast, maybe like this:
```suggestion
  const std::string resp_code{static_cast<char>(negotiated_abort_code)};
```

libminifi/include/core/state/UpdatePolicy.h
lordgamez: The UpdatePolicyBuilder's constructor is protected and make_unique cannot access it.

extensions/python/PythonBindings.cpp
lordgamez: Replaced the std::string_view with std::string to make sure it is null-terminated in https://github.com/apache/nifi-minifi-cpp/pull/2040/commits/f66ce5c95061c24f82febd382375e4cec86c41fb

libminifi/src/c2/HeartbeatJsonSerializer.cpp
lordgamez: Added comment that magic_enum ensures null-terminated strings in this case in https://github.com/apache/nifi-minifi-cpp/pull/2040/commits/f66ce5c95061c24f82febd382375e4cec86c41fb Unfortunately I could not ensure it with a valid static_assert

libminifi/test/unit/SiteToSiteTests.cpp
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/2040/commits/f66ce5c95061c24f82febd382375e4cec86c41fb

libminifi/include/core/state/UpdatePolicy.h
szaszm: I see the bool ctor is public, and I see no reason not to make the move constructor public too. Why not do that and avoid an explicit new?

```suggestion
    auto policy = std::unique_ptr<UpdatePolicyBuilder>(UpdatePolicyBuilder(enable_all));
```

libminifi/include/core/state/UpdatePolicy.h
```diff
@@ -107,7 +102,7 @@ class UpdatePolicyBuilder {
   }
 
   std::unique_ptr<UpdatePolicy> build() {
-    std::unique_ptr<UpdatePolicy> new_policy = std::unique_ptr<UpdatePolicy>(new UpdatePolicy(*(current_policy_.get())));
+    std::unique_ptr<UpdatePolicy> new_policy = std::unique_ptr<UpdatePolicy>(new UpdatePolicy(*(current_policy_.get())));  // NOLINT(clang-analyzer-cplusplus.NewDeleteLeaks)
```
szaszm, line 105:
```suggestion
    auto new_policy = std::unique_ptr<UpdatePolicy>(UpdatePolicy(*current_policy_));
```

libminifi/include/core/state/UpdatePolicy.h
lordgamez: I think you are mixing up the `UpdatePolicyBuilder` and the `UpdatePolicy` classes. We are instantiating an `UpdatePolicyBuilder` here and its bool ctor is also protected on line 113.

libminifi/include/core/state/UpdatePolicy.h
lordgamez: Here the comment says "Defines an immutable policy map" where the copy constructor is protected on purpose, with the UpdatePolicyBuilder defined as a friend class. I'm not sure this is a good design or if it is still valid, but I think refactoring the `UpdatePolicy` and `UpdatePolicyBuilder` classes should be figured out separately and it is out of scope of this PR.

libminifi/include/core/state/UpdatePolicy.h
szaszm: You're right, I mixed them up. Would it be feasible to enable public move of UpdatePolicyBuilder, and make the change I suggested above?

szaszm (approved):

PR #2045, opened 2025-10-09
MINIFICPP-2644 Add proxy controller service support for AWS and Azure processors

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2644

-------------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


CONTROLLERS.md
```diff
@@ -244,6 +245,24 @@ In the list below, the names of required properties appear in bold. Any other pr
 | **File**                  |               |                  | Path to a file to store state                                                                                                                          |
 
 
+## ProxyConfigurationService
+
+### Description
+
+Provides a set of configurations for different MiNiFi C++ components to use a proxy server.
```
szaszm: The proxy protocol is not specified anywhere. Typical options include http, https, socks4, socks4a, socks5. We should document what this service uses and leave the door open for more protocols in the future.

minifi-api/include/minifi-cpp/controllers/ProxyConfigurationServiceInterface.h
```diff
@@ -0,0 +1,42 @@
...
+#include "minifi-cpp/core/ControllerServiceApiDefinition.h"
+
+namespace org::apache::nifi::minifi::controllers {
+
+struct ProxyConfiguration {
+  std::string proxy_host;
+  std::optional<uint16_t> proxy_port;
+  std::optional<std::string> proxy_user;
+  std::optional<std::string> proxy_password;
+};
```
szaszm: details like this should not be on the API. username/password can be optional and replaced by other auth methods in some other proxy protocols. Possibly even hostname and port can vary.

CONTROLLERS.md
lordgamez: Good point, we only support HTTP proxy at the moment. The controller is based on the NiFi counterpart https://nifi.apache.org/components/org.apache.nifi.proxy.StandardProxyConfigurationService/ where the proxy type can be specified if there are additional options. We can add that in the future if we want to support more protocols. Currently the cloud providers usually support only HTTP proxies, the newer AWS SDK should support HTTPS as well, we can introduce that later if needed. Added additional info to the description in https://github.com/apache/nifi-minifi-cpp/pull/2045/commits/2b46713093565c2e54d6e0bbea6ec8723679096c

minifi-api/include/minifi-cpp/controllers/ProxyConfigurationServiceInterface.h
lordgamez: As we only support HTTP proxies for now only these parameters are used, they can be extended or refactored when other protocols are introduced.

szaszm: If we add this type definition to the API, we can't change it later without breaking API.

lordgamez: I'm not sure what would be the best way to define this to adjust it to the future C API. In the current form I think it will only be extended, so non of the current field will be changed in the future. We should discuss this how these cases should be handled in the future, because I think additional fields being added to a struct on the API can be common. We can either define some additional `reserved` fields on the C API for future extensions, we can define some accessors on the C API so the struct itself will not be published to the API, or we can just break it when needed if we can predict that it will not be changed for a long time. I think we still have some time for think about this as controller services will not be supported on the C API for a while. What do you suggest in this case?

lordgamez: We could change this to have separate getter function for each field like we do on the SSLContextServiceInterface, in that way that can be extended when new properties are introduced, what do you think?

szaszm: I think a getter-only interface could work. I would also add a proxy type getter, even if the type enum only has one possible value today, so older processors can detect future versions of the service that are incompatible.

I think we could also do a flat struct with the first field being a type enum field, if the interface always returns a heap allocated object: that way future versions can grow the struct without breaking its layout. I'd prefer this approach.

lordgamez: I updated the interface to use a getter-only interface to be consistent with the SSLContextServiceInterface, I also added a type enum in https://github.com/apache/nifi-minifi-cpp/pull/2045/commits/12e4262eaac5fa2972484f30210ac73132e9de35

extensions/aws/processors/AwsProcessor.cpp
```diff
@@ -69,10 +69,21 @@ std::optional<Aws::Auth::AWSCredentials> AwsProcessor::getAWSCredentials(
...
-  proxy.password = minifi::utils::parseOptionalProperty(context, ProxyPassword, flow_file).value_or("");
+  auto proxy_controller_service = minifi::utils::parseOptionalControllerService<minifi::controllers::ProxyConfigurationServiceInterface>(context, ProxyConfigurationService, getUUID());
+  if (proxy_controller_service) {
+    proxy.host = proxy_controller_service->getHost();
+    auto port_opt = proxy_controller_service->getPort();
+    proxy.port = port_opt ? *port_opt : 0;
+    auto username_opt = proxy_controller_service->getUsername();
+    proxy.username = username_opt ? *username_opt : "";
+    auto password_opt = proxy_controller_service->getPassword();
+    proxy.password = password_opt ? *password_opt : "";
```
fgerlits: very minor, but why not
```suggestion
    proxy.host = proxy_controller_service->getHost();
    proxy.port = proxy_controller_service->getPort().value_or(0);
    proxy.username = proxy_controller_service->getUsername().value_or("");
    proxy.password = proxy_controller_service->getPassword().value_or("");
```
as it is done in the else branch?

libminifi/include/controllers/ProxyConfigurationService.h
```diff
@@ -0,0 +1,109 @@
...
+      : ControllerServiceImpl(name, uuid) {
+  }
+
+  MINIFIAPI static constexpr const char* Description = "Provides a set of configurations for different MiNiFi C++ components to use a proxy server. "
+      "Currently these properties can only be used for HTTP proxy configuration, not other protocols are supported at this time.";
+
+  MINIFIAPI static constexpr auto ProxyServerHost = core::PropertyDefinitionBuilder<>::createProperty("Proxy Server Host")
+      .withDescription("Proxy server hostname or ip-address.")
+      .isRequired(true)
+      .build();
```
fgerlits: this could have a `NON_BLANK_VALIDATOR` to catch some bad configurations (e.g. when the port is set but the host isn't)

CONTROLLERS.md
```diff
@@ -244,6 +245,24 @@ In the list below, the names of required properties appear in bold. Any other pr
 | **File**                  |               |                  | Path to a file to store state                                                                                                                          |
 
 
+## ProxyConfigurationService
+
+### Description
+
+Provides a set of configurations for different MiNiFi C++ components to use a proxy server. Currently these properties can only be used for HTTP proxy configuration, not other protocols are supported at this time.
```
fgerlits:
```suggestion
Provides a set of configurations for various MiNiFi C++ components to use a proxy server. Currently these properties can only be used for HTTP proxy configuration, no other protocols are supported at this time.
```

libminifi/include/controllers/ProxyConfigurationService.h
fgerlits, line 40: this property could have a NON_BLANK_VALIDATOR

extensions/aws/processors/AwsProcessor.cpp
lordgamez: Good point, updated in https://github.com/apache/nifi-minifi-cpp/pull/2045/commits/fa7ded8afdf472df14255ff2cbba4be47eed6271

CONTROLLERS.md
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/2045/commits/fa7ded8afdf472df14255ff2cbba4be47eed6271

libminifi/include/controllers/ProxyConfigurationService.h
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/2045/commits/fa7ded8afdf472df14255ff2cbba4be47eed6271

fgerlits (approved):

extensions/aws/processors/AwsProcessor.h
```diff
@@ -131,6 +132,11 @@ class AwsProcessor : public core::ProcessorImpl {  // NOLINT(cppcoreguidelines-s
       .withValidator(core::StandardPropertyValidators::NON_BLANK_VALIDATOR)
       .supportsExpressionLanguage(true)
       .build();
+  EXTENSIONAPI static constexpr auto ProxyType = core::PropertyDefinitionBuilder<magic_enum::enum_count<minifi::controllers::ProxyType>()>::createProperty("Proxy Type")
+      .withDescription("Proxy type")
+      .withDefaultValue(magic_enum::enum_name(minifi::controllers::ProxyType::HTTP))
+      .withAllowedValues(magic_enum::enum_names<minifi::controllers::ProxyType>())
+      .build();
```
szaszm: this belongs in the ProxyConfigurationService, not in the processors. In NiFi the same controller service specifies all proxy details, including the type/protocol.

extensions/azure/storage/DataLakeStorageClient.h
```diff
@@ -39,6 +41,7 @@ struct AzureDataLakeStorageParameters {
   std::string file_system_name;
   std::string directory_name;
   std::optional<uint64_t> number_of_retries;
+  std::optional<minifi::controllers::ProxyConfiguration> proxy_configuration;
```
szaszm: Instead of an optional member, it could default to type==DIRECT, which means no proxy. That would make one more impossible state unrepresentable, which is considered a best practice nowadays.

extensions/aws/processors/AwsProcessor.cpp
```diff
@@ -66,15 +66,25 @@ std::optional<Aws::Auth::AWSCredentials> AwsProcessor::getAWSCredentials(
...
+
+  auto proxy_controller_service = minifi::utils::parseOptionalControllerService<minifi::controllers::ProxyConfigurationServiceInterface>(context, ProxyConfigurationService, getUUID());
+  if (proxy_controller_service) {
+    proxy.proxy_type = proxy_controller_service->getProxyType();
+    proxy.proxy_host = proxy_controller_service->getHost();
+    proxy.proxy_port = proxy_controller_service->getPort().value_or(0);
+    proxy.proxy_user = proxy_controller_service->getUsername().value_or("");
+    proxy.proxy_password = proxy_controller_service->getPassword().value_or("");
+  } else {
+    proxy.proxy_type = minifi::utils::parseOptionalEnumProperty<minifi::controllers::ProxyType>(context, ProxyType).value_or(minifi::controllers::ProxyType::HTTP);
```
szaszm: We should add a DIRECT option and make it the default.

minifi-api/include/minifi-cpp/controllers/ProxyConfigurationServiceInterface.h
```diff
@@ -0,0 +1,44 @@
...
+#include "minifi-cpp/core/ControllerServiceApiDefinition.h"
+
+namespace org::apache::nifi::minifi::controllers {
+
+enum class ProxyType {
+  HTTP,
+  HTTPS
+};
+
+class ProxyConfigurationServiceInterface : public virtual core::controller::ControllerService {
```
szaszm: I'd keep the naming consistent with NiFi and call this ProxyConfigurationService

libminifi/include/controllers/ProxyConfigurationService.h
```diff
@@ -0,0 +1,116 @@
...
+
+#include "minifi-cpp/controllers/ProxyConfigurationServiceInterface.h"
+#include "controllers/ProxyConfiguration.h"
+#include "core/controller/ControllerService.h"
+#include "core/PropertyDefinitionBuilder.h"
+#include "minifi-cpp/core/PropertyValidator.h"
+
+namespace org::apache::nifi::minifi::controllers {
+
+class ProxyConfigurationService : public core::controller::ControllerServiceImpl, public ProxyConfigurationServiceInterface {
```
szaszm: I'd keep the naming consistent with NiFi and call this StandardProxyConfigurationService

extensions/azure/tests/DeleteAzureBlobStorageTests.cpp
```diff
@@ -339,4 +339,30 @@ TEST_CASE_METHOD(DeleteAzureBlobStorageTestsFixture, "Test Azure blob delete wit
   REQUIRE(failed_flowfiles[0] == TEST_DATA);
 }
 
+TEST_CASE_METHOD(DeleteAzureBlobStorageTestsFixture, "Test Azure blob delete using proxy", "[azureBlobStorageDelete]") {
```
szaszm: Since this logic is in shared code, I'd only add one unit test, not one for each processor.

extensions/aws/processors/AwsProcessor.h
lordgamez: It is also present in the controller service, but the proxy properties were already present in the AWS processors before introducing the proxy controller service, so this needs to be added here too. Maybe we can deprecate it in the future and remove them to only use the proxy controller service. In NiFi 2.0 they also removed the proxy properties leaving the proxy configuration service property. I would not suggest using the existing proxy properties for host, port, username, password, but require adding a controller service just for the proxy type.

extensions/azure/storage/DataLakeStorageClient.h
lordgamez: I'm not sure I'm convinced about this. What do you mean by "one more impossible state unrepresentable"? It seem a bit unintuitive to me. When the user adds a proxy controller service their intention is to use a proxy, if they do not want to use a proxy they should not add a service at all. What would be the benefit of adding a proxy service and configure it to not to use a proxy?

libminifi/include/controllers/ProxyConfigurationService.h
lordgamez: I'm not sure it's a good idea, I think we should be consistent with our own naming conventions. For example we have AzureStorageCredentialsService and SSLContextService, while these are named StandardAzureCredentialsControllerService and StandardSSLContextService in NiFi. I also think adding the "Standard" prefix just adds confusion for the user when there are no other proxy services so there is no need for the additional adjective.

extensions/azure/tests/DeleteAzureBlobStorageTests.cpp
lordgamez: I kept the tests for the Put and List processors as the List processors are a bit different and removed the others in https://github.com/apache/nifi-minifi-cpp/pull/2045/commits/f70ac212f7eefc06ab27f7471fd92a5c8f9f39ba

martinzink (approved):

extensions/azure/storage/DataLakeStorageClient.h
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/2045/commits/8890e9fe63b26c4f38dc59af04906e40c737b5cf

extensions/aws/processors/AwsProcessor.cpp
lordgamez: Added DIRECT option in https://github.com/apache/nifi-minifi-cpp/pull/2045/commits/8890e9fe63b26c4f38dc59af04906e40c737b5cf, but I kept the HTTP as the default as the main intention should be to use a proxy when a controller is added.

core-framework/include/controllers/ProxyConfiguration.h
```diff
@@ -18,16 +17,18 @@
...
-  std::string host;
-  uint32_t port = 0;
-  std::string username;
-  std::string password;
+struct ProxyConfiguration {
+  ProxyType proxy_type;
+  std::string proxy_host;
+  std::optional<uint16_t> proxy_port;
+  std::optional<std::string> proxy_user;
+  std::optional<std::string> proxy_password;
```
szaszm, line 31: I'd make proxy_port a raw uint16_t, it's always going to be set anyway, or we can use 0 as a default value if need be. Also let's move the username + password into a separate struct with not optional string fields, and have an optional<ProxyCredentials> here.

extensions/azure/storage/DataLakeStorageClient.h
szaszm: We can remove the optional member now

extensions/azure/tests/ListAzureDataLakeStorageTests.cpp
```diff
@@ -256,4 +256,27 @@ TEST_CASE_METHOD(ListAzureDataLakeStorageTestsFixture, "Both SAS Token and Stora
   REQUIRE_THROWS_AS(test_controller_.runSession(plan_, true), minifi::Exception);
 }
 
+TEST_CASE_METHOD(ListAzureDataLakeStorageTestsFixture, "List data lake storage files using proxy", "[azureDataLakeStorageParameters]") {
+  auto proxy_configuration_service = plan_->addController("ProxyConfigurationService", "ProxyConfigurationService");
+  plan_->setProperty(proxy_configuration_service, "Proxy Server Host", "https://host");
+  plan_->setProperty(proxy_configuration_service, "Proxy Server Port", "1234");
+  plan_->setProperty(proxy_configuration_service, "Proxy User Name", "username");
+  plan_->setProperty(proxy_configuration_service, "Proxy User Password", "password");
+  plan_->setProperty(proxy_configuration_service, "Proxy Type", "HTTPS");
```
szaszm: HTTPS is no longer supported, we should change it to HTTP

extensions/azure/storage/AzureDataLakeStorageClient.cpp
```diff
@@ -35,13 +35,24 @@ AzureDataLakeStorageClient::AzureDataLakeStorageClient() {
   utils::AzureSdkLogger::initialize();
 }
 
-std::unique_ptr<Azure::Storage::Files::DataLake::DataLakeFileSystemClient> AzureDataLakeStorageClient::createClient(
-    const AzureStorageCredentials& credentials, const std::string& file_system_name, std::optional<uint64_t> number_of_retries) {
+std::unique_ptr<Azure::Storage::Files::DataLake::DataLakeFileSystemClient> AzureDataLakeStorageClient::createClient(const AzureStorageCredentials& credentials,
+    const std::string& file_system_name, std::optional<uint64_t> number_of_retries, const std::optional<minifi::controllers::ProxyConfiguration>& proxy_configuration) {
```
szaszm: An optional non-owning reference can be more simply represented as a pointer
```suggestion
    const std::string& file_system_name, std::optional<uint64_t> number_of_retries, minifi::controllers::ProxyConfiguration* const proxy_configuration) {
```

extensions/azure/storage/DataLakeStorageClient.h
lordgamez: Removed optional in https://github.com/apache/nifi-minifi-cpp/pull/2045/commits/29da0bdc1203b7bf69d6813c1d8ebeec2f9763f1

core-framework/include/controllers/ProxyConfiguration.h
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/2045/commits/29da0bdc1203b7bf69d6813c1d8ebeec2f9763f1

extensions/azure/tests/ListAzureDataLakeStorageTests.cpp
lordgamez: Removed in https://github.com/apache/nifi-minifi-cpp/pull/2045/commits/29da0bdc1203b7bf69d6813c1d8ebeec2f9763f1

extensions/azure/storage/AzureDataLakeStorageClient.cpp
lordgamez: Removed optional so it is a const ref now in https://github.com/apache/nifi-minifi-cpp/pull/2045/commits/29da0bdc1203b7bf69d6813c1d8ebeec2f9763f1

szaszm (approved):

PR #2081, opened 2025-12-18
MINIFICPP-2696 Add option for optimizationForSmallDB in DatabaseConte…

martinzink:
I’ve went ahead and this some testing so we can set a sensible default value for this new configurable value. Based on my testing it seems the processing overhead is negligible, unfortunately the memory gains are also not what one expect (flow specific)

I’ve created a bunch of flows and tried it various cache sizes.

- Simple GenerateFlowFile → LogAttribute 5MB Text (random)
  - processing throughput remained the same while achieving 10% lower memory footprint on avarage, but the memory usage was not consistent
- TailFile(30k lines) → MergeFile → LogAttribute, with GetFile → LogAttribute (to increase memory usage)
  - Processing time of the tailfile remained within run-to-run variance with 20-30% lower memory footprint

I’ve also noticed that compaction time and periodic https://man7.org/linux/man-pages/man3/malloc_trim.3.html makes memory footprint even smaller (much more noticably in some cases)

Methodology
Processing time was measured via log messages, memory footprint was measured via prometheus AgentStatus metrics
Baseline was setting the new variable to an empty string (which basically mimics the current main behaviour)

**I couldn’t measure performance losses but only tried with a small set of relatively simple flows, since this PR now changes the default behavior its important that we emphasise this change and the method to disable this in the release logs.**

### DIFF
--- a/extensions/rocksdb-repos/DatabaseContentRepository.cpp
+++ b/extensions/rocksdb-repos/DatabaseContentRepository.cpp
@@ -58,37 +58,50 @@ bool DatabaseContentRepository::initialize(const std::shared_ptr<minifi::Configu
 
   setCompactionPeriod(configuration);
 
-  const auto cache_size = configuration->get(Configure::nifi_dbcontent_optimize_for_small_db_cache_size)
+  const auto cache_size = configuration->get(Configure::nifi_dbcontent_optimize_for_small_db_cache_size).or_else([] { return std::make_optional<std::string>("8 MB"); })
     | utils::andThen([](const auto& cache_size_str) -> std::optional<uint64_t> {
       return parsing::parseDataSize(cache_size_str) | utils::toOptional();
     });
 
   std::shared_ptr<rocksdb::Cache> cache = nullptr;
+  std::shared_ptr<rocksdb::WriteBufferManager> wbm = nullptr;
   if (cache_size) {
     cache = rocksdb::NewLRUCache(*cache_size);
+    wbm = std::make_shared<rocksdb::WriteBufferManager>(0, cache);
+    logger_->log_trace("Using {} sized cache for DatabaseContentRepository", *cache_size);
+  } else {
+    logger_->log_trace("Cache limitation disabled for DatabaseContentRepository");
   }
 
-  auto set_db_opts = [encrypted_env, &cache] (minifi::internal::Writable<rocksdb::DBOptions>& db_opts) {
+  auto set_db_opts = [encrypted_env, &cache, &wbm] (minifi::internal::Writable<rocksdb::DBOptions>& db_opts) {
     minifi::internal::setCommonRocksDbOptions(db_opts);
     if (encrypted_env) {
       db_opts.set(&rocksdb::DBOptions::env, encrypted_env.get(), EncryptionEq{});
     } else {
       db_opts.set(&rocksdb::DBOptions::env, rocksdb::Env::Default());
     }
-    if (cache) {
-      db_opts.call(&rocksdb::DBOptions::OptimizeForSmallDb, &cache);
-    }
+    db_opts.optimizeForSmallDb(cache, wbm);
   };
   auto set_cf_opts = [&configuration, &cache] (rocksdb::ColumnFamilyOptions& cf_opts) {
     cf_opts.OptimizeForPointLookup(4);
     cf_opts.merge_operator = std::make_shared<StringAppender>();
     cf_opts.max_successive_merges = 0;
-    if (auto compression_type = minifi::internal::readConfiguredCompressionType(configuration, Configure::nifi_content_repository_rocksdb_compression)) {
+    cf_opts.max_write_buffer_number = 2;
+    cf_opts.write_buffer_size = 4_MB;
+    if (const auto compression_type = internal::readConfiguredCompressionType(configuration, Configure::nifi_content_repository_rocksdb_compression)) {
       cf_opts.compression = *compression_type;
     }
     if (cache) {
-      cf_opts.OptimizeForSmallDb(&cache);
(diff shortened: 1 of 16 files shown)


extensions/rocksdb-repos/database/OpenRocksDb.cpp
szaszm, line 1: at this point we have 4 instances of copy-pasta, I think extracting the logic is warranted

libminifi/test/unit/LogMetricsPublisherTests.cpp
```diff
@@ -86,8 +85,49 @@ TEST_CASE_METHOD(LogPublisherTestFixture, "Verify empty metrics if no valid metr
...
+  std::string regex_pattern =
+      // 1. Log Header & Opening
+      R"(\[[\d\-\s\:\.]+\]\s*\[[^\]]+\]\s*\[)" + std::move(level) + R"(\]\s*\{\s*)" +
+      R"(\"LogMetrics\":\s*\{\s*)" +
+
+      // 2. (skips other nodes, stops if a \n[ starts (other log line))
+      R"((?:(?!\n\[)[\s\S])*?)" +
+
+      // 3. RepositoryMetrics Node
+      R"(\"RepositoryMetrics\":\s*\{\s*)" +
```
fgerlits: Why do we need to escape the `"` characters? This would be marginally less unreadable without these backslashes.

szaszm (approved):

libminifi/test/unit/LogMetricsPublisherTests.cpp
martinzink: refactored it because the huge regex matching was unfeasable on CI machines
[refactor LogMetricsPublisherTests so it doesn't really on huge regex_match](https://github.com/apache/nifi-minifi-cpp/pull/2081/changes/72bef5f063a2d873474d0de065e346a74845d9d7)

fgerlits (approved):

lordgamez (approved):

extensions/rocksdb-repos/database/RocksDbUtils.h
```diff
@@ -49,6 +49,18 @@ class Writable {
...
+  template <typename Method, typename... Args>
+  decltype(auto) call(Method method, Args&&... args) {
+    return std::invoke(method, target_, std::forward<Args>(args)...);
+  }
+
+  void optimizeForSmallDb(std::shared_ptr<rocksdb::Cache> cache, std::shared_ptr<rocksdb::WriteBufferManager> wbm) {
+    if (!cache || !wbm) { return; }
+    target_.OptimizeForSmallDb(&cache);
+    target_.write_buffer_manager = wbm;
+    target_.max_open_files = 20;
```
fgerlits, line 61: do we want to set `is_modified_` here?

extensions/rocksdb-repos/DatabaseContentRepository.cpp
```diff
@@ -58,21 +58,51 @@ bool DatabaseContentRepository::initialize(const std::shared_ptr<minifi::Configu
 
   setCompactionPeriod(configuration);
 
-  auto set_db_opts = [encrypted_env] (minifi::internal::Writable<rocksdb::DBOptions>& db_opts) {
+  const auto cache_size = configuration->get(Configure::nifi_dbcontent_optimize_for_small_db_cache_size).or_else([] { return std::make_optional<std::string>("8 MB"); })
+    | utils::andThen([](const auto& cache_size_str) -> std::optional<uint64_t> {
+      return parsing::parseDataSize(cache_size_str) | utils::toOptional();
+    });
```
fgerlits: This disables the cache whenever the configuration setting cannot be parsed, if I read it correctly.  So not only `""`, but also `"two million bytes"`, `"8 NB"` etc will lead to a silently disabled cache.

I think it would be better to only disable the cache if the configuration is set to `""` and fail to start up if it is set to an invalid value.

libminifi/test/libtest/unit/TestUtils.cpp
```diff
@@ -260,4 +260,43 @@ std::error_code sendMessagesViaSSL(const std::vector<std::string_view>& contents
   return {};
 }
 
+std::vector<LogMessageView> extractLogMessageViews(const std::string& log_str) {
+  std::vector<LogMessageView> messages;
+  const std::regex header_pattern(R"((\[[\d\-\s\:\.]+\]) (\s*\[[^\]]+\]) \[(.*)\])");
```
fgerlits: minor, but this will pick up garbage for the log level if the payload contains a `]` character; also I wouldn't include the `[]` characters in the timestamp and the class name:
```suggestion
  const std::regex header_pattern(R"(\[([\d\-\s\:\.]+)\]\s+\[(.*?)\]\s+\[(.*?)\])");
```

extensions/rocksdb-repos/database/RocksDbUtils.h
martinzink: if thats set to  modified it will refuse to load the repository so probably not

extensions/rocksdb-repos/DatabaseContentRepository.cpp
martinzink: good idea, [invalid nifi.database.content.repository.optimize.for.small.db.cache.…](https://github.com/apache/nifi-minifi-cpp/pull/2081/commits/b80ec78eaf15429ecfe28c5ae9ddda70e30645b0)

libminifi/test/libtest/unit/TestUtils.cpp
fgerlits: some tests may need to be updated, because the class name no longer contains the `[]` characters

szaszm (approved):

PR #2083, opened 2026-01-07
MINIFICPP-2700 Terminate if ContentRepository creation failed

martinzink:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/libminifi/include/core/RepositoryFactory.h
+++ b/libminifi/include/core/RepositoryFactory.h
@@ -32,8 +32,12 @@ namespace org::apache::nifi::minifi::core {
  * @param configuration_class_name configuration class name
  * @param fail_safe determines whether or not to make the default class if configuration_class_name is invalid
  * @param repo_name name of the repository
+ * @param logger pointer to a logger so it can log out errors
  */
-std::unique_ptr<core::ContentRepository> createContentRepository(const std::string& configuration_class_name, bool fail_safe = false, const std::string& repo_name = "");
+std::unique_ptr<core::ContentRepository> createContentRepository(const std::string& configuration_class_name,
+    bool fail_safe = false,
+    const std::string& repo_name = "",
+    logging::Logger* logger = nullptr);
 
 /**
  * Create a repository represented by the configuration class name
--- a/libminifi/include/core/RepositoryFactory.h
+++ b/libminifi/include/core/RepositoryFactory.h
@@ -27,23 +27,10 @@
 
 namespace org::apache::nifi::minifi::core {
 
-/**
- * Create a context repository
- * @param configuration_class_name configuration class name
- * @param fail_safe determines whether or not to make the default class if configuration_class_name is invalid
- * @param repo_name name of the repository
- * @param logger pointer to a logger so it can log out errors
- */
 std::unique_ptr<core::ContentRepository> createContentRepository(const std::string& configuration_class_name,
-    const std::string& repo_name = "",
-    logging::Logger* logger = nullptr);
+    const std::string& repo_name,
+    logging::Logger& logger);
 
-/**
- * Create a repository represented by the configuration class name
- * @param configuration_class_name configuration class name
- * @param fail_safe determines whether or not to make the default class if configuration_class_name is invalid
- * @param repo_name name of the repository
- */
 std::unique_ptr<core::Repository> createRepository(const std::string& configuration_class_name, const std::string& repo_name = "");
 
 }  // namespace org::apache::nifi::minifi::core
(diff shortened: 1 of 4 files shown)


libminifi/src/core/RepositoryFactory.cpp
szaszm, line 1: It may be better to inject a logger as a function parameter instead of creating one locally for a semi-related class.

martinzink: good idea, [review changes](https://github.com/apache/nifi-minifi-cpp/pull/2083/commits/e09d59e3bd3b8f325d070ca8c4bcdc147837c1aa)

libminifi/src/core/RepositoryFactory.cpp
```diff
@@ -44,6 +47,10 @@ std::unique_ptr<core::ContentRepository> createContentRepository(const std::stri
       return std::make_unique<core::repository::FileSystemRepository>(repo_name);
     }
     if (fail_safe) {
```
adamdebreceni: I am wondering if this is a good idea (not the change itself) if the user requests some persistent content repo and we log an error, store stuff in-memory then it gets wiped, the user might not check the error or even if they do there might have already been some processing happening and we have no way of getting to the data

martinzink: yeah thats a valid point, prior to this PR we dont even notified the user via the logs, but we might want to be more strict thats true

martinzink: im usually on the side to fail fast and fail hard :D so we might just remove this fail_safe parameter and crash in that case

martinzink: removed the fail_safe option, [remove fail_safe option](https://github.com/apache/nifi-minifi-cpp/pull/2083/commits/68da2bfc0e38b37f2f13a886529ef4c9ff7ba2d9)

adamdebreceni (approved):

libminifi/include/core/RepositoryFactory.h
```diff
@@ -32,8 +32,11 @@ namespace org::apache::nifi::minifi::core {
  * @param configuration_class_name configuration class name
  * @param fail_safe determines whether or not to make the default class if configuration_class_name is invalid
```
fgerlits: the documentation of the `fail_safe` parameter should be removed (also below, before the `createRepository` function)

libminifi/src/core/RepositoryFactory.cpp
```diff
@@ -28,33 +28,31 @@ using namespace std::literals::chrono_literals;
...
-    } else {
-      throw std::runtime_error("Support for the provided configuration class could not be found");
-    }
-  } catch (const std::runtime_error&) {
-    if (fail_safe) {
-      return std::make_unique<core::repository::VolatileContentRepository>("fail_safe");
-    }
+
+  auto return_obj = core::ClassLoader::getDefaultClassLoader().instantiate<core::ContentRepository>(class_name_lc,
+                                                                                                  class_name_lc);
```
fgerlits: this indentation looks bad; it may be better to keep this on one line

libminifi/src/core/RepositoryFactory.cpp
```diff
@@ -28,33 +28,31 @@ using namespace std::literals::chrono_literals;
...
 
-  throw std::runtime_error("Support for the provided configuration class could not be found");
+  if (class_name_lc == "volatilecontentrepository") {
+    return std::make_unique<core::repository::VolatileContentRepository>(repo_name);
+  }
+  if (class_name_lc == "filesystemrepository") {
+    return std::make_unique<core::repository::FileSystemRepository>(repo_name);
+  }
+
+  logger->log_critical("Could not create the configured content repository ({})", configuration_class_name);
```
fgerlits: If we allow `logger` to be null, then please check it before dereferencing. It may be better to force it to be non-null?

martinzink: 👍 https://github.com/apache/nifi-minifi-cpp/pull/2083/changes/3812bc5013cb4d3b8c8597e73677870c39e9a210

libminifi/include/core/RepositoryFactory.h
martinzink: good idea, and since the only nontrivial argument was fail_safe I doubt this documentation style comment add anything useful so I've removed it entirely https://github.com/apache/nifi-minifi-cpp/pull/2083/changes/3812bc5013cb4d3b8c8597e73677870c39e9a210

libminifi/src/core/RepositoryFactory.cpp
martinzink: 👍 https://github.com/apache/nifi-minifi-cpp/pull/2083/changes/3812bc5013cb4d3b8c8597e73677870c39e9a210

fgerlits (approved):

szaszm (approved):

lordgamez (approved):

PR #2097, opened 2026-02-06
MINIFICPP-2712 Generate modular docs

martinzink:
With this PR, ./minifi --docs <output_dir> will generate additional md files one per extension <output_dir>/modules

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/minifi_main/AgentDocs.cpp
+++ b/minifi_main/AgentDocs.cpp
@@ -34,7 +34,9 @@
 #include "range/v3/range/conversion.hpp"
 #include "range/v3/view/join.hpp"
 #include "range/v3/view/transform.hpp"
+#include "utils/RegexUtils.h"
 #include "utils/StringUtils.h"
+#include "utils/file/FileUtils.h"
 
 namespace {
 
@@ -42,33 +44,24 @@ namespace minifi = org::apache::nifi::minifi;
 
 std::string formatName(std::string_view name_view, bool is_required) {
   std::string name{name_view};
-  if (is_required) {
-    return "**" + name + "**";
-  } else {
-    return name;
-  }
+  if (is_required) { return "**" + name + "**"; }
+  return name;
 }
 
 std::string formatAllowedValues(const minifi::core::Property& property) {
-  if (property.getValidator().getEquivalentNifiStandardValidatorName() == minifi::core::StandardPropertyValidators::BOOLEAN_VALIDATOR.getEquivalentNifiStandardValidatorName()) {
+  if (property.getValidator().getEquivalentNifiStandardValidatorName() ==
+      minifi::core::StandardPropertyValidators::BOOLEAN_VALIDATOR.getEquivalentNifiStandardValidatorName()) {
     return "true<br/>false";
-  } else {
-    const auto allowed_values = property.getAllowedValues();
-    return allowed_values
-        | ranges::views::join(std::string_view{"<br/>"})
-        | ranges::to<std::string>();
   }
+  const auto allowed_values = property.getAllowedValues();
+  return allowed_values | ranges::views::join(std::string_view{"<br/>"}) | ranges::to<std::string>();
 }
 
 std::string formatDescription(std::string_view description_view, bool is_sensitive = false, bool supports_expression_language = false) {
   std::string description{description_view};
   minifi::utils::string::replaceAll(description, "\n", "<br/>");
-  if (is_sensitive) {
-    description += "<br/>**Sensitive Property: true**";
-  }
-  if (supports_expression_language) {
-    description += "<br/>**Supports Expression Language: true**";
-  }
+  if (is_sensitive) { description += "<br/>**Sensitive Property: true**"; }
+  if (supports_expression_language) { description += "<br/>**Supports Expression Language: true**"; }
   return description;
 }
 
@@ -103,7 +96,7 @@ void writeHeader(std::ostream& docs, const std::vector<std::pair<std::string, mi
   docs << APACHE_LICENSE;
 
   docs << "\n\n## Table of Contents\n\n";
-  for (const auto& [name, documentation] : class_descriptions) {
+  for (const auto& [name, documentation]: class_descriptions) {
     docs << "- [" << name << "](#" << name << ")\n";
   }
 }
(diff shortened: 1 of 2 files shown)


minifi_main/AgentDocs.cpp
```diff
@@ -177,58 +162,151 @@ std::string extractClassName(const std::string& full_class_name) {
...
+    os << "\n\n## Table of Contents\n\n";
+    writeComponentParts(os, components.processors, "Processors");
+    writeComponentParts(os, components.controller_services, "Controller Services");
+    writeComponentParts(os, components.parameter_providers, "Parameter Providers");
   }
+
+  static void writeModule(const std::filesystem::path& docs_dir, const std::string_view module_name, const minifi::Components& components) {
+    const auto dir_creation_result = minifi::utils::file::create_dir(docs_dir / "modules");
+    gsl_Assert(dir_creation_result == 0);
+    std::ofstream os(docs_dir / "modules" / (std::string(module_name) + ".md"));
```
lordgamez, line 288: Could we check beforehand if the component contains any processors, controller services, or parameter providers? Currently we generate an empty markdown file with only an apache license and a Table of Contents header if nothing is available in the extension like prometheus or python script extension. I think those could be omitted.

martinzink: good idea, https://github.com/apache/nifi-minifi-cpp/pull/2097/changes/3505ccf871a2020992147c1bff9ddb678aafc562#diff-39d01781be675a93078dddc3459d8e08702407dd0361efe4d645f00355d15953R260

lordgamez (approved):

minifi_main/AgentDocs.cpp
```diff
@@ -103,7 +95,7 @@ void writeHeader(std::ostream& docs, const std::vector<std::pair<std::string, mi
   docs << APACHE_LICENSE;
 
   docs << "\n\n## Table of Contents\n\n";
-  for (const auto& [name, documentation] : class_descriptions) {
+  for (const auto& [name, documentation]: class_descriptions) {
```
fgerlits: Why did you remove the spaces before the colons (here and at several other places in this file)? The Google style guide recommends a space before and after the colon: https://google.github.io/styleguide/cppguide.html#Horizontal_Whitespace, and that's what we have in at least 90% of minifi code, as well. Please put them back.

minifi_main/AgentDocs.cpp
```diff
@@ -177,58 +162,161 @@ std::string extractClassName(const std::string& full_class_name) {
...
+  writeDescription(os, documentation);
+  writeProperties(os, documentation);
+}
 
-  for (const auto& [bundle_id, components]: ClassDescriptionRegistry::getClassDescriptions()) {
-    for (const auto &controller_service_description : components.controller_services) {
+class MonolithDocumentation {
+ public:
+  explicit MonolithDocumentation() {
+    for (const auto& [bundle_id, component]: minifi::ClassDescriptionRegistry::getClassDescriptions()) {
```
fgerlits: The second variable has type `minifi::Components`. Why did you rename it from `components` to `component`? Please revert this.

minifi_main/AgentDocs.cpp
```diff
@@ -177,58 +162,161 @@ std::string extractClassName(const std::string& full_class_name) {
...
+
+  std::vector<std::pair<std::string, minifi::ClassDescription>> controller_services;
+  std::vector<std::pair<std::string, minifi::ClassDescription>> processors;
+  std::vector<std::pair<std::string, minifi::ClassDescription>> parameter_providers;
+};
+
+class ModularDocumentation {
+ public:
+  static void write(const std::filesystem::path& docs_dir) {
+    for (auto& [bundle_id, component]: minifi::ClassDescriptionRegistry::getMutableClassDescriptions()) {
```
fgerlits: does this have to be mutable?

fgerlits: I would name this `components`, as well

minifi_main/AgentDocs.cpp
```diff
@@ -177,58 +162,161 @@ std::string extractClassName(const std::string& full_class_name) {
...
+      sortComponents(component);
+      writeModule(docs_dir, bundle_id.name, component);
+    }
+  }
+
+ private:
+  static void sortComponents(minifi::Components& components) {
+    std::ranges::sort(components.processors, {}, &minifi::ClassDescription::short_name_);
+    std::ranges::sort(components.controller_services, {}, &minifi::ClassDescription::short_name_);
+    std::ranges::sort(components.parameter_providers, {}, &minifi::ClassDescription::short_name_);
```
fgerlits: This would put `PutSQL` before `PutSmb`, for example. I don't know if we currently have two such names within one module, but we might in the future. I would prefer a case-insensitive sort, unless you have a good reason to prefer this case-sensitive one.

minifi_main/AgentDocs.cpp
martinzink: I've just auto formated the file, it seems like our clang-format didnt agree with our stated style guide
I've modified that aswell [fix clang-format](https://github.com/apache/nifi-minifi-cpp/pull/2097/commits/f6217406d972a5f576bf37ab146397ce791bfed8)

minifi_main/AgentDocs.cpp
martinzink: 👍 https://github.com/apache/nifi-minifi-cpp/pull/2097/changes/460772fb41d29785bf65de9403478823650a5743#diff-39d01781be675a93078dddc3459d8e08702407dd0361efe4d645f00355d15953R187-R188

minifi_main/AgentDocs.cpp
martinzink: if we want to sort them yeah, I agree it might not be nicest but either this or we copy them into another struct or copy them in the loop?

minifi_main/AgentDocs.cpp
martinzink: https://github.com/apache/nifi-minifi-cpp/pull/2097/changes/460772fb41d29785bf65de9403478823650a5743#diff-39d01781be675a93078dddc3459d8e08702407dd0361efe4d645f00355d15953R262-R265

minifi_main/AgentDocs.cpp
martinzink: https://github.com/apache/nifi-minifi-cpp/pull/2097/changes/460772fb41d29785bf65de9403478823650a5743#diff-39d01781be675a93078dddc3459d8e08702407dd0361efe4d645f00355d15953R253-R256

fgerlits: I would prefer a copy, eg.
```
    for (auto [bundle_id, components] : minifi::ClassDescriptionRegistry::getClassDescriptions()) {
```
because the design is that the mutable version is only used during static initialization, and the structure is const afterwards. I know this is a one-shot run of minifi, so it doesn't really matter -- if you want to keep it as it is, I don't mind too much.

martinzink: [copy to avoid getMutableClassDescriptions](https://github.com/apache/nifi-minifi-cpp/pull/2097/commits/f60eb4b533029e3ec4643dcb93ca583a59a5ebc7)

fgerlits (approved):

PR #2133, opened 2026-03-11
MINIFICPP-2718 Windows based docker tests

martinzink:
This PR allows windows based docker containers to run the behave test suite with some limitations:

- I only enabled a handful of features because some of the tests are requiring thirdparty containers (linux only and windows doesnt allow mixed (windows and linux) containers)
- Even the tests that dont require thirdparty containers might rely on extra linux commands that needs to be replicated in powershell
- Its unfeasable to enable these in github CI, because this requires nested virtualization (which is quite rare for VMs).

So this is only the MVP for this feature, but it is still a huge help for future developments especially for using it to develop external extensions using the upcoming C api. It cuts down the local verification time.

I've also replaced the previous m2crypto based ssl_utils with the industry standard cryptography, this should help with the setup on macs and windows systems (m2crypto was quite difficult to install on these systems)

---

### For all changes:
     in the commit message?

### For code changes:

### Note:


behave_framework/pyproject.toml
```diff
@@ -1,18 +1,21 @@
...
 ]
 
 [tool.setuptools]
 package-dir = {"" = "src"}
-packages = ["minifi_test_framework"]
+
+[tool.setuptools.packages.find]
+where = ["src"]
+include = ["minifi_test_framework*"]
+
```
martinzink: this is required for wheel building

extensions/couchbase/tests/features/containers/couchbase_server_container.py
```diff
@@ -66,7 +65,7 @@ def run_post_startup_commands(self):
 
         return True
 
-    @retry_check(max_tries=12, retry_interval=5)
+    @retry_check(12, 5)
```
szaszm: what's the reason for this change?

behave_framework/src/minifi_test_framework/containers/container_windows.py
```diff
@@ -0,0 +1,390 @@
...
+        if not self.container:
+            return False
+
+        command = (
+            f"powershell -Command \"if ((Test-Path -LiteralPath '{directory_path}' -PathType Container) "
+            f"-and (Get-ChildItem -LiteralPath '{directory_path}' -Force | Select-Object -First 1)) "
+            f"{{ exit 0 }} else {{ exit 1 }}\""
+        )
+
+        exit_code, _ = self.exec_run(command)
```
szaszm: why not `_run_powershell` here?

extensions/couchbase/tests/features/containers/couchbase_server_container.py
martinzink: it was mismatched with the actual naming of the parameters
def retry_check(max_tries: int = 5, retry_interval_seconds: int = 1)

should I instead correct and keep the named arguments?

szaszm: I think I'd keep the named arguments, but I leave it up to you to decide which is better.

martinzink: 👍 [retry_check with named arguments](https://github.com/apache/nifi-minifi-cpp/pull/2133/commits/e3429cc8fde010de4065173597e0c38b77821718)

behave_framework/src/minifi_test_framework/containers/container_windows.py
martinzink: good idea, [review changes](https://github.com/apache/nifi-minifi-cpp/pull/2133/commits/febe6b1a59d67a70b60a3c59192f447b2293a02b)

docker/installed/win.Dockerfile
lordgamez, line 1: The usual convention for naming Dockerfiles is `Dockerfile.<target>` so I would rename this to Dockerfile.win

behave_framework/src/minifi_test_framework/containers/container_linux.py
```diff
@@ -479,3 +484,11 @@ def directory_contains_file_with_minimum_size(self, directory_path: str, expecte
             return True
 
         return False
+
+    def get_memory_usage(self) -> int | None:
+        exit_code, output = self.exec_run(["awk", "/VmRSS/ { printf \"%d\\n\", $2 }", "/proc/1/status"])
+        if exit_code != 0:
+            return None
+        memory_usage_in_bytes = int(output.strip()) * 1024
```
fgerlits, line 492: very minor, but why do we add the `\n` in line 489 only to strip it out in line 492? also, I would include "kB" in the regex, just in case the format of `/proc/1/status` changes in the future:
```suggestion
        exit_code, output = self.exec_run(["awk", "/VmRSS.*kB/ { printf \"%d\", $2 }", "/proc/1/status"])
        if exit_code != 0:
            return None
        memory_usage_in_bytes = int(output) * 1024
```

behave_framework/src/minifi_test_framework/containers/container_protocol.py
```diff
@@ -0,0 +1,62 @@
...
+        ...
+
+    def clean_up(self):
+        ...
+
+    def exec_run(self, command):
+        ...
+
+    def directory_contains_file_with_content(self, directory_path: str, expected_content: str) -> bool:
+        ...
```
fgerlits, line 40: It's a bit strange that we say these stubs return `bool`, `str` or `int`, but in fact we return `None`. Maybe raise an exception instead?

behave_framework/src/minifi_test_framework/containers/container_windows.py
```diff
@@ -0,0 +1,392 @@
...
+    def __init__(self, image_name: str, container_name: str, network: Network, command: str | None = None, entrypoint: str | None = None):
+        super().__init__()
+        self.image_name: str = image_name
+        self.container_name: str = container_name
+        self.network: Network = network
+
+        self.client: docker.DockerClient = docker.from_env()
+        self.container: Optional[Container] = None
+        self.files: List[File] = []
+        self.dirs: List[Directory] = []
```
fgerlits, line 49: minor, but `List` is deprecated, we should use `list` instead: https://docs.python.org/3/library/typing.html#aliases-to-built-in-types

behave_framework/src/minifi_test_framework/containers/container_windows.py
```diff
@@ -0,0 +1,392 @@
...
+
+    def _normalize_path(self, path: str) -> str:
+        clean_path = path.strip().replace("/", "\\")
+        if clean_path.startswith("\\"):
+            clean_path = clean_path[1:]
+
+        # If it doesn't already have a drive letter, assume C:
+        if ":" not in clean_path:
+            return f"C:\\{clean_path}"
+        return clean_path
```
fgerlits, line 66: would `os.path.normpath()` work instead of this function?

behave_framework/src/minifi_test_framework/containers/container_windows.py
```diff
@@ -0,0 +1,392 @@
...
+
+        exit_code, _ = self._run_powershell(ps_script)
+        return exit_code == 0
+
+    def directory_contains_file_with_content(self, directory_path: str, expected_content: str) -> bool:
+        if not self.container:
+            return False
+
+        win_path = self._normalize_path(directory_path)
+        escaped_content = expected_content.replace("'", "''")
```
fgerlits: in PS, things are usually escaped with a backtick:
```suggestion
        escaped_content = expected_content.replace("'", "`'")
```

but if this works, too, we can leave it as it is

extensions/standard-processors/tests/features/core_functionality.feature
```diff
@@ -33,8 +33,8 @@ Feature: Core flow functionalities
 
     When all instances start up
 
-    Then at least one file with the content "first_custom_text" is placed in the "/tmp/output" directory in less than 20 seconds
-    And at least one file with the content "second_custom_text" is placed in the "/tmp/output" directory in less than 20 seconds
+    Then at least one file with the content "first_custom_text" is placed in the "/tmp/output" directory in less than 200 seconds
```
martinzink, line 36: why?

behave_framework/src/minifi_test_framework/containers/container_windows.py
```diff
@@ -0,0 +1,392 @@
...
+        win_path = self._normalize_path(directory_path)
+        ps_script = (
+            f"if ((Test-Path -LiteralPath '{win_path}' -PathType Container) "
+            f"-and (Get-ChildItem -LiteralPath '{win_path}' -Force | Select-Object -First 1)) "
+            f"{{ exit 0 }} else {{ exit 1 }}"
+        )
+
+        exit_code, _ = self._run_powershell(ps_script)
+
+        return exit_code == 0
```
fgerlits, line 371: we could log the error (if any) here, too, as in the other functions

behave_framework/src/minifi_test_framework/containers/container_windows.py
```diff
@@ -0,0 +1,392 @@
...
+
+    def nonempty_dir_exists(self, directory_path: str) -> bool:
+        if not self.container:
+            return False
+
+        win_path = self._normalize_path(directory_path)
+        ps_script = (
+            f"if ((Test-Path -LiteralPath '{win_path}' -PathType Container) "
+            f"-and (Get-ChildItem -LiteralPath '{win_path}' -Force | Select-Object -First 1)) "
+            f"{{ exit 0 }} else {{ exit 1 }}"
```
fgerlits, line 366: When I run this on a test directory, I get True on both empty and non-empty directories. A StackOverflow answer suggests adding `Measure-Object`:
```suggestion
            f"if ((Test-Path -LiteralPath '{win_path}' -PathType Container) "
            f"-and ((Get-ChildItem -LiteralPath '{win_path}' -Force | "
            f"        Select-Object -First 1 | Measure-Object).Count -gt 0)) "
            f"{{ exit 0 }} else {{ exit 1 }}"
```

behave_framework/src/minifi_test_framework/core/ssl_utils.py
```diff
@@ -13,149 +13,79 @@
...
+def make_self_signed_cert(common_name: str) -> tuple[Certificate, RSAPrivateKey]:
+    key = rsa.generate_private_key(public_exponent=65537, key_size=2048)
 
-    return ca_cert, ca_key
+    subject = issuer = x509.Name([x509.NameAttribute(NameOID.COMMON_NAME, common_name), ])
 
+    cert = x509.CertificateBuilder().subject_name(subject).issuer_name(issuer).public_key(key.public_key()).serial_number(
+        x509.random_serial_number()).not_valid_before(datetime.datetime.now(datetime.timezone.utc)).not_valid_after(
+        datetime.datetime.now(datetime.timezone.utc) + datetime.timedelta(days=3650)).add_extension(
+        x509.SubjectKeyIdentifier.from_public_key(key.public_key()), critical=False, ).add_extension(x509.BasicConstraints(ca=True, path_length=None),
```
fgerlits, line 47: nitpicking, but I think breaking into lines like this would be easier to read:
```suggestion
    cert = (x509.CertificateBuilder().subject_name(subject).issuer_name(issuer).public_key(key.public_key())
        .serial_number(x509.random_serial_number())
        .not_valid_before(datetime.datetime.now(datetime.timezone.utc))
        .not_valid_after(datetime.datetime.now(datetime.timezone.utc) + datetime.timedelta(days=3650))
        .add_extension(x509.SubjectKeyIdentifier.from_public_key(key.public_key()), critical=False)
        .add_extension(x509.BasicConstraints(ca=True, path_length=None), critical=True)
        .sign(key, hashes.SHA256()))
```

extensions/standard-processors/tests/features/core_functionality.feature
```diff
@@ -33,8 +33,8 @@ Feature: Core flow functionalities
...
+    And at least one file with the content "second_custom_text" is placed in the "/tmp/output" directory in less than 200 seconds
```
fgerlits, line 37: Is this much slower on Windows, or was the timeout only changed for testing and it can be reverted?

behave_framework/src/minifi_test_framework/containers/container_windows.py
```diff
@@ -0,0 +1,392 @@
...
+        self.client: docker.DockerClient = docker.from_env()
+        self.container: Optional[Container] = None
+        self.files: List[File] = []
+        self.dirs: List[Directory] = []
+        self.host_files: List[HostFile] = []
+        self.volumes: Dict = {}
+        self.command: str | None = command
+        self.entrypoint: str | None = entrypoint
+        self._temp_dir: Optional[tempfile.TemporaryDirectory] = None
+        self.ports: Optional[Dict[str, int]] = None
```
lordgamez, line 55: We should be consistent, either use X | None everywhere or Optional, I would go with X | None as it is the newer syntax and needs no import.

behave_framework/src/minifi_test_framework/containers/container_windows.py
```diff
@@ -0,0 +1,392 @@
...
+                environment=self.environment,
+                volumes=self.volumes,
+                network=self.network.name,
+                command=self.command,
+                entrypoint=self.entrypoint,
+                detach=True,
+                tty=False
+            )
+
+            self.container.start()
```
lordgamez, line 116: Could we replace container.create + container.start with container.run here?

behave_framework/src/minifi_test_framework/containers/container_windows.py
```diff
@@ -0,0 +1,392 @@
...
+
+        tar_stream.seek(0)
+
+        self.container.put_archive(path=dir_name, data=tar_stream)
+
+    def clean_up(self):
+        if self._temp_dir:
+            try:
+                self._temp_dir.cleanup()
+                self._temp_dir = None
```
lordgamez, line 155: Shouldn't we put the `self._temp_dir = None` in the finally? Even if the cleanup fails, we should create a new directory for our next container after cleanup.

behave_framework/src/minifi_test_framework/containers/minifi_protocol.py
```diff
@@ -0,0 +1,58 @@
...
+        ...
+
+    def set_log_property(self, key: str, value: str):
+        ...
+
+    def set_deploy_timeout_seconds(self, timeout_seconds: int):
+        ...
+
+
+def set_controller_socket_properties(minifi: MinifiProtocol):
```
lordgamez, line 33: This is already part of MinifiController as a method, isn't that enough?

behave_framework/src/minifi_test_framework/containers/container_protocol.py
martinzink: afaik when using protocols its standard to use ... even if the protocol defines return types 
https://peps.python.org/pep-0544/

behave_framework/src/minifi_test_framework/containers/container_windows.py
martinzink: we also prepend c: if its not there (which is required so we dont have to rewrite every test with /tmp/output etc), so no, maybe we could do normalize instead of the first part if you would like?

behave_framework/src/minifi_test_framework/core/ssl_utils.py
martinzink: flake8 didnt like it,
i've reformatted the whole file with ruff instead how about this?
[ruff format behave_framework/src/minifi_test_framework/core/ssl_utils.py](https://github.com/apache/nifi-minifi-cpp/pull/2133/commits/4bcc95b0cd95aa29b0a1bdc8d5f072e78e7e66ca)

extensions/standard-processors/tests/features/core_functionality.feature
martinzink: yeah just debugging reverted [reverted 200 second debug timeout](https://github.com/apache/nifi-minifi-cpp/pull/2133/commits/748f2e4b6af8b836b5b6df5c0b8af3588cd25b17)

docker/installed/Dockerfile.win
martinzink: didnt know about this convention good to know, renamed this and other in this dir
[renamed dockerfiles](https://github.com/apache/nifi-minifi-cpp/pull/2133/commits/95f6047a0c3c3ba3080c774824be4f0b7b080aed)

behave_framework/src/minifi_test_framework/containers/container_windows.py
martinzink: good idea, https://github.com/apache/nifi-minifi-cpp/pull/2133/changes/20e8b2e39e6ab0334fed85d8eb0e17a5cb59501d#diff-badda25eb0095d027910c54ff8750a8ae77d4aeb283a79641f78c1d5957f279dR155-R156

behave_framework/src/minifi_test_framework/containers/minifi_protocol.py
martinzink: 👍 https://github.com/apache/nifi-minifi-cpp/pull/2133/changes/20e8b2e39e6ab0334fed85d8eb0e17a5cb59501d#diff-2432ecfa82b0448e4b0091c323c412d7424b996a61ed2c67b2f193b67f02d7a6L33-L39

behave_framework/src/minifi_test_framework/containers/container_windows.py
martinzink: good idea, https://github.com/apache/nifi-minifi-cpp/pull/2133/changes/20e8b2e39e6ab0334fed85d8eb0e17a5cb59501d#diff-badda25eb0095d027910c54ff8750a8ae77d4aeb283a79641f78c1d5957f279dR103

behave_framework/src/minifi_test_framework/containers/container_windows.py
martinzink: 👍 https://github.com/apache/nifi-minifi-cpp/pull/2133/changes/20e8b2e39e6ab0334fed85d8eb0e17a5cb59501d

behave_framework/src/minifi_test_framework/containers/container_windows.py
martinzink: https://github.com/apache/nifi-minifi-cpp/pull/2133/changes/20e8b2e39e6ab0334fed85d8eb0e17a5cb59501d

behave_framework/src/minifi_test_framework/containers/container_windows.py
martinzink: [logging.error(f"Error running command for nonempty_dir_exists: {outpu…](https://github.com/apache/nifi-minifi-cpp/pull/2133/commits/07064ecc96ad8744f301266a0ef5fae8aa7352a9)

behave_framework/src/minifi_test_framework/containers/container_windows.py
fgerlits: How about `os.path.realpath`? That function prepends `C:`, too.

behave_framework/src/minifi_test_framework/core/ssl_utils.py
fgerlits: It looks good, except it seems to limit line length to a low value (100 chars?) -- can we increase that to e.g. 200 chars?

EDIT: I have done this in c4cd3921db87ff33de6f4948d58c6da5c66bf7fb; please double-check and resolve this thread if it looks OK.

martinzink: yeah looks good, thanks 👍

lordgamez (approved):

behave_framework/src/minifi_test_framework/containers/container_linux.py
```diff
@@ -479,3 +484,11 @@ def directory_contains_file_with_minimum_size(self, directory_path: str, expecte
             return True
 
         return False
+
+    def get_memory_usage(self) -> int | None:
+        exit_code, output = self.exec_run(["awk", "/VmRSS.*kB/ { printf \"%d\", $2 }", "/proc/1/status"])
```
szaszm, line 489: I think this only counts the main container process, not any child processes. Is this intentional?

martinzink: i dont know, this is not new or changed (just the file renamed from minifi-container.py)

behave_framework/src/minifi_test_framework/containers/container_windows.py
martinzink: I tested it, and that prepends the current working directory not `c:`

fgerlits: hm, interesting, it worked when I tried it; I got `foo\bar.txt` -> `C:\current\working\directory\foo\bar.txt` and `\foo\bar.txt` -> `C:\foo\bar.txt`

but maybe it doesn't work in all cases

fgerlits (approved):

behave_framework/src/minifi_test_framework/containers/container_windows.py
martinzink: u are right, then it should work, because its usually /tmp/input so it should work 👍 just checked

docker/installed/Dockerfile.win
```diff
@@ -0,0 +1,42 @@
...
+RUN choco install git -y
+RUN choco install vcredist140 -y
+
+COPY ${MSI_SOURCE} C:\temp\minifi.msi
+
+SHELL ["cmd", "/S", "/C"]
+
+RUN C:\Windows\System32\msiexec.exe /i C:\temp\minifi.msi /qn /norestart /L*V C:\minifi_install.log
+
+CMD ["C:\\Program Files\\ApacheNiFiMiNiFi\\nifi-minifi-cpp\\bin\\minifi.exe"]
```
szaszm:
```suggestion
CMD ["C:\\Program Files\\ApacheNiFiMiNiFi\\nifi-minifi-cpp\\bin\\minifi.exe"]

```

szaszm (approved):

PR #2148, opened 2026-03-30
MINIFICPP-2760 Input/OutputStreamCallback should return std::expected

martinzink:
This PR addresses a valid critisim during the review in https://github.com/apache/nifi-minifi-cpp/pull/2120#discussion_r3000932337

I've switched to an expected result while keeping the i64 on the C API.
We already abused (and unfortuently will continue to abuse but less so ) the u64 and size_t to hide our errors in them.  We used static_cast<u64>(-1) to signal errors, but now we have multiple erros, so this PR switches most of the code (but not all) to instead signal the errors in the unexpected value and one could be certain that if the expected has value it is a valid size.
I've implemented the back and forth conversions from the old i64, and hacked u64 to the new std::expected<u64, MinifiIoStatus>

---

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/aws/processors/PutS3Object.cpp
+++ b/extensions/aws/processors/PutS3Object.cpp
@@ -288,11 +288,11 @@ void PutS3Object::onTrigger(core::ProcessContext& context, core::ProcessSession&
       if (flow_file->getSize() <= multipart_threshold_) {
         logger_->log_info("Uploading S3 Object '{}' in a single upload", put_s3_request_params->object_key);
         result = s3_wrapper_->putObject(*put_s3_request_params, stream, flow_file->getSize());
-        return io::IoResult::fromSizeT(flow_file->getSize());
+        return io::IoResult::from(flow_file->getSize());
       }
       logger_->log_info("S3 Object '{}' passes the multipart threshold, uploading it in multiple parts", put_s3_request_params->object_key);
       result = s3_wrapper_->putObjectMultipart(*put_s3_request_params, stream, flow_file->getSize(), multipart_size_);
-      return io::IoResult::fromSizeT(flow_file->getSize());
+      return io::IoResult::from(flow_file->getSize());
     } catch(const aws::s3::StreamReadException& ex) {
       logger_->log_error("Error occurred while uploading to S3: {}", ex.what());
       return io::IoResult::error();
--- a/extensions/aws/processors/PutS3Object.cpp
+++ b/extensions/aws/processors/PutS3Object.cpp
@@ -283,20 +283,19 @@ void PutS3Object::onTrigger(core::ProcessContext& context, core::ProcessSession&
   }
 
   std::optional<minifi::aws::s3::PutObjectResult> result;
-  session.read(flow_file, [this, &flow_file, &put_s3_request_params, &result](const std::shared_ptr<io::InputStream>& stream) -> int64_t {
+  session.read(flow_file, [this, &flow_file, &put_s3_request_params, &result](const std::shared_ptr<io::InputStream>& stream) -> io::IoResult {
     try {
       if (flow_file->getSize() <= multipart_threshold_) {
         logger_->log_info("Uploading S3 Object '{}' in a single upload", put_s3_request_params->object_key);
         result = s3_wrapper_->putObject(*put_s3_request_params, stream, flow_file->getSize());
-        return gsl::narrow<int64_t>(flow_file->getSize());
-      } else {
-        logger_->log_info("S3 Object '{}' passes the multipart threshold, uploading it in multiple parts", put_s3_request_params->object_key);
-        result = s3_wrapper_->putObjectMultipart(*put_s3_request_params, stream, flow_file->getSize(), multipart_size_);
-        return gsl::narrow<int64_t>(flow_file->getSize());
+        return io::IoResult::fromSizeT(flow_file->getSize());
       }
+      logger_->log_info("S3 Object '{}' passes the multipart threshold, uploading it in multiple parts", put_s3_request_params->object_key);
+      result = s3_wrapper_->putObjectMultipart(*put_s3_request_params, stream, flow_file->getSize(), multipart_size_);
+      return io::IoResult::fromSizeT(flow_file->getSize());
     } catch(const aws::s3::StreamReadException& ex) {
       logger_->log_error("Error occurred while uploading to S3: {}", ex.what());
-      return -1;
+      return io::IoResult::error();
     }
   });
   if (!result.has_value()) {
--- a/extensions/bustache/ApplyTemplate.cpp
+++ b/extensions/bustache/ApplyTemplate.cpp
@@ -43,7 +43,7 @@ void ApplyTemplate::onTrigger(core::ProcessContext& context, core::ProcessSessio
   }
 
   std::string template_file = context.getProperty(Template.name, flow_file.get()).value_or("");
-  session.write(flow_file, [&template_file, &flow_file, this](const auto& output_stream) {
+  session.write(flow_file, [&template_file, &flow_file, this](const auto& output_stream) -> io::IoResult {
     logger_->log_info("ApplyTemplate reading template file from {}", template_file);
     // TODO(szaszm): we might want to return to memory-mapped input files when the next todo is done. Until then, the agents stores the whole result in memory anyway, so no point in not doing the same
     // with the template file itself
@@ -61,8 +61,11 @@ void ApplyTemplate::onTrigger(core::ProcessContext& context, core::ProcessSessio
 
     // TODO(calebj) write ostream reciever for format() to prevent excessive copying
     std::string ostring = bustache::to_string(format(data));
-    output_stream->write(gsl::make_span(ostring).as_span<const std::byte>());
-    return gsl::narrow<int64_t>(ostring.length());
+    const auto write_res = output_stream->write(gsl::make_span(ostring).as_span<const std::byte>());
+    if (write_res < 0) {
+      return io::IoResult::error();
+    }
+    return io::IoResult::fromSizeT(ostring.length());
   });
   session.transfer(flow_file, Success);
 }
--- a/extensions/bustache/ApplyTemplate.cpp
+++ b/extensions/bustache/ApplyTemplate.cpp
@@ -62,7 +62,7 @@ void ApplyTemplate::onTrigger(core::ProcessContext& context, core::ProcessSessio
     // TODO(calebj) write ostream reciever for format() to prevent excessive copying
     std::string ostring = bustache::to_string(format(data));
     output_stream->write(gsl::make_span(ostring).as_span<const std::byte>());
-    return io::IoResult::fromSizeT(ostring.length());
+    return io::IoResult::from(ostring.length());
   });
   session.transfer(flow_file, Success);
 }
--- a/extensions/bustache/ApplyTemplate.cpp
+++ b/extensions/bustache/ApplyTemplate.cpp
@@ -61,7 +61,7 @@ void ApplyTemplate::onTrigger(core::ProcessContext& context, core::ProcessSessio
 
     // TODO(calebj) write ostream reciever for format() to prevent excessive copying
     std::string ostring = bustache::to_string(format(data));
-    const auto write_res = output_stream->write(gsl::make_span(ostring).as_span<const std::byte>());
+    output_stream->write(gsl::make_span(ostring).as_span<const std::byte>());
     return io::IoResult::fromSizeT(ostring.length());
   });
   session.transfer(flow_file, Success);
--- a/extensions/bustache/ApplyTemplate.cpp
+++ b/extensions/bustache/ApplyTemplate.cpp
@@ -62,9 +62,6 @@ void ApplyTemplate::onTrigger(core::ProcessContext& context, core::ProcessSessio
     // TODO(calebj) write ostream reciever for format() to prevent excessive copying
     std::string ostring = bustache::to_string(format(data));
     const auto write_res = output_stream->write(gsl::make_span(ostring).as_span<const std::byte>());
-    if (write_res < 0) {
-      return io::IoResult::error();
-    }
     return io::IoResult::fromSizeT(ostring.length());
   });
   session.transfer(flow_file, Success);
--- a/extensions/libarchive/FocusArchiveEntry.cpp
+++ b/extensions/libarchive/FocusArchiveEntry.cpp
@@ -154,7 +154,7 @@ la_ssize_t FocusArchiveEntry::ReadCallback::read_cb(struct archive * a, void *d,
   return gsl::narrow<la_ssize_t>(read);
 }
 
-int64_t FocusArchiveEntry::ReadCallback::operator()(const std::shared_ptr<io::InputStream>& stream) const {
+io::IoResult FocusArchiveEntry::ReadCallback::operator()(const std::shared_ptr<io::InputStream>& stream) const {
   auto input_archive = processors::archive_read_unique_ptr{archive_read_new()};
   struct archive_entry *entry = nullptr;
   int64_t nlen = 0;
@@ -169,7 +169,7 @@ int64_t FocusArchiveEntry::ReadCallback::operator()(const std::shared_ptr<io::In
   // Read each item in the archive
   if (archive_read_open(input_archive.get(), &data, ok_cb, read_cb, ok_cb)) {
     logger_->log_error("FocusArchiveEntry can't open due to archive error: {}", archive_error_string(input_archive.get()));
-    return nlen;
+    return io::IoResult::fromI64(nlen);
   }
 
   while (context_->isRunning()) {
@@ -181,12 +181,7 @@ int64_t FocusArchiveEntry::ReadCallback::operator()(const std::shared_ptr<io::In
 
     if (res < ARCHIVE_OK) {
       logger_->log_error("FocusArchiveEntry can't read header due to archive error: {}", archive_error_string(input_archive.get()));
-      return nlen;
-    }
-
-    if (res < ARCHIVE_WARN) {
-      logger_->log_warn("FocusArchiveEntry got archive warning while reading header: {}", archive_error_string(input_archive.get()));
-      return nlen;
+      return io::IoResult::error();
     }
 
     auto entryName = archive_entry_pathname(entry);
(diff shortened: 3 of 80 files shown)


extensions/aws/processors/PutS3Object.cpp
```diff
@@ -293,20 +293,19 @@ void PutS3Object::onTrigger(core::ProcessContext& context, core::ProcessSession&
...
-        return gsl::narrow<int64_t>(flow_file->getSize());
-      } else {
-        logger_->log_info("S3 Object '{}' passes the multipart threshold, uploading it in multiple parts", put_s3_request_params->object_key);
-        result = s3_wrapper_.putObjectMultipart(*put_s3_request_params, stream, flow_file->getSize(), multipart_size_);
-        return gsl::narrow<int64_t>(flow_file->getSize());
+        return io::IoResult::fromSizeT(flow_file->getSize());
       }
+      logger_->log_info("S3 Object '{}' passes the multipart threshold, uploading it in multiple parts", put_s3_request_params->object_key);
+      result = s3_wrapper_.putObjectMultipart(*put_s3_request_params, stream, flow_file->getSize(), multipart_size_);
+      return io::IoResult::fromSizeT(flow_file->getSize());
```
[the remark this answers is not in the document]
martinzink: we dont support 32 bit builds, and if we were we would have this problem on a lot of different places (we use size_t in a bunch of places)

extensions/mqtt/processors/ConsumeMQTT.cpp
```diff
@@ -163,21 +163,21 @@ std::queue<ConsumeMQTT::SmartMessage> ConsumeMQTT::getReceivedMqttMessages() {
...
+  const size_t len = stream->write(reinterpret_cast<uint8_t*>(message_.contents->payload), gsl::narrow<size_t>(message_.contents->payloadlen));
   if (io::isError(len)) {
     success_status_ = false;
     logger_->log_error("Stream writing error when processing message");
-    return -1;
+    return io::IoResult::error();
   }
 
-  return gsl::narrow<int64_t>(len);
+  return io::IoResult::fromSizeT(gsl::narrow<int64_t>(len));
```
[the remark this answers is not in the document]
martinzink: fixed https://github.com/apache/nifi-minifi-cpp/pull/2148/changes/d0cc52597cee192b10d9f46d08340029f6a8c6d5#diff-bcdc140bec3984114489502b01dfea7dd83390b110643ce0ebd3aefbb81d6badR180

extensions/bustache/ApplyTemplate.cpp
```diff
@@ -61,8 +61,11 @@ void ApplyTemplate::onTrigger(core::ProcessContext& context, core::ProcessSessio
 
     // TODO(calebj) write ostream reciever for format() to prevent excessive copying
     std::string ostring = bustache::to_string(format(data));
-    output_stream->write(gsl::make_span(ostring).as_span<const std::byte>());
-    return gsl::narrow<int64_t>(ostring.length());
+    const auto write_res = output_stream->write(gsl::make_span(ostring).as_span<const std::byte>());
+    if (write_res < 0) {
+      return io::IoResult::error();
+    }
+    return io::IoResult::fromSizeT(ostring.length());
```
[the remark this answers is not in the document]
martinzink: good point fixed, https://github.com/apache/nifi-minifi-cpp/pull/2148/changes/d0cc52597cee192b10d9f46d08340029f6a8c6d5#diff-920237a9f4b39ba595639dc10b0b316826c369990c09d6bc41a8fa24fc6d2ce5R64-R66

minifi-api/common/include/minifi-cpp/io/StreamCallback.h
```diff
@@ -16,24 +16,101 @@
...
+  nonstd::expected<ReadWrite, MinifiIoStatus> result_;
 };
 
-// FlowFile IO Callback functions for input and output
-// throw exception for error
-using InputStreamCallback = std::function<int64_t(const std::shared_ptr<InputStream>& input_stream)>;
-using OutputStreamCallback = std::function<int64_t(const std::shared_ptr<OutputStream>& output_stream)>;
-using InputOutputStreamCallback = std::function<std::optional<ReadWriteResult>(const std::shared_ptr<InputStream>& input_stream, const std::shared_ptr<OutputStream>& output_stream)>;
+using OutputStreamCallback = std::function<IoResult(const std::shared_ptr<OutputStream>& output_stream)>;
+using InputStreamCallback = std::function<IoResult(const std::shared_ptr<InputStream>& output_stream)>;
```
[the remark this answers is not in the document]
martinzink: https://github.com/apache/nifi-minifi-cpp/pull/2148/changes/d0cc52597cee192b10d9f46d08340029f6a8c6d5#diff-e3c3d53f326186152d21dc0bc5bb5e0339747cf2584945707e5d0141180c6595R112

extensions/libarchive/FocusArchiveEntry.cpp
```diff
@@ -169,7 +169,7 @@ int64_t FocusArchiveEntry::ReadCallback::operator()(const std::shared_ptr<io::In
   // Read each item in the archive
   if (archive_read_open(input_archive.get(), &data, ok_cb, read_cb, ok_cb)) {
     logger_->log_error("FocusArchiveEntry can't open due to archive error: {}", archive_error_string(input_archive.get()));
-    return nlen;
+    return io::IoResult::fromI64(nlen);
```
fgerlits:
```suggestion
    return io::IoResult::zero();
```
since there is no way for `nlen` to be anything other than zero here

extensions/libarchive/FocusArchiveEntry.cpp
```diff
@@ -181,12 +181,7 @@ int64_t FocusArchiveEntry::ReadCallback::operator()(const std::shared_ptr<io::In
 
     if (res < ARCHIVE_OK) {
       logger_->log_error("FocusArchiveEntry can't read header due to archive error: {}", archive_error_string(input_archive.get()));
-      return nlen;
-    }
-
-    if (res < ARCHIVE_WARN) {
-      logger_->log_warn("FocusArchiveEntry got archive warning while reading header: {}", archive_error_string(input_archive.get()));
-      return nlen;
+      return io::IoResult::error();
```
fgerlits: I would keep this as `zero()` so that we don't change the logic, unless you understand it fully and you're sure that it was wrong.

Also, I would fix the warning vs error distinction instead of removing it:
* check for errors first, log error, return zero
* then check for retry/warning, log retry/warning, return zero.

extensions/rocksdb-repos/tests/SwapTests.cpp
```diff
@@ -56,14 +56,14 @@ class OutputProcessor : public core::ProcessorImpl {
...
-      auto ret = output->write(as_bytes(std::span(id)));
-      if (minifi::io::isError(ret)) {
-        return -1;
+    session.write(ff, [&] (const std::shared_ptr<minifi::io::OutputStream>& output) -> io::IoResult {
+      const size_t ret = output->write(as_bytes(std::span(id)));
+      if (io::isError(ret)) {
+        return io::IoResult::error();
       }
-      return gsl::narrow<int64_t>(ret);
+      return io::IoResult::fromSizeT(ret);
```
fgerlits: nitpicking, but fromSizeT checks for error internally, so lines 63-65 can be removed

libminifi/src/core/ProcessSession.cpp
```diff
@@ -258,14 +258,14 @@ void ProcessSessionImpl::write(core::FlowFile &flow, const io::OutputStreamCallb
...
+      if (callback_result.is_cancelled()) {
+        stream->close();
+        content_session_->remove(claim);
+        claim.reset();
+        return;
+      }
 
-    if (callback_result < 0) {
       throw Exception(FILE_OPERATION_EXCEPTION, "Failed to process flowfile content");
     }
```
fgerlits: I liked the previous flow better, without nesting the `if`s.

libminifi/src/sitetosite/CompressionOutputStream.cpp
```diff
@@ -103,7 +103,7 @@ size_t CompressionOutputStream::compressAndWrite() {
   }
 
   // Write the compressed data
-  ret = internal::pipe(buffer_stream, internal_stream_);
+  ret = minifi::io::IoResult(internal::pipe(buffer_stream, internal_stream_)).toI64();  // TODO(mzink) feels wrong
```
fgerlits: Yes, it does... Maybe we could add a `toSizeT` method? Or keep the return value of `pipe` as `size_t` and convert it later?

 (Also, the copy seems unnecessary.)

minifi-api/common/include/minifi-cpp/io/StreamCallback.h
```diff
@@ -16,24 +16,101 @@
  */
 #pragma once
 
+#include <cinttypes>
 #include <functional>
 #include <memory>
 #include <optional>
 
+#include "../../minifi-api/include/minifi-c/minifi-c.h"
```
fgerlits, line 23: elsewhere we have
```suggestion
#include "minifi-c/minifi-c.h"
```
would that work here?

minifi-api/common/include/minifi-cpp/io/StreamCallback.h
```diff
@@ -16,24 +16,101 @@
  */
 #pragma once
 
+#include <cinttypes>
 #include <functional>
 #include <memory>
 #include <optional>
```
fgerlits: `#include <optional>` can be removed

fgerlits (commented):
Thank you for doing this! I think it improves readability significantly.

minifi-api/common/include/minifi-cpp/io/StreamCallback.h
martinzink: unfortunetly it wont work, this is included from files where the search path only includes minifi-api/include/minifi-c so you need `#include "minifi-c.h"` (e.g. extension-framework/cpp-extension-lib/src/core/ProcessSession.cpp) and also from files where the search path only includes minifi-api/include so you would need `#include "minifi-c/minifi-c.h"` (e.g. core-framework/src/core/ContentSession.cpp)

We should probably fix this but didnt want to burden this PR (and more so myself) with the cmake magic that controls these

libminifi/src/sitetosite/CompressionOutputStream.cpp
martinzink: yeah, I forgot this comment and that i still had todos with it,

maybe this as a comprimise?

https://github.com/apache/nifi-minifi-cpp/pull/2148/changes/21f5ece018ff1324ddb15bf6c4bed0c126a127ab#diff-df9fae27f6ebe2efade9fb444f123205d3efba0dd458c51f8cd9a37ffa3fe276R106-R107

libminifi/src/core/ProcessSession.cpp
martinzink: agree, firstly I went with raw std::expected and not a custom class that justified this change but than later I moved to custom class and didnt think this through a second time

https://github.com/apache/nifi-minifi-cpp/commit/21f5ece018ff1324ddb15bf6c4bed0c126a127ab#diff-d55088efe0f1ad2a600dba414bfee749c5fb634ab44418146623f892bd6756a4R262-R269

extensions/rocksdb-repos/tests/SwapTests.cpp
martinzink: 👍 https://github.com/apache/nifi-minifi-cpp/commit/21f5ece018ff1324ddb15bf6c4bed0c126a127ab#diff-204b48ece2bb6eb91b320c67870d012681c2d1d60002c4802db64731da5c84a8L63-L65

extensions/libarchive/FocusArchiveEntry.cpp
martinzink: no, the zero change was unintentional, good catch 👍, 

the warn part was removed because it was unreachable code (since ARCHIVE_OK is 0 and ARCHIVE_WARNING is -20 iirc), we should probably redo this whole part and ditch the less than operators and check every known error code because the < WARN is also wrong because 
```
#define	ARCHIVE_EOF	  1	/* Found end of archive. */
#define	ARCHIVE_OK	  0	/* Operation was successful. */
#define	ARCHIVE_RETRY	(-10)	/* Retry might succeed. */
#define	ARCHIVE_WARN	(-20)	/* Partial success. */
/* For example, if write_header "fails", then you can't push data. */
#define	ARCHIVE_FAILED	(-25)	/* Current operation cannot complete. */
/* But if write_header is "fatal," then this archive is dead and useless. */
#define	ARCHIVE_FATAL	(-30)	/* No more operations are possible. */
```

but again this seems out of the scope fo this PR, Ive created a JIRA ticket so we could understand and handle the different returns correctly

[added TODO to remind for MINIFICPP-2761](https://github.com/apache/nifi-minifi-cpp/pull/2148/commits/0417f390bbe1dae7a71a4eb08d2582c2dfc2e14c)[added TODO to remind for MINIFICPP-2761](https://github.com/apache/nifi-minifi-cpp/pull/2148/commits/0417f390bbe1dae7a71a4eb08d2582c2dfc2e14c)

https://issues.apache.org/jira/browse/MINIFICPP-2761

extensions/libarchive/FocusArchiveEntry.cpp
martinzink: 👍 https://github.com/apache/nifi-minifi-cpp/commit/21f5ece018ff1324ddb15bf6c4bed0c126a127ab#diff-26d52bd8c669d38dc6f969e095bbbb61ec6c9da9628715eb7312cffe64cfab48R172

minifi-api/common/include/minifi-cpp/io/StreamCallback.h
martinzink: https://github.com/apache/nifi-minifi-cpp/pull/2148/changes/21f5ece018ff1324ddb15bf6c4bed0c126a127ab#diff-e3c3d53f326186152d21dc0bc5bb5e0339747cf2584945707e5d0141180c6595L22

libminifi/src/sitetosite/CompressionOutputStream.cpp
fgerlits: Yes, but the return value at the end should be `gsl::narrow<size_t>(pipe_res.toI64())` or something similar, not `ret`.

In fact, it would be nice to get rid of the reusing of `ret` in this function, and use separate variables each time.

martinzink: [review changes](https://github.com/apache/nifi-minifi-cpp/pull/2148/commits/35e35c3c938e866668a9ab1545218fa3d40f6ea8)

yeah that makes sense, its still strange that we only return the last operations size and not totalling it, but I probably dont want to poke this anymore in this PR

fgerlits (approved):

extensions/libarchive/FocusArchiveEntry.cpp
```diff
@@ -179,14 +179,9 @@ int64_t FocusArchiveEntry::ReadCallback::operator()(const std::shared_ptr<io::In
...
-    if (res < ARCHIVE_OK) {
+    if (res < ARCHIVE_OK) {  // TODO(MINIFICPP-2761)
       logger_->log_error("FocusArchiveEntry can't read header due to archive error: {}", archive_error_string(input_archive.get()));
-      return nlen;
-    }
-
-    if (res < ARCHIVE_WARN) {
-      logger_->log_warn("FocusArchiveEntry got archive warning while reading header: {}", archive_error_string(input_archive.get()));
-      return nlen;
+      return io::IoResult::zero();
```
adamdebreceni, line 184: nlen seems to be updated in this loop or is it always 0 at this point?

minifi-api/common/include/minifi-cpp/io/StreamCallback.h
```diff
@@ -16,24 +16,100 @@
...
+
+  static IoResult fromI64(int64_t i64_val) {
+    if (i64_val < 0) { return IoResult(nonstd::make_unexpected(static_cast<MinifiIoStatus>(i64_val))); }
+    return IoResult(gsl::narrow<uint64_t>(i64_val));
+  }
+
+  static IoResult fromSizeT(const size_t val) {
+    if (isError(val)) { return IoResult::error(); }
+    return IoResult(gsl::narrow<uint64_t>(val));
+  }
```
adamdebreceni, line 55: I think we should either clarify how these differ (usage wise) and let that be reflected in the name or drop the type from the name (or if we want to prevent casts we could move to a template and branch with std::is_same_v and friends)

extensions/libarchive/MergeContent.h
```diff
@@ -82,35 +82,39 @@ class BinaryConcatenationMerge : public MergeBin {
...
         if (!isFirst && !demarcator_.empty()) {
           const auto write_ret = stream->write(reinterpret_cast<const uint8_t*>(demarcator_.data()), demarcator_.size());
-          if (io::isError(write_ret))
-            return -1;
+          if (io::isError(write_ret)) {
+            return io::IoResult::error();
+          }
           write_size_sum += write_ret;
         }
         const auto len = serializer_.serialize(flow, stream);
```
lordgamez, line 103: nitpick: I would rename this as we do not return an integer anymore

extensions/libarchive/FocusArchiveEntry.cpp
martinzink: thats valid 👍  [review change in FocusArchiveEntry.cpp](https://github.com/apache/nifi-minifi-cpp/pull/2148/commits/949213372202dac54da723bce6f8fe257b877ba7)

minifi-api/common/include/minifi-cpp/io/StreamCallback.h
martinzink: [IoResult::fromSizeT and IoResult::fromI64 -> IoResult::from](https://github.com/apache/nifi-minifi-cpp/pull/2148/commits/e26ba0f3807f611605459254842300d2f3104e08)

extensions/libarchive/MergeContent.h
martinzink: [rename len in MergeContent.h](https://github.com/apache/nifi-minifi-cpp/pull/2148/commits/3303dfb6c8829e08b7813e65fd11ef1b543c02f6)

lordgamez (approved):

adamdebreceni (approved):

PR #2151, opened 2026-03-31
MINIFICPP-2762 Replace AWS S3 client with AWS CRT S3 client

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2762

----------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


extensions/aws/utils/AWSInitializer.cpp
```diff
@@ -34,6 +34,12 @@ AWSInitializer& AWSInitializer::get() {
 }
 
 AWSInitializer::~AWSInitializer() {
+  shutdown();
+}
+
+void AWSInitializer::shutdown() {
+  if (is_shut_down_) return;
+  is_shut_down_ = true;
```
szaszm: I think this initializer can be used concurrently from multiple threads, so this needs to be thread-safe. I think changing the flag to an atomic bool and making this change should be sufficient.
```suggestion
  bool is_shut_down = false;
  if (!is_shut_down_.compare_exchange_strong(&is_shut_down, true)) return;
```

extensions/aws/tests/AWSTestMain.cpp
szaszm, line 1: is this still necessary, now that aws shutdown is called on extension unload?

szaszm (approved):

extensions/aws/processors/ListS3.cpp
```diff
@@ -43,17 +43,10 @@ void ListS3::onSchedule(core::ProcessContext& context, core::ProcessSessionFacto
   }
   state_manager_ = std::make_unique<minifi::utils::ListingStateManager>(state_manager);
 
-  auto common_properties = getCommonELSupportedProperties(context, nullptr);
-  if (!common_properties) {
-    throw Exception(PROCESS_SCHEDULE_EXCEPTION, "Required property is not set or invalid");
-  }
-
```
szaszm, line 50: we should remove the EL support flag from the affected properties

szaszm (changes requested):
my approve was a bit too quick, I just noticed the issue with the EL support change

extensions/aws/tests/AWSTestMain.cpp
lordgamez: Good catch, removed in https://github.com/apache/nifi-minifi-cpp/pull/2151/commits/bebc42d718e876c40827af1132d1b92bb9141471

extensions/aws/processors/ListS3.cpp
lordgamez: The `setSupportsExpressionLanguage` flag was removed from the affected properties in AwsProcessor.h and removed from the PROCESSORS.md, is there anything else I'm missing?

adamdebreceni (approved):

martinzink (approved):

szaszm (approved):

PR #2155, opened 2026-04-08
MINIFICPP-2768 use gcc-toolset-14 on rocky, remove expected-lite

martinzink:
compiler support changes: gcc 13, clang 18
removed: expected-lite and other gcc12 workarounds

---

### For all changes:
     in the commit message?

### For code changes:

### Note:


.github/workflows/compiler-support.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        platform: [ { id: gcc-12, base_image: gcc:12-bookworm },
+        platform: [ { id: gcc-13, base_image: gcc:13-bookworm },
```
szaszm, line 17: let's replace the 15 image tag with 15-trixie, the image authors seem to have moved on to debian trixie base with 14 and 15
https://hub.docker.com/_/gcc

core-framework/src/utils/Cron.cpp
```diff
@@ -46,14 +46,15 @@ namespace org::apache::nifi::minifi::utils {
 namespace {
 
 // https://github.com/HowardHinnant/date/issues/550
-// Due to https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78714
-// the month parsing with '%b' and the weekday parsing with '%a' is case-sensitive in gcc11
-// This has been fixed in gcc13
+// Due to libstdc++ bug https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78714
+// the month parsing with '%b' and the weekday parsing with '%a' can be case-sensitive.
+// Normalizing to Title Case prevents parsing failures across all GCC versions.
```
szaszm, line 51: If you want to drop GCC < 13, then we can drop the workaround too.

docker/rockylinux/Dockerfile
```diff
@@ -52,16 +52,13 @@ RUN cd $MINIFI_BASE_DIR && \
 RUN groupadd -g ${GID} ${USER} && useradd -g ${GID} ${USER} && \
     chown -R ${USER}:${USER} ${MINIFI_BASE_DIR}
 
-# Patching standard header to avoid https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105651
-RUN patch -p1 /opt/rh/gcc-toolset-12/root/usr/include/c++/12/bits/basic_string.tcc ${MINIFI_BASE_DIR}/thirdparty/libstdc++/avoid_bogus_Wrestrict_PR105651.patch
```
szaszm, line 56: please remove the patch file too

extensions/standard-processors/processors/InvokeHTTP.cpp
```diff
@@ -79,16 +79,16 @@ void HttpClientStore::returnClient(http::HTTPClient& client) {
...
   if (trimmed_input.ends_with("/s") || trimmed_input.ends_with("/S")) {
     return trimmed_input.substr(0, trimmed_input.size() - 2);
   }
-  return nonstd::make_unexpected(core::ParsingErrorCode::GeneralParsingError);
+  return std::unexpected(core::ParsingErrorCode::GeneralParsingError);
 }
 }  // namespace
 
-nonstd::expected<uint64_t, std::error_code> invoke_http::parseDataTransferSpeed(const std::string_view input) {
+std::expected<uint64_t, std::error_code> invoke_http::parseDataTransferSpeed(const std::string_view input) {
```
szaszm, line 91: monadic operations are standard now, so we should replace utils::andThen with .and_then too

libminifi/src/core/ProcessContextImpl.cpp
```diff
@@ -99,10 +99,10 @@ uint8_t ProcessContextImpl::getMaxConcurrentTasks() const { return processor_.ge
 
 void ProcessContextImpl::yield() { processor_.yield(); }
 
-nonstd::expected<std::string, std::error_code> ProcessContextImpl::getProperty(const std::string_view name, const FlowFile* flow_file) const {
+std::expected<std::string, std::error_code> ProcessContextImpl::getProperty(const std::string_view name, const FlowFile* flow_file) const {
   const auto property = getProcessorInfo().getSupportedProperty(name);
   if (!property) {
-    return nonstd::make_unexpected(PropertyErrorCode::NotSupportedProperty);
+    return std::unexpected(PropertyErrorCode::NotSupportedProperty);
```
szaszm: I would use braced init with `std::unexpected` to make it clear that this is a constructor call, and for the other benefits of direct-list-initialization, e.g. no implicit narrowing. https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es23-prefer-the--initializer-syntax

libminifi/test/unit/ExpectedTest.cpp
szaszm, line 1: the monadic operations are standard, we don't need these utilities or their tests anymore

extensions/sql/CMakeLists.txt
```diff
@@ -23,15 +23,22 @@ endif()
...
 
-include_directories(SYSTEM ../../thirdparty/rapidjson-1.1.0/include/ ../../thirdparty/rapidjson-1.1.0/include/rapidjson)
-include_directories(".")
-
 file(GLOB SOURCES  "*.cpp" "services/*.cpp" "processors/*.cpp"  "data/*.cpp")
 
 add_minifi_library(minifi-sql SHARED ${SOURCES})
 
-if(WIN32)
+# Modern CMake: Target-specific includes instead of global include_directories()
```
szaszm: I don't think this needs a comment, it's expected knowledge of people working with cmake nowadays that legacy global and directory-based commands are to be avoided.

extensions/sql/CMakeLists.txt
```diff
@@ -23,15 +23,22 @@ endif()
...
+target_include_directories(minifi-sql
+    PRIVATE
+        "."
+    SYSTEM PRIVATE
+        "../../thirdparty/rapidjson-1.1.0/include/"
+        "../../thirdparty/rapidjson-1.1.0/include/rapidjson"
```
szaszm: instead of adding the dependency include dirs explicitly, the proper approach would be using target_link_libraries against the rapidjson target

core-framework/common/include/utils/expected.h
szaszm, line 1: the monadic operations have been standardized, so our implementations can be removed

libminifi/test/unit/ExpectedTest.cpp
martinzink: yeah i rather do that in a followup PR, it might have been a bad idea to purge expected lite in the firstplace with these compiler chagnes

extensions/standard-processors/processors/InvokeHTTP.cpp
martinzink: good point however those rewrites can be not straightforward, so not gonna pollute this PR with it

core-framework/src/utils/Cron.cpp
martinzink: yeah it comes up with the later toolkits aswell so we should keep this for all glibc

docker/rockylinux/Dockerfile
martinzink: removed 👍

libminifi/src/core/ProcessContextImpl.cpp
martinzink: 👍 replaced them in [() -> {}](https://github.com/apache/nifi-minifi-cpp/pull/2155/commits/1cae8bceda621416a81d377eaa29b514561513af)

extensions/sql/CMakeLists.txt
martinzink: reverted all this sql changes in favor of https://github.com/apache/nifi-minifi-cpp/pull/2155/changes#diff-9ae861c2a31b58ad29d721cd542711af853d6b783554941173b8a55fe22ecf43R56

core-framework/src/utils/Cron.cpp
szaszm: libstdc++*
glibc is the GNU C standard library, with the version macro `__GLIBC__`
libstdc++ is the GNU C++ standard library, with the version macro `__GLIBCXX__`
It's confusing, maybe they started out as the same project, but they're separate now, and libstdc++ is closely related to GCC (versioned and released together), but totally independent of glibc.

extensions/sql/CMakeLists.txt
```diff
@@ -53,7 +53,7 @@ else()
         CMAKE_COMMAND ""
         UPDATE_COMMAND ""
         INSTALL_COMMAND make install
-        CONFIGURE_COMMAND ./autogen.sh && ./configure --prefix=${IODBC_BYPRODUCT_DIR} --with-pic
+        CONFIGURE_COMMAND ./autogen.sh && ./configure --prefix=${IODBC_BYPRODUCT_DIR} --with-pic CFLAGS=-std=gnu17
```
szaszm: I would append to the CFLAGS instead of replacing it. Plus I'm not sure why this applies from the end of the command, normally per command envvar overrides go before the command, or you use a shell-wide override with export in a separate command

extensions/sql/CMakeLists.txt
martinzink: reverted all this sql changes in favor of https://github.com/apache/nifi-minifi-cpp/pull/2155/changes#diff-9ae861c2a31b58ad29d721cd542711af853d6b783554941173b8a55fe22ecf43R56

extensions/sql/CMakeLists.txt
martinzink: iodbc's configure handles these as env variables explicitly [sql add instead of overwrite](https://github.com/apache/nifi-minifi-cpp/pull/2155/commits/75ecadeb5f06ff52a7c306a67d196c10c6141141)

szaszm (approved):

szaszm (approved):

cmake/Couchbase.cmake
```diff
@@ -30,18 +30,26 @@ set(COUCHBASE_CXX_CLIENT_BUILD_TOOLS OFF CACHE BOOL "" FORCE)
 set(COUCHBASE_CXX_CLIENT_POST_LINKED_OPENSSL ON CACHE BOOL "" FORCE)
 set(COUCHBASE_CXX_CLIENT_INSTALL OFF CACHE BOOL "" FORCE)
 
+# Fix for MSVC + ASIO C++20/23 concept evaluation bugs
+if(MSVC)
+    add_compile_definitions(ASIO_DISABLE_CONCEPTS)
+    # Ensure Windows 10 API target so ASIO networking features are fully enabled
+    add_compile_definitions(_WIN32_WINNT=0x0A00)
```
szaszm: Does this mean dropping pre-win10 windows versions? That's something that I would consider carefully.

extensions/couchbase/CMakeLists.txt
```diff
@@ -21,20 +21,21 @@ if (NOT (ENABLE_ALL OR ENABLE_COUCHBASE))
...
 
 add_minifi_library(minifi-couchbase SHARED ${SOURCES})
 
-target_include_directories(minifi-couchbase PRIVATE BEFORE "${CMAKE_CURRENT_SOURCE_DIR}" "controllerservices" "processors" ${COUCHBASE_INCLUDE_DIR})
-target_link_libraries(minifi-couchbase ${LIBMINIFI} couchbase_cxx_client_static hdr_histogram_static snappy llhttp::llhttp)
+target_include_directories(minifi-couchbase PRIVATE BEFORE "${CMAKE_CURRENT_SOURCE_DIR}" "controllerservices" "processors")
+target_include_directories(minifi-couchbase SYSTEM PRIVATE ${COUCHBASE_INCLUDE_DIR})
+# Demote preprocessor warnings so <ciso646> doesn't break the build
+if (CMAKE_CXX_COMPILER_ID MATCHES "GNU")
+target_compile_options(minifi-couchbase PRIVATE -Wno-error=cpp)
```
szaszm:
```suggestion
    target_compile_options(minifi-couchbase PRIVATE -Wno-error=cpp)
```

fgerlits (approved):

adamdebreceni (approved):

lordgamez (approved):

PR #2167, opened 2026-05-04
MINIFICPP-2792 Fix S3 upload above 2GB on Windows

lordgamez:
- Set content length to avoid narrowing size to int on Windows
- Add MinifiToAwsInputStream to wrap MiNiFi InputStream and avoid stream copy

https://issues.apache.org/jira/browse/MINIFICPP-2792

-----------------

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/extensions/aws/s3/MinifiToAwsInputStream.cpp
+++ b/extensions/aws/s3/MinifiToAwsInputStream.cpp
@@ -1,0 +1,72 @@
+/**
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "MinifiToAwsInputStream.h"
+
+#include <algorithm>
+#include <span>
+
+namespace org::apache::nifi::minifi::aws::s3 {
+
+MinifiInputStreamBuf::int_type MinifiInputStreamBuf::underflow() {
+  if (gptr() < egptr()) {
+    return traits_type::to_int_type(*gptr());
+  }
+  const uint64_t stream_pos = stream_->tell();
+  if (stream_pos >= start_pos_ + content_length_) {
+    return traits_type::eof();
+  }
+  const auto remaining = (start_pos_ + content_length_) - stream_pos;
+  const auto to_read = std::min<uint64_t>(utils::configuration::DEFAULT_BUFFER_SIZE, remaining);
+  const auto bytes_read = stream_->read(std::span(reinterpret_cast<std::byte*>(buffer_.data()), gsl::narrow<size_t>(to_read)));
+  if (io::isError(bytes_read)) {
+    owner_->setstate(std::ios_base::badbit);
+    return traits_type::eof();
+  }
+  if (bytes_read == 0) {
+    return traits_type::eof();
+  }
+  setg(buffer_.data(), buffer_.data(), buffer_.data() + bytes_read);
+  return traits_type::to_int_type(*gptr());
+}
+
+MinifiInputStreamBuf::pos_type MinifiInputStreamBuf::seekoff(off_type off, std::ios_base::seekdir way, std::ios_base::openmode which) {
+  if (!(which & std::ios_base::in)) {
+    return pos_type(off_type(-1));
+  }
+  pos_type new_virtual_pos;
+  if (way == std::ios_base::beg) {
+    new_virtual_pos = pos_type(off);
+  } else if (way == std::ios_base::cur) {
+    const auto phys_pos = static_cast<off_type>(stream_->tell()) - static_cast<off_type>(egptr() - gptr());
+    new_virtual_pos = pos_type(phys_pos - static_cast<off_type>(start_pos_) + off);
+  } else {
+    new_virtual_pos = pos_type(static_cast<off_type>(content_length_) + off);
+  }
+  return seekpos(new_virtual_pos, which);
+}
+
+MinifiInputStreamBuf::pos_type MinifiInputStreamBuf::seekpos(pos_type pos, std::ios_base::openmode which) {
+  if (!(which & std::ios_base::in)) {
+    return pos_type(off_type(-1));
+  }
+  stream_->seek(start_pos_ + static_cast<size_t>(off_type(pos)));
+  setg(buffer_.data(), buffer_.data(), buffer_.data());  // invalidate read buffer
+  return pos;
+}
+
+}  // namespace org::apache::nifi::minifi::aws::s3
--- a/extensions/aws/s3/MinifiToAwsInputStream.cpp
+++ b/extensions/aws/s3/MinifiToAwsInputStream.cpp
@@ -22,48 +22,63 @@
 
 namespace org::apache::nifi::minifi::aws::s3 {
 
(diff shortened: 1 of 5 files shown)


extensions/aws/s3/MinifiToAwsInputStream.h
```diff
@@ -0,0 +1,105 @@
...
+    const uint64_t stream_pos = stream_->tell();
+    if (stream_pos >= start_pos_ + content_length_) {
+      return traits_type::eof();
+    }
+    const auto remaining = (start_pos_ + content_length_) - stream_pos;
+    const auto to_read = std::min<uint64_t>(utils::configuration::DEFAULT_BUFFER_SIZE, remaining);
+    const auto bytes_read = stream_->read(std::span(reinterpret_cast<std::byte*>(buffer_.data()), gsl::narrow<size_t>(to_read)));
+    if (io::isError(bytes_read)) {
+      owner_->setstate(std::ios_base::badbit);
+      return traits_type::eof();
```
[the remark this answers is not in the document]
lordgamez: This stream is directly passed to the AWS SDK. Minifi is not handling the error in case of read failure, but the AWS SDK does, and it checks the state instead of handling and exception. I tested that in case of read failure AWS returns a failure with AWS_IO_STREAM_READ_FAILED with this solution so it should be okay.

extensions/aws/s3/S3Wrapper.cpp
```diff
@@ -68,32 +69,11 @@ std::string S3Wrapper::getEncryptionString(Aws::S3Crt::Model::ServerSideEncrypti
...
-}
-
 std::optional<PutObjectResult> S3Wrapper::putObject(const PutObjectRequestParameters& put_object_params, const std::shared_ptr<io::InputStream>& stream, uint64_t flow_size) {
-  uint64_t read_size{};
-  auto data_stream = readFlowFileStream(stream, flow_size, read_size);
   auto request = createPutObjectRequest<Aws::S3Crt::Model::PutObjectRequest>(put_object_params);
-  request.SetBody(data_stream);
+  auto aws_stream = std::make_shared<MinifiToAwsInputStream>(stream, flow_size);
+  request.SetBody(aws_stream);
+  request.SetContentLength(static_cast<long long>(flow_size));  // NOLINT(runtime/int,google-runtime-int) AWS SDK expects long long for content length
```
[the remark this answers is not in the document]
lordgamez: We do not have support for AWS docker tests for Windows yet, so it is not possible. We also tried testing large >1GB flow files in the past and it did not work in the CI environment. This test should be considered in the future, but not viable at the moment

extensions/aws/s3/MinifiToAwsInputStream.cpp
```diff
@@ -0,0 +1,72 @@
...
+  if (stream_pos >= start_pos_ + content_length_) {
+    return traits_type::eof();
+  }
+  const auto remaining = (start_pos_ + content_length_) - stream_pos;
+  const auto to_read = std::min<uint64_t>(utils::configuration::DEFAULT_BUFFER_SIZE, remaining);
+  const auto bytes_read = stream_->read(std::span(reinterpret_cast<std::byte*>(buffer_.data()), gsl::narrow<size_t>(to_read)));
+  if (io::isError(bytes_read)) {
+    owner_->setstate(std::ios_base::badbit);
+    return traits_type::eof();
+  }
```
szaszm, line 44: Is this the correct way to propagate IO errors through iostreams?

extensions/aws/s3/MinifiToAwsInputStream.cpp
```diff
@@ -0,0 +1,72 @@
...
+  setg(buffer_.data(), buffer_.data(), buffer_.data() + bytes_read);
+  return traits_type::to_int_type(*gptr());
+}
+
+MinifiInputStreamBuf::pos_type MinifiInputStreamBuf::seekoff(off_type off, std::ios_base::seekdir way, std::ios_base::openmode which) {
+  if (!(which & std::ios_base::in)) {
+    return {off_type(-1)};
+  }
```
szaszm, line 58: A code comment explaining what and why is happening would be helpful

extensions/aws/s3/MinifiToAwsInputStream.cpp
```diff
@@ -0,0 +1,72 @@
...
+}
+
+MinifiInputStreamBuf::pos_type MinifiInputStreamBuf::seekpos(pos_type pos, std::ios_base::openmode which) {
+  if (!(which & std::ios_base::in)) {
+    return {off_type(-1)};
+  }
+  stream_->seek(start_pos_ + static_cast<size_t>(off_type(pos)));
+  setg(buffer_.data(), buffer_.data(), buffer_.data());  // invalidate read buffer
+  return pos;
+}
```
fgerlits, line 88: Should we check that `0 <= pos <= content_length_`? (or should that be `< content_length_`?)

lordgamez: Good point, added check in https://github.com/apache/nifi-minifi-cpp/pull/2167/commits/ee61619f21d579ef538406b96a4e291dc51cf3e2

martinzink (approved):

fgerlits (approved):

extensions/aws/s3/MinifiToAwsInputStream.cpp
lordgamez: Yes, I tested this manually by setting the error state explicitly and AWS SDK propagated the error correctly to the client application.

extensions/aws/s3/MinifiToAwsInputStream.cpp
lordgamez: Added comments in https://github.com/apache/nifi-minifi-cpp/pull/2167/commits/9f4ca8f6a14ef9a7f7430ddf2f91a06b43c58ab3

szaszm:
Could you please specify which parts of the code were AI generated and which were hand written?

lordgamez:
> Could you please specify which parts of the code were AI generated and which were hand written?

What I can say is that the initial version of MinifiInputStreamBuf was generated using AI and then edited afterwards, but as this was a month ago with multiple iterations I cannot say specifically which parts were changed, rewritten specifically.

extensions/aws/s3/MinifiToAwsInputStream.h
```diff
@@ -0,0 +1,60 @@
...
+
+ private:
+  std::shared_ptr<io::InputStream> stream_;
+  uint64_t start_pos_;
+  uint64_t content_length_;
+  std::vector<char> buffer_;
+  gsl::not_null<std::basic_ios<char>*> owner_;
+};
+
+class MinifiToAwsInputStream : private MinifiInputStreamBuf, public std::basic_iostream<char> {
```
szaszm: I'm fine with AI code as long as it matches the level quality I would expect from the author. Here, I think any human would have used the std::iostream alias, and in general people would use the std::X alias over the std::basic_X<char> full type, when only char is used.

lordgamez: I don't think this is the case here, the AWS SDK uses typedefs using std::basic_X types in its stream definitions, that's why std::basic_X types are used here explicitly.

```
namespace Aws
{

// Serves no purpose other than to help my conversion process
typedef std::basic_ifstream< char, std::char_traits< char > > IFStream;
typedef std::basic_ofstream< char, std::char_traits< char > > OFStream;
typedef std::basic_fstream< char, std::char_traits< char > > FStream;
typedef std::basic_istream< char, std::char_traits< char > > IStream;
typedef std::basic_ostream< char, std::char_traits< char > > OStream;
typedef std::basic_iostream< char, std::char_traits< char > > IOStream;
typedef std::istreambuf_iterator< char, std::char_traits< char > > IStreamBufIterator;

using IOStreamFactory = std::function< Aws::IOStream*(void) >;


} // namespace Aws
```

lordgamez: But regardless how the AWS SDK defines these types, it's a good point that we should shorten these types in our context. It probably didn't occur to me because I was checking the AWS defined types and it was explicit there. I think it does not make a difference if that was generated by AI or not, if it's a valid comment as in this case, then it should be addressed by the author. Updated in https://github.com/apache/nifi-minifi-cpp/pull/2167/commits/18863d9b7899de9483af0aeff986d481c60dcf8d

szaszm (approved):

PR #2168, opened 2026-05-04
MINIFICPP-2784 Fixing OpenSSL cmake

martinzink:
https://issues.apache.org/jira/browse/MINIFICPP-2784
The bundled openssl was not strict enough and sometimes it used the system headers while building, this caused build errors on systems where the openssl headers were not compatible with our bundled version. (e.g. arch linux, trixie, etc)

This also fixes the currently failing CI jobs namely
* 🟢 [MiNiFi-CPP memcheck](https://github.com/martinzink/nifi-minifi-cpp/actions/runs/25181169091) 
* 🟢 [Check supported Compilers](https://github.com/martinzink/nifi-minifi-cpp/actions/runs/25181160322) 

---

### For all changes:
     in the commit message?

### For code changes:

### Note:


cmake/BundledLibcURL.cmake
```diff
@@ -16,91 +16,57 @@
 # under the License.
 
 function(use_bundled_curl SOURCE_DIR BINARY_DIR)
```
lordgamez: I would remove this `use_bundled_curl` function and only include this cmake where used, also rename the removing the `Bundled` prefix to be consistent with other FetchContent using cmake files. Same for the cmake/BundledLibSSH2.cmake file.

martinzink: [BundledLibcURL.cmake -> FetchLibcURL.cmake, BundledLibSSH2.cmake -> F…](https://github.com/apache/nifi-minifi-cpp/pull/2168/commits/68b2ac27c9616e033886f71d3606238cbfea529a)

lordgamez (approved):

fgerlits (approved):

cmake/ssl/FindCrypto.cmake
```diff
@@ -0,0 +1,56 @@
...
+if(NOT crypto_LIBRARY OR NOT EXISTS "${crypto_INCLUDE_DIR}")
+    message(FATAL_ERROR "Failed to locate bundled crypto (libcrypto) inside ${OPENSSL_ROOT_DIR}")
+endif()
+
+set(CRYPTO_FOUND TRUE)
+set(crypto_FOUND TRUE)
+set(crypto_SHARED_LIBRARY "")
+set(crypto_STATIC_LIBRARY "${crypto_LIBRARY}")
+
+add_library(AWS::crypto UNKNOWN IMPORTED)
```
szaszm, line 46: Do we need an AWS::crypto target when the AWS SDK was patched to use OpenSSL::Crypto?

martinzink: AWS SDK still has some references like
`list(APPEND PLATFORM_LIBS  AWS::crypto)`
we could try to eradicate them aswell and replace them with OpenSSL::Crypto via patching but this seems more long lived because an update might introduces another AWS::crypto dependency

szaszm: makes sense, let's leave it then, thx

szaszm (approved):

PR #2170, opened 2026-05-05
MINIFICPP-2798 Mock library for C extensions

martinzink:
Separates the CPP wrapper into a abstract interface and the Cffi layer (that commincates via the stable API), and also introduces a Mock alternative implementation, with its easier to create new unit tests that only test the processors controllers services business logic without the need to depend on the libminifi library.

---

### For all changes:
     in the commit message?

### For code changes:

### Note:


extension-framework/cpp-extension-lib/mocklib/src/MockProcessSession.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+#include "MockStreams.h"
+
+namespace org::apache::nifi::minifi::mock {
+
+MockProcessSession::~MockProcessSession() {
+  // api::core::FlowFile requires that somebody takes ownership before it goes out of scope
+  // the tested processor should either remove or transfer all FlowFiles it handles
+  for (auto& ff : removed_flow_files_) {
+    delete ff.release();  // NOLINT(cppcoreguidelines-owning-memory)
```
adamdebreceni: isn't this the same as ff.reset()?

martinzink: No because ff is a 

```
struct EnsureMovedFromDeleter {
  void operator()(MinifiFlowFile* ff) {
    if (ff) {
      throw std::logic_error("Each flowfile should be either transferred or removed");
    }
  }
};

using FlowFile = std::unique_ptr<MinifiFlowFile, EnsureMovedFromDeleter>;
```

if we call ff.reset() or simply let the unique_ptr go out of scope the custom deleter will run and throw, with release and delete we skip the custom deleter (which is warranted in this mock environment)

adamdebreceni: you are right my bad

adamdebreceni (approved):

extension-framework/cpp-extension-lib/mocklib/src/MockProcessSession.cpp
martinzink: we further discussed this on other channels and came up with a simpler solution that doesnt require this ugliness
[simplify](https://github.com/apache/nifi-minifi-cpp/pull/2170/commits/97edf58a31903572b384daf0b319023cc94b8b11)

extension-framework/cpp-extension-lib/include/api/utils/Ssl.h
```diff
@@ -0,0 +1,55 @@
...
+  }
+};
+
+struct SslServerOptions {
+  SslData cert_data;
+  ClientAuthOption client_auth_option;
+
+  SslServerOptions(SslData cert_data, ClientAuthOption client_auth_option)
+      : cert_data(cert_data),
+      client_auth_option(client_auth_option) {}
```
szaszm: we could remove this constructor and leave the struct to be an aggregate
```suggestion
```

extension-framework/cpp-extension-lib/mocklib/src/MockProcessSession.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+#include "MockProcessSession.h"
+
+#include "MockStreams.h"
+
+namespace org::apache::nifi::minifi::mock {
+
+api::core::FlowFile MockProcessSession::create(const api::core::FlowFile*) {
+  auto new_ff = api::core::FlowFile{new MinifiFlowFile};  // NOLINT(clang-analyzer-cplusplus.NewDeleteLeaks)
+  return new_ff;
```
lordgamez: Can we merge these two lines?

extension-framework/cpp-extension-lib/mocklib/src/MockProcessSession.cpp
```diff
@@ -0,0 +1,88 @@
...
+}
+api::core::FlowFile MockProcessSession::get() {
+  if (input_flow_files_.empty()) { return nullptr; }
+
+  auto ff = std::move(input_flow_files_.back());
+  input_flow_files_.pop_back();
+
+  return api::core::FlowFile{ff.release()};
```
lordgamez, line 33:
```suggestion
  return {ff.release()};
```

lordgamez (approved):
Great stuff 👍 LGTM just some minor nitpicks.

extension-framework/cpp-extension-lib/mocklib/src/MockProcessSession.cpp
martinzink: `Cannot convert braced-init-list to return type api::core::FlowFile: chosen constructor is explicit`

extension-framework/cpp-extension-lib/mocklib/src/MockProcessSession.cpp
martinzink: [review changes](https://github.com/apache/nifi-minifi-cpp/pull/2170/commits/d23b6b8bfa23516d0c999d51ab2eb420a1fb893d)

extension-framework/cpp-extension-lib/mocklib/include/MockProcessSession.h
```diff
@@ -0,0 +1,80 @@
...
+  std::optional<std::string> getAttribute(api::core::FlowFile& ff, std::string_view key) override;
+  [[nodiscard]] std::map<std::string, std::string> getAttributes(const api::core::FlowFile& ff) const override;
+  [[nodiscard]] std::string getFlowFileId(const api::core::FlowFile& ff) const override;
+  [[nodiscard]] uint64_t getFlowFileSize(const api::core::FlowFile& ff) const override;
+
+  void addInputFlowFile(std::unique_ptr<MinifiFlowFile> flow_file);
+
+ private:
+  std::vector<std::unique_ptr<MinifiFlowFile>> input_flow_files_;
+  std::map<std::string, std::vector<std::unique_ptr<MinifiFlowFile>>> transferred_flow_files_;
```
szaszm: I'm guessing relationships are the key, I would add an inline comment to indicate that. Optionally you could also replace it with a Relationship object, the same one as session transfer takes
```suggestion
  std::map<std::string /* relationship name */, std::vector<std::unique_ptr<MinifiFlowFile>>> transferred_flow_files_;
```

extension-framework/cpp-extension-lib/include/api/utils/Ssl.h
fgerlits, line 1: This looks like a copy of `extension-framework/include/utils/net/Ssl.h`. Can we remove the old file, or do we need both?

martinzink: Currently we need them both because there is no common lib above the legacy cpp extensions and the new c api based cpp extensions thats appropriate for this file.

image

We could refactor this, but thats probably out of the scope of this PR

extension-framework/cpp-extension-lib/mocklib/include/MockProcessSession.h
martinzink: good idea, https://github.com/apache/nifi-minifi-cpp/pull/2170/changes/62fa423ff10fb91159958404c532cb4daccde878#diff-47d53d9e90b5146cb43fa62e059648d9f5f6386797cecb1d4b27fcacb47e3d26L76-L77

i didnt want to replace it with the Relationship object since its not copyable afaik

extension-framework/cpp-extension-lib/include/api/utils/Ssl.h
martinzink: 👍 https://github.com/apache/nifi-minifi-cpp/pull/2170/changes/62fa423ff10fb91159958404c532cb4daccde878#diff-8802588662245e0a79b9fa9c322e8a6f27712e81a90ce63062b877cb8ec3d589L48-L51

fgerlits (approved):

szaszm (approved):

PR #2174, opened 2026-05-14
MINIFICPP-2813 Upgrade thirdparty libraries part 1

lordgamez:
This is the first part of upgrading the thirdparty libraries before releasing MiNiFi C++ 1.0:
- Upgrade jsoncons lib to v1.7.0
- Upgrade paho.mqtt.c library to v1.3.16
- Upgrade grpc to v1.80.0
- Upgrade benchmark lib to v1.9.5
- Upgrade ArgParse lib to v3.2
- Upgrade librdkafka to v2.14.1
- Upgrade Absiel lib to version 20260107.1
- Upgrade RpMalloc to v1.4.5
- Upgrade liblzma to v5.8.3
- Upgrade lz4 library to v1.10.0
- Upgrade zlib to v1.3.2
- Upgrade Lua to version 5.4.8
- Upgrade magic_enum lib to v0.9.8
- Upgrade zstd to v1.5.7
- Upgrade libssh2 to version 1.11.1
- Upgrade json-schema-validator to v2.4.0 and nlohmann_json to v3.12.0
- Upgrade kubernetes lib to v0.14.0 and libwebsockets lib to v4.5.8
- Upgrade mimalloc to version 3.3.2
- Upgrade yaml-cpp to version 0.9.0
- Upgrade MbedTLS to v3.6.6 and Open62541 to v1.5.4
- Upgrade alpine docker base image to version 3.22.4
- Upgrade gsl-lite to version 1.1.0

**Note**: Each upgrade is done in a separate commit, so it is recommended to review the PR commit by commit for easier understanding.

-------------

### For all changes:
     in the commit message?

### For code changes:

### Note:


cmake/KubernetesClientC.cmake
```diff
@@ -43,21 +42,23 @@ set(WEBSOCKETS_PC ${Bash_EXECUTABLE} -c "set -x &&\
...
-        URL_HASH        SHA256=6a85a1bccf25acc7e8e5383e4934c9b32a102880d1e4c37c70b27ae2a42406e1
+        URL             https://github.com/warmcat/libwebsockets/archive/refs/tags/v4.5.8.tar.gz
+        URL_HASH        SHA256=b6ade658f4af3a823d0dc806ae5ef0623f0f4f5e2aeb895a0f77c4783840c30e
         PATCH_COMMAND "${WEBSOCKETS_PC}"
         SYSTEM
 )
 
 FetchContent_MakeAvailable(yaml websockets)
 
+target_compile_options(websockets PRIVATE -fPIC)
```
szaszm: is this line necessary? why?

extensions/rocksdb-repos/FlowFileRepository.cpp
```diff
@@ -151,7 +151,8 @@ void FlowFileRepository::initialize_repository() {
   const auto it = opendb->NewIterator(options);
   for (it->SeekToFirst(); it->Valid(); it->Next()) {
     utils::Identifier container_id;
-    auto eventRead = FlowFileRecord::DeSerialize(gsl::make_span(it->value()).as_span<const std::byte>(), content_repo_, container_id);
+    const auto slice = it->value();
+    auto eventRead = FlowFileRecord::DeSerialize(std::span<const std::byte>(reinterpret_cast<const std::byte*>(slice.data()), slice.size()), content_repo_, container_id);
```
szaszm: I didn't check yet, but we should see if we can avoid the reinterpret_cast in favor of as_bytes or as_writable_bytes.

cmake/KubernetesClientC.cmake
lordgamez: I'm not sure why was this cflag needed originally and neither @fgerlits have a reason why this was needed. This change was only added, because the global cmake flag caused issues with other libraries. I checked and the library works correctly without compiling it with the -fPIC flag, so I removed it in https://github.com/apache/nifi-minifi-cpp/pull/2174/commits/9269fbda0a0c740b99fc96fd9cb6b9494dc6232e

extensions/rocksdb-repos/FlowFileRepository.cpp
lordgamez: It seems to be working, updated in https://github.com/apache/nifi-minifi-cpp/pull/2174/commits/9269fbda0a0c740b99fc96fd9cb6b9494dc6232e

cmake/ZLIB.cmake
```diff
@@ -0,0 +1,45 @@
...
+include(FetchContent)
+
+set(ZLIB_BUILD_TESTING OFF CACHE BOOL "" FORCE)
+set(ZLIB_BUILD_SHARED OFF CACHE BOOL "" FORCE)
+
+FetchContent_Declare(
+    zlib
+    URL "https://github.com/madler/zlib/archive/refs/tags/v1.3.2.tar.gz"
+    URL_HASH "SHA256=b99a0b86c0ba9360ec7e78c4f1e43b1cbdf1e6936c8fa0f6835c0cd694a495a1"
+    SYSTEM
```
martinzink, line 27: With OVERRIDE_FIND_PACKAGE we might be able to get rid of the dummy FindZLIB.cmake

martinzink (commented):
One small improvement idea, but if that causes any compilation issue feel free to ignore it.

Otherwise LGTM great work 👍

fgerlits (approved):

cmake/ZLIB.cmake
lordgamez: Good idea, but unfortunately for some reason it does not seem to work with RockyLinux, maybe the FindZLIB.cmake module is different in the rocky image.
```
2026-05-20T11:12:40.7138217Z #12 420.4 CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
2026-05-20T11:12:40.7139389Z #12 420.4   Could NOT find ZLIB (missing: ZLIB_LIBRARY ZLIB_INCLUDE_DIR)
2026-05-20T11:12:40.7140224Z #12 420.4 Call Stack (most recent call first):
2026-05-20T11:12:40.7141317Z #12 420.4   /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:600 (_FPHSA_FAILURE_MESSAGE)
2026-05-20T11:12:40.7142623Z #12 420.4   /usr/share/cmake/Modules/FindZLIB.cmake:200 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
2026-05-20T11:12:40.7143601Z #12 420.4   CMakeLists.txt:166 (find_package)
2026-05-20T11:12:40.7144259Z #12 420.4 
2026-05-20T11:12:40.7145102Z #12 420.4 
2026-05-20T11:12:40.7145688Z #12 420.4 -- Configuring incomplete, errors occurred!
```
I am not sure if it's worth debugging, I would keep the current solution for now.

martinzink (approved):

extensions/rocksdb-repos/FlowFileRepository.cpp
szaszm: There are a lot more instances of this pattern, could you go through all of them in this diff?

extensions/mqtt/tests/ConsumeMQTTTests.cpp
```diff
@@ -94,6 +96,18 @@ class TestConsumeMQTTProcessor : public minifi::processors::ConsumeMQTT {
...
+  auto* raw_msg = static_cast<MQTTAsync_message*>(std::malloc(sizeof(MQTTAsync_message)));  // NOLINT(cppcoreguidelines-owning-memory,cppcoreguidelines-no-malloc)
+  auto* payload_copy = static_cast<char*>(std::malloc(payload.size() + 1));  // NOLINT(cppcoreguidelines-owning-memory,cppcoreguidelines-no-malloc)
+  std::memcpy(payload_copy, payload.data(), payload.size());
+  payload_copy[payload.size()] = '\0';
+  *raw_msg = MQTTAsync_message{.struct_id = {'M', 'Q', 'T', 'M'}, .struct_version = struct_version,
+                               .payloadlen = gsl::narrow<int>(payload.size()), .payload = payload_copy,
+                               .qos = qos, .retained = retained, .dup = dup, .msgid = msgid, .properties = {}};
+  return std::unique_ptr<MQTTAsync_message, TestConsumeMQTTProcessor::MQTTMessageDeleter>(raw_msg);  // NOLINT(clang-analyzer-unix.Malloc)
+}
+
```
szaszm, line 110: There are a lot of unexplained unsafe operations in this block. Is this old code moved from somewhere else? If not, why did you do it this way?

CMakeLists.txt
```diff
@@ -302,7 +301,14 @@ include(GslLite)
 
 # Add necessary definitions based on the value of STRICT_GSL_CHECKS, see gsl-lite README for more details
 list(APPEND GslDefinitions gsl_CONFIG_DEFAULTS_VERSION=1)
-list(APPEND GslDefinitionsNonStrict gsl_CONFIG_CONTRACT_VIOLATION_THROWS gsl_CONFIG_NARROW_THROWS_ON_TRUNCATION=1)
+list(APPEND GslDefinitions gsl_FEATURE_BYTE=1)
+list(APPEND GslDefinitions gsl_CONFIG_NARROW_THROWS_ON_TRUNCATION=1)
```
szaszm: Why did you move this from NonStrict to default/strict mode?

extensions/mqtt/tests/ConsumeMQTTTests.cpp
lordgamez: This was a bug that previously we were using `new` for the test message allocation, while the `MQTTMessageDeleter` was using `MQTTAsync_freeMessage` which uses `free` for deallocation. The problem did not occur before as we weren't using the `PAHO_HIGH_PERFORMANCE` cmake option, which made the mqtt library use its own allocator that had a heap tracker. As the heap tracker did not track our `new` created message it was a noop when calling `MQTTAsync_freeMessage` and leaked the message object.

With the introduction of `PAHO_HIGH_PERFORMANCE` cmake option it crashed because it uses `free` directly on the `new` created message. With this change `malloc` is used to create the message to avoid the issue.

CMakeLists.txt
lordgamez: This was my mistake putting it in the `GslDefinitionsNonStrict` definitions which is only used in Debug mode, it should be defined in all cases.

extensions/rocksdb-repos/FlowFileRepository.cpp
lordgamez: I could only find one instance in the diff and one other instance in a separate test, fixed in https://github.com/apache/nifi-minifi-cpp/pull/2174/commits/fa0d11adc25fd5a7d514bc76bf7b726c28a96608 Please point them out if I missed any other instances.

CMakeLists.txt
szaszm: I don't think it was a mistake. The idea with STRICT_GSL_CHECKS was that by default, in debug mode, contract violations crash (std::terminate), while in release mode, they throw, but by changing this option to strict, you can force release builds to crash early too. A narrowing conversion that results in value change is a contract violation, similar to failed gsl_Expects.

In the new version of gsl-lite, the defaults have changed, so the configuration needs to be rewritten. For example, gsl_CONFIG_NARROW_THROWS_ON_TRUNCATION=1 is the default now, but we probably only want it in non strict mode. In strict mode, it should terminate, not throw. There are other config and default changes in their docs too, that's why I'm advocating for a full rewrite of this section.

lordgamez: In https://github.com/apache/nifi-minifi-cpp/pull/2174/commits/02aac2d186f9d1c460d2bca9453073c3fd3a1524 I added `gsl_CONFIG_NARROW_THROWS_ON_TRUNCATION=0` to `GslDefinitionsNonStrict` and removed `gsl_CONFIG_DEFAULTS_VERSION=1` as it is already the default. For other options that have changed their defaults since the previously used version:

- The new defaults of `gsl_FEATURE_STRING_SPAN` and `gsl_CONFIG_ALLOWS_SPAN_COMPARISON` are consistent with C++20 `std::span` so we can leave it at that
- `gsl_FEATURE_BYTE`: we already set it to 1 manually
- `gsl_CONFIG_DEPRECATE_TO_LEVEL`: default changed to 9 which is good to have the highest deprecation level on by default
- `gsl_CONFIG_INDEX_TYPE`: defaults to a more domain specific index type which we do not conflict with so we can leave it at that
- `gsl_CONFIG_NOT_NULL_EXPLICIT_CTOR`: default 1 is used for explicit ctor due to https://github.com/Microsoft/GSL/issues/395 so the default makes sense, it is only suggested to change it as a workaround if the legacy code conflicts with it
- `gsl_CONFIG_TRANSPARENT_NOT_NULL`: the default is 1 using a transparent not_null type which conforms with our usage
- default runtime contract violation handling: new default is `gsl_CONFIG_CONTRACT_VIOLATION_ASSERTS` instead of `gsl_CONFIG_CONTRACT_VIOLATION_TERMINATES` which is a subtle change calling `std::abort` instead of `std::terminate` but with more information in case of a crash. Reason given by gsl-lite: the mode enabled by `gsl_CONFIG_CONTRACT_VIOLATION_ASSERTS` is consistent with the behavior of the `assert()` macro while retaining runtime contract checks even if `NDEBUG` is defined. With this explanation I think we can keep that as well on default setting.

szaszm: gsl_CONFIG_NARROW_THROWS_ON_TRUNCATION: If we wanted to keep the old behavior, it would need to be 1 in non-strict/release mode and 0 in normal mode. But we can also reevaluate this behavior. Back when we introduced gsl-lite, narrow truncation was treated as any other contract check failure, but today the C++ Core Guidelines specifies that narrow throws on truncation. I think we can change the behavior of narrow to match modern guidelines and leave this setting on default.

gsl_FEATURE_BYTE: why do we already set it to 1 manually?

lordgamez: gsl_CONFIG_NARROW_THROWS_ON_TRUNCATION: Good point, matching the C++ core guidlines sounds good to me, removed it to use the default in https://github.com/apache/nifi-minifi-cpp/pull/2174/commits/a0639e35ad2b4907e97fe10bee74224e59cd7d40

gsl_FEATURE_BYTE: I just realized we do not need this anymore, I thought it is used for something different, but it only provides byte, as_bytes, as_writable_bytes etc. implementations that we already use the std library version of. I must have been a leftover from when older compilers were supported. Removed it in https://github.com/apache/nifi-minifi-cpp/pull/2174/commits/a0639e35ad2b4907e97fe10bee74224e59cd7d40

CMakeLists.txt
```diff
@@ -301,8 +300,12 @@ target_compile_definitions(RapidJSON INTERFACE RAPIDJSON_HAS_STDSTRING)
 include(GslLite)
 
 # Add necessary definitions based on the value of STRICT_GSL_CHECKS, see gsl-lite README for more details
-list(APPEND GslDefinitions gsl_CONFIG_DEFAULTS_VERSION=1)
```
szaszm, line 304: Why did you choose to delete this definition? I'd keep the config defaults version explicitly specified.

thirdparty/libwebsockets/openssl3.patch
szaszm, line 1: what's the source of this patch? Is it still necessary?

CMakeLists.txt
lordgamez: I removed it because that's the default, but we can keep it to be explicit if you prefer.

thirdparty/libwebsockets/openssl3.patch
lordgamez: I wrote it, it is needed to specify the flags manually for the bundled openssl otherwise the build tries to check these flags through compiling examples using the system installed openssl to check which feature is available.

szaszm: Can you add this info to the top of the file? (that you wrote it and because otherwise it would use the system openssl otherwise)

fyi: This patch captured my attention because it's related to openssl, and patching crypto code should be a last resort.

CMakeLists.txt
szaszm: I think I'd prefer to keep this one explicitly specified, in order to not accidentally defeat gsl-lite backwards compatibility efforts by automatically adopting a breaking change in default settings.

extensions/rocksdb-repos/FlowFileRepository.cpp
szaszm: I couldn't find any more either, sorry for exaggerating their prevalence. On my first review, I got the impression that there are many instances of this pattern, but there aren't as many as I thought. 2 days ago in my last reply, I was replying under this assumption, without reviewing all the files again.

CMakeLists.txt
lordgamez: Readded in https://github.com/apache/nifi-minifi-cpp/pull/2174/commits/5e4bfe44f8326b5bfbb833bf5a90ede76b0faa21

thirdparty/libwebsockets/openssl3.patch
lordgamez: Added in https://github.com/apache/nifi-minifi-cpp/pull/2174/commits/5e4bfe44f8326b5bfbb833bf5a90ede76b0faa21

szaszm (approved):

PR #2176, opened 2026-05-15
MINIFICPP-2816 Add controller_service::api_implementations to C api

martinzink:

### For all changes:
     in the commit message?

### For code changes:

### Note:


core-framework/include/core/controller/ControllerServiceFactoryImpl.h
```diff
@@ -34,13 +34,13 @@ class ControllerServiceFactoryImpl : public ControllerServiceFactory {
       : class_name_(core::className<T>()) {
   }
 
-  explicit ControllerServiceFactoryImpl(std::string group_name)
-      : group_name_(std::move(group_name)),
+  explicit ControllerServiceFactoryImpl(std::string module_name)
```
martinzink, line 37: Ive renamed it from group_name to module_name, because this groupname was not the groupname in C2 but rather the artifact name

core-framework/include/core/ObjectFactory.h
```diff
@@ -31,7 +31,7 @@ class ObjectFactoryImpl : public ObjectFactory {
 
   ObjectFactoryImpl() = default;
 
-  std::string getGroupName() const override {
+  std::string getModuleName() const override {
     return group_;
```
adamdebreceni: we could also update the member

libminifi/src/minifi-c.cpp
```diff
@@ -164,7 +164,7 @@ class CControllerServiceFactory : public minifi::core::controller::ControllerSer
     return std::make_unique<minifi::utils::CControllerService>(class_description_, std::move(metadata));
   }
 
-  [[nodiscard]] std::string getGroupName() const override { return group_name_; }
+  [[nodiscard]] std::string getModuleName() const override { return group_name_; }
```
adamdebreceni: we could also update the group_name_ as well here

adamdebreceni (approved):

libminifi/src/core/state/nodes/AgentInformation.cpp
```diff
@@ -17,46 +17,91 @@
...
+  auto pos = sv.rfind('.');
+  return (pos == std::string_view::npos) ? sv : sv.substr(pos + 1);
+}
+
+bool minifiSystemAllowedType(const std::string_view allowed_type) {
+  return allowed_type == controllers::SSLContextServiceInterface::ProvidesApi.type
+      || allowed_type == core::RecordSetWriter::ProvidesApi.type
+      || allowed_type == core::RecordSetReader::ProvidesApi.type
+      || allowed_type == controllers::ProxyConfigurationServiceInterface::ProvidesApi.type
+      || allowed_type == controllers::AttributeProviderService::ProvidesApi.type;
```
lordgamez, line 58: Would it be possible to
1) create a static vector to store these types in and the controller classes could register themselves in at startup?
2) or if that's too complicated at least extract these to an array where they would be listed manually and check if the passed type exists there

martinzink: yeah the first one seems a good idea, but thats a whole new PR i think, I did the second one here [refactor minifiSystemAllowedType](https://github.com/apache/nifi-minifi-cpp/pull/2176/commits/5955a539da1399c685caa9a23f493ef5a92e6c48) not sure if thats what u had in mind

lordgamez (approved):

extension-framework/cpp-extension-lib/include/api/core/Resource.h
```diff
@@ -181,6 +191,20 @@ void useControllerServiceClassDefinition(Fn&& fn) {
               static_cast<Class*>(self)->disable();
             } catch (...) {}
           },
+          .get_interface = [](void* self, MinifiStringView interface_name) -> void* {
+            try {
+              const std::string_view name_view{interface_name.data, interface_name.length};
+
+              if constexpr (requires { Class::ProvidedInterfaces; }) {
```
szaszm: tiny thing, doesn't really matter, but I'd only create the view if it's needed.
```suggestion
              if constexpr (requires { Class::ProvidedInterfaces; }) {
                const std::string_view name_view{interface_name.data, interface_name.length};
```

extensions/stable-api-sandbox/CMakeLists.txt
szaszm, line 1: I'm thinking we should call this stable-api-tests or something like that, so even at first glance it's clear that this is a test-only extension.

extensions/stable-api-sandbox/ZooProcessor.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include "AnimalControllerServices.h"
+#include "api/core/ProcessContext.h"
+
+namespace org::apache::nifi::minifi::api_sandbox {
+
+MinifiStatus ZooProcessor::onTriggerImpl(api::core::ProcessContext& process_context, api::core::ProcessSession& process_session) {
+  if (const auto can_fly_opaque = process_context.getControllerService(CanFlyService)) {
+    if (*can_fly_opaque) {
+      const auto can_fly_controller_name = process_context.getProperty(CanFlyService, nullptr) | utils::orThrow("Should be here");
+      const CanFlyControllerApi* can_fly = reinterpret_cast<CanFlyControllerApi*>(*can_fly_opaque);
```
szaszm:
```suggestion
      const CanFlyControllerApi* const can_fly = reinterpret_cast<CanFlyControllerApi*>(*can_fly_opaque);
```

extensions/stable-api-sandbox/ZooProcessor.cpp
```diff
@@ -0,0 +1,46 @@
...
+      logger_->log_critical("Can {} fly? {}", can_fly_controller_name, can_fly->canFly());
+    }
+  }
+  if (const auto num_of_legs_opaque = process_context.getControllerService(NumberOfLegsService)) {
+    if (*num_of_legs_opaque) {
+      const auto num_of_legs_name = process_context.getProperty(NumberOfLegsService, nullptr) | utils::orThrow("Should be here");
+      const NumberOfLegsControllerApi* num_legs = reinterpret_cast<NumberOfLegsControllerApi*>(*num_of_legs_opaque);
```
szaszm:
```suggestion
      const NumberOfLegsControllerApi* const num_legs = reinterpret_cast<NumberOfLegsControllerApi*>(*num_of_legs_opaque);
```

extensions/stable-api-sandbox/ZooProcessor.h
```diff
@@ -0,0 +1,63 @@
...
+
+class ZooProcessor : public api::core::ProcessorImpl {
+ public:
+  EXTENSIONAPI static constexpr const char* Description = "Test ZooProcessor";
+
+  EXTENSIONAPI static constexpr auto CanFlyService = core::PropertyDefinitionBuilder<>::createProperty("Can fly service")
+                                                         .withDescription("Test CanFlyService")
+                                                         .isRequired(true)
+                                                         .withAllowedTypes<CanFlyControllerApi>()
+                                                         .build();
```
szaszm: Continuation indentation should be 4 spaces deeper. I'm against trying to maintain alignments, because we get right-heavy lines like these, plus renames and similar changes break alignment.

extensions/stable-api-sandbox/ZooProcessor.h
```diff
@@ -0,0 +1,63 @@
...
+
+  EXTENSIONAPI static constexpr auto NumberOfLegsService = core::PropertyDefinitionBuilder<>::createProperty("Number of legs service")
+                                                               .withDescription("Test NumberOfLegsService")
+                                                               .isRequired(true)
+                                                               .withAllowedTypes<NumberOfLegsControllerApi>()
+                                                               .build();
+
```
szaszm, line 45: same here

minifi-api/common/include/minifi-cpp/core/ProvidedControllerServiceInterface.h
```diff
@@ -0,0 +1,43 @@
...
+#include <string_view>
+#include <type_traits>
+
+#include "core/ClassName.h"
+
+namespace org::apache::nifi::minifi::core {
+
+struct ProvidedInterface {
+  std::string_view name;
+  void* (*cast)(void* instance);
```
szaszm, line 31:
```suggestion
  /// fully-qualified class name of the implemented ControllerService interface
  std::string_view name;
  /// casts a pointer of the ControllerService from its dynamic type to a pointer of an implemented interface type. Needed to support cases where they are not the same, like multiple inheritance, virtual inheritance, or contained non-first member implementing the interface
  void* (*cast)(void* instance);
```

extensions/stable-api-testing/tests/ZooTests.cpp
```diff
@@ -0,0 +1,71 @@
...
+    CHECK(controller.getProcessor()->setProperty(ZooProcessor::CanFlyService.name, "duck"));
+    CHECK(controller.getProcessor()->setProperty(ZooProcessor::NumberOfLegsService.name, "dog_with_jetpack"));
+    controller.trigger();
+    CHECK(LogTestController::getInstance().contains("[org::apache::nifi::minifi::api_sandbox::ZooProcessor] [critical] Can duck fly? true"));
+    CHECK(LogTestController::getInstance().contains("[org::apache::nifi::minifi::api_sandbox::ZooProcessor] [critical] dog_with_jetpack has 4 legs"));
+  }
+  {
+    LogTestController::getInstance().clear();
+    CHECK(controller.getProcessor()->setProperty(ZooProcessor::CanFlyService.name, "duck"));
+    CHECK(controller.getProcessor()->setProperty(ZooProcessor::NumberOfLegsService.name, "invalid"));
```
szaszm, line 63: shouldn't this throw an error?

libminifi/test/unit/ComponentManifestTests.cpp
```diff
@@ -38,6 +39,25 @@ SerializedResponseNode& get(SerializedResponseNode& node, const std::string& fie
...
 
+namespace org::apache::nifi::minifi::test::apple {
+class ExampleApacheService : public core::controller::ControllerServiceBase, public core::controller::ControllerServiceHandle {
+ public:
+  using ControllerServiceBase::ControllerServiceBase;
+
+  void initialize() override {}
+  void onEnable() override {}
+
+  [[nodiscard]] ControllerServiceHandle* getControllerServiceHandle() override { return this; }
```
szaszm, line 50: Maybe this should be added to the macro ADD_COMMON_VIRTUAL_FUNCTIONS_FOR_CONTROLLER_SERVICES, tho it doesn't strictly belong in the scope of this change, so you decide

minifi-api/common/include/minifi-cpp/core/ProvidedControllerServiceInterface.h
```diff
@@ -0,0 +1,43 @@
...
+};
+
+template<typename Interface, typename ConcreteService>
+void* castProvidedControllerServiceInterface(void* instance) {
```
szaszm:
```suggestion
void* castProvidedControllerServiceToInterface(void* instance) {
```

minifi-api/common/include/minifi-cpp/core/ProvidedControllerServiceInterface.h
```diff
@@ -0,0 +1,43 @@
...
+#include "core/ClassName.h"
+
+namespace org::apache::nifi::minifi::core {
+
+struct ProvidedInterface {
+  std::string_view name;
+  void* (*cast)(void* instance);
+};
+
+template<typename Interface, typename ConcreteService>
```
szaszm, line 36:
```suggestion
/// casts a pointer of the ControllerService from its dynamic type to a pointer of an implemented interface type. Needed to support cases where they are not the same, like multiple inheritance, virtual inheritance, or contained non-first member implementing the interface
template<typename Interface, typename ConcreteService>
```

minifi-api/common/include/minifi-cpp/core/ProvidedControllerServiceInterface.h
```diff
@@ -0,0 +1,43 @@
...
+void* castProvidedControllerServiceInterface(void* instance) {
+  auto* concrete = static_cast<ConcreteService*>(instance);
+  auto* interface_ptr = static_cast<Interface*>(concrete);
+  return interface_ptr;
+}
+
+template<typename Interface, typename ConcreteService>
+constexpr ProvidedInterface createProvidedInterface() {
+  return ProvidedInterface{className<Interface>(), &castProvidedControllerServiceInterface<Interface, ConcreteService>};
```
szaszm:
```suggestion
  return ProvidedInterface{className<Interface>(), &castProvidedControllerServiceToInterface<Interface, ConcreteService>};
```

libminifi/src/core/state/nodes/AgentInformation.cpp
```diff
@@ -17,46 +17,94 @@
...
+}
+
+bool minifiSystemAllowedType(const std::string_view allowed_type) {
+  constexpr auto system_allowed_types = std::to_array<std::string_view>({
+      controllers::SSLContextServiceInterface::ProvidesApi.type,
+      core::RecordSetWriter::ProvidesApi.type,
+      core::RecordSetReader::ProvidesApi.type,
+      controllers::ProxyConfigurationServiceInterface::ProvidesApi.type,
+      controllers::AttributeProviderService::ProvidesApi.type});
+  return ranges::contains(system_allowed_types, allowed_type);
```
szaszm: I like the old version with inline comparisons better. No need to introduce an STL container and ranges when it doesn't have a benefit, and line patterns are easily readable, even if they are somewhat repetitive.

szaszm: I just noticed that @lordgamez suggested the change, so CC.

lordgamez: It's not really what I suggested, but we discussed that the originally suggested improvement should be a separate PR. Until that is implemented the same issue persists that when any new system allowed type is introduced it needs to be manually inserted here which is not trivial to find and easily missed, does not really matter if it's inlined or extracted to a container. My original intent to alternatively extract the constants was to move them to a more accessible and discoverable place when introducing a new option, but there's not really a viable place for it now, so it can be reverted.

szaszm: ah okay. If it's a temporary solution, I don't insist on changing it, but I would still prefer the simpler version.

extensions/stable-api-testing/tests/ZooTests.cpp
martinzink: no we dont validate here, thats only done on C2 side, we theoratically could so its might be a good feature

extensions/stable-api-sandbox/ZooProcessor.h
martinzink: ive tweaked the clang format and reformatted the file
https://github.com/apache/nifi-minifi-cpp/pull/2176/changes/4fac593e0c7ed8871caab93bb93309974b3a4af1

extensions/stable-api-sandbox/ZooProcessor.h
martinzink: https://github.com/apache/nifi-minifi-cpp/pull/2176/changes/4fac593e0c7ed8871caab93bb93309974b3a4af1

libminifi/src/core/state/nodes/AgentInformation.cpp
```diff
@@ -17,46 +17,93 @@
...
+  if (allowed_type.starts_with(APACHE_GROUP_STR)) {
+    return std::string{APACHE_GROUP_STR};
+  }
+  return std::string{allowed_type.substr(0, allowed_type.find_last_of('.'))};
+}
+
+constexpr std::string_view shortNameFromFullName(std::string_view sv) {
+  auto pos = sv.rfind('.');
+  return (pos == std::string_view::npos) ? sv : sv.substr(pos + 1);
+}
```
szaszm, line 51: you could switch these and call shortNameFromFullName from allowedTypeGroupName. Unless you would rather keep them conceptually separate, regardless of the type compatibility.

extensions/stable-api-testing/AnimalControllerServices.h
```diff
@@ -0,0 +1,67 @@
...
+
+namespace org::apache::nifi::minifi::api_testing {
+class DogController : public api::core::ControllerServiceImpl, public CanFlyControllerApi, public NumberOfLegsControllerApi {
+ public:
+  EXTENSIONAPI static constexpr const char* Description = "Test DogController";
+  EXTENSIONAPI static constexpr auto HasJetpack = core::PropertyDefinitionBuilder<>::createProperty("Has Jetpack")
+                                                      .withDescription("Whether or not the dog has a jetpack")
+                                                      .withDefaultValue("false")
+                                                      .withValidator(core::StandardPropertyValidators::BOOLEAN_VALIDATOR)
+                                                      .build();
```
szaszm:
```suggestion
  EXTENSIONAPI static constexpr auto HasJetpack = core::PropertyDefinitionBuilder<>::createProperty("Has Jetpack")
      .withDescription("Whether or not the dog has a jetpack")
      .withDefaultValue("false")
      .withValidator(core::StandardPropertyValidators::BOOLEAN_VALIDATOR)
      .build();
```

extensions/stable-api-testing/AnimalControllerServices.h
```diff
@@ -0,0 +1,67 @@
...
+#pragma once
+
+#include "AnimalControllerServiceApis.h"
+#include "api/core/ControllerServiceImpl.h"
+#include "api/utils/Export.h"
+#include "core/PropertyDefinitionBuilder.h"
+#include "minifi-cpp/core/ProvidedControllerServiceInterface.h"
+
+namespace org::apache::nifi::minifi::api_testing {
+class DogController : public api::core::ControllerServiceImpl, public CanFlyControllerApi, public NumberOfLegsControllerApi {
```
szaszm, line 28:
```suggestion
/// DogController and DuckController are testing that one ControllerService type can implement multiple interfaces
class DogController : public api::core::ControllerServiceImpl, public CanFlyControllerApi, public NumberOfLegsControllerApi {
```

extensions/stable-api-testing/AnimalControllerServices.h
martinzink: [reformatted entire stable-api-testing dir](https://github.com/apache/nifi-minifi-cpp/pull/2176/commits/215e3f0c938f4e8ff9a9ac7fef53daee7511fffe)

libminifi/src/core/state/nodes/AgentInformation.cpp
martinzink: good idea 👍 [small refactor in AgentInformation.cpp](https://github.com/apache/nifi-minifi-cpp/pull/2176/commits/c20b3341734e19cd6bfe338a2221ec0334d971e3)

szaszm (approved):

libminifi/src/core/state/nodes/AgentInformation.cpp
martinzink: yeah since the temp solution wasnt anyones favorite i've reverted that to the simpler version [Revert "refactor minifiSystemAllowedType"](https://github.com/apache/nifi-minifi-cpp/pull/2176/commits/35d49cab7e09b7cdc82bd5d87bcd7f4cd24646e7)

libminifi/src/core/state/nodes/AgentInformation.cpp
martinzink: it turns out its not equivalent, the shortname gets from foo.bar.baz **baz** and we need **foo.bar** here so [Revert "small refactor in AgentInformation.cpp"](https://github.com/apache/nifi-minifi-cpp/pull/2176/commits/c3771a8a796e98b8d7f3eaa9e7c161b96f5ce958)

PR #2201, opened 2026-06-18
MINIFICPP-2849 Implement LMDB based content repository

lordgamez:
https://issues.apache.org/jira/browse/MINIFICPP-2849

-----------------

### For all changes:

### For code changes:

### Note:


extensions/lmdb/LmdbStream.cpp
```diff
@@ -0,0 +1,116 @@
...
+}
+
+size_t LmdbStream::tell() const {
+  return offset_;
+}
+
+size_t LmdbStream::write(const uint8_t* value, size_t size) {
+  if (!write_enable_) { return STREAM_ERROR; }
+  if (size != 0 && IsNullOrEmpty(value)) { return STREAM_ERROR; }
+  value_.append(reinterpret_cast<const char*>(value), size);
```
lordgamez, line 110: LMDB does not have an append function when writing a value like RocksDB's `Merge` function, so instead of rereading the original value, appending to it, then writing back the new value, all the writes are buffered until the stream is closed, that's when the actual write and commit happens. Currently all content repository streams are used either for write-only or read-only use cases, so there should be no use case where reads and writes are mixed. This should be addressed in a separate PR to change the content repository interface to use separate OutputStream and InputStream types for reads and writes to enforce this, which would also result in separate LmdbInputStream and LmdbOutputStream types (same for RocksDB).

cmake/LMDB.cmake
```diff
@@ -0,0 +1,44 @@
...
+        URL      https://github.com/LMDB/lmdb/archive/refs/tags/LMDB_1.0.0-branch.tar.gz
+        URL_HASH SHA256=8d3e790194e43a72f172f34c442ea4737b2d1433fc0983f2ef70bae999bc2d28
+        PATCH_COMMAND "${PC}"
+        SOURCE_SUBDIR "libraries/liblmdb"
+        SYSTEM
+)
+
+if (WIN32)
+    get_directory_property(MINIFI_SAVED_COMPILE_DEFS COMPILE_DEFINITIONS)
+    remove_definitions(-DWIN32_LEAN_AND_MEAN)
```
szaszm: can you explain why you did this?

lordgamez: LMDB fails to compile on Windows when WIN32_LEAN_AND_MEAN is defined, which is automatically added to the compile definitions in CMakeLists.txt so it is included in all thirdparties used with FetchContent, so it needs to be removed separately for LMDB on Windows.

szaszm: I'd like to make absolutely sure that we only undefine WIN32_LEAN_AND_MEAN while compiling LMDB, or find another workaround. If you look up what it does, it prevents windows.h from including a bunch of additional headers that usually end up unused, and may end up conflicting with other headers, while also increasing compile times. Alternatively, we could look up the transitively included headers that LMDB relies on, and include them explicitly with target_compile_options(target PRIVATE -include foo.h)

Ideally Microsoft would've made the lightweight header the default, with the option to opt in to additional features, but for historical reasons, the heavyweight header is the default, and you can opt out of the extra features. (bloat)

lordgamez: When I checked this according to my understanding this change should only impact the lmdb subdirectory, so it should not impact any other compilation. Anyway I can still check if there is another workaround.

lordgamez: There were 2 issues:
- Defining WIN32_LEAN_AND_MEAN removed `winternl.h` needed for `NTSTATUS` usage, that needed to be added explicitly
- After including `winternl.h` the function pointer names in `mdb.c` clashed with the Windows symbols, those needed to be renamed

Added patch in: https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/52d93b20633060c479678d6ac9d71c21e018b572

thirdparty/lmdb/add-cmake-file.patch
```diff
@@ -0,0 +1,29 @@
+lmdb only comes with a Makefile which only works on Linux so we create our own CMakeLists.txt file to be platform independent
+
+diff -rupN orig/libraries/liblmdb/CMakeLists.txt patched/libraries/liblmdb/CMakeLists.txt
```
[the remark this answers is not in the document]
lordgamez: The line is meant as a comment for the patch and already skipped in the patch command.

thirdparty/lmdb/add-cmake-file.patch
```diff
@@ -0,0 +1,29 @@
...
+--- orig/libraries/liblmdb/CMakeLists.txt	2026-06-12 13:02:00.868592927 +0200
++++ patched/libraries/liblmdb/CMakeLists.txt	2026-06-12 11:51:51.369241534 +0200
+@@ -0,0 +1,23 @@
++cmake_minimum_required(VERSION 3.25)
++project(lmdb C)
++
++if(MSVC)
++    add_compile_options(/W3 /WX-)
++else()
++    add_compile_options(-W -Wall -Wno-unused-parameter -Wbad-function-cast -Wuninitialized -O2 -g)
```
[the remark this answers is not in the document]
lordgamez: Good point, these options were only mirrored from the original Makefile of the project but should not be necessary.

extensions/lmdb/LmdbStream.cpp
```diff
@@ -0,0 +1,116 @@
...
+  dirty_ = true;
```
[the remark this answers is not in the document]
lordgamez: This is already addressed in the comment https://github.com/apache/nifi-minifi-cpp/pull/2201#discussion_r3435648962 and can be open for discussion.

thirdparty/lmdb/add-cmake-file.patch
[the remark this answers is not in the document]
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/92f698e12994676f394a3cfdb0ea53573a490652

extensions/lmdb/LmdbStream.cpp
```diff
@@ -0,0 +1,116 @@
...
+    return false;
+  }
+
+  rc = mdb_txn_commit(txn);
+  if (rc != MDB_SUCCESS) {
+    logger_->log_error("Failed to commit LMDB transaction during close: {}", mdb_strerror(rc));
+    return false;
+  }
+  return true;
+}
```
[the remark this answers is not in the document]
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/92f698e12994676f394a3cfdb0ea53573a490652

extensions/lmdb/LmdbStream.cpp
```diff
@@ -0,0 +1,116 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "LmdbStream.h"
+
+#include <string>
+#include <utility>
+
```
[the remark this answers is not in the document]
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/92f698e12994676f394a3cfdb0ea53573a490652

extensions/lmdb/LmdbContentRepository.cpp
```diff
@@ -0,0 +1,283 @@
...
+
+  if (!max_db_size) {
+    logger_->log_error("Invalid max DB size configuration for LMDB Content Repository");
+    mdb_env_close(lmdb_env_);
+    lmdb_env_ = nullptr;
+    return false;
+  }
+
+  logger_->log_info("Setting LMDB max DB size to {} bytes", *max_db_size);
+  mdb_env_set_mapsize(lmdb_env_, gsl::narrow<size_t>(*max_db_size));
```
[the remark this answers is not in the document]
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/92f698e12994676f394a3cfdb0ea53573a490652

extensions/lmdb/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+file(GLOB SOURCES  "*.cpp")
+
+add_minifi_library(minifi-lmdb SHARED ${SOURCES})
+
+target_include_directories(minifi-lmdb PUBLIC ${LMDB_INCLUDE_DIR})
+target_link_libraries(minifi-lmdb PUBLIC lmdb)
+target_link_libraries(minifi-lmdb PUBLIC minifi-api minifi-extension-framework)
+target_link_libraries(minifi-lmdb PRIVATE $<LINK_ONLY:core-minifi>)
+
+register_extension(minifi-lmdb "LMDB" LMDB "This Enables persistent provenance, flowfile, and content repositories using LMDB" "extensions/lmdb/tests")
```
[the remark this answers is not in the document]
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/92f698e12994676f394a3cfdb0ea53573a490652

extensions/lmdb/tests/LmdbContentSessionTests.cpp
```diff
@@ -0,0 +1,301 @@
...
+  LmdbContentSessionController(const LmdbContentSessionController&) = delete;
+  LmdbContentSessionController& operator=(LmdbContentSessionController&&) = delete;
+  LmdbContentSessionController& operator=(const LmdbContentSessionController&) = delete;
+
+  std::shared_ptr<core::ContentRepository> content_repository_;
+};
+
+namespace {
+
+std::shared_ptr<io::OutputStream> operator<<(std::shared_ptr<io::OutputStream> stream, const std::string& str) {
```
fgerlits: nitpick, but `stream` should be `const &` (also in `>>`)

extensions/lmdb/tests/LmdbStreamTests.cpp
```diff
@@ -0,0 +1,273 @@
...
+    if (const int rc = mdb_env_create(&lmdb_env_)) {
+      throw std::runtime_error("Failed to create LMDB environment: " + std::string(mdb_strerror(rc)));
+    }
+    if (const int rc = mdb_env_set_mapsize(lmdb_env_, 100ULL * 1024 * 1024); rc != MDB_SUCCESS) {
+      throw std::runtime_error("Failed to set LMDB map size: " + std::string(mdb_strerror(rc)));
+    }
+
+    if (const int rc = mdb_env_open(lmdb_env_, db_path_.c_str(), MDB_NOTLS, 0664)) {
+      throw std::runtime_error("Failed to open LMDB environment " + db_path_ + ": " + std::string(mdb_strerror(rc)));
+    }
```
fgerlits: I would add `; rc != MDB_SUCCESS` in lines 29 and 36, too

extensions/lmdb/LmdbContentRepository.cpp
```diff
@@ -0,0 +1,288 @@
...
+
+  const auto writeResource = [&lmdb_content_repository](const std::shared_ptr<ResourceClaim>& resource_claim, const std::shared_ptr<io::BaseStream>& stream, bool is_append) {
+    auto outStream = lmdb_content_repository->write(*resource_claim, is_append);
+    if (outStream == nullptr) { throw Exception(REPOSITORY_EXCEPTION, "Couldn't open the underlying resource for write: " + resource_claim->getContentFullPath()); }
+    const auto size = stream->size();
+    if (outStream->write(stream->getBuffer()) != size) {
+      throw Exception(REPOSITORY_EXCEPTION, "Failed to write " + std::string(is_append ? "appended" : "new") + " resource: " + resource_claim->getContentFullPath());
+    }
+    auto lmdb_out_stream = std::dynamic_pointer_cast<io::LmdbStream>(outStream);
+    if (lmdb_out_stream == nullptr) { throw Exception(REPOSITORY_EXCEPTION, "Couldn't cast output stream to LmdbStream for commit: " + resource_claim->getContentFullPath()); }
```
fgerlits: should we check this before the `write()`?

extensions/lmdb/LmdbContentRepository.cpp
```diff
@@ -0,0 +1,288 @@
...
+    if (!lmdb_out_stream->commit()) { throw Exception(REPOSITORY_EXCEPTION, "Failed to commit " + std::string(is_append ? "appended" : "new") + " resource: " + resource_claim->getContentFullPath()); }
+  };
```
fgerlits, line 62: Very minor nitpick, but these `throw Exception...` lines look too long to me for a single-line format. I would put all of them on separate lines as you did at lines 46-48.

extensions/lmdb/LmdbContentRepository.cpp
```diff
@@ -0,0 +1,288 @@
...
+  for (const auto& resource : append_state_) {
+    writeResource(resource.first, resource.second.stream, true);
+  }
+
+  managed_resources_.clear();
+  append_state_.clear();
+}
+
+bool LmdbContentRepository::initialize(const std::shared_ptr<minifi::Configure>& configuration) {
+  if (const int rc = mdb_env_create(&lmdb_env_)) {
```
fgerlits: I would add the explicit comparison here:
```suggestion
  if (const int rc = mdb_env_create(&lmdb_env_); rc != MDB_SUCCESS) {
```

extensions/lmdb/LmdbContentRepository.cpp
```diff
@@ -0,0 +1,288 @@
...
+    // Default to 10 GB if the property is not set
+    return std::make_optional<uint64_t>(10ULL * 1024 * 1024 * 1024);
+  });
+
+  if (!max_db_size) {
+    logger_->log_error("Invalid max DB size configuration for LMDB Content Repository");
+    mdb_env_close(lmdb_env_);
+    lmdb_env_ = nullptr;
+    return false;
+  }
```
fgerlits, line 96: I don't think `max_db_size` can be `nullopt` here: if the property was unset or empty, then we are using the default value, and if it is an invalid number, then we throw earlier. Maybe we should set `max_db_size` to `nullopt` instead of throwing?

extensions/lmdb/LmdbContentRepository.cpp
```diff
@@ -0,0 +1,288 @@
...
+    auto cursor_guard = gsl::finally([cursor] { mdb_cursor_close(cursor); });
+
+    MDB_val key{};
+    MDB_val val{};
+    int rc = mdb_cursor_get(cursor, &key, &val, MDB_FIRST);
+
+    while (rc == MDB_SUCCESS) {
+      std::string key_string = std::string(static_cast<char*>(key.mv_data), key.mv_size);
+
+      std::lock_guard<std::mutex> lock(count_map_mutex_);
```
fgerlits, line 243: Should we lock the `count_map_mutex_` for the whole duration of the iteration? Allowing modifications to `count_map_` while we are checking the orphans makes it difficult to reason about what is happening.

extensions/lmdb/LmdbStream.h
```diff
@@ -0,0 +1,66 @@
...
+
+#include "core/logging/LoggerFactory.h"
+#include "io/BaseStream.h"
+#include "lmdb.h"
+
+namespace org::apache::nifi::minifi::io {
+
+class LmdbStream : public io::BaseStreamImpl {
+ public:
+  explicit LmdbStream(std::string path, MDB_env* lmdb_env, MDB_dbi* lmdb_handle, bool write_enable = false);
```
fgerlits: I can't see any place where we use the default, and it's better not to have it, so I would remove it:
```suggestion
  explicit LmdbStream(std::string path, MDB_env* lmdb_env, MDB_dbi* lmdb_handle, bool write_enable);
```

extensions/lmdb/LmdbStream.cpp
```diff
@@ -0,0 +1,119 @@
...
+  rc = mdb_txn_commit(txn);
+  if (rc != MDB_SUCCESS) {
+    logger_->log_error("Failed to commit LMDB transaction during close: {}", mdb_strerror(rc));
+    return false;
+  }
+
+  dirty_ = false;
+  return true;
+}
```
fgerlits, line 97: nitpick: the error logs should say "commit" instead of "close"

extensions/lmdb/LmdbStream.cpp
```diff
@@ -0,0 +1,119 @@
...
+  offset_ = offset;
+}
+
+size_t LmdbStream::tell() const {
+  return offset_;
+}
+
+size_t LmdbStream::write(const uint8_t* value, size_t size) {
+  if (!write_enable_) { return STREAM_ERROR; }
+  if (size != 0 && IsNullOrEmpty(value)) { return STREAM_ERROR; }
```
fgerlits: `value` is just a pointer, so `IsNullOrEmpty` is just a null check; I would change this line to
```suggestion
  if (size != 0 && value == nullptr) { return STREAM_ERROR; }
```
to make this clearer

extensions/lmdb/LmdbContentRepository.h
```diff
@@ -0,0 +1,87 @@
...
+#include "lmdb.h"
+
+namespace org::apache::nifi::minifi::core::repository {
+
+class LmdbContentRepository : public core::ContentRepositoryImpl {
+ public:
+  explicit LmdbContentRepository(std::string_view name = className<LmdbContentRepository>(), const utils::Identifier& uuid = {})
+      : core::ContentRepositoryImpl(name, uuid) {}
+
+  ~LmdbContentRepository() override {
```
szaszm: The class should follow the rule of 5. Disable copy and move?
Destructors should be noexcept when possible.
```suggestion
  LmdbContentRepository(const LmdbContentRepository&) = delete;
  LmdbContentRepository(LmdbContentRepository&&) = delete;
  LmdbContentRepository& operator=(const LmdbContentRepository&) = delete;
  LmdbContentRepository& operator=(LmdbContentRepository&&) = delete;

  ~LmdbContentRepository() noexcept override {
```

extensions/lmdb/LmdbContentRepository.h
```diff
@@ -0,0 +1,87 @@
...
+    stop();
```
szaszm: Don't call virtual functions of the same class from constructors and destructors. Qualify the function name to clarify what gets called.
```suggestion
    LmdbContentRepository::stop();
```

extensions/lmdb/LmdbContentRepository.h
```diff
@@ -0,0 +1,87 @@
...
+    if (lmdb_env_) {
+      mdb_dbi_close(lmdb_env_, lmdb_handle_);
+      mdb_env_close(lmdb_env_);
+    }
+  }
```
szaszm, line 51: In the name of the single responsibility principle, I would extract the lifetime management of LMDB handles to a separate type. LmdbContentRepository should have one responsibility: implement a minifi c++ content repository in terms of LMDB. We can use helper types and functions to extract individual steps, and have this class only have the minimum necessary complexity.

extensions/lmdb/LmdbContentRepository.h
```diff
@@ -0,0 +1,87 @@
...
+  class Session : public BufferedContentSession {
+   public:
+    explicit Session(std::shared_ptr<ContentRepository> repository);
+
+    void commit() override;
+  };
+
+  std::shared_ptr<ContentSession> createSession() override;
+  bool initialize(const std::shared_ptr<minifi::Configure>& configuration) override;
+  std::shared_ptr<io::BaseStream> write(const minifi::ResourceClaim& claim, bool append = false) override;
```
szaszm: Don't use default arguments on virtual functions

extensions/lmdb/LmdbContentRepository.cpp
```diff
@@ -0,0 +1,288 @@
...
+#include "minifi-cpp/utils/gsl.h"
+#include "utils/Locations.h"
+
+namespace org::apache::nifi::minifi::core::repository {
+
+LmdbContentRepository::Session::Session(std::shared_ptr<ContentRepository> repository) : BufferedContentSession(std::move(repository)) {}
+
+void LmdbContentRepository::Session::commit() {
+  auto lmdb_content_repository = std::dynamic_pointer_cast<LmdbContentRepository>(repository_);
+  if (!lmdb_content_repository) { throw Exception(REPOSITORY_EXCEPTION, "Session's repository is not an LmdbContentRepository"); }
```
szaszm: If we require LmdbContentRepository, we should say so in the constructor signature
```suggestion
LmdbContentRepository::Session::Session(std::shared_ptr<LmdbContentRepository> repository) : BufferedContentSession(std::move(repository)) {}

void LmdbContentRepository::Session::commit() {
  auto lmdb_content_repository = std::dynamic_pointer_cast<LmdbContentRepository>(repository_);
  if (!lmdb_content_repository) { throw Exception(REPOSITORY_EXCEPTION, "Session's repository is not an LmdbContentRepository"); }
```

extensions/lmdb/LmdbContentRepository.h
```diff
@@ -0,0 +1,87 @@
...
+    }
+  }
+
+  EXTENSIONAPI static constexpr auto Properties = std::array<core::PropertyReference, 0>{};
+  EXTENSIONAPI static constexpr bool SupportsDynamicProperties = false;
+  EXTENSIONAPI static constexpr bool SupportsDynamicRelationships = false;
+
+  class Session : public BufferedContentSession {
+   public:
+    explicit Session(std::shared_ptr<ContentRepository> repository);
```
szaszm:
```suggestion
    explicit Session(std::shared_ptr<LmdbContentRepository> repository);
```

extensions/lmdb/tests/LmdbContentSessionTests.cpp
```diff
@@ -0,0 +1,301 @@
...
+  REQUIRE(content == "committed");
+}
+
+TEST_CASE("LmdbContentRepository::Session commit throws when underlying repository is not LmdbContentRepository", "[lmdb]") {
+  auto unrelated_repository = std::make_shared<core::repository::VolatileContentRepository>();
+  unrelated_repository->initialize(std::make_shared<ConfigureImpl>());
+
+  core::repository::LmdbContentRepository::Session session(unrelated_repository);
+  REQUIRE_THROWS_WITH(session.commit(), Catch::Matchers::ContainsSubstring("Session's repository is not an LmdbContentRepository"));
+}
```
szaszm: If Session is changed to only accept LmdbContentRepository, then this will no longer compile, therefore no runtime testing is necessary or possible.
```suggestion
```

extensions/lmdb/LmdbContentRepository.cpp
```diff
@@ -0,0 +1,288 @@
...
+  }
+
+  return true;
+}
+
+void LmdbContentRepository::start() {}
+void LmdbContentRepository::stop() {}
+
+std::shared_ptr<ContentSession> LmdbContentRepository::createSession() {
+  return std::make_shared<Session>(sharedFromThis<ContentRepository>());
```
szaszm:
```suggestion
  return std::make_shared<Session>(sharedFromThis<LmdbContentRepository>());
```

extensions/lmdb/LmdbContentRepository.cpp
```diff
@@ -0,0 +1,288 @@
...
+    return false;
+  }
+
+  // Reserve virtual address space for the DB file (max size it can grow to)
+  const auto max_db_size = configuration->get(Configure::nifi_content_repository_lmdb_max_db_size) | utils::andThen([](auto max_db_size_str) -> std::optional<uint64_t> {
+    if (max_db_size_str.empty()) { return std::nullopt; }
+    return parsing::parseDataSize(max_db_size_str) | utils::orThrow(fmt::format("{} was set to invalid value: '{}'", Configure::nifi_content_repository_lmdb_max_db_size, max_db_size_str));
+  }) | utils::orElse([] {
+    // Default to 10 GB if the property is not set
+    return std::make_optional<uint64_t>(10ULL * 1024 * 1024 * 1024);
```
szaszm: from Literals.h
```suggestion
    return std::make_optional<uint64_t>(10_GiB);
```

extensions/lmdb/tests/LmdbContentSessionTests.cpp
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/f0f3414be0d37ff0f9ecadc88893e83fda6620bc

extensions/lmdb/tests/LmdbStreamTests.cpp
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/f0f3414be0d37ff0f9ecadc88893e83fda6620bc

extensions/lmdb/LmdbContentRepository.cpp
lordgamez: Good idea, moved in https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/f0f3414be0d37ff0f9ecadc88893e83fda6620bc

extensions/lmdb/LmdbContentRepository.cpp
lordgamez: Good point, changed in https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/f0f3414be0d37ff0f9ecadc88893e83fda6620bc

extensions/lmdb/LmdbContentRepository.cpp
lordgamez: Added in https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/f0f3414be0d37ff0f9ecadc88893e83fda6620bc

extensions/lmdb/LmdbContentRepository.cpp
lordgamez: Makes sense, changed to return nullopt instead of throwing, it should not make a difference in the usage https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/f0f3414be0d37ff0f9ecadc88893e83fda6620bc

extensions/lmdb/LmdbContentRepository.cpp
lordgamez: Makes sense, changed in https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/f0f3414be0d37ff0f9ecadc88893e83fda6620bc

extensions/lmdb/LmdbStream.h
lordgamez: Removed in https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/f0f3414be0d37ff0f9ecadc88893e83fda6620bc

extensions/lmdb/LmdbStream.cpp
lordgamez: Fixed in https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/f0f3414be0d37ff0f9ecadc88893e83fda6620bc

extensions/lmdb/LmdbStream.cpp
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/f0f3414be0d37ff0f9ecadc88893e83fda6620bc

extensions/lmdb/LmdbContentRepository.h
lordgamez: Good point, updated in https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/f0f3414be0d37ff0f9ecadc88893e83fda6620bc

extensions/lmdb/LmdbContentRepository.h
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/f0f3414be0d37ff0f9ecadc88893e83fda6620bc

extensions/lmdb/LmdbContentRepository.h
lordgamez: Agreed, this was extracted in the `LmdbWrapper` class in https://github.com/apache/nifi-minifi-cpp/pull/2203 As it would be a bit complicated changing both PRs to move that here and we will probably handle and merge both PRs close together or at least in the same release I would keep that change in the follow up PR for now.

extensions/lmdb/LmdbContentRepository.h
lordgamez: Good point, updated to use explicit parameter for all repository writes removing default value in https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/f0f3414be0d37ff0f9ecadc88893e83fda6620bc

extensions/lmdb/LmdbContentRepository.cpp
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/f0f3414be0d37ff0f9ecadc88893e83fda6620bc

extensions/lmdb/LmdbContentRepository.h
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/f0f3414be0d37ff0f9ecadc88893e83fda6620bc

extensions/lmdb/tests/LmdbContentSessionTests.cpp
lordgamez: Removed in https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/f0f3414be0d37ff0f9ecadc88893e83fda6620bc

extensions/lmdb/LmdbContentRepository.cpp
lordgamez: Updated in https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/f0f3414be0d37ff0f9ecadc88893e83fda6620bc

extensions/lmdb/LmdbContentRepository.cpp
lordgamez: Good idea, updated in https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/f0f3414be0d37ff0f9ecadc88893e83fda6620bc

fgerlits (approved):

cmake/MiNiFiOptions.cmake
```diff
@@ -90,6 +90,7 @@ endif()
 add_minifi_option(ENABLE_ALL "Enables all extensions" OFF)
 add_minifi_option(ENABLE_CIVET "Enables CivetWeb components." ON)
 add_minifi_option(ENABLE_ROCKSDB "Enables the RocksDB extension." ON)
+add_minifi_option(ENABLE_LMDB "Enables the LMDB extension." ON)
```
martinzink: Could we name this MINIFI_LMBD instead? or anything else but prefixed with MINIFI?

martinzink (approved):
LGTM, only nitpick is the MinifiOption naming

cmake/MiNiFiOptions.cmake
lordgamez: Renamed it in https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/aa0f2616843ca4786f30a960a4c84a8b6eb2aa7e
The only problem was that the bootstrap script checked for the `ENABLE_` prefix to choose the extension options. I tweaked it a bit to move `MINIFI_LMDB` and `MINIFI_RUST` to the extension options and `AWS_ENABLE_UNITY_BUILD` to the build options.

martinzink: (not in this PR) but maybe we could do MINIFI_BUILD_ and MINIFI_EXT_ prefixes?

lordgamez: Yes, that sounds good

szaszm (changes requested):
I'm not sure LMDB-based repositories are a good direction long term. In theory it works better than RocksDB in every aspect for our use case, but I want to evaluate other repository implementation options before releasing this and committing to supporting it.

We can merge this later to keep the code in sync and avoid merge conflicts until we finish the evaluation of other options, but please mark it as experimental in the docs, and make it disabled by default.

lordgamez:
> I'm not sure LMDB-based repositories are a good direction long term. In theory it works better than RocksDB in every aspect for our use case, but I want to evaluate other repository implementation options before releasing this and committing to supporting it.
> 
> We can merge this later to keep the code in sync and avoid merge conflicts until we finish the evaluation of other options, but please mark it as experimental in the docs, and make it disabled by default.

Updated in https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/b045aa6b258580c1bc7c97fb8e736f5f7317d3eb

thirdparty/lmdb/all/patches/fix-windows-symbols.patch
szaszm, line 1: why is this necessary or even working?

extensions/lmdb/LmdbContentRepository.cpp
```diff
@@ -0,0 +1,299 @@
...
+}
+
+void LmdbContentRepository::start() {}
+void LmdbContentRepository::stop() {}
+
+std::shared_ptr<ContentSession> LmdbContentRepository::createSession() {
+  return std::make_shared<Session>(sharedFromThis<LmdbContentRepository>());
+}
+
+std::shared_ptr<io::BaseStream> LmdbContentRepository::write(const minifi::ResourceClaim& claim, bool) {
```
szaszm: AI review
1. The append parameter to write() is silently ignored

LmdbStream always loads the existing value for the key (loadValue()), and subsequent write() calls always append new bytes to value_ regardless of the caller's intent. This happens to work today because BufferedContentSession guarantees a brand-new (empty) key for every write()/overwrite path and only sets is_append=true when genuine appending is intended — so the two code paths currently converge on the same correct behavior. However, this is fragile:

    If a resource-claim path is ever reused while a stale LMDB entry exists (e.g., a failed removeKey() that ended up on purge_list_ and hasn't been retried yet, followed by path reuse), a "new" write would silently concatenate onto leftover bytes instead of overwriting — silent data corruption rather than a clear failure.
    It also means the append flag on the interface is effectively decorative for this backend, which is misleading to future maintainers and out of step with the RocksDB/FileSystem implementations that do act on it.

Suggest either truncating/deleting the existing value explicitly when append == false (defensive), or at least asserting/logging if a "new" write ever encounters a pre-existing key.

extensions/lmdb/LmdbContentRepository.cpp
```diff
@@ -0,0 +1,299 @@
...
+    logger_->log_error("Failed to read LMDB database stats: {}", mdb_strerror(rc));
+  }
+  mdb_txn_abort(txn);
+  return stat;
+}
+
+uint64_t LmdbContentRepository::getRepositorySize() const {
+  const auto stat = getDbStat();
+  return stat.ms_psize * (stat.ms_branch_pages + stat.ms_leaf_pages + stat.ms_overflow_pages);
+}
```
szaszm, line 291: AI review
2. getRepositorySize() likely under-reports actual footprint
C++

uint64_t LmdbContentRepository::getRepositorySize() const {
  const auto stat = getDbStat();
  return stat.ms_psize * (stat.ms_branch_pages + stat.ms_leaf_pages + stat.ms_overflow_pages);
}

This reflects only live/used B-tree pages, not the actual on-disk/mapped file size. Given the documentation explicitly calls out that "the database file never shrinks" as a caveat, users querying repository size after deletions will see a number that doesn't match the real file footprint they were warned about. Consider deriving this from the mapped file size (e.g., via mdb_env_info/mdb_env_stat combined with actual file size) if the intent is to reflect disk usage.

Human comment
Consider adding a new metric for lmdb repositories only for live vs full size.

extensions/lmdb/LmdbContentRepository.cpp
szaszm: I'm only sharing sloppypasta because I think this could be a genuine concern, and the suggested solution, disabling / asserting on the unsupported path, would be an improvement.

thirdparty/lmdb/all/patches/fix-windows-symbols.patch
lordgamez: The `winternl.h` include was needed with newer MSVC version for the NTSTATUS symbol, but that caused a conflict of the existing Windows types with the custom `Nt*` typedefs in this file so they needed to be renamed.

extensions/lmdb/LmdbContentRepository.cpp
lordgamez: I changed it to use the filesystem size for now on in https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/3512381965980998b36e06aa9838ab88270726a2 as non of the statistics given by LMDB really reflect the actual disk footprint that is the most useful for us. I would keep that for now and we can think of implementing other useful stats later after a repository interface refactor. Currently only `getRocksDbStats` is available on the interface, so that should be changed to a generic custom repository stats to be able to implement repository specific statistics.

extensions/lmdb/LmdbContentRepository.cpp
lordgamez: I think here it was just using the same design as RocksDB where when creating the RocksDBStream was also ignoring the append argument on write. I think here we can handle the case and add a warning when we overwrite an existing value as it is unusual, but can happen in the aforementioned erroneous case. I'm still on the edge of either it should be a warning or an assertion, as if that happens we can still proceed, or could be an intentional use case of the LmdbStream so I left it as a warning for now. Updated in https://github.com/apache/nifi-minifi-cpp/pull/2201/commits/3512381965980998b36e06aa9838ab88270726a2

szaszm (approved):
I still don't think LMDB will be the right repository replacement, but I approve merging this, to avoid conflicts, while we evaluate all the options and settle on a specific new repository implementation.

PR #2217, opened 2026-07-24
MINIFICPP-2787 Do not encrypt empty properties and parameters

fgerlits:
This also fixes a bug: if a sensitive value was set as
```
  sensitive_property:
```
instead of
```
  sensitive_property: ""
```
in a YAML flow configuration, then the value was null instead of an empty string, and it got expanded to the string "null", and then encrypted.

https://issues.apache.org/jira/browse/MINIFICPP-2787

---

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/libminifi/test/libtest/unit/Catch.h
+++ b/libminifi/test/libtest/unit/Catch.h
@@ -17,9 +17,11 @@
 
 #pragma once
 
+#include <chrono>
+#include <expected>
 #include <optional>
 #include <string>
-#include <chrono>
+
 #include "fmt/format.h"
 #include "catch2/catch_test_macros.hpp"
 #include "catch2/catch_tostring.hpp"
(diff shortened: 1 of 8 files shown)


libminifi/test/libtest/unit/Catch.h
```diff
@@ -56,6 +58,19 @@ struct StringMaker<std::chrono::file_clock::duration> {
...
+template <typename T, typename E>
+struct StringMaker<std::expected<T, E>> {
+  static std::string convert(const std::expected<T, E>& expected) {
+    if (!expected) {
+      return "error: " + Catch::Detail::stringify(expected.error());
+    }
+    if constexpr (!std::is_void_v<T>) {
+      return Catch::Detail::stringify(*expected);
+    }
+    return "OK";
```
szaszm: I would represent expected alternatives like this. What do you think?
```suggestion
    if (expected.has_value() && std::is_void_v<T>) {
      return "expected(void)";
    } else if (expected.has_value()) {
      return fmt::format("expected({})", Catch::Detail::stringify(expected.value()));
    } else {
      return fmt::format("unexpected({})", Catch::Detail::stringify(expected.error()));
    }
```

fgerlits: I have updated the debug strings to what you suggested, in 8eef269f100ab235defbfd2ea99bb3de06ab0a4b. I have kept the structure as it was, because your version doesn't compile.

lordgamez (approved):

martinzink (approved):

adamdebreceni (approved):

szaszm (approved):

PR #2220, opened 2026-07-24
MINIFICPP-2869 Improve rust api property validation and property getters

martinzink:

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/minifi_rust/extensions/minifi_rs_playground/features/controller_apis.feature
+++ b/minifi_rust/extensions/minifi_rs_playground/features/controller_apis.feature
@@ -23,6 +23,6 @@ Feature: Testing controller service api casting
     And the "Extra information" property of the Wolfie the magical controller service is set to "The dog (Canis familiaris or Canis lupus familiaris) is a domesticated descendant of wolves."
     When the MiNiFi instance starts up
 
-    Then the Minifi logs contain the following message: "[minifi_rs_playground::processors::zoo_processor::ZooProcessorRs] [critical] Can DogControllerRs { has_jetpack: true, extra_info: Some("The dog (Canis familiaris or Canis lupus familiaris) is a domesticated descendant of wolves.") } fly?" in less than 10 seconds
+    Then the Minifi logs contain the following message: "[minifi_rs_playground::processors::zoo_processor::ZooProcessorRs] [critical] Can DogControllerRs { has_jetpack: true, extra_info: Some("The dog (Canis familiaris or Canis lupus familiaris) is a domesticated descendant of wolves.") } fly? true" in less than 10 seconds
     And the Minifi logs do not contain errors
     And the Minifi logs do not contain warnings
--- a/minifi_rust/extensions/minifi_rs_playground/features/controller_apis.feature
+++ b/minifi_rust/extensions/minifi_rs_playground/features/controller_apis.feature
@@ -23,6 +23,6 @@ Feature: Testing controller service api casting
     And the "Extra information" property of the Wolfie the magical controller service is set to "The dog (Canis familiaris or Canis lupus familiaris) is a domesticated descendant of wolves."
     When the MiNiFi instance starts up
 
-    Then the Minifi logs contain the following message: "[minifi_rs_playground::processors::zoo_processor::ZooProcessorRs] [critical] Can DogControllerRs { has_jetpack: true, extra_info: "The dog (Canis familiaris or Canis lupus familiaris) is a domesticated descendant of wolves." } fly? true" in less than 10 seconds
+    Then the Minifi logs contain the following message: "[minifi_rs_playground::processors::zoo_processor::ZooProcessorRs] [critical] Can DogControllerRs { has_jetpack: true, extra_info: Some("The dog (Canis familiaris or Canis lupus familiaris) is a domesticated descendant of wolves.") } fly?" in less than 10 seconds
     And the Minifi logs do not contain errors
     And the Minifi logs do not contain warnings
--- a/minifi_rust/extensions/minifi_rs_playground/src/controller_services/dog_controller_service.rs
+++ b/minifi_rust/extensions/minifi_rs_playground/src/controller_services/dog_controller_service.rs
@@ -18,11 +18,11 @@
 use crate::controller_services::animal_controller_apis::{
     CanFlyControllerApi, NumberOfLegsControllerApi,
 };
-use minifi_native::ControllerServiceApi;
 use minifi_native::macros::ComponentIdentifier;
+use minifi_native::{ControllerServiceApi, property_constraint};
 use minifi_native::{
     ControllerServiceDefinition, EnableControllerService, GetProperty, Logger, MinifiError,
-    Property, ProvidedInterface, StandardPropertyValidator, create_provided_interface,
+    Property, ProvidedInterface, create_provided_interface,
 };
 
 pub(crate) const HAS_JETPACK: Property = Property {
@@ -32,9 +32,7 @@ pub(crate) const HAS_JETPACK: Property = Property {
     is_sensitive: false,
     supports_expr_lang: false,
     default_value: Some("false"),
-    validator: StandardPropertyValidator::BoolValidator,
-    allowed_values: &[],
-    allowed_type: None,
+    constraints: property_constraint::<bool>(),
 };
 
 pub(crate) const EXTRA_INFO: Property = Property {
@@ -44,9 +42,7 @@ pub(crate) const EXTRA_INFO: Property = Property {
     is_sensitive: false,
     supports_expr_lang: false,
     default_value: None,
-    validator: StandardPropertyValidator::AlwaysValidValidator,
-    allowed_values: &[],
-    allowed_type: None,
+    constraints: None,
 };
 
 #[allow(dead_code)] // extra_info is only used by {:?}
@@ -73,11 +69,10 @@ impl EnableControllerService for DogControllerRs {
     where
         Self: Sized,
     {
-        let has_jetpack = context.get_bool_property(&HAS_JETPACK)?.ok_or(
-            MinifiError::missing_required_property("Has jetpack is required"),
-        )?;
-
-        let extra_info = context.get_property(&EXTRA_INFO)?.unwrap_or("".into());
+        let has_jetpack = context.get_req_property::<bool>(&HAS_JETPACK)?;
+        let extra_info = context
+            .get_property::<String>(&EXTRA_INFO)?
+            .unwrap_or("".into());
 
         Ok(Self {
             has_jetpack,
--- a/minifi_rust/extensions/minifi_rs_playground/src/controller_services/dog_controller_service.rs
+++ b/minifi_rust/extensions/minifi_rs_playground/src/controller_services/dog_controller_service.rs
@@ -19,37 +19,25 @@ use crate::controller_services::animal_controller_apis::{
     CanFlyControllerApi, NumberOfLegsControllerApi,
 };
 use minifi_native::macros::ComponentIdentifier;
-use minifi_native::{ControllerServiceApi, property_constraint};
 use minifi_native::{
-    ControllerServiceDefinition, EnableControllerService, GetProperty, Logger, MinifiError,
-    Property, ProvidedInterface, create_provided_interface,
+    ControllerServiceApi, ControllerServiceDefinition, EnableControllerService, GetProperty,
+    Logger, MinifiError, Property, PropertyDefinition, ProvidedInterface,
+    create_provided_interface, property_definitions,
 };
 
-pub(crate) const HAS_JETPACK: Property = Property {
-    name: "Has Jetpack",
-    description: "Whether or not the dog has a jetpack",
-    is_required: true,
-    is_sensitive: false,
-    supports_expr_lang: false,
-    default_value: Some("false"),
-    constraints: property_constraint::<bool>(),
-};
+pub(crate) const HAS_JETPACK: Property<bool> =
+    Property::new("Has Jetpack", "Whether or not the dog has a jetpack").with_default("false");
 
-pub(crate) const EXTRA_INFO: Property = Property {
-    name: "Extra information",
-    description: "We need this to verify the casting was done correctly",
-    is_required: false,
-    is_sensitive: false,
-    supports_expr_lang: false,
-    default_value: None,
-    constraints: None,
-};
+pub(crate) const EXTRA_INFO: Property<Option<String>> = Property::new(
+    "Extra information",
+    "We need this to verify the casting was done correctly",
+);
 
 #[allow(dead_code)] // extra_info is only used by {:?}
 #[derive(Debug, ComponentIdentifier)]
 pub(crate) struct DogControllerRs {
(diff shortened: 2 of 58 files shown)


minifi_rust/minifi_native/src/api/property.rs
```diff
@@ -44,69 +58,100 @@ pub struct Property {
...
 
-    fn get_size_property(&self, property: &Property) -> Result<Option<u64>, MinifiError> {
-        if property.validator != DataSizeValidator {
+    fn get_property<T: PropertyType>(
+        &self,
+        property: &Property,
+    ) -> Result<Option<T::Output>, MinifiError> {
+        if T::EXPECTED_CONSTRAINTS != NoConstraints
+            && T::EXPECTED_CONSTRAINTS != property.constraints
+        {
```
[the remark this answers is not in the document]
martinzink: it doesnt move out and it does compile, the property is already reference

minifi_rust/minifi_native/src/api/errors.rs
```diff
@@ -89,6 +92,18 @@ impl From<NulError> for MinifiError {
...
+
+impl From<std::convert::Infallible> for MinifiError {
+    fn from(_: std::convert::Infallible) -> Self {
+        unreachable!("Infallible errors can never happen")
+    }
+}
+
 impl MinifiError {
     pub(crate) fn to_status(&self) -> minifi_status {
         match self {
```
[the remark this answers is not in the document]
martinzink: missing required attribute is not validation error

minifi_rust/extensions/minifi_rs_playground/src/controller_services/dog_controller_service.rs
```diff
@@ -32,9 +32,7 @@ pub(crate) const HAS_JETPACK: Property = Property {
     is_sensitive: false,
     supports_expr_lang: false,
     default_value: Some("false"),
-    validator: StandardPropertyValidator::BoolValidator,
-    allowed_values: &[],
-    allowed_type: None,
+    constraints: property_constraint::<bool>(),
```
szaszm: After our private discussion, I'll wait to see if you can refactor this to move the type to a Property<T> type parameter, but the current version is an improvement too.

minifi_rust/extensions/minifi_rs_playground/features/controller_apis.feature
```diff
@@ -23,6 +23,6 @@ Feature: Testing controller service api casting
     And the "Extra information" property of the Wolfie the magical controller service is set to "The dog (Canis familiaris or Canis lupus familiaris) is a domesticated descendant of wolves."
     When the MiNiFi instance starts up
 
-    Then the Minifi logs contain the following message: "[minifi_rs_playground::processors::zoo_processor::ZooProcessorRs] [critical] Can DogControllerRs { has_jetpack: true, extra_info: "The dog (Canis familiaris or Canis lupus familiaris) is a domesticated descendant of wolves." } fly? true" in less than 10 seconds
+    Then the Minifi logs contain the following message: "[minifi_rs_playground::processors::zoo_processor::ZooProcessorRs] [critical] Can DogControllerRs { has_jetpack: true, extra_info: Some("The dog (Canis familiaris or Canis lupus familiaris) is a domesticated descendant of wolves.") } fly?" in less than 10 seconds
```
fgerlits: the ` true` at the end should not be removed, I think

minifi_rust/extensions/minifi_rs_playground/src/processors/generate_flow_file/properties.rs
```diff
@@ -15,64 +15,34 @@
...
-    default_value: Some("1"),
-    validator: StandardPropertyValidator::U64Validator,
-    allowed_values: &[],
-    allowed_type: None,
-};
+pub(crate) const BATCH_SIZE: Property<u64> = Property::new(
+    "Batch Size",
+    "The number of FlowFiles to be transferred in each invocation",
+)
+.with_default("1");
```
szaszm, line 30: shouldn't this be indented 2 levels as a continuation?

minifi_rust/minifi_native/src/api/property.rs
```diff
@@ -16,102 +16,291 @@
...
+
+    pub const fn name(&self) -> &'static str {
+        self.name
+    }
+
+    pub const fn definition(&self) -> PropertyDefinition {
+        PropertyDefinition {
+            name: self.name,
+            description: self.description,
+            is_required: K::IS_REQUIRED || self.default_value.is_some(),
```
fgerlits: I know the existing C++ code is inconsistent, but if the property has a default value, then I would say it should *not* be required, meaning that Minifi should accept flow definitions in which the property is not specified (and return the default value from `get_property`).
```suggestion
            is_required: K::IS_REQUIRED && self.default_value.is_none(),
```

Also, if `IS_REQUIRED` is not always the same as `is_required`, they should have different names. I would negate `IS_REQUIRED` and rename it to `IS_OPTIONAL`.

szaszm: I'd say don't try to be smart here, and wire it through directly. That way if the C++ implementation changes behavior around edge cases (like #2241), the Rust implementation will automatically work right too.

minifi_rust/minifi_native/src/c_ffi/c_ffi_process_context.rs
```diff
@@ -65,9 +65,9 @@ unsafe extern "C" fn get_property_callback(
 
 impl<'a> ProcessContext for CffiProcessContext<'a> {
     type FlowFile = CffiFlowFile<'a>; // FlowFile shouldn't outlive the ProcessContext
-    fn get_property(
+    fn get_raw_property<K: PropertySchema + ?Sized>(
```
szaszm: What does K stand for? Key for something? Because I think that's supposed to be the type of the property value, not the property key (name).

minifi_rust/extensions/minifi_rs_playground/src/processors/generate_flow_file/properties.rs
martinzink: it doesnt seem that way no, cargo fmt formatted this i think its based on this 

> [Multi-line elements](https://doc.rust-lang.org/style-guide/expressions.html#multi-line-elements)
If any element in a chain is formatted across multiple lines, put that element and any later elements on their own lines.

minifi_rust/minifi_native/src/api/property.rs
martinzink: 👍 good idea, [is_required: K::IS_REQUIRED](https://github.com/apache/nifi-minifi-cpp/pull/2220/commits/d4d650cdbf041bd5054957ddac1d17feb328ea22)

minifi_rust/minifi_native/src/c_ffi/c_ffi_process_context.rs
martinzink: i dont think it stands for anything, but u are right maybe P for Property? [K -> P in Property context](https://github.com/apache/nifi-minifi-cpp/pull/2220/commits/644ccc8273fb29cca6b4490c89fe5a10b78b09b2)

minifi_rust/extensions/minifi_rs_playground/src/processors/generate_flow_file/properties.rs
szaszm: It's one thing to put it on its own line, and I have nothing against that. It's another thing how you indent it, and here a statement continuation doesn't get indented as a continuation. Maybe splitting the call within the parentheses to multiple lines caused the formatter to end up undoing the continuation indentation for further lines. I believe this is a problem with the formatter, but I didn't find any clear confirmation of that in the rules.

minifi_rust/minifi_native/src/api/property.rs
```diff
@@ -16,102 +16,291 @@
...
+            default_value: self.default_value,
+            constraints: P::CONSTRAINT,
         }
+    }
 
-        if let Some(property_val) = self.get_property(property)? {
-            Ok(Some(humantime::parse_duration(property_val.as_str())?))
-        } else {
-            Ok(None)
+    pub(crate) const fn with_marker<K2: ?Sized + PropertySchema>(&self) -> Property<K2> {
```
szaszm: What does this do?

minifi_rust/extensions/minifi_rs_playground/src/processors/generate_flow_file/properties.rs
szaszm: tbh it's fine to leave it as is, but the formatting seems like a formatter bug to me

minifi_rust/minifi_native/src/api/property.rs
martinzink: this is used for the wrapped the controller services, with this we can change the type of the property (creating an alternative one) during get_controller_service calls with this so one could simply write 
```
pub(crate) const DUMMY_CONTROLLER_SERVICE: Property<Option<DummyControllerService>> = Property::new(
    "Dummy Controller Service",
    "Optional dummy controller service",
);
```
and the output type can be an Option<&DummyControllerService> via the ControllerServiceValue PropertySchema

also this is how the separation is done so the controllerservices can only be queried via get_controller_service and not get_property

fgerlits (approved):

minifi_rust/minifi_native/src/api/property.rs
szaszm: Even after reading this a couple of times, looking at the code, and jumping around to gather context, I still find this really confusing. Could you write a comment somewhere in the code that explains the part from a Property declaration to retrieving a controller service, including jumping through types and traits? I imagine 2-3 paragraphs could be enough to explain this, and I believe it would help future readers too.

minifi_rust/minifi_native/src/api/process_context.rs
```diff
@@ -15,115 +15,39 @@
...
-    fn get_property(
+    fn get_raw_property<P: PropertySchema + ?Sized>(
         &self,
-        property: &Property,
+        property: &Property<P>,
         flow_file: Option<&Self::FlowFile>,
     ) -> Result<Option<String>, MinifiError>;
 
-    fn get_bool_property(
+    fn get_raw_controller_service<Cs, P>(
```
szaszm: What is the difference between a "raw" controller service and one that's not "raw"? I remember that with properties, you get the string value instead of the parsed one, but that doesn't make sense here. Could you add short comments above the respective functions, explaining what they do?

martinzink: raw means agent managed in rust context, so there are RawProcessors and RawControllerServices these are the objects are created and destroyed in minifi code (via callbacks) these wrap the user written Processors and ControllerServices (in options) so the user written Processors are created during schedule and user Written ControllerServices are created during enable.

The C api strictly oeprates with these raw types, but the extensions writer shouldnt need to know about these.

In property context, since the agent deals with strings only raw_property means the actual string set, and property is the parsed verified typed object.

minifi_rust/minifi_native/src/api/process_context.rs
```diff
@@ -15,115 +15,42 @@
...
-    }
+        property: &Property<P>,
+    ) -> Result<Option<&RawCs>, MinifiError>
+    where
+        RawCs: RawControllerService + ComponentIdentifier + 'static,
+        P: PropertySchema + ?Sized;
 
-    fn get_raw_controller_service<Cs>(
+    /// Returns the enabled ControllerService (managed by RawControllerService)
+    fn get_controller_service<Cs>(
```
szaszm, line 43: Whose responsibility is it to cast the controller service to the interface type expected by the property and the extension? Because when a controller service uses multiple inheritance or encapsulation, the pointer value between the pointer to the interface and the pointer to the dynamic typed controller service object can differ. I see get_controller_service_api is supposed to handle this, but why is get_controller_service not the same as get_controller_service_api, doing the conversion from the potentially processor-unknown to the processor-known type?

minifi_rust/minifi_native/src/api/property.rs
```diff
@@ -16,102 +16,303 @@
...
+/// These values will be translated to fill out the
+/// validator, allowed_value, allowed_types, is_required on the agent side
+pub trait PropertySchema {
+    const CONSTRAINT: Option<PropertyConstraints>;
+    const IS_REQUIRED: bool;
+}
+
+/// The requiredness of the property is enforced via this Option impl
+/// If the property is required it should be registered as Property<T>
+/// If the property is not required it should be registered as Property<Option<T>
```
szaszm:
```suggestion
/// If the property is not required it should be registered as Property<Option<T>>
```

minifi_rust/minifi_native/src/api/property.rs
```diff
@@ -16,102 +16,303 @@
...
+    const IS_REQUIRED: bool = false;
+}
 
-    fn get_size_property(&self, property: &Property) -> Result<Option<u64>, MinifiError> {
-        if property.validator != DataSizeValidator {
-            return Err(MinifiError::validation_err(format!(
-                "to use get_size_property {:?} must have DataSizeValidator",
-                property
-            )));
+/// Trait required to register property as Property<T> or Property<Option<T>
```
szaszm:
```suggestion
/// Trait required to register property as Property<T> or Property<Option<T>>
```

minifi_rust/minifi_native/src/api/process_context.rs
martinzink: this one uses exact types, the get_controller_service_api is the one that casts to dyn Trait so not used in producetion yet, usually the caller knows the exact ControllerService type

minifi_rust/minifi_native/src/api/property.rs
martinzink: https://github.com/apache/nifi-minifi-cpp/pull/2220/changes/288e205a9fef79af4737904a69f88f0110ee41b6

minifi_rust/minifi_native/src/api/property.rs
martinzink: https://github.com/apache/nifi-minifi-cpp/pull/2220/changes/288e205a9fef79af4737904a69f88f0110ee41b6

minifi_rust/minifi_native/src/api/process_context.rs
szaszm: I think the processor shouldn't need to care about the dynamic type of a controller service, only what type it expects to get, and if those match, then the conversion is a no-op / identity.

martinzink: yeah but the return type of these differ enough so its not trivial to merge them, one returns a simple Option<&T> the other Option<Box<&Trait>>

minifi_rust/minifi_native/src/api/property.rs
martinzink: https://github.com/apache/nifi-minifi-cpp/pull/2220/changes/f71f18fcc02dfaa9600f1a1a9c86f0565e049911#r3871519483

szaszm (approved):

lordgamez (approved):

PR #2221, opened 2026-07-27
MINIFICPP-2867 Implement ForkEnrichment/JoinEnrichmentAttributes

martinzink:

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/extensions/standard-processors/processors/ForkEnrichment.cpp
+++ b/extensions/standard-processors/processors/ForkEnrichment.cpp
@@ -1,0 +1,61 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "ForkEnrichment.h"
+
+#include "core/Resource.h"
+#include "minifi-cpp/core/ProcessSession.h"
+#include "utils/ProcessorConfigUtils.h"
+
+namespace org::apache::nifi::minifi::standard {
+void ForkEnrichment::initialize() {
+  setSupportedProperties(Properties);
+  setSupportedRelationships(Relationships);
+  ProcessorImpl::initialize();
+}
+
+void ForkEnrichment::onSchedule(core::ProcessContext& context, core::ProcessSessionFactory& session_factory) {
+  max_batch_size_ = utils::parseOptionalU64Property(context, MaxBatchSize);
+  if (max_batch_size_ && *max_batch_size_ == 0) {
+    throw Exception(PROCESSOR_EXCEPTION, "Max Batch Size property is invalid");
+  }
+
+  ProcessorImpl::onSchedule(context, session_factory);
+}
+
+void ForkEnrichment::onTrigger(core::ProcessContext&, core::ProcessSession& session) {
+  uint64_t processed = 0;
+  while (const auto original = session.get()) {
+    const auto enrichment = session.clone(*original);
+
+    original->setAttribute(ENRICHMENT_ROLE, "ORIGINAL");
+    enrichment->setAttribute(ENRICHMENT_ROLE, "ENRICHMENT");
+
+    const std::string group_id = utils::IdGenerator::getIdGenerator()->generate().to_string();
+    original->setAttribute(ENRICHMENT_GROUP_ID, group_id);
+    enrichment->setAttribute(ENRICHMENT_GROUP_ID, group_id);
+
+    session.transfer(original, Original);
+    session.transfer(enrichment, Enrichment);
+    if (max_batch_size_ && ++processed >= max_batch_size_) {
+      break;
+    }
+  }
+}
+
+REGISTER_RESOURCE(ForkEnrichment, Processor);
+}  // namespace org::apache::nifi::minifi::standard
--- a/extensions/standard-processors/processors/JoinEnrichmentAttributes.cpp
+++ b/extensions/standard-processors/processors/JoinEnrichmentAttributes.cpp
@@ -1,0 +1,160 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "JoinEnrichmentAttributes.h"
+
+#include "core/Resource.h"
+#include "minifi-cpp/core/ProcessSession.h"
+#include "utils/AttributeErrors.h"
+#include "utils/EnrichmentUtils.h"
+#include "utils/ProcessorConfigUtils.h"
+
+namespace org::apache::nifi::minifi::standard {
+const core::Relationship JoinEnrichmentAttributes::Self("__self__", "Marks the FlowFile to be owned by this processor");
+
+void JoinEnrichmentAttributes::initialize() {
+  setSupportedProperties(Properties);
+  setSupportedRelationships(Relationships);
+  ProcessorImpl::initialize();
+}
+
+void JoinEnrichmentAttributes::onSchedule(core::ProcessContext& context, core::ProcessSessionFactory& session_factory) {
+  using namespace std::literals::chrono_literals;
+  if (const auto timeout = utils::parseOptionalDurationProperty(context, TimeoutProperty); timeout && *timeout > 0ms) {
+    time_out_tracker_.emplace(*timeout);
+  }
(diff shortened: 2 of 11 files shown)


extensions/standard-processors/processors/JoinEnrichmentAttributes.cpp
```diff
@@ -0,0 +1,160 @@
...
+    session.transfer(flow_file, Invalid);
+    session.transfer(previous_node.mapped(), Invalid);
+    if (!std::ranges::contains(session_flow_files_, previous_node.mapped()->getUUID())) {
+      session.add(previous_node.mapped());
+    }
+    return;
+  }
+
+  if (const auto pair_node = pair_map.extract(group_id)) {
+    logger_->log_trace("Match found");
```
lordgamez: We could log the group id to see which group id the match was found for.

extensions/standard-processors/processors/JoinEnrichmentAttributes.h
```diff
@@ -0,0 +1,139 @@
...
+    std::chrono::steady_clock::time_point timestamp;
+    std::string group_name;
+  };
+
+  std::chrono::steady_clock::duration time_out_;
+  std::deque<TimeStampedGroup> queue_;
+};
+}  // namespace join_enrichment_attributes
+
+using MapType = std::unordered_map<std::string, std::shared_ptr<core::FlowFile>, utils::string::transparent_string_hash, std::equal_to<>>;
```
lordgamez: This could be renamed to be a bit more descriptive.

extensions/standard-processors/processors/JoinEnrichmentAttributes.cpp
```diff
@@ -0,0 +1,160 @@
...
+    core::ProcessSession& session) const {
+  const auto cloned = session.clone(*original);
+  for (const auto& [k, v] : enrichment->getAttributes()) {
+    if (k != ENRICHMENT_ROLE) {
+      cloned->setAttribute(k, v);
+    }
+  }
+  cloned->setAttribute(ENRICHMENT_ROLE, "JOINED");
+  if (!std::ranges::contains(session_flow_files_, original->getUUID())) {
+    session.add(original);
```
lordgamez, line 62: Why is session.add needed here, but not when transferring flow files to Invalid or Self relationships?

extensions/standard-processors/processors/JoinEnrichmentAttributes.h
```diff
@@ -0,0 +1,139 @@
...
+    ORIGINAL,
+    ENRICHMENT,
+  };
+
+  void handleFlowFile(std::shared_ptr<core::FlowFile> flow_file, core::ProcessSession& session, std::chrono::steady_clock::time_point current_time);
+  void join(const std::shared_ptr<core::FlowFile>& original, const std::shared_ptr<core::FlowFile>& enrichment, core::ProcessSession& session) const;
+
+  core::FlowFileStore flow_file_store_;
+  // We need to track current session's FlowFiles (we cant add those)
+  std::vector<utils::Identifier> session_flow_files_;
```
lordgamez: Only addition and contains functionality is used for this container without need for ordering, it may be better to use unordered_set for this.

extensions/standard-processors/processors/JoinEnrichmentAttributes.cpp
martinzink: We only want to add these to the session if they are not from this session. (ProcessSession asserts that a flowfile is only in one category (added, deleted, modified/updaetd) and every transferred flowfile is in session

extensions/standard-processors/processors/JoinEnrichmentAttributes.h
martinzink: 👍 https://github.com/apache/nifi-minifi-cpp/pull/2221/changes/074a90bbed1383dde9af0383bb3c4063456a01cb#diff-328942b134e20b3cc74348741cdfc3d529c1305195f9df4f84d4f6937a143c0aR76 ?

extensions/standard-processors/processors/JoinEnrichmentAttributes.cpp
martinzink: https://github.com/apache/nifi-minifi-cpp/pull/2221/changes/074a90bbed1383dde9af0383bb3c4063456a01cb#diff-e7508634e044ab79b696f8f9ae6e5e40613364813b1e6a987f551897553954feR104

extensions/standard-processors/processors/JoinEnrichmentAttributes.h
martinzink: https://github.com/apache/nifi-minifi-cpp/pull/2221/changes/074a90bbed1383dde9af0383bb3c4063456a01cb#diff-328942b134e20b3cc74348741cdfc3d529c1305195f9df4f84d4f6937a143c0aR140

lordgamez (approved):

extensions/standard-processors/processors/JoinEnrichmentAttributes.h
```diff
@@ -0,0 +1,147 @@
...
+  using ProcessorImpl::ProcessorImpl;
+
+  EXTENSIONAPI static constexpr const char* Description =
+      "Rejoins the forked FlowFiles coming from ForkEnrichment processor, the resulting FlowFile will have the Original's content and all attributes "
+      "from both of them (prioritizing Enrichment's).";
+
+  EXTENSIONAPI static constexpr auto Invalid = core::RelationshipDefinition{"invalid",
+      "Any FlowFiles without the requisite attributes will be routed here"};
+  EXTENSIONAPI static constexpr auto Joined = core::RelationshipDefinition{"joined",
+      "The resultant FlowFile with Records joined together from both the original and enrichment FlowFiles will be routed to this relationship"};
```
fgerlits, line 89: attributes, not records

extensions/standard-processors/processors/ForkEnrichment.cpp
```diff
@@ -0,0 +1,61 @@
...
+  ProcessorImpl::onSchedule(context, session_factory);
+}
+
+void ForkEnrichment::onTrigger(core::ProcessContext&, core::ProcessSession& session) {
+  uint64_t processed = 0;
+  while (const auto original = session.get()) {
+    const auto enrichment = session.clone(*original);
+
+    original->setAttribute(ENRICHMENT_ROLE, "ORIGINAL");
+    enrichment->setAttribute(ENRICHMENT_ROLE, "ENRICHMENT");
```
fgerlits, line 46: I would move the `EnrichmentRole` enum from `JoinEnrichmentAttributes` to somewhere which is visible from here, and use that instead of the "ORIGINAL" and "ENRICHMENT" strings.

extensions/standard-processors/processors/JoinEnrichmentAttributes.h
```diff
@@ -0,0 +1,147 @@
...
+ private:
+  enum class EnrichmentRole {
+    ORIGINAL,
+    ENRICHMENT,
+  };
+
+  void handleFlowFile(std::shared_ptr<core::FlowFile> flow_file, core::ProcessSession& session, std::chrono::steady_clock::time_point current_time);
+  void join(const std::shared_ptr<core::FlowFile>& original, const std::shared_ptr<core::FlowFile>& enrichment, core::ProcessSession& session) const;
+
+  core::FlowFileStore flow_file_store_;
```
fgerlits, line 138: This is old code, but `FlowFileStore` assumes that after we move from `incoming_files_`, it will be empty. This is probably true in practice, but it isn't guaranteed. We should do a `clear` to be safe.

extensions/standard-processors/processors/JoinEnrichmentAttributes.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+  const auto role = flow_file->getAttribute(ENRICHMENT_ROLE) | utils::toExpected(make_error_code(core::AttributeErrorCode::MissingAttribute)) |
+      utils::andThen(parsing::parseEnum<EnrichmentRole>);
+  if (!role) {
+    logger_->log_warn("{} has invalid role due to {}", flow_file->getId(), role.error());
+    session.transfer(flow_file, Invalid);
+    return;
+  }
+
+  std::string group_id = *(flow_file->getAttribute(ENRICHMENT_GROUP_ID));
```
fgerlits, line 88: we should check or assert that the attribute exists before dereferencing it

extensions/standard-processors/processors/JoinEnrichmentAttributes.cpp
```diff
@@ -0,0 +1,160 @@
...
+  auto& my_map = role == EnrichmentRole::ENRICHMENT ? enrichments_ : originals_;
+  auto& pair_map = role == EnrichmentRole::ENRICHMENT ? originals_ : enrichments_;
+
+  if (const auto previous_node = my_map.extract(group_id)) {
+    logger_->log_warn("Encountered duplicate {} for {}, routing both to Invalid", magic_enum::enum_name(*role), group_id);
+    session.transfer(flow_file, Invalid);
+    session.transfer(previous_node.mapped(), Invalid);
+    if (!std::ranges::contains(session_flow_files_, previous_node.mapped()->getUUID())) {
+      session.add(previous_node.mapped());
+    }
```
fgerlits, line 99: Does this work? I would expect that we need to add the flow file first before we can transfer it. If this is the correct order, then please add a comment explaining why.

extensions/standard-processors/tests/features/enrichment.feature
fgerlits, line 1: Can you add a couple more attributes to the original flow file, please? One which is not touched by the enrichment branch, and one which is overwritten by it. At the end, both (together with the existing one, all three) should show up in the log with the correct value.

extensions/standard-processors/tests/unit/JoinEnrichmentAttributesTests.cpp
```diff
@@ -0,0 +1,139 @@
...
+      test_controller(minifi::test::utils::make_processor<JoinEnrichmentAttributes>("JoinEnrichmentAttributes"));
+  const auto proc = test_controller.getProcessor();
+  CHECK(test_controller.plan->setProperty(proc, JoinEnrichmentAttributes::TimeoutProperty.name, "1 ms"));
+
+  const auto first_trigger = test_controller.trigger(minifi::test::InputFlowFileData{.content = "first",
+      .attributes = {{std::string{ENRICHMENT_GROUP_ID}, "foo"}, {std::string{ENRICHMENT_ROLE}, "ORIGINAL"}, {"first_attr", "1"}}});
+  // First trigger no output (it holds the original waiting for its pair)
+  CHECK(std::ranges::all_of(first_trigger, [](const auto& res) -> bool { return res.second.empty(); }));
+
+  std::this_thread::sleep_for(1ms);
```
fgerlits, line 115: are you sure this is not going to be flaky? maybe we could sleep for 2 ms, to make flakiness less likely

PR #2225, opened 2026-08-04
MINIFICPP-2749 Add EncryptContentPGP and DecryptContentPGP

martinzink:

### For all changes:

### For code changes:

### Note:


minifi_rust/extensions/minifi_pgp/src/controller_services/private_key_service/controller_service_definition.rs
```diff
@@ -0,0 +1,31 @@
...
+use minifi_native::{
+    ControllerServiceDefinition, Property, PropertyDefinition, ProvidedInterface,
+    property_definitions,
+};
+
+pub(super) const KEY_FILE: Property<Option<SecretKeyFile>> = Property::new(
+    "Key File",
+    "File path to PGP Secret Key encoded in binary or ASCII Armor",
+)
+.supports_expression_language();
```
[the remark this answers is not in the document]
martinzink: EL language is valid without Flowfiles

minifi_rust/extensions/minifi_pgp/src/processors/decrypt_content.rs
```diff
@@ -0,0 +1,356 @@
...
+        if self.decompress_data && decrypted_msg.is_compressed() {
+            decrypted_msg = decrypted_msg
+                .decompress()
+                .map_err(MinifiError::other)
+                .route_err_to_failure()?
+        };
+
+        let attributes_to_add = Self::extract_attributes_from_decrypted_message(&decrypted_msg);
+        let _written_bytes =
+            std::io::copy(&mut decrypted_msg.into_inner(), output_stream).route_err_to_failure()?;
```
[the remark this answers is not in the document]
martinzink: good catch and the fix is not straightforward, so we will skip this feature for now and address it in a followup PR

minifi_rust/extensions/minifi_pgp/minifi_pgp.md
```diff
@@ -0,0 +1,119 @@
...
+
+| Name               | Default Value | Allowable Values | Description                                                                                                                                                                          |
+|--------------------|---------------|------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| **File Encoding**  | BINARY        | ASCII<br/>BINARY | File Encoding for encryption                                                                                                                                                         |
+| Symmetric Password |               |                  | Password used for encrypting data with Password-Based Encryption<br/>**Sensitive Property: true**                                                                                    |
```
szaszm, line 71: Can we rename this to Passphrase to match NiFi?

minifi_rust/extensions/minifi_pgp/minifi_pgp.md
```diff
@@ -0,0 +1,119 @@
...
+
+Decrypt contents of OpenPGP messages.
+
+### Properties
+
+In the list below, the names of required properties appear in bold. Any other properties (not in bold) are considered optional. The table also indicates any default values, and whether a property supports the NiFi Expression Language.
+
+| Name                | Default Value | Allowable Values | Description                                                                                                 |
+|---------------------|---------------|------------------|-------------------------------------------------------------------------------------------------------------|
+| Symmetric Password  |               |                  | Password used for decrypting data encrypted with Password-Based Encryption<br/>**Sensitive Property: true** |
```
szaszm, line 40: Can we rename this to Passphrase to match NiFi?

minifi_rust/extensions/minifi_pgp/src/test_utils/mod.rs
szaszm, line 1: Can we limit this file, and other test utilities files, to `#[cfg(test)]`?

minifi_rust/extensions/minifi_pgp/src/processors/encrypt_content/processor_definition.rs
szaszm, line 1: I'd keep all of the processor-related metadata in the same file as the processor implementation

szaszm (commented):
It's great to see the new Rust API working beautifully in practice, I really like the API design. A few minor comments below. I'm about halfway through, and will finish with a more thorough review next week.

minifi_rust/extensions/minifi_pgp/features/steps/steps.py
```diff
@@ -0,0 +1,88 @@
...
+    checking_steps,  # noqa: F401
+    configuration_steps,  # noqa: F401
+    core_steps,  # noqa: F401
+    flow_building_steps,  # noqa: F401
+)
+
+
+@step("an EncryptContentPGP processor with a PGPPublicKeyService is set up")
+def step_encrypt_content_with_service(context: MinifiTestContext):
+    dir_path = os.path.dirname(os.path.realpath(__file__))
```
lordgamez: It may be better to add a `context.resource_dir` in the environment.py and reference that in every step instead of always calculating the directory and using relative paths separately.

minifi_rust/extensions/minifi_pgp/features/encrypt_decrypt.feature
```diff
@@ -0,0 +1,69 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+@SUPPORTS_WINDOWS
+Feature: Test PGP extension's encryption and decryption capabilities
+
+  Background: The pgp library is successfully built on linux
```
lordgamez: The background description "successfully built on linux" is a bit strange with the added `@SUPPORTS_WINDOWS` tag

minifi_rust/extensions/minifi_pgp/features/encrypt_decrypt.feature
```diff
@@ -0,0 +1,69 @@
...
+
+  Scenario: The pgp library is loaded into minifi
```
lordgamez, line 19: Should this be tested for all extensions? If this is the problem would it be obvious from the other failure logs?

minifi_rust/extensions/minifi_pgp/minifi_pgp.md
lordgamez, line 1: Shouldn't this be part of the PROCESSORS.md and CONTROLLERS.md or we want to keep these separately?

minifi_rust/extensions/minifi_pgp/features/steps/steps.py
szaszm, line 1: I think we need a test covering the most secure happy path: one where private keys are protected with a long passphrase. Changing existing tests is fine for me too.

minifi_rust/extensions/minifi_pgp/src/controller_services/public_key_service/controller_service_definition.rs
```diff
@@ -0,0 +1,43 @@
...
+use crate::controller_services::key_file_property::PublicKeyFile;
+use crate::controller_services::key_property::PublicKey;
+use minifi_native::{
+    ControllerServiceDefinition, Property, PropertyDefinition, ProvidedInterface,
+    property_definitions,
+};
+
+pub(crate) const KEYRING_FILE: Property<Option<PublicKeyFile>> = Property::new(
+    "Keyring File",
+    "File path to PGP Keyring or Public Key encoded in binary or ASCII Armor",
```
szaszm: Does raw binary work at all in a minifi flow definition? I suspect not, so I'd remove the claim from the description, even if we leave the handling code unchanged, and there may be a few lucky binary keys that happen to be JSON/YAML escape-able

minifi_rust/extensions/minifi_pgp/minifi_pgp.md
martinzink: passphrase for me signals its used to decrypt the private key, but i can change it if we want to be more in line with nifi

minifi_rust/extensions/minifi_pgp/src/test_utils/mod.rs
martinzink: its already is look at minifi_rust/extensions/minifi_pgp/src/lib.rs 
```
#[cfg(test)]
mod test_utils;
```

minifi_rust/extensions/minifi_pgp/src/processors/encrypt_content/processor_definition.rs
martinzink: still unsure which looks better
[move defs into super files](https://github.com/apache/nifi-minifi-cpp/pull/2225/commits/b83eae8992cd0d121959d6e1c29314e2d8ab363b)

minifi_rust/extensions/minifi_pgp/features/steps/steps.py
martinzink: good idea, [python changes](https://github.com/apache/nifi-minifi-cpp/pull/2225/commits/1be37bdcb073f1a065afee468d0657f792fcc6be)

minifi_rust/extensions/minifi_pgp/features/encrypt_decrypt.feature
martinzink: yeah thats valid, thats a leftover thing because i had to explictily build the extension to work (im on mac so cant use the locally built artifact) but thats not the case anymore with cargo behave-alpine/debian it works automatically so no need for that anymore, removed it in [python changes](https://github.com/apache/nifi-minifi-cpp/pull/2225/commits/1be37bdcb073f1a065afee468d0657f792fcc6be)

minifi_rust/extensions/minifi_pgp/features/encrypt_decrypt.feature
martinzink: yeah probably not, maybe the fact that this one requires no warnings and no errors, (other test permit warnings) but i can remove it if it seems too basic to be included

minifi_rust/extensions/minifi_pgp/minifi_pgp.md
martinzink: good idea to include them, but i would also keep the per extension documentation [add pgp documentation to root docs](https://github.com/apache/nifi-minifi-cpp/pull/2225/commits/bd385900fcd9bdd724decf01196579d03ec511be)

minifi_rust/extensions/minifi_pgp/src/controller_services/key_property.rs
szaszm, line 1: If there is an easy way to reduce the code duplication without unifying public and private key types, and without exploding code complexity, it might be worth exploring.

minifi_rust/extensions/minifi_pgp/src/controller_services/public_key_service.rs
```diff
@@ -0,0 +1,259 @@
...
+        "Keyring File",
+        "File path to PGP Keyring or Public Key encoded in binary or ASCII Armor",
+    )
+    .supports_expression_language();
+
+    pub(crate) const KEYRING: Property<Option<PublicKey>> = Property::new(
+        "Keyring",
+        "PGP Keyring or Public Key encoded in ASCII Armor",
+    )
+    .sensitive();
```
szaszm: Public keys are not sensitive

minifi_rust/extensions/minifi_pgp/src/processors/decrypt_content.rs
```diff
@@ -0,0 +1,399 @@
...
+            pks.get_the_ring()
+        } else {
+            TheRing::default()
+        };
+
+        ring.decrypt_options = ring.decrypt_options.enable_gnupg_aead();
+
+        if let Some(sym_passwd) = &self.symmetric_password {
+            ring.message_password.push(sym_passwd);
+        }
```
szaszm, line 71: What if both a private key and a symmetric passphrase are used? How does that work? Does it work at all? Does it double encrypt and decrypt the message, once with asymmetric and once with symmetric encryption? Does roundtrip with EncryptContent work in that case? If not, should we disable that and exforce an exclusive-OR relationship between the different key types, in schedule?

minifi_rust/extensions/minifi_pgp/src/processors/decrypt_content.rs
```diff
@@ -0,0 +1,399 @@
...
+    use crate::utils;
+    use minifi_native::{
+        OutputAttribute, ProcessorDefinition, ProcessorInputRequirement, Property,
+        PropertyDefinition, Relationship, property_definitions,
+    };
+
+    pub(super) const LITERAL_DATA_FILENAME: OutputAttribute = OutputAttribute {
+        name: "pgp.literal.data.filename",
+        relationships: &["success"],
+        description: "Filename from decrypted Literal Data (Note that OpenPGP signatures do not include the formatting octet, the file name, and the date field of the Literal Data packet in a signature hash; therefore, those fields are not protected against tampering in a signed document. Therefore a lot of implementations omit these inherently malleable metadata)",
```
szaszm: Should we ignore the unprotected metadata instead of using them when they're present? As a defense-in-depth measure, I think we should protect flowfile attributes from untrusted values, as some of our processors treat them as trusted data, opening attack vectors.

minifi_rust/extensions/minifi_pgp/src/processors/decrypt_content.rs
```diff
@@ -0,0 +1,399 @@
...
+    };
+
+    pub(super) const LITERAL_DATA_MODIFIED: OutputAttribute = OutputAttribute {
+        name: "pgp.literal.data.modified",
+        relationships: &["success"],
+        description: "Modified Date from decrypted Literal Data (Note that OpenPGP signatures do not include the formatting octet, the file name, and the date field of the Literal Data packet in a signature hash; therefore, those fields are not protected against tampering in a signed document. Therefore a lot of implementations omit these inherently malleable metadata)",
```
szaszm: same here

minifi_rust/extensions/minifi_pgp/src/processors/encrypt_content.rs
```diff
@@ -0,0 +1,391 @@
...
+    use crate::utils;
+    use minifi_native::{
+        OutputAttribute, ProcessorDefinition, ProcessorInputRequirement, Property,
+        PropertyDefinition, Relationship, property_definitions,
+    };
+
+    pub(crate) const FILE_ENCODING: Property<FileEncoding> =
+        Property::new("File Encoding", "File Encoding for encryption")
+            .with_default(FileEncoding::Binary.into_str());
+    pub(crate) const SYMMETRIC_PASSWORD: Property<Option<utils::Password>> = Property::new(
```
szaszm:
```suggestion

    pub(crate) const SYMMETRIC_PASSWORD: Property<Option<utils::Password>> = Property::new(
```

minifi_rust/extensions/minifi_pgp/src/utils.rs
```diff
@@ -0,0 +1,37 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+use minifi_native::{
+    MinifiError, PropertyConstraints, PropertySchema, PropertyType, StandardPropertyValidator,
+};
+
+pub(crate) struct Password {}
+
+impl PropertySchema for Password {
```
szaszm, line 24: Can we make this sensitive by default?

CONTROLLERS.md
```diff
@@ -245,6 +247,39 @@ In the list below, the names of required properties appear in bold. Any other pr
...
+
+PGP Private Key Service provides Private Keys loaded from files or properties
+
+### Properties
+
+In the list below, the names of required properties appear in bold. Any other properties (not in bold) are considered optional. The table also indicates any default values, and whether a property supports the NiFi Expression Language.
+
+| Name           | Default Value | Allowable Values | Description                                                                                             |
+|----------------|---------------|------------------|---------------------------------------------------------------------------------------------------------|
+| Key File       |               |                  | File path to PGP Secret Key encoded in binary or ASCII Armor<br/>**Supports Expression Language: true** |
```
szaszm: If this property allows inline key encoded in binary or ascii, then the property needs to be sensitive. I think we should disallow inline keys instead, or use the sensitive Key property for that purpose.

I would also rename the properties to match NiFi, even at a slight cost to clarity.

minifi_rust/extensions/minifi_pgp/minifi_pgp.md
szaszm: I'd rather be matching NiFi even if it's at a slight cost to clarity. Password and passphrase are used similarly, passphrase just signals that it can (and should) be longer than a word.

minifi_rust/extensions/minifi_pgp/src/processors/encrypt_content/processor_definition.rs
szaszm: If you prefer them separate, then it's ultimately your call as the author. I prefer them in a single file.

CONTROLLERS.md
martinzink: no this is a file path only

minifi_rust/extensions/minifi_pgp/src/utils.rs
martinzink: thats a great improvement idea, but probably out of the scope of this PR

minifi_rust/extensions/minifi_pgp/src/processors/decrypt_content.rs
martinzink: it can be decrypted with either, its quite smart, pgp generates a single session token it uses that to encrypt the message/file symmetrically and then it encrypts the session token twice, once with public key, once with the configured symmetric key, then saves both results, during decrypiption if it can decrypt the sesion key with either private or symmetric key it can decrypt the file.

minifi_rust/extensions/minifi_pgp/src/controller_services/public_key_service/controller_service_definition.rs
martinzink: this is. afile path, we dont support explicit binary keys its either ASCII via property or ASCII/BINARY via file

minifi_rust/extensions/minifi_pgp/features/encrypt_decrypt.feature
lordgamez: Well this having the requirement of having no warnings and errors could be useful we can keep it for now, maybe remove it later when all tests require this.

lordgamez (approved):

minifi_rust/extensions/minifi_pgp/src/controller_services/public_key_service/controller_service_definition.rs
szaszm: Okay, I've misinterpreted the text, but all good then.

CONTROLLERS.md
szaszm: again, I've misinterpreted the text, but the second part still stands: I'd rename the properties to match NiFi

minifi_rust/extensions/minifi_pgp/features/steps/steps.py
martinzink: We already test it only the passphrase is on the smaller side, but Im not sure what coverage do we get from increasing the passphrase from current 'whiterabbit' to something like 'correCt-bAttery-horSe-staple'

minifi_rust/extensions/minifi_pgp/src/controller_services/key_property.rs
martinzink: Since they are completely different types, its either this or macros or add our own trait which we would have to implement twice so not much better. I've tried it originally but didnt like it, feel free to play with it, maybe you can come up with something better.

minifi_rust/extensions/minifi_pgp/src/controller_services/public_key_service.rs
martinzink: https://github.com/apache/nifi-minifi-cpp/pull/2225/changes/dd9c84f9e0e120ee3430f61aeae6aebb7a0e26e4#diff-0c1a5f6a8f18176688013bc55ddbf3d326638d7f6b6dde57f2d9650df430cafbL75-L76

minifi_rust/extensions/minifi_pgp/src/processors/decrypt_content.rs
martinzink: https://github.com/apache/nifi-minifi-cpp/pull/2225/changes/dd9c84f9e0e120ee3430f61aeae6aebb7a0e26e4#diff-04a53c5b836714443ecb6da3dd630ee5c5ca30adff3012af7a13f5f71dcd7f19L174-L175

minifi_rust/extensions/minifi_pgp/src/processors/decrypt_content.rs
martinzink: https://github.com/apache/nifi-minifi-cpp/pull/2225/changes/dd9c84f9e0e120ee3430f61aeae6aebb7a0e26e4#diff-04a53c5b836714443ecb6da3dd630ee5c5ca30adff3012af7a13f5f71dcd7f19L174-L175

minifi_rust/extensions/minifi_pgp/minifi_pgp.md
martinzink: changed the properties to match nifi in [review changes](https://github.com/apache/nifi-minifi-cpp/pull/2225/commits/dd9c84f9e0e120ee3430f61aeae6aebb7a0e26e4) and [review changes 2](https://github.com/apache/nifi-minifi-cpp/pull/2225/commits/d3982f49f43641588f3dfe65c5bdc12ef23d1c2f)

minifi_rust/extensions/minifi_pgp/features/steps/steps.py
szaszm: fair, I think I missed the passphrase part of the second test, mb

minifi_rust/extensions/minifi_pgp/src/controller_services/private_key_service.rs
```diff
@@ -0,0 +1,249 @@
...
+    };
+
+    pub(super) const KEY_FILE: Property<Option<SecretKeyFile>> = Property::new(
+        "Key File",
+        "File path to PGP Secret Key encoded in binary or ASCII Armor",
+    )
+    .supports_expression_language();
+
+    pub(super) const KEY: Property<Option<SecretKey>> =
+        Property::new("Key", "Secret Key encoded in ASCII Armor").sensitive();
```
szaszm, line 88: Is this identical to the Keyring property in the NiFi equivalent? If so, I'd rename this one too, and Key file to Keyring file.

minifi_rust/extensions/minifi_pgp/src/processors/encrypt_content.rs
```diff
@@ -0,0 +1,392 @@
...
+        _logger: &L,
+    ) -> Result<Self, MinifiError>
+    where
+        Self: Sized,
+    {
+        let file_encoding = context.get_property(&FILE_ENCODING)?;
+        let symmetric_password = context.get_property(&SYMMETRIC_PASSPHRASE)?;
+
+        let public_key_service = context.get_controller_service(&PUBLIC_KEY_SERVICE)?;
+        let public_key_search = context.get_raw_property(&PUBLIC_KEY_SEARCH)?;
```
szaszm: I forgot how get_raw_property differs from get_property or get_controller_service. Could you add a comment for get_raw_property? Does it skip validation, EL eval, or both?
If it doesn't evaluate EL, then it looks good to me. If it does, then it can result in an empty expansion where the expression depends on flow file attributes, resulting in an error even when the configuration is correct.

martinzink: yeah i was under the impression that it doesnt eval EL, but it does so thats not gonna work here, and in that case there is no point of haing get_raw_property anymore, maybe we could add some support to the c api, `minifi_process_context_is_property_set` ?

minifi_rust/extensions/minifi_pgp/src/controller_services/private_key_service.rs
martinzink: funnily enough i did follows nifi's naming convention, it uses Key and Key file in private key context and KeyRing Key Ring File in public context 🤷 
https://nifi.apache.org/docs/nifi-docs/components/org.apache.nifi/nifi-pgp-service-nar/1.28.0/org.apache.nifi.pgp.service.standard.StandardPGPPublicKeyService/index.html
https://nifi.apache.org/docs/nifi-docs/components/org.apache.nifi/nifi-key-service-nar/1.28.0/org.apache.nifi.key.service.StandardPrivateKeyService/index.html

minifi_rust/extensions/minifi_pgp/src/processors/encrypt_content.rs
martinzink: [get_raw_property does EL, so its not a good idea to check that in sch…](https://github.com/apache/nifi-minifi-cpp/pull/2225/commits/70673484965fe0d999c4e771f5090cb03d22eb3f)

we should create a new PR that addresses this missing capability in our stable API

adamdebreceni (approved):

minifi_rust/extensions/minifi_pgp/src/controller_services/private_key_service.rs
szaszm: https://nifi.apache.org/components/org.apache.nifi.pgp.service.standard.StandardPGPPrivateKeyService/

StandardPGPPrivateKeyService is not the same as StandardPrivateKeyService, super confusing. I navigated the docs by clicking on the controller service type from the relevant processor pages. Also I suggest looking at the latest docs, I wouldn't want to reimplement legacy interfaces when we could use the latest.

minifi_rust/extensions/minifi_pgp/src/processors/encrypt_content.rs
szaszm: I'd also consider separating property query from EL eval, and leave the eval to the extension code. But that doesn't concern this PR.

PR #2227, opened 2026-08-04
MINIFICPP-2876 Manifest bundles invalid groupname

martinzink:

### For all changes:

### For code changes:

### Note:


extensions/standard-processors/tests/unit/ManifestTestHelper.h
szaszm, line 1: Is this supposed to be under standard-processors?

szaszm:
I think copilot got overwhelmed by the 13k diff from the manifest change

extensions/standard-processors/tests/unit/ManifestTestHelper.h
martinzink: probably not, removed it entirely and moved its contents (after proper refactor (removed redundant helpers etc)) to TestUtils.h [refactor ManifestTestHelper.h into TestUtils.h](https://github.com/apache/nifi-minifi-cpp/pull/2227/commits/6c3fa33993170b97621c406fe030f09abeb87cc6)

szaszm (approved):

core-framework/include/core/Resource.h
```diff
@@ -57,6 +57,11 @@ class StaticClassType {
 #else
       auto module_name = "minifi-system";
 #endif
+#ifdef MINIFI_EXTENSION_GROUP_NAME
+    auto group_name = MAKESTRING(MINIFI_EXTENSION_GROUP_NAME);
+#else
+    auto group_name = "org.apache.nifi.minifi";
+#endif
```
fgerlits, line 64: Can `MINIFI_EXTENSION_GROUP_NAME` be undefined? It may be better to force it to be defined, and generate a compile error otherwise.

minifi-api/include/minifi-cpp/agent/agent_docs.h
```diff
@@ -51,39 +57,125 @@ struct ClassDescription {
...
-  std::vector<ClassDescription> controller_services;
-  std::vector<ClassDescription> parameter_providers;
-  std::vector<ClassDescription> other_components;
+struct BundleCoordinate {
+  std::string name;
+  std::string group_name;
+  std::string version;
+
+  auto operator<=>(const BundleCoordinate& rhs) const = default;
+};
```
fgerlits, line 66: I think the previous name, `BundleIdentifier`, described the purpose of this struct better than `BundleCoordinate`. Can we revert to that?

extensions/python/PythonCreator.h
```diff
@@ -174,7 +175,8 @@ class PythonCreator : public minifi::core::CoreComponentImpl {
       .inputRequirement_ = toString(processor->getInputRequirement()),
       .isSingleThreaded_ = processor->isSingleThreaded()};
 
-    minifi::ClassDescriptionRegistry::getMutableClassDescriptions()[details].processors.push_back(description);
+    auto& bundle_components = minifi::ClassDescriptionRegistry::getMutableClassDescriptions().try_emplace(details, details).first->second;
+    bundle_components.addClassDescription(description, ResourceType::Processor);
```
fgerlits: This `try_emplace(details, details).first->second` looks a bit confusing, and it is repeated several times. Can we add a helper function so that we can write something like
```suggestion
    auto& bundle_components = minifi::ClassDescriptionRegistry::getMutableComponents(bundle_identifier);
    bundle_components.addClassDescription(description, ResourceType::Processor);
```
instead?

minifi-api/include/minifi-cpp/agent/agent_docs.h
szaszm: BundleCoordinate supposedly comes from nifi.
BundleCoordinate: https://github.com/apache/nifi/blob/main/nifi-framework-api/src/main/java/org/apache/nifi/bundle/BundleCoordinate.java
There is also a Bundle type in nifi-api: https://github.com/apache/nifi-api/blob/main/src/main/java/org/apache/nifi/flow/Bundle.java
I don't have a strong preference either way, because I don't know how well or how directly all of this maps to the corresponding NiFi types.

adamdebreceni (approved):

minifi-api/include/minifi-cpp/agent/agent_docs.h
martinzink: yeah, I changed it to mimic NiFi

core-framework/include/core/Resource.h
martinzink: sure thing, https://github.com/apache/nifi-minifi-cpp/pull/2227/changes/2f49406fee65c9dcf3738559f7c2dc0dc0c3e2ad
since we only defined this in the register_extension macro i had to add this into almost every tests/cmakelist (anything that includes Resource.h)

extensions/python/PythonCreator.h
martinzink: 👍 https://github.com/apache/nifi-minifi-cpp/pull/2227/changes/2f49406fee65c9dcf3738559f7c2dc0dc0c3e2ad

fgerlits (approved):

PR #2254, opened 2026-09-03
MINIFICPP-2877 Fix ASan and LSan warnings

fgerlits:
ASan/LSan fixes:
* fix a SEGV-at-shutdown error in Tcp/UdpServer
* fix a return-string_view-to-temporary-string bug in TestUtils
* fix a leak in the libssh2 library
* free the shared_ptr result of async_resolve when it is no longer needed
* fix a leak in HTTPSiteToSiteTest
* fix a leak by restoring the old controller service provider if an update fails
* fix a leak in the Python bindings
* fix an alloc-dealloc-mismatch bug: new + free (test code only)

ODR violation warnings are suppressed, because the way we link and dlopen
libraries causes many false positives. If two symbols are defined with the
same name but different sizes, they will still be flagged.

Upgrade Abseil to the latest release to fix a leak, and upgrade Grpc (which uses
Abseil) to the latest release. These required some new valgrind suppressions
(generated by Claude Code, like the earlier suppressions).

Add a new ASan/LSan job to the weekly memcheck CI run.

---

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/cmake/Abseil.cmake
+++ b/cmake/Abseil.cmake
@@ -28,8 +28,8 @@ set(PC ${Bash_EXECUTABLE}  -c "set -x &&\
 
 FetchContent_Declare(
         absl
-        URL      https://github.com/abseil/abseil-cpp/archive/refs/tags/20260526.0.tar.gz
-        URL_HASH SHA256=6e1aee535473414164bf83e4ebc40240dec71a4701f8a642d906e95bea1aea0c
+        URL      https://github.com/abseil/abseil-cpp/archive/refs/tags/20260817.0.tar.gz
+        URL_HASH SHA256=f7e05179df39c45434cad433f5783840bb3788ef322976f9138bc6b72b3a107d
         PATCH_COMMAND "${PC}"
         OVERRIDE_FIND_PACKAGE
         SYSTEM
--- a/cmake/BuildTests.cmake
+++ b/cmake/BuildTests.cmake
@@ -17,6 +17,22 @@
 
 include(GetCatch2)
 
+if (MINIFI_ADVANCED_ASAN_BUILD)
+    file(MAKE_DIRECTORY "${CMAKE_BINARY_DIR}/asan_logs")
+
+    # Route each test's AddressSanitizer/LeakSanitizer output to <build>/asan_logs/<test-name>.<pid>
+    # and suppress odr-violation warnings (only if the two symbols have the same size). Hundreds of global
+    # symbols are present in two or more .so's, so we don't want to suppress each individually.
+    function(add_test)
+        _add_test(${ARGV})
+        cmake_parse_arguments(MINIFI_TEST "" "NAME" "COMMAND" ${ARGV})
+        if (MINIFI_TEST_NAME)
+            set_property(TEST "${MINIFI_TEST_NAME}" APPEND PROPERTY
+                ENVIRONMENT "ASAN_OPTIONS=detect_odr_violation=1:log_path=${CMAKE_BINARY_DIR}/asan_logs/${MINIFI_TEST_NAME}")
+        endif()
+    endfunction()
+endif()
+
 ### test functions
 MACRO(GETSOURCEFILES result curdir)
     FILE(GLOB children RELATIVE ${curdir} ${curdir}/*)
--- a/cmake/BundledIodbc.cmake
+++ b/cmake/BundledIodbc.cmake
@@ -33,7 +33,7 @@ ExternalProject_Add(
     CMAKE_COMMAND ""
     UPDATE_COMMAND ""
     INSTALL_COMMAND make install
-    CONFIGURE_COMMAND bash "-c" "./autogen.sh && ./configure --prefix=${IODBC_BYPRODUCT_DIR} --with-pic CFLAGS='${CMAKE_C_FLAGS} -std=gnu17'"
+    CONFIGURE_COMMAND bash "-c" "./autogen.sh && ./configure --prefix=${IODBC_BYPRODUCT_DIR} --with-pic CFLAGS='${PASSTHROUGH_CMAKE_C_FLAGS} -std=gnu17'"
     STEP_TARGETS build
     BUILD_BYPRODUCTS "${IODBC_BYPRODUCT_DIR}/${IODBC_BYPRODUCT}"
     EXCLUDE_FROM_ALL TRUE
--- a/extension-framework/include/utils/net/ConnectionHandler.h
+++ b/extension-framework/include/utils/net/ConnectionHandler.h
@@ -17,6 +17,8 @@
 
 #pragma once
 
+#include <vector>
+
 #include <asio/read.hpp>
 
 #include "utils/net/AsioCoro.h"
@@ -64,7 +66,7 @@ class ConnectionHandler final : public ConnectionHandlerBase {
   [[nodiscard]] asio::awaitable<std::error_code> setupUsableSocket(asio::io_context& io_context) override;
   [[nodiscard]] bool hasUsableSocket() const { return socket_ && socket_->lowest_layer().is_open(); }
 
-  asio::awaitable<std::error_code> establishNewConnection(const asio::ip::tcp::resolver::results_type& endpoints, asio::io_context& io_context_);
+  asio::awaitable<std::error_code> establishNewConnection(const std::vector<asio::ip::tcp::endpoint>& endpoints, asio::io_context& io_context_);
   [[nodiscard]] asio::awaitable<std::tuple<std::error_code, size_t>> write(const asio::const_buffer& buffer) override;
   [[nodiscard]] asio::awaitable<std::tuple<std::error_code, size_t>> read(asio::mutable_buffer& buffer) override;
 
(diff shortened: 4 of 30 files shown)


extension-framework/include/utils/net/Server.h
```diff
@@ -67,9 +75,21 @@ class Server {
...
+  // Spawn a coroutine on io_context_ with a cancellation slot so stop() can end it and let the context drain
+  // gracefully. Must be called from the io_context thread (i.e. from run() before io_context_.run(), or from within
+  // a coroutine running on it); cancellation_signals_ is only ever touched on that thread, so it needs no locking.
+  template<typename T>
+  void asyncSpawn(asio::awaitable<T> coroutine) {
+    const auto cancellation_signal_it = cancellation_signals_.emplace(cancellation_signals_.end());
+    asio::co_spawn(io_context_, std::move(coroutine),
+        asio::bind_cancellation_slot(cancellation_signal_it->slot(),
+            [this, cancellation_signal_it](std::exception_ptr, auto&&...) { cancellation_signals_.erase(cancellation_signal_it); }));
+  }
```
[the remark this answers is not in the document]
fgerlits: Exceptions were also swallowed silently before this PR; I have added a comment about this.

libminifi/include/core/controller/StandardControllerServiceNode.h
```diff
@@ -31,10 +31,9 @@ namespace org::apache::nifi::minifi::core::controller {
 
 class StandardControllerServiceNode : public ControllerServiceNode {
  public:
-  explicit StandardControllerServiceNode(std::shared_ptr<ControllerService> service, std::shared_ptr<ControllerServiceProvider> provider, std::string id,
-                                         std::shared_ptr<Configure> configuration)
+  explicit StandardControllerServiceNode(std::shared_ptr<ControllerService> service, ControllerServiceProvider* provider, std::string id, std::shared_ptr<Configure> configuration)
       : ControllerServiceNode(std::move(service), std::move(id), std::move(configuration)),
-        provider(std::move(provider)),
+        provider(provider),
```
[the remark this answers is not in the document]
fgerlits: Good point. I will revert this change, for now, as fixing it would be a bigger change, and should go in a separate pull request.

cmake/BuildTests.cmake
```diff
@@ -17,6 +17,22 @@
...
+    # and suppress odr-violation warnings (only if the two symbols have the same size). Hundreds of global
+    # symbols are present in two or more .so's, so we don't want to suppress each individually.
+    function(add_test)
+        _add_test(${ARGV})
+        cmake_parse_arguments(MINIFI_TEST "" "NAME" "COMMAND" ${ARGV})
+        if (MINIFI_TEST_NAME)
+            set_property(TEST "${MINIFI_TEST_NAME}" APPEND PROPERTY
+                ENVIRONMENT "ASAN_OPTIONS=detect_odr_violation=1:log_path=${CMAKE_BINARY_DIR}/asan_logs/${MINIFI_TEST_NAME}")
+        endif()
+    endfunction()
```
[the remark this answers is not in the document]
fgerlits: If you redefine a built-in CMake function, the original function will be available inside the new implementation with an underscore prepended, so this does work. This is a widely used CMake technique (so CMake is unlikely to drop it, since that would break many projects), but it is not officially supported, and has some issues (e.g., redefining the same function a second time will not work).

I don't think Copilot's solutions would work without creating an infinite call loop. We could write a `minifi_add_test` wrapper which adds the ASAN options when needed, but then all `add_test` calls would need to be changed to `minifi_add_test` and all future test writers will need to remember to use `minifi_add_test` instead of `add_test`. Or we could define the ASAN_OPTIONS globally, but then ASan results would be written to files named `asan.<pid>` (without the name of the test), which is not great.

I still think redefining `add_test` as above is the cleanest way to do this, but I will keep this comment open so others can comment.

cmake/BundledIodbc.cmake
```diff
@@ -33,7 +33,7 @@ ExternalProject_Add(
     CMAKE_COMMAND ""
     UPDATE_COMMAND ""
     INSTALL_COMMAND make install
-    CONFIGURE_COMMAND bash "-c" "./autogen.sh && ./configure --prefix=${IODBC_BYPRODUCT_DIR} --with-pic CFLAGS='${CMAKE_C_FLAGS} -std=gnu17'"
+    CONFIGURE_COMMAND bash "-c" "./autogen.sh && ./configure --prefix=${IODBC_BYPRODUCT_DIR} --with-pic CFLAGS='${PASSTHROUGH_CMAKE_C_FLAGS} -std=gnu17'"
```
fgerlits, line 36: This is unrelated to ASan, but it was clearly a bug, and it's a one-line fix. I will remove it from this PR if requested.

libminifi/test/libtest/integration/HTTPHandlers.h
```diff
@@ -40,9 +40,6 @@
 
 namespace org::apache::nifi::minifi::test {
 
-static std::atomic<int> transaction_id;
-static std::atomic<int> transaction_id_output;
```
fgerlits, line 44: This is also unrelated to ASan, but we used a local variable named `transaction_id` somewhere where this header was included, and clang-tidy was complaining about hiding a global variable. So I moved these two to the .cpp file.

libminifi/test/unit/FileStreamTests.cpp
```diff
@@ -124,36 +124,6 @@ TEST_CASE("TestFileBadArgumentNoChange2", "[TestLoader]") {
   REQUIRE(std::string(reinterpret_cast<char*>(data), verifybuffer.size()) == "tempFile");
 }
 
-TEST_CASE("TestFileBadArgumentNoChange3", "[TestLoader]") {
-  TestController testController;
-  auto path = testController.createTempDirectory() / "tstFile.ext";
```
fgerlits, line 129: I removed this test because now it is basically identical to the previous one, and gcc 15.2 did not like the `string` constructor at line 154. This test case used to test `stream.read(nullptr, 0)`, but we no longer support this pointer syntax, only `std::span`s.

extension-framework/include/utils/net/ConnectionHandler.h
```diff
@@ -64,7 +66,7 @@ class ConnectionHandler final : public ConnectionHandlerBase {
   [[nodiscard]] asio::awaitable<std::error_code> setupUsableSocket(asio::io_context& io_context) override;
   [[nodiscard]] bool hasUsableSocket() const { return socket_ && socket_->lowest_layer().is_open(); }
 
-  asio::awaitable<std::error_code> establishNewConnection(const asio::ip::tcp::resolver::results_type& endpoints, asio::io_context& io_context_);
+  asio::awaitable<std::error_code> establishNewConnection(const std::vector<asio::ip::tcp::endpoint>& endpoints, asio::io_context& io_context_);
```
szaszm: Why change this type? If we change it, why not go with something more generic, like std::span?

fgerlits:
successful run of the new memcheck job: https://github.com/fgerlits/nifi-minifi-cpp/actions/runs/35078415362

extension-framework/include/utils/net/ConnectionHandler.h
fgerlits: There was a leak in `ConnectionHandler<SocketType>::setupUsableSocket` (lines 147-163 in the same file) because the `resolve_result` did not get deleted properly. I don't completely understand why this happened(*), but the fix was to copy the endpoints from the `resolve_result` and let it go out of scope before the next `co_await` is called.

Using `span` instead of `vector` as the argument type is a good idea, done in 0dcbf880b53a9921b5bbd43f73c2e4240f816130.

(*) it may be related to https://gcc.gnu.org/bugzilla/show_bug.cgi?id=124584, as I have checked that it happens on GCC 15, 16, but doesn't happen on GCC 13, 14 and Clang 20.

martinzink (commented):
could we now perhaps expand the memchecked tag or maybe even remove the tag and run everything not just memchecked tagged in the valgrind ci?

otherwise lgtm 👍

fgerlits:
> could we now perhaps expand the memchecked tag or maybe even remove the tag and run everything not just memchecked tagged in the valgrind ci?
> 
> otherwise lgtm 👍

Good idea, but I'd prefer to put that in a separate PR, since it's not related to ASan, and it's likely to find a lot of new issues.  Jira: https://issues.apache.org/jira/browse/MINIFICPP-2905

martinzink (approved):

szaszm (approved):

cmake/Abseil.cmake
```diff
@@ -28,8 +28,8 @@ set(PC ${Bash_EXECUTABLE}  -c "set -x &&\
 
 FetchContent_Declare(
         absl
-        URL      https://github.com/abseil/abseil-cpp/archive/refs/tags/20260526.0.tar.gz
-        URL_HASH SHA256=6e1aee535473414164bf83e4ebc40240dec71a4701f8a642d906e95bea1aea0c
+        URL      https://github.com/abseil/abseil-cpp/archive/refs/tags/20260817.0.tar.gz
```
lordgamez, line 31: Did the older version have an issue that needed the update or was Abseil updated for another reason? I see that the Conan package version was not updated as Conan Center's version is not updated yet, so there will be a difference between the dependency versions of the package built from source and the one downloaded from Conan. I don't think that's a problem now, but we should make a note to update that one too when possible.

lordgamez (approved):

cmake/Abseil.cmake
fgerlits: There was a leak in the 20260526 version which is fixed in the 20260817 version, so I upgraded instead of patching.

We should review and upgrade both Conan and non-Conan dependency versions periodically. Maybe after each release, so there is time to find and fix any issues before the next release?

szaszm: In the age of AI security scanning, and lots of CVEs in each release, I think I'd time the updates closer to before a release, to minimize the number of known unpatched vulnerabilities in our libs.

PR #2261, opened 2026-09-11
MINIFICPP-2900 Fix leak of NodeID allocation in OPC processors

lordgamez:
Claude Code found a potential memory leak when allocating node id in OPC processors. In case of `translateBrowsePathsToNodeIdsRequest` calls or any other cases where `UA_NodeId` is allocated, and uses a node id type that is heap allocated like a string nodeid, the owner's responsibility is to free that node id using `UA_NodeId_clear` call when the id is no longer in use. This commit adds a RAII wrapper for the `UA_NodeId` to mitigate this issue. 

The PR additionally renames the OPC files to align with the filename conventions of the repository.

https://issues.apache.org/jira/browse/MINIFICPP-2900

-----------

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/extensions/opc/include/OPCCommon.h
+++ b/extensions/opc/include/OPCCommon.h
@@ -58,6 +58,43 @@ enum class OPCNodeDataType{
   String
 };
 
+// RAII owner for a UA_NodeId that calls UA_NodeId_clear to free node id allocation
+class NodeId {
+ public:
+  NodeId() = default;
+  explicit NodeId(UA_NodeId id) noexcept : id_(id) {}  // takes ownership of an already-built node id
+  NodeId(const NodeId&) = delete;
+  NodeId& operator=(const NodeId&) = delete;
+  NodeId(NodeId&& other) noexcept : id_(other.id_) { other.id_ = UA_NODEID_NULL; }
+  NodeId& operator=(NodeId&& other) noexcept {
+    if (this != &other) {
+      UA_NodeId_clear(&id_);
+      id_ = other.id_;
+      other.id_ = UA_NODEID_NULL;
+    }
+    return *this;
+  }
+  ~NodeId() { UA_NodeId_clear(&id_); }
+
+  static NodeId copyOf(const UA_NodeId& id) {
+    NodeId result;
+    UA_NodeId_copy(&id, &result.id_);
+    return result;
+  }
+
+  operator const UA_NodeId&() const noexcept { return id_; }  // NOLINT(google-explicit-constructor) implicit passthrough to the C API is intended
+  [[nodiscard]] const UA_NodeId& get() const noexcept { return id_; }
+
+  // Returns a pointer to the (cleared) node id for an open62541 out-parameter to write a freshly created node id into.
+  UA_NodeId* receive() noexcept {
+    UA_NodeId_clear(&id_);
+    return &id_;
+  }
+
+ private:
+  UA_NodeId id_ = UA_NODEID_NULL;
+};
+
 struct NodeData;
 
 class Client;
@@ -73,7 +110,7 @@ class Client {
   UA_ReferenceDescription * getNodeReference(UA_NodeId node_id);
   void traverse(UA_NodeId node_id, const std::function<NodeFoundCallBackFunc>& cb, const std::string& base_path = "", uint64_t max_depth = 0, bool fetch_root = true);
   bool exists(UA_NodeId node_id);
-  UA_StatusCode translateBrowsePathsToNodeIdsRequest(const std::string& path, std::vector<UA_NodeId>& found_node_ids, int32_t namespace_index,
+  UA_StatusCode translateBrowsePathsToNodeIdsRequest(const std::string& path, std::vector<NodeId>& found_node_ids, int32_t namespace_index,
     const std::vector<UA_UInt32>& path_reference_types, const std::shared_ptr<core::logging::Logger>& logger);
 
   template<typename T>
--- a/extensions/opc/include/OPCCommon.h
+++ b/extensions/opc/include/OPCCommon.h
@@ -74,7 +74,7 @@ class NodeId {
     }
     return *this;
   }
-  ~NodeId() { UA_NodeId_clear(&id_); }
+  ~NodeId() noexcept { UA_NodeId_clear(&id_); }
 
   static NodeId copyOf(const UA_NodeId& id) {
     NodeId result;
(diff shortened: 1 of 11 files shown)


szaszm:
Can you have Claude Code do a thorough review too, before we waste human reviewer time on trivial issues? It's giving better results than Copilot usually does.

The first things I've noticed:
- No gsl::owner annotation when taking ownership of the node id
- Destructor should be noexcept

lordgamez:
> Can you have Claude Code do a thorough review too, before we waste human reviewer time on trivial issues? It's giving better results than Copilot usually does.
> 
> The first things I've noticed:
> 
>     * No gsl::owner annotation when taking ownership of the node id
> 
>     * Destructor should be noexcept

I ran Claude Code to review the commit, but it did not find any issues that was relevant to the commit and needs to be addressed. 
I added noexcept to the destructor in https://github.com/apache/nifi-minifi-cpp/pull/2261/commits/7cbb097173228045e2e26afb43104153e138a5ee
`gsl::owner` is not applicable here as the UA_NodeId struct itself is stack allocated and the heap allocation depends on the actual type in its union member. The ownership logic is implemented in the UA_* types.
The added test confirms the memory leak issue (failing without the fix and passing with the fix) when ran through valgrind as usually only the UA string node id type is heap allocated.

szaszm:
Since the clear function operates with UA_NodeId, I would treat the internals as a black box for the purpose of cleanup, and always call UA_NodeId_clear on it, regardless of the union values. I'd treat UA_NodeId as a resource handle and UA_NodeId_clear its corresponding resource cleanup function, and resource handles can be owning or non-owning, that's why I think we should use gsl::owner.

lordgamez:
> Since the clear function operates with UA_NodeId, I would treat the internals as a black box for the purpose of cleanup, and always call UA_NodeId_clear on it, regardless of the union values. I'd treat UA_NodeId as a resource handle and UA_NodeId_clear its corresponding resource cleanup function, and resource handles can be owning or non-owning, that's why I think we should use gsl::owner.

I agree the internals should be treated as a black box. One of the reasons for this PR is that we had previously not tested node IDs with string-based underlying types, which is why the issue didn't show up in our memchecks. The fix now handles all union type cases the same way, clearing them every time, even though for integer node ID types the clear operation is a no-op.

That said, I still don't understand how gsl::owner could be used here. In the NodeId RAII wrapper, ownership is taken at construction and held until the end of its lifetime, so ownership never changes over the course of this solution. Also, the `id_` member is of type `UA_NodeId`, not `UA_NodeId*`, and `gsl::owner` can only be applied to a pointer type. In every function where we actually return a pointer, we deliberately return a non-owning one, since `NodeId` retains ownership of the `id_` member itself. Maybe I'm missing something, could you elaborate on how you pictured using `gsl::owner` here?

szaszm:
Ownership of UA_NodeId is transferred to the RAII wrapper in its constructor, so I'd annotate the constructor parameter and the data member as owner.

> Also, the id_ member is of type UA_NodeId, not UA_NodeId*, and gsl::owner can only be applied to a pointer type.

I wasn't sure about this restriction, but if gsl::owner only applies to pointer types, then indeed we can't use it for this, and a comment is the best we can do.

lordgamez:
> Ownership of UA_NodeId is transferred to the RAII wrapper in its constructor, so I'd annotate the constructor parameter and the data member as owner.
> 
> > Also, the id_ member is of type UA_NodeId, not UA_NodeId*, and gsl::owner can only be applied to a pointer type.
> 
> I wasn't sure about this restriction, but if gsl::owner only applies to pointer types, then indeed we can't use it for this, and a comment is the best we can do.

I see now, that was the source of my confusion. Making it at construction would make sense, but when I checked the definition I saw that it can only be applied to a pointer, so I didn't understand where we would use it, that's why I pointed out previously that the `UA_NodeId` member itself is stack allocated. We can add a comment if nothing else.

```
template <class T, std::enable_if_t<std::is_pointer<T>::value, bool> = true>
using owner = T;
```

extensions/opc/include/OPCCommon.h
```diff
@@ -58,6 +58,43 @@ enum class OPCNodeDataType{
...
+  }
+
+  operator const UA_NodeId&() const noexcept { return id_; }  // NOLINT(google-explicit-constructor) implicit passthrough to the C API is intended
+  [[nodiscard]] const UA_NodeId& get() const noexcept { return id_; }
+
+  // Returns a pointer to the (cleared) node id for an open62541 out-parameter to write a freshly created node id into.
+  UA_NodeId* receive() noexcept {
+    UA_NodeId_clear(&id_);
+    return &id_;
+  }
```
szaszm, line 92: what if the write fails, do we handle those error cases correctly? wouldn't it be cleaner to keep the out param local to the call, then wrap the result manually in this NodeId wrapper, and move-assign to existing wrapper objects as necessary?

extensions/opc/include/OPCCommon.h
```diff
@@ -58,6 +58,43 @@ enum class OPCNodeDataType{
...
+  NodeId(NodeId&& other) noexcept : id_(other.id_) { other.id_ = UA_NODEID_NULL; }
+  NodeId& operator=(NodeId&& other) noexcept {
+    if (this != &other) {
+      UA_NodeId_clear(&id_);
+      id_ = other.id_;
+      other.id_ = UA_NODEID_NULL;
+    }
+    return *this;
+  }
+  ~NodeId() noexcept { UA_NodeId_clear(&id_); }
```
szaszm, line 77: Is this valid for UA_NODEID_NULL or already cleared objects?

lordgamez: Yes, UA_NODEID_NULL is a numeric node id with 0 namespace and 0 id, when UA_NodeId_clear is called that is what the id is set to after the free operation. The UA_NodeId_clear operation is no-op for all numeric node ids, so nothing is done when it is called for an already cleared object.

extensions/opc/include/OPCCommon.h
lordgamez: This is already handled in the OPC UA client code in `src/client/ua_client_highlevel.c`. We are using this value as the out parameter of the add node operation, and the add node only touches the out parameter if the internal call succeeds otherwise it remains untouched:
```
/* Move the id of the created node */
retval = response.results[0].statusCode;
if(retval == UA_STATUSCODE_GOOD && outNewNodeId) {
    *outNewNodeId = response.results[0].addedNodeId;
    UA_NodeId_init(&response.results[0].addedNodeId);
}
```

martinzink (approved):

szaszm (approved):

fgerlits (approved):

